PR #204, opened 2014-11-22
Fix error not being shown when failing to verify SMS code

dmdeller:
Hi there,

I was trying to build and run the textSecure branch on my iPhone, and I couldn't get past the verify SMS code step. When I pressed the 'Verify' button, nothing happened. I dug into it and found a TODO for error handling, so I figured I may as well give that a try.

I ended up making some changes to TSAccountManager to return an NSError instead of a typedef'd enum (TSRegistrationFailure). This is similar to how Cocoa, AFNetworking, etc. bubble up errors (and is recommended by the GitHub style guide :)). This approach allows me to handle failures from various different places in the same way, as long as they all return an NSError.

I would suggest similar changes to PushManager - in particular, showing a UIAlertView from somewhere other than a view controller is discouraged, and won't be possible after migrating to the new UIAlertController. Taking it further, the approach of using NSErrors for network failures could probably be done throughout the app (maybe it already is - I haven't looked thoroughly). But I didn't want to get too deep into a refactoring without seeing what you guys think first.

Finally, I was going to try adding some more strings for additional error codes/conditions, but I wasn't sure how localization is handled in this project (adding English text before other translations are available?), so I thought it best to leave it alone for now.

Ultimately, I still can't proceed past the SMS code verification step (I think because I can't connect to textsecure-service-staging.whispersystems.org) but at least I get a nice error alert now :)

### DIFF
--- a/Signal/src/textsecure/Account/TSAccountManager.h
+++ b/Signal/src/textsecure/Account/TSAccountManager.h
@@ -9,6 +9,9 @@
 #import <Foundation/Foundation.h>
 #import "TSConstants.h"
 
+static NSString *const TSRegistrationErrorDomain = @"TSRegistrationErrorDomain";
+static NSString *const TSRegistrationErrorUserInfoHTTPStatus = @"TSHTTPStatus";
+
 typedef NS_ENUM(NSUInteger, TSRegistrationFailure) {
     kTSRegistrationFailureAuthentication,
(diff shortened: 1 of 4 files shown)


stuartcw:
Is this the same problem where you can't verify unless you have all notifications on? I have tried so many times to verify until today where I read a thread saying it was a known bug. After enabling them I got in straight away...

Issue #249, opened 2014-12-29
Group avatar fixes


FredericJacobs:
- [ ] Avatars (TSAttachmentStreams) should have a YapDatabase relationship to the group they belong to, and be deleted when the avatar is changed or the group is deleted.
- [x] Group avatars should be properly compressed/resized.

corbett:
@FredericJacobs  I am not as adept in the YapDatabase relationship regime yet to be confident I will do the right thing for part a) of this issue. if it seems easy to you and/or you can give me a TLDR of how this works that would be helpful. otherwise I will dive in soon.

corbett:
working on a) of this issue now.

FredericJacobs:
You can see the `TSInteraction` relationship to a thread.

I always write a test to see if it works like I expect it to: https://github.com/WhisperSystems/Signal-iOS/blob/master/Signal/test/textsecure/TSMessageStorageTests.m#L74

corbett:
cool yes I see that, I was more basing my work on the TSAttachmentStream setup. I'm unclear the purpose of the structs... e.g.

extern const struct TSMessageRelationships {
    __unsafe_unretained NSString *threadUniqueId;
} TSMessageRelationships;

FredericJacobs:
Uhm, it's not really necessary in this case. Don't bother doing a struct.

corbett:
yes, that's what I concluded.

FredericJacobs:
In this case: 
`TSAttachmentStream < ---- RELATIONSHIP --- > TSGroupThread`

corbett:
I took a more of a glance at this today. actually the group avatar is
stored as a UIImage blob in the GroupModel. I wrote a test case and it is
correctly deleted. now whether we should actually store the group avatars
as tsattachmentstreams instead is another question.

On Sun, Jan 4, 2015 at 11:56 PM, Frederic Jacobs <notifications@github.com>
wrote:

> In this case:
> TSAttachmentStream < ---- RELATIONSHIP --- > TSGroupThread
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/WhisperSystems/Signal-iOS/issues/249#issuecomment-68655709>
> .
>



-- 
Dr. Corbett Moran
Rockets @ http://www.spacex.com
Physics @ http://www.ICS.uzh.ch  <http://www.ICS.uzh.ch>
Dev @ http://circleof6app.com
Dev @ https://whispersystems.org
[...]

PR #256, opened 2015-01-01
Group avatar fixes pr

corbett:
• refactor to put a method resize with quality in a catagory used elsewhere in the code
•  -03 optimization only for release/ad hoc but not debug (to play well with debugger)
• resizes group avatars to max 100 pixels keeping aspect ratio.

addresses 1/2 of https://github.com/WhisperSystems/Signal-iOS/issues/249


FredericJacobs:
Ok, so I merged this.

I edited the commits to:
- fix the group avatar bug where it was sent as an attachment
- replace the image resize pod by the category itself in the code. Pods are awesome for dependencies that evolve quickly over time and that do provide big functionality. In this case, adding a Pod dependency for a single method isn't worth it IMO because it's really annoying to audit pod's security.

corbett:
OK, yes that makes sense. I agree that we don't want too many pods. I
however think that the resize functionality may be something we often
deploy and this one was simple. that said, this sounds fine.

On Fri, Jan 2, 2015 at 4:27 PM, Frederic Jacobs <notifications@github.com>
wrote:

> Closed #256 <https://github.com/WhisperSystems/Signal-iOS/pull/256>.
>
> —
> <https://github.com/WhisperSystems/Signal-iOS/pull/256#event-213476950>.
>



-- 
Dr. Corbett Moran
Dev @ http://circleof6app.com
www.christinecorbettmoran.com

PR #257, opened 2015-01-02
more descriptive info message strings

corbett:
• more descriptive info message strings on group update
• UI of info messages doesn't display multiliners well, but the info string will be stored for use by future UI improvments.


FredericJacobs:
Looks good to me, well done :+1:

corbett:
OK, we will need to change the UI of the TSInfoMessages now to potentially display longer messages so some more elaborate updates don't result in ... (which should be changed anyway as the strings aren't guaranteed to be one line).

PR #248, opened 2014-12-29
Groups bugfixes

corbett:
1,2,5,7,8,9,10,11 fixed/wontfix/notes
bugfixes for groups-bugfixes (remaining: synchronicity issues 3,4,6)
FIXED 1) Resign first responder in new group view when tapping camera icon. Otherwise, keyboard hides action sheet choices.
FIXED 2) Have a list of participants UI
3) Non-blocking UI on group create with attachment
4) I created a group on iOS, name doesn't show up on Android, neither does the avatar.
-will be fixed by proper treatment of 3)
FIXED 5) I shouldn't be able to remove myself from the participants list.
6) Sending is terribly laggy
-will be fixed by 3)
WONTFIX 7) Message should be Group Created and not Group updated if new group.
-this is not the case on android! I am punting this for now and we can fix when we do the localized strings.
FIXED 8) Leaving group doesn't appear to be working (tried with android group)
FIXED (with caveats) 9) Receiving messages in deleted group causes weird messageviewcontroller to show up
-ignore messages from a group you have left according to your understanding? what should the correct behavior be here? doing that for now. 
-tableViewCellTappedDelete (the fact that the group model lives in a thread the group model goes away when the thread does!) this will cause a major refactoring, unfortunately to be changed. for now this causes a group leave event and then deletes the thread, which makes sense to the user. not sure what android does here. message leave received. 
-seems like we should actually store the groups separately (forever)! and then we will no if a user leaves, but is later readded! otherwise a group update message is sent... not a new group message. and without a group to compare to what should we do. 
WONTFIX 10) Avatar Image doesn't have reference-counting and will never be deleted, even if group is.
I am punting this for now. we should open an issue. generically this issue should also make sure the group avatar is small. 
FIXED 11) Android user leaving group is causing a "Name changed" info message

PR #258, opened 2015-01-02
group messaging ack

corbett:
• DO NOT MERGE
• this is only a sketch of what would need to be changed to ack only in the following situations:
-not a group message
-Is a group update message
-I have heard about the group (i.e. previously have received an update message about it and haven't left it)
• we can see that the code is quite ugly. we have to go all the way through encryption prior to the ack. I think independent of security issues this approach is not clean
• this is just meant to be a point for discussions. if we end up going this route I will revisit this branch. also untested just a sketch.

### DIFF
--- a/Signal/src/textsecure/Messages/TSMessagesManager.h
+++ b/Signal/src/textsecure/Messages/TSMessagesManager.h
@@ -17,10 +17,11 @@
 
 @property (readonly) YapDatabaseConnection *dbConnection;
 
-- (void)handleMessageSignal:(IncomingPushMessageSignal*)messageSignal;
+// returns yes if this message should be acked
+- (BOOL)handleMessageSignal:(IncomingPushMessageSignal*)messageSignal;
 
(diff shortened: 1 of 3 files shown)


FredericJacobs:
I'll leave the last word to @moxie0 (See discussion in https://github.com/WhisperSystems/Signal-iOS/pull/248) but I don't think that this is the right way to go.

corbett:
cool, yea me neither if there is another way. just wanted to flesh out the code changes necessary for this way. I will write a message to Moxie cc you and write the droids/bros as needed.

FredericJacobs:
@corbett: I think this can be closed, right?

corbett:
yes

On Sat, Jan 3, 2015 at 1:10 PM, Frederic Jacobs <notifications@github.com>
wrote:

> @corbett <https://github.com/corbett>: I think this can be closed, right?
>
> —
> <https://github.com/WhisperSystems/Signal-iOS/pull/258#issuecomment-68594457>
> .
>



-- 
Dr. Corbett Moran
Dev @ http://circleof6app.com
www.christinecorbettmoran.com

PR #265, opened 2015-01-04
Some subtle changes in tone

LotP:
Using a more formal language and avoid to "speak" to the user directly, which is unusual to the German user.
Sie / Du corrected, some unusual language cleared. 
Still some inconsistencies.


FredericJacobs:
Hi @LotP,

Thanks for the contribution but localizations are fetched in version control before every release. We do use Transifex for translations:
https://www.transifex.com/projects/p/signal-ios/

Issue #233, opened 2014-12-23
Improved group management + group fingerprint UX


FredericJacobs:

corbett:
can we discuss what improvements are meant by this issue ?

FredericJacobs:
I can think of:
1) [ ] When adding participants to a group, "Add people" can be selected and crashes the app.
2) [ ] Tapping a group member in the participants list should open a private conversation with that member and I should be able to verify key from there.
3) [ ]  Group update messages are not working as intended. Messages regarding members leaving or joining are not displayed, only "Group updated"
4) [ ] Image should not be changed if I edit the group name or other attributes.

corbett:
ok... great yes these make sense...

FredericJacobs:
I can also think of 
[ ] If no image selected, don't upload the UI placeholder blue camera icon.

corbett:
there were two issues with "Group update messages are not working as intended."
a) the code which did this (caveat: that the TSInfoMessage cells do not expanding to display the message being JSQDisplayedMessageCollectionViews) somehow disappeared from master probably in a bad merge. I have redone a PR for that: https://github.com/WhisperSystems/Signal-iOS/pull/300
b) re:caveat, to change the info messages we will have to go into ```JSQMessagesViewController.h```/```JSQDisplayedMessageCollectionViewCell.xib``` internals. I am loathe to do that but I see @dtsbourg has done before. an alternative would be not to rely on this message class. 
thoughts?

PR #267, opened 2015-01-04
Group management 233 pr

corbett:
• disabling selection of "Add People"/"Recipients" and associated failure modes
• brings up contact information sheet on click from group
• addresses several points in #233


FredericJacobs:
All good.

PR #269, opened 2015-01-04
Group avatar 233

corbett:
• don't upload group avatar unless user picked one
• display default of group avatar if none chosen in message thread
further addresses issues in #233


FredericJacobs:
All good :+1:

PR #286, opened 2015-01-06
End error message sentences with a dot

mxmerz:
All info messages in ``TSInfoMessage.m`` and two of the error messages in ``TSErrorMessage.m`` already ended in a dot. Also, all sentences should end in a dot. Therefore, the other error messages should also end in a dot.

And yes, I know that this is pedantic. And I’m sorry.


FredericJacobs:
Thanks @mxmerz. Better consistency now.

PR #301, opened 2015-01-07
default avatar for contact

### LINKED ISSUE #282: Conversations without a image should have a default avatar
The empty white circle feels like a failed load.

corbett:
fixes #282; note that the default avatar for the group and the individual case were what I found in the repo we'll want to have @abolishme or whoever create or find better assets for that.


FredericJacobs:
I don't think #282 was a bug. It was a feature in Signal before. We always had empty circles and it was part of the design. Better check in with Tyler about that.

corbett:
In any case we can replace with an empty image as needed, with the existing code from this pr. If you type in dummy_image.png it will be a blank circle again (as the image will not be found) or we can set it to nil.

corbett:
Fred, it isn't a bug, there should be a default avatar. It won't be this one ( a standin) but I'll still want this code.

abolishme:
something like this is what i suggest we do for the empty avatar states:

https://github.com/ge0ff/CRNInitialsImageView

corbett:
SGTM if Fred is on board. what about groups? and we'll have issues for that
right?

On Mon, Jan 12, 2015 at 12:40 AM, Tyler Reinhard <notifications@github.com>
wrote:

> something like this is what i suggest we do for the empty avatar states:
>
> https://github.com/ge0ff/CRNInitialsImageView
>
> —
> <https://github.com/WhisperSystems/Signal-iOS/pull/301#issuecomment-69552576>
> .
>



-- 
Dr. Corbett Moran
Dev @ http://circleof6app.com
www.christinecorbettmoran.com

FredericJacobs:
I suggest we use a placeholder for groups with no avatar and CRNInitialsImageViews for contact threads without picture.

abolishme:
I agree, @FredericJacobs

dtsbourg:
No need to add dependencies, JSQMessages' ```AvatarImageFactory``` can do the job : 
[JSQMessagesAvatarImageFactory](
https://github.com/jessesquires/JSQMessagesViewController/blob/7ef5b4bbc2f51eb636a00e867c676b8ceb369f68/JSQMessagesViewController/Factories/JSQMessagesAvatarImageFactory.m)

corbett:
superceded by #339

PR #341, opened 2015-01-14
removal of dead code and tuning of archive time

corbett:


Signal/src/call/RecentCallManager.m
```diff
@@ -79,7 +79,7 @@ - (void)addRecentCall:(RecentCall*)recentCall {
         uint64_t callDateSeconds = (uint64_t)[recentCall.date timeIntervalSince1970];
         TSCall *call = [[TSCall alloc] initWithTimestamp:callDateSeconds*1000 withCallNumber:recentCall.phoneNumber.toE164 callType:recentCall.callType inThread:thread];
         if(recentCall.isArchived) { //for migration only from Signal versions with RedPhone only
-            thread.archivalDate = [NSDate dateWithTimeIntervalSince1970:(callDateSeconds+10000)];
+            thread.archivalDate = [NSDate dateWithTimeIntervalSince1970:(callDateSeconds+60)]; // archive has to happen in the future of the original call
```
corbett: 1,5,20,30 don't work, 60 does, the comparison is perhaps comparing against a date with the granularity of a minute, so we add a minute in the future from the original call to make sure the call is archived. this would only go awry if the user:
-makes a call to Bob from Signal < 2 
-archives said call to Bob
-upgrades to Signal 2.0
-gets a message from Bob
in less than a minute, that is all that has to happen in the space of 60 seconds. In that case the thread will remain archived (the message is still registered in the thread just in the archive tab) until a message from Bob is received and more than a minute has passed since the original call. 

I think this case is very edge and I'm OK with it.

corbett:
now ready for review @FredericJacobs

FredericJacobs:
LGTM

PR #350, opened 2015-01-15
 1.0 constraints

abolishme:


corbett:
PR needs to be into design

PR #373, opened 2015-01-16
2.0 1.0c1 countrycode

corbett:


corbett:
- [ ] test on device for iPhone 5/6/6+
- [ ] search crashing as we are missing CountryCodeTableViewCell.xib (should figure out how to stylize within storyboard or reintroduce this file)

PR #390, opened 2015-01-18
average image quality

joyceyan:
image quality is average by default instead of checking if the user has configured a preference for image quality. i've also removed ability to configure image quality in the settings page.


corbett:
tested on device, looks good.

PR #391, opened 2015-01-19
search by contact's name or number

joyceyan:
When composing a new message, you can search through your contacts by searching by your contact's name or by their phone number(s). Since phone numbers are stored in E164, we can strip the search text of illegal characters like parentheses and spaces, and then see if the search text is "contained" in the E164 formatted phone numbers of their contacts.


corbett:
I will merge this for now, better will be to format the search string as e164 and then compare the two.

PR #395, opened 2015-01-19
to unknown contact sends install signal link via SMS PR 1/2

joyceyan:
In the messages composer, users can search within their contacts by their name or by their phone number. Now, messaging numbers not in the "contacts" book is enabled. If a user searches for a phone number that has >7 digits in it, then we can fire up iMessage to this number with a link to install Signal (link not inputted yet) in the message.

### DIFF
--- a/Signal/src/view controllers/MessageComposeTableViewController.m
+++ b/Signal/src/view controllers/MessageComposeTableViewController.m
@@ -12,10 +12,15 @@
 #import "MessagesViewController.h"
 #import "SignalsViewController.h"
 
+#import <MessageUI/MFMessageComposeViewController.h>
+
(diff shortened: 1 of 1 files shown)


corbett:
Issues:
-check out PhoneNumberUtil tryParsePhoneNumberFromUserSpecifiedText and bestEffortFormatPartialUserSpecifiedTextToLookLikeAPhoneNumber to make sure that the phone number works with or without a country code specified (you can tell if there is a country code because it starts with a +) and formats nicely
-ask the user for confirmation that they are OK to launch a regular insecure message

corbett:
also missing a cancel button on the sms, but I can fix that if we fix everything else.

PR #406, opened 2015-01-19
Moved and styled settings

abolishme:
two notes: the destroy account button needs a confirmation modal on tap, and i the settings page won't launch in this commit because i moved the table view section order and apparently the view controller has hardcoded their locations.


corbett:
will be done in separate pr

PR #413, opened 2015-01-20
Implement sms delegate + polishing regular SMS invites

joyceyan:
I finished the issues that persisted since my last PR.

Changes made:
- MFMessageComposeViewControllerDelegate delegate implemented, so the inability to cancel out of an iMessage screen should be fixed
- plus sign removed from when a user is typing
- clarified that users send the text to their friend explicitly to invite them to use Signal; not meant to replace the main SMS app
- included link to install Signal in the message


joyceyan:
accidentally sent this PR to master, see PR #414 which is a PR to wboc15

Issue #439, opened 2015-01-21
99 issues for 2.0 


corbett:
## @corbett and @FredericJacobs 
- [ ] #370  Set "Conversation Safe Words" label in active conversation view
- [x] center the update/leave/members vertically on group convo dropdown
- [x] light gray hairline at top of inbox/archive switcher.
- [ ] light gray hairline at the bottom of the contact details dropdown.
- [ ] left transition for settings (some prototyping in reverse_settings_transition where I create a nave stack with the settings artificially at base)
- [x] text and image together (android has capability of sending)
- [x] not use directories backed up in iCloud for attachment storage
- [x] New Message view needs way to be refreshed when placeholder is displayed.

## @abolishme
*waiting for tasks*

## Completed
- [x] App Store-sized Icon + vectorial icon in repo TOP PRIORITY
- [x] tap and hold "my fingerprint" button for five seconds just does "abort secure session" ... not "shred", which was previously the red button on the fingerprint screen. remove red button. (ask moxie for details). if user wants to clear history, they delete the conversation in the inbox/archive view. (REMOVED)
- [x] improve twitter link image
- [x] constraints on the invisible buttons on the fingerprintview controller are off (check by setting the background to something non clear). TOP PRIORITY
- [x] on 6Plus bottom inbox/archive bar not constrained on left MEDIUM PRIORITY
- [x] delete button in inbox slightly poorly constrained on 6Plus (bleeds in) MEDIUM PRIOIRTY
- [x] constraint on table view cell, would be nice to have a longer preview message string (cuts off halfway through due to constraints)
- [x] review status bar changes and make suggestions (colors, size?) 
- [x] paperclip for attach should be ows_ blue (need an icon for this then that is that color)
- [x] in the build new conversation table, when you tap the phone icon, it should automatically launch a new call, and add the conversation to the inbox in the background (this should be a button before I hook it up) 
- [x] crash on search country code
- [x] in add new message hide phone if user not redphone user. (this should be a button before I hook it up)
- [x] BUG inbox on first launch doesn't update names (unless you select archive/inbox then updates) doing this selection in code doesn't work. mysterious yapdb related bug introduced. (@fredericjacobs will fix)
- [x] contact details dropdown should have translucent "vibrancy".
- [x] calls updates in JSQ styling. 
- [x] group updates in JSQ styling. 
- [x] crash on search country code
- [x] fix identity key tap to verify
- [x] image full screen view no longer has save button
[...]

Issue #312, opened 2015-01-10
Swiping left on a conversation deletes without user confirmation


JohanDoe:
In the overview of conversations if you swipe left, you see a red cross appear on the right, suggesting you should press this cross to confirm the deletion of the conversation.

However, if you swipe left 'too much' the conversation just disappears without the possibility to confirm. If you accidentally swiped a certain conversation, it is gone. 

I suggest that swiping left always allows the user to confirm the possible deletion (i.e. to press the red cross), regardless of how forceful the swipe to the left is.

abolishme:
this enhancement will be made!

JohanDoe:
Nice! Tx and best.

FredericJacobs:
Fixed and implemented.

Issue #303, opened 2015-01-07
Image description is missing


rmueller83:
If I send an image from TextSecure Android with an additional text, the text is not displayed on iOS Signal.

![ts1](https://cloud.githubusercontent.com/assets/2698076/5650656/b0f1d75e-96a0-11e4-8a3e-1470efd6f4c9.png)

![ts2](https://cloud.githubusercontent.com/assets/2698076/5650658/b544adae-96a0-11e4-9af0-2e1d3f19b31b.PNG)

corbett:
thanks, yes this is a known issue on our end, I will do some work to fix
it.

On Wed, Jan 7, 2015 at 6:10 PM, rmueller83 <notifications@github.com> wrote:

> If I send an image from TextSecure Android with an additional text, the
> text is not displayed on iOS Signal.
>
> [image: ts1]
> <https://cloud.githubusercontent.com/assets/2698076/5650656/b0f1d75e-96a0-11e4-8a3e-1470efd6f4c9.png>
>
> [image: ts2]
> <https://cloud.githubusercontent.com/assets/2698076/5650658/b544adae-96a0-11e4-9af0-2e1d3f19b31b.PNG>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/WhisperSystems/Signal-iOS/issues/303>.
>



-- 
Dr. Corbett Moran
Rockets @ http://www.spacex.com
Physics @ http://www.ICS.uzh.ch  <http://www.ICS.uzh.ch>
Dev @ http://circleof6app.com
Dev @ https://whispersystems.org
www.christinecorbettmoran.com

corbett:
@jackflips

PR #456, opened 2015-01-22
viewing audio messages implemented

joyceyan:
things changed
- pbxproj modified because play/pause icon resources were added in
- videoURL renamed to mediaURL, we'll try to go through more thoroughly to rename things accordingly
- moved around some methods in MessagesViewController.m so all the audio methods are grouped together
- audio progress bar displayed for each audio message

//TODO
- displaying audio duration not working (initializeAudioBubbles)
- UI of playing audio breaks when the application goes in background and you open up the app again


corbett:
hey... this causes persistent crashes, so I can't merge it at this stage. that said could you please be sure not to delete your original fork/branch as I'll use as prototyping work to get me started when we get back to putting audio as a priority soon. then we can merge this all in with the TODOs done. suggested issues #312 #303, fixing crash on searching for country in country code selector, fixing search bar off issue in message compose.

corbett:
fixing search bar off issue has been done. the updated great issue would be to style TSInfoMessage for group updates, phone calls as tyler desires or creating new subclasses in the fork of JSQ that do so. styel for that is mentioned here: 
https://github.com/WhisperSystems/Signal-iOS/issues/439

corbett:
if you want to change the delete maybe look into using
https://github.com/MortimerGoro/MGSwipeTableCell

PR #457, opened 2015-01-22
Remove audio

corbett:
audio support isn't there yet, and may be punted for launch. @joyceyan and @jackflips did some great prototyping work in https://github.com/WhisperSystems/Signal-iOS/pull/456 for future support.


corbett:
wrong pr...

Issue #465, opened 2015-01-22
Deleting individual messages causes a crash


jackflips:

FredericJacobs:
Fixed

Issue #398, opened 2015-01-19
Shredding all communication caused a crash


jackflips:
'Cannot fetch message because row 0 is >= numberOfItemsInSection 0'

corbett:
good bug!

PR #469, opened 2015-01-22
deleting single messages won't crash app

jackflips:
see [#465](https://github.com/WhisperSystems/Signal-iOS/issues/465).


jackflips:
and [#398](https://github.com/WhisperSystems/Signal-iOS/issues/398)

FredericJacobs:
Caches always causing crashes!

Reviewing this now.

jackflips:
Of course you're right. This should be better!

PR #484, opened 2015-01-23
can see leftover duration of audio messages

joyceyan:
Earlier I tried putting in the duration of each audio message, but found this was non-trivial. Even though that's difficult, I realized it's pretty easy to let the user see how much time is left on each audio message even if they can't see how long the message is before playing it.


joyceyan:
Changes here were included in PR #487

PR #487, opened 2015-01-24
Alert before deleting a thread + show remaining duration of audio message left

joyceyan:
Before swiping to delete a thread, a UIAlertView pops up and warns the user. They can choose to cancel or proceed with permanent deletion of the thread. The wording could maybe be improved, but currently the title of the alert is "Warning" and the message is "Are you sure you want to delete this entire thread permanently?"

The minor changes to audio-related stuff is just letting the user know how much time is left in the audio message they're currently playing.

### DIFF
--- a/Signal/src/view controllers/InboxTableViewCell.h
+++ b/Signal/src/view controllers/InboxTableViewCell.h
@@ -24,7 +24,7 @@ typedef enum : NSUInteger {
 
 @end
 
-@interface InboxTableViewCell : UITableViewCell  <UIScrollViewDelegate>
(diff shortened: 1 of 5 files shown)


corbett:
@joyceyan can you separate out the audio changes in a separate PR, in general one PR per feature. unfortunately this isn't the designed behavior. we should be minimizing alerts. the delete should function as on iOS design, where the user taps on the red box to delete.

joyceyan:
audio changes can be seen in PR #484 which i've now re-opened

corbett:
@joyceyan thanks I'll review that now!

PR #513, opened 2015-01-26
fixes UX bug with compose from archive

corbett:
where user clicks compose and selects contact or group from the archive on returning the archive is still selected.


FredericJacobs:
Merged as ab6dc0d989595e00aa983f7370fa3008bd778b94

PR #514, opened 2015-01-26
Message id fixes

FredericJacobs:


corbett:
this looks good, tested on device as well. in general I prefer to review PRs separately for logical changes vs accumulated mega rebases from multiple authors. on board with rebasing before final merges.

PR #527, opened 2015-01-29
Archive empty simple

corbett:
• can archive empty thread
• added creation date without messing up thread order
• switch to inbox on message compose, so user on back button sees what they have just created

Please test extensively prior to merge, as this has caused issues in the past with thread ordering.


FredericJacobs:
Merged as 553a38288ec2495aab6d7acdb24870ced68ed49b :+1:

PR #532, opened 2015-01-29
fixed issue with nav bar in message view no longer being tappable after ...

corbett:
...modal presented


FredericJacobs:
Merged as 499cdfa56f5d17c73a1e463c5d65c1db4e1f788e

PR #533, opened 2015-01-30
Fix for group avatar on create/update from iOS. Previously broken.

corbett:
• new avatar and updated avatar functional
• rounded corners in place, both in new/update group view and in inbox


FredericJacobs:
Merged as 797492fc130001e67b36612bf78db434ca462455

PR #537, opened 2015-01-30
Pinning cert and not public keys

FredericJacobs:


Signal/src/network/http/RPServerRequestsManager.m
```diff
@@ -36,7 +36,7 @@ - (id)init{
         NSURL *endPointURL = [NSURL URLWithString:[NSString stringWithFormat:@"https://%@:%hu", endpoint.hostname, endpoint.port]];
         NSURLSessionConfiguration *sessionConf = NSURLSessionConfiguration.defaultSessionConfiguration;
         self.operationManager = [[AFHTTPSessionManager alloc] initWithBaseURL:endPointURL sessionConfiguration:sessionConf];
-        self.operationManager.securityPolicy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModeNone];
+        self.operationManager.securityPolicy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModePublicKey];
         self.operationManager.securityPolicy.allowInvalidCertificates = YES;
```
corbett: why do we want allowInvalidCertificates=YES?

FredericJacobs: Yes, invalidCertificates means that they are not part of the trust store (not signed by a OS-approved CA).

corbett: OK! rest of this makes sense. maybe wait and see if Square merges in if that makes sense too.

PR #550, opened 2015-01-31
• upgrade from existing beta

corbett:
• update to upgrade from app on store to test

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -21,7 +21,7 @@
 #import "UIColor+OWS.h"
 #import "CodeVerificationViewController.h"
 #import "MIMETypeUtil.h"
-
+#import "TSDatabaseView.h"
 #import <PastelogKit/Pastelog.h>
 
 #define kSignalVersionKey @"SignalUpdateVersionKey"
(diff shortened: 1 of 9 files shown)


Signal/src/environment/VersionMigrations.m
```diff
@@ -16,6 +16,8 @@
 #import "RecentCallManager.h"
 #import "SignalKeyingStorage.h"
 #import "UICKeyChainStore.h"
+#import "TSStorageManager.h"
+#import "TSDatabaseView.h"
```
corbett: +#import "TSDatabaseView.h" is not needed

Signal/src/view controllers/UITests/SignalsViewController.m
```diff
@@ -88,7 +88,7 @@ -(void)viewWillAppear:(BOOL)animated
 - (void)viewDidAppear:(BOOL)animated{
     [super viewDidAppear:animated];
     
-    if (![TSAccountManager isRegistered]){
+    if (![TSAccountManager isRegistered] && ![Environment.preferences getIsMigratingToVersion2Dot0]){
```
corbett: this was one place that the migration was broke; when the tab view controller was removed this became the first view and then would display the signup flow as migration was in progress!

Signal/src/AppDelegate.m
```diff
@@ -90,13 +101,15 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
     self.notificationTracker = [NotificationTracker notificationTracker];
     
-    [self performUpdateCheck];
     CategorizingLogger* logger = [CategorizingLogger categorizingLogger];
     [logger addLoggingCallback:^(NSString *category, id details, NSUInteger index) {}];
     [Environment setCurrent:[Release releaseEnvironmentWithLogging:logger]];
     [Environment.getCurrent.phoneDirectoryManager startUntilCancelled:nil];
     [Environment.getCurrent.contactsManager doAfterEnvironmentInitSetup];
     
+    [self performUpdateCheck]; // this call must be made after environment has been initialized because in general upgrade may depend on environment
```
corbett: this was a second place the existing migration had broken, performUpdateCheck must be called after we have an environment

Signal/src/AppDelegate.m
```diff
@@ -49,22 +49,33 @@ @implementation AppDelegate
...
     NSString *currentVersion = [Environment.preferences setAndGetCurrentVersion];
+    // TODO: remove
     if (!previousVersion) {
         DDLogError(@"No previous version found. Possibly first launch since install.");
         [Environment resetAppData]; // We clean previous keychain entries in case their are some entries remaining.
     }
-    else if ([Environment.preferences getIsMigratingToVersion2Dot0] || [currentVersion compare:previousVersion options:NSNumericSearch] == NSOrderedDescending){
-        if([self isVersion:previousVersion atLeast:@"1.0.2"]) {
-            [VersionMigrations migrateFrom1Dot0Dot2ToVersion2Dot0]; // this is only necessary for older apps
+    else if(([self isVersion:previousVersion atLeast:@"1.0.2" andLessThan:@"2.0"]) || [Environment.preferences getIsMigratingToVersion2Dot0] ) {
```
corbett: getIsMigratingToVersion2Dot0 is stored in preferences not DB as we may not in general have a DB already if the migration was halted in the midst

PR #556, opened 2015-01-31
simplest fix to do group avatar deletion

corbett:
• associated tests
• for review; pay attention to handling as I haven't dealt with edges before. 
• eventually TSGroupModel should be refactored to have a TSAttachmentStream instead of an image blob.


Signal/src/Storyboard/Storyboard.storyboard
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
corbett: I am not sure how storyboard changes got in, maybe due to rebasing on master. these should not be part of the PR

PR #598, opened 2015-02-16
Tyler Design Review

FredericJacobs:
## Design to-dos
- [x] Review Iconography. Group placeholders are missing. Submit PR or post pixelapse link.
- [x] Asset sizes are incorrect. The "+" specially stands out for instance as being way to big.
- [x] Are we keeping the "iPhone-like" call icon? Smartphones are rarely used for phone calls. Iconography might be misleading and adding a warning might not be giving the best UX.
- [x] Tyler needs to set aspect ratio constraints in the signup view.
- [x] Approving [new ringtone](https://soundcloud.com/fredericjacobs/ring-ring-candidate) or proposing another one. (**Approved** by @abolishme , multiple ringtones would be nice but would require server-side changes ... so might not be worth it)
- [x] Comment on any other design issues in current build.

## Implementor's notes.

I'm aware that swiping behaviour is not yet adapted. Tableview to empty state should be properly animated.


abolishme:
getting a crash whenever i try to add a new group with a name but no recipients.

abolishme:
getting truncated text on the string `Add People` in the group builder
screen shot 2015-02-17 at 2 39 59 am

abolishme:
the photo selection dialog that slides up when user attaches photo from camera roll has a blue navigation bar that i believe is set to be translucent when it should be opaque, which is causing a slightly lighter blue.

FredericJacobs:
Just pushed a fix for the group chat issue.

FredericJacobs:
Everything in this thread was addressed.

PR #600, opened 2015-02-17
If you get a textsecure message from an unknown number, allow TextSecure interactions

corbett:
I decided to just check if a sender isn't a text secure user and there's an incoming message that isn't a call in a thread => that means you can reply. More complicated methods of handling would be to do the network call to check if a sender is a user (this isn't robust offline/poor network-so should probably be done in addition to the below anyway as a sanity check) and additionally cache that result somehow. This works for most cases, with the edge case that if you get a TextSecure message from an unknown contact and delete all incoming messages on the thread but leave the thread open, you will no longer be able to reply. Happy to tweak incorporating these ideas.


FredericJacobs:
Addressed in another PR using TSRecipients.

PR #602, opened 2015-02-19
Allows retry of failed downloads.

FredericJacobs:


corbett:
this looks good. did you test with audio, video, and with pictures?

FredericJacobs:
On download it doesn't make any difference.
TSAttachment pointers are the same for all media kinds at that point. But yeah, I tried video, not audio.

PR #604, opened 2015-02-19
fixed issue on plus position in group create

corbett:


FredericJacobs:
Merged as b37683c0ed71620a329c951083a36c7c994d0d21

PR #612, opened 2015-02-22
Display all contacts and allows you to invite them to use Signal

joyceyan:
In the compose message view, you can see all your contacts listed in a separate section in the UITableViewController. If you select a contact that is not a Signal user, then you will be prompted to invite them to use Signal. I don't have an iOS device (but will hopefully get my hands on one very soon) so I can't confirm that the SMS prompt works, but I think calling [contact fullName] and [contact allPhoneNumbers] should work.

I also noticed that there was something called an "inviteContactButton" which I assume was set up to let you click on the button, and then select what contacts you want to invite to Signal. I'm not sure if this UX is natural enough, but you can check it out. I just copied Android's layout of showing "TextSecure Contacts" and "All Contacts", but perhaps labeling the second section "Invite Contacts to Signal" instead of "All Contacts" would make this more clear.

### DIFF
--- a/Signal/src/view controllers/MessageComposeTableViewController.m
+++ b/Signal/src/view controllers/MessageComposeTableViewController.m
@@ -27,8 +27,10 @@
 {
     UIButton* sendTextButton;
     NSString* currentSearchTerm;
-    NSArray* contacts;
+    NSArray* contacts; // signal contacts
     NSArray* searchResults;
+    NSArray* allContacts; // signal + non signal contacts you can invite to signal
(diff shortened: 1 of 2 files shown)


joyceyan:
@JavaJens I didn't know that existed. I updated the link accordingly. Also unsure what the distinction is between the two links, but I just arbitrarily chose one.

JavaJens:
Yah, me neither. Sadly I never got a reply to my in-line comment...

PR #662, opened 2015-03-04
Allow the entry of the country dial code to also pre-select/filter

dirkx:
Very minor patch that makes entering the country during login easier by also detecting the dial prefix. Useful for countries which are way down the list or are not that consistently spelled in latin-languages.


FredericJacobs:
Thanks, merged in `testing`.

PR #663, opened 2015-03-05
Feature: TouchID Security support for the Screen Security Privacy

fbartho:
Sometimes people hand their device to others, or have them stolen while unlocked. This feature allows a user to lock access to the Signal App to only be used via verified biometric fingerprint.

A future improvement could enable the "Fallback Password" feature of LocalAuthentication, but to do that would require a more complex UI state-machine, and the addition of either a pod or a viewcontroller to handle that. This would require more complexity in settings, and keychain management.

Some Localizable strings were added, following existing project conventions, but they were only added to the english strings file, please let me know if you need a separate listing.


hubert3:
Nice work, although I would prefer a solution using kSecAccessControlUserPresence on a keychain item which contains the key needed to decrypt my authentication token / stored messages, rather than LocalAuthentication framework.

LocalAuthentication can be bypassed on a jailbroken device and does nothing for data stored at rest.

fbartho:
@hubert3 this pull request is purely a privacy feature akin to the existing "Screen Security" privacy setting. This PR doesn't change any data model involved, else that would be a much more complex feature.

Additionally, jail broken devices can't be presumed to have any security whatsoever. Key loggers or keychain, all are insecure. This PR isn't attempting to address or debate that question in any way.

In a non-jail broken context, the only "trusted" party is Apple.

Is the PR as given useful to you? Changing the whole security model seems out of scope for my available time.

hubert3:
I haven't tried it yet, but it sounds like a nice addition. I agree my suggestion would be a larger change.

I'm just wary of people using LocalAuthentication inappropriately or promising more security benefit from it than it really delivers, see e.g. this analysis of how Dropbox uses it:

http://www.andreas-kurtz.de/2014/10/ios-8-touch-id-local-authentication-caveats.html

FredericJacobs:
Hey @fbartho,

Thanks for this great contribution. I will fully review your pull request tomorrow. I'm focusing on some critical bug fixes today.

I think that we do want it to be part of the keychain authentication in the future.
The local SQLite database is encrypted with SQLCipher. Right now, the key is generated randomly on setup and stored in the keychain. We probably want the users to be able to provide a passphrase in the future but hasn't been implemented until now because of the extra UI needed not being a priority.

This could be a good temporary fix for people desiring local authenticating if we make it clear about how it works and under what threat model it's useful.

hubert3:
@FredericJacobs I think LocalAuthentication would be a nice extra security measure for screen lock once the app is already running and has opened the database (i.e. dbPassword is in memory)

Longer term I think it would be great if dbPassword could (optionally) have the kSecAccessControlUserPresence flag applied to it. 

This means you'd have to provide a fingerprint or re-enter the device passcode at app launch to be able to decrypt the database, which offers better security than kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly which is used at the moment (see line 145 TSStorageManager.m).

FredericJacobs:
@fbartho: I tested it on a iPhone 5S where TouchID isn't enabled and it allowed me to toggle the TouchID setting. Then when the user opens the app, it stays stuck on the launch screen.

FredericJacobs:
Looks like this is stalled. Closing. I'm happy to review another rebased PR with the mentioned issues addressed.

fbartho:
Yeah, @FredericJacobs I had limited contributor availability. Sorry I wasn't able to further advance the changes on this. Cheers, and good luck.

noahemmet:
Hi! Would people have time to review this PR if I rebased it first?

Also, how'd y'all feel about including a link to something like [Police Can Force You to Use Your Fingerprint to Unlock Your Phone](http://www.theatlantic.com/technology/archive/2016/05/iphone-fingerprint-search-warrant/480861/) (or an ACLU equivalent), maybe under the toggle switch, as a risk reminder to users?

michaelkirk:
> Hi! Would people have time to review this PR if I rebased it first?

Yeah, do it.

noahemmet:
Great, I'll get started. (PS, I think merging the [string localization PR](https://github.com/WhisperSystems/Signal-iOS/pull/1420) might make rebasing a bit easier here, but I can work around it for now!)

aledalgrande:
+1 for this feature

PR #595, opened 2015-02-15
mockup for recording audio messages

joyceyan:
If you uncomment out lines 1483 and 1488-1490 of MessagesViewController.m, then it should theoretically actually record audio (but not do anything with the audio file). But since I don't have a device and only the simulator to play with, calling "record" or "stop" in AVAudioRecorder causes my simulator to crash so I haven't tested that functionality. At the very least, this is a mockup for recording audio messages.

PR #694, opened 2015-03-09
Recording audio

jackflips:
@FredericJacobs @corbett Haven't been able to test sending proper because of certificate issues but it should be in order :)


FredericJacobs:
What's the reasoning for switching to EZAudio lib?

FredericJacobs:
Did you check with @joyceyan? She also worked on some recording UI in #595

corbett:
You should also note that you can have cert if you install app from store and then install app from Xcode on top of

joyceyan:
image

Some auto-layout issues on iPhone 6

jackflips:
@FredericJacobs I'm using EZAudio instead of SCWaveform because it made it simple to get live audio data to generate the waveform on the fly. It is also generally nicer and more easily modifiable. My PR is effectively the same as @joyceyan's mockup - it just has a little more UI on top.

FredericJacobs:
Great, thanks.
Will review later today!

FredericJacobs:
@joyceyan: Do you want to give a peer-review of this PR too since you worked on audio recording too?

Signal/src/view controllers/MessagesViewController.m
```diff
@@ -202,7 +232,15 @@ -(void) initializeTextView {
     [self.inputToolbar.contentView.textView  setFont:[UIFont ows_regularFontWithSize:17.f]];
     self.inputToolbar.contentView.leftBarButtonItem = _attachButton;
     
-    self.inputToolbar.contentView.rightBarButtonItem = _messageButton;
+    _recordRecognizer = [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(recording:)];
+    [_recordButton addGestureRecognizer:_recordRecognizer];
+}
+
+- (NSString *) applicationDocumentsDirectory
```
joyceyan: what does this do?

jackflips: It removes the send button and attaches a long tap gesture recognizer to the record button. Recording is initialized when the user taps and holds on that button.

TheStash: @jackflips so by default the "mic" button is shown which you can use to start recording, if you input text in the text field it switches to a "send" button?

jackflips: Yep. That happens in textViewDidChange:

joyceyan:
Cool! Since I still don't have a device yet I unfortunately can't do very thorough testing, but this looks good to me.. especially with the width of the text field changed :)

FredericJacobs:
Great! Thanks for the review @joyceyan

FredericJacobs:
I promoted it to the `testing` branch. I will review and test it some more over the next week and if it all goes well, it should be part of 2.0.2. Great work @joyceyan & @jackflips!

jackflips:
Sweet :+1:

FredericJacobs:
Seems to be working like expected.

FredericJacobs:
@jackflips 1) How does one cancel a recording? 2) You never delete the recording file?

jackflips:
1. The user can tap the x button on the left to cancel after recording.
2. It is overwritten on each subsequent recording.

FredericJacobs:
I rebased it on top of the testing branch. It's currently on `audio`.

I think the audio file should be deleted after the NSData is sent or after the user cancels.

FredericJacobs:
Also, there is no reason not to store it in the temp folder rather than in the Documents one.

FredericJacobs:
@jackflips : I'm sorry! I made a mistake in a rebasing conflict and accidentally removed the x button from the list of assets. Fixed now on the branch.

2) Also added code for better file management.

FredericJacobs:
@jackflips: It does appear that the playback of the the sound is a little out of sync with the waveform. Can you look into that? See `audio` branch.

jackflips:
@FredericJacobs I tested both the compose waveform and the waveforms in the chat proper and they seem fine to me. Perhaps you can send me an audio file you are testing with via signal?

Issue #673, opened 2015-03-05
Too much space below blue messages


rocknjohn:
Spacing below gray messages is as expected. Spacing below the blue messages is overly large. This is most obvious after the user sends multiple messages.

https://imgur.com/IqIVtke

I'd suggest the spacing beneath the blue messages should match the spacing beneath the gray messages. 
Using Signal 2.0

PR #729, opened 2015-03-19
Issue #673 - Spacing between incoming and outgoing messages

aaronraimist:
This patch fixes issue #673 by only displaying extra space below messages if needed. This gives room for senders name (in groups) and "Delivered" text below messages.


FredericJacobs:
Thanks, merged into `testing` as a38101c570ab55feeb9d110dd4a561e4ad04

PR #732, opened 2015-03-19
Fixed build after SCWaveForm major release

agentcoops:
The version of SCWaveForm was not fixed in the Podfile and a major release recently occurred (2.0.0) that broke API compatibility and thereby the Signal build. Set SCWaveForm version to last known to work with Signal.


FredericJacobs:
Thanks, but current version is no more using SCWaveForm. See `testing` branch

PR #740, opened 2015-03-24
audio recording minor fixes

jackflips:


FredericJacobs:
Hey Jack.
After the audio attachement is sent, I still have an issue of synchronization with the waveform.

Can you check the `audio` branch? Might also have to upgrade your CocoaPods version by running `gem update`.

jackflips:
Hey @FredericJacobs,
I was at first able to reproduce the out of audio-out-of-sync issue, but after deleting the conversation and recording a whole bunch of new messages I can't seem to reproduce it. I'll try with another device now...

jackflips:
I've tested sending audio from 3 devices to each other and can't seem to reproduce the bug. @FredericJacobs can you send me an audio file that causes the error to happen in Signal?

Issue #649, opened 2015-03-03
URLs in chats are not showing up as links


BBKrWw:
It would be nice if we could copy/paste a URL or type one, and have it show up as a clickable link.

1337sup3rh4x0r:
I second that! Links should be highlited and clickable

samer:
A temporary fix that works for me is to tap and hold the message bubble, then tap on "copy". The URL should now be underlined and tappable.

pmciano:
@samer You don't even have to tap on copy. You can tap on the message again, or anywhere else in the conversation, and the URL will become clickable.

samer:
@pmciano indeed! :+1:

maltemayer:
@FredericJacobs Milestone for 2.0.2?

sammcj:
+1

maltemayer:
This is still a problem, I wonder why it wasn’t set as milestone for 2.0.2.
People I convinced using Signal, easily getting angry about things like this. There are already complaining about messages not being sent, I hope they wont find out about this URL issue untill the fix.
It would be very sad losing user over issues like this.

sammcj:
It's great that signal is making decent quality encryption available to everyone but I agree - the UI seems to put off most people I've tried to convince to use it.

hubert3:
I thought this was intentional to stop malicious links that could exploit your phone being opened too easily, but I guess it's just a bug

sammcj:
There should at least be an option for it then right?

lordmortis:
Is this a bug or is this working as intended?

PR #776, opened 2015-05-01
Fix Issue #649

lordmortis:
Data detectors are connected to if a textview is selectable. For some reason their selectability was disabled, so re-enabling fixes issue #649 

It doesn't appear to have any knock on effects? I think it was introduced by accident in 402df723 

Images still appear to be viewable.


FredericJacobs:
Thanks for the pull request. 

Two questions:
1) Looks like the date selectors are not activated.
2) If a link is selectable, long-pressing the message does not permit to delete a message.

lordmortis:
Date selectors *should* be a thing we can enable by changing values on the text box, i'll take a look.

How do you want to handle the long-press thing? I can't just re-enable selectors, or the data detectors won't work. Do we still want it to be long-press, or do we want some other gesture to allow deletion?

FredericJacobs:
I think long press for deletion is the correct gesture. We need to figure out a way to bound the gesture recognizer in some way. If we can detect touches in a bubble outside of the data selector area, things are solved.

FredericJacobs:
Merged as 0f57804eefcb837a83db96e35e2f2c70e7c80fe5

PR #796, opened 2015-05-26
Feature/better push notification messages

lordmortis:
I think this format for the notifications is a little closer to how other apps do it? 

I also think having "Signal" as the title may be a little redundant because the icon will still be present and a swipe on the message will still open the app? 

I'm now showing the user or group name as the "title" (so that's in bold instead of "Signal") 
If the message bodies are to be shown, they are displayed as the text body (in the case of the group, the name of the user it's from is displayed first with a ":" separating them and the message)

If the names are show but message bodies are not then either "New Message from (user)" is displayed or "New Message!" is displayed

If neither is shown, the original "Signal" and "New Message" is displayed.

Hope you like it :)


FredericJacobs:
I have fixed the issue, shortening the notification messages length when possible. Will push along with the 2.1 tag.

lordmortis:
How do Facebook Messenger and Google Hangouts do it? :/

FredericJacobs:
They don't use the `alertTitle` property AFAIK. It's all in the `alertBody`.

FredericJacobs:
The alertTitle property is I believed filled by iOS to the title of your application's name before being displayed. So whatever you set there will be overwritten.

lordmortis:
It's working in the notifications view, but not on the lock screen (gets overwritten)

I also suspect it will do the correct thing on the watch (it seems to use the alertTitle on my Pebble)

FredericJacobs:
Ok, I'll set the `alertTitle` property. It doesn't hurt and hopefully it displays correctly on a iWatch.

lordmortis:
Yeah, will still have to include the group/person in the body of the message sadly. I could have sworn when i did the original modification it wasn't doing this, but i must be wrong. 

Having the user/group name in bold *does* make it easier to read off the watch however. 

Here's now google hangouts does it.
img_1725
img_1723

lordmortis:
Some more information here: http://stackoverflow.com/questions/29425358/how-to-configure-local-notification-for-the-apple-watch

https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/ApplePushService.html#//apple_ref/doc/uid/TP40008194-CH100-SW1 <- remote notification payloads can have titles

FredericJacobs:
If you want to work on the `alertTitles`, please work on top of this branch: https://github.com/WhisperSystems/Signal-iOS/tree/notifications-advanced

PR #841, opened 2015-07-10
Minor improvements

ruuki:
declared constants instead of string literals
avoided using self within block
removed unused method decleration from header file


FredericJacobs:
Merged as 040e4c750b62895f5cb61a31cc9f416a46d3c2ed

PR #865, opened 2015-08-09
Attempt to reset the keyboard back to the default

brd:
This should reset the keyboard after sending a message back to the default keyboard from the Emoji or the number/special characters keyboard.

I can't get it to install on my iPhone due to no Apple iOS Developer account, but it looks like it should do the trick. At least according to: https://github.com/jessesquires/JSQMessagesViewController/issues/971

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -574,6 +574,14 @@ typedef enum : NSUInteger {
         TSOutgoingMessage *message = [[TSOutgoingMessage alloc] initWithTimestamp:[NSDate ows_millisecondTimeStamp] inThread:self.thread messageBody:text attachments:nil];
         
         [[TSMessagesManager sharedManager] sendMessage:message inThread:self.thread];
+
+        // Reset the keyboard from special characters to Default
+        [self.keyboardController endListeningForKeyboard];
+        [self.inputToolbar.contentView.textView resignFirstResponder];
+        [self.inputToolbar.contentView.textView setKeyboardType:UIKeyboardTypeDefault];
+        [self.inputToolbar.contentView.textView becomeFirstResponder];
+        [self.keyboardController beginListeningForKeyboard];
+
         [self finishSendingMessage];
     }
 }


chrisballinger:
@brd You can install to device with Xcode 7 beta without a paid account

FredericJacobs:
What exactly is the reasoning behind this? It seems that Apple is not doing this in iMessage.

brd:
@chrisballinger I tried and still cannot get it to work even after finding the build config and telling it not to sign.

@FredericJacobs iMessage does reset the keyboard back to the qwerty keyboard when you press send. To me when I send a message after ending the message with say a question mark, and I want to send another immediately, it is nice to have the keyboard reset back to qwerty because I always start with a letter.

chrisballinger:
@brd You still have to sign, you just don't need a paid account... I think.

brd:
Still no luck, can someone else test this for me?

FredericJacobs:
Looking at this tonight.

FredericJacobs:
Actually, I'm gonne refactor the entire class since it needs a code cleanup.

philiprenich:
While a small UX change, this is actually one of my biggest annoyances with the app right now. I would love to see this get thrown into an update soon! It really would help make the texting a more seemless and smooth experience - bringing it inline with iMessage and regular texting - which is what it needs to be for greater adoption.

michaelkirk:
> iMessage does reset the keyboard back to the qwerty keyboard when you press send. 

For me on iOS 9.3 this is true with the numeric keyboard/punctuation keyboard, but not the emoji.

Typing in numbers from numeric keyboard
image

Press send, and QWERTY keyboard **exposed**.
image

Typing in emoji from emoji keyboard
image

Press send, and emoji keyboard **remains**.
image

WhatsApp has the exact same behavior, which leads me to think we should follow suit here.

It's reasonable that people who send an emoji, often want to send another emoji, while people who send a punctuation are usually ready to type more characters.

philiprenich:
iOS 9.3 for me as well and I don't have that experience. Numeric keyboard stays upon send.

michaelkirk:
Sorry - I was clarifying the behavior of the native messaging client (and whatsapp) which I think we should strive for I. This case.

Signal behaves for me as you've described it.

michaelkirk:
We want to stick with the existing behavior.

Issue #871, opened 2015-08-18
wou want


RiseT:
Search the translatable strings within this repository for '**wou want**' and you'll find a typo ... ;-)

PR #876, opened 2015-08-19
Change the translation for FINGERPRINT_SHRED_KEYMATERIAL_CONFIRMATION…

orta:
Fixes - https://github.com/WhisperSystems/Signal-iOS/issues/871

Previously said `wou want` now says `you want`.


FredericJacobs:
Thanks for the pull request.
But we never update the strings directly. It's delegated to Transifex.
Just the english string is modified.

PR #884, opened 2015-09-01
Some nits & add corner rounding to the message text box.

mahyar:
I've always thought the lack of corner rounding in the message box didn't look that good, so I added some.

### DIFF
--- a/Signal/src/environment/VersionMigrations.m
+++ b/Signal/src/environment/VersionMigrations.m
@@ -33,7 +33,7 @@
 
 + (void)performUpdateCheck{
     NSString *previousVersion     = Environment.preferences.lastRanVersion;
-//    NSString *currentVersion      = [Environment.preferences setAndGetCurrentVersion]; //currently unused
+    /*NSString *currentVersion  =*/ [Environment.preferences setAndGetCurrentVersion];
     BOOL     VOIPRegistration     = [[PushManager sharedManager] supportsVOIPPush]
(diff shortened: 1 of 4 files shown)


mahyar:
I also added a 1px line to the top of the bar and put the current version set command back in.
screen shot 2015-09-01 at 8 57 39 am

rocknjohn:
+1 rounded corners
-1 on blue top border. The existing neutral is cleaner.

While the design of this area is being considered: 
- Could we try a faint neutral gray border around the message box? Very faint, like lighter than even iMessage has. But it would aid visual distinction of the message area from the surrounding area and may make it feel more polished.
- When the "Send" button is disabled and gray, the gray has a dark enough tone to not be clear it is disabled. Could the gray text color of the disabled "Send" button be made a bit lighter?

FredericJacobs:
Ping @abolishme

mahyar:
Since your designer is AWOL maybe we can try rounded corners at least?  It shouldn't be too risky.

michaelkirk:
> Could we try a faint neutral gray border around the message box? Very faint, like lighter than even iMessage has. But it would aid visual distinction of the message area from the surrounding area and may make it feel more polished.

> When the "Send" button is disabled and gray, the gray has a dark enough tone to not be clear it is disabled. Could the gray text color of the disabled "Send" button be made a bit lighter?

FYI @rocknjohn these are both coming in #1255

Issue #957, opened 2015-11-19
No exit button when scanning fingerprint


tbrixen:
When tapping to scan another user's fingerprint the camera openes. 
But the only way of getting back to Signal (without scanning) is force-closing the app.

PR #971, opened 2015-11-27
Fix for issue #957 (quit button not displaying when scanning fingerprint)

rygen:
The AVCaptureVideoPreviewLayer was being inserted as the topmost layer, therefore the quit button wouldn't show up when ScanIdentityBarcodeViewController is presented.


FredericJacobs:
Thanks, it was merged as https://github.com/WhisperSystems/Signal-iOS/commit/25293fd40b402959bfbe802dbeffe4f068186d0d.

PR #976, opened 2015-11-30
Improve visibility of X in quit button

rygen:
When viewing the quit button over the Scan Identity Barcode View Controller, I found the white X to be hard to distinguish itself from the surrounding circle. 
In my tests visibility was slightly improved with X being transparent.

This doesn't change its appearance on the Present Identity QR Code View Controller, since its background is already white.

Please submit to your designer for review.

[FREEBIE]


michaelkirk:
The new safety number screens addressed this differently.

Sorry your PR rotted for so long =(

rygen:
No problem! :)

PR #995, opened 2015-12-08
Cocoapods needs latest openssl - 1.0.205

swedakonsult:
OpenSSL to 1.0.205
*  ```pods install``` failed because https://openssl.org/source/openssl-1.0.2d.tar.gz has been moved to it's long term storage and replaced by 1.0.2.e
* Retrieved the new version by looking at changes made on December 3, 2015, to https://raw.githubusercontent.com/CocoaPods/Specs/master/Specs/OpenSSL/1.0.205/OpenSSL.podspec.json

FREEBIE

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -3,7 +3,7 @@ source 'https://github.com/CocoaPods/Specs.git'
 inhibit_all_warnings!
 link_with ["Signal", "SignalTests"]
 
-pod 'OpenSSL',                    '~> 1.0.204.1'
+pod 'OpenSSL',                    '~> 1.0.205'
 pod 'libPhoneNumber-iOS',         '~> 0.8.7'
 pod 'AxolotlKit'               
 pod 'PastelogKit',                '~> 1.3'
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -21,7 +21,7 @@ PODS:
   - AFNetworking/UIKit (2.6.3):
     - AFNetworking/NSURLConnection
     - AFNetworking/NSURLSession
-  - AxolotlKit (0.8):
+  - AxolotlKit (0.7):
     - 25519 (~> 2.0.1)
     - HKDFKit (~> 0.0.3)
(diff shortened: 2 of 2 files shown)


swedakonsult:
I'm not finding the Jenkins files - which repo are they in?
I'll start by reverting the .lock changes to see if that passes the test.

drichardson:
@swedakonsult The error message from travis is:

> error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.

I'm not familiar with travis, but somehow you need to `pod install` before the build starts.

swedakonsult:
Thanks @drichardson, I found that but was looking for a way to possibly resolve this by looking at the build files. It does seem like the validation is performed at the wrong time, but I have no previous experience of Cocoapods so I might not be able to fix the root cause.

swedakonsult:
I'd be concerned about merging these until the tests are updated in Jenkins. I have limited knowledge of Cocoapods but it seems to me like there's a chance that the merged change of updating the Podsfile might not have been part of the test deployment. It looks like the verification is only comparing the file in the repo with the file somewhere else (not clear where it's sourced, from the test error statement). The test should maybe perform a ```pod install``` or ```pod update``` at some point and then verify that it looks the same as the checked-in file?

FredericJacobs:
This is already fixed on some branches.

swedakonsult:
@FredericJacobs can you expand on the message for closing this pull request?
1. Does it mean that it's already been merged into the main branch so that building the source is possible?
2. Is the pull request strategy for this project to merge the pull requests into other branches prior to merges into the delivery branch?
3. Could you provide a link to any documentation describing which branch(es) are supposed to be used for breaking fixes?

It would also be very useful if you could provide a link to the branch that fixes this. Mainly because it might help someone that is running into this same problem to get a better understanding of when it will be fixed in the main branch.

PR #1014, opened 2015-12-20
Add subtitle to Screen Security setting to explain its function.

kunickiaj:
First time contributor to this project, so please let me know if the Podfile shouldn't be updated.
I did update OpenSSL to 1.0.205 since the binaries for 1.0.204.1 are no longer being hosted on the openssl site at the link in that version of the cocoapod.

Added an English string to the localization files, which should be translated for each language supported.

The text I came up with is: *Obfuscates Signal thumbnail in app switcher* as the subtitle to the Screen Security setting.


kunickiaj:
Looks like the build fails, but would be resolved by: https://github.com/WhisperSystems/Signal-iOS/pull/1006 which runs `pod install` in the travis build, or by updating openssl in the precompiled pods dependency.

forthrin:
The text sounds good if that's what the setting does. Remember to change the text if what it goes is extended or changed.

FredericJacobs:
Great @kunickiaj.

I'll look into it. Don't worry about CI

FredericJacobs:
I'll rebase that tomorrow. 

Note for future contributions, we never update the localizations directly (besides the English one). Translations are done on Transifex.

kunickiaj:
Good to know, thanks! Hopefully the next contribution will actually include
more than a couple lines :)

FredericJacobs:
I changed it to a section footer to have more room for the translations and look more like the iOS settings hints.

img_0006

Issue #1025, opened 2015-12-23
Improve Look and Feel to fit in more with iOS 


jaredStef:
Many parts of the app seem inconsistent and could be better made to fit in with the operating system.
- [ ] Made cells of message groups larger
- [ ] Extend Cell Insets to far right
- [x]  Round the corners more on the message text field 
- [x] Make the send button the same blue the app uses which would imply that it could be tapped and leads to nicer ui
- [ ] Use standard system Alert View Controller and just change the color when picking images
- [ ] Add back button to settings VC so it is more consistent and less confusing than having it segue back when the X is tapped. 
- [x] Make Network Status Not animate when tapped

I know Objective-C and work with iOS apps, so I am willing to add all of these changes myself. I have been using the app a lot recently after convincing my friends to switch from using Skype due to privacy concerns and want the user experience of this app to be the best it can be.

nikklassen:
This is not strictly a look and feel issue, but I thought this was a good place to mention it since this issue is actively being worked on. I would like the font size the be configurable to aid in the accessibility of the app. Is that something that could be added with these changes?

jaredStef:
+1 Everyone should be able to use the app, so we should make it as accessible as possible. My work on my current PR has been postponed a week or two due to me finishing a college class, but when I get to it I would be glad to help. Can anyone else work on this issue?

FredericJacobs:
@nikklassen The messages in the conversation view are DynamicType, so the font size from iOS settings should be taken into account. It has still to be done in other places. It's the #471 issue

nikklassen:
@FredericJacobs Great, thanks! Sorry I missed that one

michaelkirk-signal:
Some of this has been addressed, and the remaining items seem unlikely to change. e.g. the presentation of the settings controller is intentionally modal.

If you have a specific bug to report, please open a new issue. If you'd like to propose other design changes, the community forum is a better place to have that kind of discussion: https://community.signalusers.org

PR #1026, opened 2015-12-23
Fix UX Issues Discussed in Issue #1025

jaredStef:
Filed this PR to discuss the UX issues discussed in my issue request #1025 

1. Made cells of message groups larger
2. Extend Cell Insets to far right
3. Round the corners more on the message text field
4. Make the send button the same blue the app uses which would imply that it could be tapped and leads to nicer ui
5. Use standard system Alert View Controller and just change the color when picking images
6. Add back button to settings VC so it is more consistent and less confusing than having it segue back when the X is tapped.
7. Make Network Status Not animate when tapped

3,4 are fixed in PR #884 

## Completed (Need Testing)  
5, 7

1, 2, & 6 need to be discussed

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -1766,37 +1766,20 @@ typedef enum : NSUInteger {
 - (void)didPressAccessoryButton:(UIButton *)sender {
     [self dismissKeyBoard];
 
-    UIView *presenter = self.parentViewController.view;
+    UIAlertController *alertController = [UIAlertController alertControllerWithTitle:NULL message:NULL preferredStyle: UIAlertControllerStyleAlert];
+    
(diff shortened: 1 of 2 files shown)


FredericJacobs:
Hi @jaredStef,

I agree on changing the image picker dialog to a native dialog. I think we should then just drop `DJWActionSheet` everywhere as a dependency and use `UIAlertController` with `UIAlertControllerStyleActionSheet`  style instead.

Can you make a pull request / commit with just that change? It will be easier for reviewing it.

Thanks!

michaelkirk:
This PR is accounted for elsewhere, so I'm closing it.

Network Section become unselectable in https://github.com/WhisperSystems/Signal-iOS/commit/bd377e65af35547fcfa404f1c8fcd9f1931fbee8#comments

And your conversion to AlertViewControllers has been incorporated into another more complete PR here: #1106.

Thanks for your attention to UX!

PR #1027, opened 2015-12-24
Switch to native UIAlertController and drop DJWActionSheet dependency 

jaredStef:
I am currently working on removing it's other uses and I will push a commit when I have done so

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -1765,39 +1765,23 @@ typedef enum : NSUInteger {
 
 - (void)didPressAccessoryButton:(UIButton *)sender {
     [self dismissKeyBoard];
-
-    UIView *presenter = self.parentViewController.view;
-
-    [DJWActionSheet showInView:presenter
-                     withTitle:nil
-             cancelButtonTitle:NSLocalizedString(@"TXT_CANCEL_TITLE", @"")
-        destructiveButtonTitle:nil
-             otherButtonTitles:@[
-                 NSLocalizedString(@"TAKE_MEDIA_BUTTON", @""),
-                 NSLocalizedString(@"CHOOSE_MEDIA_BUTTON", @"")
-             ] //,@"Record audio"]
-                      tapBlock:^(DJWActionSheet *actionSheet, NSInteger tappedButtonIndex) {
-                        if (tappedButtonIndex == actionSheet.cancelButtonIndex) {
-                            DDLogVerbose(@"User Cancelled");
-                        } else if (tappedButtonIndex == actionSheet.destructiveButtonIndex) {
-                            DDLogVerbose(@"Destructive button tapped");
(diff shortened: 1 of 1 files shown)


FredericJacobs:
Good, looking forward to reviewing that.

michaelkirk:
Thanks for working on this @jaredStef! I'm definitely in favor of getting rid of all the DJWActionSheets. They are kind of tacky looking, plus the AlertControllers have a nicer API, and when in doubt, "go native". =)

Signal/src/view controllers/MessagesViewController.m
```diff
@@ -1765,39 +1765,23 @@ - (void)audioRecorderDidFinishRecording:(AVAudioRecorder *)recorder successfully
...
-                                    [self recordAudio];
-                                    break;
-                                default:
-                                    break;
-                            }
-                        }
-                      }];
-}
+    
+    UIAlertController *alertController = [UIAlertController alertControllerWithTitle:NULL message:NULL preferredStyle:UIAlertControllerStyleAlert];
```
michaelkirk: We probably want an action sheet style alert controller.

And style nit: you should be passing in `nil` instead of `NULL` for things that can be an object. http://stackoverflow.com/questions/557582/null-vs-nil-in-objective-c

Signal/src/view controllers/MessagesViewController.m
```diff
@@ -1765,39 +1765,23 @@ - (void)audioRecorderDidFinishRecording:(AVAudioRecorder *)recorder successfully
...
+    
+    UIAlertAction *takeMedia = [UIAlertAction actionWithTitle:NSLocalizedString(@"TAKE_MEDIA_BUTTON", @"") style:UIAlertControllerStyleAlert handler:^(UIAlertAction * _Nonnull action) {
+        [self takePictureOrVideo];
+    }];
+    
+    UIAlertAction *chooseMedia = [UIAlertAction actionWithTitle:NSLocalizedString(@"CHOOSE_MEDIA_BUTTON", @"") style:UIAlertControllerStyleAlert handler:^(UIAlertAction * _Nonnull action) {
```
michaelkirk: This actually fails to compile because you are using an UIAlertControllerStyle constant for what should be a UIAlertActionStyle constant. (This wouldn't happen in swift! ( :fist: *shakes fist at sky*))

jaredStef:
Yeah I really need to go over it and finish it. I am in the middle of registering an LLC, and launching a app on the App Store, but I plan on completing this when I am out on Spring break starting March 24. Thanks for the pointers! I presume you'll be taking over since Fredrick was hired by Apple?

michaelkirk:
@jaredStef - I've built on top of your changes here: https://github.com/WhisperSystems/Signal-iOS/pull/1106

Thanks for getting this started!

michaelkirk:
superseded by #1106

PR #1049, opened 2016-01-19
SignalUITests added

matthewkotila:
Added UI tests to Signal in Swift. Some tests require the app to be
verified with a real number, and some require a “Test” contact that is
verified with Signal (can be self).

### DIFF
--- a/SignalUITests/SignalUITests.swift
+++ b/SignalUITests/SignalUITests.swift
@@ -1,0 +1,737 @@
+//
+//  SignalUITests.swift
+//  SignalUITests
+//
+//  Created by Matthew Kotila on 1/13/16.
+//  Copyright © 2016 Open Whisper Systems. All rights reserved.
+//
+
+import XCTest
+
+class SignalUITests: XCTestCase {
+    
+    override func setUp() {
+        super.setUp()
+        continueAfterFailure = false
+        XCUIApplication().launch()
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+    }
+    
+    // requires unverified app
+    func testCountryCodeSelectionScreenNavigation() {
+        
+        let app = XCUIApplication()
+        app.buttons["Country Code"].tap()
+        
+        XCTAssert(app.staticTexts["Select Country Code"].exists)
+        
+    }
+    
+    // requires unverified app
+    func testCountryCodeSelectionScreenBackNavigation() {
+        
+        let app = XCUIApplication()
+        app.buttons["Country Code"].tap()
+        app.navigationBars["Select Country Code"].buttons["btnCancel  white"].tap()
+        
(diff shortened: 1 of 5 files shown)


SignalUITests/SignalUITests.swift
```diff
@@ -0,0 +1,737 @@
...
+        super.setUp()
+        continueAfterFailure = false
+        XCUIApplication().launch()
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+    }
+    
+    // requires unverified app
```
michaelkirk: Do you propose manually getting the app into this state and then running the tests? Or would you be able to get the app in this state as part of the test suite?

michaelkirk:
excited to have some uitests :+1:

SignalUITests/SignalUITests.swift
matthewkotila: Hi Michael, I'm not extremely versed in UI tests, so when I encountered the problem of "application state" (i.e. testing UI pre-verification and post-verification), I simply wrote tests for an app that was pre-verified, and then verified it manually using the iOS simulator with valid information, and then wrote tests for the post-verified app. I'm unsure if there's a way to automatically input information (such as contacts, verification codes, etc.) once and start a different set of UI tests intended for a verified app. Additionally, I'm unsure if there's a way to automatically "wipe" an app in order to test pre-verification UI. All very interesting questions in the realm of UI testing!

michaelkirk: These are just my thoughts on how to do testing well, I don't have commit
access to this project:

It's important that the test suite be able to run without manual
intervention. What I have done with other ios apps is add something like a
testing category to the app delegate to the testing target (only!) which
exposes useful interfaces for faking data.

E.g. addFakeContact, registerPhone, unRegisterPhone, addMessageThread might
be useful for Signal iOS.

Then you can use these methods in your test setup blocks to get the app in
a useful state for your tests.

michaelkirk:
Hey @matthewkotila - I'm excited about having some UITests for Signal iOS. I did a little work based on your commit, you can see the WIP here: https://github.com/michaelkirk/Signal-iOS

I'm seeing some pretty flaky results though - often the app fails to start. Are you seeing that? Are you running the tests on your device on a simulator?

michaelkirk:
I didn't forget about this! The flaky test results were seemingly related to the background-fetch permission which we no longer use. If you have any interest in revisiting UITests, I'd happily merge some stable tests. 

It *is* important that they be runnable automatically though.  I did some abandoned work in to that end here:
https://github.com/michaelkirk/Signal-iOS/tree/uitest

e.g. faking registration and such.

I'm going to close this as it's woefully out of date at this point, but if you're still interested in working on it, please do.

PR #1054, opened 2016-01-24
Update Storyboard.storyboard

140483:
Use of the "x" (cross) system item instead of a personnalised thicker one to fit more to iOS design


michaelkirk:
Sticking with existing design unless we do a full design overhaul. Thanks!

Issue #1076, opened 2016-02-12
Problem with installing openssl from pod


ArkadiuszBanas:
During pod installing I've received: 
`Installing OpenSSL (1.0.205)

[!] Error installing OpenSSL
[!] /usr/bin/curl -f -L -o /var/folders/_k/3smslx552gq2w0l43sx_zs0r0000gn/T/d20160212-81963-1u45fd5/file.tgz https://openssl.org/source/openssl-1.0.2e.tar.gz --create-dirs`

The problem is 'openssl-1.0.2e.tar.gz' not existing anymore. Podfile needs to be updated.

ArkadiuszBanas:
Comment to commit: 
There's no sens to specify pod version as X.X.XXX with ~> mark. 
There was two ways to fix that:
1. '~> 1.0' 
2. '1.0.216'. 

I've choosed first option to be consistent with rest of podfile.
Of course it's FREEBIE :)

ArkadiuszBanas:
Previous pull request was abandoned.

michaelkirk:
This has been fixed in master.

PR #1077, opened 2016-02-12
#1076 - Problem with installing openssl from pod - fixed

ArkadiuszBanas:
1.0.205 is not exisitng anymore.
~> sign will not work if we adding last part of version signature.
Removing last part will fix the problem.

FREEBIE


ArkadiuszBanas:
I have some problem with Travis verification. I've modified my commit to check if it will help.

PR #1078, opened 2016-02-12
Useless didReceiveMemoryWarning methods removed from view controllers

ArkadiuszBanas:
During code review I've found few useless didReceiveMemoryWarning methods in view controllers. I've decided to remove is. 
Less is better :)


ArkadiuszBanas:
It didn't pass becuase of problem with pod (openssl). My previous patch should fix that.

PR #1079, opened 2016-02-12
#1076 - OpenSSL updated to 1.0.206.

ArkadiuszBanas:
It's fix for error during pod install.

FREEBIE

PR #1081, opened 2016-02-14
fix unused_string.py uses python interpreter

michaelkirk:
and some readability improvements

### DIFF
--- a/Scripts/unused_strings.py
+++ b/Scripts/unused_strings.py
@@ -1,9 +1,10 @@
-#!/usr/bin/env sh
+#!/usr/bin/env python
 """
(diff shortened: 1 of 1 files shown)


ArkadiuszBanas:
Fail because of:
`Installing OpenSSL (1.0.205)

[!] Error installing OpenSSL

[!] /usr/bin/curl -f -L -o /var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/d20160214-3593-1vnvk6n/file.tgz https://openssl.org/source/openssl-1.0.2e.tar.gz --create-dirs
`
It's already fix but somebody have to submit it (https://github.com/WhisperSystems/Signal-iOS/pull/1079)

Issue #1094, opened 2016-02-28
Feature request: Different new message sound


fathorse:
I've found that there is a general consensus among my Signal using friends that the current NewMessage.aifc sound should be updated from the current "clomp" to something that's slightly more pleasant to the ear. Alternatively, add a handful of sounds and let the user choose. Thanks! :)

michaelkirk:
I'm not sure if we're talking about the same thing, but I think it's a problem with the audio levels on the second beat.

See my new PR #1095

caveman1973:
I see no problem with the sound.

The real problem is not being able to choose a different sound for different contacts.

michaelkirk:
@caveman1973 you might be interested in https://github.com/WhisperSystems/Signal-iOS/issues/1009

mkb:
I'm not keen on the notification sound either. Custom sounds (#1009) would be ideal but changing the built-in sound would be sufficient.

richardsj:
Yes please - any sound would be nicer than the current one.  There is a pretty nice list here https://notificationsounds.com/notification-sounds - and some of these could be shortened so that the effect is shorter and more subtle.  A few loud ones in there and a menu - job done!

michaelkirk-signal:
This was addressed a while ago, when we added a selection of system sounds and changed the default.

PR #1095, opened 2016-02-28
Limit filter for NewMessage.aifc to avoid blown out audio

michaelkirk:
limited NewMessage output to -3DB. Previously was over-amplified, leading to unpleasnt distortion

Disclaimer, I'm not an audio engineer, my audio processing experience comes from recording me and my friends making music when I was a teenager =)

I've applied a limiting filter to clip the loudest audio for the NewMessage audio clip, which I think solves #1094 

Here's a video of an analysis I did of the audio file - we're peaking out on the second stroke, which results in a harsh clip: https://vimeo.com/157036596

You can see that the new audio clip peaks at -3db.


richardsj:
any sound would be nicer than the current one. There is a pretty nice list here https://notificationsounds.com/notification-sounds - and some of these could be shortened so that the effect is shorter and more subtle. A few loud ones in there and a menu - job done!

Issue #866, opened 2015-08-11
Signal eats too much traffic


JimLoose:
I don't use signal that much, especially not that much for multimedia.
Anyway Settings > Mobile > shows me 155MB for what?

What needs that much traffic in Signal? (much more than Whatsapp (8MB))

![img_0924](https://cloud.githubusercontent.com/assets/195000/9192894/6afdda68-400f-11e5-8373-68c75277cec5.PNG)

How much is your data usage? It would be great if you could please look into that issue.

hubert3:
Probably this feature: https://whispersystems.org/blog/contact-discovery/

Which means you can discover contacts on Signal without uploading your entire address book, as Whatsapp and most other apps do.

TuxDev1337:
So how to disable that? I don't want that anyway.

TuxDev1337:
I would appreciate if somebody could investigate this. For me Signal is unusable if I am not at home, since I don't want to spent all of my data volume for something that is not really useful for me. There should be an opt-out for that contact checking.

FredericJacobs:
We are working on this. Will post an update soon.

TuxDev1337:
Awesome :D Thank you!

evanchaney:
I have also experienced this issue. I am seeing Signal consume cellular data at a rate of about 5–10MB per day without launching the app or allowing it to use background refresh.

evanchaney:
Lending some credence to hubert3’s guess, I have updated 3 contacts in the iOS contacts app over the last week which looks like it triggers a refresh of Signal’s contacts (presumably in the background):

https://github.com/WhisperSystems/Signal-iOS/commit/e8ea00d71a4b72a4b7a2e9bfaf2026dafbe1b064

JimLoose:
> We are working on this. Will post an update soon.
??

This is still a very big issue without any clear explanation or fix!
My contract only has 500MB and Signal eats most of it, that's why i and most of my friends removed it for the now!

dan-tum:
Same here, in the last 3 months I have 450 MB of Signal cellular usage while only using it with one single friend and I only use texting. It has used more data than Safari, Facebook and Alien Blue, apps which I use very often.

alasyon:
I'm having the same issue. Uses 70-100MB per day, sending 10-20 text-only messages. My data cap is 2GB/month, so if I use Signal then it single-handedly uses the entire cap. Makes it unusable for myself and most of my iPhone owning friends.

[...]

PR #1098, opened 2016-03-05
Upgrade OpenSSL dependency version

jschneier:
FREEBIE


haffenloher:
This has already been done in the [`opensslupdate`](https://github.com/WhisperSystems/Signal-iOS/tree/opensslupdate) branch.

jschneier:
@haffenloher thanks for pointing that out. Do you know why it hasn't been committed? I'd like to hack around and contribute but having a broken master makes that less than simple....

JohanDoe:
I have the feeling no-one is working on the master / Signal OS since @FredericJacobs left. Maybe @moxie0 knows more about the state of play here? Seems like more and more people are waiting for some of the PR being committed to the master and the next version (see for example #866).

Issue #913, opened 2015-10-08
Cannot delete a message in a conversation


mmkassem:
The users should be able to delete a single (or multiple) message(s) they have either sent or received.

iOS 9.1 public beta
Signal 2.1.3

michaelkirk:
I might be misunderstanding, but it seems like this has been around for a while:

Does this video capture the behavior you want to see?
https://vimeo.com/158448243

If not, can you provide more detail?

wdormann:
That shows deleting a failed message.   I see no way of doing that for a successfully sent or received message.

Issue #704, opened 2015-03-10
Ability to save and delete videos


lskjdf22:
Pictures are currently savable. It would be nice to be able to save a video locally as well. There may be a way to do this in the UI, but it is not apparent to me.

FredericJacobs:
No. It is currently not possible. Neither is it to delete a video. Still wondering about the UX of this. Probably just long tapping the video in the conversation view.

fxer:
Is there something about adding the save option to videos that is much more difficult than adding it to still images on iOS?

yaelwrites:
Assuming this feature is still unavailable?

michaelkirk:
closed long ago in #1105

PR #1102, opened 2016-03-10
present action sheet to delete sent or received messages.

michaelkirk:
Note: This will fail until #1076 is merged.

#913: There is currently no way to delete individual messages in Signal-iOS. Was there ever? 

This solves  #913 by presenting an action sheet when any non-media message is tapped.).

Deleting a media message is not part of this PR, and discussed here: https://github.com/WhisperSystems/Signal-iOS/issues/704. That may entail rethinking this feature so we can get consistent delete across message types.

Also, in line with https://github.com/WhisperSystems/Signal-iOS/pull/1027, I'm preferring the native action sheet to the DJWActionSheet for more native look and feel.

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -919,13 +919,11 @@ typedef enum : NSUInteger {
 
     switch (messageItem.messageType) {
         case TSOutgoingMessageAdapter:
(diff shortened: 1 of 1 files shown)


michaelkirk:
Closing this for now - I've got some better UX forthcoming!

michaelkirk:
OK! Found the root cause of this (presumed?) regression. I actually had never tried deleting a message before so was unfamiliar with the previously existing UX.

Issue #1103, opened 2016-03-11
Inconsistent "delete message" UX


michaelkirk:
Behavior and design of delete tools changes depending on media type.

1. Text Messages
 - can't delete (see #913)
2. Images
  - Tap image to view fullscreen
  - press "share" button
  - press "delete"
3. Videos
 - can't delete (see #671)

This is kind of a meta issue, and would be resolved *if* we do #781, but even if we don't do "multiple delete" whatever interface we have needs to be consistent regardless of media type of the message. 

The existing UI for Images isn't correct - hitting a share button to delete images doesn't make sense.

michaelkirk:
Fixed in #1105

Issue #1023, opened 2015-12-22
copy&paste images don't work


JimLoose:
when i copy an image from imessage or the photos app and paste it into a message in signal
- it either doesn't work at all
- or only pastes the image name (not the image)

2-4601:
Dupe of #793.

FredericJacobs:
Thanks @2-4601 🍭

Issue #825, opened 2015-06-19
Long pressing an image should show "Save/Copy/Delete" options


rocknjohn:
When in the conversation view, long pressing on an _unexpanded_ image should show "Save/Copy/Delete" options. This would provide a similar, intuitive behavior to the behavior that already exists for text messages where a user can long press to show "Copy/Delete" options for an individual text message.

Zahlentheorie:
+1

Additionally, can't save sent videos yet.

pstasiak:
I made some investigation. To implement this feature, some changes in pod `JSQMessagesViewController` are needed. I see that there is custom fork in `podfile` by @FredericJacobs. Do you see any potential problem in merging PR that I would make in `JSQMessagesViewController` repository to your fork? Did you made big API changes?

michaelkirk:
closed long ago in #1105

PR #1105, opened 2016-03-11
consistent Delete/Copy/Save UX across media types

michaelkirk:
Delete/Copy/Save media items via long press menu from within conversation view.

Fix for #1103 and #704

The *new* functionality is 
 * delete: video, audio, photo, and animated GIF messages
 * copy: video (as mp4), photo (as jpg), and animated GIF messages (was already possible to copy plain text message)
 * save: video (as mp4), photo (as jpg), and animated GIF messages (intentionally not possible to save plain text message or audio messages)

### Motivation

This is in line with how people expect to interact with messages (e.g. see WhatsApp and the Native Messaging app), and we were already doing it with Text messages. But for some reason we were doing these actions differently for images, and not at all for videos.

### Previous UX

In order to delete an image, first, you'd tap on the image to bring up the full screen view:

image

Then, tap "Share" button to open an action sheet, then press delete:
image

### Proposed UX

Now we're doing this consistently across message types using the edit menu, via long-press:

image

Consequently the "share" icon has been removed from the fullscreen image view:
image


rocknjohn:
Looks nice Michael. The native action sheets look much nicer. The DJW animations had felt unsettling too.

But wouldn't it make sense to have "Save/Copy/Delete" all available when long pressing on an image (https://github.com/WhisperSystems/Signal-iOS/issues/825), instead of only "Delete". These are all common actions.

michaelkirk:
@rocknjohn - I agree with you. Plus that's the behavior of native messaging app. 

Here's what would need to be done:

- [ ] Move "save" action from "share" menu to "edit" menu
- [ ] Kill the "share" menu all together until there is an actual "share" action you can do with it.
- [ ] Add a "copy" action for photos edit menu
- [ ] Add a "copy" and a "save" action for videos

I won't get to it until later this week, per other working obligations. Thanks for the feedback!

michaelkirk:
OK - some progress here. This PR is getting pretty large...

I've killed off the "share" menu all together, since the items in it are more conventionally found in the edit-menu (that's *this* guy, which you get to by long pressing).

image

The bad news is that until #1023 is complete, we won't be able to paste anything into Signal. So, silly enough, you can copy these things, but you can still only paste text into Signal at this point.

The feature feels incomplete without *also* fixing the paste situation, but this PR is already getting quite big. I'll update the description to reflect the current state of things.

Signal/src/view controllers/TSMessageAdapter.m
```diff
@@ -237,6 +241,79 @@ - (NSDate *)date {
...
+        //TODO implement save for video
+    }
+    return NO;
+}
+
+- (void)performEditingAction:(SEL)action {
+    if (action == @selector(delete:)) {
+        DDLogDebug(@"Deleting interaction with uniqueId: %@", self.interaction.uniqueId);
+        [self.interaction remove];
+    } else if (action == @selector(copy:)) {
```
michaelkirk: This is screaming for refactoring based on inheritance, but I was trying not to move too much at once.

pstasiak:
> But wouldn't it make sense to have "Save/Copy/Delete" all available when long pressing on an image (#825), instead of only "Delete". These are all common actions.

Hey, I've just created #1127 with an update of JSQMessagesViewController with copy media functionality.

greycubesgav:
+1 for this to be merged and committed to the App Store available verion.

JohanDoe:
I am using Signal 2.4.0 from the app store. The share button has indeed disappeared. But I only see copy/delete using the edit menu via long-press on an image. 

Is this intentional?

michaelkirk:
@JohanDoe - That's not intentional, and sounds like a bug.

You should be able to save anything except plain text and audio messages. Can you file a new bug and include screenshots?

JohanDoe:
@michaelkirk strange. Created a new conversation thread for the bug report screenshots and now everything is as it used to be. Can save images, also in old thread. 

Can't reproduce right now. Will keep an eye on it and if I can reproduce I will submit a bug report.

michaelkirk:
👀 

Thanks @JohanDoe!

PR #1106, opened 2016-03-12
Native actionsheets

michaelkirk:
Note: This will fail to build until #1076 is merged.

Finishing what @jaredStef  started over in #1027, this replaces DJWAction sheets with their native counterpart.
## Motivation

**Primarily**
We're using native action sheets in some places, and DJW in others. We need consistency.

**Secondarily**
- DJW are a little flashier, but I find the animations corny.
- Also, using native alert controller presentation gives us a little better keyboard hide/show behavior out of the box. (see https://github.com/michaelkirk/Signal-iOS/commit/2c51ef4fd24ec5287c6dd9a09faa7f04ae24d857)
- Also, the API for native alert controllers is much nicer. =)

I've also replaced two of the action sheets which we were using as "confirmation screens" before doing "really" destructive actions. Alerts should be used in this scenario, whereas action-sheets are for more common place interactions (e.g. when deleting something from a collection would be an appropriate time to use an action sheet). 

For a good example of where we're already doing this, see how we're using an alert when deleting an account here: https://github.com/WhisperSystems/Signal-iOS/blob/master/Signal/src/view%20controllers/SettingsTableViewController.m#L193

With this PR, I've left alone the last remaining DJWActionSheet, which gets removed in: https://github.com/WhisperSystems/Signal-iOS/pull/1105

If these are both merged, we could delete DJWActionSheet from the codebase all together. :+1:

**Translations**

Because github doesn't render .strings diffs, you can see it here: https://gist.github.com/michaelkirk/484d9e9fa23b4d8d61d8

Do I need to do anything else special to accommodate new translations? How can I get these up to transifex?
### Previously

image
Here's what the old attachment action sheet looks like
### New ActionSheet screenshots

image
Adding image to group action sheet

image
Key Change action sheet

image
Unsent message action sheet
### New Alert Screenshots

image
Clear history logs Alert

image
Resetting session Alert


Elland:
What is necessary to get this merged?

michaelkirk:
> What is necessary to get this merged?

Now it at least needs a rebase, though at this point it might be easier to start from scratch.

michaelkirk:
superseeded by #1495

Issue #1153, opened 2016-04-13
Crash when deleting last message in thread


michaelkirk:
### Bug description
Crash when deleting last message in thread

### Steps to reproduce
- long press on last element in thread
- delete

**Actual result:** App crashes (message is deleted, which you can verify by restarting the app)
**Expected result:** No Crash

### Device info
**Device**: iPhone6
**iOS version**: 9.3.1
**Signal version:** HEAD (c0bb704d2d4a55b9309777dce93fc5e786a07ed5)

michaelkirk:
Fixed with 9f572881f8c7385ca6af9baf40173830c1ca3e3a

PR #1154, opened 2016-04-13
Remove proper element from cache on delete

michaelkirk:
### Description

Fixes crash on deleting the last interaction #1153 (introduced in #1152).

Once an element is deleted, the indexOf is potentially offset, so it's
important we clear the cache with the proper interactionId.

It's going to be pretty tough to write a test for this as is. It's worth
considering extracting the data source from the controller, but that's
going to be a sizable refactor and I'm trying to minimize code churn at
this stage...

// FREEBIE

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -308,15 +308,6 @@ typedef enum : NSUInteger {
     self.inputToolbar.contentView.textView.editable = NO;
 }
 
-- (void)willDeleteThreadWithId:(NSString *)threadId {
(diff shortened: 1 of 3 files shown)


Signal/src/view controllers/MessagesViewController.m
```diff
@@ -1609,8 +1610,6 @@ - (void)yapDatabaseModified:(NSNotification *)notification {
       for (YapDatabaseViewRowChange *rowChange in messageRowChanges) {
           switch (rowChange.type) {
               case YapDatabaseViewChangeDelete: {
-                  TSInteraction *interaction = [self interactionAtIndexPath:rowChange.indexPath];
```
FredericJacobs: What happens if a deletion happens that is triggered by some other code than the one above? I think it should still be supported.

Issue #1150, opened 2016-04-12
intermittent: "Load Earlier Messages" shows for empty thread. Crashes when tapped


michaelkirk:
- [x] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description

Sometimes when I open an empty message thread I see the "Load Earlier Messages" button.

### Steps to reproduce
- tap compose, select recipient you have no message history with
- leave thread empty, return to inbox
- relaunch app and go to thread

**Actual result:** Sometimes, I see an empty thread with the "Load Earlier Messages" button displayed.
**Expected result:** An empty thread without the "Load Earlier Messages" button displayed.

### Screenshots

![image](https://cloud.githubusercontent.com/assets/217057/14464575/486b006c-009d-11e6-8dec-db3abf69299b.png)

### Device info
**Device**: iphone6
**iOS version**: 9.3.1
**Signal version:** HEAD (9d4befd)

big-r81:
I can't reproduce this yet...

The only occurence of this message is in JSQMessagesLoadEarlierHeaderView...

FredericJacobs:
I noticed that. It has to do with switching threads. Probably, need to reset those when reloading data after changing `threadId` in conversation view or something.

PR #1187, opened 2016-05-16
Reset compose view on thread switch

### LINKED ISSUE #1179: Text input field expanded when multi-line draft in other conversation thread exists
- [X ] I have searched open and closed issues for duplicates

### Bug description
"New Message" field is expanded when there is a multi-line draft in a different conversation thread. On message send the extra lines/spaces do not go through.

### Steps to reproduce
- Go to existing conversation or create new, referenced as Thread A below
- Write a multi-line message in the text input field
- Press the back arrow (top left corner)
- Go to a different conversation thread or create a new one, referenced as Thread B below

**Actual result:** 
Thread B text input field is expanded to same height of Thread A's draft. Height resumes to normal after sending Thread A's draft or it a 1 line draft.  Message sent in Thread B does not include extra spaces
**Expected result:** Thread B input field should not be expanded

### Screenshots
blob:chrome-extension%3A//bikioccmkafdpakkkcpdbppfkghcmihk/fae9740a-bde4-4a97-9116-ae7d1e375caf
blob:chrome-extension%3A//bikioccmkafdpakkkcpdbppfkghcmihk/62ecf658-43a1-4ab7-a949-fa8cce16fb85

### Device info

**Device**: iPhone 6s
**iOS version**: 9.3.1
**Signal version:** 2.3

### Link to debug log
https://gist.github.com/9701fd2dcc91094a82cf519434644fcf

michaelkirk:
### Description

Fixes #1179 to make sure text box is proper size and makes sure to pop the keyboard when draft text is present.

### DIFF
--- a/Signal/src/view controllers/UITests/SignalsViewController.m
+++ b/Signal/src/view controllers/UITests/SignalsViewController.m
@@ -324,10 +324,8 @@ static NSString *const kShowSignupFlowSegue = @"showSignupFlow";
 
 - (void)presentThread:(TSThread *)thread keyboardOnViewAppearing:(BOOL)keyboardOnViewAppearing {
     dispatch_async(dispatch_get_main_queue(), ^{
-      if (!_mvc) {
-          _mvc = [[UIStoryboard storyboardWithName:@"Storyboard" bundle:NULL]
-              instantiateViewControllerWithIdentifier:@"MessagesViewController"];
-      }
+      _mvc = [[UIStoryboard storyboardWithName:@"Storyboard" bundle:NULL]
+                instantiateViewControllerWithIdentifier:@"MessagesViewController"];
 
       if (self.presentedViewController) {
           [self.presentedViewController dismissViewControllerAnimated:YES completion:nil];
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -139,10 +139,9 @@ typedef enum : NSUInteger {
 
     if ([_thread isKindOfClass:[TSGroupThread class]] &&
         ![((TSGroupThread *)_thread).groupModel.groupMemberIds containsObject:[TSAccountManager localNumber]]) {
+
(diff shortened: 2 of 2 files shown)


Signal/src/view controllers/UITests/SignalsViewController.m
```diff
@@ -324,10 +324,8 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
 
 - (void)presentThread:(TSThread *)thread keyboardOnViewAppearing:(BOOL)keyboardOnViewAppearing {
     dispatch_async(dispatch_get_main_queue(), ^{
-      if (!_mvc) {
-          _mvc = [[UIStoryboard storyboardWithName:@"Storyboard" bundle:NULL]
-              instantiateViewControllerWithIdentifier:@"MessagesViewController"];
-      }
+      _mvc = [[UIStoryboard storyboardWithName:@"Storyboard" bundle:NULL]
+                instantiateViewControllerWithIdentifier:@"MessagesViewController"];
```
michaelkirk: FYI @FredericJacobs, I think we should roll back the "shared MessageViewsController" from this release. We've had too many separate issues as a result of it, and we're not really leveraging for any substantial advantage at this point (e.g. there is no iPad layout).

Did want to give you a chance to comment though.

big-r81: We can have a shared MessageViewController! Load the drafted message in viewDidAppear and clear the compose Textview on viewWillDissapear...

michaelkirk: Apart from this issue, it's caused a couple others (#1150, #1152), and it offers no measurable value to this release, while introducing the risk of other yet-to-be-discovered issues.

I'm not opposed to reusing the MessageViewController in theory, and see how it paves the way for a split view iPad layout, but we don't have that in this release anyway. If you'd like to submit a PR with your work, it might be useful for a future iPad layout. 

Thanks for looking into it @big-r81.

PR #1206, opened 2016-06-09
Fix new localizations

### LINKED ISSUE #1205: New localizations not accessible
Though we added new localizations for:

- Persian (fa)
- Indonesian (id)
- Macedonian (Macedonia) (mk_MK)
- Shona (sn)
- Galacian (gl)

These localizations are not being used by the app. I failed to actually mark these localizations as available before the release. People choosing these localizations will see english or whatever their supported fallback localization is. 😢 

Also, getting Farsi to look decent requires touching up our custom nav bar spacing, since it's RTL.

michaelkirk:
Fixes #1205


FredericJacobs:
I don't think this addresses the issue. The reason why we didn't have support for some of these languages in the past is because:
- Fix RTL before adding RTL translations. Iranians, Israelis, Arabs ... are used to using UIs in English and know most important words. Breaking the usability of the UI because broken RTL support is worse.
- iOS doesn't support some of those languages. Therefore, you can't set your iPhone to some of these languages. And therefore there is no way to use these localizations.
- I've been reluctant to add a switch within the app to use those languages. Because that would add another setting with additional complexity for switching languages. Not being a user or not getting constant user feedback from people using that switch didn't make it worth the investment of doing "non-standard" localization. 

Before accepting a new translation on Transifex I:
- checked that the translation was supported by iOS
- checked that the localization isn't a duplicate, or too close to an existing one.

The worst thing we can do to upset translators is get them to translate the entire app and then be like: "Sorry, we actually can't include the thing you worked hours on. But thanks!"

Hope this explains some decisions we made in the past.

michaelkirk:
> Fix RTL before adding RTL translations. Iranians, Israelis, Arabs ... are used to using UIs in English and know most important words. Breaking the usability of the UI because broken RTL support is worse.

I agree. It was pulled here: https://github.com/WhisperSystems/Signal-iOS/pull/1206/commits/89a8b61f82822e6c2fba5cfe497bb3485d0fc121

> iOS doesn't support some of those languages. Therefore, you can't set your iPhone to some of these languages. And therefore there is no way to use these localizations.

Here’s a screenshot of the new Shona translation:

image

This corresponds to this configuration in Settings.app.
image

Shona isn’t a supported “iPhone Language” (presumably this means that Apple hasn't localized all of *their* apps and underlying OS strings to Shona) but you *can* select it as a “preferred language" and then see Signal localized to Shona. Same goes for Galacian, Macedonian, and Farsi (pending RTL).

> The worst thing we can do to upset translators is get them to translate the entire app and then be like: "Sorry, we actually can't include the thing you worked hours on. But thanks!"

Agreed. I regret not getting them in the 2.3 release, but, excepting Farsi, I will have them in a point release shortly. Depending on the amount of work to get RTL looking good, Farsi may have to wait a bit longer.

FredericJacobs:
Oh, good. When we originally considered Shona, it was not yet in the "preferred language" list.

michaelkirk:
> Oh, good. When we originally considered Shona, it was not yet in the "preferred language" list.

That's likely. I think it wasn't until ios8 that most localizations become available.

PR #1238, opened 2016-06-28
Fix: No Call icon for registered Signal contact

michaelkirk:
From support:

## No Call icon for registered Signal contact

both users iPhone 6, iOS 9.3.2, Signal 2.3.2. User can receive calls, can send/receive messages, but does not see a call icon to place a call to this contact.  Have 're-registered' within app and deleted contact from iPhone and saved again to adjust format.  Still no call icon

## Description

Since all supported Signal users now have Text+Voice, there's no use carrying the distinction of RedPhone vs TextSecure throughout the app.


michaelkirk:
Not squashing this one since the individual commits provide better context for style related changes.

mawed:
What's with a standalone registration of signal desktop?

Only textsecure capability.

PR #1259, opened 2016-07-18
Fixes #130

### LINKED ISSUE #130: improve error message on hitting network that doesn't allow access to port Signal/RedPhone server requires
Several users are having a difficult time registering with the server because they are on WiFi networks that do not allow access to the port number that the Signal/RedPhone server is listening on. Switching to their cellular network usually allows them to complete the process successfully.

Perhaps another short-term solution would be to change the "We couldn't reach the Signal server" message to something like"We couldn't reach the Signal server. Please try registering from another WiFi network, or using mobile data."

PanosSakkos:
Improved error message when network doesn’t allow to access Signal port.


Signal/Images.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -77,6 +77,11 @@
       "idiom" : "ipad",
       "filename" : "launch-icon-ipad@2x.png",
       "scale" : "2x"
+    },
```
PanosSakkos: Revert this diff

Signal/Images.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -77,6 +77,11 @@
       "idiom" : "ipad",
```
PanosSakkos: Add FREEBIE in the commit message

michaelkirk:
replacing with #1285. Thanks!

PR #1269, opened 2016-07-20
Narrow the bubbles a bit.

michaelkirk:
This is closer to the 2.3 version.

Seems like upstream sizing has changed since our upgrade. This wider size calculation makes sense if you're losing space to the avatar, but since we're not using avatars the full width bubbles can make the whole view seem a bit crowded when you have large blobs of text.

Previously this:
image

Has been changed to this:
image

// FREEBIE


cascheberg:
The pictured conversation is so entertaining that I read all of it. But then I noticed that the text is cut off at the end in the new version of the bubble spacing.

michaelkirk:
@cascheberg. Woah. Good catch.

I can't reproduce it now, I'll keep a look out. 👀

PR #1292, opened 2016-07-22
Improve audio waveform scrolling performance

### LINKED ISSUE #1258: rendering Audio message gets really slow with large audio file
### Steps to reproduce
- Receive big audio message (e.g. a 6 minute Adele tune)
- Scroll back and forth past the audio message
- See the Audio Message wave form rendered

**Actual result:** There is noticeable lag, where the UI freezes before the audio message scrolls on screen.
**Expected result:** Ideally it should be quick to render *every* time, but in any case, we should at least be able to cache it so that any expensive rendering is only done once.

Another possibility here is to use the JSQ AudioPlayer, instead of the waveform generator.

vlovich:

- [X] I have read the [README](https://github.com/WhisperSystems/Signal-iOS/blob/master/README.md) and [CONTRIBUTING](https://github.com/WhisperSystems/Signal-iOS/blob/master/CONTRIBUTING.md) documents
- [X] I have signed the [Contributor Licence Agreement](https://whispersystems.org/cla/)

### Contributor checklist

 * iPhone Simulator 6S, 9.3

- - - - - - - - - -

### Description
The object is already cached & the waveform view constructed when scrolling,
so there's no need to re-create the view.

This doesn't fix the problem with the wave-form generation on the first
view of the message.  That is a more critical problem in the underlying
SCWaveformView class which renders on the main thread instead of doing
it asynchronously.

I've verified by hacking (due to being on the simulator) a large 6-minute mp3
into the chat conversation & then scrolling back & forth.  The first display still
hangs until SCWaveformView is done, but further scrolling remains smooth.

FIXES #1258

### DIFF
--- a/Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
+++ b/Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
@@ -143,14 +143,16 @@
         NSURL *url =
             [MIMETypeUtil simLinkCorrectExtensionOfFile:_attachment.mediaURL ofMIMEType:_attachment.contentType];
 
-        AVURLAsset *asset         = [[AVURLAsset alloc] initWithURL:url options:nil];
-        _waveform                 = [[SCWaveformView alloc] init];
-        _waveform.frame           = CGRectMake(42.0, 0.0, size.width - 84, size.height);
-        _waveform.asset           = asset;
-        _waveform.progressColor   = [UIColor whiteColor];
-        _waveform.backgroundColor = [UIColor colorWithRed:229 / 255.0f green:228 / 255.0f blue:234 / 255.0f alpha:1.0f];
-        [_waveform generateWaveforms];
-        _waveform.progress = 0.0;
+        if (!self.waveform) {
+            AVURLAsset *asset         = [[AVURLAsset alloc] initWithURL:url options:nil];
+            self.waveform                 = [[SCWaveformView alloc] init];
+            self.waveform.frame           = CGRectMake(42.0, 0.0, size.width - 84, size.height);
+            self.waveform.asset           = asset;
+            self.waveform.progressColor   = [UIColor whiteColor];
+            self.waveform.backgroundColor = [UIColor colorWithRed:229 / 255.0f green:228 / 255.0f blue:234 / 255.0f alpha:1.0f];
+            [self.waveform generateWaveforms];
+            self.waveform.progress = 0.0;
+        }
 
         _audioBubble = [[UIView alloc] initWithFrame:CGRectMake(0.0, 0.0, size.width, size.height)];
         _audioBubble.backgroundColor =


Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -143,14 +143,16 @@ - (UIView *)mediaView {
...
 
-        AVURLAsset *asset         = [[AVURLAsset alloc] initWithURL:url options:nil];
-        _waveform                 = [[SCWaveformView alloc] init];
-        _waveform.frame           = CGRectMake(42.0, 0.0, size.width - 84, size.height);
-        _waveform.asset           = asset;
-        _waveform.progressColor   = [UIColor whiteColor];
-        _waveform.backgroundColor = [UIColor colorWithRed:229 / 255.0f green:228 / 255.0f blue:234 / 255.0f alpha:1.0f];
-        [_waveform generateWaveforms];
-        _waveform.progress = 0.0;
+        if (_waveform == NULL) {
```
michaelkirk: style nit: `if (!self.waveform)`

Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -143,14 +143,16 @@ - (UIView *)mediaView {
...
+            AVURLAsset *asset         = [[AVURLAsset alloc] initWithURL:url options:nil];
+            _waveform                 = [[SCWaveformView alloc] init];
```
michaelkirk: I know this came before you, but prefer property access over direct ivar manipulation unless you are in an initializer. So all these `_waveform` should be replaced with `self.waveform`.

michaelkirk:
Thanks for taking a look! I just had some small style feedback before merging. If you'd like, I can fix it up, otherwise, I'll leave it to you.

Bonus points if you move the initial render to a background thread. :+1:

vlovich:
Unfortunately the render can't be moved to a background thread because of the way SCWaveformView is implemented (not well from a performance perspective).

I'd have to rewrite a large chunk of the SCWaveFormView code, get it upstreamed, & then update this code to use the newer release (which I tried & it has even worse performance than now due to reading the entire file on the UI thread).

big-r81:
Are there any reason why we use version 1.0 instead of upstream?

michaelkirk:
I can't speak for @FredericJacobs intentions, but from my own experience, it can be easier, or at least more transparent, to audit the application for stability and security when small stable dependencies are copied into your own source repository and mostly left untouched, rather than blindly updating a bunch of small dependencies all the time.

That said, if this has been fixed upstream, we should update. Are you willing to check?

big-r81:
> Are you willing to check?
Sure.

@vlovich @michaelkirk How did you get an audio file into the simulator or into signal?

michaelkirk:
Hint: Xcode Beta iOS 10 simulators can receive push notifications, thus they are able to register like a full fledged device. You just need a spare number that receive sms.

vlovich:
I can register like a full fledged device, but to get the audio file into the simulator, I hard-coded a path to a file no my computer, hacked the code to "allow" audio recordings, & hacked the audio path to always use that file rather than the mic.  Once I had a few sent, I simply profiled.  I didn't have an Android device which would be needed to test it out without changing the iOS code because that's the only way to send an audio file currently.

Newer upstream doesn't have this problem fixed.  In fact, as far as I could tell from a quick browse of the performance, it's massively worse.

vlovich:
Anything else needed to merge this in?

michaelkirk:
> Anything else needed to merge this in?

Code looks good - just need to test it and merge it in. Thanks for you patience.

michaelkirk:
Thanks @vlovich! :+1:

PR #1299, opened 2016-08-06
Respect users name display settings

### LINKED ISSUE #1158: Order for Japanese names
Hi,

I am using Signal in Japanese and noticed that the family name and the given name are displayed in the wrong order. For Japanese names written in Japanese, the family name should come before the given name. Moreover, the space between the family name and the given name can be omitted.

For example, if the given name is G and the family name is F, Signal currently shows
G F
but it should be either
F G
or
FG

FG (the space being omitted) is better if it is written in a sentence (e.g., "Missed a call from xxx") since Japanese sentences don't use space. It is safer to show FG everywhere, which is what is done in  Apple's Mail and iMessage.

Since the reverse order appears quite awkward and may give people a bad impression, I would like to ask you to look into this problem.

The problem appears in iPhone 6s with iOS 9.3.1 and Signal 2.2.

Thank you.

abcdev:

### Contributor checklist

  - iPhone 5s, iOS 10 beta 4
- [x] I have made the choice whether I want the [BitHub reward](https://github.com/WhisperSystems/Signal-Android/wiki/BitHub-Rewards)

---
### Description

Fixes #1158 by respecting users name display settings. Depends on [WhisperSystems/SignalServiceKit/pull/22](https://github.com/WhisperSystems/SignalServiceKit/pull/22)

### DIFF
--- a/Signal/src/views/ContactTableViewCell.m
+++ b/Signal/src/views/ContactTableViewCell.m
@@ -31,7 +31,7 @@
         [[NSMutableAttributedString alloc] initWithString:contact.fullName];
 
     BOOL firstNameDisplay = ABPersonGetCompositeNameFormat() == kABPersonCompositeNameFormatFirstNameFirst ? YES : NO;
-    BOOL firstNameSortOrdering = ABPersonGetSortOrdering() == kABPersonCompositeNameFormatFirstNameFirst ? YES : NO;
+    BOOL sortByFirstName = ABPersonGetSortOrdering() == kABPersonSortByFirstName ? YES : NO;
 
     UIFont *firstNameFont;
     UIColor *firstNameFontColor;
@@ -41,7 +41,7 @@
     UIColor *lastNameFontColor;
     NSRange lastNameRange;
 
-    if ((firstNameSortOrdering && contact.firstName) || !contact.lastName) {
+    if ((sortByFirstName && contact.firstName) || !contact.lastName) {
         firstNameFont = [UIFont ows_mediumFontWithSize:_nameLabel.font.pointSize];
         firstNameFontColor = [UIColor blackColor];
         lastNameFont  = [UIFont ows_regularFontWithSize:_nameLabel.font.pointSize];
--- a/Signal/src/views/ContactTableViewCell.m
+++ b/Signal/src/views/ContactTableViewCell.m
@@ -30,34 +30,53 @@
     NSMutableAttributedString *fullNameAttributedString =
         [[NSMutableAttributedString alloc] initWithString:contact.fullName];
 
-    UIFont *firstNameFont;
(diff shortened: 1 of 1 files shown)


Signal/src/views/ContactTableViewCell.m
```diff
@@ -30,34 +30,53 @@ - (NSAttributedString *)attributedStringForContact:(Contact *)contact {
     NSMutableAttributedString *fullNameAttributedString =
         [[NSMutableAttributedString alloc] initWithString:contact.fullName];
 
+    BOOL firstNameDisplay = ABPersonGetCompositeNameFormat() == kABPersonCompositeNameFormatFirstNameFirst ? YES : NO;
+    BOOL firstNameSortOrdering = ABPersonGetSortOrdering() == kABPersonCompositeNameFormatFirstNameFirst ? YES : NO;
```
michaelkirk: `ABPersonGetSortOrdering` doesn't return `kABPersonCompositeNameFormatFirstNameFirst`

See https://developer.apple.com/reference/addressbook/1619730-sort_order

Signal/src/views/ContactTableViewCell.m
```diff
@@ -30,34 +30,53 @@ - (NSAttributedString *)attributedStringForContact:(Contact *)contact {
     NSMutableAttributedString *fullNameAttributedString =
         [[NSMutableAttributedString alloc] initWithString:contact.fullName];
 
+    BOOL firstNameDisplay = ABPersonGetCompositeNameFormat() == kABPersonCompositeNameFormatFirstNameFirst ? YES : NO;
```
michaelkirk: `BOOL displayGivenNameFirst = ABPersonGetCompositeNameFormat() == kABPersonCompositeNameFormatFirstNameFirst`

Signal/src/views/ContactTableViewCell.m
michaelkirk: `BOOL sortByFirstName` would be clearer

michaelkirk:
It seems to me like it would be clearer to build two [attributed strings passing the attributes in a dictionary](https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Classes/NSAttributedString_Class/#//apple_ref/occ/instm/NSAttributedString/initWithString:attributes:), then concatenating the strings, rather than dealing with the NSRanges/length calculations.

Signal/src/views/ContactTableViewCell.m
michaelkirk: Should be `kABPersonSortByFirstName`

abcdev:
@michaelkirk I tried to stay as close as possible to the original code. But I could change that of course.

PR #1305, opened 2016-08-15
Enable screen security by default

michaelkirk:
When enabled, the screen security option obscures the app in the application switcher.


rmueller83:
Isn't this contrary to https://github.com/WhisperSystems/Signal-Android/issues/5529 / https://github.com/WhisperSystems/Signal-Android/pull/5587 ?

michaelkirk:
From the linked issues:

> It's quite common to see non-passphrase users wonder why screenshots in
Signal [Android] don't work.

Screen protection in iOS does not block screenshots. It's only about the
task switcher, so I don't think the issues addressed are relevant.

PR #1332, opened 2016-09-18
code verification: show complete phone number

nixnuex:
### Contributor checklist
 * iPhone 6S, iOS 9.3
- [X] I have made the choice whether I want the [BitHub reward](https://github.com/WhisperSystems/Signal-Android/wiki/BitHub-Rewards)

- - - - - - - - - -

### Description
During account creation in the verification code step the user should be presented with his complete international phone number (including country code), entered in the previous step.

### DIFF
--- a/Signal/src/view controllers/CodeVerificationViewController.m
+++ b/Signal/src/view controllers/CodeVerificationViewController.m
@@ -45,7 +45,8 @@
 - (void)viewWillAppear:(BOOL)animated {
     [super viewWillAppear:animated];
     [self enableServerActions:YES];
-    [_phoneNumberEntered setText:_formattedPhoneNumber];
+    [_phoneNumberEntered setText:
+        [PhoneNumber bestEffortFormatPartialUserSpecifiedTextToLookLikeAPhoneNumber:[TSAccountManager localNumber]]];
     [self adjustScreenSizes];
 }
 


michaelkirk:
I'm having a hard time understanding why this is desirable. Can you help me understand the goal?

The user typed in their number as they would recognize it, and given the context of their locale, that should be sufficient for us to unambiguously determine their e164 for internal purposes. How is it helpful to show them the (potentially alien) e164 format when that's not what they typed in?

nixnuex:
Thanks for your comment. The purpose of this commit is a better user experience. In some countries, it is common to either use the non-international phone number like 0151 12345678 for non-international usage or the drop the zero and use the international one +49 151 12345678. When the user is asked for country code and phone number in the verification step, he/she will select the country code +49 from the list and enter the phone number without leading zero: 151 12345678.
If this last part is then presented to the user on the "enter verif code" dialogue, he/she might be confused because standing on it's own, _it is not a valid phone number_. This could trigger e.g. an unwanted return to the previous page.
This commit results in the phone number being presented in the "nicest" possible way, like +49 151 12345678 and not in e164 format like +4915112345678.

michaelkirk:
Right. This sounds reasonable. I'll have a look a little later this week.

michaelkirk:
Thanks!

PR #1333, opened 2016-09-18
script to generate global .strings file

### LINKED ISSUE #1331: Generate localizable strings with `genstrings`
Our current manual process for managing localizable strings is a pain.

1. time consuming
2. error prone
3. redundant
4. doesn't remove unused strings
5. makes it less likely that translation hints get updated

Our current source strings live in a single file, at https://github.com/WhisperSystems/Signal-iOS/blob/master/Signal/translations/en.lproj/Localizable.strings

When new strings are needed they are added manually to that file (mostly manually anyway, see `Signal/translations/bin/genstrings-for-file` for a small step in automation). Recently I've started also including translation context hints to NSLocalizedString, and also manually copying them into our strings file. What a waste! 

The proposed process should populate en.lproj with string file(s) containing all the strings used in Signal-iOS, SignalServiceKit, and JSQMessagsViewController.

Pull requests welcome!

nixnuex:
### Contributor checklist
 * OS X 10.11.6
- [X] I have made the choice whether I want the [BitHub reward](https://github.com/WhisperSystems/Signal-Android/wiki/BitHub-Rewards)

- - - - - - - - - -

### Description
Tool to automize the generation of a .strings file for localization.

 fixes #1331


michaelkirk:
This is a start!  And already an improvement if we go through the effort of re-ordering our strings file. 

The biggest pain is genstrings blows away your existing translations. e.g. after running this you see:

```
-"ACTION_REQUIRED_TITLE" = "Action Required";
+"ACTION_REQUIRED_TITLE" = "ACTION_REQUIRED_TITLE";
```

Have you done anything like this before? http://www.stevestreeting.com/2010/05/18/os-x-localisation-incremental-genstrings-and-utf-8-files/

Signal/translations/bin/auto-genstrings
```diff
@@ -0,0 +1,13 @@
...
+OUTPUT_DIR="Signal/translations/en.lproj/"
+
+if [ ! -d "Signal/src" ]; then
+  echo "Please run this tool from the repository's base directory"
+  exit 1
+fi
+
+mkdir -p "${OUTPUT_DIR}"
+
+find $TARGETS -name "*.m" -print0 | xargs -0 genstrings -a -o "$OUTPUT_DIR"
```
michaelkirk: I don't think that we want the `-a` as it will just duplicate all the existing strings to the end of the file.

nixnuex:
I see. The golden solution should append only new strings to the end of the file, leaving previously translated intact. The `-a` option is the only way to directly get everything into one file, as far as I know. `genstrings` is really not that versatile. I will have a look how to build something on top of it.

michaelkirk:
If it makes it easier, having one file isn’t really a requirement - we could easily upload multiple translatable resources to our translation service (transifex).

Here are the goals as I see them:
- making sure I didn’t omit any strings from source
- comments in .strings are kept up to date with what is in the source
- strings are removed from .strings if they no longer appear in the source

Ideally all of that without having to manually undo all the overwriting genstrings does by default.

nixnuex:
I rewrote the script. It is now capable of generating an up-to-date .strings file and merge the existing definitions into it. It keeps the format including comments as it is generated by genstrings.

Hope it helps!

michaelkirk:
OH MY GOD🙀 😍 😍 😍 

A savior descended and she was known as awk!

Thanks for this, really! It might be a minute before I can review it, but super excited to incorporate it into our work flow.

michaelkirk:
I was auditing this today, and found that "END_CALL_BUTTON_TITLE" was missing from the output strings.

Something funny about that string is it's defined in LocalizableText.h, which assigns NSLocalizedStrings to constants, then those constants are used. I think it's sort of... from a different era, so I'm open to a solution which blows away those constants and inlines the strings if that fixes the issue.

michaelkirk:
nit: could you make the script executable?

nixnuex:
Adjusted the file permission.

Concerning the strings defined in LocalizableText.h: As this tool is heavily based on Apple's genstrings command, which does not seem to parse comment/precompiler lines in the code, I fear this is the wrong place to fix this issue. I think the cleaner solution would be to inline the strings and make them accessible to genstrings.

michaelkirk:
> I think the cleaner solution would be to inline the strings and make them accessible to genstrings.

Agreed. Are you planning to do that, or waiting on me? Let me know either way.

nixnuex:
I won't have time for the next couple of days, so I would appreciate if you could take care of this.

michaelkirk:
I made some progress on this, it looks good so far, but I want to leave the boring task of auditing the strings file until tomorrow when my brain is fresh.

michaelkirk:
This is getting closer, but I'm still seeing these two oddities:

Some existing translations not found, maybe because of the lack of a space around the `=` in the original? Honestly I could skip this one assuming that going forward errors like this will be rare because we won't be manually editing the strings file.

```
-/* Activity indicator title, shown upon returning to the device manager, until you complete the provisioning process on desktop */
-WAITING_TO_COMPLETE_DEVICE_LINK_TEXT="Complete setup on Signal Desktop.";
+/* Activity indicator title, shown upon returning to the device manager, until you complete the provisioning process on desktop */
+"WAITING_TO_COMPLETE_DEVICE_LINK_TEXT" = "WAITING_TO_COMPLETE_DEVICE_LINK_TEXT";
```


Though this one is more severe, seems like multi-line strings aren't correctly captured:
```
+/* Paragraph(s) shown alongside keying material when verifying privacy with {{contact name}} */
+"PRIVACY_VERIFICATION_INSTRUCTIONS" = ";
-/* Paragraph(s) shown alongside keying material when verifying privacy with {{contact name}} */
-"PRIVACY_VERIFICATION_INSTRUCTIONS" = "If you wish to verify the security of your end-to-end encryption with %@, compare the numbers above with the numbers on their device.
.
```

Also, I found two unlocalized strings because of this. =)

nixnuex:
Thanks for digging into this. The first example `WAITING_TO_COMPLETE_DEVICE_LINK_TEXT` indeed fails because of the missing quotes in the original file. Missing spaces around the `=` are caught. I would not change this behavior.

For the second example: Yes, multiline strings are a problem. All lines are treated individually. Would it be an option to replace the line breaks in the original localized strings by `\n`? If not, I could try  to work around by replacing line breaks in the original and switching back to real line breaks in the new .strings file.

PR #1336, opened 2016-09-20
Fix groups breaking without avatar

michaelkirk:
- Group updates were appearing as "unsupported attachment"
- Couldn't message in a new group unless it had attachment.

// FREEBIE

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -158,12 +158,13 @@ typedef enum : NSUInteger {
         return;
     }
 
-    if ([_thread isKindOfClass:[TSGroupThread class]] &&
-        ![((TSGroupThread *)_thread).groupModel.groupMemberIds containsObject:[TSAccountManager localNumber]]) {
-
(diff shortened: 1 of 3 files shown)


Signal/src/view controllers/MessagesViewController.m
```diff
@@ -158,12 +158,13 @@ - (void)hideInputIfNeeded {
...
     }
 
-    if ([_thread isKindOfClass:[TSGroupThread class]] &&
-        ![((TSGroupThread *)_thread).groupModel.groupMemberIds containsObject:[TSAccountManager localNumber]]) {
-
-        [self inputToolbar].hidden = YES; // user has requested they leave the group. further sends disallowed
-        [self.inputToolbar endEditing:TRUE];
-        self.navigationItem.rightBarButtonItem = nil; // further group action disallowed
+    if ([_thread isKindOfClass:[TSGroupThread class]]) {
+        TSGroupThread *groupThread = (TSGroupThread *)self.thread;
```
michaelkirk: this is just a readability change. Bug fix is in SSK. https://github.com/WhisperSystems/SignalServiceKit/pull/31

PR #1369, opened 2016-10-05
disappearing messages

michaelkirk:


cottsak:
So I'm reading https://whispersystems.org/blog/disappearing-messages/ and I can't satisfy my burning curiosity:

> Who is the agent that is responsible for the destruction of all copies ("end points" in the end-to-end-secure context) of disappearing messages?

It's obviously not a 'server', like in the case of Snapchat I suspect, since this is end-to-end encrypted thus a centralised agent has no context to perform those responsibilities. So who does it in Signal's implementation?

And:

> How, as a user, can I be confident that the "other client" will actually carry out the destruction of said message?

I'm asking this question here since I don't know where else to put it local to the context of the change. I also don't know Obj-C well enough to easily capture the answer myself from the code.

Can anyone @michaelkirk @moxie0 help me understand how this works?

michaelkirk:
Here's the key paragraph from the blog post

> Disappearing messages are a way for you and your friends to keep your message history tidy. They are a collaborative feature for conversations where all participants want to automate minimalist data hygiene, not for situations where your contact is your adversary — after all, if someone who receives a disappearing message really wants a record of it, they can always use another camera to take a photo of the screen before the message disappears.

So what we're saying is, even when you trust the remote participant to *receive* your message, there is value in helping them delete it after a while. Most scenarios can be reduced to: If you lose control of your phone, or your partner loses control of their phone, it would be great if the device didn't have years of your correspondence stored on it.

> @cottsak said: Who is the agent that is responsible for the destruction of all copies ("end points" in the end-to-end-secure context) of disappearing messages?

Each client is responsible for deleting their own local copy of the message, using the timer duration included in the contents of the trusted message they received.

> @cottsak said:  How, as a user, can I be confident that the "other client" will actually carry out the destruction of said message?

If you *don't* trust the remote recipient, then you shouldn't be messaging them anything you would want to revoke in the first place. Regardless of any super-crypto-hacker-drm scheme which "makes sure" the message is deleted from the remote phone, if the recipient is an adversary, they can just snap a picture from a different phone. Why bother with any byzantine system so trivially circumvented?

So, to rephrase, this feature doesn't allow you to prevent an adversarial recipient from keeping a recording of the message, it's a tool to help you and your friends keep your message history tidy in order to mitigate the damage of a future compromise. That's the value proposition.

If you'd like to discuss this further, please take it to the forum: https://whispersystems.discoursehosting.net/t/automatically-disappearing-messages/473

PR #1387, opened 2016-10-12
timer animation too fast

### LINKED ISSUE #1384: Timer animation is too fast
Messages look like there are almost out of time even when they're only 1/2 done

michaelkirk:
fixes #1384


michaelkirk:
merged in https://github.com/WhisperSystems/Signal-iOS/pull/1389

PR #1388, opened 2016-10-12
restart animations when returning from background

### LINKED ISSUE #1385: timers appear empty when switching back from another app
### Steps to reproduce
- Set disappearing timer to 1 day
- Send some messages
- Double tap home to switch to any other app
- Double tap home to return to Signal

**Actual result:** Timers appear empty
**Expected result:** Timers should be appropriately filled.

michaelkirk:
fixes #1385


michaelkirk:
merged in https://github.com/WhisperSystems/Signal-iOS/pull/1389

michaelkirk:
fixed in https://github.com/WhisperSystems/Signal-iOS/commit/da82c01f6e00486e002a3bedac930a5e63ff71dc

PR #1399, opened 2016-10-15
Attempt to make travis more reliable

michaelkirk:
Since XCode8 things have been really flaky. Some folks claim the SE
simulator launches faster, so giving that a try.

// FREEBIE


michaelkirk:
Ha. Apparently not.

PR #1400, opened 2016-10-17
Fix issue 757

mdouglass:
### First time contributor checklist

- [x] I have read the [README](https://github.com/WhisperSystems/Signal-iOS/blob/master/README.md) and [CONTRIBUTING](https://github.com/WhisperSystems/Signal-iOS/blob/master/CONTRIBUTING.md) documents
- [x] I have signed the [Contributor Licence Agreement](https://whispersystems.org/cla/)

### Contributor checklist

- [x] I'm following the [code, UI and style conventions](https://github.com/WhisperSystems/Signal-iOS/blob/master/CONTRIBUTING.md#code-conventions)
- [x] My commits are rebased on the latest master branch
- [x] My commits are in nice logical chunks
- [x] My contribution is fully baked and is ready to be merged as is
- [x] I have tested my contribution on these devices:
 * iPhone 7, iOS 10.0.3
 * iPhone 6S, iOS 9.3
- [x] I have made the choice whether I want the [BitHub reward](https://github.com/WhisperSystems/Signal-Android/wiki/BitHub-Rewards) or not by omitting or adding the word `FREEBIE` in my commit message

- - - - - - - - - -

### Description
Second half of the fix for #757. The first half of this fix is in SignalServiceKit. This half makes sure that local changes are given proper details.

PR #1403, opened 2016-10-18
Scanning an invalid QR code shows an error dialog

### LINKED ISSUE #1347: Scanning non-linking code should give a useful response
### Bug description

When you're on the link desktop screen and you scan some random code, you don't get any good feedback that the code you scanned is invalid. 

### Steps to reproduce
- Running at least 2.5.0(currently in beta)
- Go to settings > linked devices > link new device
- Scan some random QR code, e.g. a fingerprint QR code.

**Actual result:**
- See "Link this Device?" alert.
- When you tap to confirm, the alert is dismissed and you're still looking at the QR Scanning screen.

**Expected result:** 

### Should see alert

- **Title:** Invalid Linking Code.
- **Description:** The code you scanned doesn't look like a valid linking code.
- **Actions** 
  - **Retry:** Dismisses the alert and restarts the capture session.
  - **Cancel:** Pops you back to the "Linked Devices" manager screen.

mdouglass:
### Contributor checklist

 * iPhone 7, 10.0
 * iDevice 6s, 9.3

- - - - - - - - - -

### Description
From the error dialog you are able to back out of scanning or try again. Adds English localization for this message, which will probably require another pass at explanatory text is not always my forte.

Adds an extra flag in OWSQRCodeScanningViewController to make sure that we don’t handle captureResults when capture is disabled (this was racing before because the call to [capture stop] is async and that was causing the cancel case to present the error dialog multiple times.

Fixes #1347 

// FREEBIE


mdouglass:
I'd actually seen that as well and had debugged into it since it didn't make any sense to me that the scan was working on first run. It turns out that the call to `self.capture.delegate = self;` is implicitly a call to `start` (https://github.com/TheLevelUp/ZXingObjC/blob/master/ZXingObjC/client/ZXCapture.m#L139).

I'm happy to make the change anyway if you'd like since it will make it clearer what's going on (and add the atomic at the same time).

mdouglass:
All the travis tests seem to be failing with a similar linking error. Is this a known issue?

michaelkirk:
> All the travis tests seem to be failing with a similar linking error. Is this a known issue?

Since upgrading to XCode8 earlier this month it's common that somewhere between building the tests and launching them the build fails. This is exacerbated during peak hours.

It might be related to: http://github.com/travis-ci/travis-ci/issues/6422 which I've been following. In the meanwhile, I have the unenviable responsibility of looking at every failed build, and if it fails in this particular way I restart it. 

Travis is passing now 👍 .
But the CLA is not 😞. I'll take your CLA from https://github.com/WhisperSystems/Signal-iOS/pull/1400 as good enough.

mdouglass:
Do I need to be redoing the CLA each time? I thought that would just be with the first commit.

michaelkirk:
> On Oct 19, 2016, at 2:51 PM, Matthew Douglass <notifications@github.com> wrote:
> 
> Do I need to be redoing the CLA each time? I thought that would just be with the first commit.
> 
No, you don’t need to redo the CLA. Our CLA server sometimes sends false negatives as it has here.

michaelkirk:
Can you rebase and update?

You'll likely have conflicts in the Localizable.strings which was largely reorganized recently. My advice would be to just run `Signal/translations/bin/auto-genstrings` after rebasing.

mdouglass:
No problem, will do.  Are there any gotchas or things to know with running the genstrings tool or is it pretty automatic.

mdouglass:
I had a little weirdness with the genstrings tool (it kept reporting: `Bad entry in file Pods/JSQMessagesViewController/JSQMessagesViewController/Categories/NSBundle+JSQMessages.m (line = 39): Argument is not a literal string.`).  But I believe I was able to reapply my changes correctly and everything seems to working correctly.

I can't get the Localizable.strings to diff though -- is there a reason they are UTF-16LE instead of UTF-8 documents? I thought the recommendation was to store them as UTF-8 now.

michaelkirk:
Should just work. In any case it's under revision control so you can't do
much damage. Give it a shot.

michaelkirk:
In the immediate, you could add something like this to your gitconfig.

```
[diff "localizablestrings"]
    textconv = "iconv -f utf-16 -t utf-8”
```

I suppose the reason we’re using it is that `genstrings` generates UTF-16 by default, but since we’ve got our own wrapper around genstrings anyway, it probably makes sense to just store it in UTF-8. Then we could even see string diffs in GH. What a day that will be.

Let’s stick with UTF-16 for this PR though.

mdouglass:
I think we crossed on that last comment.  The PR is already updated.

nixnuex:
> I can't get the Localizable.strings to diff though -- is there a reason they are UTF-16LE instead of UTF-8 documents? I thought the recommendation was to store them as UTF-8 now.

Just a side remark since you brought it up: I left it as UTF16 since genstrings generates UTF16 files. I went back and figured out, that Apple now recommends UTF8 and automatically converts it to UTF16:

>[Resource Programming Guide->String Resources](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html) Note: It is recommended that you save strings files using the UTF-8 encoding, which is the default encoding for standard strings files. Xcode automatically transcodes strings files from UTF-8 to UTF-16 when they’re copied into the product.

I propose a change to the infant auto-genstrings and will take care of it.

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/bbfffdf79f0ac374ee3c811006e224c49a55e850

Should I interpret your flipping the endianness of the Localizable.strings file as a strong vote in support of UTF-8?

mdouglass:
No, damn, didn't even realize I had done that.

PR #1407, opened 2016-10-19
Unifies bubble sizes for media bubbles

### LINKED ISSUE #1270: Bubble for video attachment should render with same AR as video
When I am sent a video with portrait aspect ratio, I see a landscape bubble, but I should see a bubble with portrait aspect ratio, as I do with image attachments.

Here's a screenshot of a video with landscape aspect ratio followed by a video with portrait aspect ratio:

image

vs receiving images which produce bubbles properly shaped to the aspect ratio of their media content.
screen shot 2016-07-20 at 5 57 25 pm

However, the photos look mis-aligned because of their different width.
- [ ] Width should be fixed, and uniform.
  - [ ] photo
  - [ ] GIF
  - [ ] Video
- [ ] Height should be determined by aspect ratio, cropping at some point if it's a really tall message.
  - [ ] photo
  - [ ] GIF
  - [ ] Video

mdouglass:
### Contributor checklist

  - iPhone 7, iOS 10.0
  - iPhone 6, iOS 9.3
  - iPhone 4s (simulator), iOS 9.3
---
### Description

Media bubbles (for photo, video and animated) sizes are calculated using shared logic. The bubbles are fixed width and their height is calculated based on the aspect ratio of the underlying image (clamped to a reasonable min/max height).

Fixes #1270

// FREEBIE


michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/bbfffdf79f0ac374ee3c811006e224c49a55e850

Should I interpret your flipping the endianness of the Localizable.strings file as a strong vote in support of UTF-8?

michaelkirk:
On an iphone5:

Before:
image

After:
image

You've done as specified, but the result is that we're not leveraging as much of our screen as we should. What used to be the max width for vertical media, appears to now be the max width for all media.  This is especially noticeable with the horizontal video, which is all together too small.

I think matching the width from the "before" version would better utilize our screen.

However that would exacerbate the opposite problem with the vertical video. If we allowed it to be that width, while maintaining AR, a single message would cover most of the screen, and likely feel too big.

Taking a cue from other apps, maybe we should fudge the aforementioned strict AR requirement. e.g slightly crop the media so that what we end with is something vaguely golden rectangle, rather than the true AR, while still showing the difference between portrait and landscape media.

What do you think?

Signal/src/Models/TSMessageAdapaters/TSAnimatedAdapter.m
```diff
@@ -130,41 +130,4 @@ - (void)performEditingAction:(SEL)action
     }
 }
```
michaelkirk: I love a PR with more red than green. =)

mdouglass:
Dis some comparisons of Messages behavior with the new behavior.  On iPhone 7 they are fairly similar (not surprising since I based the PR on that).  But on the older iPhone there is a definite divergence.

Based on your comments and these comparisons, I'd suggest:
1) Make the iPhone 7 bubble about 10% wider than in the original PR, but otherwise maintain the fixed width, variable height model.
2) Make the iPhone 5 bubble work differently. In whichever dimension is larger use a fixed size and float the other dimension to match the aspect ratio. I think that should give a better feeling on the smaller screen. 

Messages - iPhone 4
messages - iphone 4

Messages - iPhone 7
messages - iphone 7

Signal - iPhone 7
signal - iphone 7

mdouglass:
Updated per my earlier comments.  See screenshots:

iPhone 5s (simulator)
simulator - iphone 5s

iPhone 7 (simulator)
simulator - iphone 7

michaelkirk:
Looks good - Thank you!

PR #1420, opened 2016-10-26
convert localizable strings to utf-8

michaelkirk:
Motivation
- no special config to diff translated strings
- don't worry about endianness
- see string diffs on github

/cc @nixnuex @mdouglass


noahemmet:
Sounds like you've got it covered, but I've also had luck with this approach: http://imnotyourson.com/diffable-and-mergeable-localizable-strings/

PR #1425, opened 2016-10-28
fix registration without background modes and keep push notifications in sync

michaelkirk:
See details in commits.


michaelkirk:
merged in https://github.com/WhisperSystems/Signal-iOS/commit/62f9606bf5f59ffcde5da88ecac2c9fde1f4843c

PR #1479, opened 2016-11-21
Fixes #1390: Make settings menu more obvious

### LINKED ISSUE #1390: Settings menu should be more obvious for contacts
Plenty of folks are having trouble finding our conversation settings menu.

Also related is that some folks literally *can't* get to the settings, e.g. tapping on the contact name doesn't do anything.

https://github.com/WhisperSystems/Signal-iOS/issues/1402
https://github.com/WhisperSystems/Signal-iOS/issues/788

Likely because we're detecting and strapping behavior to the native navbar title, which sometimes fails. We probably won't have this issue if we're defining our own title view.

jschneier:
### Contributor checklist

 * iPhone SE, iOS 10.1

- - - - - - - - - -

### Description

* Always show settings button
* Remove timer button

See the linked issue. The timer link goes to the settings anyway and we are already using the `...` for group messages so it make sense to just extend that to all convos. I'm still not sure if the phone should be the far right item or not for consistency with group contacts. I also think there is a bit too much space between the icons but that might just be me.

I'm open to other ideas if the timer icon link is something we 100% want. You need to get to settings to configure it anyway so the discoverability of that feature is pretty poor at the moment.

Also, how do we feel about making it a modal for consistency with the other settings.

iPhone SE screenshot:
img_0753

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -501,23 +501,17 @@ typedef enum : NSUInteger {
                                                                       action:@selector(callAction)];
         callButton.imageInsets = UIEdgeInsetsMake(0, -10, 0, 10);
         [barButtons addObject:callButton];
-    } else if ([self.thread isGroupThread]) {
-        UIBarButtonItem *manageGroupButton =
-            [[UIBarButtonItem alloc] initWithImage:[UIImage imageNamed:@"contact-options-action"]
-                                             style:UIBarButtonItemStylePlain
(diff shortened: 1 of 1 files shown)


michaelkirk:
Thanks for braving ahead with this.

> The timer link goes to the settings anyway

It serves as a status icon to show when disappearing messages are disabled, so I'm inclined to keep it around.

I went for a different approach here: #1498

charlesmchen:
Fixed by #1773 which will be available in our next release.

Issue #624, opened 2015-03-02
Contact listed twice in new message dialog


Schabernack:
One contact is listed twice in the signal new message dialog. The contact is actually the merged view of two different datasources (icloud, caldav) for the same person (edit: linked contacts: link contacts), both with the phone number associated with it.  

If the contact is mentioned only once in the apple contact app, i feel signal should align and also show only the merged contact.

hubert3:
+1, also seeing this on a few contacts

corbett:
the user curates their contacts within apple's native contact app
currently. That means if there are two contacts in the native app with the
same phone number, and that phone number is a signal user, the contact will
appear twice. It's intentional currently but something that also frustrates
me as I have many duplicates in the native app.  we'll discuss what to do
about the case further.

On Mon, Mar 2, 2015 at 4:34 PM, hubert3 <notifications@github.com> wrote:

> +1, also seeing this on a few contacts
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/WhisperSystems/Signal-iOS/issues/624#issuecomment-76862512>
> .
>



-- 
Dr. Corbett Moran
Physics @ http://www.tapir.caltech.edu <http://www.ICS.uzh.ch>
Dev @ http://circleof6app.com
Dev @ https://whispersystems.org
www.christinecorbettmoran.com

bs:
Seeing this also. Problematic if one uses their phone to manage contacts from multiple sources, which I imagine many users do. (gmail, fb, icloud, ...)

michaelkirk:
Fixed in next release with https://github.com/WhisperSystems/Signal-iOS/commit/06ca3c92909e59c47320a9cc2bb6333222b2368e#diff-8418c8ea3481450157e2ef89c76ed1c1R382

Issue #720, opened 2015-03-16
Multiple AddressBooks Support


mmkassem:
When using multiple contacts accounts (ex. iCloud, Google, ...etc), a contact info can be duplicated and appears more than once in Signal. It is better to show only one contact per phone number especially if both the name and the phone number match

riyapenn:
Similar to https://github.com/WhisperSystems/Signal-iOS/issues/624

michaelkirk:
Fixed in next release with https://github.com/WhisperSystems/Signal-iOS/commit/06ca3c92909e59c47320a9cc2bb6333222b2368e#diff-8418c8ea3481450157e2ef89c76ed1c1R382

PR #1485, opened 2016-11-23
Do not list duplicate contacts (#624, #720)

roostr:
### Contributor checklist

 * iPhone 6s, iOS 10.1.1

- - - - - - - - - -

### Description
When starting a new conversation, I often see duplicate entries for many of my contacts, because I use multiple contact management backend systems and iOS is merging my contacts (but also reporting them as distinct AddressBook records). This patch filters out all contacts that have the same first name, last name, and list of phone numbers. It sufficiently deduplicates my list for me.

Commit message:

Currently, users who have "linked contacts" (contacts that are the same
person, but from different sources [like iCloud, Facebook, Google,
CalDAV] that iOS links together) can see multiple entries for
the same contact. This patch will deduplicate contacts that have the
same name and phone number, which resolves all duplications on my
device.

As #720 appears to be a 100% dupe of #624, this fixes both.

### DIFF
--- a/Signal/src/contact/OWSContactsManager.m
+++ b/Signal/src/contact/OWSContactsManager.m
@@ -255,9 +255,71 @@ void onAddressBookChanged(ABAddressBookRef notifyAddressBook, CFDictionaryRef in
     CFRelease(allPeople);
     NSArray *filteredContacts = [sortedPeople filteredArrayUsingPredicate:predicate];
 
-    return [filteredContacts map:^id(id item) {
+    NSArray *contacts = [filteredContacts map:^id(id item) {
       return [self contactForRecord:(__bridge ABRecordRef)item];
     }];
+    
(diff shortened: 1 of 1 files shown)


michaelkirk:
Thanks for looking into this, but we may have worked past eachother.  Is this still an issue as of: https://github.com/WhisperSystems/Signal-iOS/commit/06ca3c92909e59c47320a9cc2bb6333222b2368e#diff-8418c8ea3481450157e2ef89c76ed1c1R382 (yesterday).

I de-duped on just e164 phone number, as threads are unique per e164.

e.g. if Alice and Bob have the same phone number, Signal considers them the same person. You can't have one thread with Alice and another with Bob.

roostr:
Ahh yes, I missed your recent work. I just pulled your commit, and I can confirm it fixes #624 and #720. Looks nice btw, I like the avatars and the new Invite flow.

Issue #1402, opened 2016-10-17
Can't get to conversation settings for some contacts


michaelkirk:
- Enter conversation
- Tap on conversation title

_Expected behavior:_ push to conversation settings screen.

_Actual behavior:_ nothing happens.

This seems to only affect certain conversation threads.

2-4601:
Haven't seen/used Signal iOS for months but could this be related to #788?

michaelkirk:
Thanks for your eternal vigilance. =)

I wouldn’t be surprised if it *is* related as I re-purposed the existing gesture recognition code.

MarcoZehe:
Note that double-tapping on the conversation header with VoiceOver on, which simulates a normal tap gesture, never opens the conversation settings in any conversation for VoiceOver users. Might be related to this.

riyapenn:
Current work around for users is to remove the last name of the contact.

1. Go to contact's entry in iPhone contact's list
2. Edit entry to remove the last name of this contact
3. Save the change
4. Open Signal
5. Select existing conversation thread or create new conversation thread with contact
6. Tap on contact name

riyapenn:
Log for user where this was an issue and then resolved by editing (adding) the last name where only first name and company existed prior to resolving: https://gist.github.com/ddbcbaf15ba1578a6db3cd574a3c3df4

charlesmchen:
Hopefully this is fixed by https://github.com/WhisperSystems/Signal-iOS/pull/1773 which will be available in our next release.

Issue #788, opened 2015-05-22
Tap & tap-and-hold on contact's name occasionally not working


2-4601:
I have tried the fingerprint verification screen on three of my friends' iPhones: 4S, 5 and 5S.
4S and 5S work as expected but for some reason on the **iPhone 5** the screen does not show up.
More specifically when in a conversation and I:
- tap the the contact's name the phone number does not show up
- tap and hold the contact's name the fingerprint screen does not show up

In both cases basically nothing happens. The name does not react to any touches.

In this specific device there are two Signal contacts. One of them is **only** a RedPhone user so I would assume the fingerprint screen would not work since there's no long-term key like there's in TextSecure. But the other contact (me) is both a RedPhone and a TextSecure user and the fingerprint screen should work - yet it does not.

More details:
- Signal was first installed on this device in fall 2014 i.e. 1.x version
- The user's own fingerprint is visible and copyable in the settings
- Texts and calls work normal
- Signal 2.0.3, iPhone 5, iOS 8.3
- This was not tested prior 2.0.3

I understand this is hard to fix since it's reproducible only on a single device/install.
Would the tap and long-press on the contact's name show in a debug log? I could post it next time I see my friend in person.

Thanks.

FredericJacobs:
Is the contact's name really short? might be that it makes it to hard to tap on?

2-4601:
Nope. It's 10 characters long (space not included)

FredericJacobs:
Looked through the code but I must say that I'm a little lost about what to do here.

I'm unable to reproduce :(

2-4601:
Thanks for looking into this.

Would this show in a debug log?

If not I'll just advice a reinstall.

FredericJacobs:
I honestly don't see how a reinstall would help. Nothing in the state of the application would influence this.

Have you tried with different names?

2-4601:
There's only one other contact and that contact is RedPhone only, so we have not been able to try with other names.

The name with the issue works in two other Signal installs without problems. It's an ascii only, 5 letters first name, 5 letters surname.

Could iOS contact restrictions affect this?

I will try with different names the next time I see my friend (need to introduce other Signal users for that).

2-4601:
I experienced this today with another device: iPhone 4S, iOS 8.3, Signal 2.1.1

All the contact names had this issue: no reaction to either tap nor tap and hold
I tried changing the contact names but it did not help.

[...]

PR #1498, opened 2016-11-29
Make settings more obvious

### LINKED ISSUE #1390: Settings menu should be more obvious for contacts

michaelkirk:
Added chevron to contact name to indicate it's tappable via a custom title view

Here's what it looks like in the most crowded context.

And with one bar button...

Note that I removed the "..." button from groups since it's redundant with tapping on the group name, and hopefully the new iconography makes it obvious enough.

Fixes #1390 (Settings should be more obvious)

I expect this to also solve:
Fixes https://github.com/WhisperSystems/Signal-iOS/issues/1402 (Can't get to conversation settings for some contacts)
Fixes https://github.com/WhisperSystems/Signal-iOS/issues/788 (https://github.com/WhisperSystems/Signal-iOS/issues/788)

Which were due to not finding the navbar label.

Closes #1479 with an alternate approach.

// FREEBIE


charlesmchen:
Superceded by https://github.com/WhisperSystems/Signal-iOS/pull/1773

Issue #1519, opened 2016-12-02
Avatar for contact with emoji needs design/change


riyapenn:
- [X ] I have searched open and closed issues for duplicates (but feel like I've overlooked something so this could also be mentioned in another issue) However, https://github.com/WhisperSystems/Signal-iOS/issues/1512  regarding unsaved contacts is related.

----------------------------------------

### Bug description
If a contact's name starts with an emoji character in the contact list entry, a character is displayed in the Avatar

### Steps to reproduce
- Have contact name in contact list with an avatar
- view message compose contact list, view invite contact list or open conversation thread and go to inbox view of contact list 

OR
- Have contact saved in contact list with JUST a phone number, NO NAME.
- view message compose contact list, view invite contact list or open conversation thread and go to inbox view of contact list 

**Actual result:** Character with an A is displayed
**Expected result:** Suggestion is a cleaner avatar or different character. For example, Signal Desktop displays diamond with a question mark (image below)

### Screenshots

iOS image of avatar for contact with name '[Emoji] Google Voice' 
![avatar emoji name](https://cloud.githubusercontent.com/assets/4898366/20820741/59c090ae-b7f2-11e6-9834-2d5d6e2af79d.png)

Signal Desktop handling of emoji as first character of contact name:
<img width="61" alt="desktopemojiavatar" src="https://cloud.githubusercontent.com/assets/4898366/20820795/a4fd62a4-b7f2-11e6-854d-6291de784302.png">

### Device info

**Device**: iPhone 5C
**iOS version**: 10.1.1
**Signal version:** 2.6.6

michaelkirk:
One approach would be to restrict initials to alphanumeric characters only.

Stanzi97:
To be honest, who (except 12-year old girls, who are not the target group of Signal) have contacts which start with a Smily/Emoji??

michaelkirk:
@Stanzi97 your comment is rude, unhelpful, and based on incorrect assumptions. Use respect when communicating here.

The current behavior is clearly broken and we should fix it.

PR #1523, opened 2016-12-04
Mkirk/fixup contacts

### LINKED ISSUE #1512: Avatar for unknown contact is confusing
When you receive a message from an unknown contact their avatar is just a "+" in a circle.

Worse than just unhelpful, this gives the false impression that this should be an actionable button.

The reasoning is the "+" is the "intials" for their display name (which is just their phone number if they're not in your contacts), but if you had e.g. "David Bowie" in your contacts it would appear as "DB".

michaelkirk:
Fixes crash when composing message and contact has no first name.

fixes #1512
closes #1520 
closes #1299

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -32,7 +32,7 @@ PODS:
   - JSQMessagesViewController (7.3.4):
     - JSQSystemSoundPlayer (~> 2.0.1)
   - JSQSystemSoundPlayer (2.0.1)
-  - libPhoneNumber-iOS (0.8.17)
+  - libPhoneNumber-iOS (0.9.1)
   - Mantle (2.1.0):
     - Mantle/extobjc (= 2.1.0)
   - Mantle/extobjc (2.1.0)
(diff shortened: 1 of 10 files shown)


roostr:
>     NSRange rangeOfLetters = [self.contactName rangeOfCharacterFromSet:[NSCharacterSet letterCharacterSet]];

I considered something like this when trying a solution that could also solve #1519, but ultimately couldn't find good documentation and shied away.

Here's what scared me off: I used [this sample code](http://stackoverflow.com/a/15742659) to look at the number of elements in a few different character sets, and found:

uppercaseLetterCharacterSet: 2229 elements
lowercaseLetterCharacterSet: 2599 elements
lettersCharacterSet: 173130 elements

It seems like a join on upper and lower may be an alternative solution, as letters includes a lot more extra stuff. From a few comments I found on SO, that extra stuff includes marks that modify letters (like when people do funky things with unicode to make text ladder up and down your screen). That's when I stopped going down the rabbit hole (as I couldn't find any good documentation on what to expect from the different character sets), and settled on a different strategy.

All that said, the code proposed looks good to ship, and if a problem surfaces in the field, this can always be revisited. So this is just an FYI of what I found when going down a similar path.

Issue #1210, opened 2016-06-11
accessibility text for buttons / magic tap to end phone call


lilmike:
- [x] I have searched open and closed issues for duplicates

---
### Bug description

Describe here the issue that you are experiencing.
I am blind, so use voice over on iOS to read the screen out loud. In signal, however, there are a few slight inaccessibilities plus a few improvements that coudl be made.
1. There are a few buttons that are labelled, instead of with a label like call or make a phone call, with the (I think) name of the image of the icon. For example, the phone button on the message page says 'phone-white' and the add media button says 'media-blue' or similar. Also, the group button on the contacts page shows 'group-white'.
2. In the phone call, it would be nice if voice over users coudl use the "magic tap" to end the phone call, instead of having to have the phone up to their ear to hear voice over, then finding the end button. The magic tap is a special voice over gesture which generally ends or begins something, such as music playback, a phone call, etc. It is a double tap with two fingers.
### Steps to reproduce
- using hyphens as bullet points
- list the steps
- that reproduce the bug
1. Turn on voice over, either in settings, or sometimes by tripple tapping the home button.
2. Navigate to signal, and find a chat you want to open.
3. double tap to open it, then touch the button for making a phone call, and adding media. Each will be spoken as 'phone-white' or 'media-blue'.
4. Go back, and double tap compose. You will notice, when touching the group chat button, that it is spoken as 'group-white'.
5. Go back to a message, and click the button to make a phone call. When you want to end the call, try double tapping with two fingers on the screen. It will not end the call. Instead, you will need to double tap the end button, which can be hard to find with the phone held up to your ear.

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)
**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)
### Device info

**Device**: iPhone 5s.
**iOS version**: 9.2

cannona:
I am also a VO user, and so am also effected by this issue.  If there is anything I can do to help get this resolved, please don't hesitate to ask.

TL;DR: me too!

paddyblifu:
I' also a vo-user too and I've found the following issue:

[...]

Issue #864, opened 2015-08-04
Labeling buttons to improve accessibility for the blind


MahfudNay:
Hello

Congratulation for Signal app, it is a fantastic app, now it is the only app that blind users can use for private messages.
To improve the accessibility, you have several tools from Apple here https://developer.apple.com/accessibility/
I would like to suggest some changes to improve accessibility for blind users for Signal app.
I can press buttons but some of them are not correctly labeled, I know what happen because I've saw the result of the action but VoiceOver reads something different that I do not understand.
1. Delete a conversation: when I put down a finger to see what actions i can get, VoiceOver does not read anything when I'm on the delete  option and the app do not ask me before if I'm sure to want to delete the conversation, so, I've deleted when I didn't want to do that.
2. Attach something: Voiceover reads "BtnAttachment Blue". I recommend you to label it as "Attach something, Attach or share".
1. BtnCancel white: I recommend to label it as "Cancel or back". That happen for example when I check the settings of the app.
   1. BtnPhone white: in an individual conversation, VoiceOver reads that button to call someone. I recommend to label it as "Phone, call or phone (name of the person) or call to (name of the person).
   2. Separated  information for a message: when I receive a message, I have to flick 3 times to read all information about the message: A. Time. B. Content of the message. C. Name of the person. 
      I recommend you to try to get that VoiceOver reads all that information just in once.
   3. The name of the person doesn't appear for each message in a conversation. When I read the content of the conversation, I have to guess if the message that I am reading is mine or from the my friend. If you get that VoiceOver say the name of the person before read the content of the message, then, it would be perfect :).

FredericJacobs:
Oh, wow. Thanks for writing this up.

I'll be in touch with you soon. I would love you to be part of our beta-program so that we can have shorter feedback loop (so that we don't have to wait for Apple to approve the test updates).

MahfudNay:
Hello

Thank you very much.
I will be very happy to  participate in the beta program, I wait  for your instructions.
I’m sure that  working together we are going to get a good app :).

Mahfud




Mahfud Chaaban
mahfud.chaaban@gmail.com
www.tecnocharlas.es <http://www.tecnocharlas.es/>

riyapenn:
Additional voiceover details:
[...]

PR #1529, opened 2016-12-07
Improve accessibility a bit

jschneier:
### Contributor checklist

 * iPhone SE, 10.1.1

- - - - - - - - - -
This is a start at #864 and #1210. Any feedback on how well this conveys meaning to the user/fits with Apple's HIG would be appreciated. Major missing pieces that still need to happen:

- [] We have to localize the Storyboards and I'm not sure how to do that - a number of the confusing buttons are built in IB and while I can set runtime defined `accessibilityLabel`s and such it doesn't localize into files without additional Xcode black magic
- [] I want to do magic tap at least on phone calls, just need to work out the state logic properly
- [] We can make things like reading out the message a lot nicer where it says things like "sent on" instead of just spitting out the date (WhatsApp does it well)
- [] I couldn't figure out a way to tap on the name to get to the conversation settings with VO enabled, maybe #1498 will help with that, you can do it in WhatsApp just fine

#1498 conflicts with this so hopefully that can be merged first and then this can be rebased/updated. This is my first time using VO, quite an experience.

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -236,6 +236,8 @@ typedef enum : NSUInteger {
     self.messageAdapterCache = [[NSCache alloc] init];
 
     _attachButton = [[UIButton alloc] init];
+    _attachButton.accessibilityLabel = NSLocalizedString(@"ATTACHMENT_LABEL", @"Accessibility label for attaching photos");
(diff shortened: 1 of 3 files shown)


michaelkirk:
Awesome. Thank you!

Would you mind opening issues for the other things you outlined?

PR #1532, opened 2016-12-08
Prevent iRate prompt when handling local notifications

### LINKED ISSUE #1530: keyboard overlays "Rate Signal" when responding to notification
## Actual Behavior

From a user report:

> I opened Signal to respond to a message from a friend, and the notification asking me to rate Signal appeared on top of the message from my friend, with the virtual keyboard on top of the notification. Which made it impossible to either dismiss the notification or use Signal.

## Expected Behavior

You should not see the rating pop-up when the app was launched in response to a notification.

guillett:
### Contributor checklist

 * iPhone5,2, iOS Version: 9.3.5

- - - - - - - - - -

### Description
Local notification prevents iRate prompt. I added a flag to prevent the next prompt to be presented.
Should fix #1530

Testing was a bit tricky as notifications are hard to generate. When the app was going to the background I created a new message and a local notification to ensure that:
- no prompt was presented when clicking on the notification
- a prompt was presented when clicking directly on the icon

I originally implemented another strategy that prevented useCount increment but the limiting condition may not be the useCount so I changed to the pushed alternatives.

I thought about an alternative adding a delay as a condition, for instance: Can't prompt in less than 30 seconds. That would delay the prompt to the next ```ApplicationEnterForeground```occurrence.

Let me know if you think alternatives (one of those or another one) would be preferred.

### DIFF
--- a/Libraries/iRate/iRate.h
+++ b/Libraries/iRate/iRate.h
@@ -165,6 +165,7 @@ typedef NS_ENUM(NSUInteger, iRateErrorCode) {
 - (void)promptIfAllCriteriaMet;
 - (void)openRatingsPageInAppStore;
 - (void)logEvent:(BOOL)deferPrompt;
+- (void)preventPromptAtNextTest;
 
 @end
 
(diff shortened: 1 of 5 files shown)


michaelkirk:
Rather than tracking state in the iRate library, I think we should use the launch state in the app delegate to determine whether or not to load iRate at all.

e.g. http://stackoverflow.com/questions/16393673/detect-if-the-app-was-launched-opened-from-a-push-notification

guillett:
The ```(BOOL)application: didFinishLaunchingWithOptions:``` is only triggered at launch time not when the app is brought back from the background to the foreground so that would not prevent prompt in that situation.

michaelkirk:
@guillett right. 

Thanks!

Issue #1254, opened 2016-07-13
Attachment never completes downloading when app is killed mid-download.


michaelkirk:
### Steps to reproduce
- Alice sends Bob a big attachment
- Bob sees "Attachment is downloading" info message
- Bob quits his app via the app switcher before it completes downloading (or it crashes, or his phone dies, etc.)
- Bob relaunches Signal, goes to Alices thread
- Bob still sees "Attachment is downloading"

**Actual result:** 
- Download never completes
- You can't tap to restart the download

**Expected result:** 
- Bob should see "Download stalled. Tap to restart" (we don't want to restart automatically lest we hammer someones bandwidth on a shoddy connection)
- You can tap to restart the download and see "Attachment is downloading"
## proposal

On startup have an async task mark all pointers that are downloading as failed. The trick will be ensuring we're not marking freshly enqueued pointers as failed.

anatoli26:
Same problem here. Why "high priority" was removed? I believe it's a critical problem. One records a voice message, sends it, thinks the peer received it, but on the other end the only thing that appears is "attachment is downloading" for eternity, with the only option to delete it.

Expected behavior is to timeout quickly and either re-download the file automatically or to mark it as failed and offer an option to start re-download.

michaelkirk:
> Why "high priority" was removed?

Instead of just labeling a bunch of things "high priority", we're ordering things in the "prioritized" milestone. It doesn't mean it's less important.

> What help is needed to resolve this bug?

Pull requests are appreciated.

michaelkirk:
This was fixed and will be released in 2.10.0

PR #1534, opened 2016-12-08
Fix show error message when download fails

michaelkirk:
Previously we had our "queued" and "failed" logic backwards.

// FREEBIE


jschneier:
Yeah this is much better. And makes #1254 more obvious.

PR #1580, opened 2017-01-04
Update SignalServiceKit pod.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -32,7 +32,7 @@ PODS:
   - JSQMessagesViewController (7.3.4):
     - JSQSystemSoundPlayer (~> 2.0.1)
   - JSQSystemSoundPlayer (2.0.1)
-  - libPhoneNumber-iOS (0.9.2)
+  - libPhoneNumber-iOS (0.9.1)
   - Mantle (2.1.0):
     - Mantle/extobjc (= 2.1.0)
   - Mantle/extobjc (2.1.0)
@@ -154,7 +154,7 @@ SPEC CHECKSUMS:
   HKDFKit: c058305d6f64b84f28c50bd7aa89574625bcb62a
   JSQMessagesViewController: 39fed975e3c9f8eba7292071e29eeb541d105e66
   JSQSystemSoundPlayer: c5850e77a4363ffd374cd851154b9af93264ed8d
-  libPhoneNumber-iOS: a8bffdec18c37728360f6771fe021302f1e0b497
+  libPhoneNumber-iOS: 81ad1e6bfcf46e668636333269afbfe60399a55b
   Mantle: 2fa750afa478cd625a94230fbf1c13462f29395b
   OpenSSL: 246ffb948e9d56466727fd318134af35f5aa764e
   PastelogKit: 7b475be4cf577713506a943dd940bcc0499c8bca
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -32,7 +32,7 @@ PODS:
   - JSQMessagesViewController (7.3.4):
     - JSQSystemSoundPlayer (~> 2.0.1)
   - JSQSystemSoundPlayer (2.0.1)
-  - libPhoneNumber-iOS (0.9.1)
+  - libPhoneNumber-iOS (0.9.2)
   - Mantle (2.1.0):
     - Mantle/extobjc (= 2.1.0)
   - Mantle/extobjc (2.1.0)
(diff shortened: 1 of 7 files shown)


Signal/src/AppDelegate.m
```diff
@@ -239,7 +239,11 @@ - (void)applicationDidBecomeActive:(UIApplication *)application {
                                                [TSSocketManager becomeActiveFromForeground];
                                                [[Environment getCurrent].contactsManager verifyABPermission];
                                            }];
-
+    
+    if ([TSAccountManager isRegistered]) {
+        [[PushManager sharedManager] applicationDidBecomeActive];
```
charlesmchen: I'm not convinced that `PushManager` is the ideal place to hang this method, but it already owns the messageFetcherJob, so it's expedient to do so... for now.

michaelkirk: If there is a blocking migration, e.g. if we add a new database view, it can take a long time for `[TSAccountManager isRegistered]` to return, so in AppDelegate I've gotten into the practice of putting things that read from the DB into an async block like #235.

michaelkirk: It might make sense to move the messageFetcherJob to the Environment, and have this code and the push manager access the fetcher job from the Environment.

I'm also OK with it here given the cost/benefit of hammering things onto the environment.

Podfile.lock
```diff
@@ -154,7 +154,7 @@ SPEC CHECKSUMS:
   HKDFKit: c058305d6f64b84f28c50bd7aa89574625bcb62a
   JSQMessagesViewController: 39fed975e3c9f8eba7292071e29eeb541d105e66
   JSQSystemSoundPlayer: c5850e77a4363ffd374cd851154b9af93264ed8d
-  libPhoneNumber-iOS: 81ad1e6bfcf46e668636333269afbfe60399a55b
+  libPhoneNumber-iOS: a8bffdec18c37728360f6771fe021302f1e0b497
```
michaelkirk: Pod update updates all the dependencies below SSK too. 

Could you update libphonenumber-ios separately for future git bisectabilitiy?

I would do this by manually editing the podfile.lock (or `git co -p master -- Podfile.lock`!), then re-running `pod install` to get a sensible PODFILE CHECKSUM.

Signal/src/network/PushManager.m
```diff
@@ -162,6 +162,10 @@ - (void)application:(UIApplication *)application didReceiveRemoteNotification:(N
     }
 }
 
+- (void)applicationDidBecomeActive {
```
michaelkirk: There is no consistent style across the codebase, but I'm trying to converge on one by using `git clang-format`. You might consider adding it to your git hooks.

Signal/src/AppDelegate.m
charlesmchen: > I've gotten into the practice of putting things that read from the DB into an async block...

I'm concerned about thready safety, since this moves all of the logic in that completion block off the main thread.  Is all of the logic in that completion thread-safe?  Is any of it expensive? Or is it calling `[TSAccountManager ifRegistered]` that is expensive?

charlesmchen: > Environment

It'd be nice to try to set out the responsibilities of each of these entities.

Signal/src/network/PushManager.m
charlesmchen: I love it. clang-format takes a (massive) config file - are you just using defaults?

charlesmchen: I also intend to start using a precommit script to do things like normalize our file headers, etc.

Signal/src/AppDelegate.m
michaelkirk: `[TSAccountManager isRegistered]` **can** be expensive.

It reads from the DB, so it has to wait for the DB to be ready before it can run. DB setup is normally not expensive, unless the user is launching for the first time after a new release which adds a database extension like a new view. Depending on the size of the users database, populating a new extension can be very expensive.

michaelkirk: I see no problem with dispatching message fetching off the main thread.

If your concern was with thread safety at the storage level specifically, I don't see a problem there either. But it would be good to dive into the concurrency idioms. They're different: https://github.com/yapstudios/YapDatabase/wiki/Thread-Safety

Signal/src/network/PushManager.m
michaelkirk: https://github.com/WhisperSystems/Signal-iOS/blob/master/.clang-format

Signal/Signal-Prefix.pch
```diff
@@ -16,6 +16,7 @@
     #endif
 
     #import "iOSVersions.h"
+    #import "Asserts.h"
```
michaelkirk: #import <SignalServiceKit/Asserts.h> since it's not in the local sources.

Signal/src/AppDelegate.m
charlesmchen: Got it; done.

Signal/Signal-Prefix.pch
charlesmchen: Done.

Podfile.lock
charlesmchen: Done.

charlesmchen:
PTAL @michaelkirk

michaelkirk:
LGTM

PR #1584, opened 2017-01-06
Rework startup logging.

charlesmchen:
* Instantiate AppVersion asap.
* Ensure logging is configured before other startup logic.

Previously, we were configuring logging after Environment. But some of the logic used by Environment's initialization process does logging (and should). So it seems like we should configure logging before Environment. 

Is this safe? The proper ordering of these startup operations isn't clear. Is there anything in the logging configuration that depends on Environment? Do I need to investigate further?

PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -23,6 +23,7 @@
 #import <SignalServiceKit/OWSIncomingMessageReadObserver.h>
 #import <SignalServiceKit/OWSMessageSender.h>
 #import <SignalServiceKit/TSAccountManager.h>
+#import <SignalServiceKit/AppVersion.h>
 
 NSString *const AppDelegateStoryboardMain = @"Main";
 NSString *const AppDelegateStoryboardRegistration = @"Registration";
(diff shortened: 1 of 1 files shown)


Signal/src/AppDelegate.m
```diff
@@ -52,6 +53,23 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     CategorizingLogger *logger = [CategorizingLogger categorizingLogger];
     [logger addLoggingCallback:^(NSString *category, id details, NSUInteger index){
     }];
+    
+    BOOL loggingIsEnabled;
+    
+#ifdef DEBUG
+    // Specified at Product -> Scheme -> Edit Scheme -> Test -> Arguments -> Environment to avoid things like
```
michaelkirk: What can you specify in the scheme that affects the code here? I feel like this comment is referring to code that no longer exists. Or does it make sense to you?

Signal/src/AppDelegate.m
```diff
@@ -52,6 +53,23 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+    // the phone directory being looked up during tests.
+    loggingIsEnabled = TRUE;
+    [DebugLogger.sharedLogger enableTTYLogging];
+#elif RELEASE
+    loggingIsEnabled = Environment.preferences.loggingIsEnabled;
```
michaelkirk: This looks a little funny since we're now accessing something on the environment before we've "set" the environment. It might work since it's ultimately the storage singleton, but at a minimum, can you confirm you've tested the behavior in a release build with file logging enabled/disabled?

Toggle file logging in (settings > privacy > enable debug log), and read the contents of the debug log by 
1. emailing it to yourself 
2. pretending like you're going to submit a GH issue which just copies it to your pasteboard buffer.

charlesmchen:
Closing in favor of #1636

PR #1587, opened 2017-01-09
Sync CallKit mute button state with in app mute button.

michaelkirk:
**note:** this isn't targeting master.

Fixes a problem where the system call screen (CallKit) mute button could get out of sync with the in-app call screen's mute button.

This includes a refactor which, I think, better separates concerns and consequently reduces dependencies in the CallViewController.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/NonCallKitCallUIAdaptee.swift
+++ b/Signal/src/call/NonCallKitCallUIAdaptee.swift
@@ -48,7 +48,9 @@ class NonCallKitCallUIAdaptee: CallUIAdaptee {
     }
 
     internal func answerCall(_ call: SignalCall) {
-        // NO-OP
+        CallService.signalingQueue.async {
+            self.callService.handleAnswerCall(call)
+        }
     }
 
     internal func declineCall(_ call: SignalCall) {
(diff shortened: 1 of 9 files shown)


Signal/src/call/SignalCall.swift
```diff
@@ -16,28 +16,42 @@ enum CallState: String {
...
+}
+
 /**
  * Data model for a WebRTC backed voice/video call.
  */
 @objc class SignalCall: NSObject {
 
     let TAG = "[SignalCall]"
 
+    var delegate: CallDelegate?
```
charlesmchen: Shouldn't this be weak?

Signal/src/call/NonCallKitCallUIAdaptee.swift
```diff
@@ -43,23 +43,32 @@ class NonCallKitCallUIAdaptee: CallUIAdaptee {
         }
     }
 
-    public func reportMissedCall(_ call: SignalCall, callerName: String) {
+    internal func reportMissedCall(_ call: SignalCall, callerName: String) {
```
charlesmchen: Huh, I'm surprised these should be internal, since they're implementing public methods in the protocol.

michaelkirk: I've made them public.

Signal/src/call/SignalCall.swift
michaelkirk: thank you!

PR #1593, opened 2017-01-10
Fix font lookup on iOS before 8.2.

charlesmchen:
I also simplified the iOS version checking.

PTAL @michaelkirk


Signal/src/environment/iOSVersions.h
```diff
@@ -9,94 +9,6 @@
 #import <Foundation/Foundation.h>
 #include <Availability.h>
 
-// Source: https://github.com/carlj/CJAMacros/blob/master/CJAMacros/CJAMacros.h
+#define SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(major, minor) \
```
charlesmchen: AFAIK you only ever need a single method. I wanted to move away from the existing macros since they use constant-ized iOS versions and the constants are incomplete and non-trivial to update. Far simpler (and just as readable) to pass the major minor versions in as ints.

Signal/src/util/UIFont+OWS.m
```diff
@@ -12,19 +12,35 @@
 @implementation UIFont (OWS)
 
 + (UIFont *)ows_thinFontWithSize:(CGFloat)size {
-    return [UIFont systemFontOfSize:size weight:UIFontWeightThin];
+    if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(8, 2)) {
+        return [UIFont systemFontOfSize:size weight:UIFontWeightThin];
```
charlesmchen: Apple changes the system font every so often, so I think we're right to use the system font methods when possible, but we need to fall back to hard-coded font names before 8.2.

charlesmchen:
PTAL @michaelkirk

michaelkirk:
nice cleanup. LGTM

PR #1595, opened 2017-01-10
Charlesmchen/webrtc/call view

charlesmchen:
### This PR now ready for review.

The redesign isn't nearly done, but this PR accomplishes a significant restructuring of this view.

* Add PureLayout
* Update Obj-C/Swift bridging header to enable programmatic layout in Swift.
* Add UIView+OWS category to make common layout idioms easier and more readable.  I expect this category to grow significantly over time as we continue to do programmatic layout.
* Update storyboards.
   * Strip contents of new Call View from storyboards.
   * Change labels of call views in storyboards.
* Rework CallViewController to create its views and their layout programmatically.   

PTAL @michaelkirk


charlesmchen:
I'm continuing to explore our options, design-wise.  Here are some current screenshots:

screen shot 2017-01-11 at 12 44 23 pm
screen shot 2017-01-11 at 12 44 02 pm
screen shot 2017-01-11 at 12 43 19 pm
screen shot 2017-01-11 at 12 42 43 pm
screen shot 2017-01-11 at 12 42 06 pm
screen shot 2017-01-11 at 12 41 31 pm

There's lots more to be done:

* We should regenerate the button assets.
  * We should add padding to the button assets to increase the hot area for these buttons.
  * The asset styles don't match.
  * The "message" and "video" assets aren't acceptable.
  * We need down- and possibly disabled states for these buttons.
  * The icons we need aren't obscure - but I haven't yet identified a good place to buy iOS icons that are appropriate.
* We'll probably need to tweak the margins using physical devices.
* The positioning of the hangup button isn't consistent between "incoming" and "ongoing modes". I haven't decided whether or not to align them.
* The layout will look terrible on 4S, but should be functional. At the very least the avatar will be tiny.
* etc.

Signal/src/call/SignalCall.swift
```diff
@@ -58,6 +58,9 @@ enum CallState: String {
         return lhs.localId == rhs.localId
     }
 
+    static func newCallSignalingId() -> UInt64 {
```
charlesmchen: I've exposed this as a function so that tests and debugging logic can generate suitable "call signaling ids."

Signal/src/view controllers/CallViewController.swift
```diff
@@ -205,6 +226,195 @@ class CallViewController: UIViewController {
         callStateDidChange(call.state)
     }
 
+    func createViews() {
+        // Dark blurred background.
+        let blurEffect = UIBlurEffect(style: .dark)
+        blurView = UIVisualEffectView(effect: blurEffect)
+        self.view.addSubview(blurView)
```
charlesmchen: I checked out using a `UIVibrancyEffect`, but it made the text less readable.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -287,21 +519,25 @@ class CallViewController: UIViewController {
...
-    @IBAction func didPressSpeakerphone(sender speakerphoneButton: UIButton) {
+    func didPressSpeakerphone(sender speakerphoneButton: UIButton) {
         Logger.info("\(TAG) called \(#function)")
         speakerphoneButton.isSelected = !speakerphoneButton.isSelected
         audioService.isSpeakerphoneEnabled = speakerphoneButton.isSelected
     }
 
-    @IBAction func didPressAnswerCall(sender: UIButton) {
+    func didPressTextMessage(sender speakerphoneButton: UIButton) {
+        Logger.info("\(TAG) called \(#function)")
```
charlesmchen: My inclination is to eventually make a UIButton category/extension so that we can gather concerns and define button behavior where the button is created.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -205,6 +226,191 @@ class CallViewController: UIViewController {
...
 
+    func createViews() {
+        // Dark blurred background.
+        let blurEffect = UIBlurEffect(style: .dark)
+        blurView = UIVisualEffectView(effect: blurEffect)
+        self.view.addSubview(blurView)
+
+        // Contact views
+        contactNameLabel = UILabel()
+        contactNameLabel.font = UIFont.ows_lightFont(withSize:ScaleFromIPhone5To7Plus(32, 40))
```
charlesmchen: Responsive font sizes.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -205,6 +226,191 @@ class CallViewController: UIViewController {
...
+        self.view.addSubview(ongoingCallView)
+        let ongoingCallRows = [
+            rowWithSubviews(subviews:[textMessageButton, videoButton],
+                            fixedHeight:buttonSize),
+            rowWithSubviews(subviews:[muteButton, speakerPhoneButton,],
+                            fixedHeight:buttonSize),
+            rowWithSubviews(subviews:[hangUpButton],
+                            fixedHeight:buttonSize),
+            ]
+        let ongoingCallRowSpacing = ScaleFromIPhone5To7Plus(20, 25)
```
charlesmchen: Responsive font sizes.

charlesmchen: Responsive, constant-ized margins and spacing.

Signal/src/UIView+OWS.h
```diff
@@ -0,0 +1,51 @@
...
+// value (for iPhone 5) to the current device based on screen width,
+// linearly interpolating through the origin.
+CGFloat ScaleFromIPhone5(CGFloat iPhone5Value);
+
+// A set of helper methods for doing layout with PureLayout.
+@interface UIView (OWS)
+
+// Pins the width of this view to the width of its superview, with uniform margins.
+- (void)autoPinWidthToSuperviewWithMargin:(CGFloat)margin;
+- (void)autoPinWidthToSuperview;
```
charlesmchen: Ideally, these convenience methods don't just DRY up repetitive layout code, their names are higher-level and more semantic than raw AL operations.

charlesmchen:
PTAL @michaelkirk

Signal/src/view controllers/CallViewController.swift
```diff
@@ -205,6 +226,191 @@ class CallViewController: UIViewController {
...
+                    spacer.autoVCenterInSuperview()
+                    if lastSpacer != nil {
+                        spacer.autoMatch(_:.width, to:.width, of:lastSpacer!)
+                    }
+                    lastSpacer = spacer
+                }
+
+                lastSubview = subview
+            }
+            subviews.first!.autoPinEdge(toSuperviewEdge:.left)
```
michaelkirk: I guess this could go at line 333 if you inverted the condition, and added an else. 
```
if lastSubview == nil {
    subviews.first!.autoPinEdge(toSuperviewEdge:.left)
    subviews.first!.autoPinEdge(toSuperviewEdge:.right)
} else {
  let spacer = UIView()
  ...
  lastSubview = subview
}
```

It's maybe a little clearer that we're handling every subview's pinning that way.


Also, consider renaming lastSubview -> prevSubview for clarity

Signal/src/view controllers/CallViewController.swift
```diff
@@ -205,6 +226,191 @@ class CallViewController: UIViewController {
...
+            subviews.last!.autoPinEdge(toSuperviewEdge:.right)
+
+            Logger.error("row \(subviews.count) -> \(row.subviews.count)")
+
+        } else if subviews.count == 1 {
```
michaelkirk: I forgot what condition the else was from. How about moving this up to an early return?

Signal/src/UIView+OWS.m
```diff
@@ -0,0 +1,170 @@
...
+
+CGFloat ScaleFromIPhone5(CGFloat iPhone5Value)
+{
+    CGFloat screenShortDimension = ScreenShortDimension();
+    return round(iPhone5Value * screenShortDimension / kIPhone5ScreenWidth);
+}
+
+#pragma mark -
+
+@implementation UIView (OWS)
```
michaelkirk: typically with categories I add our prefix to the methods, as in `ows_autoPinWidthToSuperview`, but these methods will be used so often, and are unlikely to clobber anything, so it seems worth breaking that convention.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -246,8 +452,18 @@ class CallViewController: UIViewController {
         self.callStatusLabel.text = textForState
 
         // Show Incoming vs. (Outgoing || Accepted) call controls
-        callControls.isHidden = callState == .localRinging
-        incomingCallControls.isHidden = callState != .localRinging
+        let isRinging = callState == .localRinging
+        for subview in allControls() {
```
michaelkirk: Maybe a little more concise
```
incomingCallControls.forEach { $0.hidden = !isRinging }
ongoingCallControls.forEach { $0.hidden = isRinging }
```

Signal/src/view controllers/CallViewController.swift
```diff
@@ -264,6 +480,19 @@ class CallViewController: UIViewController {
         }
     }
 
+    func allControls() -> [UIView] {
+        return incomingCallControls() + ongoingCallControls()
+    }
+
+    func incomingCallControls() -> [UIView] {
```
michaelkirk: I commonly see inlined computed vars for idempotent getters like this.

```
var incomingCallControls: [UIView] {
  get { 
       return [ acceptIncomingButton, declineIncomingButton, ]
  }
}
```

Signal/src/view controllers/CallViewController.swift
```diff
@@ -287,21 +516,25 @@ class CallViewController: UIViewController {
         self.dismiss(animated: true)
     }
 
-    @IBAction func didPressMute(sender muteButton: UIButton) {
+    func didPressMute(sender muteButton: UIButton) {
         Logger.info("\(TAG) called \(#function)")
         muteButton.isSelected = !muteButton.isSelected
         CallService.signalingQueue.async {
```
michaelkirk: FYI I touched these methods in https://github.com/WhisperSystems/Signal-iOS/pull/1587/files#diff-7b6a32382aa648492d9e501eb1e6269cR286

If there *is* a conflict it *should* be pretty straight forward to resolve. LMK if you have an issue.

charlesmchen:
> Also, I saw some unconventional swift formatting. I'm still getting used to swift formatting too. Can you run your code through swift lint? I just included it in the project before you started, but have it disabled in the "Swift lint" run phase because it was too noisy and I didn't want to reformat the entire code base as part of the webrtc branch.

Done.

Signal/src/UIView+OWS.m
charlesmchen: Right. I'm also trying to follow the PureLayout convention of prefixing method names with auto + layout verb.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -147,19 +148,36 @@ class CallViewController: UIViewController {
...
+
+    // MARK: Ongoing Call Controls
+
+    var ongoingCallView: UIView!
+
+    var hangUpButton: UIButton!
+    var muteButton: UIButton!
+    var speakerPhoneButton: UIButton!
+    // TODO: Which call states does this apply to?
+    var textMessageButton: UIButton!
```
charlesmchen: Okay, for now I've added a "text message" button to the incoming mode. Let's revisit the video button when we get to video.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -205,6 +226,191 @@ class CallViewController: UIViewController {
...
+                            fixedHeight:buttonSize),
+            rowWithSubviews(subviews:[hangUpButton],
+                            fixedHeight:buttonSize),
+            ]
+        let ongoingCallRowSpacing = ScaleFromIPhone5To7Plus(20, 25)
+        var lastRow : UIView?
+        for row in ongoingCallRows {
+            ongoingCallView.addSubview(row)
+            row.autoPinWidthToSuperview()
+            if lastRow != nil {
```
charlesmchen: Good idea, done.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -205,6 +226,191 @@ class CallViewController: UIViewController {
...
+        declineIncomingButton = createButton(imageName:"endcall",
+                                             action:#selector(didPressDeclineCall),
+                                             buttonSize:buttonSize)
+
+        incomingCallControlsRow = rowWithSubviews(subviews:[acceptIncomingButton, declineIncomingButton],
+                                                  fixedHeight:buttonSize)
+        self.view.addSubview(incomingCallControlsRow)
+    }
+
+    func createButton(imageName : String!, action : Selector!, buttonSize : CGFloat!) -> UIButton {
```
charlesmchen: Huh, okay.  I'm very much still unaware of best swift practices. Thanks and done.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -205,6 +226,191 @@ class CallViewController: UIViewController {
...
+        let image = UIImage(named:imageName)
+        Logger.error("button \(imageName) \(NSStringFromCGSize(image!.size))")
```
charlesmchen: Removed that - it was obsolete.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -205,6 +226,191 @@ class CallViewController: UIViewController {
...
+        button.setImage(image, for:.normal)
+        button.addTarget(self, action:action, for:.touchUpInside)
+        button.autoSetDimension(.width, toSize:buttonSize)
+        button.autoSetDimension(.height, toSize:buttonSize)
+        return button
+    }
+
+    // Creates a row view that evenly spaces its subviews horizontally.
+    // If there is only a single subview, it is centered.
+    func rowWithSubviews(subviews : Array<UIButton>, fixedHeight : CGFloat) -> UIView {
```
charlesmchen: Done.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -205,6 +226,191 @@ class CallViewController: UIViewController {
...
+                if lastSubview != nil {
+                    let spacer = UIView()
+                    spacer.isHidden = true
+                    row.addSubview(spacer)
+                    spacer.autoPinEdge(.left, to:.right, of:lastSubview!)
+                    spacer.autoPinEdge(.right, to:.left, of:subview)
+                    spacer.setContentHuggingHorizontalLow()
+                    spacer.autoVCenterInSuperview()
+                    if lastSpacer != nil {
+                        spacer.autoMatch(_:.width, to:.width, of:lastSpacer!)
```
charlesmchen: Oh nice! Done.

Signal/src/view controllers/CallViewController.swift
charlesmchen: The problem is that we don't do `first` on both lines - the other one uses `last`.  ie. the left-most button in this row should be pinned to the left side of the row and the right-most to the right.

Signal/src/view controllers/CallViewController.swift
charlesmchen: I've clarified by reorganizing the comments.

Signal/src/view controllers/CallViewController.swift
charlesmchen: Yeah, I was tempted to do that, but I didn't want to make the assumption that no controls showed up in both `incomingCallControls` and `ongoingCallControls`.  I was trying to future-proof this against later refactoring since I expect we'll continue to rework this view.

Signal/src/view controllers/CallViewController.swift
michaelkirk: Generally, prefer plain non-optional types whenever possible. As far as I can tell, there's no reason we'd ever want to pass any of these params as nil, so we'd want the compiler to error at the call site.

Unwrapping parameters in a function signature is not very common. The only valid use case I can think of is when implementing an objc API that didn't have nullability annotations, yet somehow you *know* the params will not be nil.

See http://stackoverflow.com/questions/24223120/in-swift-what-does-the-symbol-mean-in-a-function-signature

Signal/src/view controllers/CallViewController.swift
charlesmchen: That's great, thank you!

Signal/src/view controllers/CallViewController.swift
charlesmchen: Thanks for explaining.

Signal/src/view controllers/CallViewController.swift
charlesmchen: It turns out the preferred syntax is to skip `get{}` unless there is also a `set()`.

```
    var incomingCallControls: [UIView] {
        return [ incomingTextMessageButton, acceptIncomingButton, declineIncomingButton, ]
    }
```

PR #1600, opened 2017-01-12
test peerconnectionclient

michaelkirk:
This is primarily a refactor to move WebRTC integration code from the CallService to the PeerConnectionClient.

- PeerConnectionClient is already the primary point of WebRTC integration
- CallService is too big.

Also snuck  in
https://github.com/WhisperSystems/Signal-iOS/commit/24baa027c2cd30c6c42465715eaa6e43e6083721 which is a useful feature both for offline testing and in the hypothetical case where SignalService doesn't return ICEServers.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -75,7 +75,7 @@ enum CallError: Error {
 // FIXME TODO do we need to timeout?
 fileprivate let timeoutSeconds = 60
 
-@objc class CallService: NSObject, PeerConnectionClientDelegate, RTCDataChannelDelegate {
+@objc class CallService: NSObject, PeerConnectionClientDelegate {
 
     // MARK: - Properties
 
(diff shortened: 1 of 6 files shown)


Signal/src/call/CallService.swift
```diff
@@ -167,11 +167,13 @@ fileprivate let timeoutSeconds = 60
...
-            self.peerConnectionClient = peerConnectionClient
+            let peerConnectionClient = PeerConnectionClient(iceServers: iceServers, delegate: self)
+
+            // When placing an outgoing call, it's our responsibility to create the DataChannel. Recipient will not have
+            // to do this explicitly.
+            peerConnectionClient.createSignalingDataChannel()
 
-            // When calling, it's our responsibility to create the DataChannel. Receivers will not have to do this explicitly.
-            self.peerConnectionClient!.createSignalingDataChannel(delegate: self)
+            self.peerConnectionClient = peerConnectionClient
```
charlesmchen: So, there's only one peerConnectionClient at a time. If a client has an ongoing call and receives another WebRTC call, that new call's peerConnectionClient isn't set up until the user hangs up the first by confirming that they want to switch?

Signal/src/call/CallService.swift
```diff
@@ -831,205 +878,22 @@ fileprivate let timeoutSeconds = 60
...
-//        this.callId            = null;
-//        this.audioEnabled      = false;
-//        this.videoEnabled      = false;
-//        this.pendingIceUpdates = null;
-//        lockManager.updatePhoneState(LockManager.PhoneState.IDLE);
+        Logger.debug("\(TAG) in \(#function)")
 
         peerConnectionClient?.terminate()
+
         peerConnectionClient = nil
```
charlesmchen: Can we do something like this too:

```
peerConnectionClient .delegate = nil
```

This kind of defensive gesture would prevent CallService receiving any events from old connections.

Signal/src/call/CallService.swift
michaelkirk: > So, there's only one peerConnectionClient at a time.
right

> If a client has an ongoing call and receives another WebRTC call, that new call's peerConnectionClient isn't set up until the user hangs up the first by confirming that they want to switch?

Well... conceivably, but in the current iteration when you are already in a call when receiving another call, you send a "BUSY" message. =/

Issue #1161, opened 2016-04-25
When on phone call, "Mute" and "Speaker Phone" icons do not indicate status


vdcbb:
- [x] I have searched open and closed issues for duplicates

---
### Bug description

Describe here the issue that you are experiencing.
### Steps to reproduce
- using hyphens as bullet points
- list the steps
- that reproduce the bug

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)
**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)
### Screenshots

This is with speaker phone ON and mute OFF. The difference is pretty subtle, especially on older screens. 

![image](https://cloud.githubusercontent.com/assets/217057/15199940/590fa93e-1797-11e6-8ed5-09b666a5b7bf.png)

Compare to something like what native phone app does, which is much clearer:
![image](https://cloud.githubusercontent.com/assets/217057/15199983/c624f196-1797-11e6-8734-c122325f6634.png)
### Device info

**Device**: iDevice X
**iOS version**: X.Y.Z
**Signal version:** Z.Y

vdcbb:
[Bug description]
When having a conversation over a Signal phone call, the "Mute" icon and "Speaker" icon do not indicate what state they are in (i.e., tapping them does not visually indicate whether they are enabled or disabled).

[Steps to reproduce]
1) make a phone call with Signal. 
2) During the conversation, tap the "Mute" icon (microphone with line through it).
3) Notice how the icon looks exactly the same as it did before? The icon's "status" has not changed to indicate what "state" it is in (i.e., is "Mute" enabled or disabled?).

[Expected result]
The icons should indicate visually what state they are in (i.e., are they enabled or disabled?). Tapping the icon when it is enabled should change its appearance so that it communicates it is now in a disabled state and vice versa.

[Device]
iPhone 6s
iOS version: 9.3.1
Signal version: 2.2

big-r81:
Hi  vdcbb,

I can't reproduce this. If i press these buttons, they are getting lighter (white) or darker (gray)...

vdcbb:
Okay, I can now see that there is a change from a white to a light gray. This is far from obvious. I'm assuming "white" is On and "light gray" is Off?

[...]

PR #1603, opened 2017-01-13
Update the call view icons and refine the layout.

charlesmchen:
I've done a first pass at the call view icons.

* I've replaced the video and message icons. 
* I've baked a 10% margin into all of the icons to enlarge the hot area of the buttons.
  * I'm using the convention of adding "-wide" to the enlarged icons' names to avoid naming conflicts.

I'll attach screenshots below.

PTAL @michaelkirk


charlesmchen:
The 8 screenshots show every permutation of:

* iPhone 7 Plus (our largest form factor) and iPhone SE (almost our smallest form factor; I I don't have a 4S or 4S simulator that I can test on).
* Incoming and Ongoing modes.
* With and without red lines to indicate hot areas.

screen shot 2017-01-13 at 1 00 08 pm
screen shot 2017-01-13 at 12 59 47 pm
screen shot 2017-01-13 at 12 58 47 pm
screen shot 2017-01-13 at 12 58 18 pm
screen shot 2017-01-13 at 12 57 45 pm
screen shot 2017-01-13 at 12 57 22 pm
screen shot 2017-01-13 at 12 56 46 pm
screen shot 2017-01-13 at 12 56 19 pm

michaelkirk:
## 1. message button on "incoming call" screen.

The "message" button in the incoming screen surprised me. What would it do?

## 2. mute/speaker button state 

I'm a big fan of the system call screen style where the background and stroke are inverted, vs. the less distinguishable change in stroke color here.

see here: https://github.com/WhisperSystems/Signal-iOS/issues/1161 for a related issue in the existing redphone call screen.

What do you think @charlesmchen

michaelkirk:
## 3. button padding

nit: I'm assuming the images have uncropped blank space in an effort to increase their hit area. It might be more flexible to adjust in the future / change per device if we instead crop tightly and use image edge insets.

charlesmchen:
> The "message" button in the incoming screen surprised me. What would it do?

https://github.com/WhisperSystems/Signal-iOS/pull/1595#discussion_r95870505 You said:

> In response to your TODO, I think the textMessageButton could happen any time, since it just dismisses the view controller, it's not dependent on the state of the underlying signaling stuff.

Doesn't iOS offer a way to reject incoming calls with a suite of text message options like "I can't pick up right now?" "I'll call you back." and "Try me again later."  

Anyway, that doesn't sound like something we'll have time to take on for this release, but we could certainly have a "messages" button that a) hangs up b) takes you to a new/existing conversation with the caller... Or we could just skip this button.

charlesmchen:
> I'm a big fan of the system call screen style where the background and stroke are inverted, vs. the less distinguishable change in stroke color here.

That's not a bad idea.  I'll try to do that when I make these buttons stateful, which will be in a forthcoming PR.

michaelkirk:
> In response to your TODO, I think the textMessageButton could happen any time, since it just dismisses the view controller, it's not dependent on the state of the underlying signaling stuff.

Huh. I *did* say that didn't I. I gave you a technically correct but useless answer. What I meant was the functionality of sending text messages wasn't contingent on the state of the underlying webrtc signaling (as opposed to video calling, for which the call must be connected before we can enable the control). 

But *as a feature* I don't think a messaging button on the incoming call screen makes sense. I think we should nix it from the incoming call screen.

> Doesn't iOS offer a way to reject incoming calls with a suite of text message options like "I can't pick up right now?" "I'll call you back." and "Try me again later."

Yeah, hangup-with-excuse would be a good future feature.

michaelkirk:
Cool. Less the "messaging button on incoming call screen" this LGTM.

charlesmchen:
> 1. message button on "incoming call" screen.

Removed.

> 3. button padding

Done.

michaelkirk:
Icon cropping looks good, thanks for doing that. 

Except this one got missed? https://github.com/WhisperSystems/Signal-iOS/pull/1603/commits/19633a9f61d5020b8ee83eae77431170cfbf7601#diff-101775d131e9aac5c3ae6a52fdf5b673

charlesmchen:
Good catch thanks
https://github.com/WhisperSystems/Signal-iOS/commit/8be987de1c7c69c4a62c7f420ecfb52e56a39d1c

PR #1604, opened 2017-01-13
Improve buttons in call view.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/view controllers/CallViewController.swift
+++ b/Signal/src/view controllers/CallViewController.swift
@@ -338,10 +338,10 @@ class CallViewController: UIViewController, CallDelegate {
         let image = UIImage(named:imageName)
         let button = UIButton()
         button.setImage(image, for:.normal)
-        button.imageEdgeInsets = UIEdgeInsetsMake(buttonInset(),
-                                                  buttonInset(),
-                                                  buttonInset(),
-                                                  buttonInset())
+        button.imageEdgeInsets = UIEdgeInsets(top: buttonInset(),
+                                              left: buttonInset(),
+                                              bottom: buttonInset(),
+                                              right: buttonInset())
         button.addTarget(self, action:action, for:.touchUpInside)
         button.autoSetDimension(.width, toSize:buttonSize())
         button.autoSetDimension(.height, toSize:buttonSize())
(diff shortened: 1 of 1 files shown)


Signal/src/view controllers/CallViewController.swift
```diff
@@ -338,10 +338,10 @@ class CallViewController: UIViewController, CallDelegate {
...
         let button = UIButton()
         button.setImage(image, for:.normal)
-        button.imageEdgeInsets = UIEdgeInsetsMake(buttonInset(),
-                                                  buttonInset(),
-                                                  buttonInset(),
-                                                  buttonInset())
+        button.imageEdgeInsets = UIEdgeInsets(top: buttonInset(),
+                                              left: buttonInset(),
+                                              bottom: buttonInset(),
+                                              right: buttonInset())
```
charlesmchen: Style clean up.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -486,19 +486,12 @@ class CallViewController: UIViewController, CallDelegate {
...
-            } else {
-                // Show ongoing controls
-                let isOngoingCallControl = ongoingCallControls.contains(subview)
-                subview.isHidden = !isOngoingCallControl
-            }
-        }
+        incomingCallView.isHidden = !isRinging
+        incomingCallView.isUserInteractionEnabled = isRinging
+        ongoingCallView.isHidden = isRinging
+        ongoingCallView.isUserInteractionEnabled = !isRinging
```
charlesmchen: Fixes the "unresponsive hangup button" issue.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -549,6 +542,8 @@ class CallViewController: UIViewController, CallDelegate {
 
     func didPressTextMessage(sender speakerphoneButton: UIButton) {
         Logger.info("\(TAG) called \(#function)")
+
+        self.dismiss(animated: true)
```
charlesmchen: Is this right?  How do users return to an ongoing call?  How does the UI surface that there _is_ an ongoing call?

michaelkirk: that's the TODO 

=)

michaelkirk:
LGTM bearing in mind that the "text message" feature is incomplete for now (no way to return to the call).

PR #1605, opened 2017-01-13
Add call duration to call view.

charlesmchen:
Show call duration for connected calls.

PTAL @michaelkirk 

screen shot 2017-01-13 at 5 30 21 pm

### DIFF
--- a/Signal/src/view controllers/CallViewController.swift
+++ b/Signal/src/view controllers/CallViewController.swift
@@ -152,19 +152,18 @@ class CallViewController: UIViewController, CallDelegate {
     var thread: TSContactThread!
     var call: SignalCall!
 
-    // MARK: Layout
+    // MARK: Views
 
     var hasConstraints = false
-
-    // MARK: Background
-
     var blurView: UIVisualEffectView!
+    var dateFormatter: DateFormatter?
 
     // MARK: Contact Views
 
     var contactNameLabel: UILabel!
     var contactAvatarView: AvatarImageView!
     var callStatusLabel: UILabel!
+    var callDurationTimer: Timer?
 
     // MARK: Ongoing Call Controls
 
@@ -213,6 +212,21 @@ class CallViewController: UIViewController, CallDelegate {
         callUIAdapter = callService.callUIAdapter
         audioService = CallAudioService()
         super.init(nibName: nil, bundle: nil)
+    }
+
+    // MARK: View Lifecycle
+
+    override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+
+        callDurationTimer?.invalidate()
+        callDurationTimer = nil
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+
+        updateCallUI(callState: call.state)
     }
 
     override func viewDidLoad() {
@@ -472,7 +486,30 @@ class CallViewController: UIViewController, CallDelegate {
         case .answering:
             return NSLocalizedString("IN_CALL_SECURING", comment: "Call setup status label")
         case .connected:
-            return NSLocalizedString("IN_CALL_TALKING", comment: "Call setup status label")
+            if let call = self.call {
+                let callDuration = call.connectionDuration()
+                let callDurationDate = Date(timeIntervalSinceReferenceDate:callDuration)
+                if dateFormatter == nil {
+                    dateFormatter = DateFormatter()
+                    dateFormatter!.dateFormat = "HH:mm:ss"
+                    dateFormatter!.timeZone = TimeZone(identifier:"UTC")!
+                }
+                var formattedDate = dateFormatter!.string(from: callDurationDate)
+                if formattedDate.hasPrefix("00:") {
+                    // Don't show the "hours" portion of the date format unless the 
+                    // call duration is at least 1 hour.
+                    formattedDate = formattedDate.substring(from: formattedDate.index(formattedDate.startIndex, offsetBy: 3))
+                } else {
+                    // If showing the "hours" portion of the date format, strip any leading
+                    // zeroes.
+                    if formattedDate.hasPrefix("0") {
+                        formattedDate = formattedDate.substring(from: formattedDate.index(formattedDate.startIndex, offsetBy: 1))
+                    }
+                }
+                return formattedDate
+            } else {
+                return NSLocalizedString("IN_CALL_TALKING", comment: "Call setup status label")
+            }
         case .remoteBusy:
             return NSLocalizedString("END_CALL_RESPONDER_IS_BUSY", comment: "Call setup status label")
         case .localFailure:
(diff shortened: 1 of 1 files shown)


Signal/src/view controllers/CallViewController.swift
```diff
@@ -506,6 +528,24 @@ class CallViewController: UIViewController, CallDelegate {
 
         default: break
         }
+
+        if callState == .connected {
+            if callDurationTimer == nil {
+                let kDurationUpdateFrequencySeconds = 1 / 20.0
+                callDurationTimer = Timer.scheduledTimer(timeInterval: kDurationUpdateFrequencySeconds,
```
charlesmchen: Wait, this is a retain cycle. Working on it...

charlesmchen: Fixed it.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -472,7 +473,28 @@ class CallViewController: UIViewController, CallDelegate {
         case .answering:
             return NSLocalizedString("IN_CALL_SECURING", comment: "Call setup status label")
         case .connected:
-            return NSLocalizedString("IN_CALL_TALKING", comment: "Call setup status label")
+            if let call = self.call {
+                let callDuration = call.connectionDuration()
+                let callDurationDate = Date(timeIntervalSinceReferenceDate:callDuration)
+                let dateFormatter = DateFormatter()
```
pouriaalmassi: Would it be worth caching this DateFormatter? https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/DataFormatting/Articles/dfDateFormatting10_4.html#//apple_ref/doc/uid/TP40002369-SW10

Signal/src/view controllers/CallViewController.swift
```diff
@@ -506,6 +528,37 @@ class CallViewController: UIViewController, CallDelegate {
 
         default: break
         }
+
+        if callState == .connected {
+            if callDurationTimer == nil {
+                let kDurationUpdateFrequencySeconds = 1 / 20.0
```
michaelkirk: Just curious, how'd you arrive at 20x per second?

Signal/src/view controllers/CallViewController.swift
```diff
@@ -472,7 +473,28 @@ class CallViewController: UIViewController, CallDelegate {
         case .answering:
             return NSLocalizedString("IN_CALL_SECURING", comment: "Call setup status label")
         case .connected:
-            return NSLocalizedString("IN_CALL_TALKING", comment: "Call setup status label")
+            if let call = self.call {
+                let callDuration = call.connectionDuration()
+                let callDurationDate = Date(timeIntervalSinceReferenceDate:callDuration)
```
michaelkirk: style nit: space after `:`

Signal/src/call/SignalCall.swift
```diff
@@ -41,6 +41,16 @@ protocol CallDelegate: class {
     var state: CallState {
         didSet {
             Logger.debug("\(TAG) state changed: \(oldValue) -> \(state)")
+
+            // Update connectedDate
+            if state == .connected {
+                if connectedDate == nil {
+                    connectedDate = NSDate()
+                }
+            } else {
```
michaelkirk: Will this immediately wipe the call duration when the call ends?

ux nit: The call view hangs out for a second when the call was hung up by the remote party. I think we should continue to display the time (frozen) after the other party hangs up.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -506,6 +528,37 @@ class CallViewController: UIViewController, CallDelegate {
...
+            callDurationTimer?.invalidate()
+            callDurationTimer = nil
+        }
+    }
+
+    func updateCallDuration(timer: Timer?) {
+        updateCallUI(callState: call.state)
+    }
+
+    override func viewDidDisappear(_ animated: Bool) {
```
michaelkirk: style nit: I like to keep the view lifecyle methods together at the top, after any inits.

Signal/src/view controllers/CallViewController.swift
michaelkirk: Consider moving all the formatting to the model and out of the VC. Keeps the VC lighter, and makes it easily testable.

pouriaalmassi: Forgive me if I'm going against norms established by this project but keeping formatting at a higher level might easy access to formatters for other models and VCs. Consider the example laid out [here](http://sandmoose.com/post/119962774817/caching-nsdateformatter) and a cautionary explanation on minimizing creation of formatters by bbum [here](http://stackoverflow.com/a/4442389). 

Creating a formatter cache would be out of scope for this PR so a separate PR may be advised.

charlesmchen: I've moved this `DateFormatter` to a lazily-populated class property. Formatting feels like a view concern, so I'd rather it didn't leak into the model.  

Thanks @pouriaalmassi @michaelkirk

Signal/src/view controllers/CallViewController.swift
charlesmchen: Done. Too bad clang-format doesn't catch this. I can't find a corresponding style option. http://clang.llvm.org/docs/ClangFormatStyleOptions.html

charlesmchen:
> Consider moving all the formatting to the model and out of the VC. Keeps the VC lighter, and makes it easily testable.

Formatting seems like a view concern, and I'd rather not leak it into the model. Some people use ViewModels for this kind of logic, but this is just a handful of lines, so it doesn't seem worthwhile. I'm personally much more concerned with keeping models "dumb" and "light" than VCs.

Signal/src/view controllers/CallViewController.swift
charlesmchen: It was a bit arbitrary.  I just wanted the frame rate to be fast enough that the ticking of seconds would feel smooth and regular.

Signal/src/view controllers/CallViewController.swift
charlesmchen: +1 Done.

Signal/src/call/SignalCall.swift
charlesmchen: >I think we should continue to display the time (frozen) after the other party hangs up.

I'm using the same UILabel to display the "call status" (Dialing, Call Ended) and the "call duration", just like the iOS call screen.   The "call duration" is shown while the call is connected. When the call ends, the call status changes to something like "Call Ended," again, just like the iOS call screen.  I like how the iOS call screen works - what's the argument for deviating?

Signal/src/view controllers/CallViewController.swift
michaelkirk: I'm not sure that clang format supports swift. I believe the swift linter catches this though.

PR #1614, opened 2017-01-18
Don't start recording until call is visible

michaelkirk:
In the upcoming webrtc branch, previously we were recording before the call was completely established, causing the user to be inexplicably prompted for recording permissions before they could see an incoming call.

Bulk of the work in: https://github.com/WhisperSystems/Signal-iOS/commit/dbb29d7d7ef7635e22e8e737f9a0a94d86a052cb

> Don't require recording permissions until call is ringing.
> We do this by manually managing the RTCAudioSession.
> Unfortunately to do this we have to include a couple of RTC headers not
> exported by the default build of WebRTC.framework (see: Libraries/WebRTC)

and the first two commits are just included for expedience.

### DIFF
--- a/Libraries/WebRTC/README.md
+++ b/Libraries/WebRTC/README.md
@@ -1,0 +1,7 @@
+The RTCAudioSession.h header isn't included in the standard build of
+WebRTC, so we've vendored it here. Otherwise we're using the vanilla
+framework.
+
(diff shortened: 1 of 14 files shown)


michaelkirk:
@charlesmchen PTAL

charlesmchen:
LGTM

PR #1615, opened 2017-01-18
Disable unused CallKit buttons.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/view controllers/CallViewController.swift
+++ b/Signal/src/view controllers/CallViewController.swift
@@ -647,8 +647,8 @@ class CallViewController: UIViewController, CallDelegate {
 
     internal func stateDidChange(call: SignalCall, state: CallState) {
         DispatchQueue.main.async {
+            Logger.info("\(self.TAG) new call status: \(state)")
             self.updateCallUI(callState: state)
-            Logger.info("\(self.TAG) new call status: \(call.state)")
         }
         self.audioService.handleState(state)
     }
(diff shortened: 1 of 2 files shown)


Signal/src/view controllers/CallViewController.swift
```diff
@@ -519,7 +519,6 @@ class CallViewController: UIViewController, CallDelegate {
 
     func updateCallUI(callState: CallState) {
         let textForState = localizedTextForCallState(callState)
-        Logger.info("\(TAG) new call status: \(callState) aka \"\(textForState)\"")
```
charlesmchen: I moved this logging to avoid rampant redundant logging due to the call duration time.

michaelkirk: I took a different approach in #1614 with https://github.com/WhisperSystems/Signal-iOS/pull/1614/commits/0797df19b97b6bf2fd7999012aea5f74cdaf7d8c

Thoughts?

michaelkirk:
LGTM.

Signal/src/view controllers/CallViewController.swift
charlesmchen: I'm sorry it looks like you just took out the call state logging without replacing it with something else.  Do you want to remove it?

michaelkirk: Whoops. I didn't intend to delete the logging line, I'll put it back, but the substantive part of the commit was to *only* redraw the status label portion of the view 20 x per second, rather than the rest of the updateCallUI method.

charlesmchen: Sure, that's a reasonable optimization. But IMO the value of the logging is to surface the changes in the state, not to document calls to updateCallUI(), so my inclination is to move it to the state setter.

michaelkirk: Good point. State change is already logged in the SignalCall, so leaving out the call in updateCallUI.

PR #1619, opened 2017-01-18
fix non-callkit ringer

michaelkirk:
- Move the CallAudioService class to it's own file
- Only start/stop ringing on NonCallKitAdapter, as CallKit handles ringing for us.
- Remove dependency on the AppAudioManager, which is a RedPhone artifact, which had very little purpose in the new call flow.


Signal/src/view controllers/CallViewController.swift
```diff
@@ -187,15 +81,15 @@ class CallViewController: UIViewController, CallDelegate {
         contactsManager = Environment.getCurrent().contactsManager
         let callService = Environment.getCurrent().callService!
         callUIAdapter = callService.callUIAdapter
-        audioService = CallAudioService()
+        audioService = CallAudioService(handleRinging: callUIAdapter.hasManualRinger)
```
charlesmchen: I don't understand why CallAudioService is owned by CallVC.  CallVC could be dismissed (so user can message).  Wouldn't we want to keep ringing even though user has left CallVC?

Signal/src/call/CallAudioService.swift
```diff
@@ -0,0 +1,177 @@
...
+
+    // MARK: - Ringing
+
+    private func startRinging() {
+        guard handleRinging else {
+            Logger.debug("\(TAG) ignoring \(#function) since CallKit handles it's own ringing state")
+            return
+        }
+
+        vibrateTimer = Timer.scheduledTimer(timeInterval: vibrateRepeatDuration, target: self, selector: #selector(ringVibration), userInfo: nil, repeats: true)
```
charlesmchen: `NSTimer`/`Timer` is a classic source of leaks, because it holds a strong reference to `target`.  I think that'll bite us here.  In the past, what I've done is add a NSTimer category with methods that take a weak reference.

Signal/src/view controllers/CallViewController.swift
michaelkirk: That's valid. I'm not really sure where it belongs - it's behavior is different depending on CallKit vs. Not.

Maybe it belongs on the callUIAdapter.

Signal/src/call/CallAudioService.swift
michaelkirk: ooph. Thanks.

michaelkirk:
PTAL @charlesmchen

Signal/src/call/SignalCall.swift
```diff
@@ -50,20 +51,35 @@ protocol CallDelegate: class {
...
+            }
+        }
+    }
+
+    var isSpeakerphoneEnabled = false {
+        didSet {
+            Logger.debug("\(TAG) isSpeakerphoneEnabled changed: \(oldValue) -> \(isSpeakerphoneEnabled)")
+            for observer in observers {
+                observer.value?.speakerphoneDidChange(call: self, isEnabled: isSpeakerphoneEnabled)
+            }
```
charlesmchen: Hmm, this will intersect with my PR a little bit. Hopefully you'll merge first.

Signal/src/call/UserInterface/CallUIAdapter.swift
```diff
@@ -39,6 +40,7 @@ class CallUIAdapter {
     let TAG = "[CallUIAdapter]"
     private let adaptee: CallUIAdaptee
     private let contactsManager: OWSContactsManager
+    private let audioService: CallAudioService
```
charlesmchen: Could we eliminate the `hasManualRinger` property and push `audioService` into `NonCallKitCallUIAdaptee`? Then, `CallKitCallUIAdaptee` wouldn't even instantiate a `CallAudioService`. This way is fine too, just wondering.

Signal/src/call/UserInterface/CallUIAdapter.swift
```diff
@@ -94,6 +102,19 @@ class CallUIAdapter {
...
 
     internal func toggleMute(call: SignalCall, isMuted: Bool) {
+        // With CallKit, muting is handled by a CXAction, so it must go through the adaptee
         adaptee.toggleMute(call: call, isMuted: isMuted)
     }
+
+    internal func toggleSpeakerphone(call: SignalCall, isEnabled: Bool) {
+        // Speakerphone is not handled by CallKit (e.g. there is no CXAction), so we handle it w/o going through the 
+        // adaptee, relying on the AudioService CallObserver to put the system in a state consistent with the call's 
+        // assigned property.
```
charlesmchen: Huh, so we can't keep the speakerphone state in sync?  The CallKit UI's speakerphone button definitely has "selected" and "unselected" states.

Signal/src/util/WeakTimer.swift
```diff
@@ -0,0 +1,43 @@
...
+//
+
+/**
+ * As of iOS10, the timer API's take a block, which makes it easy to reference weak self in Swift. This class offers a 
+ * similar API that works pre iOS10.
+ *
+ * Solution modified from
+ * http://stackoverflow.com/questions/16821736/weak-reference-to-nstimer-target-to-prevent-retain-cycle/41003985#41003985
+ */
+final class WeakTimer {
```
charlesmchen: Nice. 👏

charlesmchen: There shouldn't be any problem using this class from Objective-C, right?

Signal/src/call/UserInterface/CallUIAdapter.swift
```diff
@@ -55,9 +57,13 @@ class CallUIAdapter {
             Logger.info("\(TAG) choosing non-callkit adaptee for older iOS")
             adaptee = NonCallKitCallUIAdaptee(callService: callService, notificationsAdapter: notificationsAdapter)
         }
+
+        audioService = CallAudioService(handleRinging: adaptee.hasManualRinger)
     }
 
     internal func reportIncomingCall(_ call: SignalCall, thread: TSContactThread) {
+        call.addObserverAndSyncState(observer: audioService)
```
charlesmchen: What will happen if user receives _two_ incoming calls before accepting either?  Could ring state end up in a bad state?

Signal/src/call/CallAudioService.swift
```diff
@@ -0,0 +1,189 @@
...
+            setAudioSession(category: AVAudioSessionCategoryPlayAndRecord, options: .defaultToSpeaker)
+        } else {
+            setAudioSession(category: AVAudioSessionCategoryPlayAndRecord)
+        }
+    }
+
+    // MARK: - Service action handlers
+
+    public func handleState(_ state: CallState) {
+        Logger.verbose("\(TAG) in \(#function) new state: \(state)")
```
charlesmchen: Should we assert that we're on the main thread here?

Signal/src/call/UserInterface/CallUIAdapter.swift
michaelkirk: > Huh, so we can't keep the speakerphone state in sync?

We **aren't**. But I think we could.

A solution might involve using notification center to observe changes in the audio session route. The details of which I haven't solidified, as I think it's lower priority. To be clear this would affect people who toggle speakerphone on callkit, and then switch to the system call screen. (The reverse, where someone toggles speakerphone in app then switches to CallKit works, presumably because the system call screen is doing something outside of CallKit, like observing the AudioSession route.

Signal/src/util/WeakTimer.swift
michaelkirk: I think we'd just have to mark the class as `@objc`

Signal/src/call/UserInterface/CallUIAdapter.swift
michaelkirk: I'd say it's the CallService's responsibility to ensure that `CallUIAdapter.reportIncomingCall` is only called when the app is "ready" to report an incoming call (e.g. it there's not already another call in-progress).

As far as the the ringer being in a "bad state", I don't *think* that's possible, as the call service coordinates that we're not reportingIncomingCall for multiple calls, and the audioService is intended to be idempotent WRT the signalCall's state changes.

Signal/src/call/UserInterface/CallUIAdapter.swift
michaelkirk: But CallKitUIAdaptee uses the CallAudioService, just not for ringing.

charlesmchen: Gotcha

PR #1620, opened 2017-01-18
fix notification actions

michaelkirk:
### Fix "Answer" from iOS9 notification doesn't init audio

The removed code was from an older eon. CallService shouldn't be touched
except via the CallUIAdapter since only there is the omnipresent
distinction between CallKit vs. NonCallKit made.

i.e. when the RTCAudioSession get's started depends on the
CallUIAdaptee.

// FREEBIE

### DIFF
--- a/Signal/src/call/NonCallKitCallUIAdaptee.swift
+++ b/Signal/src/call/NonCallKitCallUIAdaptee.swift
@@ -22,7 +22,9 @@ class NonCallKitCallUIAdaptee: CallUIAdaptee {
         self.notificationsAdapter = notificationsAdapter
     }
 
-    func startOutgoingCall(_ call: SignalCall) {
+    func startOutgoingCall(handle: String) -> SignalCall {
+        let call = SignalCall.outgoingCall(localId: UUID(), remotePhoneNumber: handle)
+
         CallService.signalingQueue.async {
             _ = self.callService.handleOutgoingCall(call).then {
                 Logger.debug("\(self.TAG) handleOutgoingCall succeeded")
@@ -30,6 +32,8 @@ class NonCallKitCallUIAdaptee: CallUIAdaptee {
                 Logger.error("\(self.TAG) handleOutgoingCall failed with error: \(error)")
             }
         }
+
+        return call
     }
 
     func reportIncomingCall(_ call: SignalCall, callerName: String) {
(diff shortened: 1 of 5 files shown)


michaelkirk:
PTAL @charlesmchen

Signal/src/call/NonCallKitCallUIAdaptee.swift
```diff
@@ -51,13 +55,35 @@ class NonCallKitCallUIAdaptee: CallUIAdaptee {
         notificationsAdapter.presentMissedCall(call, callerName: callerName)
     }
 
+    func answerCall(localId: UUID) {
+        CallService.signalingQueue.async {
+            guard let call = self.callService.call else {
+                assertionFailure("\(self.TAG) in \(#function) No current call.")
+                return
+            }
+
```
charlesmchen: Is the parameter the id of the call?  Should we assert that it matches the id of the "current" call?

Signal/src/call/NonCallKitCallUIAdaptee.swift
```diff
@@ -51,13 +55,35 @@ class NonCallKitCallUIAdaptee: CallUIAdaptee {
...
             self.callService.handleAnswerCall(call)
         }
     }
 
+    func declineCall(localId: UUID) {
+        CallService.signalingQueue.async {
+            guard let call = self.callService.call else {
+                assertionFailure("\(self.TAG) in \(#function) No current call.")
+                return
+            }
```
charlesmchen: Ditto

Signal/src/call/UserInterface/CallUIAdapter.swift
```diff
@@ -31,13 +35,25 @@ extension CallUIAdaptee {
...
+        }
+    }
 }
 
 /**
  * Notify the user of call related activities.
  * Driven by either a CallKit or System notifications adaptee
  */
-class CallUIAdapter {
+@objc class CallUIAdapter: NSObject {
```
charlesmchen: Why has this been converted to a NSObject?

michaelkirk: because it's now being used in objc (in PushManager.m)

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/6c14f2f5002b7698aac4998bbc33db6c439c9f82

PR #1626, opened 2017-01-20
[SSK] remove Cuba from domain fronting

michaelkirk:
Current implementation is blocked in Cuba, plus there are reports of
Signal working w/o domain fronting anyway.

PTAL @charlesmchen 

// FREEBIE

### DIFF
--- a/Signal/Signal-Info.plist
+++ b/Signal/Signal-Info.plist
@@ -21,7 +21,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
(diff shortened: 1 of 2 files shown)


Signal/Signal-Info.plist
```diff
@@ -21,7 +21,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>2.6.12</string>
+	<string>2.6.13</string>
```
charlesmchen: Do the changes in this file belong in this PR?

PR #1630, opened 2017-01-22
prevent session corruption

michaelkirk:
(based on #1626 - ignore the commit re: Cuban censorship)

Coordinate SignalProtocol encryption/decryption on a single serial queue. Previously message sending encrypted on the sending thread, while message receiving decrypted on the main thread.

Though I was not able to reliably reproduce this, this is a likely source of session corruption. E.g. when receiving and sending messages simultaneously we compute and store the state, potentially clobbering the newly generated state from sending with the newly generated state from receiving.

**Corresponding PRs:**

SignalProtocol: https://github.com/WhisperSystems/SignalProtocolKit/pull/14
SignalService: https://github.com/WhisperSystems/SignalServiceKit/pull/98

PTAL @charlesmchen

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -3,8 +3,9 @@ source 'https://github.com/CocoaPods/Specs.git'
 
 target 'Signal' do
     pod 'SocketRocket',               :git => 'https://github.com/facebook/SocketRocket.git'
-    pod 'AxolotlKit', git: 'https://github.com/WhisperSystems/SignalProtocolKit.git'
-    pod 'SignalServiceKit',           git: 'https://github.com/WhisperSystems/SignalServiceKit.git', branch: 'master'
+    pod 'AxolotlKit',                 git: 'https://github.com/WhisperSystems/SignalProtocolKit.git', branch: 'mkirk/session-corruption'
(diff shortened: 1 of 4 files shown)


Podfile
```diff
@@ -3,8 +3,9 @@ source 'https://github.com/CocoaPods/Specs.git'
 
 target 'Signal' do
     pod 'SocketRocket',               :git => 'https://github.com/facebook/SocketRocket.git'
-    pod 'AxolotlKit', git: 'https://github.com/WhisperSystems/SignalProtocolKit.git'
-    pod 'SignalServiceKit',           git: 'https://github.com/WhisperSystems/SignalServiceKit.git', branch: 'master'
+    pod 'AxolotlKit',                 git: 'https://github.com/WhisperSystems/SignalProtocolKit.git', branch: 'mkirk/session-corruption'
```
charlesmchen: Should we rename the `AxolotlKit` pod to `SignalProtocolKit`?

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/4e72ab92c8de6a4450ace111b35c71af28f25cab

PR #1631, opened 2017-01-23
Elaborate logging in app delegate.

charlesmchen:
PTAL @michaelkirk 

Hmm, I need to find the time to revisit #1584


michaelkirk:
LGTM :+1:

PR #1632, opened 2017-01-23
Charlesmchen/webrtc/video

charlesmchen:
PTAL @michaelkirk


Signal/Images.xcassets/mute-selected-wide.imageset/Contents.json
```diff
@@ -2,7 +2,7 @@
   "images" : [
     {
       "idiom" : "universal",
-      "filename" : "video-active.png",
+      "filename" : "mute-selected-wide.png",
```
charlesmchen: I know you want a different style for the "unselected" icons, but I'm trying to prioritize getting things working first.  I'll return to these icons if we have time.

Signal/src/AppDelegate.m
```diff
@@ -317,7 +317,13 @@ - (void)application:(UIApplication *)application
  */
 - (BOOL)application:(UIApplication *)application continueUserActivity:(nonnull NSUserActivity *)userActivity restorationHandler:(nonnull void (^)(NSArray * _Nullable))restorationHandler
 {
-    DDLogWarn(@"%@ called %s with userActivity: %@, but not yet supported.", self.tag, __PRETTY_FUNCTION__, userActivity);
+    if ([userActivity.activityType isEqualToString:@"INStartVideoCallIntent"]) {
+        [[Environment getCurrent].callService handleCallKitStartVideo];
```
charlesmchen: AFAIK this is the only way to observe when user taps video button in CallKit ui.

Signal/src/call/CallService.swift
```diff
@@ -608,7 +610,7 @@ fileprivate let timeoutSeconds = 60
         call.state = .connected
 
         // We don't risk transmitting any media until the remote client has admitted to being connected.
-        peerConnectionClient.setAudioEnabled(enabled: true)
+        peerConnectionClient.setAudioEnabled(enabled: !call.isMuted)
```
charlesmchen: Caller could mute the call before callee accepts call, so we should honor current model state.

Signal/src/call/CallService.swift
```diff
@@ -713,7 +715,7 @@ fileprivate let timeoutSeconds = 60
      *
      * Can be used for Incoming and Outgoing calls.
      */
-    func handleToggledMute(isMuted: Bool) {
+    func setIsMuted(isMuted: Bool) {
```
charlesmchen: There's a lot of "toggle mute" methods that are actually setters.  I've renamed them appropriately.

Signal/src/call/Speakerbox/CallKitCallManager.swift
```diff
@@ -48,14 +49,22 @@ final class CallKitCallManager: NSObject {
...
         requestTransaction(transaction)
     }
 
+//    func setHasVideo(call: SignalCall, hasVideo: Bool) {
+//        let update = CXCallUpdate()
+//        update.hasVideo = call.hasVideo
+//        
+//        // Report the incoming call to the system
+//        provider.reportCall(with: call.localId, updated update: update)
+//    }
```
charlesmchen: Whoops, gotta tidy this up before you review.

Signal/src/AppDelegate.m
michaelkirk: I'm only a little surprised.

Just to keep in mind the comment below, we'll eventually also use this method to initiate a CallKit voice call when someone chooses to initiate a Signal call from their contacts, or from the "callback" action on the system call screen when a call fails.

Signal/src/call/CallService.swift
michaelkirk: This is a good change. Did you test that this behaves as expected? E.g. the call is initially muted, but it's possible for the caller to un-mute?

The reason I wanted to double check was that there is some plumbing related to setting up the AudioSession in RTCAudioSession that occurs only once you `setAudioEnabled(enabled: true)`.

Signal/src/AppDelegate.m
charlesmchen: We should also look into how this fits into having an "intents extension".

Signal/src/call/CallService.swift
charlesmchen: Yup, this works.

charlesmchen:
> Still TODO of course is the actual display of video. Did you want to merge this independent of that?

Yup.

a) We're committed to shipping video.
b) Small branches!
c) If we ever decided to ship without video, we could hide/disable video with a tiny commit.

Signal/src/call/Speakerbox/CallKitCallUIAdaptee.swift
```diff
@@ -139,7 +153,7 @@ final class CallKitCallUIAdaptee: NSObject, CallUIAdaptee, CXProviderDelegate {
         CallService.signalingQueue.async {
             self.callService.handleOutgoingCall(call).then {
                 action.fulfill()
-            }.catch { error in
+            }.catch { _ in
                 self.callManager.removeCall(call)
```
charlesmchen: Done.

charlesmchen:
> This is a good change. Did you test that this behaves as expected? E.g. the call is initially muted, but it's possible for the caller to un-mute?

These "pre-mute" changes work for the mute button in the call view, but the callkit UI doesn't seem available on caller's phone while ringing, so the second change is future-proofing against that ever changing.

Issue #1627, opened 2017-01-21
User prompted to re-register but still receives notifications, OS keychain corrupts/loses the database decryption key


riyapenn:
- [ X ] I have searched open and closed issues for duplicates most relevant: 

----------------------------------------

### Bug description
User still receiving notifications for Signal, but when opening Signal they are prompted to re-register; last opened Signal probably the night before (before battery died)

### Steps to reproduce
1. Phone died
2. Left it charging all night
3. Turn on phone
4. Received notifications of incoming messages and saw a badge count
5. Opened Signal

**Actual result:** See registration page prompting to re-register
**Expected result:** Normal behavior of messaging

### Device info

**Device**: iPhone 7+
**iOS version**: 10.2
**Signal version:** 2.6.12

### Link to debug log

User 1: https://gist.github.com/cad2e42437d0414720b8abbada725d07
User 2: https://gist.github.com/46a4401eae612a451c16753fbfa7a9bf 
User 3: https://gist.github.com/cf4aa713af70a0b47cef23c82b159df1 
User 4: https://gist.github.com/e3f38da03854e96615f8d084c6aa3596 
User 5: https://gist.github.com/3308be2ff38dc008032751643527c414 
User 6: https://gist.github.com/80011e66dc06f5c7691c6a69cbf4dc48

michaelkirk:
Maybe related to this: https://forums.developer.apple.com/thread/4743?start=105&tstart=0

FredericJacobs:
+1 user report on this

riyapenn:
Resurfacing? https://gist.github.com/anonymous/c0d452ba95da30279244b2371d1632f8

riyapenn:
User on 2.6.15 is stating this is resurfacing.  https://gist.github.com/anonymous/6ccbe5903127eec4599bc523356ba42a

PR #1636, opened 2017-01-23
Bail on startup if DB password is inaccessible

michaelkirk:
Most likely this would be  because the user hasn't unlocked their device
since last restart.

This behavior existed once before, but the startup ordering is pretty
delicate. So, we're now redundantly checking in SSK in case this
delicate startup logic gets mis-ordered again.

Also fixed the AppDelegate method to check for the proper
applicationState, since it will never be "active" in didFinishLaunching.

fixes https://github.com/WhisperSystems/Signal-iOS/issues/1627

corresponding SSK: https://github.com/WhisperSystems/SignalServiceKit/pull/99

PTAL @charlesmchen 

// FREEBIE

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -386,7 +386,7 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
         return;
     }
 
-    if ([TSStorageManager isDatabasePasswordAccessible]) {
+    if (![TSStorageManager isDatabasePasswordAccessible]) {
         UILocalNotification *notification = [[UILocalNotification alloc] init];
         notification.alertBody            = NSLocalizedString(@"PHONE_NEEDS_UNLOCK", nil);
         [[UIApplication sharedApplication] presentLocalNotificationNow:notification];
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -58,13 +58,26 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 }
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
-    // Initializing logger
-    CategorizingLogger *logger = [CategorizingLogger categorizingLogger];
-    [logger addLoggingCallback:^(NSString *category, id details, NSUInteger index){
-    }];
+    BOOL loggingIsEnabled;
+#ifdef DEBUG
+    // Specified at Product -> Scheme -> Edit Scheme -> Test -> Arguments -> Environment to avoid things like
+    // the phone directory being looked up during tests.
+    loggingIsEnabled = TRUE;
+    [DebugLogger.sharedLogger enableTTYLogging];
+#elif RELEASE
+    loggingIsEnabled = Environment.preferences.loggingIsEnabled;
+#endif
+    if (loggingIsEnabled) {
+        [DebugLogger.sharedLogger enableFileLogging];
+    }
 
     // XXX - careful when moving this. It must happen before we initialize TSStorageManager.
     [self verifyDBKeysAvailableBeforeBackgroundLaunch];
+
+    // Initializing env logger
+    CategorizingLogger *logger = [CategorizingLogger categorizingLogger];
+    [logger addLoggingCallback:^(NSString *category, id details, NSUInteger index){
+    }];
 
     // Setting up environment
     [Environment setCurrent:[Release releaseEnvironmentWithLogging:logger]];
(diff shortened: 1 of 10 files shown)


Signal/src/AppDelegate.m
```diff
@@ -378,31 +380,23 @@ - (void)application:(UIApplication *)application
...
-- (void)verifyBackgroundBeforeKeysAvailableLaunch {
-    if ([self applicationIsActive]) {
+- (void)verifyDBKeysAvailableBeforeBackgroundLaunch
+{
+    if (UIApplication.sharedApplication.applicationState != UIApplicationStateBackground) {
         return;
     }
 
-    if (![[TSStorageManager sharedManager] databasePasswordAccessible]) {
+    if ([TSStorageManager isDatabasePasswordAccessible]) {
```
charlesmchen: Wait, shouldn't this have a logical not `!` ?

Signal/src/AppDelegate.m
```diff
@@ -63,6 +63,9 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     [logger addLoggingCallback:^(NSString *category, id details, NSUInteger index){
     }];
 
+    // XXX - careful when moving this. It must happen before we initialize TSStorageManager.
+    [self verifyDBKeysAvailableBeforeBackgroundLaunch];
```
charlesmchen: I'd really like to review our ordering of operations in app startup. It looks like this is being done _before_ logging is enabled down below which is probably a serious issue.

Signal/src/environment/PropertyListPreferences.m
```diff
@@ -130,9 +120,21 @@ - (void)setHasRegisteredVOIPPush:(BOOL)enabled
     [self setValueForKey:PropertyListPreferencesKeyHasRegisteredVoipPush toValue:@(enabled)];
 }
 
+- (BOOL)loggingIsEnabled
+{
+    NSNumber *preference = [NSUserDefaults.standardUserDefaults objectForKey:PropertyListPreferencesKeyEnableDebugLog];
```
charlesmchen: This looks good, but it doesn't look like it will migrate the old preferences value. When we load the database, I proposed that we:

* Check the old preference. If it's set to "disable logging", then:
* Remove the old preference if possible or set it to "enable logging" if not (to avoid overwriting the new property twice) and:
* Set new new preference to "disable logging".

Signal/src/AppDelegate.m
```diff
@@ -378,31 +378,23 @@ - (void)application:(UIApplication *)application
...
+{
+    if (UIApplication.sharedApplication.applicationState != UIApplicationStateBackground) {
         return;
     }
 
-    if (![[TSStorageManager sharedManager] databasePasswordAccessible]) {
+    if (![TSStorageManager isDatabasePasswordAccessible]) {
         UILocalNotification *notification = [[UILocalNotification alloc] init];
         notification.alertBody            = NSLocalizedString(@"PHONE_NEEDS_UNLOCK", nil);
         [[UIApplication sharedApplication] presentLocalNotificationNow:notification];
```
charlesmchen: I'm not sure we should present a "needs unlock" local notification. I think it'd be a better UX to say "New Messages in Signal!" or something positive like that rather than a negative or imperative message.

michaelkirk: OK. I removed it. The user will see a subsequent "New Message" notification 30 seconds later anyway.

Signal/src/environment/PropertyListPreferences.m
michaelkirk: I opted for an explicit one time migration vs lazily-migrating for future code simplicity: https://github.com/WhisperSystems/Signal-iOS/pull/1636/commits/907e122d62afa0b0e612c750273a338edb35aadb

michaelkirk:
PTAL @charlesmchen

Signal/src/environment/PropertyListPreferences.m
charlesmchen: That seems good.  

Hmm, so the migration happens _after_ the "is logging enabled" preference is checked at app startup? I see you turn off logging at that time. Since no PII or user activity is logged by that point - maybe just a line or two about the current app version and the "app launched" event, that seems fine.

michaelkirk: > Hmm, so the migration happens after the "is logging enabled" preference is checked at app startup?

The migration needs to read the database, but the purpose of these changes are to enable logging before the database is set up. So it's a pickle. I think one-time logging generic startup information is a reasonable concession. Here's the output:

```
2017/01/23 20:07:19:147  [VersionMigrations] Checking migrations. currentVersion: 2.6.13.1, lastRanVersion: 2.6.12.0
2017/01/23 20:07:19:151  [OWSDatabaseMigrationRunner] Skipping previously run migration: <OWS100RemoveTSRecipientsMigration: 0x16e3ae60> {
    uniqueId = 100;
}
2017/01/23 20:07:19:152  [OWSDatabaseMigrationRunner] Skipping previously run migration: <OWS101ExistingUsersBlockOnIdentityChange: 0x16e0e2c0> {
    uniqueId = 101;
}
2017/01/23 20:07:19:153  [OWSDatabaseMigrationRunner] Running migration: <OWS102MoveLoggingPreferenceToUserDefaults: 0x16e43b50> {
    uniqueId = 102;
}
2017/01/23 20:07:19:202  [AppDelegate] applicationDidBecomeActive.
2017/01/23 20:07:19:360  [OWS102MoveLoggingPreferenceToUserDefaultsMigrationId] copying existing logging preference to NSUserDefaults
2017/01/23 20:07:48:517  [OWS102MoveLoggingPreferenceToUserDefaults] assigning existing value: 0
2017/01/23 20:08:02:004  [OWS102MoveLoggingPreferenceToUserDefaults] Disabling file logger after one-time log settings migration.
```

PR #1645, opened 2017-01-26
Charlesmchen/webrtc/video2 

charlesmchen:
Roughs out the video-related views.

* Adds views for local and remote video to call view.
* Adds plumbing for the local and remote video tracks from the `PeerConnectionClient` to `CallService`.
* Adds plumbing for the local and remote video tracks from the `CallService` to `CallViewController`.
* Ensures observers of `SignalCall` and `CallService` are always invoked on main thread.
* Improve thread safety of `SignalCall` and `CallService`.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -80,11 +80,24 @@ enum CallError: Error {
 // FIXME TODO do we need to timeout?
 fileprivate let timeoutSeconds = 60
 
-@objc class CallService: NSObject, PeerConnectionClientDelegate {
+// All Observer methods will be invoked from the main thread.
+protocol CallServiceObserver: class {
+    /**
+     * Fired whenever the local or remote video track become active or inactive.
+     */
+    func didUpdateVideoTracks(localVideoTrack: RTCVideoTrack?,
+                              remoteVideoTrack: RTCVideoTrack?)
+}
+
+// This class' state should only be accessed on the signaling queue, _except_
+// the observer-related state which only be accessed on the main thread.
+@objc class CallService: NSObject, CallObserver, PeerConnectionClientDelegate {
 
     // MARK: - Properties
 
     let TAG = "[CallService]"
+
+    var observers = [Weak<CallServiceObserver>]()
 
     // MARK: Dependencies
 
@@ -104,7 +117,18 @@ fileprivate let timeoutSeconds = 60
     var peerConnectionClient: PeerConnectionClient?
     // TODO code cleanup: move thread into SignalCall? Or refactor messageSender to take SignalRecipient identifier.
     var thread: TSContactThread?
-    var call: SignalCall?
+    var call: SignalCall? {
+        didSet {
+            assertOnSignalingQueue()
+
+            oldValue?.removeObserver(self)
+            call?.addObserverAndSyncState(observer: self)
+
+            DispatchQueue.main.async { [weak self] in
+                self?.updateIsVideoEnabled()
+            }
+        }
+    }
 
     /**
      * In the process of establishing a connection between the clients (ICE process) we must exchange ICE updates.
@@ -122,6 +146,26 @@ fileprivate let timeoutSeconds = 60
 
     // Used to coordinate promises across delegate methods
     var fulfillCallConnectedPromise: (() -> Void)?
+
+    weak var localVideoTrack: RTCVideoTrack? {
+        didSet {
+            assertOnSignalingQueue()
+
+            Logger.info("\(self.TAG) \(#function)")
+
+            fireDidUpdateVideoTracks()
+        }
+    }
+
+    weak var remoteVideoTrack: RTCVideoTrack? {
+        didSet {
+            assertOnSignalingQueue()
+
+            Logger.info("\(self.TAG) \(#function)")
+
+            fireDidUpdateVideoTracks()
+        }
+    }
 
     required init(accountManager: AccountManager, contactsManager: OWSContactsManager, messageSender: MessageSender, notificationsAdapter: CallNotificationsAdapter) {
         self.accountManager = accountManager
@@ -591,7 +635,7 @@ fileprivate let timeoutSeconds = 60
 
         // We don't risk transmitting any media until the remote client has admitted to being connected.
         peerConnectionClient.setAudioEnabled(enabled: !call.isMuted)
-        peerConnectionClient.setVideoEnabled(enabled: call.hasVideo)
+        peerConnectionClient.setLocalVideoEnabled(enabled: shouldHaveLocalVideoTrack())
     }
 
     /**
@@ -731,7 +775,7 @@ fileprivate let timeoutSeconds = 60
         }
 
         call.hasVideo = hasVideo
-        peerConnectionClient.setVideoEnabled(enabled: hasVideo)
+        peerConnectionClient.setLocalVideoEnabled(enabled: shouldHaveLocalVideoTrack())
     }
 
     func handleCallKitStartVideo() {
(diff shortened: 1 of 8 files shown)


Signal/src/call/PeerConnectionClient.swift
```diff
@@ -131,35 +141,45 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
             return
         }
 
+        // TODO: What are the best values to use here?
+        let mediaConstraintsDictionary = [
+            kRTCMediaConstraintsMinWidth: "240",
+            kRTCMediaConstraintsMinHeight: "320",
+            kRTCMediaConstraintsMaxWidth: "240",
+            kRTCMediaConstraintsMaxHeight: "320"
```
charlesmchen: WebRTC seems to ignore these values, so I'll probably take them out.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -342,9 +378,60 @@ class CallViewController: UIViewController, CallObserver {
...
+
+        var constraints: [NSLayoutConstraint] = []
+
+        // We fill the screen with the remote video. The remote video's
+        // aspect ratio may not (and in fact will very rarely) match the 
+        // aspect ratio of the current device, so parts of the remote
+        // video will be hidden offscreen.  
+        //
+        // It's better to trim the remote video than to adopt a letterboxed
+        // layout.
```
charlesmchen: @michaelkirk thoughts?

Signal/src/call/CallService.swift
```diff
@@ -833,17 +877,38 @@ fileprivate let timeoutSeconds = 60
...
+    }
+
     // MARK: Helpers
 
     /**
      * Ensure that all `SignalCall` and `CallService` state is synchronized by only mutating signaling state in 
      * handleXXX methods, and putting those methods on the signaling queue.
+     *
+     * TODO: We might want to move this queue and method to OWSDispatch so that we can assert this in
+     *       other classes like SignalCall as well.
```
michaelkirk: 👍

Signal/src/call/SignalCall.swift
```diff
@@ -39,51 +42,97 @@ protocol CallObserver: class {
...
     let localId: UUID
-    
+
     var hasVideo = false {
         didSet {
-            Logger.debug("\(TAG) hasVideo changed: \(oldValue) -> \(hasVideo)")
-            for observer in observers {
-                observer.value?.hasVideoDidChange(call: self, hasVideo: hasVideo)
+            // This should only occur on the signaling queue.
+            objc_sync_enter(self)
```
charlesmchen: SignalCall state is accessed and its methods called from more than one thread, not just the signaling queue. I'd like to propose that we try to move SignalCall to being accessed only on the main thread. Until that time, I've protected all of its state through synchronization.

More generally, I'd like to propose that we have an explicit understanding that all state should only be accessed from one thread. 

I'd also like to propose that signaling queue become a private implementation detail of ServiceCall.  

Looking forward to hearing your thoughts about this.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -132,19 +139,36 @@ class CallViewController: UIViewController, CallObserver {
...
+
         createContactViews()
         createOngoingCallControls()
         createIncomingCallControls()
     }
 
+    func createVideoViews() {
+        remoteVideoView = RTCEAGLVideoView()
+        remoteVideoView.delegate = self
+        localVideoView = RTCCameraPreviewView()
```
charlesmchen: Unfortunately, we currently ask for camera permissions when ringing starts.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -541,27 +630,87 @@ class CallViewController: UIViewController, CallObserver {
...
-            self.updateCallUI(callState: call.state)
+        AssertIsOnMainThread()
+        self.updateCallUI(callState: call.state)
+    }
+
+    // MARK: - Video
+
+    internal func updateLocalVideoTrack(localVideoTrack: RTCVideoTrack?) {
+        AssertIsOnMainThread()
+        if self.localVideoTrack == localVideoTrack {
```
charlesmchen: Done. I've been wondering why `guard` exists.  It just formalizes conventional early-exit practice.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -541,27 +630,87 @@ class CallViewController: UIViewController, CallObserver {
...
+        self.localVideoTrack = localVideoTrack
+
+        var source: RTCAVFoundationVideoSource?
+        if localVideoTrack?.source is RTCAVFoundationVideoSource {
+            source = localVideoTrack?.source as! RTCAVFoundationVideoSource
+        }
+        localVideoView.captureSession = source?.captureSession
+        let isHidden = source == nil
+        Logger.info("\(TAG) \(#function) isHidden: \(isHidden)")
+        localVideoView.isHidden = source == nil
```
charlesmchen: Done, thanks.

Signal/src/view controllers/CallViewController.swift
```diff
@@ -541,27 +630,87 @@ class CallViewController: UIViewController, CallObserver {
...
+
+    internal func updateRemoteVideoTrack(remoteVideoTrack: RTCVideoTrack?) {
+        AssertIsOnMainThread()
+        if self.remoteVideoTrack == remoteVideoTrack {
+            return
+        }
+
+        self.remoteVideoTrack?.remove(remoteVideoView)
+        self.remoteVideoTrack = nil
+        remoteVideoView.renderFrame(nil)
```
charlesmchen: Right, that's my best guess too, but the docs aren't very forthcoming. Anyway, I'm doing this because the demo app does and it certainly makes sense that we'd want to flush the contents of the remote video view.

charlesmchen:
PTAL @michaelkirk

Signal/src/call/SignalCall.swift
michaelkirk: > I've protected all of its state through synchronization.

I'm not sure I understand that claim. Accessing it from the main thread while it's mutated on the signalingQueue would still give an inconsistent state. In other words - we're not doing any locking on access. I think I understand your concern there, but I don't see how this addresses it.

My question, reworded: If all the methods you've added `objc_sync` to are already being run on the same serial queue, then what difference could adding `objc_sync` possibly make?

> More generally, I'd like to propose that we have an explicit understanding that all state should only be accessed from one thread.

Sounds like a good rule of thumb, but I'm not sure I'm comfortable committing to this 100%. For example, I'm not sure that meshes with YapDatabase usage conventions, but maybe I'm misunderstanding. In any case, I appreciate your insight, and look forward to your highlighting specific related examples in code reviews going forward.

Signal/src/call/SignalCall.swift
```diff
@@ -108,26 +157,49 @@ protocol CallObserver: class {
...
         observers = []
+
+        objc_sync_exit(self)
     }
 
     // MARK: Equatable
 
     static func == (lhs: SignalCall, rhs: SignalCall) -> Bool {
-        return lhs.localId == rhs.localId
+        objc_sync_enter(self)
```
charlesmchen: I've reverted this.

Signal/src/view controllers/CallViewController.swift
michaelkirk: Really? I would be less surprised if it were when peerConnection is initialized, but not until ringing starts seems arbitrary. What are we doing that fires the request?

Signal/src/call/SignalCall.swift
michaelkirk: Ah, as discussed offline, this has to do with synchronizing with the list of observers, which is mutated on the main thread.

greemlin:
Hi, Is there a way to adjust the video bandwidth? (bitrates, framerates etc) in order to reduce bandwidth/costs?

PR #1646, opened 2017-01-26
Charlesmchen/webrtc/video3

charlesmchen:
This PR builds on #1645.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -166,6 +166,15 @@ protocol CallServiceObserver: class {
             fireDidUpdateVideoTracks()
         }
     }
+    var isRemoteVideoEnabled = false {
+        didSet {
+            assertOnSignalingQueue()
+
+            Logger.info("\(self.TAG) \(#function)")
+
+            fireDidUpdateVideoTracks()
+        }
+    }
 
     required init(accountManager: AccountManager, contactsManager: OWSContactsManager, messageSender: MessageSender, notificationsAdapter: CallNotificationsAdapter) {
         self.accountManager = accountManager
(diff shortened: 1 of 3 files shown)


Signal/src/call/CallService.swift
```diff
@@ -834,6 +843,8 @@ protocol CallServiceObserver: class {
         } else if message.hasVideoStreamingStatus() {
             Logger.debug("\(TAG) remote participant sent VideoStreamingStatus via data channel")
 
+            self.isRemoteVideoEnabled = message.videoStreamingStatus.enabled()
+
             // TODO: translate from java
```
michaelkirk: TO-DONE!

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -433,7 +432,9 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
         Logger.debug("\(TAG) didOpen dataChannel:\(dataChannel)")
         CallService.signalingQueue.async {
             Logger.debug("\(self.TAG) set dataChannel")
+            assert(self.dataChannel == nil)
             self.dataChannel = dataChannel
+            dataChannel.delegate = self
```
michaelkirk: 😞 . Good catch!

Signal/src/call/CallService.swift
```diff
@@ -1026,9 +1038,26 @@ protocol CallServiceObserver: class {
         // It's only safe to access the class properties on the signaling queue, so
         // we dispatch there...
         CallService.signalingQueue.async {
-            Logger.info("\(self.TAG) \(#function): \(self.shouldHaveLocalVideoTrack())")
+            guard let call = self.call else {
+                return
```
michaelkirk: I'd say both of these guards deserve a log.

charlesmchen: It's expected for either `call` or `peerConnectionClient` to be nil at this point - for example, this method is called when call is set to nil.

Signal/src/call/CallService.swift
charlesmchen: Does that mean I should remove this commented-out code?

michaelkirk: Yeah, sorry for not being clear.

PR #1648, opened 2017-01-27
Make sure WebRTC preferences are synced *every* call

michaelkirk:
This slows the UI, but only for people who have locally opted into
WebRTC calls, and the alternative is that users are likely to have stale
settings the first time a pair of people opt-in.

// FREEBIE

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -51,6 +51,7 @@
 #import <JSQMessagesViewController/UIColor+JSQMessages.h>
 #import <JSQSystemSoundPlayer.h>
 #import <MobileCoreServices/UTCoreTypes.h>
+#import <SignalServiceKit/ContactsUpdater.h>
 #import <SignalServiceKit/MimeTypeUtil.h>
 #import <SignalServiceKit/OWSAttachmentsProcessor.h>
 #import <SignalServiceKit/OWSDisappearingMessagesConfiguration.h>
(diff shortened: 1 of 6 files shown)


michaelkirk:
PTAL @charlesmchen

Signal/src/view controllers/MessagesViewController.m
```diff
@@ -626,32 +627,62 @@ - (void)showFingerprintWithTheirIdentityKey:(NSData *)theirIdentityKey theirSign
...
+                (remoteWantsWebRTC ? @"YES" : @"NO"));
+            if (localWantsWebRTC && remoteWantsWebRTC) {
+                [self placeWebRTCCall];
+            } else {
+                [self placeRedphoneCall];
+            }
+        }
+        failure:^(NSError *_Nonnull error) {
+            DDLogWarn(@"%@ looking up call recipient: %@ failed with error: %@", self.tag, self.thread, error);
+            SignalAlertView(NSLocalizedString(@"UNABLE_TO_PLACE_CALL", @"Alert Title"), error.localizedDescription);
```
charlesmchen: I don't think we should ever surface error descriptions to the user. I suggest: "Could not reach service." (indicating rightfully that user should check their connectivity).

Signal/src/view controllers/MessagesViewController.m
```diff
@@ -626,32 +627,62 @@ - (void)showFingerprintWithTheirIdentityKey:(NSData *)theirIdentityKey theirSign
...
+        }];
+}
+
+- (void)placeRedphoneCall
+{
+    DDLogInfo(@"%@ Placing redphone call to: %@", self.tag, self.thread);
```
charlesmchen: Should we wrap this logging with a `#ifdef DEBUG`?  The log files are unencrypted.

Signal/src/view controllers/MessagesViewController.m
```diff
@@ -626,32 +627,62 @@ - (void)showFingerprintWithTheirIdentityKey:(NSData *)theirIdentityKey theirSign
...
+
+    OWSAssert(number != nil);
+    OWSAssert(contact != nil);
+
+    [Environment.phoneManager initiateOutgoingCallToContact:contact atRemoteNumber:number];
+}
+
+- (void)placeWebRTCCall
+{
+    DDLogInfo(@"%@ Placing WebRTC call to: %@", self.tag, self.thread);
```
charlesmchen: Ditto

Signal/src/view controllers/MessagesViewController.m
michaelkirk: We're not surfacing error descriptions, we're surfacing localizedErrorDescriptions. Arguably it should be localizedRecoverySuggestion, but we're not actively using it.

Network failures can happen for lots of reasons. e.g. offline, or the account were unregistered

In the case of disconnected network, the error reads: "Signal was unable to connect to the internet. Please try from another WiFi network or use mobile data."

Signal/src/view controllers/MessagesViewController.m
michaelkirk: I think a logging a call is valuable diagnostics. We rely on the scrubbing log formatter to remove phone numbers in non-debug.

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/fff061ff3fb664f6e4da06c391580df37ebd9978

PR #1652, opened 2017-01-27
Show/hide call view controls in remote video mode if user taps screen.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/views/OWSAnyTouchGestureRecognizer.h
+++ b/Signal/src/views/OWSAnyTouchGestureRecognizer.h
@@ -1,0 +1,16 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+// This custom GR can be used to detect touches when they
+// begin in a view.  In order to honor touch dispatch, this
+// GR will ignore touches that:
+//
+// * Are not single touches.
+// * Are not in the view for this GR.
+// * Are inside a visible, interaction-enabled subview.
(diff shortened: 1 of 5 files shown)


Signal/src/views/OWSAnyTouchGestureRecognizer.h
```diff
@@ -0,0 +1,16 @@
...
+#import <UIKit/UIKit.h>
+
+// This custom GR can be used to detect touches when they
+// begin in a view.  In order to honor touch dispatch, this
+// GR will ignore touches that:
+//
+// * Are not single touches.
+// * Are not in the view for this GR.
+// * Are inside a visible, interaction-enabled subview.
+@interface OWSAnyTouchGestureRecognizer : UIGestureRecognizer
```
charlesmchen: We use a custom GR in order to be more responsive, specifically to recognize as soon as the user taps the screen, not when a "tap" is completed.

charlesmchen: I'd like to use this kind of GR for certain other interactions in the UI as well.

michaelkirk: I like how responsive it is. 👍

charlesmchen: @michaelkirk  Is this work LGTM?

michaelkirk: Yup, LGTM here and #1656

PR #1656, opened 2017-01-30
Charlesmchen/webrtc/rework video controls

charlesmchen:
### Builds upon #1652

I wrongfully switched call view layouts based on "has remote video" state, not "has local video".

PTAL @michaelkirk

### DIFF
--- a/Signal/src/view controllers/CallViewController.swift
+++ b/Signal/src/view controllers/CallViewController.swift
@@ -458,6 +458,7 @@ class CallViewController: UIViewController, CallObserver, CallServiceObserver, R
         }
 
         self.localVideoConstraints = constraints
+        updateCallUI(callState: call.state)
     }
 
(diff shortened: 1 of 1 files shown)


Signal/src/view controllers/CallViewController.swift
```diff
@@ -73,6 +73,12 @@ class CallViewController: UIViewController, CallObserver, CallServiceObserver, R
     var remoteVideoConstraints: [NSLayoutConstraint] = []
     var localVideoConstraints: [NSLayoutConstraint] = []
 
+    var areRemoteVideoControlsHidden = false {
```
michaelkirk: nit so that you know I actually read it: maybe `shouldRemoteVideoControlsBeHidden`

charlesmchen: Done.

PR #1657, opened 2017-01-30
Charlesmchen/webrtc/video refinements 

charlesmchen:
* Prevent screen from dimming or device from locking during video call.
* Remove camera constraints.
* Improve thread safety in call ui adapter and adaptees.
   * I've moved the activity in these classes to the main thread.
* Show alert when user tries to activate local video without camera permissions.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -795,8 +795,9 @@ protocol CallServiceObserver: class {
             break
         }
 
-        // TODO: We could also check this whenever the app returns from the background,
-        //       since the user could disable this priviledge at all times.
+        // We don't need to worry about the user granting or remotinv this permission
+        // during a call while the app is in the background, because changing this
+        // permission kills the app.
         if authStatus != .authorized {
             DispatchQueue.main.async {
                 let title = NSLocalizedString("CAMERA_PERMISSION_MISSING_TITLE", comment: "Alert title when camera is not authorized")
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -1114,6 +1114,10 @@ protocol CallServiceObserver: class {
                 }
             }
         }
+        
+        // Prevent screen from dimming during video call.
+        let hasLocalOrRemoteVideo = localVideoTrack != nil || remoteVideoTrack != nil
+        UIApplication.shared.isIdleTimerDisabled = hasLocalOrRemoteVideo
     }
 }
 
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -779,6 +779,37 @@ protocol CallServiceObserver: class {
     func setHasLocalVideo(hasLocalVideo: Bool) {
         assertOnSignalingQueue()
 
+        let authStatus = AVCaptureDevice.authorizationStatus(forMediaType:AVMediaTypeVideo)
+        switch authStatus {
+        case .notDetermined:
+            Logger.debug("\(TAG) authStatus: AVAuthorizationStatusNotDetermined")
+            break
+        case .restricted:
+            Logger.debug("\(TAG) authStatus: AVAuthorizationStatusRestricted")
+            break
+        case .denied:
+            Logger.debug("\(TAG) authStatus: AVAuthorizationStatusDenied")
+            break
+        case .authorized:
+            Logger.debug("\(TAG) authStatus: AVAuthorizationStatusAuthorized")
+            break
+        }
+
+        // TODO: We could also check this whenever the app returns from the background,
+        //       since the user could disable this priviledge at all times.
+        if authStatus != .authorized {
+            DispatchQueue.main.async {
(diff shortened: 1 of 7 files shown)


Signal/src/call/CallService.swift
```diff
@@ -773,6 +779,38 @@ protocol CallServiceObserver: class {
...
+            break
+        }
+
+        // We don't need to worry about the user granting or remotinv this permission
+        // during a call while the app is in the background, because changing this
+        // permission kills the app.
+        if authStatus != .authorized {
+            DispatchQueue.main.async {
+                let title = NSLocalizedString("CAMERA_PERMISSION_MISSING_TITLE", comment: "Alert title when camera is not authorized")
+                let message = NSLocalizedString("CAMERA_PERMISSION_MISSING_BODY", comment: "Alert body when camera is not authorized")
```
charlesmchen: @michaelkirk The string extraction script currently has two warnings:
```
Warning: Key "MISSED_CALL" used with multiple comments "notification body" & "notification title"
Warning: Key "MSGVIEW_MISSED_CALL" used with multiple comments "notification body" & "notification title. Embeds {{Caller's Name}}"
```

Signal/src/call/CallService.swift
```diff
@@ -773,6 +779,38 @@ protocol CallServiceObserver: class {
...
+            break
+        case .denied:
+            Logger.debug("\(TAG) authStatus: AVAuthorizationStatusDenied")
+            break
+        case .authorized:
+            Logger.debug("\(TAG) authStatus: AVAuthorizationStatusAuthorized")
+            break
+        }
+
+        // We don't need to worry about the user granting or remotinv this permission
```
michaelkirk: s/remotinv/removing

Signal/src/call/CallService.swift
```diff
@@ -1114,6 +1156,10 @@ protocol CallServiceObserver: class {
                 }
             }
         }
+
+        // Prevent screen from dimming during video call.
+        let hasLocalOrRemoteVideo = localVideoTrack != nil || remoteVideoTrack != nil
+        UIApplication.shared.isIdleTimerDisabled = hasLocalOrRemoteVideo
```
michaelkirk: It's not obvious to me that we're re-enabling the idleTimer once the call ends.

Signal/src/call/CallService.swift
```diff
@@ -773,6 +779,38 @@ protocol CallServiceObserver: class {
...
+            Logger.debug("\(TAG) authStatus: AVAuthorizationStatusAuthorized")
+            break
+        }
+
+        // We don't need to worry about the user granting or remotinv this permission
+        // during a call while the app is in the background, because changing this
+        // permission kills the app.
+        if authStatus != .authorized {
+            DispatchQueue.main.async {
+                let title = NSLocalizedString("CAMERA_PERMISSION_MISSING_TITLE", comment: "Alert title when camera is not authorized")
```
michaelkirk: This is pretty similar to https://github.com/WhisperSystems/Signal-iOS/blob/master/Signal/src/util/UIViewController+CameraPermissions.m

Is there anything we could reuse or extract? That code has the added nicety of pointing the user to the settings app.

Signal/src/call/CallService.swift
michaelkirk: Should this be set on the main queue?

Signal/src/view controllers/CallViewController.swift
```diff
@@ -662,8 +662,6 @@ class CallViewController: UIViewController, CallObserver, CallServiceObserver, R
     func didPressVideo(sender: UIButton) {
         Logger.info("\(TAG) called \(#function)")
         let hasLocalVideo = !sender.isSelected
-        audioModeVideoButton.isSelected = hasLocalVideo
```
michaelkirk: Why remove this? I guess because it's redundant with updateCallUI? 
We're not handling the mute/speaker buttons in that way though.

Not a blocking concern at all, just wondering if you had a design goal here.

Signal/src/call/CallService.swift

Signal/src/call/CallService.swift
charlesmchen: Good catch. I've merged these strings.

Signal/src/call/CallService.swift
charlesmchen: I've added a comment.

charlesmchen: I wondered about that.  The documentation doesn't indicate that to that affect, but it _is_ part of UIKit and UIKit main-thread only.  I'm switching to be safe. Thanks!

Signal/src/view controllers/CallViewController.swift
charlesmchen: The view state should be updated by observing changes to the model state.  It is, but we also have the convention of optimistically updating the view in these kinds of handlers.  That's hopefully redundant. Here, it was worse because the CallService will "reject" requests from the video to start video if the camera permissions are missing.

PR #1658, opened 2017-01-30
Better concurrent call handling

michaelkirk:
This addresses a couple issues related to receiving calls when already on a call. 

So. previously, when you initiated a Signal Call to Possum, connected and started talking. Then received a non-Signal call from Squirrel, you'd see this:

squirrel

Which is weird, because you shouldn't be able to  "Accept" the new call without specifying what to do with the old call (e.g. "Hold" or "End" the existing call).

Turns out, we never marked the outgoing SignalCall as connected, so that's fixed in e425d35
With that, the interrupting call screen looks like this:

hold accept

Note the "hold & accept" option. We didn't intend on supporting hold initially, but I took a stab at it, just to see how difficult it would be in: https://github.com/WhisperSystems/Signal-iOS/commit/969b73cad8ed3fb542848e398a4620a9f1d40f9e

It worked, in that you could interrupt the Signal call with Possum by holding it, talk to Squirrel, then upon ending the call with Squirrel you'd be back talking to Possum, no problem.

BUT, before ending the interrupting call from Squirrel, if you *chose* to swap between held calls from the callkit screen, Signal call audio with Possum would not properly resume (unless you returned to the Signal app on your own volition, but that's not a thing we could reasonably expect users to discover.)

So I backed away form supporting Call muting for now, and went the route of fully disabling it for *outgoing* calls the same way we already were for *incoming* calls. Now the same call screen looks like this:

end accept

finé

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/SignalCall.swift
+++ b/Signal/src/call/SignalCall.swift
@@ -133,6 +133,8 @@ protocol CallObserver: class {
         }
     }
 
+    var isOnHold = false
+
     var connectedDate: NSDate?
 
     var error: CallError?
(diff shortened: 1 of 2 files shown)


Signal/src/call/SignalCall.swift
```diff
@@ -133,6 +133,8 @@ protocol CallObserver: class {
         }
     }
 
+    var isOnHold = false
```
charlesmchen: Remove this?

PR #1660, opened 2017-01-30
More space for non-callkit incoming call buttons

michaelkirk:
The "decline" and "accept" were too close together. Added a spacer, the
same size as an extra button between the two.

before:

image

after:

image

// FREEBIE

### DIFF
--- a/Signal/src/view controllers/CallViewController.swift
+++ b/Signal/src/view controllers/CallViewController.swift
@@ -255,7 +255,7 @@ class CallViewController: UIViewController, CallObserver, CallServiceObserver, R
         spacerView.autoSetDimension(.height, toSize:buttonSize())
 
         incomingCallView = createContainerForCallControls(controlGroups : [
-            [acceptIncomingButton, spacerView, declineIncomingButton ]
+            [acceptIncomingButton, declineIncomingButton ]
             ])
     }
 
@@ -322,7 +322,12 @@ class CallViewController: UIViewController, CallObserver, CallServiceObserver, R
                     spacer.autoPinEdge(.right, to:.left, of:subview)
                     spacer.setContentHuggingHorizontalLow()
                     spacer.autoVCenterInSuperview()
-                    if lastSpacer != nil {
+
+                    if subviews.count == 2 {
+                        // special case to hardcode the spacer's size when there is only 1 spacer.
+                        spacer.autoSetDimension(.width, toSize: ScaleFromIPhone5To7Plus(46, 60))
+                    } else if lastSpacer != nil {
+                        // else there is more than one spacer, and this isn't the first one.
                         spacer.autoMatch(.width, to:.width, of:lastSpacer!)
                     }
                     lastSpacer = spacer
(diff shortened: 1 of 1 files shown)


Signal/src/view controllers/CallViewController.swift
```diff
@@ -244,8 +244,12 @@ class CallViewController: UIViewController, CallObserver, CallServiceObserver, R
         declineIncomingButton = createButton(imageName:"hangup-active-wide",
                                              action:#selector(didPressDeclineCall))
 
+        let spacerView = UIView(frame: CGRect(x: 0, y: 0, width: buttonSize(), height: 0))
+        spacerView.autoSetDimension(.width, toSize:buttonSize())
+        spacerView.autoSetDimension(.height, toSize:buttonSize())
+
         incomingCallView = createContainerForCallControls(controlGroups : [
-            [acceptIncomingButton, declineIncomingButton ]
+            [acceptIncomingButton, spacerView, declineIncomingButton ]
```
charlesmchen: rowWithSubviews() Already creates spacers!  

It's busted because we currently only make sure the spacers all have the same width.  I suggest we replace:
```
spacer.autoMatch(.width, to:.width, of:lastSpacer!)
```
with something like:
```
// Only use spacing for two-button rows.
let spacerWidth = subviews.count == 2 ? ScaleFromIPhone5To7Plus(10, 20) : 0
spacer.autoSetDimension(.width, spacerWidth)
```

charlesmchen: I'd like to suggest that they are too far apart in the screenshots above. They should both be easily accessible with the thumb, even on large form factors like the 7+.

michaelkirk:
PTAL @charlesmchen 

New spacing looks like this...

## 5

image

## 6

iphone 6 noncallkit ringer

## 7+

iphone 7 noncallkit ringer

Signal/src/view controllers/CallViewController.swift
```diff
@@ -250,6 +250,10 @@ class CallViewController: UIViewController, CallObserver, CallServiceObserver, R
         declineIncomingButton = createButton(imageName:"hangup-active-wide",
                                              action:#selector(didPressDeclineCall))
 
+        let spacerView = UIView(frame: CGRect(x: 0, y: 0, width: buttonSize(), height: 0))
+        spacerView.autoSetDimension(.width, toSize:buttonSize())
+        spacerView.autoSetDimension(.height, toSize:buttonSize())
```
charlesmchen: This code can be removed, right?

michaelkirk: 😞

Signal/src/view controllers/CallViewController.swift
```diff
@@ -318,7 +322,12 @@ class CallViewController: UIViewController, CallObserver, CallServiceObserver, R
                     spacer.autoPinEdge(.right, to:.left, of:subview)
                     spacer.setContentHuggingHorizontalLow()
                     spacer.autoVCenterInSuperview()
-                    if lastSpacer != nil {
+
+                    if subviews.count == 2 {
```
charlesmchen: This won't handle the case where there's more than two buttons in a row. You could modify 

```if subviews.count == 2 ```

to read:

```if row.subviews.count == 2 ```

On the other hand, it might just be better to assert in the `else...if` clause since this would be would too much space for a 3 or 4 button row.

michaelkirk:
LGTM. LGTY?

charlesmchen:
LGTM thanks for your patience

PR #1661, opened 2017-01-30
Add drop shadows to text in call view.

charlesmchen:
PTAL @michaelkirk


Signal/src/view controllers/CallViewController.swift
```diff
@@ -183,11 +183,17 @@ class CallViewController: UIViewController, CallObserver, CallServiceObserver, R
         contactNameLabel = UILabel()
         contactNameLabel.font = UIFont.ows_lightFont(withSize:ScaleFromIPhone5To7Plus(32, 40))
         contactNameLabel.textColor = UIColor.white
+        contactNameLabel.layer.shadowOffset = CGSize(width: 0, height: 0)
```
michaelkirk: also CGSize.zero

charlesmchen: Done!

charlesmchen:
> I was also having a hard time reading the navbar (time, signal strength). Did you experience that?

I plan on doing something around that once I can repro the issue you saw.

PR #1666, opened 2017-01-31
Specify AVAudioSession modes for calls.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallAudioService.swift
+++ b/Signal/src/call/CallAudioService.swift
@@ -54,8 +54,14 @@ import Foundation
 
     private func ensureIsEnabled(call: SignalCall) {
         // Auto-enable speakerphone when local video is enabled.
-        if call.isSpeakerphoneEnabled || call.hasLocalVideo {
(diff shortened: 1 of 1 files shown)


Signal/src/call/CallAudioService.swift
```diff
@@ -54,8 +54,14 @@ import Foundation
 
     private func ensureIsEnabled(call: SignalCall) {
         // Auto-enable speakerphone when local video is enabled.
-        if call.isSpeakerphoneEnabled || call.hasLocalVideo {
-            setAudioSession(category: AVAudioSessionCategoryPlayAndRecord, options: .defaultToSpeaker)
+        if call.hasLocalVideo {
+            setAudioSession(category: AVAudioSessionCategoryPlayAndRecord,
+                            mode: AVAudioSessionModeVideoChat,
```
charlesmchen: The docs are very vague about the effects of AVAudioSessionModeVideoChat and AVAudioSessionModeVoiceChat. It's not just which mics that are affected but also the audio processing.

michaelkirk:
LGTM

PR #1670, opened 2017-01-31
[SSK] fix crash when messaging newly unregistered user

michaelkirk:
PTAL @charlesmchen.

Once the corresponding [SSK PR](https://github.com/WhisperSystems/SignalServiceKit/pull/103) is merged, I'll update the branch in the podfile before merging this.


michaelkirk:
In feb5a9e

PR #1673, opened 2017-01-31
Ensure audio state when call connects.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallAudioService.swift
+++ b/Signal/src/call/CallAudioService.swift
@@ -32,7 +32,7 @@ import Foundation
 
     internal func stateDidChange(call: SignalCall, state: CallState) {
         AssertIsOnMainThread()
-        self.handleState(state)
+        self.handleState(call:call)
     }
 
     internal func muteDidChange(call: SignalCall, isMuted: Bool) {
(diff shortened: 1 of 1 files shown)


Signal/src/call/CallAudioService.swift
```diff
@@ -32,7 +32,7 @@ import Foundation
 
     internal func stateDidChange(call: SignalCall, state: CallState) {
         AssertIsOnMainThread()
-        self.handleState(state)
+        self.handleState(call:call)
```
michaelkirk: Why this change?

Previously we were passing an immutable state to this method. But now we're passing something for which the state could change (on a different thread).

(also nit: space after `:` per linter)

Signal/src/call/CallAudioService.swift
```diff
@@ -104,12 +104,12 @@ import Foundation
...
     }
 
-    private func handleConnected() {
+    private func handleConnected(call: SignalCall) {
         Logger.debug("\(TAG) \(#function)")
         stopRinging()
 
         // disable start recording to transmit call audio.
-        setAudioSession(category: AVAudioSessionCategoryPlayAndRecord)
+        ensureIsEnabled(call: call)
```
michaelkirk: This fix makes sense. :+1:

Signal/src/call/CallAudioService.swift
michaelkirk: Oh. I see why.

PR #1675, opened 2017-01-31
Log reconnect events as such.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
Ask and ye shall receive.

PR #1676, opened 2017-01-31
Charlesmchen/webrtc/thread safety

charlesmchen:
* Everything expensive about the signaling process is done inside PeerConnectionClient.  Therefore, I've made the signaling queue a private implementation detail of that class.  
* `CallService` and `SignalCall` are now main-thread-only.
* All of the code that interacts with `CallService` and `SignalCall` can now do so more safely and simply.

I'll annotate this PR in-line.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -110,7 +110,8 @@ protocol CallServiceObserver: class {
     static let fallbackIceServer = RTCIceServer(urlStrings: ["stun:stun1.l.google.com:19302"])
 
     // Synchronize call signaling on the callSignalingQueue to make sure any appropriate requisite state is set.
-    static let signalingQueue = DispatchQueue(label: "CallServiceSignalingQueue")
+//    static let signalingQueue = DispatchQueue(label: "CallServiceSignalingQueue")
+    static let signalingQueue = DispatchQueue.main
 
     // MARK: Ivars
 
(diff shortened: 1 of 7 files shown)


Signal/src/call/CallService.swift
```diff
@@ -109,33 +108,28 @@ protocol CallServiceObserver: class {
...
         didSet {
-            assertOnSignalingQueue()
+            AssertIsOnMainThread()
 
             oldValue?.removeObserver(self)
             call?.addObserverAndSyncState(observer: self)
 
-            DispatchQueue.main.async { [weak self] in
-                self?.updateIsVideoEnabled()
-            }
```
charlesmchen: Most of this PR consists of this kind of simplification.  Most everything moves to the main thread.

Signal/src/call/CallService.swift
```diff
@@ -260,22 +252,24 @@ protocol CallServiceObserver: class {
 
             self.peerConnectionClient = peerConnectionClient
 
-            return self.peerConnectionClient!.createOffer()
-        }.then(on: CallService.signalingQueue) { (sessionDescription: HardenedRTCSessionDescription) -> Promise<Void> in
```
charlesmchen: A lot of whitespace was fixed. If you want, I'll go back and revert this. I know you're not a fan.

charlesmchen:
I'm going to close this PR so that I can clean it up some.

PR #1678, opened 2017-02-01
Charlesmchen/webrtc/thread safety 

charlesmchen:
### SSK PR: https://github.com/WhisperSystems/SignalServiceKit/pull/104

* Everything expensive about the signaling process is done inside PeerConnectionClient.  Therefore, I've made the signaling queue a private implementation detail of that class.  
* `CallService` and `SignalCall` are now main-thread-only.
* All of the code that interacts with `CallService` and `SignalCall` can now do so more safely and simply.

I'll annotate this PR in-line.

PTAL @michaelkirk


Signal/src/call/CallService.swift
```diff
@@ -109,24 +108,19 @@ protocol CallServiceObserver: class {
...
         didSet {
-            assertOnSignalingQueue()
+            AssertIsOnMainThread()
 
             oldValue?.removeObserver(self)
             call?.addObserverAndSyncState(observer: self)
 
-            DispatchQueue.main.async { [weak self] in
-                self?.updateIsVideoEnabled()
-            }
```
charlesmchen: Most of this PR consists of this kind of simplification. Most everything moves to the main thread.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -70,12 +70,24 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
              dataChannelSignaling = "signaling"
     }
 
+    // A state in this class should only be accessed on this queue in order to
+    // serialize access.
+    //
+    // This queue is also used to perform expensive calls to the WebRTC API.
+    private static let signalingQueue = DispatchQueue(label: "CallServiceSignalingQueue")
```
charlesmchen: * `signalingQueue` is now a private implementation detail of this class.  
* `PeerConnectionClient`'s properties are now private and `PeerConnectionClient` takes sole responsibility for ensuring that it's state is only accessed on `signalingQueue`.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -70,12 +70,24 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
+
     // Delegate is notified of key events in the call lifecycle.
-    public weak var delegate: PeerConnectionClientDelegate!
+    private weak var delegate: PeerConnectionClientDelegate!
+
+    func setDelegate(delegate: PeerConnectionClientDelegate?) {
+        PeerConnectionClient.signalingQueue.async {
+            self.delegate = delegate
+        }
```
charlesmchen: Nearly all of the signalling-related concurrency is now async, reducing the surface area for deadlock.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -128,17 +142,23 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
     // MARK: - Media Streams
 
     public func createSignalingDataChannel() {
-        let dataChannel = peerConnection.dataChannel(forLabel: Identifiers.dataChannelSignaling.rawValue,
-                                                     configuration: RTCDataChannelConfiguration())
-        dataChannel.delegate = self
+        AssertIsOnMainThread()
+        
+        PeerConnectionClient.signalingQueue.sync {
```
charlesmchen: This is the only exception: the main thread blocks on the `signalingQueue` when creating the data channel, to facilitate testing. Since `signalingQueue` never blocks on the main thread, there is no risk of dead lock.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -205,111 +233,161 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-                    reject(error)
-                    return
-                }
+    public func createOffer() -> Promise<HardenedRTCSessionDescription> {
+        AssertIsOnMainThread()
 
-                fulfill(HardenedRTCSessionDescription(rtcSessionDescription: sessionDescription))
-            })
+        return Promise { fulfill, reject in
+            AssertIsOnMainThread()
```
charlesmchen: Promise-related code can be hard to read; the concurrency in particular isn't always self-evident.  I've been prolific with asserts to make this more clear.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -326,115 +404,145 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     // MARK: - Data Channel
 
-    func sendDataChannelMessage(data: Data) -> Bool {
-        guard let dataChannel = self.dataChannel else {
-            Logger.error("\(TAG) in \(#function) ignoring sending \(data) for nil dataChannel")
-            return false
+    public func sendDataChannelMessage(data: Data) -> Promise<Bool> {
```
charlesmchen: The only API change of significance is that sendDataChannelMessage() now returns a promise.

charlesmchen: The result of `sendDataChannelMessage()` is only used for logging; it might be simpler to just pull that logging into `sendDataChannelMessage()`.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -326,115 +404,145 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
 
     /** Called when media is received on a new stream from remote peer. */
-    public func peerConnection(_ peerConnection: RTCPeerConnection, didAdd stream: RTCMediaStream) {
-        Logger.debug("\(TAG) didAdd stream:\(stream) video tracks: \(stream.videoTracks.count) audio tracks: \(stream.audioTracks.count)")
-
-        if stream.videoTracks.count > 0 {
-            remoteVideoTrack = stream.videoTracks[0]
-            if let delegate = delegate {
-                delegate.peerConnectionClient(self, didUpdateRemote: remoteVideoTrack)
+    internal func peerConnection(_ peerConnection: RTCPeerConnection, didAdd stream: RTCMediaStream) {
```
charlesmchen: Some of these `RTCPeerConnection` delegate callbacks were actually being invoked on a WebRTC thread. That could have caused some hard-to-reproduce bugs.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -326,115 +404,145 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-        case .connected, .completed:
-            if let delegate = delegate {
-                delegate.peerConnectionClientIceConnected(self)
-            }
-        case .failed:
-            Logger.warn("\(self.TAG) RTCIceConnection failed.")
-            if let delegate = delegate {
-                delegate.peerConnectionClientIceFailed(self)
+    internal func peerConnection(_ peerConnection: RTCPeerConnection, didChange newState: RTCIceConnectionState) {
+        PeerConnectionClient.signalingQueue.async {
```
charlesmchen: Ditto.

Signal/src/call/SignalCall.swift
```diff
@@ -45,27 +45,17 @@ protocol CallObserver: class {
...
+            for observer in observers {
+                observer.value?.hasLocalVideoDidChange(call: self, hasLocalVideo: hasLocalVideo)
             }
         }
     }
 
     var state: CallState {
         didSet {
-            // This should only occur on the signaling queue.
-            objc_sync_enter(self)
```
charlesmchen: The `objc_sync_...` stuff can now go away. RIP.

Signal/src/call/PeerConnectionClient.swift
charlesmchen: Okay, I've done this.

Signal/src/call/NonCallKitCallUIAdaptee.swift
```diff
@@ -27,12 +27,10 @@ class NonCallKitCallUIAdaptee: CallUIAdaptee {
 
         let call = SignalCall.outgoingCall(localId: UUID(), remotePhoneNumber: handle)
 
-        CallService.signalingQueue.async {
-            _ = self.callService.handleOutgoingCall(call).then {
-                Logger.debug("\(self.TAG) handleOutgoingCall succeeded")
+        self.callService.handleOutgoingCall(call).then {
```
michaelkirk: probably gettting a warning here without explicitly ignoring the returned value, right?

i.e. put back the `_ = `

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -70,12 +72,24 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
+    func setDelegate(delegate: PeerConnectionClientDelegate?) {
+        PeerConnectionClient.signalingQueue.async {
+            self.delegate = delegate
+        }
+    }
 
     // Connection
 
-    internal var peerConnection: RTCPeerConnection!
+    private var peerConnection: RTCPeerConnection!
```
michaelkirk: IIRC this was internal because it's being used by a test. I should have left comment.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -128,17 +144,23 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
     // MARK: - Media Streams
 
     public func createSignalingDataChannel() {
-        let dataChannel = peerConnection.dataChannel(forLabel: Identifiers.dataChannelSignaling.rawValue,
-                                                     configuration: RTCDataChannelConfiguration())
-        dataChannel.delegate = self
+        AssertIsOnMainThread()
+
+        PeerConnectionClient.signalingQueue.sync {
```
michaelkirk: Should we make this return a promise and refactor the callsite so the signaling queue stuff can run async?

Signal/src/call/WebRTCCallMessageHandler.swift
```diff
@@ -30,7 +31,7 @@ class WebRTCCallMessageHandler: NSObject, OWSCallMessageHandler {
         Logger.verbose("\(TAG) handling offer from caller:\(callerId)")
 
         let thread = TSContactThread.getOrCreateThread(contactId: callerId)
-        CallService.signalingQueue.async {
+        DispatchQueue.main.async {
```
michaelkirk: I would have guessed that this is already main thread. Can we get rid of the dispatch and just assertMainQueue?

Signal/src/call/UserInterface/CallUIAdapter.swift
```diff
@@ -30,24 +30,29 @@ protocol CallUIAdaptee {
...
         NotificationCenter.default.post(name: NSNotification.Name(rawValue: callNotificationName), object: call)
     }
 
     internal func reportMissedCall(_ call: SignalCall, callerName: String) {
+        AssertIsOnMainThread()
+
         notificationsAdapter.presentMissedCall(call, callerName: callerName)
     }
 
+    // TODO: who calls this?
```
michaelkirk: It's used by a UILocalNotification action. I'd say don't look into it much at this point as I'm making changes to it in the NSUserActivity feature I'm working on. I can double check that notification actions are working as part of that.

Signal/src/call/Speakerbox/CallKitCallUIAdaptee.swift
```diff
@@ -268,15 +262,15 @@ final class CallKitCallUIAdaptee: NSObject, CallUIAdaptee, CXProviderDelegate {
...
 
         // Synchronous to ensure call is terminated before call is displayed as "ended"
-        CallService.signalingQueue.sync {
-            self.callService.handleLocalHungupCall(call)
-        }
+        self.callService.handleLocalHungupCall(call)
 
-        // Signal to the system that the action has been successfully performed.
-        action.fulfill()
+        DispatchQueue.main.async {
```
michaelkirk: presumably we're already on the main thread?

Signal/src/call/Speakerbox/CallKitCallUIAdaptee.swift
```diff
@@ -268,15 +262,15 @@ final class CallKitCallUIAdaptee: NSObject, CallUIAdaptee, CXProviderDelegate {
         //        call.endSpeakerboxCall()
 
         // Synchronous to ensure call is terminated before call is displayed as "ended"
```
michaelkirk: This comment no longer makes sense.

Signal/src/call/SignalCall.swift
michaelkirk: 🔥 🔥 🔥 🔥

Signal/src/call/PeerConnectionClient.swift
michaelkirk: 👍

Signal/src/call/CallService.swift
```diff
@@ -250,8 +241,9 @@ protocol CallServiceObserver: class {
             return Promise(error: CallError.assertionError(description: errorDescription))
         }
 
-        return getIceServers().then(on: CallService.signalingQueue) { iceServers -> Promise<HardenedRTCSessionDescription> in
+        return getIceServers().then(on: DispatchQueue.main) { iceServers -> Promise<HardenedRTCSessionDescription> in
```

Signal/src/call/NonCallKitCallUIAdaptee.swift
charlesmchen: Nope, I'm not seeing any such warning?  That's really a possible warning in swift?  It seems like there would be mostly false positives for such a check.

Signal/src/call/PeerConnectionClient.swift
charlesmchen: Ah, and I've broken the tests.  Fixed.  I've added test-specific accessors/utility methods to avoid having production code use these pathways.

Signal/src/call/NonCallKitCallUIAdaptee.swift
michaelkirk: Yes that's really a possible warning in swift.

Signal/src/call/PeerConnectionClient.swift
charlesmchen: Perhaps PeerConnectionClient's constructor should just call createSignalingDataChannel().
We always do it immediately after instantiating the PCC.  Anyway, no need for a promise, right?

Signal/src/call/WebRTCCallMessageHandler.swift
charlesmchen: Done.

Signal/src/call/UserInterface/CallUIAdapter.swift
charlesmchen: K, thanks. The assert should be useful when you test.

Signal/src/call/Speakerbox/CallKitCallUIAdaptee.swift
charlesmchen: Removed.

Signal/src/call/Speakerbox/CallKitCallUIAdaptee.swift
charlesmchen: Good catch, removed.

PR #1679, opened 2017-02-01
Add button to start a new conversation with non-contact based on phon…

charlesmchen:
…e number in search field.

With great effort I resisted the desire to go further than the narrowest possible fix.

PTAL @michaelkirk 

img_4279
img_4280


charlesmchen:
I'm tempted to replace the "search by name or number", since if a user enters a "non-contact phone number" they're not _searching_.  But I can't come up with better copy. "enter a name or number" is worse imo.

img_4277

JohanDoe:
Could a solution to your "search by name or number" issue be to implement this iMessage style?

So it basically says "To:" and you just start typing. No match means non contact number. 

I personally like it. ;-)

Signal/src/view controllers/MessageComposeTableViewController.m
```diff
@@ -296,11 +302,24 @@ - (void)filterContentForSearchText:(NSString *)searchText
...
+        self.showNewConversationForNonContactButton = YES;
+        self.nonContactPhoneNumber = formattedNumber;
     } else {
         self.sendTextButton.hidden = YES;
+        self.showNewConversationForNonContactButton = NO;
+        _nonContactPhoneNumber = nil;
     }
 }
 
+- (void)setShowNewConversationForNonContactButton:(BOOL)showNewConversationForNonContactButton {
```
michaelkirk: nit: clang-format

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -481,6 +481,9 @@
 /* No comment provided by engineer. */
 "NETWORK_STATUS_TEXT" = "You can check your network status by looking at the colored bar above your inbox.";
 
+/* Text for button to start a new conversation with a non-contact */
+"NEW_CONVERSATION_FOR_NON_CONTACT_FORMAT" = "New conversation with %@";
```
michaelkirk: So we now have two cells that could appear when typing a phone number into the new message contact picker... 

1. Send SMS to +12312312312
2. New conversation with: +12312312312

The distinction of the two seems not very obvious. I think better copy might help a little. Something like:

1. Invite +12312312312 to Signal via SMS
2. Start a Signal conversation with: +12312312312

I think that's an easy change that improves the situation, the copy might be too long as is, but I think including "Invite" in the SMS option is key.

...

I also want to talk through the UX when typing in the wrong number, or when the user mistakenly expects to be able to contact someone who is not a Signal user. Alice may not know (or care!) that Bob is not on Signal. She reasonably assumes that she'll be able to contact whichever number she types in.

So the flow in this PR is currently:

0. Alice types in bobs number, hits "new conversation"
1. Taken to a new Signal thread with number +1231231231
2. When Alice sends a message, she'll see the "Unregistered User" error.

On Android, I believe the "Unregistered User" error would instead just fall back to sending an SMS. But we have it a little harder here.

One thought would be to make the "Unregistered User" error an actionable alert that would lead you to the invite via SMS flow. If we did that, I'd consider removing the "Invite +12312312312 via SMS" from the contact picker all together.  As a side note, invitation is *also* discoverable via the "Invite Friends to Signal" entry in the contacts picker.

I realize that this is, in some sense, and existing UX bug, but given that we're letting people more easily start conversations with potentially non-signal users I think we need to re-evaluate.

@charlesmchen thoughts? Good Idea? Bad Idea? Do it later?

charlesmchen: > I also want to talk through the UX when typing in the wrong number,

Yes, that's just one problem. Another problematic scenario is where the current user has no signal contacts.  This whole UI changes in that case, and this fix doesn't affect that UI.  There's a ton of stuff I'd like to change here, and I'm trying to constrain this PR.

charlesmchen:
@JohanDoe

> Could a solution to your "search by name or number" issue be to implement this iMessage style?

Unlike iMessage, Signal currently draws a strong distinction between 1:1 conversations and group conversations. So the iMessage approach, while more elegant isn't currently possible in Signal.

Signal/src/view controllers/MessageComposeTableViewController.m
charlesmchen: I don't get it - what's misformatted?

michaelkirk: bracket should be on it's own line

PR #1680, opened 2017-02-01
Avoid crashes when closing peer connection client.

charlesmchen:
This work prevents crashes due to a number of edge cases, mostly concurrency-related.

* Adds asserts that `CallService. peerConnectionClient` is nil when we create a new `PeerConnectionClient`.  This isn't always true.  These asserts should help us identify a repro.
* `PeerConnectionClient` does a lot of dispatching to the signaling queue and the main thread. I've  used weak references and added checks in these places around termination.
* The exact order of operations during termination matters, as noted by the existing comment in `terminateInternal()`.  I've slightly reworked this logic.

It's quite easy to repro crashes before this branch that are not reproducible after this branch.  Just start & stop many calls to another device without even accepting the call on the second device.  It matters how quickly you hang up the call while its ringing - you can hang up immediately, right before the "ringing" state, right after the "ringing" state, or later.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -250,6 +250,7 @@ protocol CallServiceObserver: class {
             // to do this explicitly.
             peerConnectionClient.createSignalingDataChannel()
 
+            assert(self.peerConnectionClient == nil, "Unexpected PeerConnectionClient instance")
             self.peerConnectionClient = peerConnectionClient
 
             return self.peerConnectionClient!.createOffer()
@@ -395,6 +396,7 @@ protocol CallServiceObserver: class {
         }.then() { (iceServers: [RTCIceServer]) -> Promise<HardenedRTCSessionDescription> in
             // FIXME for first time call recipients I think we'll see mic/camera permission requests here,
             // even though, from the users perspective, no incoming call is yet visible.
+            assert(self.peerConnectionClient == nil, "Unexpected PeerConnectionClient instance")
             self.peerConnectionClient = PeerConnectionClient(iceServers: iceServers, delegate: self)
 
             let offerSessionDescription = RTCSessionDescription(type: .offer, sdp: callerSessionDescription)
@@ -994,7 +996,6 @@ protocol CallServiceObserver: class {
         Logger.debug("\(TAG) in \(#function)")
 
         PeerConnectionClient.stopAudioSession()
-        peerConnectionClient?.setDelegate(delegate:nil)
         peerConnectionClient?.terminate()
 
         peerConnectionClient = nil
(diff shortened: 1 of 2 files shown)


charlesmchen:
It's worth noting that this doesn't fix all of the known issues - but it adds asserts that make them much easier to observe when they do happen.

Signal/src/call/CallService.swift
```diff
@@ -994,7 +996,6 @@ protocol CallServiceObserver: class {
         Logger.debug("\(TAG) in \(#function)")
 
         PeerConnectionClient.stopAudioSession()
-        peerConnectionClient?.setDelegate(delegate:nil)
```
michaelkirk: Why would we want CallService to receive delegate methods after termination?

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -392,16 +433,22 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-        audioTrack = nil
-        localVideoTrack = nil
-        remoteVideoTrack = nil
+
+        localVideoTrack?.isEnabled = false
+        remoteVideoTrack?.isEnabled = false
+
+        peerConnection.delegate = nil
+        peerConnection.close()
+        peerConnection = nil
```
michaelkirk: A lot of the crashes I saw while developing I interpreted as "tried to free (e.g.) audiotrack while tearing down peer connection but unexpectedly found something still had a reference to it." 

I was able to avoid these types of crashes by making sure peerConnection client held no references to these objects *before* rtcPeerConnection was dealloc'd

I would have expected to see fewer crashes if you first nil'd the references to all of the resources "owned" by RTCPeerConnection, e.g. dataChannel, audioSender, AudioTrack, videoSender, etc. **then** nilled peerConnection. If this is working, then fine. But something to keep in mind if we see more crashes here.

Signal/src/call/CallService.swift
charlesmchen: Clearing of the delegate now occurs inside the next statement: `peerConnectionClient?.terminate()`. Before PCC clears its delegate, it invokes a few of the delegate method to ensure that the views & `CallService` release any references they have to the audio and video tracks.

Signal/src/call/PeerConnectionClient.swift
charlesmchen: Yes, that's reasonable and that reordering of operations isn't the important part of these changes. I've moved the clearing of the tracks & senders above the clearing of the peerConnection.  But it doesn't seem to make things any safer or less safe.  It's the other changes in this PR that seem to matter in my testing.

charlesmchen:
There are still some issues (tracked in trello cards) but things are a _lot_ more stable now.

PR #1681, opened 2017-02-02
CallKit call from contacts

michaelkirk:
e.g. from the system contacts app

image

Or the "Callback" action from a failed call.

If you try to call a non-signal contact from the contacts app you just get a message "Contact is not a Signal User". 

Ideally we'd instead shuttle people towards an invite flow, but that might be for another day.


Signal/src/AppDelegate.m
```diff
@@ -333,10 +334,39 @@ - (void)application:(UIApplication *)application
 - (BOOL)application:(UIApplication *)application continueUserActivity:(nonnull NSUserActivity *)userActivity restorationHandler:(nonnull void (^)(NSArray * _Nullable))restorationHandler
 {
     if ([userActivity.activityType isEqualToString:@"INStartVideoCallIntent"]) {
+        DDLogInfo(@"%@ got start video call intent", self.tag);
```
charlesmchen: You know, when I added this `INStartVideoCallIntent` logic, I was trying to handle the case where the user taps the "video" button in the CallKit UI.  I find it surprising that If the user taps the "signal" button in the contacts app (which has a "video" icon), the Signal App receives a `INStartAudioCallIntent` activity.

Signal/src/AppDelegate.m
```diff
@@ -333,10 +334,39 @@ - (void)application:(UIApplication *)application
...
+
+        INInteraction *interaction = [userActivity interaction];
+        INIntent *intent = interaction.intent;
+
+        if (![intent isKindOfClass:[INStartAudioCallIntent class]]) {
+            DDLogError(@"%@ unexpected class for start call audio: %@", self.tag, intent);
+            return NO;
+        }
+        INStartAudioCallIntent *startCallIntent = (INStartAudioCallIntent *)intent;
+        NSString *_Nullable handle = startCallIntent.contacts.firstObject.personHandle.value;
```
charlesmchen: It might be nice to add a comment here documenting what type of value `handle` is and how its formatted.

Signal/src/call/OutboundCallInitiator.swift
```diff
@@ -0,0 +1,100 @@
...
+        self.contactsUpdater = contactsUpdater
+    }
+
+    /**
+     * |handle| is a user formatted phone number, e.g. from a system contacts entry
+     */
+    public func initiateCall(handle: String) -> Bool {
+        Logger.info("\(TAG) in \(#function) with handle: \(handle)")
+        guard let recipientId = PhoneNumber(fromUserSpecifiedText: handle).toE164() else {
+            Logger.warn("\(TAG) unable to parse signalId from phone number: \(handle)")
```
charlesmchen: If a contact has multiple phone numbers (work cell, home cell, fax, etc.) which one(s) does signal receive?  If more than one, shouldn't we try to parse/lookup all of them and use the first one that corresponds to a signal user?

Signal/src/AppDelegate.m
charlesmchen: Aha https://trello.com/c/0frpAkvR

michaelkirk: Actually there are two buttons in the contact app. One for signal audio, or one for signal video. The audio sends the INStartAudioCallIntent, the video sends INStartVideoCallIntent.

Somewhat confusingly, as you've observed, the INStartVideoCallIntent you receive when initiating a video call from contacts is the same intent the app receives when "upgrading" an existing audio call to video.

charlesmchen:
Per offline conversation, feel free to merge this and address the comments in a follow-up PR.

michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/e272684ea0a15d0c7a48478184f5ff4c7221856b

PR #1683, opened 2017-02-02
Can receive calls while in settings

michaelkirk:
The existing presented settings (or fingerprint verification) modals
were preventing the display of the incoming call modal.

// FREEBIE


michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/d8df4b9e358c0cb0cad3dea1c35cb627151285cc

PR #1685, opened 2017-02-02
CallKit opt out

michaelkirk:
based on #1683 

Allows opting out of call kit on iOS10.

### DIFF
--- a/Signal/src/call/UserInterface/CallUIAdapter.swift
+++ b/Signal/src/call/UserInterface/CallUIAdapter.swift
@@ -76,11 +76,11 @@ extension CallUIAdaptee {
             // So we use the non-CallKit call UI.
             Logger.info("\(TAG) choosing non-callkit adaptee for simulator.")
             adaptee = NonCallKitCallUIAdaptee(callService: callService, notificationsAdapter: notificationsAdapter)
-        } else if #available(iOS 10.0, *) {
+        } else if #available(iOS 10.0, *), Environment.getCurrent().preferences.isCallKitEnabled() {
             Logger.info("\(TAG) choosing callkit adaptee for iOS10+")
             adaptee = CallKitCallUIAdaptee(callService: callService, notificationsAdapter: notificationsAdapter)
         } else {
-            Logger.info("\(TAG) choosing non-callkit adaptee for older iOS")
+            Logger.info("\(TAG) choosing non-callkit adaptee")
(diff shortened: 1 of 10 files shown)


Signal/src/call/UserInterface/CallUIAdapter.swift
```diff
@@ -76,11 +76,11 @@ extension CallUIAdaptee {
             // So we use the non-CallKit call UI.
             Logger.info("\(TAG) choosing non-callkit adaptee for simulator.")
             adaptee = NonCallKitCallUIAdaptee(callService: callService, notificationsAdapter: notificationsAdapter)
-        } else if #available(iOS 10.0, *) {
+        } else if #available(iOS 10.0, *), Environment.getCurrent().preferences.isCallKitEnabled() {
```
charlesmchen: Am I misreading the swift syntax?  Why is there a comma and not a &&

Signal/src/view controllers/AdvancedSettingsTableViewController.m
```diff
@@ -101,26 +120,61 @@ - (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger
...
+    AdvancedSettingsTableViewControllerSection settingsSection = (AdvancedSettingsTableViewControllerSection)section;
+    switch (settingsSection) {
+        case AdvancedSettingsTableViewControllerSectionLogging:
+            return NSLocalizedString(@"LOGGING_SECTION", nil);
+        case AdvancedSettingsTableViewControllerSectionCalling:
+            return NSLocalizedString(@"SETTINGS_SECTION_TITLE_CALLING", @"settings topic header for table section");
+        case AdvancedSettingsTableViewControllerSectionPushNotifications:
+            return NSLocalizedString(@"PUSH_REGISTER_TITLE", @"Used in table section header and alert view title contexts");
+        default:
+            return 0;
```
michaelkirk: return nil, and log?

Signal/src/call/UserInterface/CallUIAdapter.swift
michaelkirk: welcome to swift. =/

michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/2a9aa4c852584120c762a3bf32166b77eabc2b27

PR #1686, opened 2017-02-02
WebRTC and CallKit changes

michaelkirk:
This is a mondo-pull request, including WebRTC audio/video calling and CallKit integration.

These changes were reviewed individually as they made their way into the development branch, but with this PR I wanted to publicly acknowledge their inclusion into master.


Signal/src/environment/NotificationsManager.m
```diff
@@ -55,20 +63,94 @@ - (void)notifyUserForCall:(TSCall *)call inThread:(TSThread *)thread {
...
-            if ([[Environment preferences] notificationPreviewType] == NotificationNoNameNoPreview) {
-                notification.alertBody = [NSString stringWithFormat:NSLocalizedString(@"MISSED_CALL", nil)];
-            } else {
-                notification.userInfo = @{Signal_Call_UserInfo_Key : cThread.contactIdentifier};
-                notification.category = Signal_CallBack_Category;
-                notification.alertBody =
-                    [NSString stringWithFormat:NSLocalizedString(@"MSGVIEW_MISSED_CALL", nil), [thread name]];
+            switch (self.notificationPreviewType) {
+                case NotificationNoNameNoPreview: {
+                    notification.alertBody = [NSString stringWithFormat:NSLocalizedString(@"MISSED_CALL", nil)];
```
charlesmchen: Why format?

charlesmchen: Also, maybe add translation comments on these strings?

Signal/src/environment/NotificationsManager.m
```diff
@@ -55,20 +63,94 @@ - (void)notifyUserForCall:(TSCall *)call inThread:(TSThread *)thread {
...
+                    break;
+                }
+                case NotificationNamePreview:
+                case NotificationNameNoPreview: {
+                    notification.userInfo = @{ Signal_Call_UserInfo_Key : cThread.contactIdentifier };
+                    notification.category = Signal_CallBack_Category;
+                    notification.alertBody =
+                        [NSString stringWithFormat:NSLocalizedString(@"MSGVIEW_MISSED_CALL", nil), [thread name]];
```
charlesmchen: translation comment, etc.?

Signal/src/environment/NotificationsManager.m
```diff
@@ -20,7 +22,8 @@
 @interface NotificationsManager ()
 
 @property SystemSoundID newMessageSound;
-@property (nonatomic, readonly) id<ContactsManagerProtocol> contactsManager;
+@property NSMutableDictionary<NSString *, UILocalNotification *> *currentNotifications;
```
michaelkirk: done. also added asserts to make sure we're only inspecting and mutating it on main thread.

Signal/src/environment/NotificationsManager.m
michaelkirk: done.

Signal/src/UserInterface/Strings.swift
```diff
@@ -1,12 +1,14 @@
...
-        static let missedCallNotificationBody = NSLocalizedString("MISSED_CALL", comment: "notification title")
-        static let missedCallNotificationBodyWithCallerName = NSLocalizedString("MSGVIEW_MISSED_CALL", comment: "notification title. Embeds {{Caller's Name}}")
-    }
+/**
+ * Strings re-used in multiple places should be added here.
+ */
+@objc class CallStrings: NSObject {
+    static let callBackButtonTitle = NSLocalizedString("CALLBACK_BUTTON_TITLE", comment: "notification action")
+    static let missedCallNotificationBody = NSLocalizedString("MISSED_CALL", comment: "notification title")
+    static let missedCallNotificationBodyWithCallerName = NSLocalizedString("MSGVIEW_MISSED_CALL", comment: "notification title. Embeds {{Caller's Name}}")
```
charlesmchen: I thought it over and remembered that it can be a better practice (imo) to not try to reuse strings _unless_ they're something really basic like "OK", "CANCEL", etc. Other strings - though similar in one language, might be need different translations in different languages. Sometimes that's because a language like English will use the same word for different parts of speech ("count" is a verb and a noun, for example). Another example is that more verbose languages like Italian often have longer copy and you'll need to abbreviate or reword it in some contexts (a cramped button, for example) but not others.

michaelkirk: I tend to agree, but I think in this case it makes sense to re-use, since the distinction is between "call back" a redphone call and "call back" an RTC call.

PR #1682, opened 2017-02-02
Handle “busy signal” appropriately.

charlesmchen:
PTAL @michaelkirk

PR #1687, opened 2017-02-02
Prevent simultaneous incoming and outgoing calls from leaving CallSer…

charlesmchen:
Prevent simultaneous incoming and outgoing calls from leaving CallService in a bad state.

Related to, but independent of: https://github.com/WhisperSystems/Signal-iOS/pull/1682

PTAL @michaelkirk


michaelkirk:
LGTM

charlesmchen:
Merged into master.

PR #1688, opened 2017-02-02
Stop any lingering audio session if provider resets.

michaelkirk:
I've never actually seen this happen, but I think it makes sense.

+clean up some comments.

### DIFF
--- a/Signal/src/call/Speakerbox/CallKitCallUIAdaptee.swift
+++ b/Signal/src/call/Speakerbox/CallKitCallUIAdaptee.swift
@@ -201,20 +201,12 @@ final class CallKitCallUIAdaptee: NSObject, CallUIAdaptee, CXProviderDelegate {
         AssertIsOnMainThread()
         Logger.debug("\(self.TAG) \(#function)")
 
-        // TODO
-        // copied from Speakerbox, but is there a corallary with peerconnection, since peer connection starts the audio
-        // session when adding an audiotrack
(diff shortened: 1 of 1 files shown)


Signal/src/call/Speakerbox/CallKitCallUIAdaptee.swift
```diff
@@ -201,20 +201,12 @@ final class CallKitCallUIAdaptee: NSObject, CallUIAdaptee, CXProviderDelegate {
         AssertIsOnMainThread()
         Logger.debug("\(self.TAG) \(#function)")
 
-        // TODO
-        // copied from Speakerbox, but is there a corallary with peerconnection, since peer connection starts the audio
-        // session when adding an audiotrack
-//        stopAudio()
+        // Stop any in-progress WebRTC related audio.
+        PeerConnectionClient.stopAudioSession()
```
charlesmchen: I don't understand why these static audio methods are on PCC.

michaelkirk: > why these static audio methods are on PCC.

Because it's interfacing with a WebRTC API. But I'm also not convinced that's a good reason, feel free to move them if there's a better spot.

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/43e7defa28555b1e415cfd1f022b8100f0aeeca8

PR #1689, opened 2017-02-03
Don't crash when incoming call on NonCallKit iOS10

michaelkirk:
Previous logic assumed "VoiceChat" mode, but when the ringer goes off,
we set "SoloAmbient" which is incompatible with that mode, causing
assertion failure.

// FREEBIE

PTAL @charlesmchen


michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/a328759f0dcd3a9f827a7aab0f13b9318890c45e

PR #1691, opened 2017-02-03
Be even more cautious when tearing down a PeerConnectionClient.

charlesmchen:
Eliminates a last class of edge cases around receiving "did add stream" events after a PCC has been torn down.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -281,6 +281,7 @@ protocol CallServiceObserver: class {
             peerConnectionClient.createSignalingDataChannel()
 
             assert(self.peerConnectionClient == nil, "Unexpected PeerConnectionClient instance")
+            Logger.debug("\(self.TAG) setting peerConnectionClient in \(#function)")
             self.peerConnectionClient = peerConnectionClient
 
             return self.peerConnectionClient!.createOffer()
@@ -428,6 +429,7 @@ protocol CallServiceObserver: class {
             // FIXME for first time call recipients I think we'll see mic/camera permission requests here,
             // even though, from the users perspective, no incoming call is yet visible.
             assert(self.peerConnectionClient == nil, "Unexpected PeerConnectionClient instance")
+            Logger.debug("\(self.self.TAG) setting peerConnectionClient in \(#function)")
             self.peerConnectionClient = PeerConnectionClient(iceServers: iceServers, delegate: self)
 
             let offerSessionDescription = RTCSessionDescription(type: .offer, sdp: callerSessionDescription)
@@ -921,8 +923,13 @@ protocol CallServiceObserver: class {
     /**
      * The connection has been established. The clients can now communicate.
      */
-    internal func peerConnectionClientIceConnected(_ peerconnectionClient: PeerConnectionClient) {
+    internal func peerConnectionClientIceConnected(_ peerConnectionClient: PeerConnectionClient) {
         AssertIsOnMainThread()
+
+        if peerConnectionClient != self.peerConnectionClient {
+            Logger.debug("\(self.TAG) \(#function) Ignoring event from obsolete peerConnectionClient")
+            return
+        }
 
         self.handleIceConnected()
     }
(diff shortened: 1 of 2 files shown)


Signal/src/call/CallService.swift
```diff
@@ -281,6 +281,7 @@ protocol CallServiceObserver: class {
             peerConnectionClient.createSignalingDataChannel()
 
             assert(self.peerConnectionClient == nil, "Unexpected PeerConnectionClient instance")
+            Logger.debug("\(self.TAG) setting peerConnectionClient in \(#function)")
```
charlesmchen: I've also added some logging around the setting of the pcc property. It should help us diagnose the "unexpected pcc" issue the next time we repro it.

Signal/src/call/CallService.swift
```diff
@@ -921,18 +923,28 @@ protocol CallServiceObserver: class {
     /**
      * The connection has been established. The clients can now communicate.
      */
-    internal func peerConnectionClientIceConnected(_ peerconnectionClient: PeerConnectionClient) {
+    internal func peerConnectionClientIceConnected(_ peerConnectionClient: PeerConnectionClient) {
         AssertIsOnMainThread()
 
+        if peerConnectionClient != self.peerConnectionClient {
```
michaelkirk: nit these should be `guard` statements to communicate to the ready we're anticipating early return

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -423,10 +423,7 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
         AssertIsOnMainThread()
         Logger.debug("\(TAG) in \(#function)")
 
-        delegate.peerConnectionClient(self, didUpdateLocal: nil)
-        delegate.peerConnectionClient(self, didUpdateRemote: nil)
-
-        PeerConnectionClient.signalingQueue.async {
+        PeerConnectionClient.signalingQueue.sync {
```
charlesmchen: I believe this is the most significant element of these changes: we block until the pcc's termination logic is complete.

Signal/src/call/CallService.swift
```diff
@@ -1028,21 +1060,27 @@ protocol CallServiceObserver: class {
 
         Logger.debug("\(TAG) in \(#function)")
 
-        PeerConnectionClient.stopAudioSession()
-        peerConnectionClient?.terminate()
-
-        peerConnectionClient = nil
+        self.localVideoTrack = nil
+        self.remoteVideoTrack = nil
+        self.fireDidUpdateVideoTracks()
```
michaelkirk: Isn't `fireDidUpdateVideoTracks()` redundant with the didSet?

Signal/src/call/CallService.swift
```diff
@@ -1028,21 +1060,27 @@ protocol CallServiceObserver: class {
...
         localVideoTrack = nil
         remoteVideoTrack = nil
         isRemoteVideoEnabled = false
+
+        var peerConnectionClientCopy = peerConnectionClient
```
michaelkirk: Could you clarify the intention of this reference juggling with a code comment?

Signal/src/call/CallService.swift

Signal/src/call/CallService.swift
charlesmchen: Removed.

Signal/src/call/CallService.swift
charlesmchen: I've added a comment. This "copy" dance won't have any effect, but it's about future-proofing this code and being extra-conservative since WebRTC classes are so fragile and prone to throwing exceptions on teardown.

charlesmchen:
PTAL @michaelkirk

Signal/src/call/CallService.swift
```diff
@@ -1028,21 +1060,33 @@ protocol CallServiceObserver: class {
...
+        // We make a local copy of peerConnectionClient and then clear
+        // the peerConnectionClient property immediately. This class' 
+        // PeerConnectionClientDelegate methods exit early if their 
+        // peerConnectionClient argument doesn't match the peerConnectionClient
+        // property.
+        //
+        // In practice this won't matter, since PeerConnectionClient invokes
+        // its delegate methods asynchronously, but we don't want to bake that
+        // assumption into this logic.
+        var peerConnectionClientCopy = peerConnectionClient
```
michaelkirk: I find this a little arcane, but maybe you have a clearer vision of the ways this can break. LGTM

charlesmchen: Okay, I've removed this. As noted, I was just trying to future-proof us against later changes to PCC, but better not to include anything confusing.

PR #1692, opened 2017-02-03
Don’t assert when two devices call each other simultaneously.

charlesmchen:
This shouldn't assert; it's not a bug, just an edge case that we want to handle properly.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/UserInterface/CallUIAdapter.swift
+++ b/Signal/src/call/UserInterface/CallUIAdapter.swift
@@ -47,7 +47,7 @@ extension CallUIAdaptee {
         AssertIsOnMainThread()
 
         guard self.callService.call == nil else {
-            assertionFailure("unexpectedly found an existing call when trying to call back: \(recipientId)")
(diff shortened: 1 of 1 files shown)


Signal/src/call/UserInterface/CallUIAdapter.swift
```diff
@@ -47,7 +47,7 @@ extension CallUIAdaptee {
         AssertIsOnMainThread()
 
         guard self.callService.call == nil else {
-            assertionFailure("unexpectedly found an existing call when trying to call back: \(recipientId)")
+            Logger.debug("unexpectedly found an existing call when trying to start outgoing call: \(recipientId)")
```
michaelkirk: can you bump it up to `info` so that we get it in production builds? 

Seems like a source of potentially confusing behavior in the wild, and is not likely to spam the logs a bunch.

charlesmchen: Done.

PR #1693, opened 2017-02-03
Modify handling of INStartVideoCallIntent and INStartAudioCallIntent …

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -412,6 +412,7 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
         }
         if ([Environment getCurrent].callService.call != nil) {
             DDLogWarn(@"%@ ignoring INStartAudioCallIntent due to ongoing WebRTC call.", self.tag);
+            return NO;
         }
 
         OutboundCallInitiator *outboundCallInitiator = [Environment getCurrent].outboundCallInitiator;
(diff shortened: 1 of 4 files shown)


Signal/src/AppDelegate.m
```diff
@@ -359,6 +406,14 @@ - (BOOL)application:(UIApplication *)application continueUserActivity:(nonnull N
             return NO;
         }
 
+        if ([Environment getCurrent].phoneManager.hasOngoingCall) {
+            DDLogWarn(@"%@ ignoring INStartAudioCallIntent due to ongoing RedPhone call.", self.tag);
+            return NO;
+        }
+        if ([Environment getCurrent].callService.call != nil) {
+            DDLogWarn(@"%@ ignoring INStartAudioCallIntent due to ongoing WebRTC call.", self.tag);
```
michaelkirk: `return NO`?

charlesmchen: Good catch!

Signal/src/AppDelegate.m
```diff
@@ -359,6 +406,15 @@ - (BOOL)application:(UIApplication *)application continueUserActivity:(nonnull N
...
+            return NO;
+        }
+
```
charlesmchen: `OutboundCallInitiator` is asynchronous - i.e. it does the "supports webrtc?" request before actually starting a call. We might want to modify it to make this check as well before starting a call. Hmm, I guess I'll do that.

charlesmchen: @michaelkirk I've pushed another commit that does this.

michaelkirk: You mean make the "is busy?" check before starting the call?

michaelkirk: we seem to have some sort of concurrency bug in our GH commenting. 😉

PR #1694, opened 2017-02-03
Callkit option should only be visible when "Video Call (Beta)" is ena…

charlesmchen:
https://trello.com/c/9gKHyBDg

PTAL @michaelkirk


michaelkirk:
LGTM

PR #1695, opened 2017-02-03
Prevent system edge swipe gestures from showing/hiding call controls.

charlesmchen:
https://trello.com/c/jpT95nK8

PTAL @michaelkirk


michaelkirk:
LGTM

PR #1696, opened 2017-02-03
Mkirk/coordinate busy

michaelkirk:
scenarios:

Alice is on redphone call with Bob.
Charlie calls Alice via webRTC
Charlie sees busy <-- new behavior

Alice is on webRTC call with Bob.
Charlie calls Alice via Redphone
Charlie sees busy  <-- new behavior

Also https://github.com/WhisperSystems/Signal-iOS/commit/df2d3e5a36afc4d5cc7a94b3f69b02740d14cf96 fixes a related bug where subsequent redphone calls were inappropriately not allowed due to improperly checking for an "ongoing" redphone call.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/phone/PhoneManager.m
+++ b/Signal/src/phone/PhoneManager.m
@@ -6,6 +6,7 @@
 #import "CallAudioManager.h"
 #import "PhoneManager.h"
 #import "RecentCallManager.h"
+#import "Signal-Swift.h"
 
 @implementation PhoneManager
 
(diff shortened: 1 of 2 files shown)


Signal/src/phone/PhoneManager.m
```diff
@@ -139,7 +141,16 @@ - (void)incomingCallWithSession:(ResponderSessionDescriptor *)session {
...
+    if (!self.curCallController) {
+        return NO;
+    }
+
+    return [self.curCallController callState].futureTermination.isIncomplete;
+}
+
+- (BOOL)hasOngoingWebRTCCall
+{
+    return !![Environment getCurrent].callService.call;
```
charlesmchen: Why are there two negation operators here?

charlesmchen: `[Environment getCurrent].callService.call != nil`?

Signal/src/phone/PhoneManager.m
```diff
@@ -139,7 +141,16 @@ - (void)incomingCallWithSession:(ResponderSessionDescriptor *)session {
 
 - (BOOL)hasOngoingCall
 {
-    return self.curCallController != nil;
+    if (!self.curCallController) {
+        return NO;
+    }
+
+    return [self.curCallController callState].futureTermination.isIncomplete;
```
charlesmchen: Heh, I just opened a PR that's very similar - I'll close it. I think your first condition is unnecessary though.

charlesmchen: https://github.com/WhisperSystems/Signal-iOS/pull/1698 has been closed. I've tested it and it works properly in every scenario I tried.

Signal/src/phone/PhoneManager.m
michaelkirk: It's an idiomatic way to return a bool from something truthy

michaelkirk: I think it's less clear though. I'll change it.

PR #1698, opened 2017-02-03
Fix “has red phone call” check.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/phone/PhoneManager.m
+++ b/Signal/src/phone/PhoneManager.m
@@ -139,7 +139,7 @@
 
 - (BOOL)hasOngoingCall
 {
-    return self.curCallController != nil;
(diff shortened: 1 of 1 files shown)


Signal/src/phone/PhoneManager.m
```diff
@@ -139,7 +139,7 @@ - (void)incomingCallWithSession:(ResponderSessionDescriptor *)session {
 
 - (BOOL)hasOngoingCall
 {
-    return self.curCallController != nil;
+    return [currentCallControllerObservable.currentValue callState].futureTermination.isIncomplete;
```
charlesmchen: You were right. The test I was using remained true after a red phone call is hung up.

PR #1700, opened 2017-02-03
Ensure audio service observation of call is always correctly wired up.

charlesmchen:
Previously, `CallUIAdapter` ensured that `CallAudioService` was always wired up to observe any call by doing so in the "only two" pathways for new calls: `reportIncomingCall()` and `startOutgoingCall()`.  The problem is that `OutboundCallInitiator` now uses a new pathway: `startAndShowOutgoingCall()`.  Rather than adding the right wiring there, I decided to DRY this up and by making `CallUIAdapter` observe `CallService` and setting up the observation anytime `CallService`'s call property is set.  That should handle all of these cases and future-proof us as well.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -83,6 +83,11 @@ fileprivate let timeoutSeconds = 60
 // All Observer methods will be invoked from the main thread.
 protocol CallServiceObserver: class {
     /**
+     * Fired whenever the call changes.
+     */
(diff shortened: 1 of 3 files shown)


charlesmchen:
If it's not clear, the symptom was that `CallAudioService` wasn't observing `SignalCall` on caller devices, so speakerphone and mute were broken.

michaelkirk:
Lgtm

PR #1701, opened 2017-02-03
Fix bug in the busy logic.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -414,7 +414,7 @@ protocol CallServiceObserver: class {
         Logger.verbose("\(TAG) receivedCallOffer for thread:\(thread)")
         let newCall = SignalCall.incomingCall(localId: UUID(), remotePhoneNumber: thread.contactIdentifier(), signalingId: callId)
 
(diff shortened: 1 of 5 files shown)


Signal/src/call/CallService.swift
```diff
@@ -414,7 +414,7 @@ protocol CallServiceObserver: class {
         Logger.verbose("\(TAG) receivedCallOffer for thread:\(thread)")
         let newCall = SignalCall.incomingCall(localId: UUID(), remotePhoneNumber: thread.contactIdentifier(), signalingId: callId)
 
-        guard call == nil || Environment.getCurrent().phoneManager.hasOngoingCall() else {
+        guard call == nil && !Environment.getCurrent().phoneManager.hasOngoingRedphoneCall() else {
```
charlesmchen: This is the real fix. The rest of this PR is just to improve readability.

riyapenn:
Would this address https://github.com/WhisperSystems/Signal-Android/issues/6157 (i've attached iOS log)?

charlesmchen:
@riyapenn I don't think so.

PR #1702, opened 2017-02-04
Charlesmchen/call thread safety4 

charlesmchen:
This PR rolls up a number of unrelated fixes.

I'll discuss them in inline comments.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -283,11 +283,7 @@ protocol CallServiceObserver: class {
         return getIceServers().then { iceServers -> Promise<HardenedRTCSessionDescription> in
             Logger.debug("\(self.TAG) got ice servers:\(iceServers)")
 
-            let peerConnectionClient = PeerConnectionClient(iceServers: iceServers, delegate: self)
-
-            // When placing an outgoing call, it's our responsibility to create the DataChannel. Recipient will not have
-            // to do this explicitly.
-            peerConnectionClient.createSignalingDataChannel()
+            let peerConnectionClient = PeerConnectionClient(iceServers: iceServers, delegate: self, callType: .Outgoing)
 
             assert(self.peerConnectionClient == nil, "Unexpected PeerConnectionClient instance")
             Logger.debug("\(self.TAG) setting peerConnectionClient in \(#function)")
@@ -439,7 +435,7 @@ protocol CallServiceObserver: class {
             // even though, from the users perspective, no incoming call is yet visible.
             assert(self.peerConnectionClient == nil, "Unexpected PeerConnectionClient instance")
             Logger.debug("\(self.self.TAG) setting peerConnectionClient in \(#function)")
-            self.peerConnectionClient = PeerConnectionClient(iceServers: iceServers, delegate: self)
+            self.peerConnectionClient = PeerConnectionClient(iceServers: iceServers, delegate: self, callType: .Incoming)
 
             let offerSessionDescription = RTCSessionDescription(type: .offer, sdp: callerSessionDescription)
             let constraints = RTCMediaConstraints(mandatoryConstraints: nil, optionalConstraints: nil)
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -75,7 +75,6 @@ enum CallError: Error {
     case disconnected
     case externalError(underlyingError: Error)
     case timeout(description: String)
-    case unexpected(description: String)
 }
 
 // FIXME TODO do we need to timeout?
@@ -286,7 +285,7 @@ protocol CallServiceObserver: class {
         return getIceServers().then { iceServers -> Promise<HardenedRTCSessionDescription> in
             Logger.debug("\(self.TAG) got ice servers:\(iceServers)")
 
-            let peerConnectionClient = PeerConnectionClient(iceServers: iceServers, delegate: self, callType: .Outgoing)
+            let peerConnectionClient = PeerConnectionClient(iceServers: iceServers, delegate: self, callType: .outgoing)
 
             assert(self.peerConnectionClient == nil, "Unexpected PeerConnectionClient instance")
             Logger.debug("\(self.TAG) setting peerConnectionClient in \(#function)")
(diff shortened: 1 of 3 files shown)


Signal/src/call/CallService.swift
```diff
@@ -283,11 +286,7 @@ protocol CallServiceObserver: class {
         return getIceServers().then { iceServers -> Promise<HardenedRTCSessionDescription> in
             Logger.debug("\(self.TAG) got ice servers:\(iceServers)")
 
-            let peerConnectionClient = PeerConnectionClient(iceServers: iceServers, delegate: self)
-
-            // When placing an outgoing call, it's our responsibility to create the DataChannel. Recipient will not have
-            // to do this explicitly.
-            peerConnectionClient.createSignalingDataChannel()
+            let peerConnectionClient = PeerConnectionClient(iceServers: iceServers, delegate: self, callType: .Outgoing)
```
charlesmchen: I've moved creation of the data channel into PeerConnectionClient's constructor. The constructor now takes an Incoming/Outgoing enum.

Signal/src/call/CallService.swift
```diff
@@ -428,6 +427,9 @@ protocol CallServiceObserver: class {
         let backgroundTask = UIApplication.shared.beginBackgroundTask {
             let timeout = CallError.timeout(description: "background task time ran out before call connected.")
             DispatchQueue.main.async {
+                guard self.call == newCall else {
```
charlesmchen: There are many places where we were responding to events related to obsolete calls.  I've added a bunch of `guard` statements that finally fix a number of nagging issues (i.e. unexpected PCC).

Signal/src/call/CallService.swift
```diff
@@ -996,7 +1011,6 @@ protocol CallServiceObserver: class {
         }
 
         self.localVideoTrack = videoTrack
-        self.fireDidUpdateVideoTracks()
```
charlesmchen: These method calls were redundant with the `didSet` logic.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -110,10 +115,10 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     private var videoSender: RTCRtpSender?
     private var localVideoTrack: RTCVideoTrack?
-    private var remoteVideoTrack: RTCVideoTrack?
+    private weak var remoteVideoTrack: RTCVideoTrack?
```
charlesmchen: I spent a lot of time wrestling with exceptions in the destructor of RTCVideoTrack.  RTCVideoTrack is apparently very fragile and will throw exceptions in its destructor if you retain a strong reference to it _anywhere_.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -536,22 +545,23 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     /** Called when media is received on a new stream from remote peer. */
     internal func peerConnection(_ peerConnection: RTCPeerConnection, didAdd stream: RTCMediaStream) {
+        guard stream.videoTracks.count > 0 else {
+            return
+        }
+        weak var remoteVideoTrack = stream.videoTracks[0]
```
charlesmchen: This is a critical change - see above.

charlesmchen:
I'm a little wary of prematurely declaring victory, but this certainly resolves all of the crashes, asserts and deadlock repros that I know of.  The app is also more consistent about returning to a "good state" after edge cases (like multiple frequent calls from more than one device). 

There's only one known issue: it is still possible to get the app into a state where CallService's call property is set even though there is no ongoing call, which leaves the app unable to send or receive calls.  Presumably we're missing a `[CallService handleFailedCall:]` somewhere.

charlesmchen:
> There's only one known issue: it is still possible to get the app into a state where CallService's call property is set even though there is no ongoing call, which leaves the app unable to send or receive calls. Presumably we're missing a [CallService handleFailedCall:] somewhere.

It's worth noting that with this issue the app now eventually returns to a good state after 10-30 seconds.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -139,21 +144,25 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
+            // When placing an outgoing call, it's our responsibility to create the DataChannel. 
+            // Recipient will not have to do this explicitly.
+            createSignalingDataChannel()
+        }
     }
 
     // MARK: - Media Streams
 
-    public func createSignalingDataChannel() {
+    fileprivate func createSignalingDataChannel() {
```
michaelkirk: I think this should be `private` as we're no longer using it outside of this class.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -64,6 +64,11 @@ protocol PeerConnectionClientDelegate: class {
  */
 class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelDelegate {
 
+    enum CallType {
+        case Incoming
```
michaelkirk: style nit, should be lowercased `incoming, outgoing`

Signal/src/call/CallService.swift
```diff
@@ -75,6 +75,7 @@ enum CallError: Error {
     case disconnected
     case externalError(underlyingError: Error)
     case timeout(description: String)
+    case unexpected(description: String)
```
michaelkirk: What's your intention for how `unexpected` should be used in distinction to the existing `assertionError`? They seem really close.

Signal/src/call/PeerConnectionClient.swift
charlesmchen: Done.

Signal/src/call/PeerConnectionClient.swift
charlesmchen: Done.

Signal/src/call/CallService.swift
charlesmchen: I'm a little wary of how we're using assertionError, since they're used in guards, not assertion.  Maybe this should be guardError?  Anyway, I've switched to assertionError until we come up with something better.

michaelkirk:
That is a fair point. I’m less concerned that it stay named “assertionError”, I just didn’t want to have two different ways of doing the same thing.

Feel free to rename all usages to `unexpected` or `failedPrecondition` or something.

PR #1704, opened 2017-02-06
Synchronize access to remoteVideoTrack.

charlesmchen:
Yet another attempt to get this right.  Achieving thread safety by dispatching async to threads/queues introduces delays wherein a remote video track may be deallocated after its associated `RTCPeerConnection`, which causes exceptions to be thrown in its destructor.  Therefore, we use synchronization to access PCC's `remoteVideoTrack` property synchronously.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/PeerConnectionClient.swift
+++ b/Signal/src/call/PeerConnectionClient.swift
@@ -588,7 +588,7 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
                     if strongSelf.peerConnection != nil {
                         remoteVideoTrack = strongSelf.remoteVideoTrack
                     }
-                    objc_sync_exit(self)
+                    objc_sync_exit(strongSelf)
 
                     delegate.peerConnectionClient(strongSelf, didUpdateRemote: remoteVideoTrack)
                 }
--- a/Signal/src/call/PeerConnectionClient.swift
+++ b/Signal/src/call/PeerConnectionClient.swift
@@ -115,7 +115,14 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     private var videoSender: RTCRtpSender?
     private var localVideoTrack: RTCVideoTrack?
-    private weak var remoteVideoTrack: RTCVideoTrack?
+    // RTCVideoTrack is fragile and prone to throwing exceptions and/or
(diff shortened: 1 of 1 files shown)


Signal/src/call/PeerConnectionClient.swift
```diff
@@ -548,19 +558,38 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
+                    
+                    // See the comments on the remoteVideoTrack property.
+                    //
+                    // We only access the remoteVideoTrack property if peerConnection is non-nil.
+                    var remoteVideoTrack: RTCVideoTrack?
+                    objc_sync_enter(strongSelf)
+                    if strongSelf.peerConnection != nil {
+                        remoteVideoTrack = strongSelf.remoteVideoTrack
+                    }
+                    objc_sync_exit(self)
```
michaelkirk: ` objc_sync_exit(strongSelf)`?

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -115,7 +115,14 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     private var videoSender: RTCRtpSender?
     private var localVideoTrack: RTCVideoTrack?
-    private weak var remoteVideoTrack: RTCVideoTrack?
+    // RTCVideoTrack is fragile and prone to throwing exceptions and/or
```
michaelkirk: I think audioTrack, videoSender, audioSender, suffer similarly when PeerConnectionClient holds a strong reference to them at the time RTCPeerConnection deallocs.

Signal/src/call/PeerConnectionClient.swift
charlesmchen: Great catch - thank you!

Signal/src/call/PeerConnectionClient.swift
charlesmchen: Yes, I've been wondering if I should do something similar for these properties.  I've just never seen an issue with them during any of my (at this point insanely rigorous) testing.

charlesmchen: Ah, I see why.  All of those other properties are only ever mutated on the main thread during the PCC constructor.  By definition PCC.terminate() can't be called until after that constructor returns. So I think we're safe here.

PR #1706, opened 2017-02-06
Add “signal” copy to call view.

charlesmchen:
img_4285

PTAL @michaelkirk


michaelkirk:
Are you bothered by the two ":" not lining up?
To me it has the affect of the timer baseline "feeling" too high.
Would it be better if the first colon were just omitted?

LGTM

charlesmchen:
> Would it be better if the first colon were just omitted?

Done. It's a feature of San Francisco https://medium.com/@mach/the-secret-of-san-francisco-fonts-4b5295d9a745#.w36d1if1l

PR #1709, opened 2017-02-06
Refine video icons.

charlesmchen:
I've modified the "disabled/unselected" state for the video icon to match the "disabled/unselected" state for the mute icons.  I've also improved the alignment of the video icons.  

I took another run at the "disabled/unselected" state for the speaker icons, but just come up with something better than the versions we have now (which don't match).

PTAL @michaelkirk


charlesmchen:
img_6544
img_6543
img_6542

michaelkirk:
LGTM

PR #1713, opened 2017-02-07
Better translation comments

michaelkirk:
in response to some confusion received from translators

// FREEBIE


michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/dc9ffe40e51f533982602a3c2a2341d46983bd44

PR #1715, opened 2017-02-07
Add “interstitial call view” that is shown during lengthy “webrtc sup…

charlesmchen:
…ported” check.

* The "interstitial call" view is shown briefly while the OutboundCallInitiator checks the WebRTC preference of the call recipient.
* In most cases, this check will complete very quickly.  In that case, users should not notice the "interstitial call" view.  It waits a second to appear, while still intercepting all touches.
* To avoid the "don't dismiss modal while it is being presented" issue, the "interstitial call" view is presented and dismissed without animation.
* The "interstitial call" view includes a "cancel" button.
* Is it the most handsome view in the world? No.  Could it be more consistent with the call view?  Yes.  However, this view will no longer be necessary once we remove RedPhone calling, which should be "soon".  It should be fairly easy to rip this view out.

img_4288

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -251,15 +251,15 @@ protocol CallServiceObserver: class {
     }
 
     class func presentCallInterstitialNotificationName() -> String {
-        return  "PresentCallInterstitialNotification"
+        return  "CallServicePresentCallInterstitialNotification"
     }
 
     class func dismissCallInterstitialNotificationName() -> String {
-        return  "DismissCallInterstitialNotification"
+        return  "CallServiceDismissCallInterstitialNotification"
     }
 
     class func callWasCancelledByInterstitialNotificationName() -> String {
-        return  "CallWasCancelledByInterstitialNotification"
+        return  "CallServiceCallWasCancelledByInterstitialNotification"
     }
 
     // MARK: - Service Actions
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -250,6 +250,18 @@ protocol CallServiceObserver: class {
         return  "CallServiceActiveCallNotification"
     }
 
+    class func presentCallInterstitialNotificationName() -> String {
+        return  "PresentCallInterstitialNotification"
+    }
+
+    class func dismissCallInterstitialNotificationName() -> String {
+        return  "DismissCallInterstitialNotification"
+    }
+
+    class func callWasCancelledByInterstitialNotificationName() -> String {
+        return  "CallWasCancelledByInterstitialNotification"
+    }
+
     // MARK: - Service Actions
 
     /**
--- a/Signal/src/call/OutboundCallInitiator.swift
+++ b/Signal/src/call/OutboundCallInitiator.swift
@@ -14,11 +14,37 @@ import Foundation
     let contactsManager: OWSContactsManager
     let contactsUpdater: ContactsUpdater
 
+    var cancelledCallTokens: [String] = []
+
     init(redphoneManager: PhoneManager, contactsManager: OWSContactsManager, contactsUpdater: ContactsUpdater) {
         self.redphoneManager = redphoneManager
 
         self.contactsManager = contactsManager
         self.contactsUpdater = contactsUpdater
+
+        super.init()
+
+        NotificationCenter.default.addObserver(self,
+                                               selector:#selector(callWasCancelledByInterstitial),
+                                               name:Notification.Name(rawValue: CallService.callWasCancelledByInterstitialNotificationName()),
+                                               object:nil)
+    }
+
+    deinit {
+        NotificationCenter.default.removeObserver(self)
+    }
+
+    func callWasCancelledByInterstitial(notification: NSNotification) {
+        AssertIsOnMainThread()
+
+        let callToken = notification.object as! String
+        cancelCallToken(callToken)
+    }
+
+    func cancelCallToken(_ callToken: String) {
+        AssertIsOnMainThread()
+
+        cancelledCallTokens.append(callToken)
     }
 
     /**
@@ -44,6 +70,10 @@ import Foundation
             return self.initiateRedphoneCall(recipientId: recipientId)
         }
 
+        // A temporary unique id used to identify this call during the 
+        let callToken = NSUUID().uuidString
+        presentCallInterstitial(callToken)
+
         // Since users can toggle this setting, which is only communicated during contact sync, it's easy to imagine the
         // preference getting stale. Especially as users are toggling the feature to test calls. So here, we opt for a
         // blocking network request *every* time we place a call to make sure we've got up to date preferences.
@@ -52,6 +82,10 @@ import Foundation
         // SignalRecipient *recipient = [SignalRecipient recipientWithTextSecureIdentifier:self.thread.contactIdentifier];
         self.contactsUpdater.lookupIdentifier(recipientId,
                                               success: { recipient in
+                                                guard !self.cancelledCallTokens.contains(callToken) else {
+                                                    Logger.error("\(self.TAG) OutboundCallInitiator aborting due to cancelled call.")
+                                                    return
+                                                }
 
                                                 guard !Environment.getCurrent().phoneManager.hasOngoingRedphoneCall() else {
                                                     Logger.error("\(self.TAG) OutboundCallInitiator aborting due to ongoing RedPhone call.")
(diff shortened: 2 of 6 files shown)


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -70,6 +70,12 @@
 /* No comment provided by engineer. */
 "AUDIO_PERMISSION_MESSAGE" = "Signal requires access to your microphone to work properly. You can grant this permission in the Settings app >> Privacy >> Microphone >> Signal";
 
+/* Title for call interstitial view */
+"CALL_INTERSTITIAL_CALLING_LABEL" = "Calling...";
+
+/* Label for cancel button on call interstitial view */
+"CALL_INTERSTITIAL_CANCEL_BUTTON" = "Cancel";
```
michaelkirk: Care to use our existing cancel text, or does this deserve it's own translation?

Signal/src/call/OutboundCallInitiator.swift
```diff
@@ -113,4 +150,17 @@ import Foundation
         return true
     }
 
+    private func presentCallInterstitial(_ callToken: String) {
```
michaelkirk: nit: I would leave this parameter named, e.g. `presentCallInterstitial(callToken: String)`

Idiomatically, as it reads now, I'd expect you're handing it some kind of `CallInterstitial`.

Signal/src/call/CallService.swift
```diff
@@ -249,6 +249,18 @@ protocol CallServiceObserver: class {
         return  "CallServiceActiveCallNotification"
     }
 
+    class func presentCallInterstitialNotificationName() -> String {
+        return  "PresentCallInterstitialNotification"
+    }
```
michaelkirk: For strings accessed in multiple classes like this, I like to prefix the class name, i.e. "CallServicePresentCallInterstitialNotification"

Signal/src/call/OutboundCallInitiator.swift
```diff
@@ -44,6 +70,10 @@ import Foundation
             return self.initiateRedphoneCall(recipientId: recipientId)
         }
 
+        // A temporary unique id used to identify this call during the 
+        let callToken = NSUUID().uuidString
+        presentCallInterstitial(callToken)
```
michaelkirk: not: why not pass the NSUUID around?

Signal/src/call/OutboundCallInitiator.swift
```diff
@@ -52,6 +82,10 @@ import Foundation
         // SignalRecipient *recipient = [SignalRecipient recipientWithTextSecureIdentifier:self.thread.contactIdentifier];
         self.contactsUpdater.lookupIdentifier(recipientId,
                                               success: { recipient in
+                                                guard !self.cancelledCallTokens.contains(callToken) else {
+                                                    Logger.error("\(self.TAG) OutboundCallInitiator aborting due to cancelled call.")
```
michaelkirk: I don't think this is an error -  maybe Logger.info?

Signal/src/call/OutboundCallInitiator.swift
```diff
@@ -113,4 +150,17 @@ import Foundation
...
     }
 
+    private func presentCallInterstitial(_ callToken: String) {
+        AssertIsOnMainThread()
+
+        let notificationName = CallService.presentCallInterstitialNotificationName()
+        NotificationCenter.default.post(name: NSNotification.Name(rawValue: notificationName), object: callToken)
+    }
+
+    private func dismissCallInterstitial(_ callToken: String) {
```
michaelkirk: ditto nit: `dismissCallInterstitial(callToken: String)`

Signal/src/view controllers/CallInterstitialViewController.swift
```diff
@@ -0,0 +1,166 @@
...
+    var contentView: UIView!
+
+    // MARK: Initializers
+
+    required init?(coder aDecoder: NSCoder) {
+        assert(false)
+        super.init(coder: aDecoder)
+    }
+
+    required init() {
```
michaelkirk: that storyboard is looking nervous. 👀  🔪

Signal/src/view controllers/CallInterstitialViewController.swift
```diff
@@ -0,0 +1,166 @@
...
+
+        dialingLabel.autoPinWidthToSuperview()
+        dialingLabel.autoVCenterInSuperview()
+
+        cancelCallButton.autoSetDimension(.height, toSize:ScaleFromIPhone5To7Plus(50, 60))
+        cancelCallButton.autoPinWidthToSuperview()
+        cancelCallButton.autoPinEdge(toSuperviewEdge:.bottom, withInset:ScaleFromIPhone5To7Plus(23, 41))
+    }
+
+    func cancelCallButtonPressed(sender button: UIButton) {
```
michaelkirk: nit: probably doesn't belong in `MARK: create views`

Signal/src/view controllers/CallInterstitialViewController.swift
```diff
@@ -0,0 +1,166 @@
...
+
+    // MARK: - Methods
+
+    func cancelCall() {
+        guard !wasCallCancelled else {
+            return
+        }
+        wasCallCancelled = true
+
+        assert(callToken != nil)
```
michaelkirk: note to self: where is this set?

Signal/src/view controllers/SignalsViewController.m
```diff
@@ -150,22 +158,71 @@ - (UIViewController *)previewingContext:(id<UIViewControllerPreviewing>)previewi
...
-- (void)handleActiveCallNotification:(NSNotification *)notification
+- (void)handlePresentCallInterstitialNotification:(NSNotification *)notification
 {
     AssertIsOnMainThread();
+    
+    NSString *callToken = notification.object;
+    OWSAssert(callToken != nil);
+    
+    OWSCallInterstitialViewController *viewController = [OWSCallInterstitialViewController new];
+    viewController.callToken = callToken;
```
michaelkirk: nit: As I understand it, every `CallInterstitialViewController` needs a `callToken`. 

Rather than setting it right here, and `assert`ing it's presence in CallInterstitialViewController, my impulse would be to make `callToken` a `let` and pass it to a `required` initializer - offloading some of this contract to language semantics.

e.g.  `[[OWSCallInterstitialViewController alloc] initWithCallToken:callToken];`

Signal/src/view controllers/SignalsViewController.m
```diff
@@ -150,22 +158,71 @@ - (UIViewController *)previewingContext:(id<UIViewControllerPreviewing>)previewi
...
+    void(^presentInterstitial)() = ^{
+        viewController.modalPresentationStyle = UIModalPresentationOverFullScreen;
+        [self presentViewController:viewController
+                           animated:NO
+                         completion:nil];
+    };
+    
+    // Dismiss any other modals so we can present call modal.
+    if (self.presentedViewController) {
+        [self dismissViewControllerAnimated:YES completion:^{
```
michaelkirk: nit: `[self dismissViewControllerAnimated:YES completion:presentInterstitial];`

Signal/translations/en.lproj/Localizable.strings

Signal/src/call/OutboundCallInitiator.swift
charlesmchen: Sure, either way would work. I'm not going to bother changing it, okay?

Signal/src/call/OutboundCallInitiator.swift
charlesmchen: Done.

Signal/src/call/OutboundCallInitiator.swift
charlesmchen: Done.

Signal/src/call/OutboundCallInitiator.swift
charlesmchen: Done.

Signal/src/call/CallService.swift

Signal/src/view controllers/CallInterstitialViewController.swift
charlesmchen: Hah.  Sadly, I doubt I'll ever find the time to remove it, but I absolutely intend to do all new views programmatically.

Signal/src/view controllers/CallInterstitialViewController.swift
charlesmchen: Done.

Signal/src/view controllers/CallInterstitialViewController.swift
charlesmchen: Eh, I'm generally adverse to (doing the idiomatic thing and) using constructors to set these kinds of required properties, because obj-c and swift constructors get so lengthy and awkward so quickly.

But, since you prodded me, I've done it here.

Signal/src/view controllers/SignalsViewController.m
charlesmchen: Done, see above.

Signal/src/view controllers/SignalsViewController.m
charlesmchen: Done.

Signal/src/call/OutboundCallInitiator.swift
michaelkirk: whoops, meant "nit", not "not". Totally fine to leave it.

PR #1717, opened 2017-02-07
Fix layout issue with inbox unread count on home view.

charlesmchen:
### Before

inbox-layout-wrong

### After

inbox-layout-right

PTAL @michaelkirk


michaelkirk:
LGTM!

PR #1718, opened 2017-02-08
End outgoing call with "No Answer" after 2 minutes

michaelkirk:
Also removed some dead and confusing related code from the CallInitiator refactor.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -78,7 +78,7 @@ enum CallError: Error {
 }
 
 // Should be roughly synced with Android client for consistency
-fileprivate let timeoutSeconds = 120
+fileprivate let connectingTimeoutSeconds = 120
 
 // All Observer methods will be invoked from the main thread.
 protocol CallServiceObserver: class {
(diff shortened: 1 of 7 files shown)


Signal/src/call/CallService.swift
```diff
@@ -77,8 +77,8 @@ enum CallError: Error {
     case timeout(description: String)
 }
 
-// FIXME TODO do we need to timeout?
-fileprivate let timeoutSeconds = 60
+// Should be roughly synced with Android client for consistency
+fileprivate let timeoutSeconds = 120
```
charlesmchen: Maybe rename to `connectingTimeoutSeconds`?  This doesn't apply once you're connected, right?

Signal/src/call/SignalCall.swift
```diff
@@ -17,6 +17,10 @@ enum CallState: String {
     case remoteBusy // terminal
 }
 
+enum CallDirection {
+    case outgoing, incoming
+}
```
charlesmchen: Hmm, I wonder if its worth trying to DRY this up with the enum in PCC.  Maybe not.

michaelkirk: done!

Signal/src/call/CallService.swift
michaelkirk: done.

michaelkirk:
rebased and merged here: https://github.com/WhisperSystems/Signal-iOS/commit/bf1ed9a277f8f8dafbecb94d90f8f1cdb6e97843

Issue #1724, opened 2017-02-08
Contacts list needs to refresh in any view that uses it.


charlesmchen:
The "start new conversation" and "add member to group" lists use the contacts list.  A common use pattern is that the user will search for a contact in this view, not find them, leave the app, add a contact, return to the app.  In this scenario, these views don't currently update the contact list, so they have to leave and re-enter the view to see the contact they just created.

charlesmchen:
This work is done and will ship in 2.7.0.x.

PR #1729, opened 2017-02-08
Update views that show contacts to reflect updates.

charlesmchen:
https://github.com/WhisperSystems/Signal-iOS/issues/1724

PTAL @michaelkirk

### DIFF
--- a/Signal/Signal-Prefix.pch
+++ b/Signal/Signal-Prefix.pch
@@ -15,8 +15,9 @@
     static const NSUInteger ddLogLevel = DDLogLevelInfo;
     #endif
 
-    #import "iOSVersions.h"
     #import <SignalServiceKit/Asserts.h>
+    #import "iOSVersions.h"
+    #import "OWSDispatch.h"
 
     #define SignalAlertView(title,msg) [[[UIAlertView alloc] initWithTitle:title message:msg delegate:nil cancelButtonTitle:NSLocalizedString(@"OK", @"") otherButtonTitles:nil, nil] show]
     #define SignalReportError [Pastelog reportErrorAndSubmitLogsWithAlertTitle:NSLocalizedString(@"ERROR_WAS_DETECTED_TITLE", @"") alertBody:NSLocalizedString(@"ERROR_WAS_DETECTED_SUBMIT", @"")];
--- a/Signal/Signal-Prefix.pch
+++ b/Signal/Signal-Prefix.pch
@@ -1,5 +1,5 @@
 //
-// Prefix header for all source files of the 'RedPhone' target in the 'RedPhone' project
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
 #import <Availability.h>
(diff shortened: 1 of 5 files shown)


Signal/Signal-Prefix.pch
```diff
@@ -15,8 +15,9 @@
     static const NSUInteger ddLogLevel = DDLogLevelInfo;
     #endif
 
-    #import "iOSVersions.h"
     #import <SignalServiceKit/Asserts.h>
+    #import "iOSVersions.h"
+    #import "OWSDispatch.h"
```
charlesmchen: I've added this to the PCH since I'd like to start sprinkling concurrency asserts throughout the codebase.

charlesmchen: Next time I'm in SSK, I'd like to make the same change to its pch.

michaelkirk: should be `<SignalServiceKit/OWSDispatch.h>`

Signal/src/contact/OWSContactsManager.m
```diff
@@ -99,18 +101,26 @@ - (void)setupAddressBook {
...
-        failure:^(NSError *error) {
-            DDLogWarn(@"%@ Failed to intersect contacts with error: %@. Rescheduling", self.tag, error);
-
-            [NSTimer scheduledTimerWithTimeInterval:60
-                                             target:self
-                                           selector:@selector(intersectContacts)
-                                           userInfo:nil
-                                            repeats:NO];
-        }];
+                                                                           success:^{
```
michaelkirk: nit: I think this indentation change is a failure of your clang-formatter. 😞

Signal/src/contact/OWSContactsManager.m
```diff
@@ -7,6 +7,8 @@
 
 typedef BOOL (^ContactSearchBlock)(id, NSUInteger, BOOL *);
 
+NSString *const OWSContactsManagerContactsDidChangeNotification = @"OWSContactsManagerContactsDidChangeNotification";
```
michaelkirk: nit: This is a little misleading, because we fire this in response to updating the `SignalRecipients` (via the poorly named `ContactsUpdater`, not explicitly after the Contacts change. 

Though, through a chain of notifications we update our SignalRecipients anytime the Contacts are updated, I think they're currently equivalent, but `OWSContactsManagerSignalRecipientsDidChangeNotification` or `OWSContactsManagerDidIntersectContactsNotification` might have less surprise potential down the road.

Or maybe I'm thinking too hard about it.

charlesmchen:
> should be <SignalServiceKit/OWSDispatch.h>

Done.

Signal/src/contact/OWSContactsManager.m
charlesmchen: I know, I really want to get to the bottom of this.

Signal/src/contact/OWSContactsManager.m
charlesmchen: Ah, that makes sense. Thanks for clarifying - I've taken your suggestion.  

In the long run, I wonder if "signal contacts" is clearer than "signal recipients".

michaelkirk: Yeah, maybe disambiguating both ways... something like renaming `Contact` to `SystemContact` and `SignalRecipient` to `SignalContact`

Issue #1725, opened 2017-02-08
Failed calls should show up in conversation history as such.


charlesmchen:
When you make a call and it fails, conversation reads “you called <person>”. It should read “you _tried to call_ <person>”.

charlesmchen:
Noticed another issue. If you call someone and they decline the call, you get a "you called Alice." message and a "missed call from Alice" message.  You should only see a "you tried to call Alice" message.

![img_4291](https://cloud.githubusercontent.com/assets/625803/22754504/be7eb7fe-ee0d-11e6-8b0f-9efca56e3eaa.PNG)

charlesmchen:
This work is done.

PR #1735, opened 2017-02-08
Improve the call status messages in conversation view.

charlesmchen:
https://github.com/WhisperSystems/Signal-iOS/issues/1725

img_4292

Essentially, calls start in an "incomplete incoming" or "incomplete outgoing" state and are updated to "incoming" or "outgoing" when the call connects.

PTAL @michaelkirk


charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/110

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -472,9 +475,15 @@
 "MSGVIEW_RECEIVED_CALL" = "You received a call from %@.";
 
 /* No comment provided by engineer. */
+"MSGVIEW_THEY_TRIED_TO_CALL_YOU" = "%@ tried to call you.";
```
charlesmchen: Interestingly, this message is only seen if (after future changes) you leave the call view and enter the conversation view _while receiving an incoming call_.  I went for the past tense to be consistent, and since these messages represent "events", but it might be more accurate to say "Alice is trying to call you."

Signal/src/call/CallService.swift
```diff
@@ -388,11 +389,17 @@ protocol CallServiceObserver: class {
     public func handleMissedCall(_ call: SignalCall, thread: TSContactThread) {
         AssertIsOnMainThread()
         // Insert missed call record
-        let callRecord = TSCall(timestamp: NSDate.ows_millisecondTimeStamp(),
-                                withCallNumber: thread.contactIdentifier(),
-                                callType: RPRecentCallTypeMissed,
-                                in: thread)
-        callRecord.save()
+        if call.callRecord == nil {
```
michaelkirk: nit: Testing if a thing is nil, and then doing something with that thing is often a good spot for an `if let`. 

So, specifically, if you flip the condition you could do this:
```
if let callRecord = call.callRecord {
    callRecord.updateCallType(RPRecentCallTypeMissed)
} else {
    callRecord = TSCall(timestamp: NSDate.ows_millisecondTimeStamp(),
                                      withCallNumber: thread.contactIdentifier(),
                                      callType: RPRecentCallTypeMissed,
                                      in: thread)
}
```

The motivation being that code with less `!`'s is a little more resilient to future refactors, because the compiler will remind you if that thing is possibly nil rather than crashing at runtime if it is.

Signal/src/call/CallService.swift
```diff
@@ -388,11 +389,17 @@ protocol CallServiceObserver: class {
...
+            call.callRecord = TSCall(timestamp: NSDate.ows_millisecondTimeStamp(),
+                                     withCallNumber: thread.contactIdentifier(),
+                                     callType: RPRecentCallTypeMissed,
+                                     in: thread)
+        } else if (call.callRecord!.callType == RPRecentCallTypeIncoming) {
+            call.callRecord!.updateCallType(RPRecentCallTypeMissed)
+        }
+
+        assert(call.callRecord != nil)
```
michaelkirk: :+1: I've been doing conservative things like this too. It really shouldn't be nil, but don't explode about it.

Signal/src/call/CallService.swift
```diff
@@ -1152,6 +1161,33 @@ protocol CallServiceObserver: class {
...
+    internal func callRecordDidChange(call: SignalCall, callRecord: TSCall?) {
+        AssertIsOnMainThread()
+
+        updateCallRecordType(call: call)
+    }
+
+    internal func updateCallRecordType(call: SignalCall?) {
+        AssertIsOnMainThread()
+
+        guard call != nil else {
```
michaelkirk: nit: ditto about rebinding `call` and `callRecord` to a local non-optional.

```
    return
}
```

then use the non-optional `call` (sans `!`) in the rest of the method.

Signal/src/call/CallService.swift
```diff
@@ -1152,6 +1161,33 @@ protocol CallServiceObserver: class {
         // Do nothing
     }
 
+    internal func callRecordDidChange(call: SignalCall, callRecord: TSCall?) {
+        AssertIsOnMainThread()
+
+        updateCallRecordType(call: call)
+    }
+
+    internal func updateCallRecordType(call: SignalCall?) {
```
michaelkirk: nit: do you dare move this to something like `call.completeAnyCallRecord()` method. This class is so effing huge. 

Also I think that would let us get rid of the `callRecordDidChange` machinery which is adding a few lines of code for how little it's doing.

Signal/src/call/CallService.swift

Signal/src/call/CallService.swift
charlesmchen: I'd love to decompose CallService, but I'm ambivalent about this. It moves business logic into the model.  I prefer dumb models.  Anyway, I've done this.

Signal/src/call/CallService.swift

Issue #1728, opened 2017-02-08
Simplify interactions around attachments


charlesmchen:
* It should be very easy to view/play a image, video or audio attachment from conversation view.
* It should be very easy to save a image, video or audio attachment from conversation view or "attachment view".
* It should be very easy to dismiss "attachment view".

Different messaging apps use different gestures for these actions - we should be permissive and accept multiple gestures.  Additionally, in some cases we may want to use the "any touch" gesture recognizer which is faster than a tap.

michaelkirk:
I agree with this, but it would be good to propose more specific changes before forging ahead here.

charlesmchen:
@michaelkirk Good point:

Proposed changes:

* Swipe-up to dismiss "photo attachment view" and "video attachment view". This is a common interaction in other clients.
* Double-tap (or more taps, and taps with multiple fingers) to dismiss those same views. Many users don't distinguish between tap and double-tap and/or perform those gestures sloppily.  Currently double tap with a single touch zooms a photo which is unlikely to be the intent - and is redundant with pinch-to-zoom.
* Long press shows "copy/delete/save" in conversation view but not in "photo attachment view" and "video attachment view".  It should work in both.
* "Copy/delete/save" controls are positioned outside the image/video, which is not consistent with other apps.  It should appear just above/below the touch. Current positioning is also ambiguous about which attachment will be affected, since the controls can easily appear above a different attachment.
* Audio attachment's "play" button sometimes disappears if pressed.  
* Tapping the audio attachment's "play" button doesn't work - you have to tap the waveform.
* "Photo attachment view" and "video attachment view" animate in. First, I think these animations are slow and unnecessary - they make interacting with multiple attachments cumbersome.  The animations should either be much faster or removed. Secondly, these animations are inconsistent. Photos "drop down from above" while video "drop in and zoom in from upper-left hand corner."  If we retain these animations, they should be consistent.
* "Video attachment view" has a black square-cornered letterbox background and a toolbar with a "done" button. "Image attachment view" has a semi-transparent background, rounded corners and no toolbar or "done" button. I thing we should make them consistent, probably by adopting the video style.
* Unlike image and video attachments, there's no way to make the audio attachment play in full screen mode.  We don't need to change this, but its inconsistent.  If we went fullscreen, it'd be much easier to scrub the audio.  

Thoughts @michaelkirk ?

riyapenn:
[...]

PR #1751, opened 2017-02-11
Charlesmchen/attachment views

charlesmchen:
https://github.com/WhisperSystems/Signal-iOS/issues/1728

* Add save/copy menu to the image attachment view.
* Let users dismiss image attachment view with tap, double-tap or swipe.
* Remove rounded corners from image attachment view.
* Clean up present & dismiss animations for image attachment view.
* Fix present & dismiss animations for video attachment view.
* Ensure video attachment view is cleaned up.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/UIView+OWS.h
+++ b/Signal/src/UIView+OWS.h
@@ -2,6 +2,7 @@
 //  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
+#import "PureLayout.h"
 #import <UIKit/UIKit.h>
 
 // A convenience method for doing responsive layout. Scales between two
--- a/Signal/src/UIView+OWS.h
+++ b/Signal/src/UIView+OWS.h
@@ -2,7 +2,7 @@
 //  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
-#import "PureLayout.h"
+#import <PureLayout/PureLayout.h>
 #import <UIKit/UIKit.h>
 
 // A convenience method for doing responsive layout. Scales between two
--- a/Signal/src/view controllers/FullImageViewController.m
+++ b/Signal/src/view controllers/FullImageViewController.m
@@ -1,44 +1,60 @@
 //
-//  FullImageViewController.m
-//  Signal
-//
-//  Created by Dylan Bourgeois on 11/11/14.
-//  Animated GIF support added by Mike Okner (@mikeokner) on 11/27/15.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
 #import "FLAnimatedImage.h"
 #import "FullImageViewController.h"
 #import "UIUtil.h"
-
-#define kImageViewCornerRadius 5.0f
+#import "UIView+OWS.h"
+#import "TSPhotoAdapter.h"
+#import "TSMessageAdapter.h"
+#import "TSAnimatedAdapter.h"
 
 #define kMinZoomScale 1.0f
 #define kMaxZoomScale 8.0f
-#define kTargetDoubleTapZoom 3.0f
 
 #define kBackgroundAlpha 0.6f
 
+// In order to use UIMenuController, the view from which it is
+// presented must have certain custom behaviors.
+@interface AttachmentMenuView : UIView
+
+@end
+
+#pragma mark -
+
+@implementation AttachmentMenuView
+
+- (BOOL)canBecomeFirstResponder {
+    return YES;
+}
+
+// We only use custom actions in UIMenuController.
+- (BOOL)canPerformAction:(SEL)action
+              withSender:(id)sender
+{
+    return NO;
+}
+
+@end
+
+#pragma mark -
+
 @interface FullImageViewController () <UIScrollViewDelegate, UIGestureRecognizerDelegate>
 
-@property (nonatomic, strong) UIView *backgroundView;
+@property (nonatomic) UIView *backgroundView;
+@property (nonatomic) UIScrollView *scrollView;
+@property (nonatomic) UIImageView *imageView;
+@property (nonatomic) UIButton *shareButton;
 
-@property (nonatomic, strong) UIScrollView *scrollView;
+@property (nonatomic) CGRect originRect;
+@property (nonatomic) BOOL isPresenting;
+@property (nonatomic) BOOL isAnimated;
+@property (nonatomic) NSData *fileData;
 
-@property (nonatomic, strong) UIImageView *imageView;
-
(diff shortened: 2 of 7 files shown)


Signal/src/UIView+OWS.h
```diff
@@ -2,6 +2,7 @@
 //  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
+#import "PureLayout.h"
```
michaelkirk: nit: #import <PureLayout/PureLayout.h>

Signal/src/view controllers/FullImageViewController.m
```diff
@@ -128,56 +159,106 @@ - (void)populateImageView:(UIImage *)image {
...
-    [self.singleTap requireGestureRecognizerToFail:self.doubleTap];
-
-    self.singleTap.delegate = self;
-    self.doubleTap.delegate = self;
+    UITapGestureRecognizer *singleTap = [[UITapGestureRecognizer alloc] initWithTarget:self
+                                                                                action:@selector(imageDismissGesture:)];
+    singleTap.delegate = self;
+    [self.view addGestureRecognizer:singleTap];
+    
+    UITapGestureRecognizer *doubleTap = [[UITapGestureRecognizer alloc] initWithTarget:self
```
michaelkirk: Why diverge from iMessage behavior, which is to zoom on double tap?

Signal/src/view controllers/FullImageViewController.m
```diff
@@ -128,56 +159,106 @@ - (void)populateImageView:(UIImage *)image {
...
 
-    self.view.userInteractionEnabled = NO;
+- (void)performEditingActionWithSelector:(SEL)selector {
+    OWSAssert(self.messageItem.messageType == TSIncomingMessageAdapter ||
+              self.messageItem.messageType == TSOutgoingMessageAdapter);
+    OWSAssert([self.messageItem isMediaMessage]);
+    OWSAssert([self.messageItem isKindOfClass:[TSMessageAdapter class]]);
+    OWSAssert([self.messageItem conformsToProtocol:@protocol(OWSMessageEditing)]);
+    OWSAssert([[self.messageItem media] isKindOfClass:[TSPhotoAdapter class]] ||
+              [[self.messageItem media] isKindOfClass:[TSAnimatedAdapter class]]);
```
michaelkirk: Can this be a `TSVideoAttachmentAdapter`?

Signal/src/view controllers/MessagesViewController.m
```diff
@@ -1365,9 +1389,26 @@ - (void)handleWarningTap:(TSInteraction *)interaction
     }
 }
 
-
 - (void)moviePlayBackDidFinish:(id)sender {
-    DDLogDebug(@"playback finished");
+    DDLogDebug(@"%@ %s", self.tag, __PRETTY_FUNCTION__);
```
michaelkirk: is this intentionally a no-op?

Signal/src/view controllers/MessagesViewController.m
```diff
@@ -1365,9 +1389,26 @@ - (void)handleWarningTap:(TSInteraction *)interaction
...
+}
+
+- (void)moviePlayerWillExitFullscreen:(id)sender {
```
michaelkirk: I would have expected doing the same thing on *will*Exit and *did*Exit to be redundant. Can you leave a comment to explain how you arrived at this?

Signal/src/view controllers/FullImageViewController.m
```diff
@@ -215,9 +301,9 @@ - (void)presentFromViewController:(UIViewController *)viewController {
 
 - (void)dismiss {
     self.view.userInteractionEnabled = NO;
-    [UIView animateWithDuration:0.4f
```
michaelkirk: 👏 👏 👏

Signal/src/view controllers/FullImageViewController.m
```diff
@@ -128,56 +159,106 @@ - (void)populateImageView:(UIImage *)image {
...
 }
 
 #pragma mark - Gesture Recognizers
 
-- (void)imageDoubleTapped:(UITapGestureRecognizer *)doubleTap {
-    CGPoint tap          = [doubleTap locationInView:doubleTap.view];
-    CGPoint convertCoord = [self.scrollView convertPoint:tap fromView:doubleTap.view];
-    CGRect targetZoomRect;
-    UIEdgeInsets targetInsets;
+- (void)imageDismissGesture:(UIGestureRecognizer *)sender {
```
michaelkirk: nit: clang-format

Signal/src/UIView+OWS.h
charlesmchen: Done.

Signal/src/view controllers/FullImageViewController.m
charlesmchen: Here's my argument: 

* 90% of the time, users who double-tap want to dismiss.
* People rarely want to zoom photos.  
* If People do want to zoom into photos, they can pinch.  Pinching is a far better interaction for zoom - it's continuous, has continuous feedback, is reversible, let's users pan and scale at the same time, etc.  Double-tap just zooms into a predetermined scale.  Users will almost certainly have to pinch afterward to scale/pan to the portion of the image of interest.
* There's a huge fraction of users who never learn to distinguish single-tap and double-tap.  They double tap everything.  

Is that persuasive?

Signal/src/view controllers/FullImageViewController.m
charlesmchen: I think clang-format isn't changing these because our .clang-format doesn't include a `BreakBeforeBraces` directive.  Not sure if we inherit one from `BasedOnStyle: WebKit`.  I've tried specifying `Stroustrup` (which I think is what you prefer), but it doesn't seem to be working.

Signal/src/view controllers/FullImageViewController.m
charlesmchen: No, one thing I'd like to do is generalize `FullImageViewController` to a `MediaViewController` or `AttachmentViewController` and use it for audio and video as well.  But I haven't done that yet.

Signal/src/view controllers/MessagesViewController.m
charlesmchen: Removed.

Signal/src/view controllers/MessagesViewController.m
charlesmchen: Done.

Signal/src/view controllers/FullImageViewController.m
michaelkirk: > Is that persuasive?

I'd assume iMessage knows all of this, and arrived at their conclusion for a reason, but I won't block the PR on it.

PR #1763, opened 2017-02-14
Flush logs before submitting logs.

charlesmchen:
I've verified that very recent activity shows up in submitted logs after this change.

PTAL @michaelkirk


charlesmchen:
@michaelkirk  I just pushed another commit. PTAL

charlesmchen:
@michaelkirk I've moved the follow-on changes to a separate PR: https://github.com/WhisperSystems/Signal-iOS/pull/1764

PR #1764, opened 2017-02-14
Flush logs before exiting the app.

charlesmchen:
Builds upon: https://github.com/WhisperSystems/Signal-iOS/pull/1763

PTAL @michaelkirk


michaelkirk:
LGTM

PR #1767, opened 2017-02-15
Update Cocoapods and Carthage submodule.

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/Signal-Carthage/pull/2


michaelkirk:
LGTM

Issue #1776, opened 2017-02-16
WebRTC setting issues.


charlesmchen:
https://twitter.com/georgemaschke/status/832151758688370688

The WebRTC setting is updated pessimistically - only after the service updates the change.  We do this in a "best effort" to keep the local and service values for this setting aligned.  

This presents various problems.  I deliberately decided not to present an activity indicator, locking the UI during the update, etc. This setting is temporary and will go away as soon as we make WebRTC the default, which could happen as soon as our next release. This means that some users will occasionally fail to update this preference without realizing it.  I consider this acceptable for a temporary, beta setting that will work most of the time.

Unfortunately, this is compounded by a bug that ignores successful updates if the user dismisses the settings before the request completes.  I'm going to fix that.

PR #1777, opened 2017-02-16
Don’t ignore WebRTC setting changes if user dismisses settings before…

charlesmchen:
… success.

Fixes: https://github.com/WhisperSystems/Signal-iOS/issues/1776

PTAL @michaelkirk


michaelkirk:
LGTM

Issue #1747, opened 2017-02-11
Sharing with UIActivityViewController


charlesmchen:
Users should be able to "share" image, video and files (and other attachments?) from a conversation to other apps.

charlesmchen:
Done.

kiritmodi2702:
UIActivityViewController : https://iosdevcenters.blogspot.com/2017/08/how-to-share-content-with.html

PR #1783, opened 2017-02-16
Add sharing of attachments.

charlesmchen:
PTAL @michaelkirk 

img_6918
img_6917
img_6916
img_6915

### DIFF
--- a/Signal/src/Models/TSMessageAdapaters/TSMessageAdapter.m
+++ b/Signal/src/Models/TSMessageAdapaters/TSMessageAdapter.m
@@ -1,11 +1,8 @@
-//  TSMessageAdapter.m
 //
-//  Signal
-//
-//  Created by Frederic Jacobs on 24/11/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
+#import "AttachmentSharing.h"
 #import "OWSCall.h"
 #import "TSAttachmentPointer.h"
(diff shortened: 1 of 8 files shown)


Signal/src/Models/TSMessageAdapaters/TSMessageAdapter.m
```diff
@@ -271,6 +277,29 @@ - (void)performEditingAction:(SEL)action
...
+        return nil;
+    }
+
+    TSMessage *message = (TSMessage *)self.interaction;
+
+    if (![message hasAttachments]) {
+        return nil;
+    }
+    OWSAssert(message.attachmentIds.count <= 1);
+    NSString *attachmentID = message.attachmentIds[0];
```
charlesmchen: I've made the assumption that messages have at most one attachment.

charlesmchen:
Fixes: https://github.com/WhisperSystems/Signal-iOS/issues/1747

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -169,6 +169,9 @@
 /* Short name for edit menu item to save contents of media message. */
 "EDIT_ITEM_SAVE_ACTION" = "Save";
 
+/* Short name for edit menu item to share contents of media message. */
```
michaelkirk: nit: Do you want to re-use the existing "ATTACHMENT_VIEW_SHARE_ACTION" which has an identical translation and  description?

charlesmchen: Sure. I've de-duplicated other strings too.

PR #1785, opened 2017-02-17
Improve image scaling quality.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/Models/TSMessageAdapaters/TSPhotoAdapter.m
+++ b/Signal/src/Models/TSMessageAdapaters/TSPhotoAdapter.m
@@ -51,6 +51,10 @@
         imageView.frame         = CGRectMake(0.0f, 0.0f, size.width, size.height);
         imageView.contentMode   = UIViewContentModeScaleAspectFill;
         imageView.clipsToBounds = YES;
+        // Use trilinear filters for better scaling quality at
+        // some performance cost.
+        imageView.layer.minificationFilter = kCAFilterTrilinear;
+        imageView.layer.magnificationFilter = kCAFilterTrilinear;
         [JSQMessagesMediaViewBubbleImageMasker applyBubbleImageMaskToMediaView:imageView
                                                                     isOutgoing:self.appliesMediaViewMaskAsOutgoing];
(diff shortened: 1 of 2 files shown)


Signal/src/view controllers/FullImageViewController.m
```diff
@@ -147,6 +147,10 @@ - (void)initializeImageView {
         self.imageView.userInteractionEnabled       = YES;
         self.imageView.clipsToBounds                = YES;
         self.imageView.layer.allowsEdgeAntialiasing = YES;
+        // Use trilinear filters for better scaling quality at
+        // some performance cost.
+        self.imageView.layer.minificationFilter = kCAFilterTrilinear;
+        self.imageView.layer.magnificationFilter = kCAFilterTrilinear;
```
charlesmchen: We definitely want to use the highest quality scaling in the image attachment view.

charlesmchen: Note that this only applies to still images.  Animated gifs are low-res and don't give the eye time to examine fine detail.

Signal/src/Models/TSMessageAdapaters/TSPhotoAdapter.m
```diff
@@ -50,6 +50,10 @@ - (UIView *)mediaView {
         imageView.frame         = CGRectMake(0.0f, 0.0f, size.width, size.height);
         imageView.contentMode   = UIViewContentModeScaleAspectFill;
         imageView.clipsToBounds = YES;
+        // Use trilinear filters for better scaling quality at
+        // some performance cost.
+        imageView.layer.minificationFilter = kCAFilterTrilinear;
+        imageView.layer.magnificationFilter = kCAFilterTrilinear;
```
charlesmchen: I've mixed feelings about using higher quality scaling in the messages view.  On the one hand, images will be usually scaled down quite a bit and thus vulnerable to scaling artifacts.  On the other hand, this could hypothetically affect scrolling performance.  I don't think it will though; Signal conversations aren't likely to have more than a handful of still images in view at any given time.

PR #1598, opened 2017-01-12
Charlesmchen/webrtc setting

charlesmchen:
### This PR depends on open PRs and isn't ready for review.

Adds a preference that enables WebRTC-based calling. 

* Add a preference to `PropertyListPreferences`.
* Update every we update our account preferences to include the new "is WebRTC enabled" property.
  * See https://github.com/WhisperSystems/SignalServiceKit/pull/89
* Add UI to the "Advanced Settings" VC to expose this property.
   * A best effort is made to keep client and service state in sync.

PR #1798, opened 2017-02-21
Enable WebRTC-based audio and video calls by default.

charlesmchen:
Reverts https://github.com/WhisperSystems/Signal-iOS/pull/1598/ and enables WebRTC calling by default.

See: https://github.com/WhisperSystems/SignalServiceKit/pull/121

PTAL @michaelkirk


charlesmchen:
@michaelkirk This work might introduce a crashing bug into the advanced settings view, but it's fixed in my next PR which I'm opening now.

PR #1801, opened 2017-02-22
Charlesmchen/callkit privacy

charlesmchen:
It seems like we don't need a CallKit extension after all.

* Adds a new "call kit privacy" preference. The user-facing label is "Show names & numbers in CallKit".
* "Call kit privacy" is enabled by default.
* If "call kit privacy" is disabled, calls function as before.
* If "call kit privacy" is enabled, we use "generic" ids (a constant prefix + the call's local id which is a UUID) to identify the calls with CallKit and label them as "Signal User" instead of the contact's name.
* We persist a mapping of "callkit id-to-phone number" so if users initiate Signal calls from phone's call history, we can call the right user.
* A new "callkit id" is used for each call to avoid leaking even the fact that you repeatedly called the same user.
* Remove the old "isCallKitEnabled" preference.

I've tested various scenarios:

* Successful calls (caller and callee).
* Cancelled calls (caller and callee).
* Audio and video calling from phone call history.

Not yet done:

* Move the "call kit privacy" preference from "advanced settings" to "privacy settings" view.
* We should discuss the copy.

PTAL @michaelkirk 

img_4297
img_4296

### DIFF
--- a/Signal/src/view controllers/PrivacySettingsTableViewController.m
+++ b/Signal/src/view controllers/PrivacySettingsTableViewController.m
@@ -157,10 +157,14 @@ typedef NS_ENUM(NSInteger, PrivacySettingsTableViewControllerSectionIndex) {
             return NSLocalizedString(@"SETTINGS_CALLING_HIDES_IP_ADDRESS_PREFERENCE_TITLE_DETAIL",
                 @"User settings section footer, a detailed explanation");
         case PrivacySettingsTableViewControllerSectionIndexCallKitEnabled:
-            return NSLocalizedString(@"SETTINGS_SECTION_CALL_KIT_DESCRIPTION", @"Settings table section footer.");
+            return (self.supportsCallKit
+                    ? NSLocalizedString(@"SETTINGS_SECTION_CALL_KIT_DESCRIPTION", @"Settings table section footer.")
+                    : nil);
         case PrivacySettingsTableViewControllerSectionIndexCallKitPrivacy:
-            return NSLocalizedString(@"SETTINGS_SECTION_CALL_KIT_PRIVACY_DESCRIPTION",
-                                     @"Explanation of the 'CallKit Privacy` preference.");
+            return ((self.supportsCallKit && [[Environment getCurrent].preferences isCallKitEnabled])
+                    ? NSLocalizedString(@"SETTINGS_SECTION_CALL_KIT_PRIVACY_DESCRIPTION",
+                                        @"Explanation of the 'CallKit Privacy` preference.")
+                    : nil);
         case PrivacySettingsTableViewControllerSectionIndexBlockOnIdentityChange:
             return NSLocalizedString(
                 @"SETTINGS_BLOCK_ON_IDENITY_CHANGE_DETAIL", @"User settings section footer, a detailed explanation");
@@ -262,6 +266,7 @@ typedef NS_ENUM(NSInteger, PrivacySettingsTableViewControllerSectionIndex) {
     DDLogInfo(@"%@ user toggled call kit preference: %@", self.tag, (sender.isOn ? @"ON" : @"OFF"));
     [[Environment getCurrent].preferences setIsCallKitEnabled:sender.isOn];
     [[Environment getCurrent].callService createCallUIAdapter];
+    [self.tableView reloadData];
 }
 
 - (void)didToggleEnableCallKitPrivacySwitch:(UISwitch *)sender {


charlesmchen:
Okay I pushed another commit with this part of the task:

> Move the "call kit privacy" preference from "advanced settings" to "privacy settings" view.


screen shot 2017-02-22 at 11 39 23 am
screen shot 2017-02-22 at 11 36 52 am

Podfile
```diff
@@ -6,7 +6,7 @@ target 'Signal' do
     pod 'AxolotlKit',                 git: 'https://github.com/WhisperSystems/SignalProtocolKit.git'
     #pod 'AxolotlKit',                 path: '../SignalProtocolKit'
     #pod 'SignalServiceKit',           git: 'https://github.com/WhisperSystems/SignalServiceKit.git'
-    pod 'SignalServiceKit',           git: 'https://github.com/WhisperSystems/SignalServiceKit.git', :branch => 'charlesmchen/webrtcByDefault'
+    pod 'SignalServiceKit',           git: 'https://github.com/WhisperSystems/SignalServiceKit.git', :branch => 'charlesmchen/callkitPrivacy'
```
michaelkirk: Can you be sure to change this back to master and `pod install` after merging the corresponding SSK Otherwise it's easy to end up referencing an orphaned SSK commit, which explodes when you run pod install.

Signal/src/call/Speakerbox/CallKitCallUIAdaptee.swift
```diff
@@ -234,6 +242,19 @@ final class CallKitCallUIAdaptee: NSObject, CallUIAdaptee, CXProviderDelegate {
         _ = self.callService.handleOutgoingCall(call)
         action.fulfill()
         self.provider.reportOutgoingCall(with: call.localId, startedConnectingAt: nil)
+
+        ensureCallName(call:call)
```
michaelkirk: nit: This function name is a little funny, and we're only using it in one place. I'd prefer just inlining this function. But if you want to keep it you should call it something that hints we're reporting the call, not just munging the name.

Podfile
charlesmchen: Yup, done.

Signal/src/call/Speakerbox/CallKitCallUIAdaptee.swift
charlesmchen: Inlined it.  

> we're reporting the call, not just munging the name.

You comment makes me unsure that I understand the CallKit API.  The intent her is to "set" (not munge) the name.  When "CallKit privacy" is enabled, we create new incoming and outgoing calls using a "generic" handle, not a phone number, so we have to set the `localizedCallerName` once the call is created.  Otherwise, CallKit presents the call to the user using the "generic" UUID.  So this `CXCallUpdate` is used to overwrite the UUID with the "Signal User" copy.  

> we're reporting the call

My understanding is that `CXCallUpdate` and `provider.reportCall` are how you update properties of a call.  Your comment makes it sounds like there's more to it.  What am I missing?

charlesmchen:
PTAL @michaelkirk

Signal/src/view controllers/AdvancedSettingsTableViewController.m
```diff
@@ -18,12 +18,12 @@
 
 @interface AdvancedSettingsTableViewController ()
 
-@property (nonatomic) UITableViewCell *enableCallKitCell;
+@property (nonatomic) UITableViewCell *enableCallKitPrivacyCell;
 @property (nonatomic) UITableViewCell *enableLogCell;
 @property (nonatomic) UITableViewCell *submitLogCell;
 @property (nonatomic) UITableViewCell *registerPushCell;
 
-@property (nonatomic) UISwitch *enableCallKitSwitch;
```
michaelkirk: We still want to retain the ability to enable CallKit - CallKitPrivacy is sort a "sub feature" relevant only while CallKit is enabled.

Also, at this point, it makes sense for both of these features to be in the "Privacy" settings, not "advanced".

charlesmchen: Done. Unfortunately, I had to do this work twice, because I did half of it at the airport and somehow lost the commit/work.

charlesmchen:
PTAL @michaelkirk

Signal/src/view controllers/PrivacySettingsTableViewController.m
```diff
@@ -123,6 +156,11 @@ - (nullable NSString *)tableView:(UITableView *)tableView titleForFooterInSectio
         case PrivacySettingsTableViewControllerSectionIndexCalling:
             return NSLocalizedString(@"SETTINGS_CALLING_HIDES_IP_ADDRESS_PREFERENCE_TITLE_DETAIL",
                 @"User settings section footer, a detailed explanation");
+        case PrivacySettingsTableViewControllerSectionIndexCallKitEnabled:
```
michaelkirk: We need to hide these footers conditionally upon CallKit being disabled or not supported. e.g. I see both footers (but neither cell) on my iOS9 device.

Signal/src/view controllers/PrivacySettingsTableViewController.m
```diff
@@ -106,6 +135,10 @@ - (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger
             return 1;
         case PrivacySettingsTableViewControllerSectionIndexCalling:
             return 1;
+        case PrivacySettingsTableViewControllerSectionIndexCallKitEnabled:
+            return self.supportsCallKit ? 1 : 0;
+        case PrivacySettingsTableViewControllerSectionIndexCallKitPrivacy:
+            return (self.supportsCallKit && [[Environment getCurrent].preferences isCallKitEnabled]) ? 1 : 0;
```
michaelkirk: It seems like this should hide the callkit privacy section when callkit is disabled. and yet...

Signal/src/view controllers/PrivacySettingsTableViewController.m
charlesmchen: Good catch, done.

charlesmchen: I haven't tested these changes yet.

Signal/src/view controllers/PrivacySettingsTableViewController.m
charlesmchen: Done. I wonder if we should move the three calling-related settings to the bottom of the view?

PR #1811, opened 2017-02-25
Call sounds - connecting, outgoing ring, busy sound.

michaelkirk:
Also add vibration for remote hangup so you're not talking to dead air.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/CallAudioService.swift
+++ b/Signal/src/call/CallAudioService.swift
@@ -3,16 +3,51 @@
 //
 
 import Foundation
+import AVFoundation
 
 @objc class CallAudioService: NSObject, CallObserver {
 
     private let TAG = "[CallAudioService]"
     private var vibrateTimer: Timer?
-    private let soundPlayer = JSQSystemSoundPlayer.shared()!
+    private let audioPlayer = AVAudioPlayer()
     private let handleRinging: Bool
 
-    enum SoundFilenames: String {
-        case incomingRing = "r"
+    class Sound {
+        let TAG = "[Sound]"
+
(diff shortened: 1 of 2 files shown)


Signal/src/call/CallAudioService.swift
```diff
@@ -76,89 +111,152 @@ import Foundation
...
     }
 
-    private func handleDialing() {
+    private func handleDialing(call: SignalCall) {
         Logger.debug("\(TAG) \(#function)")
+
+        if call.isSpeakerphoneEnabled {
+            setAudioSession(category: AVAudioSessionCategoryPlayAndRecord,
+                            mode: AVAudioSessionModeVoiceChat,
+                            options: [.defaultToSpeaker, .mixWithOthers])
```
charlesmchen: oh neat, so this is the way to specify multiple enum values in Swift?

Signal/src/call/CallService.swift
```diff
@@ -1198,6 +1198,7 @@ protocol CallServiceObserver: class {
             call!.hasLocalVideo)
     }
 
+    //TODO only fire this when it's changed? as of right now it gets called whenever you e.g. lock the phone while it's incoming ringing.
```
charlesmchen: Yes, we over-fire this.  We could certainly use a local property to de-duplicate these messages but I haven't bothered yet.

Signal/src/call/CallAudioService.swift
michaelkirk: Not enums exactly, but options yes: https://developer.apple.com/reference/swift/optionset

Issue #1794, opened 2017-02-19
Back button on new message screen doesn't work.


darrenw:
- [x] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description
The back button on MessageComposeTableViewController doesn't work.
I'm not particularly used to Storyboard based development but I believe the issue is that MessageComposeTableViewController is being modally presented in a navigationcontroller, and the taps on the back button it gets from calling 'useOWSBackButton' call popViewController, which doesn't dismiss the modal and so it goes nowhere.

### Steps to reproduce
- Launch signal.
- Hit the 'new message button in the top right'
- Hit the back button on the resulting screen.

**Actual result:** Nothing happens.
**Expected result:**  The new message screen should disappear and go back to the main screen.

### Device info
**Signal version:** Master built off commit 5cd1d17056dab08b5b0f30487a632746bfae61b2

charlesmchen:
@darrenw Fixed in https://github.com/WhisperSystems/Signal-iOS/pull/1813.  Fix will ship in next release (1-2 weeks).

PR #1813, opened 2017-02-27
Fix back button in “new conversation” view.

charlesmchen:
Fixes https://github.com/WhisperSystems/Signal-iOS/issues/1794

PTAL @michaelkirk


michaelkirk:
I ran and verified that this fixes the behavior, but as the compose view is presented modally, it seems that we should show the "X" instead of the "<"?

charlesmchen:
> I ran and verified that this fixes the behavior, but as the compose view is presented modally, it seems that we should show the "X" instead of the "<"?

Good point. PTAL @michaelkirk

michaelkirk:
LGTM

Issue #1821, opened 2017-03-02
Incorrect country code by country name when registering, prevents registration


riyapenn:
- [X] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description
Countries have incorrect [country code listed for registration](https://github.com/WhisperSystems/Signal-iOS/blob/c7ee430790898110f575adbda5bf34000b32242f/Signal/src/view%20controllers/CountryCodeViewController.h)

### Steps to reproduce
- Install Signal
- See registration screen
- Tap Country Code to edit country code
- Choose country listed, with incorrect country code of +0 i.e. Kosovo should be + 383
- Typing in entire phone number (country code without +) yields "Registration Error"

**Actual result:** Country code is listed as +0 and can not edit
**Expected result:** Country code to be editable manually or have correct code listed

### Screenshots
 
![screen shot 2017-03-01 at 4 55 20 pm](https://cloud.githubusercontent.com/assets/4898366/23488189/e6583726-fe9f-11e6-896a-1688889c5900.png)

# Summary of misused +0 as country code

Country | Country Code | Comments
------- | -------- | --------
Antarctica |  +672
Bouvet Island | +55
Canary Islands | +34
Ceuta & Melilla | +34
Diego Garcia | +246
French Southern Territories | +262
Kosovo | +383
Pitcairn Islands | +64
So. Georgia & So. Sandwich | +500
U.S Outlying Islands | +1

### Device info

**Device**: iPhone 5C
**iOS version**: 10.2.1
**Signal version:** 2.7.1

michaelkirk:
Fixed in the next release.

PR #1822, opened 2017-03-02
Fix calling codes in registration flow.

charlesmchen:
Fixes: https://github.com/WhisperSystems/Signal-iOS/issues/1821

PTAL @michaelkirk

### DIFF
--- a/Signal/src/view controllers/CountryCodeViewController.m
+++ b/Signal/src/view controllers/CountryCodeViewController.m
@@ -18,8 +18,67 @@ static NSString *const kUnwindToCountryCodeWasSelectedSegue = @"UnwindToCountryC
 - (void)viewDidLoad {
     [super viewDidLoad];
     [self.navigationController.navigationBar setTranslucent:NO];
-    _countryCodes = [PhoneNumberUtil countryCodesForSearchTerm:nil];
+    _countryCodes = [self countryCodesForSearchTerm:nil];
     self.title    = NSLocalizedString(@"COUNTRYCODE_SELECT_TITLE", @"");
+}
+
+- (NSArray *)countryCodesForSearchTerm:(NSString *)searchTerm {
+    NSMutableArray *result = [NSMutableArray array];
+    for (NSString *countryCode in [PhoneNumberUtil countryCodesForSearchTerm:nil]) {
+        NSString *callingCode = [self callingCodeFromCountryCode:countryCode];
+        if (callingCode != nil &&
+            ![callingCode isEqualToString:@"+0"]) {
+            [result addObject:countryCode];
+        }
+    }
+    return result;
+}
+
+- (NSString *)callingCodeFromCountryCode:(NSString *)countryCode {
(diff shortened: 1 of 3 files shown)


Signal/src/view controllers/CountryCodeViewController.m
```diff
@@ -18,10 +18,69 @@ @implementation CountryCodeViewController
...
+            ![callingCode isEqualToString:@"+0"]) {
+            [result addObject:countryCode];
+        }
+    }
+    return result;
+}
+
+- (NSString *)callingCodeFromCountryCode:(NSString *)countryCode {
+    NSString *callingCode = [PhoneNumberUtil callingCodeFromCountryCode:countryCode];
+    if ([countryCode isEqualToString:@"AQ"]) {
```
charlesmchen: `PhoneNumberUtil` uses a third-party library `NBPhoneNumberUtil`, which is returning bad data for some obscure country codes.

Signal/src/view controllers/CountryCodeViewController.m
```diff
@@ -18,10 +18,69 @@ @implementation CountryCodeViewController
...
     self.title    = NSLocalizedString(@"COUNTRYCODE_SELECT_TITLE", @"");
 }
 
+- (NSArray *)countryCodesForSearchTerm:(NSString *)searchTerm {
+    NSMutableArray *result = [NSMutableArray array];
+    for (NSString *countryCode in [PhoneNumberUtil countryCodesForSearchTerm:nil]) {
+        NSString *callingCode = [self callingCodeFromCountryCode:countryCode];
+        if (callingCode != nil &&
+            ![callingCode isEqualToString:@"+0"]) {
+            [result addObject:countryCode];
```
charlesmchen: We fix the countries that we can, but we also do this filtering for countries that we can't fix.  This future-proofs us as well.

Signal/src/view controllers/CountryCodeViewController.m
```diff
@@ -18,10 +18,69 @@ @implementation CountryCodeViewController
...
+    } else if ([countryCode isEqualToString:@"IC"]) {
+        // Canary Islands
+        callingCode = @"+34";
+    } else if ([countryCode isEqualToString:@"EA"]) {
+        // Ceuta & Melilla
+        callingCode = @"+34";
+    } else if ([countryCode isEqualToString:@"CP"]) {
+        // Clipperton Island
+        //
+        // This country code should be filtered.
```
michaelkirk: Can you add a comment as to why it should be filtered, and for Heard & MCDonald islands?

charlesmchen: Done. Essentially, I can't find reliable info on their calling codes.  I've added the most probable value for the Heard & McDonald Islands, but I'm going to filter Clipperton Island which appears to be uninhabited and not to have a valid calling code.

michaelkirk:
rebased and merged with https://github.com/WhisperSystems/Signal-iOS/commit/4983853dd05b6bbbef8b1325aebb9fa0ce477343

PR #1834, opened 2017-03-10
Charlesmchen/paste

charlesmchen:
* Gather attachment-related logic in `SignalAttachment` class.
   * We're now more consistent about applying attachment validation & conversion logic.
   * I also fixed some bugs in this logic.
* Update `MessagesViewController` to use `SignalAttachment`.
* Update `MessagesViewController` to support paste of attachments.
* Add `AttachmentApprovalViewController`.

What's not included:

* Previously we were mostly failing silently (and sometimes logging) errors around attachments. `SignalAttachment` now gathers and normalizes our validation logic and captures errors using an enum.  It'd be nice to surface these errors (using alerts?) to the user, but this PR is already so large that I'm going to do that in a follow-up PR.

PTAL @michaelkirk


Signal/Images.xcassets/file-icon-large.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "file-icon-large@1x.png",
```
charlesmchen: For large "hero" image assets that we want to resize based on device size, I like to use a single oversized asset and scale using UIImageView.

charlesmchen:
Screenshots of the "attachment approval" dialog:

screen shot 2017-03-10 at 3 36 23 pm
screen shot 2017-03-10 at 3 21 42 pm
screen shot 2017-03-10 at 3 20 06 pm
screen shot 2017-03-10 at 2 50 18 pm

This dialog is used to approve pasted attachments.  The dialog shows images (this is important since they may have been converted) but just shows file size and type for other file types.

Signal/src/view controllers/AttachmentApprovalViewController.swift
```diff
@@ -0,0 +1,217 @@
...
+        if image != nil {
+            let imageView = UIImageView(image:image)
+            imageView.layer.minificationFilter = kCAFilterTrilinear;
+            imageView.layer.magnificationFilter = kCAFilterTrilinear;
+            imageView.contentMode = .scaleAspectFit
+            attachmentPreviewView.addSubview(imageView)
+            imageView.autoPinWidthToSuperview()
+            imageView.autoPinHeightToSuperview()
+        } else {
+            createGenericPreview(attachmentPreviewView:attachmentPreviewView)
```
charlesmchen: If we can't load the image, show the "generic" preview.  This should never happen (thanks to the validation logic in `SignalAttachment`) but we want to fail gracefully if it does.

Signal/src/view controllers/MessagesViewController.h
```diff
@@ -14,6 +14,24 @@
...
+
+#pragma mark -
+
+@interface OWSMessagesInputToolbar : JSQMessagesInputToolbar
+
+@end
+
+#pragma mark -
+
 @interface MessagesViewController : JSQMessagesViewController <UIImagePickerControllerDelegate,
```
charlesmchen: To support paste, I had to use a custom text view.  `JSQMessagesViewController` doesn't make it easy to use a custom text view; I had to subclass a few JSQ classes AND clone a few JSQ .xibs, etc.

Signal/src/view controllers/MessagesViewController.m
```diff
@@ -1211,7 +1316,11 @@ - (void)showConversationSettings
         DDLogDebug(@"%@ Ignoring request to show conversation settings, since user left group", self.tag);
         return;
     }
-    [self performSegueWithIdentifier:OWSMessagesViewControllerSeguePushConversationSettings sender:self];
+    
+    OWSConversationSettingsTableViewController *settingsVC = [[UIStoryboard storyboardWithName:AppDelegateStoryboardMain bundle:NULL]
+                                                              instantiateViewControllerWithIdentifier:@"OWSConversationSettingsTableViewController"];
+    [settingsVC configureWithThread:self.thread];
+    [self.navigationController pushViewController:settingsVC animated:YES];
```
charlesmchen: To use the customized text view, I had to switch to doing programmatic segues between these views. =/

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,463 @@
...
+// This class gathers that logic.  It offers factory methods
+// for attachments that do the necessary work. 
+//
+// The return value for the factory methods will be nil if the input is nil.
+//
+// [SignalAttachment hasError] will be true for non-valid attachments.
+//
+// TODO: Perhaps do conversion off the main thread?
+// TODO: Show error on error.
+// TODO: Show progress on upload.
```
charlesmchen: `SignalAttachment` raises all sorts of interesting questions about how we validate, convert and otherwise handle attachments.  I've tried to capture these questions as TODOs here and below in this file.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,463 @@
...
+            Logger.verbose("\(TAG) Compressing attachment as image/jpeg")
+            return compressImageAsJPEG(image : image, attachment : attachment)
+        }
+    }
+
+    // If the proposed attachment already is a JPEG, and already conforms to the 
+    // file size and content size limits, don't recompress it.
+    //
+    // TODO: Should non-JPEGs always be converted to JPEG?
+    private class func isInputImageValidOutputImage(image: UIImage?, imageData: Data?, dataUTI: String!) -> Bool {
```
charlesmchen: One of the important fixes in this refactor is that we're now much more careful to avoid converting/re-encoding images as JPEG if possible. JPEG is a lossy format and needlessly re-encoding significantly degrades image quality.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,463 @@
...
+        assert(attachment.error == nil)
+
+        var imageUploadQuality = Environment.preferences().imageUploadQuality()
+
+        while true {
+            let maxSize = maxSizeForImage(image: image, imageUploadQuality:imageUploadQuality)
+            var dstImage: UIImage! = image
+            if image.size.width > maxSize ||
+                image.size.height > maxSize {
+                dstImage = imageScaled(image, toMaxSize: maxSize)
```
charlesmchen: Another place we avoid re-encoding if possible.

michaelkirk:
Despite all of my nits, this looks really great, and I'm excited to include it.

Signal/Images.xcassets/file-icon-large.imageset/Contents.json
charlesmchen: > anything with thin lines is likely better rasterized externally.

Sure - especially text.  But that doesn't apply to this image, right?

Signal/src/view controllers/AttachmentApprovalViewController.swift
charlesmchen: Yup, there's a lot of aspects of this work that prepare for that.

Signal/src/view controllers/MessagesViewController.m
charlesmchen: Done.

Signal/src/UIColor+OWS.m
```diff
@@ -107,4 +107,21 @@ + (UIColor *)backgroundColorForContact:(NSString *)contactIdentifier
...
 
++ (UIColor *)colorWithRGBHex:(unsigned long)value
+{
+    CGFloat red = ((value >> 16) & 0xff) / 255.f;
+    CGFloat green = ((value >> 8) & 0xff) / 255.f;
+    CGFloat blue = ((value >> 0) & 0xff) / 255.f;
+    return [UIColor colorWithRed:red green:green blue:blue alpha:1.f];
+}
+
++ (UIColor *)colorWithARGBHex:(unsigned long)value
```
charlesmchen: I prefer ARGB because:

* This method makes very specific assumptions about the channel ordering in the hex representation (namely that the alpha channel will be in the "highest" position of the ) and...
* ARGB is the most common ordering for this kind of hex representation for colors, right?

Signal/src/view controllers/AttachmentApprovalViewController.swift
```diff
@@ -0,0 +1,217 @@
...
+
+import Foundation
+
+class AttachmentApprovalViewController: UIViewController {
+
+    let TAG = "[AttachmentApprovalViewController]"
+
+    // MARK: Properties
+
+    let attachment: SignalAttachment!
```
charlesmchen: Done.

Signal/src/view controllers/AttachmentApprovalViewController.swift
```diff
@@ -0,0 +1,217 @@
...
+
+    var successCompletion : (() -> Void)?
+
+    // MARK: Initializers
+
+    @available(*, deprecated, message:"use attachment: constructor instead.")
```
charlesmchen: You're right. Done.

Signal/src/view controllers/AttachmentApprovalViewController.swift
```diff
@@ -0,0 +1,217 @@
...
+        self.navigationItem.title = NSLocalizedString("ATTACHMENT_APPROVAL_DIALOG_TITLE",
+                                                      comment: "Title for the 'attachment approval' dialog.")
+
+        createViews()
+    }
+
+    // MARK: - Create Views
+
+    private func createViews() {
+        let previewTopMargin = 30 as CGFloat
```
charlesmchen: Done.

Signal/src/view controllers/AttachmentApprovalViewController.swift
```diff
@@ -0,0 +1,217 @@
...
+        imageView.autoHCenterInSuperview()
+        imageView.autoPinEdge(toSuperviewEdge:.top)
+        imageView.autoSetDimension(.width, toSize:imageSize)
+        imageView.autoSetDimension(.height, toSize:imageSize)
+
+        var lastView: UIView = imageView
+
+        let labelFont = UIFont.ows_regularFont(withSize:ScaleFromIPhone5To7Plus(18, 24))
+
+        if attachment.fileExtension() != nil {
```
charlesmchen: Done.

Signal/src/view controllers/AttachmentApprovalViewController.swift
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    func cancelPressed(sender: UIButton) {
+        dismiss(animated: true, completion:nil)
+    }
+
+    func sendPressed(sender: UIButton) {
+        let successCompletion = self.successCompletion
+        dismiss(animated: true, completion: {
+            if successCompletion != nil {
```
charlesmchen: Done.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+// TODO: Perhaps do conversion off the main thread?
+// TODO: Show error on error.
+// TODO: Show progress on upload.
+class SignalAttachment: NSObject {
+
+    static let TAG = "[SignalAttachment]"
+
+    // MARK: Properties
+
+    let data: Data!
```
charlesmchen: Done.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+
+    // Attachment types are identified using UTIs.
+    //
+    // See: https://developer.apple.com/library/content/documentation/Miscellaneous/Reference/UTIRef/Articles/System-DeclaredUniformTypeIdentifiers.html
+    let dataUTI: String!
```
charlesmchen: Done.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+
+    /**
+     * Media Size constraints from Signal-Android
+     * (org/thoughtcrime/securesms/mms/PushMediaConstraints.java)
+     */
+    static let kMaxFileSizeGif = 5 * 1024 * 1024
+    static let kMaxFileSizeImage = 420 * 1024
+    static let kMaxFileSizeVideo = 100 * 1024 * 1024
+    static let kMaxFileSizeAudio = 100 * 1024 * 1024
+    // TODO: What should the max file size on "other" attachments be?
```
charlesmchen: You know, I went to https://github.com/WhisperSystems/Signal-Android/blob/master/src/org/thoughtcrime/securesms/mms/PushMediaConstraints.java and it turns out these values are out of sync with Android.  I've aligned them.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+    static let kMaxFileSizeGeneric = 100 * 1024 * 1024
+
+    // MARK: Constructor
+
+    // This method should not be called directly; use the factory
+    // methods instead.
+    internal required init(data: Data!, dataUTI: String!) {
```
charlesmchen: Done.

charlesmchen: So when is it appropriate to unwrap a parameter?

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+        self.data = data
+        self.dataUTI = dataUTI
+        super.init()
+    }
+
+    public func hasError() -> Bool {
```
charlesmchen: Done.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+        let mimeType = UTTypeCopyPreferredTagWithClass(dataUTI as CFString, kUTTagClassMIMEType)
+        guard mimeType != nil else {
+            return nil
+        }
+        return mimeType?.takeRetainedValue() as? String
+    }
+
+    // Returns the file extension for this attachment or nil if no file extension
+    // can be identified.
+    public func fileExtension() -> String? {
```
charlesmchen: Done.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+        return fileExtension?.takeRetainedValue() as? String
+    }
+
+    // Returns the set of UTIs that correspond to valid _input_ image formats
+    // for Signal attachments.
+    //
+    // Image attachments may be converted to another image format before 
+    // being uploaded.
+    //
+    // TODO: We need to finalize which formats we support.
```
charlesmchen: I've taken a run at this and `SignalAttachment` now uses `MIMETypeUtil`.

https://github.com/WhisperSystems/SignalServiceKit/pull/125

Still, there's a lot to think over here.  It's worth considering what a "supported" file type means, once we allow users to attach arbitrary files.

For example, why would we impose our current low limits on images (6 MB) but allow users to send 100 MB zip files (which might contain just a single image file)?  Especially in an age where the pictures from phones' cameras are often larger than 6MB?  I'm hoping we can use a single max attachment file size for all attachments - or only apply an exception that is _more generous_, not less, for certain attachment types.

If the app is only concerned with how to _display_ attachments, shouldn't that depend on the capabilities of the current phone, rather than being hard-coded?  Also, neither a UTI nor a MIME type is fine-grained enough to determine whether or not the current phone can play a given .mp4, for example.  It depends on the compression used, the frame rate, the content size, etc.  Maybe that doesn't really matter. http://stackoverflow.com/questions/2719214/what-are-the-video-formats-supported-by-mpmovieplayercontroller-on-the-ipad

Signal/Images.xcassets/file-icon-large.imageset/Contents.json
michaelkirk: I don't know? Assuming you checked in on all the resolutions we're using, I don't have a problem with it.

Signal/src/UIColor+OWS.m
michaelkirk: > ARGB is the most common ordering for this kind of hex representation for colors, right?

I certainly didn't think so, else I wouldn't have brought it up.

See also: https://en.wikipedia.org/wiki/RGBA_color_space

michaelkirk: > This method makes very specific assumptions about the channel ordering in the hex representation (namely that the alpha channel will be in the "highest" position of the ) and...

Right, obviously that would have to change with the name. But if you feel this is the "normal" way to do it, go for it.

Signal/src/view controllers/SignalAttachment.swift
michaelkirk: TBH I can't think of a real case when you'd ever want to do this. It would have to be a method that takes optionals but which you somehow *know* will never be nil. In any case I can think of it's better to force that determination on the caller who is more proximate to the actual value.

Signal/src/view controllers/SignalAttachment.swift
michaelkirk: > For example, why would we impose our current low limits on images (6 MB) but allow users to send 100 MB zip files (which might contain just a single image file)? 

I think we *do* have to consider normal usage patterns, not just some abstract rationale. The overwhelming majority of attachments being sent will be photos. If we can deliver those at a quality that is "good enough" for almost everyone, while saving a substantial portion of service bandwidth, that's a win.

Now, if 6MB is "good enough" is a good question...

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+    // can be found.
+    //
+    // NOTE: The attachment returned by this method may not be valid.
+    //       Check the attachment's error property.
+    public class func attachmentFromPasteboard() -> SignalAttachment? {
+        guard UIPasteboard.general.numberOfItems == 1 else {
+            // Ignore pasteboard if it contains multiple items.
+            //
+            // TODO: Should we try to use the first?
+            return nil
```
charlesmchen: Done.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+        let pasteboardUTISet = Set(UIPasteboard.general.types)
+        for dataUTI in inputImageUTISet() {
+            if pasteboardUTISet.contains(dataUTI) {
+                let imageData = UIPasteboard.general.data(forPasteboardType:dataUTI)
+                return imageAttachment(withData : imageData, dataUTI : dataUTI)
+            }
+        }
+        for dataUTI in videoUTISet() {
+            if pasteboardUTISet.contains(dataUTI) {
+                let imageData = UIPasteboard.general.data(forPasteboardType:dataUTI)
```
charlesmchen: > nit: This should probably be let videoData = ... same with let audioData below.

Just going to use `data`.  

> Or maybe there's some way to dedupe this...

We need to find the "best" dataUTI and its corresponding data.  Deduplicating that logic would mean an optional tuple, and unpacking an optional tuple would be more verbose, unless I'm missing something.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+            Logger.verbose("\(TAG) Compressing attachment as image/jpeg")
+            return compressImageAsJPEG(image : image, attachment : attachment)
+        }
+    }
+
+    private class func defaultImageUploadQuality() -> TSImageQuality {
+        // always return average image quality
+        //
+        // TODO: Why are we using this value?  Why not default to .uncropped?
+        return .medium
```
charlesmchen: Great.  

While we're at it, I'm going to make some other changes to image handling logic, namely not converting non-JPEG images to JPEG.  That seems to be a vestige of an earlier time when Signal "wanted" to attach only tiny JPEGs.  Now that we a) allow much larger attachments (the Android image attachment max file size has changed from 420kb to 6mb) b) prefer not to resize/recompress images, and c) are moving towards arbitrary attachments, it only makes sense to allow attaching PNGs (for example) without converting them to JPEG.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+            }
+        }
+        return false
+    }
+
+    // Factory method for an image attachment.
+    //
+    // NOTE: The attachment returned by this method may nil or not be valid.
+    //       Check the attachment's error property.
+    public class func imageAttachment(withImage image: UIImage?, dataUTI: String!) -> SignalAttachment! {
```
charlesmchen: Done.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+
+        // Make a placeholder attachment on which to hang errors if necessary.
+        let attachment = SignalAttachment(data : Data(), dataUTI: dataUTI)
+        attachment.image = image
+
+        Logger.verbose("\(TAG) Writing \(attachment.mimeType()) as image/jpeg")
+        return compressImageAsJPEG(image : image, attachment : attachment)
+    }
+
+    private class func compressImageAsJPEG(image: UIImage!, attachment: SignalAttachment!) -> SignalAttachment! {
```
charlesmchen: Done.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+        assert(dataUTI.characters.count > 0)
+
+        assert(data != nil)
+        guard let data = data else {
+            return nil
+        }
+
+        let attachment = SignalAttachment(data : data, dataUTI: dataUTI)
+
+        if validUTISet != nil {
```
charlesmchen: Done.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -52,6 +52,21 @@
 /* No comment provided by engineer. */
 "ATTACHMENT" = "Attachment";
 
+/* Label for 'cancel' button in the 'attachment approval' dialog. */
```

michaelkirk:
Cool - If you're going to change what types we send to include PNGs etc, you'll have to check that desktop and android do the right thing with each of them.

Signal/src/view controllers/MessagesViewController.m
```diff
@@ -2408,7 +2470,51 @@ - (void)navigationTitleTapped:(UIGestureRecognizer *)gestureRecognizer {
...
+#pragma mark - OWSTextViewPasteDelegate
+
+- (void)didPasteAttachment:(SignalAttachment * _Nullable)attachment {
+    DDLogError(@"%@ %s",
+               self.tag,
+               __PRETTY_FUNCTION__);
+    
+    if (attachment == nil ||
+        [attachment hasError]) {
+        // TODO: Add UI.
```
charlesmchen: Done.

charlesmchen:
> Cool - If you're going to change what types we send to include PNGs etc, you'll have to check that desktop and android do the right thing with each of them.

You're right. I'm following up in the standup group.

Signal/src/UIColor+OWS.m
charlesmchen: Huh.  Well, having done some googling I see you're right there is no clearly right answer.  I've just always seen hex colors use ARGB, but I guess that was anecdotal.  I'm going to remove this method since we're not even using it yet.

charlesmchen:
> Cool - If you're going to change what types we send to include PNGs etc, you'll have to check that desktop and android do the right thing with each of them.

It turns out this isn't safe; the Android client can't handle many of the formats supported by `MIMETypesUtil`.  So in addition to needing to convert some images format, we can't (yet) allow users to paste some kinds of audio and video attachments.  Therefore, I've gone back to only emitting the file formats that the iOS client was already emitting - plus PNG.   

I've also added a `allowArbitraryAttachments` constant to allow us to prepare for the future.  

And I've been able to resolve a bunch of the TODOs in this PR, which is nice.

Signal/src/view controllers/AttachmentApprovalViewController.swift
```diff
@@ -0,0 +1,215 @@
...
+
+    @available(*, unavailable, message:"use attachment: constructor instead.")
+    required init?(coder aDecoder: NSCoder) {
+        self.attachment = SignalAttachment.genericAttachment(data: nil,
+                                                             dataUTI: kUTTypeContent as String)
+        super.init(coder: aDecoder)
+        assert(false)
+    }
+
+    required init(attachment: SignalAttachment!, successCompletion : @escaping () -> Void) {
```
michaelkirk: nit: remove the `!` from `SignalAttachment!`

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,501 @@
...
+
+            Logger.verbose("\(TAG) Compressing attachment as image/jpeg")
+            return compressImageAsJPEG(image : image, attachment : attachment)
+        }
+    }
+
+    private class func defaultImageUploadQuality() -> TSImageQuality {
+        // Always default to the highest possible image quality and
+        // not converting the image if possible.
+        return .uncropped
```
michaelkirk: Just to double back, we agreed to change this back so as to conservatively match prod for now, right?

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,501 @@
...
+        }
+    }
+
+    // MARK: Video Attachments
+
+    // Factory method for video attachments.
+    //
+    // NOTE: The attachment returned by this method may not be valid.
+    //       Check the attachment's error property.
+    public class func videoAttachment(data: Data?, dataUTI: String) -> SignalAttachment {
```
michaelkirk: nit: looks like we're already checking if-nil at the caller, so we can make these `data` parameters non-optional.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,501 @@
...
+        }
+        // If pasteboard contains multiple items, use only the first.
+        let itemSet = IndexSet(integer:0)
+        guard let pasteboardUTITypes = UIPasteboard.general.types(forItemSet:itemSet) else {
+            return nil
+        }
+        let pasteboardUTISet = Set<String>(pasteboardUTITypes[0])
+        for dataUTI in inputImageUTISet {
+            if pasteboardUTISet.contains(dataUTI) {
+                let data = UIPasteboard.general.data(forPasteboardType:dataUTI, inItemSet:itemSet)![0] as! Data
```
michaelkirk: Another nit: 

Whenever I see a `!` I get nervous. Is this safe? 

Can we say that it's always true that whenever a `pasteboardUTISet.contains(dataUTI)` that `UIPasteboard.general.data(forPasteboardType:dataUTI, inItemSet:itemSet)[0]` will be a non-nil Data?

I mean, it seems like a reasonable assumption, but since this method returns an optional anyway, I'd be more comfortable writing something conservative like:

```
guard let data = UIPasteboard.general.data(forPasteboardType:dataUTI, inItemSet:itemSet)?.firstObject as Data else { 
  Logger.warn("uh oh")
  return nil
}
return imageAttachment(data: data, dataUTI: dataUTI)
```

And avoid any chance of crashing.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -0,0 +1,477 @@
...
+        return newAttachment(withData : data,
+                             dataUTI : dataUTI,
+                             validUTISet : nil,
+                             maxFileSize : kMaxFileSizeGeneric)
+    }
+
+    // MARK: Helper Methods
+
+    private class func newAttachment(withData data: Data?,
+                                     dataUTI: String!,
```
charlesmchen: Done.

Signal/src/view controllers/AttachmentApprovalViewController.swift
charlesmchen: Done.

Signal/src/view controllers/SignalAttachment.swift
charlesmchen: Done. This got verbose so I added a method to do the unpacking.

Signal/src/view controllers/SignalAttachment.swift
charlesmchen: That's right.

Signal/src/view controllers/SignalAttachment.swift
charlesmchen: I'd like to leave them optional.  I hope to use them in more places soon (e.g. the sharing extension) and it'll be nice to consolidate how nils are converted into errors surfaced to the user.

Issue #1832, opened 2017-03-10
When Registering, Search Country Codes by Country Code


charlesmchen:
Country Code search currently only matches country name - not the country code itself.  You should be able to search for +44 (UK) by typing 44.

NicolasCGN:
In 2.8.0.5 (also 2.8.0.4) is not possible to search the country by entering the first characters of its name.

charlesmchen:
Thanks @NicolasCGN .  This will be fixed in a forthcoming release.

charlesmchen:
@NicolasCGN This is fixed and will ship in a forthcoming release.

NicolasCGN:
Hmmmm. I'm still not able to filter countries in beta 2.8.3.0. No matter if I write the beginning of the country's name or the telephone prefix.

michaelkirk:
This will be fixed in 2.9.0

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/a264d9aa9b4d37655dc4648919c971bb3a068a61

PR #1835, opened 2017-03-10
Fix filtering of country codes in registration flow.

charlesmchen:
* Filtering wasn't working; it has been fixed.
* I've added support for arbitrary partial matches and matching based on country code and calling code.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/view controllers/CountryCodeViewController.m
+++ b/Signal/src/view controllers/CountryCodeViewController.m
@@ -22,109 +22,9 @@ static NSString *const kUnwindToCountryCodeWasSelectedSegue = @"UnwindToCountryC
 - (void)viewDidLoad {
     [super viewDidLoad];
     [self.navigationController.navigationBar setTranslucent:NO];
-    _countryCodes = [self countryCodesForSearchTerm:nil];
+    _countryCodes = [PhoneNumberUtil countryCodesForSearchTerm:nil];
     self.title    = NSLocalizedString(@"COUNTRYCODE_SELECT_TITLE", @"");
     self.searchBar.delegate = self;
-}
-
-- (NSArray *)countryCodesForSearchTerm:(NSString *)searchTerm {
-    NSMutableArray *result = [NSMutableArray array];
-    for (NSString *countryCode in [PhoneNumberUtil countryCodesForSearchTerm:searchTerm]) {
-        NSString *callingCode = [self callingCodeFromCountryCode:countryCode];
-        if (callingCode != nil &&
-            ![callingCode isEqualToString:@"+0"]) {
-            [result addObject:countryCode];
-        }
-    }
-    
-    if (searchTerm.length > 0) {
-        NSArray *allCountryCodes = [self countryCodesForSearchTerm:nil];
-        // We want to ignore + when using a search term like "+44".
-        // We also want to ignore punctuation, etc.
-        // To keep things simple, we just remove everything except letters and numbers.
-        NSCharacterSet *charactersToRemove = [[NSMutableCharacterSet alphanumericCharacterSet] invertedSet];
-        NSString *simplifiedSearchTerm = [[searchTerm componentsSeparatedByCharactersInSet:charactersToRemove]
-                                          componentsJoinedByString:@"" ];
-        for (NSString *split in [simplifiedSearchTerm componentsSeparatedByCharactersInSet:[NSCharacterSet whitespaceCharacterSet]]) {
-            NSString *simplifiedSplit = split.lowercaseString;
-            OWSAssert(simplifiedSplit.length > 0);
-            for (NSString *countryCode in allCountryCodes) {
-                if ([countryCode.lowercaseString rangeOfString:simplifiedSplit].location != NSNotFound) {
-                    [result addObject:countryCode];
-                    continue;
-                }
-                NSString *countryName = [PhoneNumberUtil countryNameFromCountryCode:countryCode];
-                OWSAssert(countryName.length > 0);
-                if ([countryName.lowercaseString rangeOfString:simplifiedSplit].location != NSNotFound) {
-                    [result addObject:countryCode];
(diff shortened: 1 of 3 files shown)


charlesmchen:
### (Partial) matching based on calling code.

screen shot 2017-03-10 at 6 46 53 pm

### As before, matching based on prefix.

screen shot 2017-03-10 at 6 46 58 pm

### Non-prefix matches.

screen shot 2017-03-10 at 6 47 03 pm

### Matching based on country code.

screen shot 2017-03-10 at 6 49 50 pm

charlesmchen:
Fixes: https://github.com/WhisperSystems/Signal-iOS/issues/1832

charlesmchen:
Fixes: https://github.com/WhisperSystems/Signal-iOS/issues/1832

Signal/src/view controllers/CountryCodeViewController.m
```diff
@@ -20,17 +24,61 @@ - (void)viewDidLoad {
     [self.navigationController.navigationBar setTranslucent:NO];
     _countryCodes = [self countryCodesForSearchTerm:nil];
     self.title    = NSLocalizedString(@"COUNTRYCODE_SELECT_TITLE", @"");
+    self.searchBar.delegate = self;
 }
 
 - (NSArray *)countryCodesForSearchTerm:(NSString *)searchTerm {
     NSMutableArray *result = [NSMutableArray array];
-    for (NSString *countryCode in [PhoneNumberUtil countryCodesForSearchTerm:nil]) {
+    for (NSString *countryCode in [PhoneNumberUtil countryCodesForSearchTerm:searchTerm]) {
```
michaelkirk: nit: I feel like it would be easier to test this logic and catch future regressions like this one if the logic were put into the PhoneNumberUtil (or some other non VC class), and unit test at least the happy path .

Signal/src/view controllers/CountryCodeViewController.m
```diff
@@ -20,17 +24,61 @@ - (void)viewDidLoad {
...
+    
+    if (searchTerm.length > 0) {
+        NSArray *allCountryCodes = [self countryCodesForSearchTerm:nil];
+        // We want to ignore + when using a search term like "+44".
+        // We also want to ignore punctuation, etc.
+        // To keep things simple, we just remove everything except letters and numbers.
+        NSCharacterSet *charactersToRemove = [[NSMutableCharacterSet alphanumericCharacterSet] invertedSet];
+        NSString *simplifiedSearchTerm = [[searchTerm componentsSeparatedByCharactersInSet:charactersToRemove]
+                                          componentsJoinedByString:@"" ];
+        for (NSString *split in [simplifiedSearchTerm componentsSeparatedByCharactersInSet:[NSCharacterSet whitespaceCharacterSet]]) {
```
michaelkirk: I *think* this search won't work if you search multiple words (e.g "United States").

e.g. http://swift.sandbox.bluemix.net/#/repl/58c713edfa2d56665a088d8a

Signal/src/view controllers/CountryCodeViewController.m
charlesmchen: Done. See: https://github.com/WhisperSystems/SignalServiceKit/pull/128

Note that PhoneNumberUtil resides in SSK but PhoneNumberTest resides in Signal-iOS.

Signal/src/view controllers/CountryCodeViewController.m
charlesmchen: I've switched to using `[ContactsManagerProtocol name: matchesQuery:]`.  In consolidating this logic, I also improved handling of whitespace-only searches and reverted to prefix-only searches.

charlesmchen:
PTAL @michaelkirk

Signal/src/view controllers/CountryCodeViewController.m
michaelkirk: Am I confused or is this PR out of date?  

Now that the search logic is in SSK, it seems like we shouldn't need to do any of this here.

Signal/test/phone/PhoneNumberTest.m
```diff
@@ -75,5 +81,78 @@ -(void) testTranslateCursorPosition {
...
+    XCTAssertEqualObjects([PhoneNumberUtil countryNameFromCountryCode:@"ZZZ"], @"ZZZ");
+    testThrows([PhoneNumberUtil countryNameFromCountryCode:nil]);
+}
+
+- (void)testCountryCodesForSearchTerm
+{
+// PhoneNumberUtil needs a valid OWSContactsManager in the TextSecureKitEnv,
+// but we want to avoid constructing the entire app apparatus so we pass nil
+// for the other parameters of the environemtn and disable nonnull warnings.
+#pragma clang diagnostic push
```
michaelkirk: Take it or leave it:

This is a primary driver for my preference for dependency injection vs. accessing global singletons all over the place.

If you move this test to SSK (where it probably belongs) you can take advantage of a pre-existing menagerie of fakes - e.g. https://github.com/WhisperSystems/SignalServiceKit/blob/master/tests/TestSupport/Fakes/OWSUnitTestEnvironment.m

charlesmchen: We're definitely paying the price here for the convenience of singletons over DI. 

And we should eventually move this test.  SSK doesn't seem to have any tests.  I'm worried that if I'll move it, we'll be less likely to run these tests. We already break the existing tests without noticing occasionally.

michaelkirk: SSK has some tests: https://github.com/WhisperSystems/SignalServiceKit/tree/master/tests

They are run in the context of https://github.com/WhisperSystems/SignalServiceKit/tree/master/Example/TSKitiOSTestApp

Though CI is currently failing, I try to keep an eye on it. You should too.

Issue #1837, opened 2017-03-10
Update conversation view header when group name changes.


charlesmchen:

charlesmchen:
This issue is fixed and will ship in the next release.

PR #1839, opened 2017-03-11
Charlesmchen/group name updates

charlesmchen:
Fixes: https://github.com/WhisperSystems/Signal-iOS/issues/1837

PTAL @michaelkirk

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -2079,8 +2079,18 @@ typedef enum : NSUInteger {
     return _editingDatabaseConnection;
 }
 
-
 - (void)yapDatabaseModified:(NSNotification *)notification {
+    // Currently, we update thread and message state every time
+    // the database is modified.  That doesn't seem optimal, but
+    // in practice it's efficient enough.
+    //
+    // 
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [self updateContents];
+    });
+}
+
+- (void)updateContents {
     [self updateBackButtonAsync];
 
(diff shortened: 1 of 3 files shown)


Signal/src/view controllers/MessagesViewController.m
```diff
@@ -2018,8 +2018,15 @@ - (YapDatabaseConnection *)editingDatabaseConnection {
...
 
-
 - (void)yapDatabaseModified:(NSNotification *)notification {
+    // Currently, we update thread and message state every time
+    // the database is modified.  That doesn't seem optimal, but
+    // in practice it's efficient enough.
+
+    // We need to `beginLongLivedReadTransaction` before we update our
+    // models in order to jump to the most recent commit.
+    NSArray *notifications = [self.uiDatabaseConnection beginLongLivedReadTransaction];
```
charlesmchen: This is the first key change - we need to ensure we're updating the model from the latest database state, so we need to call `beginLongLivedReadTransaction` _before_ we update our thread.

Signal/src/view controllers/MessagesViewController.m
```diff
@@ -2030,10 +2037,9 @@ - (void)yapDatabaseModified:(NSNotification *)notification {
               self.thread = [TSGroupThread threadWithGroupModel:gThread.groupModel transaction:transaction];
           }
         }];
+        [self setNavigationTitle];
```
charlesmchen: This is the second key change - the group name may have changed so we need to ensure that view and model state are aligned.

michaelkirk:
LGTM

Issue #1708, opened 2017-02-06
debug log stops recording


michaelkirk:
### Bug description

New entries stop recording to debug log. I've only seen this once but wanted to start taking notes in case it occurs again.

### Steps to reproduce

After using the app for days

go to settings > advanced > submit debug log

**Actual result:** A couple of old logs, but no recent logs

**Expected result:** The last three days of logs.

### Device info

**Device**: iPhone 5.3
**iOS version**: 10.2.1
**Signal version:** 2.6.13.1, 2.7.0.1

### Link to debug log

https://gist.github.com/anonymous/1da6ecc5b9362962cb8b5cb120bc99e7

Note the long gap before the final day. Initially the 2017-02-06 21-03.log was not included in the generated debug log, but after toggling logging off and then generating a new log, it was included.

Maybe related to:

870fb96
907e122

riyapenn:
Toggling the "Enable Debug Log" re-triggers logging information.  Will post more logs to this comment via edits
User 1: https://gist.github.com/anonymous/5bcac587c4ce5fe90cfdb9fbc36b8628

riyapenn:
Is becoming an issue where I can not get useful logs (for other issues) from users without asking them to re-enable, then send in a log after experiencing the issue again.

charlesmchen:
Fixed this in https://github.com/WhisperSystems/Signal-iOS/pull/1841

Sorry Riya - Debug Logs are broken in production right now.  It might even be worth doing a hotfix to 2.8.1.0 that just takes this change.

PR #1841, opened 2017-03-13
Convert "is logging enabled" methods to class methods so that they ca…

charlesmchen:
…n safely be used before Environment has been initialized.

Fixes: https://github.com/WhisperSystems/Signal-iOS/issues/1708

PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -83,7 +83,7 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     loggingIsEnabled = TRUE;
     [DebugLogger.sharedLogger enableTTYLogging];
 #elif RELEASE
-    loggingIsEnabled = Environment.preferences.loggingIsEnabled;
+    loggingIsEnabled = PropertyListPreferences.loggingIsEnabled;
 #endif
     if (loggingIsEnabled) {
         [DebugLogger.sharedLogger enableFileLogging];
@@ -134,8 +134,9 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     self.window.rootViewController = [storyboard instantiateInitialViewController];
     [self.window makeKeyAndVisible];
 
-    [VersionMigrations performUpdateCheck]; // this call must be made after environment has been initialized because in
-                                            // general upgrade may depend on environment
+    // performUpdateCheck must be invoked after Environment has been initialized because
+    // upgrade process may depend on Environment.
+    [VersionMigrations performUpdateCheck];
 
     // Accept push notification when app is not open
     NSDictionary *remoteNotif = launchOptions[UIApplicationLaunchOptionsRemoteNotificationKey];
--- a/Signal/src/environment/Environment.m
+++ b/Signal/src/environment/Environment.m
@@ -244,6 +244,8 @@ static Environment *environment = nil;
 
 + (PropertyListPreferences *)preferences
 {
+    OWSAssert([Environment getCurrent] != nil);
+    OWSAssert([Environment getCurrent].preferences != nil);
     return [Environment getCurrent].preferences;
 }
 
--- a/Signal/src/environment/PropertyListPreferences.h
+++ b/Signal/src/environment/PropertyListPreferences.h
@@ -43,8 +43,8 @@ extern NSString *const PropertyListPreferencesKeyEnableDebugLog;
 - (BOOL)getHasArchivedAMessage;
 - (void)setHasArchivedAMessage:(BOOL)enabled;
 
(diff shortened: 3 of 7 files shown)


Signal/src/AppDelegate.m
```diff
@@ -83,7 +83,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     loggingIsEnabled = TRUE;
     [DebugLogger.sharedLogger enableTTYLogging];
 #elif RELEASE
-    loggingIsEnabled = Environment.preferences.loggingIsEnabled;
+    loggingIsEnabled = PropertyListPreferences.loggingIsEnabled;
```
charlesmchen: Environment is initialized after this statement, so this was always set to NO in production. =/

Signal/src/environment/Environment.m
```diff
@@ -244,6 +244,8 @@ - (NotificationsManager *)notificationsManager
 
 + (PropertyListPreferences *)preferences
 {
+    OWSAssert([Environment getCurrent] != nil);
+    OWSAssert([Environment getCurrent].preferences != nil);
```
charlesmchen: This should help us catch most mistakes of this form going forward. It won't work for release-only bugs (like this one) even if we test release builds, since asserts don't fire in production.  I'm thinking about how to improve our asserts around this.

Signal/src/environment/PropertyListPreferences.h
```diff
@@ -61,8 +61,8 @@ extern NSString *const PropertyListPreferencesKeyEnableDebugLog;
 
 - (TSImageQuality)imageUploadQuality;
 
-- (nullable NSString *)lastRanVersion;
-- (NSString *)setAndGetCurrentVersion;
++ (nullable NSString *)lastRanVersion;
++ (NSString *)setAndGetCurrentVersion;
```
charlesmchen: I went ahead and converted all NSUserDefaults-backed prefs to be class methods, as another measure to future-proof us against these kinds of mistakes.

Signal/src/environment/Environment.m
michaelkirk: Not opposed to some redundancy, but I notice `getCurrent` is already asserting that it's not nil.

Signal/src/environment/PropertyListPreferences.h
michaelkirk: Looking at it another way, we would have caught this issue if *Environment* had *no* class methods except `getCurrent`, and all values pulled from that instance rather than the "convenient" class methods (since getCurrent asserts that it's not empty).

For testability design, I have a slight preference for leaving preferences as a property on the environment instance, rather than relying on more class methods, but I could let that go pretty easily.

Signal/src/environment/Environment.m
charlesmchen: So it does - but it uses `NSAssert()` which won't fire for release builds.  I've proposed modifying `OWSAssert()` to log in release builds, which would help.  And I guess if I do that, I'll do a follow-up PR to convert all (?) NSAsserts to OWSAsserts.  

https://github.com/WhisperSystems/SignalServiceKit/pull/126

Signal/src/environment/PropertyListPreferences.h
charlesmchen: > Looking at it another way, we would have caught this issue if Environment had no class methods except getCurrent, and all values pulled from that instance rather than the "convenient" class methods (since getCurrent asserts that it's not empty).

The `[Environment preferences]` convenience method uses `[Environment getCurrent]` - so removing it wouldn't help.

For lack of a better option, I'm proposing that we log assert failures for release builds: https://github.com/WhisperSystems/SignalServiceKit/pull/126

> For testability design, I have a slight preference for leaving preferences as a property on the environment instance, rather than relying on more class methods, but I could let that go pretty easily.

I think it would be worse to pre-instantiate a partially working instance of `PropertyListPreferences` before `Environment` was initialized, or any other solution I can think of.

Signal/src/AppDelegate.m
michaelkirk: One more thought - what about avoiding the prod/dev discrepancy with file logging altogether? Are we really worried about people with debug builds not being able to figure out why logging isn't being persisted?

```
#ifdef DEBUG
    [DebugLogger.sharedLogger enableTTYLogging];
#end

loggingIsEnabled = PropertyListPreferences.loggingIsEnabled;
```

charlesmchen: Sorry - I just merged.  I'll try to do this in a follow-up PR.

PR #1846, opened 2017-03-14
Limit size of outgoing text messages.

charlesmchen:
Easy way to test to send some large text (e.g. http://norvig.com/big.txt ) from desktop, copy & paste in iOS, try to send.

screen shot 2017-03-14 at 12 19 38 pm

PTAL @michaelkirk

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -853,6 +853,29 @@ typedef enum : NSUInteger {
          senderDisplayName:(NSString *)senderDisplayName
                       date:(NSDate *)date
 {
+    text = [text stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
+
+    // Limit outgoing text messages to 64kb.
+    //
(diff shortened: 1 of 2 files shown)


Signal/src/view controllers/MessagesViewController.m
```diff
@@ -852,6 +852,29 @@ - (void)didPressSendButton:(UIButton *)button
...
                       date:(NSDate *)date
 {
+    text = [text stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
+
+    // Limit outgoing text messages to 64kb.
+    //
+    // TODO: Convert large text messages to attachments
+    // which are presented as normal text messages.
+    const NSUInteger kMaxTextMessageSize = 64 * 1024;
+    if (text.length > kMaxTextMessageSize) {
```
michaelkirk: I think we want `[text lengthOfBytesUsingEncoding:NSUTF8StringEncoding]` here

charlesmchen: Good point, done.

PR #1850, opened 2017-03-15
Disable the "scroll to new message" animation.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -2177,12 +2177,21 @@ typedef enum : NSUInteger {
                                                                          forNotifications:notifications
                                                                              withMappings:self.messageMappings];
 
-    __block BOOL scrollToBottom = NO;
-
     if ([sectionChanges count] == 0 & [messageRowChanges count] == 0) {
         return;
     }
-
+    
+    __block BOOL scrollToBottom = NO;
+    const CGFloat kIsAtBottomTolerancePts = 5;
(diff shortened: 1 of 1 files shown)


michaelkirk:
Although as is, it is annoying, we haven't yet implemented another indication of new messages received while scrolling back in a thread.

charlesmchen:
> Although as is, it is annoying, we haven't yet implemented another indication of new messages received while scrolling back in a thread.

Sorry if this wasn't clear.  This doesn't affect the behavior when we receive new _incoming_ messages. This affects the behavior when we send a new _outgoing_ message. Per Moxie in standup, it feels snappier to skip the "new outgoing message rolls onscreen" message. Note that we auto-scroll to the bottom of the conversation whenever user activates the new message text field, so users should usually be at the bottom when this happens. They could add a message, scroll up, and hit "send", in which case the app should still animate to the new message. I'll check that last case.

michaelkirk:
> This doesn't affect the behavior when we receive new incoming messages. 

Huh. it sure looks like it would. `scrollToBottom` is set to `YES` whenever we're inserting a row, which would be either incoming or outgoing. What am I missing?

charlesmchen:
> Huh. it sure looks like it would. scrollToBottom is set to YES whenever we're inserting a row, which would be either incoming or outgoing. What am I missing?

You're right. PTAL

charlesmchen:
@michaelkirk There's more work to be done on this task. I'll do it in a subsequent PR.

PR #1853, opened 2017-03-15
move zxing to framework

michaelkirk:
and update build instructions

PTAL @charlesmchen 

If it looks good, I'll ff-merge the corresponding Carthage submodule commit.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,54 +1,84 @@
-## Building
+# Building
 
-1) Clone the repo to a working directory
+## 1. Clone
 
-2) [CocoaPods](http://cocoapods.org) is used to manage dependencies. Pods are setup easily and are distributed via a ruby gem. Follow the simple instructions on the website to setup. After setup, run the following command from the toplevel directory of Signal-iOS to download the dependencies for Signal-iOS:
+Clone the repo to a working directory
+
+## 2. Depenencies
+
+### tldr;
+
+Build and configure dependencies.
+
+
+```
+make dependencies
+```
+
+### Dependency Details
+
+We have a couple of dependency management tools. We us Carthage for
+managing frameworks, but because some of our dependencies are not yet
+framework compatible, we use Cocoapods to manage the remainder in a
+static library.
+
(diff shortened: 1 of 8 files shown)


BUILDING.md
```diff
@@ -1,55 +1,85 @@
...
+2.1) [CocoaPods](http://cocoapods.org) is used to manage the dependencies in our static library. Pods are setup easily and are distributed via a ruby gem. Follow the simple instructions on the website to setup. After setup, run the following command from the toplevel directory of Signal-iOS to download the dependencies for Signal-iOS:
 
 ```
 pod install
 ```
+
 If you are having build issues, first make sure your pods are up to date
+
 ```
 pod update
```
charlesmchen: I'm not sure we want to recommend a `pod update` when people have build issues.  Shouldn't it just be `pod install`?

BUILDING.md
```diff
@@ -1,55 +1,85 @@
...
-5) In the Signal target on the General tab, change the Team drop down to your own. On the Capabilities tab turn off Push Notifications and Data Protection. Only Background Modes should remain on.
+In the Signal target on the General tab, change the Team drop down to
+your own. On the Capabilities tab turn off Push Notifications and Data
+Protection. Only Background Modes should remain on.
 
-6) Some of our build scripts, like running tests, expect your Derived
+Some of our build scripts, like running tests, expect your Derived
 Data directory to be `$(PROJECT_DIR)/build`. In Xcode, go to `Preferences-> Locations`,
 and set the "Derived Data" dropdown to "Relative" and the text field
 value to "build".
```
charlesmchen: Huh, I've never done this and tests run fine for me.  Why are we using a relative dir?

michaelkirk: > Why are we using a relative dir?

No reason that I know of, I just documented that we were. (It's hardcoded in some of the build settings). I'll double check that I can build without it and update the docs if so.

BUILDING.md
michaelkirk: `pod repo update` is what we want I think. I'll update.

Issue #1338, opened 2016-09-21
Submit Debug Log should make it obvious it's copying to clipboard


michaelkirk:
When someone wants to submit a debug log they are presented with two choices.
1. email
2. issue on github

Usually by the time someone is submitting a debug log, they are doing so at the behest of support or an existing github issue. 

What people really want is to just copy it to the clipboard and they'll manage to get it to the right source. As is, people feel like they can't do that.

michaelkirk:
reopening until we incorporate the update to PastelogKit, which requires we update to the next major version of cocoalumberjack.

PR #1860, opened 2017-03-17
Intern pastelog

michaelkirk:
All of this has happened before.
All of this will happen again.

PTAL @charlesmchen 

We don't want this as an external dependency - it's just going to slow us down for future changes. 

Updated the copy per @charlesmchen  feedback over at https://github.com/WhisperSystems/PastelogKit/pull/1

finally incorporates @guillett's work to copy-to-pasteboard. (sorry for the delay 🙄 )

Fixes https://github.com/WhisperSystems/Signal-iOS/issues/1338

### DIFF
--- a/Signal/src/util/Pastelog.h
+++ b/Signal/src/util/Pastelog.h
@@ -1,0 +1,16 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+@interface Pastelog : NSObject <NSURLConnectionDelegate, NSURLConnectionDataDelegate, UIAlertViewDelegate>
+
+typedef void (^successBlock)(NSError *error, NSString *urlString);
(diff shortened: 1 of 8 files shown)


michaelkirk:
Fixes https://github.com/WhisperSystems/Signal-iOS/pull/1860

Signal/src/util/Pastelog.h
```diff
@@ -0,0 +1,16 @@
+//
```
charlesmchen: Signal/src/util/Pastelog...

Most projects I've worked on in the paste have isolated third party code in a separate directory (e.g. Signal/src/ThirdParty/Pastelog.m).  a) we can modify the precommit script to ignore this directory, to make it easier to take patches from the source. b) this makes the origin of the code clear.

PR #1863, opened 2017-03-19
Fix "Call failed" due to deadlock when immediately terminating call

michaelkirk:
It's not clear why we were ever dispatching `sync` here.

Before this:

Place a call
See "connecting..."
Hang up
UI hangs for ~5 seconds
See "call failed" on CallKit screen
Press "cancel" on CallKit screen
returned to responsive app.

PTAL @charlesmchen


michaelkirk:
PTAL @charlesmchen

PR #1864, opened 2017-03-19
Allow default browsers to be configured for links

### LINKED ISSUE #1514: Option to open links by default with googlechrome
I use Google Chrome as my default browser on iOS. Currently when I tap links they open in Safari, and there's no way to change this default.

I propose (and am happy to give a go at implementing, if it would be accepted) a new setting on the settings screen, which would only appear if Chrome is detected to be installed, which allows you to toggle whether Chrome should be the default link handler.

illicitonion:
This adds Chrome as a browser, and makes it trivial to add further
browsers.

FREEBIE

### Contributor checklist
 * iPhone SE Simulator running iOS 10.2, both with and without a browser installed which handles the googlechrome[s] link scheme.

- - - - - - - - - -

### Description
Fixes #1514 

This adds a setting to the Advanced settings screen to select the web browser which should be used for link handlers. It sniffs what browsers are installed (currently looking only for Chrome), and gives them as options. (Perhaps we should only show the setting at all if you don't only have Safari installed, but surfacing the capability through options may be informative to users too). When links in messages are clicked, your selected default browser is looked up and opened in place of Safari.

To add future browsers, BrowserUtil would need changing to detect and return schemes for the other browser. The scheme for sniffing would also need adding to Signal-Info.plist in the LSApplicationQueriesSchemes list.

Right now, I use Chrome as my main browser, and so having to copy links and paste them into the browser of my choice is a little annoying. Google apps that I use offer the option to auto-open in Chrome, or remain with Safari. Some other apps that I use give me the option when I long-press on a link of which browser I want to use.

### DIFF
--- a/Signal.xcodeproj/project.pbxproj
+++ b/Signal.xcodeproj/project.pbxproj
@@ -7,6 +7,9 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0773FC2A1E7D509700C1E683 /* OWSMessageViewCellDelegate.m in Sources */ = {isa = PBXBuildFile; fileRef = 0773FC291E7D509700C1E683 /* OWSMessageViewCellDelegate.m */; };
+		0773FC301E7D5DB100C1E683 /* DefaultBrowserSelectionViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = 0773FC2F1E7D5DB100C1E683 /* DefaultBrowserSelectionViewController.m */; };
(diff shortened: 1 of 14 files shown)


Imperiopolis-Signal:
Hi, thanks so much for your contribution. I'm sorry we didn't get a chance to properly review this. We're open to considering this feature in the future, but since this PR is so old it's not possible to integrate it as is. Feel free to open a new PR based off the latest code and ping me to take a look.

inoas:
Ref https://github.com/signalapp/Signal-iOS/issues/4243

PR #1856, opened 2017-03-16
Swipe to navigate back from messages view (#1845)

kevinmbeaulieu:
### Contributor checklist

- [x] I'm following the [code, UI and style conventions](https://github.com/WhisperSystems/Signal-iOS/blob/master/CONTRIBUTING.md#code-conventions)
- [x] My commits are rebased on the latest master branch
- [x] My commits are in nice logical chunks
- [x] My contribution is fully baked and is ready to be merged as is
- [x] I have tested my contribution on these devices:
 * iPhone 7, iOS 10.2
- [x] I have made the choice whether I want the [BitHub reward](https://github.com/WhisperSystems/Signal-Android/wiki/BitHub-Rewards) or not by omitting or adding the word `FREEBIE` in my commit message

- - - - - - - - - -

### Description
Fixes #1845.

PR #1866, opened 2017-03-19
fix swipe back

### LINKED ISSUE #1845: Swipe right in conversation does not exit anymore

### Bug description
When in a conversation it used to be possible to swipe right from the left edge of the screen to go back. This is not the case anymore and disrupts the workflow. (Compare to iMessage or Whatsapp)

### Steps to reproduce
- Go into any conversation
- Swipe right from the left edge of the screen

**Actual result:** nothing happens
**Expected result:** should drag the screen, following the finger and "close" the conversation going back to the default screen where all conversations are visible

### Device info

**Device**: iPhone 6s
**iOS version**: 10.2.1
**Signal version:** 2.8.1

michaelkirk:
fixes #1845 
closes #1856

Since we're using a custom back button we have to juggle the navigationController's interactivePopGestureRecognizer ourselves.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/view controllers/OWSConversationSettingsTableViewController.m
+++ b/Signal/src/view controllers/OWSConversationSettingsTableViewController.m
@@ -211,6 +211,10 @@ static NSString *const OWSConversationSettingsTableViewControllerSegueShowGroupM
 - (void)viewWillAppear:(BOOL)animated
 {
     [super viewWillAppear:animated];
+
+    // Since we're using a custom back button, we have to do some extra work to manage the interactivePopGestureRecognizer
+    self.navigationController.interactivePopGestureRecognizer.delegate = self;
+
     // HACK to unselect rows when swiping back
     // http://stackoverflow.com/questions/19379510/uitableviewcell-doesnt-get-deselected-when-swiping-back-quickly
     [self.tableView deselectRowAtIndexPath:[self.tableView indexPathForSelectedRow] animated:animated];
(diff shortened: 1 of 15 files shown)


Signal/src/view controllers/OWSConversationSettingsTableViewController.m
```diff
@@ -211,6 +211,10 @@ - (void)viewDidLoad
 - (void)viewWillAppear:(BOOL)animated
 {
     [super viewWillAppear:animated];
+
+    // Since we're using a custom back button, we have to do some extra work to manage the interactivePopGestureRecognizer
+    self.navigationController.interactivePopGestureRecognizer.delegate = self;
```
charlesmchen: Great. Do we also need to do this in every VC which calls "useOWSBackButton"?

michaelkirk: Good call.

I understand we use a custom back button as a way to space our custom header for the message view controller. But why do we use them elsewhere?

Unless you have a reason not to, I propose that in this PR I revert all VC's to use the standard back button except for the message view controller, and tweak the spacing on the custom back button so that it lines up there with the system default used elsewhere.

What do you think @charlesmchen?

charlesmchen: I think the reason I added it everywhere was that:

* It looks really bad if you transition from a VC "default" back button to a "custom" back button and vice versa.  So we need to use the "custom" back button in any view which can be reached from the a view using the "custom" back button.
* We need the "custom" back button in the messages VC so that we can use the left-aligned title.
* You can reach the "scan QR code" view from the messages VC _and_ the settings VC. 
* Therefore we need to use the "custom" back button in 90-100% of our VCs, so I just decided to use it everywhere.  

I'm very open to suggesting another approach.  We could, for example, add a "useCustomBackButton" property to the "scan QR code" view.

michaelkirk: > It looks really bad if you transition from a VC "default" back button to a "custom" back button and vice versa

I'm proposing trying to fix *this* problem. Then I think the others go away.

michaelkirk:
I've made some changes to make the custom back button align perfectly with the existing back button, which allows us to *only* use the custom back button where we need it, and use the default back button (and interactivePopGesture) everywhere else.

Can you take another look @charlesmchen ?

7 custom back screenshot

iphone5s custom back screenshot

michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/acf3a6e027b59a3c1804619f147b4c95768e8040

PR #1871, opened 2017-03-21
Added share location feature

christophhagen:
Added a new class ShareLocationManager, which integrates with
MessagesViewController to allow sending the current location. The
implementation tries to mimic the existing behaviour on Android.
Permissions for location services are also handled.

### Contributor checklist

 * iPhone 6, iOS 11.0 (15A5327)

- - - - - - - - - -

### Description
Adds a "share current location" feature to Signal. The behaviour is close to the behaviour already present in Signal-Android. The location coordinates are determined before a google maps static image is created, which is then sent together with a message containing the coordinates, address information and a google maps link.
The feature is accessible through the attachment picker.

## Implementation details
- Location service permissions are handled
- Error messages will be displayed if location could not be sent
- Strings required  are added to Localized.strings
- Description for location use is added to Signal-Info.plist
- CoreLocation framework is added as required dependency
- The ShareLocationManager class handles the whole process
- The size of the static maps image is 400x400 pixels (can easily be changed)
- A self-designed icon is included for the attachment picker

See the screenshots for how it looks from the sending and receiving side.

If anything about the implementation needs to be fixed, please let me know.

## Screenshots

| Attachment picker |  Location permission |
|:--------------------:|:----------------------|
Attachment picker | img_1450
| Sent location |  No permission dialog |
img_1451 | img_1453
| No internet connection | 
img_1452


chanying2:
Any plans to resolve the conflicts and merge the latest master?

christophhagen:
I can update this pull request to the latest master, but are there any implementation problems that explain why this pull request was ignored for so long?

@michaelkirk, any thoughts?

chanying2:
This PR is quite nice in my opinion, though, an ideal PR would implement a pick a location feature, rather than your immediate location, and similar a few other things, like what you are used to in other popular apps.  @christophhagen

george-gw:
Any reason why you're using the Google Maps API instead of the built-in APIs?

christophhagen:
@george-gw: To keep the UI consistent for iOS and Android. Signal for Android uses Google Maps, so I wanted to use it as well for iOS. I will soon create a new pull request (or update this one) with an updated version where both Apple and Google maps are possible.

christophhagen:
Updated this PR to the current master. Also fixed some bugs, and cleaned the code. The attachment picker now has an icon that I designed quickly. Maybe someone is willing to spend more time on that. 

I would like to see this feature soon. Hope it gets merged.

@charlesmchen @michaelkirk 
I would like to suggest that for future versions we send the location as it's own data type (for example only the coordinates in a .gpx file as an attachment). 
This has several advantages:
- No contact to external servers until the receiver side chooses to do the lookup
- Less traffic over the signal server (no static images)
- Location can be sent even if no internet connection exists at the time
- Location can be displayed on an interactive map on the receiver side.
- Each platform can use the default maps application.

Thoughts?

tcurdt:
Why was this never merged? At least I don't see it in the release.

PR #1876, opened 2017-03-22
Only initiate WebRTC calls, not RedPhone calls.

charlesmchen:
PTAL @michaelkirk


michaelkirk:
LGTM

Signal/src/AppDelegate.m
```diff
@@ -410,7 +410,8 @@ - (BOOL)application:(UIApplication *)application continueUserActivity:(nonnull N
 
         OutboundCallInitiator *outboundCallInitiator = [Environment getCurrent].outboundCallInitiator;
         OWSAssert(outboundCallInitiator);
-        return [outboundCallInitiator initiateCallWithHandle:phoneNumber];
+        [outboundCallInitiator initiateCallWithHandle:phoneNumber];
+        return YES;
```
michaelkirk: It's worth noting there is a change in behavior here - we now return `YES` here if the number was not parseable in the outboundCallInitiator. This could happen (e.g.) when Signal-calling someone via Contacts.app when they have a bogus (unparseable) phone number.

Honestly, I'm not totally clear on the implication of this, but I think it's fine, according to:

https://developer.apple.com/reference/uikit/uiapplicationdelegate/1623072-application?language=objc

> Use this method to update your iOS app so that the user can continue the activity from where they left off. If you do not implement this method or if your implementation returns NO, iOS tries to create a document for your app to open using a URL.

To clarify, I think it's OK because we are handling the NSUserActivity, in the sense that we display "not a signal user".

And just to pile on to my comment avalanche here, "some day" we should hook people in to the invite flow here.

> +15551231234 is not a Signal user yet. Would you like to invite them?
> [ No thanks ], [ Send Invite via SMS ]

If you agree, I'll open a ticket.

charlesmchen: You're right - the invite flow would be best here.  I'm going to restore the return value to make that eventual work easier.

PR #1882, opened 2017-03-23
[SSK] Better logging for envelopes

michaelkirk:
Leave last 3 digits of recipientId when scrubbing logs.

This is in line with other Signal clients, and makes it possible to
trace interactions.

PTAL @charlesmchen


michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/a61e82653a8893a6e822547b95f387ea03c5c8e3

PR #1887, opened 2017-03-24
Show alerts with relevant error messages when attachment-related erro…

charlesmchen:
…rs occur.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/view controllers/MessagesViewController.m
+++ b/Signal/src/view controllers/MessagesViewController.m
@@ -1950,7 +1950,7 @@ typedef enum : NSUInteger {
                                              DDLogWarn(@"%@ %s Invalid attachment: %@.",
                                                        self.tag,
                                                        __PRETTY_FUNCTION__,
-                                                       attachment ? [attachment errorMessage] : @"Missing data");
+                                                       attachment ? [attachment errorName] : @"Missing data");
                                              [self showErrorAlertForAttachment:attachment];
                                              failedToPickAttachment(nil);
                                          } else {
@@ -1997,7 +1997,7 @@ typedef enum : NSUInteger {
                                               DDLogWarn(@"%@ %s Invalid attachment: %@.",
                                                         self.tag,
                                                         __PRETTY_FUNCTION__,
-                                                        attachment ? [attachment errorMessage] : @"Missing data");
+                                                        attachment ? [attachment errorName] : @"Missing data");
                                               [self showErrorAlertForAttachment:attachment];
                                               failedToPickAttachment(nil);
                                           } else {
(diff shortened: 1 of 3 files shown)


Signal/src/view controllers/MessagesViewController.m
```diff
@@ -1988,21 +1989,21 @@ - (void)imagePickerController:(UIImagePickerController *)picker
...
-                           self.tag,
-                           __PRETTY_FUNCTION__,
-                           attachment ? [attachment errorMessage] : @"Missing data");
-                 failedToPickAttachment(nil);
-             } else {
-                 [self dismissViewControllerAnimated:YES
-                                          completion:^{
-                                              OWSAssert([NSThread isMainThread]);
-                                              
+             [self dismissViewControllerAnimated:YES
```
charlesmchen: One significant change - previously if users selected an invalid image in the image picker (camera roll), we would just ... do nothing. No error, not dismiss the image picker.  Now we dismiss the picker before we show the error alert.  This isn't a perfect experience, but a) it's better than doing nothing (which felt broken) and b) we'll soon support arbitrary attachments and significantly raise attachment sizes so this will happen very rarely.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -5,7 +5,7 @@
 import Foundation
 import MobileCoreServices
 
-enum SignalAttachmentError: String {
+enum SignalAttachmentError: Int {
```
charlesmchen: I'm not sure why this was a String enum.

Signal/src/view controllers/SignalAttachment.swift
```diff
@@ -97,7 +97,25 @@ class SignalAttachment: NSObject {
             assert(false)
             return nil
         }
-        return "\(error)"
```
charlesmchen: errorDescription used to be used for logging only.  Now it's used for logging and for user-facing error messages. We could separate these out into two separate methods...

charlesmchen:
img_9242

Signal/src/view controllers/SignalAttachment.swift
michaelkirk: nit: no need for `Int` either, unless you intend to use the `.rawValue` somewhere. You should be able to just `enum SignalAttachmentError {`

But (see my other comment), I think we should `enum SignalAttachmentError: LocalizedError {`

Signal/src/view controllers/SignalAttachment.swift
michaelkirk: nit: I think it would be more conventional for our errors to adopt the https://developer.apple.com/reference/foundation/localizederror protocol.

e.g. http://stackoverflow.com/a/39176551/353178

Signal/src/view controllers/SignalAttachment.swift
charlesmchen: Done.

Signal/src/view controllers/SignalAttachment.swift
charlesmchen: Done.

PR #1889, opened 2017-03-25
Copy images to pasteboard as data, not UIImage.

charlesmchen:
This is why copy & paste of images on iOS 9 isn't working.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/Models/TSMessageAdapaters/TSPhotoAdapter.m
+++ b/Signal/src/Models/TSMessageAdapaters/TSPhotoAdapter.m
@@ -113,6 +113,9 @@
     }
 
     if (action == @selector(copy:)) {
+        // We should always copy to the pasteboard as data, not an UIImage.
+        // The pasteboard should has as specific as UTI type as possible and
+        // data support should be far more general than UIImage support.
         OWSAssert(self.attachment.filePath.length > 0);
         NSString *fileExtension = [self.attachment.filePath pathExtension];
         NSArray *utiTypes = (__bridge_transfer NSArray *)UTTypeCreateAllIdentifiersForTag(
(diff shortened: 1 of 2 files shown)


Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -246,6 +246,7 @@ - (void)performEditingAction:(SEL)action
     if ([self isVideo]) {
         if (action == @selector(copy:)) {
             NSData *data = [NSData dataWithContentsOfURL:self.fileURL];
+            // TODO: This assumes all videos are mp4.
```
charlesmchen: This will change soon.

Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -260,7 +261,6 @@ - (void)performEditingAction:(SEL)action
             NSData *data = [NSData dataWithContentsOfURL:self.fileURL];
 
             NSString *pasteboardType = [MIMETypeUtil getSupportedExtensionFromAudioMIMEType:self.contentType];
-            [UIPasteboard.generalPasteboard setData:data forPasteboardType:(NSString *)UIPasteboardNameGeneral];
```
charlesmchen: Each call to `setData` corresponds to a separate "copy".  The `setData` calls below overwrote this pasteboard, so this was redundant.

Signal/src/Models/TSMessageAdapaters/TSPhotoAdapter.m
```diff
@@ -112,7 +113,18 @@ - (void)performEditingAction:(SEL)action
     }
 
     if (action == @selector(copy:)) {
-        UIPasteboard.generalPasteboard.image = self.image;
+        OWSAssert(self.attachment.filePath.length > 0);
```
michaelkirk: Because this code is non-obvious, it might be worth leaving a comment as to why we're losing the much more concise version of the above.

charlesmchen: Done.

PR #1891, opened 2017-03-25
Slightly tweak the appearance of the attachment upload progress bar.

charlesmchen:
It was too fat, which looks fine when the progress bar is empty but looks "chubby" as the progress completes.

PTAL @michaelkirk


charlesmchen:
* Thinner border.
* Reduce rounding of corners.
* Reduce overall height.

### Before

progress before

### After

progress after

PR #1894, opened 2017-03-26
Log submission of logs.

charlesmchen:
The intention is to make it easier to confirm that the logs contain the latest entries.

PTAL @michaelkirk


Signal/src/view controllers/AdvancedSettingsTableViewController.m
```diff
@@ -127,6 +127,7 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
     [tableView deselectRowAtIndexPath:indexPath animated:YES];
 
     if ([tableView cellForRowAtIndexPath:indexPath] == self.submitLogCell) {
+        DDLogDebug(@"%@ Submitting debug logs", self.tag);
```
michaelkirk: If you want to verify this with users submitting logs from the app store release, than this needs to be DDLogInfo

charlesmchen: Done.

PR #1895, opened 2017-03-26
Rework country views in registration view.

charlesmchen:
Previously, our registration view would show a "country code" label (without a country name) when using the default country (derived from the device's region setting) and show the "country name" if the country was changed.  

I've changed this to always show the "country code" label and to always show the actual country code after the calling code.  Country names tend to be long, so there isn't enough space to show the "country code" label, the country name and the calling code.

The goal is to be more consistent and to always indicate to the user their current state.

### Before (default)

screen shot 2017-03-26 at 3 57 08 pm

### Before (after country changed)

screen shot 2017-03-26 at 3 57 15 pm

### After

screen shot 2017-03-26 at 3 56 21 pm

PTAL @michaelkirk

### DIFF
--- a/Signal/src/view controllers/CountryCodeViewController.h
+++ b/Signal/src/view controllers/CountryCodeViewController.h
@@ -1,3 +1,7 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
 #import <UIKit/UIKit.h>
 
 @class CountryCodeViewController;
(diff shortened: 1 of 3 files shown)


NicolasCGN:
The string 'Already have a Signal account?' doesn't get translated in the localised app. Is there a type in the string key?

charlesmchen:
Thanks @NicolasCGN I've fixed that here: https://github.com/WhisperSystems/Signal-iOS/pull/1899

PR #1899, opened 2017-03-27
Fix translation of “already have an account?” button.

charlesmchen:
@NicolasCGN commented an hour ago:

> The string 'Already have a Signal account?' doesn't get translated in the localised app. Is there a type in the string key?

PTAL @michaelkirk


michaelkirk:
interface builder =(

LGTM.

PR #1904, opened 2017-03-28
Charlesmchen/oversize text messages and arbitrary attachments

charlesmchen:
What's included:

* Modify Debug UI to send oversize text messages and attachments with an unknown mime type.
* Add generic attachment adapter to present arbitrary attachments in messages view.
* Present oversize text message as a snippet with indication to tap.

What's not included:

* Tap on oversize text message shows full message in separate view.
* Tap on arbitrary attachment shows ... share UI?

PTAL @michaelkirk 

### Incoming and outgoing oversize text messages and arbitrary attachments.

screen shot 2017-03-28 at 3 29 17 pm

screen shot 2017-03-28 at 3 29 12 pm

PR #1911, opened 2017-03-29
Charlesmchen/oversize text message view

charlesmchen:
This work builds on https://github.com/WhisperSystems/Signal-iOS/pull/1904 which is not yet merged.

PTAL @michaelkirk 

screen shot 2017-03-29 at 12 19 48 pm
screen shot 2017-03-29 at 12 19 42 pm

### DIFF
--- a/Signal/src/view controllers/OversizeTextMessageViewController.swift
+++ b/Signal/src/view controllers/OversizeTextMessageViewController.swift
@@ -1,0 +1,150 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+import WebRTC
+import PromiseKit
+
+class OversizeTextMessageViewController: UIViewController {
+
+    let TAG = "[OversizeTextMessageViewController]"
+
+    let message: TSMessage
+
+    // MARK: Initializers
+
+    @available(*, unavailable, message:"use message: constructor instead.")
+    required init?(coder aDecoder: NSCoder) {
+        message = TSMessage()
+        super.init(coder: aDecoder)
+    }
+
+    required init(message: TSMessage) {
+        self.message = message
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    // MARK: Attachment
+
+    private func attachmentStream() -> TSAttachmentStream? {
+        guard message.hasAttachments() else {
+            Logger.error("\(TAG) message has no attachments.")
+            assert(false)
+            return nil
+        }
+        let attachmentIDRaw = message.attachmentIds[0]
+        guard attachmentIDRaw is String else {
+            Logger.error("\(TAG) message attachment id is not a string.")
+            assert(false)
+            return nil
+        }
+        let attachmentID = attachmentIDRaw as! String
+        guard let attachment = TSAttachment.fetch(withUniqueID:attachmentID) else {
+            Logger.error("\(TAG) could not load attachment.")
+            assert(false)
+            return nil
+        }
+        guard attachment.contentType == OWSMimeTypeOversizeTextMessage else {
+            Logger.error("\(TAG) attachment has unexpected content type.")
+            assert(false)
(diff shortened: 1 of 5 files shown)


Signal/src/view controllers/OversizeTextMessageViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+    // MARK: Attachment
+
+    private func attachmentStream() -> TSAttachmentStream? {
+        guard message.hasAttachments() else {
+            Logger.error("\(TAG) message has no attachments.")
+            assert(false)
+            return nil
+        }
+        let attachmentIDRaw = message.attachmentIds[0]
+        guard attachmentIDRaw is String else {
```
michaelkirk: nit:

or:
```
guard let attachmentID = message.attachmentIds[0] as? String else {
    Logger.error("\(TAG) message attachment id is not a string.")
    assertionFailure()
    return nil
}
```

Signal/src/view controllers/OversizeTextMessageViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+            Logger.error("\(TAG) could not load attachment.")
+            assert(false)
+            return nil
+        }
+        guard attachment.contentType == OWSMimeTypeOversizeTextMessage else {
+            Logger.error("\(TAG) attachment has unexpected content type.")
+            assert(false)
+            return nil
+        }
+        guard attachment is TSAttachmentStream else {
```
michaelkirk: nit: could combine a couple of these guards to be more concise and avoid the `!`
```
guard let stream = TSAttachment.fetch(withUniqueID:attachmentID) as? TSAttachmentStream else {
...
}
```

Signal/src/view controllers/OversizeTextMessageViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+
+        self.navigationItem.title = NSLocalizedString("OVERSIZE_TEXT_MESSAGE_VIEW_TITLE",
+                                                      comment: "The title of the 'oversize text message' view.")
+
+        self.view.backgroundColor = UIColor.white
+
+        let textView = UITextView()
+        textView.textColor = UIColor.black
+        textView.text = displayText()
+        textView.font = UIFont.ows_regularFont(withSize:14)
```
michaelkirk: can we use dynamic type font here?

charlesmchen:
@michaelkirk Sorry, I just saw your comments here.  I'll address them post merge.

charlesmchen:
> I feel like we need to do more to show that [Tap for More] is interface, not content. i.e. it's actionable text, rather than something the sender sent.

Agreed.  I just haven't figured out what to do yet.  I'm open to ideas.  One thing I considered is making an oversize text adapter like the generic attachment adapter.  I actually did that already and threw it away once.  But I'm starting to think it was the right approach.

charlesmchen:
> I feel like we need to do more to show that [Tap for More] is interface, not content. i.e. it's actionable text, rather than something the sender sent.

I'm doing this work under another trello card.

Signal/src/view controllers/OversizeTextMessageViewController.swift
charlesmchen: Done, nice, thank you.

Signal/src/view controllers/OversizeTextMessageViewController.swift
charlesmchen: Done.

charlesmchen:
https://github.com/WhisperSystems/Signal-iOS/commit/3765d28da9144b28cb8777f4194ab88f169d08b8

Signal/src/view controllers/OversizeTextMessageViewController.swift
charlesmchen: Done.

PR #1914, opened 2017-03-29
Improve support for arbitrary attachments.

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/SignalServiceKit/pull/152

PR #1915, opened 2017-03-29
Charlesmchen/image view vs share

charlesmchen:
This builds upon https://github.com/WhisperSystems/Signal-iOS/pull/1914 which is not yet merged.

PTAL @michaelkirk 

img_9353
img_9352


Signal/src/view controllers/FullImageViewController.m
```diff
@@ -274,21 +306,30 @@ - (void)presentFromViewController:(UIViewController *)viewController {
                      animated:NO
                    completion:^{
                        UIWindow *window = [UIApplication sharedApplication].keyWindow;
+                       // During the presentation animation, we want to seamlessly animate the image
```
charlesmchen: Decided to add some comments around the presentation animation which is complicated.

PR #1916, opened 2017-03-30
Charlesmchen/spaces in paths

charlesmchen:
PTAL @michaelkirk


charlesmchen:
PTAL @michaelkirk 

I redid this work _without_ using precommit.

PR #1917, opened 2017-03-30
Enforce singleton access for MessagesManager and MessageSender

michaelkirk:
Corresponding with https://github.com/WhisperSystems/SignalServiceKit/pull/155

PTAL @charlesmchen


Signal/src/network/PushManager.m
```diff
@@ -44,44 +43,29 @@ + (instancetype)sharedManager {
...
 {
     self = [super init];
     if (!self) {
         return self;
     }
 
-    _contactsManager = contactsManager;
-    _callUIAdapter = callService.callUIAdapter;
-
-    _messageSender = [[OWSMessageSender alloc] initWithNetworkManager:networkManager
```
michaelkirk: Much of this complexity goes back to a circular dependency that used to exist between messageSender and messagesManager.

That circular dependency was removed, so this code can be greatly simplified.

Signal/src/environment/Environment.m
```diff
@@ -13,15 +13,17 @@
 #import "TSGroupThread.h"
 #import <SignalServiceKit/ContactsUpdater.h>
 
-#define isRegisteredUserDefaultString @"isRegistered"
```
michaelkirk: unused, just some random cleanup.

Signal/src/Jobs/MessageFetcherJob.swift
```diff
@@ -13,15 +13,13 @@ class MessageFetcherJob: NSObject {
     // MARK: injected dependencies
     let networkManager: TSNetworkManager
     let messagesManager: TSMessagesManager
-    let messageSender: MessageSender
     let signalService: OWSSignalService
 
     var runPromises = [Double: Promise<Void>]()
 
-    init(messagesManager: TSMessagesManager, messageSender: MessageSender, networkManager: TSNetworkManager, signalService: OWSSignalService) {
+    init(messagesManager: TSMessagesManager, networkManager: TSNetworkManager, signalService: OWSSignalService) {
```
michaelkirk: Pruned this unused dependency.

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/3d6c153ca39ba9fc083c3316509dc3089cd12636

PR #1919, opened 2017-03-31
Add explicit calling code state to registration view.

charlesmchen:
We need to separate "view state" from "view model state" in registration view, i.e. the current calling code (e.g. "+44") from how it is presented (e.g. "+44 (UK)").  We should be used the raw calling code when constructing the phone number to verify.  

This bug doesn't have any effect in production which is how it slipped through testing, but it's not right.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/RegistrationViewController.m
+++ b/Signal/src/ViewControllers/RegistrationViewController.m
@@ -16,6 +16,8 @@ static NSString *const kCodeSentSegue = @"codeSent";
 
 @interface RegistrationViewController ()
 
+@property (nonatomic) NSString *lastCallingCode;
+
 @end
 
 @implementation RegistrationViewController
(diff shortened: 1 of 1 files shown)


Signal/src/view controllers/RegistrationViewController.m
```diff
@@ -16,6 +16,8 @@
 
 @interface RegistrationViewController ()
 
+@property (nonatomic) NSString *lastCallingCode;
```
michaelkirk: nit: At a glance "last" denotes "previous", but I think we mean the currently selected one? Maybe "selectedCallingCode" or just "callingCode"

feel free to ignore, just my initial thought when I saw it being used.

charlesmchen:
> I'm sort of surprised this didn't break more things. I guess tryParsePhoneNumberFromUserSpecifiedText filtered out the non-numeric chars?

Yup, I was very concerned when I discovered this, but the the filtering saved us.

Signal/src/view controllers/RegistrationViewController.m
charlesmchen: Fixed in the next PR.

PR #1920, opened 2017-03-31
New debug action: print sessions

michaelkirk:
Helpful for debugging session state.

PTAL @charlesmchen


michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/e828001544b7694c660ccb5ae0f19f6ad69fc834

PR #1921, opened 2017-03-31
Fix crash when viewing changed safety numbers

michaelkirk:
Instantiate fingerprintVC programmatically since MVC was not built from
the storyboard and thus doesn't have it's storyboard-defined segues.

PTAL @charlesmchen


charlesmchen:
Please don't forget to put this in the test plan.

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/afcb08c10507c29317d9280f88ea6681306b57d1

PR #1922, opened 2017-03-31
Charlesmchen/blocking2

charlesmchen:
This PR requires: https://github.com/WhisperSystems/SignalServiceKit/pull/156.

* Adds `BlockListViewController`, a page in the settings that shows the block list and lets you unblock people. 
   * It can also be used to present `AddToBlockListViewController`.
   * TODO: We should present blocklist entries using names not numbers for contacts.
* Adds `AddToBlockListViewController` a page in settings that lets you add numbers to the block list.
   * Currently, you can only add phone numbers by typing or pasting them in.  
   * TODO: I'll eventually show a list of non-blocked contacts to make it easier to block contacts.
* Modify `CountryCodeViewController` to make it usable from storyboards or programmatically.
* Pulled out the "format phone number in text field as you type" logic and refined it a bit.
* I pulled out `OWSTableViewController` so that we can use it in other contexts as well.

PTAL @michaelkirk 

### BlockListViewController

screen shot 2017-03-31 at 1 46 38 pm

### AddToBlockListViewController

screen shot 2017-03-31 at 1 47 54 pm

### CountryCodeViewController

screen shot 2017-03-31 at 1 51 08 pm

PR #1923, opened 2017-03-31
Charlesmchen/blocking3

charlesmchen:
This PR builds on https://github.com/WhisperSystems/Signal-iOS/pull/1922 which is unmerged.

* Present blocklist entries using names not numbers for contacts.
* Show a list of non-blocked contacts to make it easier to block contacts.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/AddToBlockListViewController.m
+++ b/Signal/src/ViewControllers/AddToBlockListViewController.m
@@ -4,7 +4,6 @@
 
 #import "AddToBlockListViewController.h"
 #import "ContactTableViewCell.h"
-#import "ContactsUpdater.h"
 #import "CountryCodeViewController.h"
 #import "Environment.h"
 #import "OWSContactsManager.h"
@@ -75,13 +74,6 @@ NSString *const kContactsTable_CellReuseIdentifier = @"kContactsTable_CellReuseI
 {
     [super viewDidLoad];
     [self.navigationController.navigationBar setTranslucent:NO];
-}
-
-- (void)viewWillAppear:(BOOL)animated
-{
-    [super viewWillAppear:animated];
-
-    [self refreshContacts];
 }
 
 - (void)addNotificationListeners
@@ -481,17 +473,6 @@ NSString *const kContactsTable_CellReuseIdentifier = @"kContactsTable_CellReuseI
                                                    style:UIAlertActionStyleDefault
                                                  handler:nil]];
     [self presentViewController:controller animated:YES completion:nil];
-}
-
-- (void)refreshContacts
-{
-    [[ContactsUpdater sharedUpdater] updateSignalContactIntersectionWithABContacts:self.contactsManager.allContacts
-        success:^{
-            [self updateContacts];
-        }
-        failure:^(NSError *error) {
-            DDLogError(@"%@ Error updating contacts", self.tag);
-        }];
 }
 
 #pragma mark - UIScrollViewDelegate
--- a/Signal/src/ViewControllers/AddToBlockListViewController.m
+++ b/Signal/src/ViewControllers/AddToBlockListViewController.m
@@ -25,7 +25,6 @@ NSString *const kContactsTable_CellReuseIdentifier = @"kContactsTable_CellReuseI
 
 #pragma mark -
 
-// TODO: Add a list of contacts to make it easier to block contacts.
 @interface AddToBlockListViewController () <CountryCodeViewControllerDelegate,
     UITextFieldDelegate,
     UITableViewDataSource,
--- a/Signal/src/ViewControllers/AddToBlockListViewController.m
+++ b/Signal/src/ViewControllers/AddToBlockListViewController.m
@@ -3,7 +3,11 @@
 //
 
 #import "AddToBlockListViewController.h"
+#import "ContactTableViewCell.h"
+#import "ContactsUpdater.h"
 #import "CountryCodeViewController.h"
+#import "Environment.h"
+#import "OWSContactsManager.h"
 #import "PhoneNumber.h"
 #import "StringUtil.h"
 #import "UIFont+OWS.h"
@@ -17,10 +21,15 @@ NS_ASSUME_NONNULL_BEGIN
 
 NSString * const kAddToBlockListViewControllerCellIdentifier = @"kAddToBlockListViewControllerCellIdentifier";
 
+NSString *const kContactsTable_CellReuseIdentifier = @"kContactsTable_CellReuseIdentifier";
+
 #pragma mark -
 
 // TODO: Add a list of contacts to make it easier to block contacts.
-@interface AddToBlockListViewController () <CountryCodeViewControllerDelegate, UITextFieldDelegate>
+@interface AddToBlockListViewController () <CountryCodeViewControllerDelegate,
+    UITextFieldDelegate,
+    UITableViewDataSource,
+    UITableViewDelegate>
 
 @property (nonatomic, readonly) OWSBlockingManager *blockingManager;
 
@@ -31,19 +40,18 @@ NSString * const kAddToBlockListViewControllerCellIdentifier = @"kAddToBlockList
 
 @property (nonatomic) UIButton *blockButton;
 
+@property (nonatomic) UITableView *contactsTableView;
+
 @property (nonatomic) NSString *callingCode;
+
+@property (nonatomic, readonly) OWSContactsManager *contactsManager;
+@property (nonatomic) NSArray<Contact *> *contacts;
 
 @end
 
 #pragma mark -
 
 @implementation AddToBlockListViewController
-
-- (void)viewDidLoad
-{
-    [super viewDidLoad];
-    [self.navigationController.navigationBar setTranslucent:NO];
-}
 
 - (void)loadView
 {
@@ -52,6 +60,8 @@ NSString * const kAddToBlockListViewControllerCellIdentifier = @"kAddToBlockList
     self.view.backgroundColor = [UIColor whiteColor];
     
     _blockingManager = [OWSBlockingManager sharedManager];
+    _contactsManager = [Environment getCurrent].contactsManager;
+    self.contacts = self.contactsManager.signalContacts;
 
     self.title = NSLocalizedString(@"SETTINGS_ADD_TO_BLOCK_LIST_TITLE", @"");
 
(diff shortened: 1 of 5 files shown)


charlesmchen:
@michaelkirk This PR is ready to go.

Signal/src/ViewControllers/AddToBlockListViewController.m
```diff
@@ -31,27 +38,28 @@ @interface AddToBlockListViewController () <CountryCodeViewControllerDelegate, U
 
 @property (nonatomic) UIButton *blockButton;
 
+@property (nonatomic) UITableView *contactsTableView;
+
 @property (nonatomic) NSString *callingCode;
 
+@property (nonatomic, readonly) OWSContactsManager *contactsManager;
+@property (nonatomic) NSArray<Contact *> *contacts;
```
michaelkirk: nit: maybe a custom setter with `AssertOnMainQueue` is in order given how we update it.

Signal/src/ViewControllers/BlockListViewController.m
```diff
@@ -214,6 +235,35 @@ - (void)blockedPhoneNumbersDidChange:(id)notification
...
+    NSMutableDictionary<NSString *, Contact *> *contactMap = [NSMutableDictionary new];
+    for (Contact *contact in contacts) {
+        for (PhoneNumber *phoneNumber in contact.parsedPhoneNumbers) {
+            NSString *phoneNumberE164 = phoneNumber.toE164;
+            if (phoneNumberE164.length > 0) {
+                contactMap[phoneNumberE164] = contact;
+            }
+        }
+    }
+    self.contactMap = contactMap;
```
michaelkirk: nit: `[contactMap copy]` to assign immutable

Signal/src/ViewControllers/BlockListViewController.m
```diff
@@ -145,24 +167,23 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
...
+- (void)showUnblockActionSheet:(NSString *)phoneNumber displayName:(NSString *)displayName
 {
     OWSAssert(phoneNumber.length > 0);
-
-    PhoneNumber *parsedPhoneNumber = [PhoneNumber tryParsePhoneNumberFromUserSpecifiedText:phoneNumber];
-    NSString *displayPhoneNumber = (parsedPhoneNumber ? parsedPhoneNumber.toE164 : phoneNumber);
+    OWSAssert(displayName.length > 0);
 
     NSString *title = [NSString stringWithFormat:NSLocalizedString(@"BLOCK_LIST_UNBLOCK_TITLE_FORMAT",
                                                      @"A format for the 'unblock phone number' action sheet title."),
```
michaelkirk: nit: translation comment should mention any embedded strings.

Signal/src/ViewControllers/BlockListViewController.m
```diff
@@ -214,6 +235,35 @@ - (void)blockedPhoneNumbersDidChange:(id)notification
...
+        self.contacts = self.contactsManager.signalContacts;
+        [self.tableView reloadData];
+    });
+}
+
+- (void)setContacts:(NSArray<Contact *> *)contacts
+{
+    _contacts = contacts;
+
+    NSMutableDictionary<NSString *, Contact *> *contactMap = [NSMutableDictionary new];
```
michaelkirk: nit: `self.contactMap` seems to be re-implementing `[contactManager contactForPhoneIdentifier:]` (but more efficiently!). Would you consider modifying/using that?

Signal/src/ViewControllers/BlockListViewController.m
```diff
@@ -130,6 +134,24 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
...
+- (NSString *)displayNameForIndexPath:(NSIndexPath *)indexPath
+{
+    NSString *phoneNumber = _blockedPhoneNumbers[(NSUInteger)indexPath.item];
+    PhoneNumber *parsedPhoneNumber = [PhoneNumber tryParsePhoneNumberFromUserSpecifiedText:phoneNumber];
+
+    // Try to parse and present the phone number in E164.
+    // It should already be in E164, so this should always work.
+    // If an invalid or unparsable phone number is already in the block list,
+    // present it as-is.
+    NSString *displayName = (parsedPhoneNumber ? parsedPhoneNumber.toE164 : phoneNumber);
```
michaelkirk: nit: Consider duduplicating some of this logic by using `[contactManager displayNameForPhoneIdentifier:]` when the number is parseable as e164.

Signal/src/ViewControllers/AddToBlockListViewController.m
```diff
@@ -332,6 +375,113 @@ - (BOOL)textFieldShouldReturn:(UITextField *)textField
...
+}
+
+- (void)showBlockActionSheet:(Contact *)contact
+{
+    OWSAssert(contact);
+
+    NSString *displayName = contact.fullName;
+
+    NSString *title = [NSString stringWithFormat:NSLocalizedString(@"BLOCK_LIST_BLOCK_TITLE_FORMAT",
+                                                     @"A format for the 'block phone number' action sheet title."),
```
michaelkirk: nit: comment should include it's embedded parameters.

Signal/src/ViewControllers/AddToBlockListViewController.m
```diff
@@ -332,6 +375,113 @@ - (BOOL)textFieldShouldReturn:(UITextField *)textField
...
+                                displayName];
+
+    UIAlertController *actionSheetController =
+        [UIAlertController alertControllerWithTitle:title message:nil preferredStyle:UIAlertControllerStyleActionSheet];
+
+    __weak AddToBlockListViewController *weakSelf = self;
+    UIAlertAction *unblockAction = [UIAlertAction
+        actionWithTitle:NSLocalizedString(@"BLOCK_LIST_BLOCK_BUTTON", @"Button label for the 'block' button")
+                  style:UIAlertActionStyleDefault
```
michaelkirk: nit: I think this should be UIAlertActionStyleDestructive since this is a confirmation pattern.

Signal/src/ViewControllers/AddToBlockListViewController.m
```diff
@@ -332,6 +375,113 @@ - (BOOL)textFieldShouldReturn:(UITextField *)textField
...
+
+- (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    Contact *contact = self.contacts[(NSUInteger)indexPath.item];
+
+    ContactTableViewCell *cell = [_contactsTableView cellForRowAtIndexPath:indexPath];
+    if (!cell) {
+        cell = [ContactTableViewCell new];
+    }
+    [cell configureWithContact:contact contactsManager:self.contactsManager];
```
michaelkirk: Forgive the stream of consciousness:

It feels a little strange when you block a user and they're still in this list (and then you can "block them again"). 

I think this causes it to be unclear that blocking "worked" (Yes, **even though** there is a "success" alert).

I guess we probably don't want to exclude Bob from this list in case Alice forgot that she already blocked Bob, and she's doesn't know why Bob isn't in the list of people to block...

One (stupid-simple-easy) solution would be to leave everyone in the list as is, but dismiss the view controller after selection. Rather than showing the "success" alert, the user could **see** that it worked via their updated blocked list. This would work OK as long as the user didn't have a ton of people in their blocked list, and could quickly see that Bob is in the blocked list.

Admittedly it would also be slower to block a lot of people at once with the proposed change, but I *think* that's an acceptable tradeoff for the greater good.

Also consider:

* adding an accessory view to indicate who is already blocked (or maybe a subtitle would be better yet).
* making the already-blocked contact rows unselectable.

charlesmchen:
> One nit: I feel the "Add Phone Number" and "Contact Name" labels are unncessarily bolded.

Done.

> Another though, "Add Phone Number" might be better as "Add entry" or something else generic since we can add by contact rather than just adding by "phone number".

I like the "Add New..." or "New..." copy in your mockup.  I'll do something like that when I do the "use table edit mode" card.

signal-2017-04-03-174529

Signal/src/ViewControllers/AddToBlockListViewController.m
charlesmchen: The work to improve thread safety around updates is already done in https://github.com/WhisperSystems/Signal-iOS/pull/1928

Signal/src/ViewControllers/AddToBlockListViewController.m
charlesmchen: Yup, so a few things.

* I'm already filtering blocked users from this list in https://github.com/WhisperSystems/Signal-iOS/pull/1928 .  The other option was to show them in this list but with a checkmark, but I felt it was slightly better to filter them?
* I have an open trello card to pop the "add block" view after a successful block.  Most users will only want to add a single block at a time, and this will - as you point out - give them the context to see how their block list has changed.

Signal/src/ViewControllers/AddToBlockListViewController.m
charlesmchen: I'll address in https://github.com/WhisperSystems/Signal-iOS/pull/1927 when I get there since these strings have moved.

Signal/src/ViewControllers/AddToBlockListViewController.m
charlesmchen: I'll address in #1927 when I get there since this logic has moved.

Signal/src/ViewControllers/BlockListViewController.m
charlesmchen: I'll address in #1927 when I get there since these strings have moved.

Signal/src/ViewControllers/BlockListViewController.m
charlesmchen: I'll address in #1927 when I get there since this logic has moved.

Signal/src/ViewControllers/BlockListViewController.m
charlesmchen: I'll address in #1927 when I get there since this logic has moved.

Signal/src/ViewControllers/BlockListViewController.m
charlesmchen: I'll address in https://github.com/WhisperSystems/Signal-iOS/pull/1928

charlesmchen:
> Another thing that occurs to me, which I wish I mentioned earlier, is that this is not the idiomatic way to write YapDB driven table view controllers.

Yes, `OWSBlockingManager` abstracts its data store.  I'd use the YapDB approach if these were `TSYapDatabaseObjects`, but it's just a simple list of strings.  And block lists will be so short that we don't need to worry about animating changes etc.

> The risk being that we'll have an inconsistent view of the state of the DB. The recommended approach is here: https://github.com/yapstudios/YapDatabase/wiki/YapDatabaseModifiedNotification

`kNSNotificationName_BlockedPhoneNumbersDidChange` works in much the same way, guaranteeing that observers of block list state are always updated.  The data store is a private concern of `OWSBlockingManager`, so it can guarantee that the notification will always be fired as necessary.

PR #1924, opened 2017-04-01
Charlesmchen/singleton assert

charlesmchen:
Depends on https://github.com/WhisperSystems/SignalServiceKit/pull/157

There are probably more singletons, but these are the obvious candidates.

PTAL @michaelkirk


Signal/src/UserInterface/Notifications/CallNotificationsAdapter.swift
```diff
@@ -21,7 +21,7 @@ class CallNotificationsAdapter: NSObject {
 //        if #available(iOS 10.0, *) {
 //            adaptee = UserNotificationsAdaptee()
 //        } else {
-            adaptee = NotificationsManager()
+            adaptee = Environment.getCurrent().notificationsManager
```
charlesmchen: This is a real fix caught with the assertion - we were instantiating this singleton twice.

PR #1925, opened 2017-04-01
Charlesmchen/blocking4

charlesmchen:
Builds upon https://github.com/WhisperSystems/Signal-iOS/pull/1923 which is unmerged.
Requires: https://github.com/WhisperSystems/SignalServiceKit/pull/158

PTAL @michaelkirk

PR #1927, opened 2017-04-03
Charlesmchen/blocking5

charlesmchen:
This PR depends on https://github.com/WhisperSystems/Signal-iOS/pull/1925 and its dependencies which are not yet merged.

I'm going to add in-line comments on this work, but I'm going to hold off until its dependencies are merged.


charlesmchen:
Adds a "block this user" row to the 1:1 conversation settings view.

* I've reworked `OWSConversationSettingsTableViewController` to extend `OWSTableViewController`.  The contents of the "conversation settings" view are just complicated and dynamic enough that manually managing table state with `NSIndexPath` and enums is awkward and brittle.  `OWSConversationSettingsTableViewController` has a custom header and some custom cells defined in the storyboard.  `OWSConversationSettingsTableViewController` still uses them, but now does table state management using `OWSTableViewController` which is much simpler and safer.
* `OWSConversationSettingsTableViewController` uses custom cells which try to match the layout of the default UITableViewCells.  Unfortunately that layout depends on the form factor of the device.  On some devices, our icons are not aligned.  We should create the table cells programmatically, but I deferred that change since this isn't a regression.
* `OWSConversationSettingsTableViewController`'s custom table view cells also have fixed custom heights which won't l10n well. Another reason to create them programmatically is responsive layout and sizing.
* I chose "circle with strike" to be the block icon. WDYT?
* I pulled out `BlockListUIUtils` so that we can block and unblock contacts / phone numbers from multiple areas of the UI.
   * One subtlety is that if we block a contact from a list of contacts, we want to block all numbers associated with that contact.


screen shot 2017-04-03 at 10 14 52 am
screen shot 2017-04-03 at 10 14 47 am

charlesmchen:
Nit from @michaelkirk:

> nit: the translation comment should specify what is being embedded in place of the interpolation chars.

charlesmchen:
https://github.com/WhisperSystems/Signal-iOS/pull/1923#discussion_r109541945

@michaelkirk said:

> nit: I think this should be UIAlertActionStyleDestructive since this is a confirmation pattern.

charlesmchen:
https://github.com/WhisperSystems/Signal-iOS/pull/1923#discussion_r109538574

> nit: Consider duduplicating some of this logic by using [contactManager displayNameForPhoneIdentifier:] when the number is parseable as e164.

charlesmchen:
https://github.com/WhisperSystems/Signal-iOS/pull/1923#discussion_r109538109

> nit: self.contactMap seems to be re-implementing [contactManager contactForPhoneIdentifier:] (but more efficiently!). Would you consider modifying/using that?

charlesmchen:
> nit: the translation comment should specify what is being embedded in place of the interpolation chars.

Done.

charlesmchen:
> nit: I think this should be UIAlertActionStyleDestructive since this is a confirmation pattern.

Done.

charlesmchen:
> nit: Consider duduplicating some of this logic by using [contactManager displayNameForPhoneIdentifier:] when the number is parseable as e164.

Done.

charlesmchen:
> nit: self.contactMap seems to be re-implementing [contactManager contactForPhoneIdentifier:] (but more efficiently!). Would you consider modifying/using that?

Done.

charlesmchen:
@michaelkirk PTAL

Signal/src/contact/OWSContactsManager.m
```diff
@@ -46,6 +49,41 @@ - (id)init {
     return self;
 }
 
+- (NSDictionary *)latestContactsById
+{
+    @synchronized(self)
+    {
+        return _latestContactsById;
```
michaelkirk: nit: seems like we should initialize `_latestContactsById` to an empty dict in `init`

charlesmchen:
> That looks like a regression to me, given that it is aligned before this PR, and it's not after. This affects the two devices I tested on iPhone 5c and iPhone7.

It was misaligned on some devices and I switched it to be misaligned on others.  I'll revert it, but the original issue remains unresolved.

Signal/src/ViewControllers/AddToBlockListViewController.m
```diff
@@ -273,25 +274,18 @@ - (void)tryToBlockPhoneNumber
...
-
-    _phoneNumberTextField.text = nil;
+    __weak AddToBlockListViewController *weakSelf = self;
+    [BlockListUIUtils showBlockPhoneNumberActionSheet:[parsedPhoneNumber toE164]
+                                   fromViewController:self
+                                      blockingManager:_blockingManager
+                                      contactsManager:_contactsManager
+                                      completionBlock:^(BOOL isBlocked) {
+                                          if (isBlocked) {
+                                              // Clear phone number text field is block succeeds.
```
charlesmchen: Done.

Signal/src/ViewControllers/BlockListUIUtils.h
```diff
@@ -0,0 +1,35 @@
...
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+@class Contact;
+@class OWSBlockingManager;
+@class OWSContactsManager;
+
+typedef void (^BlockActionCompletionBlock)(BOOL isBlocked);
```
charlesmchen: Done.

Signal/src/ViewControllers/BlockListUIUtils.m
```diff
@@ -0,0 +1,247 @@
...
+                    contactsManager:(OWSContactsManager *)contactsManager
+                    completionBlock:(BlockActionCompletionBlock)completionBlock
+{
+    NSMutableArray<NSString *> *phoneNumbers = [NSMutableArray new];
+    for (PhoneNumber *phoneNumber in contact.parsedPhoneNumbers) {
+        if (phoneNumber.toE164.length > 0) {
+            [phoneNumbers addObject:phoneNumber.toE164];
+        }
+    }
+    if (phoneNumbers.count < 1) {
```
charlesmchen: Done.

Signal/src/ViewControllers/BlockListUIUtils.m
```diff
@@ -0,0 +1,247 @@
...
+        [self showBlockFailedAlert:fromViewController];
+        if (completionBlock) {
```
charlesmchen: Done.

Signal/src/ViewControllers/BlockListUIUtils.m
```diff
@@ -0,0 +1,247 @@
...
+
+    UIAlertAction *unblockAction = [UIAlertAction
+        actionWithTitle:NSLocalizedString(@"BLOCK_LIST_BLOCK_BUTTON", @"Button label for the 'block' button")
+                  style:UIAlertActionStyleDestructive
+                handler:^(UIAlertAction *_Nonnull action) {
+                    [self blockPhoneNumbers:phoneNumbers
+                                displayName:displayName
+                         fromViewController:fromViewController
+                            blockingManager:blockingManager];
+                    if (completionBlock) {
```
charlesmchen: Done.

Signal/src/ViewControllers/BlockListUIUtils.m
```diff
@@ -0,0 +1,247 @@
...
+
+    UIAlertAction *unblockAction = [UIAlertAction
+        actionWithTitle:NSLocalizedString(@"BLOCK_LIST_UNBLOCK_BUTTON", @"Button label for the 'unblock' button")
+                  style:UIAlertActionStyleDestructive
+                handler:^(UIAlertAction *_Nonnull action) {
+                    [BlockListUIUtils unblockPhoneNumber:phoneNumber
+                                             displayName:displayName
+                                      fromViewController:fromViewController
+                                         blockingManager:blockingManager];
+                    if (completionBlock) {
```
charlesmchen: Done.

michaelkirk:
>> That looks like a regression to me, given that it is aligned before this PR, and it's not after. This affects the two devices I tested on iPhone 5c and iPhone7.
> It was misaligned on some devices and I switched it to be misaligned on others. I'll revert it, but the original issue remains unresolved.

Before this commit it's aligned on iPhone5c, iPhoneSE, iPhone6, and iPhone7. Which was it misaligned on? 

My concern is that in rewriting large portions of the conversation settings table some layout is getting lost.

Signal/src/ViewControllers/BlockListViewController.m
```diff
@@ -137,18 +137,7 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
...
-    // It should already be in E164, so this should always work.
-    // If an invalid or unparsable phone number is already in the block list,
-    // present it as-is.
-    NSString *displayName = (parsedPhoneNumber ? parsedPhoneNumber.toE164 : phoneNumber);
-    Contact *contact = self.contactMap[displayName];
-    if (contact && [contact fullName].length > 0) {
-        displayName = [contact fullName];
-    }
-
+    NSString *displayName = [_contactsManager displayNameForPhoneIdentifier:phoneNumber];
```
charlesmchen: I'm sorry, I don't understand this comment. Per your earlier comment, I now use `[_contactsManager  displayNameForPhoneIdentifier:]`.  What logic on 146 are you referring to?

Note: Your comment made me realize we weren't properly formatting display names for alerts, so I've added that.

Signal/src/ViewControllers/BlockListViewController.m
```diff
@@ -167,65 +156,19 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
         }
         case BlockListViewControllerSection_BlockList: {
             NSString *phoneNumber = _blockedPhoneNumbers[(NSUInteger)indexPath.item];
-            NSString *displayName = [self displayNameForIndexPath:indexPath];
-            [self showUnblockActionSheet:phoneNumber displayName:displayName];
+            [BlockListUIUtils showUnblockPhoneNumberActionSheet:phoneNumber
+                                             fromViewController:self
+                                                blockingManager:_blockingManager
+
```
charlesmchen: Done.

Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
```diff
@@ -407,38 +416,58 @@ - (IBAction)disappearingMessagesSwitchValueDidChange:(id)sender
...
 {
-    self.disappearingMessagesConfiguration.enabled = flag;
+    OWSAssert(!self.isGroupThread);
 
-    // When this message is called as a result of the switch being flipped, this will be a no-op
-    // but it allows us to resuse the method to set the switch programmatically in view setup.
-    self.disappearingMessagesSwitch.on = flag;
-    [self durationSliderDidChange:self.disappearingMessagesDurationSlider];
+    if (![sender isKindOfClass:[UISwitch class]]) {
+        DDLogError(@"%@ Unexpected sender for block user switch: %@", self.tag, sender);
```
charlesmchen: Done.

Signal/src/ViewControllers/OWSTableViewController.h
```diff
@@ -32,15 +32,30 @@ NS_ASSUME_NONNULL_BEGIN
...
+
++ (OWSTableItem *)itemWithCustomCellBlock:(OWSTableCustomCellBlock)customCellBlock
+                          customRowHeight:(CGFloat)customRowHeight
+                              actionBlock:(OWSTableActionBlock)actionBlock;
+
++ (OWSTableItem *)itemWithCustomCellBlock:(OWSTableCustomCellBlock)customCellBlock
+                              actionBlock:(OWSTableActionBlock)actionBlock;
+
+- (UITableViewCell *)customCell;
+- (NSNumber *)customRowHeight;
```
charlesmchen: I haven't gotten there yet, but I'd eventually like to let these items specify "custom row height" using either a fixed value or a block.  Once we add block support, this will be a legit method, not a getter.

Signal/src/ViewControllers/OWSTableViewController.m
```diff
@@ -95,6 +99,60 @@ + (OWSTableItem *)actionWithTitle:(NSString *)title actionBlock:(OWSTableActionB
     return item;
 }
 
++ (OWSTableItem *)itemWithCustomCell:(UITableViewCell *)customCell
+                     customRowHeight:(CGFloat)customRowHeight
+                         actionBlock:(OWSTableActionBlock)actionBlock
```
charlesmchen: Done.

Signal/src/contact/OWSContactsManager.m
charlesmchen: We already do?

charlesmchen:
> Before this commit it's aligned on iPhone5c, iPhoneSE, iPhone6, and iPhone7. Which was it misaligned on?

> My concern is that in rewriting large portions of the conversation settings table some layout is getting lost.

I've reverted the change.  I don't have a list but its definitely misaligned on some devices - maybe the "plus" devices?

charlesmchen:
PTAL @michaelkirk

Signal/src/ViewControllers/BlockListViewController.m
michaelkirk: Yup, I'm still all for using `[_contactsManager displayNameForPhooneIdentifier:]`.

Here's what I'm talking about from line 146:
```
NSString *displayName = (parsedPhoneNumber ? parsedPhoneNumber.toE164 : phoneNumber);
```

Previous behavior was that if the number was not parseable to e164, to use the raw `phoneNumber`. Now in that case, I think we'll get `nil`.

Like I said, probably doesn't matter since anyone that can be usefully blocked will have an e164 parseable phone number, but I wanted to highlight that the logic was lost in case you were depending on it.

Signal/src/contact/OWSContactsManager.m
michaelkirk: oh. You're right. 😬

Signal/src/ViewControllers/BlockListViewController.m
charlesmchen: Ah, the _old_ 146.  

> Now in that case, I think we'll get nil.

Sorry, I still don't see the issue.  We're now using `displayNameForPhoneIdentifier`.  It doesn't try to parse the phone number - it just looks up a contact by phone number.  If no contact is found, it returns the phone number, not nil.  Isn't that the desired behavior?

```
- (NSString * _Nonnull)displayNameForPhoneIdentifier:(NSString * _Nullable)identifier {
    if (!identifier) {
        return self.unknownContactName;
    }
    Contact *contact = [self contactForPhoneIdentifier:identifier];
    
    NSString *displayName = (contact.fullName.length > 0) ? contact.fullName : identifier;
    
    return displayName;
}
```

michaelkirk: Do you see that in the previous implementation of displayNameForIndexPath there was a situation where we would return phoneNumber?

It seems like there no longer is.

michaelkirk: Oh wait I'm wrong. that logic is inside of contact manager. Sheesh. Sorry.

PR #1928, opened 2017-04-03
Charlesmchen/blocking6

charlesmchen:
* Hide blocked users and current user in “new 1:1 conversation” view.
* Hide blocked users and current user in “new/edit group conversation” view.
* Hide blocked users and current user in “add to blocklist” view.
* Hide blocked users and current user in “blocklist” view.
* Improve thready safety around contact changes in views.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/AddToBlockListViewController.m
+++ b/Signal/src/ViewControllers/AddToBlockListViewController.m
@@ -16,6 +16,7 @@
 #import "ViewControllerUtils.h"
 #import <SignalServiceKit/OWSBlockingManager.h>
 #import <SignalServiceKit/PhoneNumberUtil.h>
+#import <SignalServiceKit/TSAccountManager.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -31,6 +32,7 @@ NSString *const kContactsTable_CellReuseIdentifier = @"kContactsTable_CellReuseI
     UITableViewDelegate>
 
 @property (nonatomic, readonly) OWSBlockingManager *blockingManager;
+@property (nonatomic, readonly) NSArray<NSString *> *blockedPhoneNumbers;
 
 @property (nonatomic) UIButton *countryNameButton;
 @property (nonatomic) UIButton *countryCodeButton;
@@ -59,8 +61,9 @@ NSString *const kContactsTable_CellReuseIdentifier = @"kContactsTable_CellReuseI
     self.view.backgroundColor = [UIColor whiteColor];
     
     _blockingManager = [OWSBlockingManager sharedManager];
+    _blockedPhoneNumbers = [_blockingManager blockedPhoneNumbers];
     _contactsManager = [Environment getCurrent].contactsManager;
-    self.contacts = self.contactsManager.signalContacts;
+    self.contacts = [self filteredContacts];
 
     self.title = NSLocalizedString(@"SETTINGS_ADD_TO_BLOCK_LIST_TITLE", @"");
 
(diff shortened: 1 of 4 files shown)


charlesmchen:
> nit: [contactMap copy] to assign immutable

https://github.com/WhisperSystems/Signal-iOS/pull/1923#discussion_r109537067

charlesmchen:
PTAL @michaelkirk

Signal/src/ViewControllers/AddToBlockListViewController.m
```diff
@@ -316,20 +319,63 @@ - (void)updateBlockButtonEnabling
...
-    });
+    OWSAssert([NSThread isMainThread]);
+
+    self.contacts = [self filteredContacts];
+    [self.contactsTableView reloadData];
+}
+
+- (BOOL)isContactBlocked:(Contact *)contact
+{
+    if (contact.parsedPhoneNumbers.count < 1) {
```
michaelkirk: Why does this mean that they are blocked?

Is it just a way to exclude them from the list? If so, consider adding that as a comment since it's non-obvious.

Signal/src/ViewControllers/AddToBlockListViewController.m
```diff
@@ -316,20 +319,63 @@ - (void)updateBlockButtonEnabling
...
+
+- (NSArray<Contact *> *_Nonnull)filteredContacts
+{
+    NSMutableArray<Contact *> *result = [NSMutableArray new];
+    for (Contact *contact in self.contactsManager.signalContacts) {
+        if (![self isContactBlocked:contact] && ![self isCurrentUserContact:contact]) {
+            [result addObject:contact];
+        }
+    }
+    return result;
```
michaelkirk: nit: `return [result copy]` to return immutable

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -667,46 +675,49 @@ - (Contact *)contactForIndexPath:(NSIndexPath *)indexPath {
...
-    dispatch_async(dispatch_get_main_queue(), ^{
-        [self updateSearchResultsForSearchController:self.searchController];
-        [self.tableView reloadData];
-        [self updateAfterRefreshTry];
-    });
+- (NSArray<Contact *> *_Nonnull)filteredContacts
+{
+    NSMutableArray<Contact *> *result = [NSMutableArray new];
+    for (Contact *contact in self.contactsManager.signalContacts) {
+        if (![self isContactBlocked:contact] && ![self isCurrentUserContact:contact]) {
```
michaelkirk: ### 1. messaging self

It looks like you're trying to exclude the current user from the compose view. In my experience, approximately 1/2 of all people think the app is broken when they can't message themselves. I'm hesitant to change it.

### 2. blocked users

It looks like you're excluding blocked contacts from the compose list. On the one hand - I get it. Alice has stated she doesn't want Bob messaging her, she probably doesn't want to message Bob either.

But I'm afraid that when we hide information like this, a non-insignificant number of people will get confused. Maybe Alice accidentally blocked Bob, or forgot she blocked Bob. Why isn't he in her contacts? 

Alice know's if she doesn't want to message Bob. We don't need to enforce that by hiding him here. Instead I think we should show blocked contacts in the compose view like everyone else, then when Alice actually tries to send something to him (clearly she *wants* to communicate at this point) we prompt her with the "unblock to send" action sheet.

The plus side I see to what you've proposed is that maybe Alice doesn't want to have to *see* Bob in her list of contacts at all, but I still think the "Why isn't Bob in my compose list?" is more likely to feel broken.

A compromise might be relegating blocked contacts in a separate section at the end of the list, or appearing in search in a future release.

What do you think @charlesmchen?

Signal/src/ViewControllers/AddToBlockListViewController.m
charlesmchen: I never bother with this for an object that a) we construct ourselves locally and b) is only used internally within a class - especially a small, simple class like this.  There's no chance we'll ever cast it to mutable and we'll never expose it to anyone else.  Copying unnecessarily is expensive.  Perhaps that's a premature optimization.

Signal/src/ViewControllers/MessageComposeTableViewController.m
charlesmchen: Yup, this work is outdated because these PRs have been so backed up.  I think I did most of this work over the weekend.  

There's a trello card to change some of this behavior and a) always show blocked users and self and b) label them as such c) and in some cases force the user to unblock to use them or prevent them from re-blocking a blocked user, etc.  

My original intent was to merge this before making those changes since those changes are not ship blockers and this change is better than nothing IMO.  But I guess I'll just go ahead and make those changes here.

Signal/src/ViewControllers/AddToBlockListViewController.m
michaelkirk: Copying is super cheap if you're the last reference holder. It's a good pattern for factory methods like this.

Signal/src/ViewControllers/MessageComposeTableViewController.m
michaelkirk: sounds good

charlesmchen:
> I'm concerned about removing "self" from the compose list as part of this work, but otherwise LGTM.

I've reverted that.

Signal/src/ViewControllers/AddToBlockListViewController.m
charlesmchen: Added a comment and renamed the method.

Signal/src/ViewControllers/AddToBlockListViewController.m
charlesmchen: Done.

PR #1930, opened 2017-04-04
Charlesmchen/blocking7

charlesmchen:
* Block actions in message view for blocked contact conversations.
* Block actions in message view for blocked group conversations.
* Add block status indicator view.

PTAL @michaelkirk 

screen shot 2017-04-04 at 3 36 48 pm
screen shot 2017-04-04 at 3 36 34 pm

### DIFF
--- a/Signal/src/ViewControllers/BlockListUIUtils.m
+++ b/Signal/src/ViewControllers/BlockListUIUtils.m
@@ -10,6 +10,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+typedef void (^BlockAlertCompletionBlock)();
+
 @implementation BlockListUIUtils
 
 + (void)showBlockContactActionSheet:(Contact *)contact
@@ -27,10 +29,12 @@ NS_ASSUME_NONNULL_BEGIN
     if (phoneNumbers.count < 1) {
         DDLogError(@"%@ Contact has no phone numbers", self.tag);
         OWSAssert(0);
-        [self showBlockFailedAlert:fromViewController];
-        if (completionBlock) {
-            completionBlock(NO);
-        }
+        [self showBlockFailedAlert:fromViewController
+                   completionBlock:^{
+                       if (completionBlock) {
+                           completionBlock(NO);
+                       }
+                   }];
         return;
     }
     NSString *displayName = [contactsManager displayNameForContact:contact];
@@ -81,10 +85,12 @@ NS_ASSUME_NONNULL_BEGIN
                     [self blockPhoneNumbers:phoneNumbers
                                 displayName:displayName
                          fromViewController:fromViewController
-                            blockingManager:blockingManager];
-                    if (completionBlock) {
-                        completionBlock(YES);
-                    }
+                            blockingManager:blockingManager
+                            completionBlock:^{
+                                if (completionBlock) {
+                                    completionBlock(YES);
+                                }
+                            }];
                 }];
     [actionSheetController addAction:unblockAction];
 
@@ -104,6 +110,7 @@ NS_ASSUME_NONNULL_BEGIN
               displayName:(NSString *)displayName
        fromViewController:(UIViewController *)fromViewController
           blockingManager:(OWSBlockingManager *)blockingManager
+          completionBlock:(BlockAlertCompletionBlock)completionBlock
 {
     OWSAssert(phoneNumbers.count > 0);
     OWSAssert(displayName.length > 0);
@@ -122,7 +129,8 @@ NS_ASSUME_NONNULL_BEGIN
                                                         @"The message format of the 'user blocked' "
                                                         @"alert. Embeds {{the blocked user's name or phone number}}."),
                                    [self formatDisplayNameForAlertMessage:displayName]]
-            fromViewController:fromViewController];
+            fromViewController:fromViewController
+               completionBlock:completionBlock];
 }
 
 + (void)showUnblockPhoneNumberActionSheet:(NSString *)phoneNumber
@@ -153,10 +161,12 @@ NS_ASSUME_NONNULL_BEGIN
                     [BlockListUIUtils unblockPhoneNumber:phoneNumber
                                              displayName:displayName
                                       fromViewController:fromViewController
-                                         blockingManager:blockingManager];
-                    if (completionBlock) {
-                        completionBlock(NO);
-                    }
+                                         blockingManager:blockingManager
+                                         completionBlock:^{
+                                             if (completionBlock) {
+                                                 completionBlock(NO);
+                                             }
+                                         }];
                 }];
     [actionSheetController addAction:unblockAction];
 
@@ -176,6 +186,7 @@ NS_ASSUME_NONNULL_BEGIN
                displayName:(NSString *)displayName
         fromViewController:(UIViewController *)fromViewController
            blockingManager:(OWSBlockingManager *)blockingManager
+           completionBlock:(BlockAlertCompletionBlock)completionBlock
 {
     OWSAssert(phoneNumber.length > 0);
     OWSAssert(displayName.length > 0);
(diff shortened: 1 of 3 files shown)


charlesmchen:
Discussion: 

* It wasn't 100% clear what to do when user taps the "group has blocked members indicator".  I chose to present the "block list" view. It won't indicate which users in the block list are in the group, but I don't think it's worth it to build a view that would do so and most people's block lists should be quite short.
* The "1:1" and "group" block indicators "float" about the conversation view and obscure some of its contents.  Most critically, it obscures the "load more/older messages" interaction.  I think we should discuss how to present this.

charlesmchen:
PTAL @michaelkirk

Signal/src/ViewControllers/BlockListUIUtils.m
```diff
@@ -27,10 +29,12 @@ + (void)showBlockContactActionSheet:(Contact *)contact
     if (phoneNumbers.count < 1) {
         DDLogError(@"%@ Contact has no phone numbers", self.tag);
         OWSAssert(0);
-        [self showBlockFailedAlert:fromViewController];
-        if (completionBlock) {
-            completionBlock(NO);
-        }
+        [self showBlockFailedAlert:fromViewController
```
charlesmchen: Not calling the completion block until after the confirmation/failure alerts have been dismissed, allows us to present alerts/vcs in the completion handler.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -500,6 +524,133 @@ - (void)viewWillAppear:(BOOL)animated
...
+    }
+
+    NSString *blockStateMessage = nil;
+    if ([self isBlockedContactConversation]) {
+        blockStateMessage = NSLocalizedString(@"MESSAGES_VIEW_CONTACT_BLOCKED",
+                                              @"Indicates that this 1:1 conversation has been blocked.");
+    } else {
+        int blockedGroupMemberCount = [self blockedGroupMemberCount];
+        if (blockedGroupMemberCount == 1) {
+            blockStateMessage = [NSString stringWithFormat:NSLocalizedString(@"MESSAGES_VIEW_GROUP_1_MEMBER_BLOCKED_FORMAT",
```
michaelkirk: This doesn't seem like a "format" string, why are we passing in count?

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -500,6 +524,133 @@ - (void)viewWillAppear:(BOOL)animated
...
+                                                                             @"Indicates that a single member of this group has been blocked."),
+                                 blockedGroupMemberCount];
+        } else if (blockedGroupMemberCount > 1) {
+            blockStateMessage = [NSString stringWithFormat:NSLocalizedString(@"MESSAGES_VIEW_GROUP_N_MEMBERS_BLOCKED_FORMAT",
+                                                                             @"Indicates that some members of this group has been blocked. Embeds "
+                                                                             @"{{the number of blocked user in this group}}."),
```
michaelkirk: typo: number of blocked user**s** in this group

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -500,6 +524,133 @@ - (void)viewWillAppear:(BOOL)animated
...
+        return NO;
+    }
+    NSString *contactIdentifier = ((TSContactThread *)self.thread).contactIdentifier;
+    return [[_blockingManager blockedPhoneNumbers] containsObject:contactIdentifier];
+}
+
+- (int)blockedGroupMemberCount
+{
+    if (![self.thread isKindOfClass:[TSGroupThread class]]) {
+        return 0;
```
michaelkirk: nit:I think `OWSAssert(NO); `

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -500,6 +524,133 @@ - (void)viewWillAppear:(BOOL)animated
...
+    }
+    
+    if (blockStateMessage) {
+        UILabel *label = [UILabel new];
+        label.font = [UIFont ows_mediumFontWithSize:14.f];
+        label.text = blockStateMessage;
+        label.textColor = [UIColor whiteColor];
+        
+        UIView * blockStateIndicator = [UIView new];
+        blockStateIndicator.backgroundColor = [UIColor ows_signalBrandBlueColor];
```
michaelkirk: I think we want material blue. signalBrandBlueColor is a bit darker and works well for text. e.g. compare to the navbar background.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -529,6 +529,15 @@
 /* message footer while attachment is uploading */
 "MESSAGE_STATUS_UPLOADING" = "Uploading...";
 
+/* Indicates that this 1:1 conversation has been blocked. */
+"MESSAGES_VIEW_CONTACT_BLOCKED" = "This User is Blocked";
```
michaelkirk: Maybe "You Blocked this User" to be clear that the user isn't e.g. banned from the signal service.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -529,6 +529,15 @@
...
+
+/* Indicates that a single member of this group has been blocked. */
+"MESSAGES_VIEW_GROUP_1_MEMBER_BLOCKED_FORMAT" = "1 Blocked Member of this Group";
```
michaelkirk: Maybe: Similarly, maybe "Blocked 1 Member of this Group" and "Blocked %d Members of this Group".

take it or leave it.

Signal/src/ViewControllers/MessagesViewController.m

Signal/src/ViewControllers/MessagesViewController.m

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: I'm thinking we should use a color that is more "alert-y", somewhere on the red, orange, yellow spectrum - to draw the eye and to emphasize that some functionality is disabled.

Here's `[UIColor ows_redColor]`:

screen shot 2017-04-05 at 11 29 12 am


WDYT?

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: Made a few changes here including adding asserts.

Signal/translations/en.lproj/Localizable.strings
charlesmchen: That's a really good idea. Done.

Signal/translations/en.lproj/Localizable.strings
charlesmchen: Yup.

screen shot 2017-04-05 at 11 34 27 am

Signal/src/ViewControllers/MessagesViewController.m
michaelkirk: :+1:

charlesmchen:
> Is it going to be clear enough to the user why we're popping the action sheet? E.g. we could be really explicit.

I had a different idea, since I tend to think long alert titles look cramped.  When we show these "you need to unblock to do that" alert, I now re-show the "block state indicator".

screen shot 2017-04-05 at 11 29 07 am

It clearly indicates the state (in a consistent way), it's presentation draws the eye to remind the user, and it offers the user a separate means to resolve the issue if the cancel.

WDYT?

charlesmchen:
> One issue I noticed is that if I block Alice, then message Alice, I see the "Unblock Alice?" dialog, as expected. When I accept, the message sends as expected, but when the action sheet dismisses, they text-input is not visible. (I think it's behind the keyboard).

Fixed.  I also avoided some keyboard animations around these "unblock" action sheets.

PR #1932, opened 2017-04-05
bail on scripts whenever an error is encountered

michaelkirk:
also use consistent bash

PTAL @charlesmchen


Signal/translations/bin/auto-genstrings
```diff
@@ -1,4 +1,8 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
+set -e
+set -x
```
charlesmchen: I wonder if we should add comments here?

PR #1933, opened 2017-04-05
Do not present local notifications if a read receipt arrives immediat…

charlesmchen:
…ely after.

This fixes: "my phone constantly vibrates while I'm chatting on desktop."

PTAL @michaelkirk

### DIFF
--- a/Signal/src/network/PushManager.m
+++ b/Signal/src/network/PushManager.m
@@ -23,6 +23,7 @@
 @property TOCFutureSource *registerWithServerFutureSource;
 @property UIAlertView *missingPermissionsAlertView;
 @property (nonatomic, retain) NSMutableArray *currentNotifications;
+@property (nonatomic, retain) NSMutableArray *pendingNotifications;
 @property (nonatomic) UIBackgroundTaskIdentifier callBackgroundTask;
 @property (nonatomic, readonly) OWSMessageSender *messageSender;
 @property (nonatomic, readonly) OWSMessageFetcherJob *messageFetcherJob;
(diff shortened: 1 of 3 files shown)


Signal/src/network/PushManager.m
```diff
@@ -429,19 +431,61 @@ - (BOOL)applicationIsActive {
...
+{
+    OWSAssert([NSThread isMainThread]);
+
+    NSString *threadId = notification.userInfo[Signal_Thread_UserInfo_Key];
+    if (checkForCancel && threadId != nil) {
+        [_pendingNotifications addObject:notification];
+
+        // The longer we wait, the more obsolete notifications we can suppress -
+        // but the more lag we introduce to notification delivery.
+        const CGFloat kDelaySeconds = 0.3f;
```
charlesmchen: A key line.

charlesmchen:
I don't understand how this issue doesn't affect more people on the team.

michaelkirk:
@charlesmchen what's the repro for this behavior?

Is it...

I am Alice, and I am linked with Desktop.
When I send messages to Bob from Desktop, my phone vibrates

Is that right?

Signal/src/network/PushManager.m
michaelkirk: Would it be simpler to modify the fireDate?

e.g. replace the entirety of this method implementation with:

```
- (void)presentNotification:(UILocalNotification *)notification checkForCancel:(BOOL)checkForCancel
{
    AssertIsOnMainThread();
    if (checkForCancel && threadId != nil) {
        _pendingNotification.fireDate = [NSDate dateWithTimeIntervalSinceNow:0.3f];
        _pendingNoficitation.timeZone = [NSTimeZone localTimeZone]; // IIRC this is necessary.
    }
    [[UIApplication sharedApplication] scheduleLocalNotification:notification];
    [self.currentNotifications addObject:notification];
}
```

Then most of the other changes could be omitted too.

charlesmchen: Marvelous, thank you.

PR #1934, opened 2017-04-05
Charlesmchen/blocking8

charlesmchen:
Refinements to "add to block list" view.

* Change the title to "Block".
* Add subtitles for the two sections and space separating them.
* Add a "you have no contacts" indicator.
* Show blocked contacts, but:
  * Decorate their cell with an indicator that they are blocked.
  * Don't let the user re-block a user already on the block list.

screen shot 2017-04-05 at 1 47 25 pm

PTAL @michaelkirk


charlesmchen:
> Is your intention that we'll also be able to use this indicator in the compose view contact picker? (That makes sense to me, and if not, we might want to reconsider how this is implemented)

Yup, I just haven't gotten there yet.  I'm about 1/4 of the way through the trello card.

PR #1935, opened 2017-04-05
right align blocked indicator

michaelkirk:
A proposed addendum to #1934 

My concern is that in the existing approach "(Blocked)" looks like someone's last name, as opposed to part of the interface. And in the case of mr. "Michael Really Long Last Name" I can't see that he's blocked, since it's truncated.

We might go even farther:

- color the text differently,
- size it smaller
- move it vertically out of line with the name
- add an icon

## Before:

before

## After:

after

WDYT @charlesmchen ?

### DIFF
--- a/Signal/src/views/ContactTableViewCell.m
+++ b/Signal/src/views/ContactTableViewCell.m
@@ -71,16 +71,19 @@ NS_ASSUME_NONNULL_BEGIN
     NSMutableAttributedString *attributedText =
         [[contactsManager formattedFullNameForContact:contact font:self.nameLabel.font] mutableCopy];
     if (self.isBlocked) {
-        // Add whitespace between the contact name and the blocked indicator.
-        [attributedText appendAttributedString:[[NSAttributedString alloc] initWithString:@" " attributes:nil]];
-        [attributedText appendAttributedString:[[NSAttributedString alloc]
-                                                   initWithString:NSLocalizedString(@"CONTACT_BLOCKED_INDICATOR",
-                                                                      @"An indicator that a contact has been blocked.")
-                                                       attributes:@{
-                                                           NSFontAttributeName : [UIFont
-                                                               ows_mediumFontWithSize:self.nameLabel.font.pointSize],
-                                                           NSForegroundColorAttributeName : [UIColor blackColor],
-                                                       }]];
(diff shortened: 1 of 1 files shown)


charlesmchen:
> color the text differently,

Yes, I think a middle gray?

> size it smaller

Yes. 10, 11 or 12pt.

> move it vertically out of line with the name

Why?

> add an icon

I've used a new "slashed circle" icon for blocking (see the 1:1 conversation settings).  I'd vote we _not_ add it here; these cells are cramped and I don't see how it'd add much.

Signal/src/views/ContactTableViewCell.m
```diff
@@ -71,16 +71,19 @@ - (void)configureWithContact:(Contact *)contact contactsManager:(OWSContactsMana
...
+            = NSLocalizedString(@"CONTACT_BLOCKED_INDICATOR", @"An indicator that a contact has been blocked.");
+        blockedLabel.font = [UIFont ows_mediumFontWithSize:self.nameLabel.font.pointSize];
+        blockedLabel.textColor = [UIColor blackColor];
+        [self addSubview:blockedLabel];
+        [blockedLabel sizeToFit];
+        [blockedLabel autoVCenterInSuperview];
+        [blockedLabel autoPinEdgeToSuperviewMargin:ALEdgeRight];
+        [blockedLabel autoPinEdge:ALEdgeLeft toEdge:ALEdgeRight ofView:self.textLabel];
+
+        self.accessoryView = blockedLabel;
```
charlesmchen: Wait, can't we skip all this:

```
+        [self addSubview:blockedLabel];
 +        [blockedLabel sizeToFit];
 +        [blockedLabel autoVCenterInSuperview];
 +        [blockedLabel autoPinEdgeToSuperviewMargin:ALEdgeRight];
 +        [blockedLabel autoPinEdge:ALEdgeLeft toEdge:ALEdgeRight ofView:self.textLabel];
 ```

And just do this:

```
 +        self.accessoryView = blockedLabel;
```

michaelkirk: Yes. =)

michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/e94cc18263df571d2999da8262185653a3d39aa8

PR #1936, opened 2017-04-06
Charlesmchen/blocking9

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/AddToBlockListViewController.m
+++ b/Signal/src/ViewControllers/AddToBlockListViewController.m
@@ -222,8 +222,8 @@ NSString *const kContactsTable_CellReuseIdentifier = @"kContactsTable_CellReuseI
 
     UILabel *label = [UILabel new];
     label.text = text;
-    label.font = [UIFont ows_boldFontWithSize:20.f];
-    label.textColor = [UIColor colorWithWhite:0.2f alpha:1.f];
+    label.font = [UIFont ows_mediumFontWithSize:20.f];
+    label.textColor = [UIColor colorWithWhite:0.3f alpha:1.f];
     label.textAlignment = NSTextAlignmentCenter;
     [row addSubview:label];
     [label autoPinEdgeToSuperviewEdge:ALEdgeLeft withInset:20.f];
@@ -489,7 +489,7 @@ NSString *const kContactsTable_CellReuseIdentifier = @"kContactsTable_CellReuseI
         UITableViewCell *cell = [UITableViewCell new];
         cell.textLabel.text = NSLocalizedString(
             @"SETTINGS_BLOCK_LIST_NO_CONTACTS", @"A label that indicates the user has no Signal contacts.");
-        cell.textLabel.font = [UIFont ows_regularFontWithSize:18.f];
+        cell.textLabel.font = [UIFont ows_regularFontWithSize:15.f];
         cell.textLabel.textColor = [UIColor colorWithWhite:0.5f alpha:1.f];
         cell.textLabel.textAlignment = NSTextAlignmentCenter;
         return cell;
@@ -518,6 +518,16 @@ NSString *const kContactsTable_CellReuseIdentifier = @"kContactsTable_CellReuseI
 }
 
 #pragma mark - UITableViewDelegate
+
+- (nullable NSIndexPath *)tableView:(UITableView *)tableView willSelectRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    if (self.hasNoContacts) {
+        // Don't let user select the "you have no contacts" cell.
+        return nil;
+    }
+
+    return indexPath;
+}
 
 - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath
 {
--- a/Signal/src/ViewControllers/BlockListViewController.m
+++ b/Signal/src/ViewControllers/BlockListViewController.m
@@ -96,12 +96,11 @@ typedef NS_ENUM(NSInteger, BlockListViewControllerSection) {
     }
 }
 
-- (nullable NSString *)tableView:(UITableView *)tableView titleForHeaderInSection:(NSInteger)section
+- (nullable NSString *)tableView:(UITableView *)tableView titleForFooterInSection:(NSInteger)section
 {
     switch (section) {
         case BlockListViewControllerSection_Add:
-            return NSLocalizedString(
-                                     @"SETTINGS_BLOCK_LIST_HEADER_TITLE", @"A header title for the block list table.");
+            return NSLocalizedString(@"SETTINGS_BLOCK_LIST_FOOTER_TITLE", @"A footer title for the block list table.");
         default:
             return nil;
     }
--- a/Signal/src/ViewControllers/MessageComposeTableViewController.m
+++ b/Signal/src/ViewControllers/MessageComposeTableViewController.m
@@ -3,9 +3,7 @@
 //
 
 #import "MessageComposeTableViewController.h"
-
-#import <MessageUI/MessageUI.h>
-
+#import "BlockListUIUtils.h"
 #import "ContactTableViewCell.h"
 #import "ContactsUpdater.h"
 #import "Environment.h"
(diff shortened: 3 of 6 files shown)


charlesmchen:
* Tweak appearance of “add to block list” view.
* Tweak appearance of contact cell.
* Update “new 1:1 conversation” view.
  * Don’t hide blocked users.  If selected, user needs to unblock to start conversation.
  * Add “you have no contacts” indicator.
  * Activate “block state indicator” in contact cell.

tweak2
tweak1

Signal/src/ViewControllers/AddToBlockListViewController.m
```diff
@@ -222,8 +222,8 @@ - (UIView *)createTitleRowWithText:(NSString *)text previousRow:(nullable UIView
 
     UILabel *label = [UILabel new];
     label.text = text;
-    label.font = [UIFont ows_boldFontWithSize:20.f];
-    label.textColor = [UIColor colorWithWhite:0.2f alpha:1.f];
+    label.font = [UIFont ows_mediumFontWithSize:20.f];
+    label.textColor = [UIColor colorWithWhite:0.3f alpha:1.f];
```
charlesmchen: The "add to block list" view's subtitles were too heavy to me.

Signal/src/views/ContactTableViewCell.m
```diff
@@ -75,13 +75,9 @@ - (void)configureWithContact:(Contact *)contact contactsManager:(OWSContactsMana
...
             = NSLocalizedString(@"CONTACT_BLOCKED_INDICATOR", @"An indicator that a contact has been blocked.");
-        blockedLabel.font = [UIFont ows_mediumFontWithSize:self.nameLabel.font.pointSize];
-        blockedLabel.textColor = [UIColor blackColor];
-        [self addSubview:blockedLabel];
+        blockedLabel.font = [UIFont ows_mediumFontWithSize:13.f];
+        blockedLabel.textColor = [UIColor colorWithWhite:0.5f alpha:1.f];
         [blockedLabel sizeToFit];
-        [blockedLabel autoVCenterInSuperview];
-        [blockedLabel autoPinEdgeToSuperviewMargin:ALEdgeRight];
-        [blockedLabel autoPinEdge:ALEdgeLeft toEdge:ALEdgeRight ofView:self.textLabel];
```
charlesmchen: You don't need to add `accessoryView` to the cell separately or manage its layout.  The cell does that for you. So when you set the label to be the `accessoryView` below, you're re-adding it to the view hierarchy and discarding these constraints.

I _like_ that you're using `accessoryView`, but there's an issue - we can't do that in the "new/edit group convo" view because we use "accessory" checkboxes there to indicate which users you're adding to the group.

charlesmchen:
I'm going to add in some copy changes to "block list" view.

screen shot 2017-04-05 at 9 57 55 pm

Signal/src/views/ContactTableViewCell.m
charlesmchen: > I like that you're using accessoryView, but there's an issue - we can't do that in the "new/edit group convo" view because we use "accessory" checkboxes there to indicate which users you're adding to the group.

Wait, I guess this isn't an issue.  A given signal user is either (already in the group/added to the group/blocked/none of the above).  A blocked user can't be in the group, so there's no conflict in terms of the accessories.

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -654,12 +685,36 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
     } else {
         OWSAssert(indexPath.section == MessageComposeTableViewControllerSectionContacts)
 
-        NSString *identifier = [[[self contactForIndexPath:indexPath] textSecureIdentifiers] firstObject];
+            if (!self.searchController.active && self.hasNoContacts)
```
michaelkirk: nit: weird indent.

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -654,12 +685,36 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
...
+                                         completion:^() {
+                                             [Environment messageIdentifier:contactIdentifier withCompose:YES];
+                                         }];
+        };
 
-        [self dismissViewControllerAnimated:YES
-                                 completion:^() {
-                                     [Environment messageIdentifier:identifier withCompose:YES];
-                                 }];
+        if ([self isContactBlocked:contact]) {
```
michaelkirk: It appears that you're asking the user to unblock the contact before you let them show the message view for that user. I don't think we should do this.

For one, I don't think its helpful. We *already* have a way of doing this by unblocking upon send.

Secondly, message compose contact picker is a way to "search" for old threads by contact. e.g. if I want to view my old thread with Bob, without scrolling around trying to find it in the inbox/archive, I come to the compose view and type "Bo..." and tap on Bob's avatar, which brings me to my existing thread with Bob. I shouldn't have to unblock him at this point before I can view our old thread.

Signal/src/views/ContactTableViewCell.m
```diff
@@ -75,13 +75,9 @@ - (void)configureWithContact:(Contact *)contact contactsManager:(OWSContactsMana
         blockedLabel.textAlignment = NSTextAlignmentRight;
         blockedLabel.text
             = NSLocalizedString(@"CONTACT_BLOCKED_INDICATOR", @"An indicator that a contact has been blocked.");
-        blockedLabel.font = [UIFont ows_mediumFontWithSize:self.nameLabel.font.pointSize];
-        blockedLabel.textColor = [UIColor blackColor];
-        [self addSubview:blockedLabel];
+        blockedLabel.font = [UIFont ows_mediumFontWithSize:13.f];
+        blockedLabel.textColor = [UIColor colorWithWhite:0.5f alpha:1.f];
```
michaelkirk: 👍

Signal/src/views/ContactTableViewCell.m
michaelkirk: You're right. Thanks.

Signal/src/ViewControllers/BlockListViewController.m
```diff
@@ -96,12 +96,11 @@ - (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger
...
 }
 
-- (nullable NSString *)tableView:(UITableView *)tableView titleForHeaderInSection:(NSInteger)section
+- (nullable NSString *)tableView:(UITableView *)tableView titleForFooterInSection:(NSInteger)section
 {
     switch (section) {
         case BlockListViewControllerSection_Add:
-            return NSLocalizedString(
-                                     @"SETTINGS_BLOCK_LIST_HEADER_TITLE", @"A header title for the block list table.");
+            return NSLocalizedString(@"SETTINGS_BLOCK_LIST_FOOTER_TITLE", @"A footer title for the block list table.");
```
michaelkirk: :+1: for the footer.

Signal/src/ViewControllers/MessageComposeTableViewController.m
charlesmchen: Done.

Signal/src/ViewControllers/MessageComposeTableViewController.m
charlesmchen: Point taken, I've reverted this change.  But I'm going to leave similar logic in the "new/edit group" view, since a) group members are unblocked by sending messages, etc. b) the "search" argument isn't relevant either.

michaelkirk: >  I'm going to leave similar logic in the "new/edit group" view, since 
> a) group members are unblocked by sending messages, etc. 

Is this a typo? Group members **are not** unblocked by sending messages. Right?

> b) the "search" argument isn't relevant either.


I agree that this behavior is a good feature for groups.

charlesmchen: Yup, typo. Sorry about that.

PR #1937, opened 2017-04-06
Improvements around contact cells.

charlesmchen:
* Improve handling of accessory views/types.
* Use contact cell in “new/edit group” view.
* Don’t hide blocked contacts and group members in “new/edit group” view.
* Let users “unblock-to-add” blocked contacts in “new/edit group” view.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/NewGroupViewController.m
+++ b/Signal/src/ViewControllers/NewGroupViewController.m
@@ -3,6 +3,8 @@
 //
 
 #import "NewGroupViewController.h"
+#import "BlockListUIUtils.h"
+#import "ContactTableViewCell.h"
 #import "Environment.h"
 #import "FunctionalUtil.h"
 #import "OWSContactsManager.h"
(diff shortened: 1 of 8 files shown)


charlesmchen:
img_9482

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -421,39 +454,111 @@ - (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger
...
+}
 
-    if ([[tableView indexPathsForSelectedRows] containsObject:indexPath]) {
-        [self adjustSelected:cell];
+- (void)updateContentsOfCell:(ContactTableViewCell *)cell indexPath:(NSIndexPath *)indexPath
+{
+    OWSAssert(cell);
+    OWSAssert(indexPath);
+
+    Contact *contact = contacts[(NSUInteger)indexPath.row];
```
michaelkirk: nit: `OWSAssert(contact != nil)`

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -421,39 +454,111 @@ - (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger
...
-}
+    Contact *contact = contacts[(NSUInteger)indexPath.row];
+    BOOL isBlocked = [self isContactBlocked:contact];
+    BOOL isInGroup = [self isContactInGroup:contact];
+    if (isInGroup) {
+        // Deselect.
+        [tableView deselectRowAtIndexPath:indexPath animated:YES];
+
+        NSString *displayName = [_contactsManager displayNameForContact:contact];
+        UIAlertController *controller = [UIAlertController
```
michaelkirk: just a thought:

I haven't run this yet, but is it possible or likely that the user is tapping on an *existing* member of the group in order to "deselect" them aka to "kick" them from the group?

In that case, maybe we should phrase the copy something like:

```
title: %@ is already in the group. 
body: You cannot kick people from a group. Either they have to leave, or you can start a new group without this user.
```

Pretty verbose, I know. =/

thoughts?

michaelkirk:
In dc174ad6

I left the "member is already in the group" copy as-is for now, but still kind of think it's a good idea.

WDYT @charlesmchen ?

PR #1938, opened 2017-04-06
Fix invalid assert in the OWS table views.

charlesmchen:
PTAL @michaelkirk


Signal/src/ViewControllers/OWSTableViewController.m
```diff
@@ -149,7 +149,6 @@ - (UITableViewCell *)customCell
     if (_customCellBlock) {
         return _customCellBlock();
     }
-    OWSAssert(0);
```
michaelkirk: What does it mean when this returns nil? 

Can you update the signature to return (nullable UITableViewCell *)?

charlesmchen: Done.

PR #1940, opened 2017-04-06
Dismiss keyboard if user scrolls in “new 1:1 conversation” view.

charlesmchen:
Same fix, another view.

PTAL @michaelkirk


michaelkirk:
In 80696e25

PR #1941, opened 2017-04-07
[SSK] Serialize read/write to session store

michaelkirk:
Update SSK to incorporate https://github.com/WhisperSystems/SignalServiceKit/pull/162

Included some relevant debug table methods.

PTAL @charlesmchen


Signal/src/ViewControllers/DebugUITableViewController.m
```diff
@@ -67,12 +67,29 @@ + (void)presentDebugUIForThread:(TSThread *)thread
...
-                                                                                      printAllSessions];
-                                                                              }] ]]];
+    [contents
+        addSection:[OWSTableSection
+                       sectionWithTitle:@"Session State"
+                                  items:@[
+                                      [OWSTableItem itemWithTitle:@"Print all sessions"
+                                                      actionBlock:^{
+                                                          dispatch_async([OWSDispatch sessionStoreQueue], ^{
+                                                              [[TSStorageManager sharedManager] printAllSessions];
```
charlesmchen: I wonder if we should make the usage of `sessionStoreQueue` an implementation detail of `TSStorageManager` and modify its methods to take completion handlers.

michaelkirk: Yeah, I really don't like how unstructured the current solution is.

Some of it's usages are not async though. Here's a good example of a non-trivial consequence of making the sessionStore async:
https://github.com/WhisperSystems/SignalProtocolKit/blob/master/AxolotlKit/Classes/SessionCipher.m#L106

So we'd have to change our encrypt/decrypt to be async/callback based as well.

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/acad91ebc1ef54ff208dcfb78b294b63785daf25

PR #1942, opened 2017-04-07
ensure session reset is decryptable

michaelkirk:
This is based on #1941. It's related UX-wise, but not actually caused by the same issue.

Typically we're sending an EndSession message because our session has
diverged from the remote party's session. So if we send an EndSession
message, but ~~decrypt~~ encrypt it with our old out-of-sync session, how can we
expect them to be able to decrypt it?

Instead, by deleting the existing sessions, we'll fetch a new PreKey before sending, 
then start fresh with the remote side.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/Jobs/SessionResetJob.swift
+++ b/Signal/src/Jobs/SessionResetJob.swift
@@ -1,5 +1,6 @@
-//  Created by Michael Kirk on 10/31/16.
-//  Copyright © 2016 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 import Foundation
 import PromiseKit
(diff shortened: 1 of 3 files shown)


charlesmchen:
> but decrypt it with our old out-of-sync session

You mean encrypt right?

Signal/src/Jobs/SessionResetJob.swift
```diff
@@ -24,32 +25,33 @@ class SessionResetJob: NSObject {
...
-        }, failure: {error in
-            Logger.error("\(self.TAG) failed to send EndSesionMessage with error: \(error.localizedDescription)")
-        });
+            DispatchQueue.main.async {
+                let endSessionMessage = EndSessionMessage(timestamp:NSDate.ows_millisecondTimeStamp(), in: self.thread)
+
+                self.messageSender.send(endSessionMessage, success: {
+                    Logger.info("\(self.TAG) successfully sent EndSession<essage.")
+
+                    Logger.info("\(self.TAG) deleting sessions for recipient: \(self.recipientId)")
```
charlesmchen: Should this log entry be up near `storageManager.deleteAllSessions`?

Signal/src/Jobs/SessionResetJob.swift
```diff
@@ -24,32 +25,33 @@ class SessionResetJob: NSObject {
...
+
+                    let message = TSInfoMessage(timestamp: NSDate.ows_millisecondTimeStamp(),
+                                                in: self.thread,
+                                                messageType: TSInfoMessageType.typeSessionDidEnd)
+                    message.save()
+                }, failure: {error in
+                    Logger.error("\(self.TAG) failed to send EndSesionMessage with error: \(error.localizedDescription)")
```
charlesmchen: s/EndSesionMessage/EndSessionMessage/

michaelkirk:
>> but decrypt it with our old out-of-sync session
> You mean encrypt right?

@charlesmchen - yes, you're right. Typo.

Signal/src/Jobs/SessionResetJob.swift
michaelkirk: done

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/e8056fcbb2b7d76841d49ef871ed80888811e793

Logging cleanup in:
https://github.com/WhisperSystems/Signal-iOS/commit/df381cf4b91895a2a19dd34ab06472cbd3160b88

PR #1944, opened 2017-04-07
Reformat phone number if user changes country code.

charlesmchen:
PTAL @michaelkirk


charlesmchen:

PR #1945, opened 2017-04-07
Ensure keyboard doesn't hide "safety numbers changed" interaction.

charlesmchen:
PTAL @michaelkirk


charlesmchen:

PR #1946, opened 2017-04-07
Round avatar icons.

charlesmchen:
PTAL @michaelkirk


charlesmchen:

charlesmchen:
> LGTM, see also AvatarImageView.swift

We have a bunch of redundant views for contacts and avatars. I think we have obj-c and swift contact cells - I'm not clear on how they are different.

PR #1947, opened 2017-04-07
Show attachment approval dialog for all attachments.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessagesViewController.m
+++ b/Signal/src/ViewControllers/MessagesViewController.m
@@ -2186,7 +2186,7 @@ typedef enum : NSUInteger {
                                              [self showErrorAlertForAttachment:attachment];
                                              failedToPickAttachment(nil);
                                          } else {
-                                             [self sendMessageAttachment:attachment];
+                                             [self tryToSendAttachmentIfApproved:attachment];
                                          }
                                      } else {
                                          failedToPickAttachment(nil);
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2210,35 +2210,32 @@ - (void)imagePickerController:(UIImagePickerController *)picker
 
 - (void)sendMessageAttachment:(SignalAttachment *)attachment
 {
+    OWSAssert([NSThread isMainThread]);
     // TODO: Should we assume non-nil or should we check for non-nil?
     OWSAssert(attachment != nil);
     OWSAssert(![attachment hasError]);
     OWSAssert([attachment mimeType].length > 0);
-    
-    DispatchMainThreadSafe(^{
```
charlesmchen: This method is now only called from `tryToSendAttachmentIfApproved:` which ensures that this will be called on the main thread.

charlesmchen:
Let's defer this until 2.11 - it touches too many code paths.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2767,34 +2764,42 @@ - (void)didPasteAttachment:(SignalAttachment * _Nullable)attachment {
...
+- (void)tryToSendAttachmentIfApproved:(SignalAttachment *_Nullable)attachment
+{
+    DDLogError(@"%@ %s", self.tag, __PRETTY_FUNCTION__);
+
+    DispatchMainThreadSafe(^{
+        if ([self isBlockedContactConversation]) {
+            __weak MessagesViewController *weakSelf = self;
+            [self showUnblockContactUI:^(BOOL isBlocked) {
+                if (!isBlocked) {
+                    [weakSelf didPasteAttachment:attachment];
```
michaelkirk: I think this should be:

```
[weakSelf tryToSendAttachmentifApproved:attachment];
```

PR #1950, opened 2017-04-07
Fix paste when pasteboard has multiple items.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
Low risk, let's take this in 2.10.

PR #1952, opened 2017-04-07
Reload data and invalidate layout whenever message view will appear.

charlesmchen:
PTAL @michaelkirk 

This is not for v2.10.  I'd like to merge this after and see if any problems emerge over the next sprint.

### DIFF
--- a/Signal/src/ViewControllers/MessagesViewController.m
+++ b/Signal/src/ViewControllers/MessagesViewController.m
@@ -524,7 +524,8 @@ typedef enum : NSUInteger {
                                                            ];
 
     [self ensureBlockStateIndicator];
-    
+
+    // Avoid layout corrupt issues and out-of-date message subtitles.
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -500,6 +500,9 @@ - (void)viewWillAppear:(BOOL)animated
                                                                                                                @"Short name for edit menu item to share contents of media message.")
                                                                                       action:shareSelector],
                                                            ];
+
+    [self.collectionView.collectionViewLayout invalidateLayout];
```
michaelkirk: maybe just a comment so someone isn't tempted to optimize it out later.

charlesmchen: FYI - I wonder if `toggleObservers:` isn't responsible for some of these issues.

PR #1953, opened 2017-04-08
Rework socket manager.

charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/164

This change is not for v2.10. We need to dogfood this extensively.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -140,38 +140,33 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
     // At this point, potentially lengthy DB locking migrations could be running.
     // Avoid blocking app launch by putting all further possible DB access in async thread.
-    UIApplicationState launchState = application.applicationState;
-    [[TSAccountManager sharedInstance] ifRegistered:YES runAsync:^{
-        if (launchState == UIApplicationStateInactive) {
-            DDLogWarn(@"The app was launched from inactive");
-            [TSSocketManager becomeActiveFromForeground];
-        } else if (launchState == UIApplicationStateBackground) {
-            DDLogWarn(@"The app was launched from being backgrounded");
-            [TSSocketManager becomeActiveFromBackgroundExpectMessage:NO];
-        } else {
-            DDLogWarn(@"The app was launched in an unknown way");
-        }
+    [[TSAccountManager sharedInstance]
+        ifRegistered:YES
+            runAsync:^{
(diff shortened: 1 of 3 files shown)


Signal/src/AppDelegate.m
```diff
@@ -140,38 +140,33 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
-    [[TSAccountManager sharedInstance] ifRegistered:YES runAsync:^{
-        if (launchState == UIApplicationStateInactive) {
-            DDLogWarn(@"The app was launched from inactive");
-            [TSSocketManager becomeActiveFromForeground];
-        } else if (launchState == UIApplicationStateBackground) {
-            DDLogWarn(@"The app was launched from being backgrounded");
-            [TSSocketManager becomeActiveFromBackgroundExpectMessage:NO];
-        } else {
-            DDLogWarn(@"The app was launched in an unknown way");
-        }
```
charlesmchen: Most of this is clang-format. The real change here is to replace this cascading if...else logic with:

```
[TSSocketManager requestSocketOpen];
```

Signal/src/AppDelegate.m
```diff
@@ -300,7 +295,6 @@ - (void)applicationWillResignActive:(UIApplication *)application {
           dispatch_sync(dispatch_get_main_queue(), ^{
               [self protectScreen];
               [[[Environment getCurrent] signalsViewController] updateInboxCountLabel];
-              [TSSocketManager resignActivity];
```
charlesmchen: `TSSocketManager` now observes application state and registration state events itself, so it doesn't need to be poked by the app delegate.

Signal/src/Jobs/MessageFetcherJob.swift
```diff
@@ -27,18 +28,18 @@ class MessageFetcherJob: NSObject {
         Logger.debug("\(TAG) \(#function)")
         guard signalService.isCensored  else {
             Logger.debug("\(self.TAG) delegating message fetching to SocketManager since we're using normal transport.")
-            TSSocketManager.becomeActive(fromBackgroundExpectMessage: true)
+            TSSocketManager.requestSocketOpen()
```
charlesmchen: This is the only change; the rest is from swift lint.

charlesmchen:
Not for v2.10. Benefits are very real but too sensitve to introduce this late.

PR #1954, opened 2017-04-08
Charlesmchen/flat avatar colors

charlesmchen:
A proposal - perhaps the team won't like it.  The intent is to make the default icons more cheerful, less drab.

PTAL @michaelkirk 

### Before

img_9485

### After

img_9488


charlesmchen:
Not for v2.10. Not even a production implementation. A conversation starter.

tijs:
Would love to see this implemented. May seem like a small thing but the current colors are indeed a bit depressing. Does Signal have any sort of style guide which might outline color schemes to be used etc?

PR #1957, opened 2017-04-08
Show block indicator in home view.

charlesmchen:
* Improve thread safety of home view cells.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/InboxTableViewCell.m
+++ b/Signal/src/ViewControllers/InboxTableViewCell.m
@@ -6,13 +6,14 @@
 #import "Environment.h"
 #import "OWSAvatarBuilder.h"
 #import "PropertyListPreferences.h"
+#import "Signal-Swift.h"
 #import "TSContactThread.h"
 #import "TSGroupThread.h"
 #import "TSMessagesManager.h"
 #import "Util.h"
-#import "Signal-Swift.h"
 #import <JSQMessagesViewController/JSQMessagesAvatarImageFactory.h>
 #import <JSQMessagesViewController/UIImage+JSQMessages.h>
+#import <SignalServiceKit/OWSBlockingManager.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/InboxTableViewCell.m
```diff
@@ -49,41 +50,56 @@ - (nullable NSString *)reuseIdentifier
     return NSStringFromClass(self.class);
 }
 
-- (void)configureWithThread:(TSThread *)thread contactsManager:(OWSContactsManager *)contactsManager
+- (void)configureWithThread:(TSThread *)thread
+            contactsManager:(OWSContactsManager *)contactsManager
+      blockedPhoneNumberSet:(NSSet<NSString *> *)blockedPhoneNumberSet;
 {
-    if (!_threadId || ![_threadId isEqualToString:thread.uniqueId]) {
-        dispatch_async(dispatch_get_main_queue(), ^{
```
charlesmchen: The home has been configuring its cells asynchronously.  This is something we'd like to avoid if possible because:

* It is likely to cause bad frames.
* It wasn't being done safely since cells can be reused.
* Most of the cells configuration should be cheap.

While doing this work, I discovered two other issues:

* We're not caching avatars.  That should be straightforward, and should be a significant perf boost.  For example we reload and configure the home view cells four times on app launch.  
* We're not persisting contact state, so we have a very perceptible "bad frame" on launch where we briefly show contacts as phone numbers.

charlesmchen: Oh, I'm wrong.  We _are_ caching avatars inside `OWSContactAvatarBuilder` using contactsManager.avatarCache.  I wonder if we need the to do anything off the main thread. I'll try to test on a 5c.

charlesmchen:
Not for v2.10.

charlesmchen:
PTAL @michaelkirk

Signal/src/ViewControllers/InboxTableViewCell.m
```diff
@@ -49,41 +50,56 @@ - (nullable NSString *)reuseIdentifier
...
-    NSString *snippetLabel = [[DisplayableTextFilter new] displayableText:thread.lastMessageLabel];
+    NSString *snippetLabel = (isBlocked ? NSLocalizedString(@"HOME_VIEW_BLOCKED_CONTACT_CONVERSATION",
+                                              @"A label for conversations with blocked users.")
+                                        : [[DisplayableTextFilter new] displayableText:thread.lastMessageLabel]);
 
     NSAttributedString *attributedDate = [self dateAttributedString:thread.lastMessageDate];
-    NSUInteger unreadCount             = [[TSMessagesManager sharedManager] unreadMessagesInThread:thread];
+    NSUInteger unreadCount = [[TSMessagesManager sharedManager] unreadMessagesInThread:thread];
 
-    dispatch_async(dispatch_get_main_queue(), ^{
```
michaelkirk: Moving the fast stuff out of async makes the initial repaint a *little* less painful, woo. 👍

PR #1958, opened 2017-04-09
Use numeric pad for verifiation code entry

michaelkirk:
No need for special phone characters like: *, -, +

PTAL @charlesmchen


michaelkirk:
> I've been kind of toying with the idea of using UIKeyboardTypePhonePad on the "registration" view and "block by phone number" view, and if the user taps "+", to open up the country code picker. WDYT?

Interesting idea! I'm hesitant to add what might be surprising behavior. That is, I think people have really specific expectations about what happens when they hit keys on the keyboard - it either enters some text, or modifies the keyboard. Popping up a view controller seems outside of that realm.

Also, I'd say the existing country-code-picking interface is pretty clear, so I'm not sure there's much value to add.

PR #1959, opened 2017-04-09
Don't repaint back-button unread badge

michaelkirk:
Just a little UX polish.

There's not much benefit to this being done async, and making it sync
avoids the repaint after the view appears.

Also, simplified the code around drawing the badge.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MessagesViewController.m
+++ b/Signal/src/ViewControllers/MessagesViewController.m
@@ -773,17 +773,22 @@ typedef enum : NSUInteger {
         _backButtonUnreadCountLabel.backgroundColor = [UIColor clearColor];
         _backButtonUnreadCountLabel.textColor = [UIColor whiteColor];
         _backButtonUnreadCountLabel.font = [UIFont systemFontOfSize:11];
+        _backButtonUnreadCountLabel.textAlignment = NSTextAlignmentCenter;
     }
     // This method gets called multiple times, so it's important we re-layout the unread badge
     // with respect to the new backItem.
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2679,57 +2691,27 @@ - (void)saveDraft {
...
+        return;
     }
+    _backButtonUnreadCount = unreadCount;
+
+    OWSAssert(_backButtonUnreadCountView != nil);
+    _backButtonUnreadCountView.hidden = unreadCount <= 0;
+
+    OWSAssert(_backButtonUnreadCountLabel != nil);
+    _backButtonUnreadCountLabel.text = [NSString stringWithFormat:@"%lu", (unsigned long)unreadCount];
+    [_backButtonUnreadCountLabel sizeToFit];
```
charlesmchen: If there any "max" value where we'd want to stop showing the values?  I think in some badging contexts people do MIN(99, actual_vaule) so that it tops out at 99 or whatever.

michaelkirk: Instead I've made sure we're stretching the background to accommodate multi digit label. (This was actually a regression, so thanks for bringing it up).

That should comfortably get us to 9999 unread messages, which ought to be enough for anybody.

charlesmchen: Great.

michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/4a128e69a6ae9a8550e249d17dc3322395cd61ab

PR #1960, opened 2017-04-10
Create block offer when non-contacts send you a message.

charlesmchen:
Depends on: https://github.com/WhisperSystems/SignalServiceKit/pull/165/files

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/BlockListUIUtils.h
+++ b/Signal/src/ViewControllers/BlockListUIUtils.h
@@ -44,6 +44,11 @@ typedef void (^BlockActionCompletionBlock)(BOOL isBlocked);
                           contactsManager:(OWSContactsManager *)contactsManager
                           completionBlock:(nullable BlockActionCompletionBlock)completionBlock;
 
+#pragma mark - UI Utils
+
++ (NSString *)formatDisplayNameForAlertTitle:(NSString *)displayName;
++ (NSString *)formatDisplayNameForAlertMessage:(NSString *)displayName;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Signal/src/ViewControllers/MessagesViewController.m
+++ b/Signal/src/ViewControllers/MessagesViewController.m
@@ -6,8 +6,8 @@
 #import "AppDelegate.h"
 #import "AttachmentSharing.h"
 #import "BlockListUIUtils.h"
-#import "DebugUITableViewController.h"
 #import "BlockListViewController.h"
+#import "DebugUITableViewController.h"
 #import "Environment.h"
 #import "FingerprintViewController.h"
 #import "FullImageViewController.h"
@@ -23,6 +23,7 @@
 #import "OWSIncomingMessageCollectionViewCell.h"
 #import "OWSMessagesBubblesSizeCalculator.h"
 #import "OWSOutgoingMessageCollectionViewCell.h"
+#import "OWSUnknownContactBlockOfferMessage.h"
 #import "PropertyListPreferences.h"
 #import "Signal-Swift.h"
 #import "SignalKeyingStorage.h"
@@ -383,6 +384,14 @@ typedef enum : NSUInteger {
     self.senderDisplayName = ME_MESSAGE_IDENTIFIER;
 
     [self initializeToolbars];
+
+    if ([self.thread isKindOfClass:[TSContactThread class]]) {
+        TSContactThread *contactThread = (TSContactThread *)self.thread;
+        [ThreadUtil createBlockOfferIfNecessary:contactThread
+                                 storageManager:self.storageManager
+                                contactsManager:self.contactsManager
+                                blockingManager:self.blockingManager];
+    }
 }
 
 - (void)viewDidLayoutSubviews
@@ -2003,6 +2012,8 @@ typedef enum : NSUInteger {
 {
     if ([message isKindOfClass:[TSInvalidIdentityKeyErrorMessage class]]) {
         [self tappedInvalidIdentityKeyErrorMessage:(TSInvalidIdentityKeyErrorMessage *)message];
+    } else if ([message isKindOfClass:[OWSUnknownContactBlockOfferMessage class]]) {
+        [self tappedUnknownContactBlockOfferMessage:(OWSUnknownContactBlockOfferMessage *)message];
     } else if (message.errorType == TSErrorMessageInvalidMessage) {
         [self tappedCorruptedMessage:message];
     } else {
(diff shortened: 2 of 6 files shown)


Signal/src/ViewControllers/BlockListUIUtils.h
```diff
@@ -44,6 +44,11 @@ typedef void (^BlockActionCompletionBlock)(BOOL isBlocked);
                           contactsManager:(OWSContactsManager *)contactsManager
                           completionBlock:(nullable BlockActionCompletionBlock)completionBlock;
 
+#pragma mark - UI Utils
+
++ (NSString *)formatDisplayNameForAlertTitle:(NSString *)displayName;
++ (NSString *)formatDisplayNameForAlertMessage:(NSString *)displayName;
```
charlesmchen: Eventually it'd be nice to move these functions to a new `AlertUtils` (or whatever) file.

Signal/src/util/ThreadUtil.m
```diff
@@ -84,6 +88,112 @@ + (void)sendMessageWithAttachment:(SignalAttachment *)attachment
...
 
++ (void)createBlockOfferIfNecessary:(TSContactThread *)contactThread
+                     storageManager:(TSStorageManager *)storageManager
+                    contactsManager:(OWSContactsManager *)contactsManager
+                    blockingManager:(OWSBlockingManager *)blockingManager
+{
+    OWSAssert(contactThread);
+    OWSAssert(storageManager);
+    OWSAssert(contactsManager);
+    OWSAssert(blockingManager);
```
charlesmchen: Since we call this method every time we open a 1:1 conversation, performance matters.  I've done what I can to mitigate the perf cost; I intend to measure the cost for worst-case conversations (e.g. conversations with 10k incoming messages and no outgoing messages).

charlesmchen:
Not for v2.10.

Signal/src/util/ThreadUtil.m
```diff
@@ -84,6 +88,112 @@ + (void)sendMessageWithAttachment:(SignalAttachment *)attachment
...
+
+        DDLogInfo(@"Creating block offer for unknown contact");
+
+        // We want the block offer to be the first interaction in their
+        // conversation's timeline, so we back-date it to slightly before
+        // the first incoming message (which we know is the first message).
+        TSIncomingMessage *firstMessage = firstIncomingMessage;
+        uint64_t blockOfferTimestamp = firstMessage.timestamp - 1;
+
+        // Create an error message we may or may not user.
```
michaelkirk: nit: typo, use~~r~~

michaelkirk: > Create an error message we may or may not use

clarity: What does this mean? Maybe it would make sense if I know who "we" is - the app or the user?

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2020,48 +2031,83 @@ - (void)tappedInvalidIdentityKeyErrorMessage:(TSInvalidIdentityKeyErrorMessage *
...
-                                                                               }
-                                                                               failure:^(NSError *_Nonnull error) {
-                                                                                   DDLogError(@"%@ Failed to resend key-error message with error:%@",
-                                                                                              self.tag,
-                                                                                              error);
-                                                                               }];
-                                                                          }
-                                                                      }];
+
+    UIAlertAction *acceptSafetyNumberAction = [UIAlertAction
```
michaelkirk: assuming the three previous blocks (around SN changes) is just clang-format.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2020,48 +2031,83 @@ - (void)tappedInvalidIdentityKeyErrorMessage:(TSInvalidIdentityKeyErrorMessage *
...
+        [UIAlertAction actionWithTitle:NSLocalizedString(@"BLOCK_OFFER_ACTIONSHEET_BLOCK_ACTION",
+                                           @"Action sheet that will block an unknown user.")
+                                 style:UIAlertActionStyleDestructive
+                               handler:^(UIAlertAction *_Nonnull action) {
+                                   DDLogInfo(@"%@ Blocking an unknown user.", self.tag);
+                                   [self.blockingManager addBlockedPhoneNumber:errorMessage.contactId];
+                                   // Delete the block offer.
+                                   [self.storageManager.dbConnection
+                                       readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
+                                           [errorMessage removeWithTransaction:transaction];
```
michaelkirk: I haven't tried to use this yet, but it seems like the block offer would reappear when the unknown-user sends a new message, which seems undesirable, since the receiving user has stated the don't want to block the sending user.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1021,6 +1027,9 @@
 /* In Inbox view, last message label for thread with corrupted attachment. */
 "UNKNOWN_ATTACHMENT_LABEL" = "Unknown attachment";
 
+/* No comment provided by engineer. */
+"UNKNOWN_CONTACT_BLOCK_OFFER" = "User not in your contacts. Would you like to block this user?";
```
michaelkirk: nicely worded.

Signal/src/util/ThreadUtil.m
```diff
@@ -84,6 +88,112 @@ + (void)sendMessageWithAttachment:(SignalAttachment *)attachment
...
+
+        NSMutableArray *blockOffers = [NSMutableArray new];
+        [[transaction ext:TSMessageDatabaseViewExtensionName]
+            enumerateRowsInGroup:contactThread.uniqueId
+                      usingBlock:^(
+                          NSString *collection, NSString *key, id object, id metadata, NSUInteger index, BOOL *stop) {
+
+                          if ([object isKindOfClass:[OWSUnknownContactBlockOfferMessage class]]) {
+                              hasUnknownContactBlockOffer = YES;
+                              [blockOffers addObject:object];
```
michaelkirk: I think we should `*stop = YES` at this point

Signal/src/util/ThreadUtil.m
```diff
@@ -84,6 +88,112 @@ + (void)sendMessageWithAttachment:(SignalAttachment *)attachment
...
+                          } else if ([object isKindOfClass:[TSIncomingMessage class]]) {
+                              TSIncomingMessage *incomingMessage = (TSIncomingMessage *)object;
+                              if (!firstIncomingMessage ||
```
michaelkirk: Our DB already ordered the objects in this extension, so we shouldn't have to do any sorting in the application, ditto for the outgoing messages.

Signal/src/util/ThreadUtil.m
charlesmchen: Obsolete, removed.

Signal/src/util/ThreadUtil.m
charlesmchen: Done.

Signal/src/util/ThreadUtil.m
charlesmchen: Oh nice!

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: Yes, I don't know why it went there.

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: > it seems like the block offer would reappear when the unknown-user sends a new message, 

They won't reappear until the user leaves and re-enters the 1:1 conversation view.  The idea is that it wouldn't matter much since they'd always be the first interaction in the conversation, so the block offer would almost always be offscreen once the conversation got going.  And I was trying to avoid adding more state.  But now you have me wondering... (see below).

charlesmchen:
> nit: What if the user is already blocked (e.g. through settings>privacy)? Showing a redundant offer to block might confuse users into thinking they hadn't yet successfully blocked the user.

`ThreadUtil createBlockOfferIfNecessary:` already checks for this.

charlesmchen:
> Take it or leave it: I'm not sure that the "error box" at the top of the thread is the right way to present this to a user. e.g. compare to iMessage, which presents this as text button at the bottom of the conversation view, making it feel more like interface than like part of the conversation content.

Yes, I went back and forth on this quite a bit on how to present these "block offers".  

I think there's more than one question:

* Should the offer be an interaction in the conversation or a view element/conversation overlay?
* Should the offer show up at the "top of view/top of conversation" or the  "bottom of view/bottom of conversation". 
* Should the offer disappear forever if accepted/declined?
* How much should we emphasize the block offer (e.g. info vs. error)?

And my goals were:

* Make the meaning of the offer be clear.
* Make the offer something that "went away if ignored," and not force the user to accept/decline it.
* Make the offer something hard to accept by accident.
* Not worry too much about making sure the user always saw the offer, since there are many other ways to block users.

> * Should the offer show up at the "top of view/top of conversation" or the  "bottom of view/bottom of conversation". 

With this in mind, I wanted to avoid having it at the bottom of the screen. In my mind, you want the most-used and safest UI elements clustered near the thumb's "range of easy movement" and the rare and unsafe element at the top of the screen, away from the thumb.  Since blocking is "destructive", I want it towards the top where it's unlikely to be tapped by accident.

> * Should the offer be an interaction in the conversation or a view element/conversation overlay?

In terms of presenting it as a view vs. an interaction, I liked the interaction since that would satisfy the goals of having the offer "go away if ignored," since the conversation would quickly push it offscreen.

> * How much should we emphasize the block offer (e.g. info vs. error)?

In terms of info vs. error, I guess info could be better.  My understanding is that our "error" interactions tend to be tappable and our "info" interactions aren't.  I was trying to be consistent with the user expectation of what's tappable.

> * Should the offer disappear forever if accepted/declined?

We could do this, but I was hoping it wasn't necessary.  In my experience typical conversations generate a pageful of content very quickly.

Let me know what you think @michaelkirk

charlesmchen:
PTAL @michaelkirk

PR #1961, opened 2017-04-10
Improve attachment upload progress indicator.

charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/168

* Should never show up for attachments in the failed state.
* Show show up without delay for attachments being sent.
* Show disappear when attachment upload fails in any way.

PTAL @michaelkirk 

For v2.11.


michaelkirk:
LGTM

PR #1965, opened 2017-04-10
Add explanation messages to the “block user alert” and the block sect…

charlesmchen:
…ion of the 1:1 conversation settings view.

PTAL @michaelkirk 

I also moved the "block" row of the 1:1 conversation settings view to its own group, to "set it apart" since it's destructive.

screen shot 2017-04-10 at 3 57 27 pm

screen shot 2017-04-10 at 3 57 23 pm

### DIFF
--- a/Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
+++ b/Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
@@ -261,8 +261,8 @@ static NSString *const OWSConversationSettingsTableViewControllerSegueShowGroupM
                                                                items:@[
                                                                    item,
                                                                ]];
-        section.footerTitle = NSLocalizedString(@"CONVERSATION_SETTINGS_BLOCK_FOOTER_TITLE",
-            @"A footer title for the block user option in the conversation settings");
+        section.footerTitle = NSLocalizedString(@"BLOCK_BEHAVIOR_EXPLANATION",
+            @"An explanation of the consequences of blocking another user.");
         [contents addSection:section];
     }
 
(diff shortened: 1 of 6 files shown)


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -97,6 +97,9 @@
 /* No comment provided by engineer. */
 "ATTACHMENT_QUEUED" = "New attachment queued for retrieval.";
 
+/* The message of the 'block user' action sheet. */
+"BLOCK_LIST_BLOCK_ALERT_MESSAGE" = "Blocked users will not be able to call you or send you messages.";
```
michaelkirk: I think we should should de-dupe this and CONVERSATION_SETTINGS_BLOCK_FOOTER_TITLE with the existing SETTINGS_BLOCK_LIST_FOOTER_TITLE so as to not make unnecessary work for the translators, and consequently reduce the surface for translation errors.

Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
```diff
@@ -209,10 +211,30 @@ - (void)updateTableContents
                                          }]];
     }
 
+    [firstSectionItems addObject:[OWSTableItem itemWithCustomCellBlock:^{
```
michaelkirk: yay OWSTableViewController.

charlesmchen: There are certain pieces of code I have rewritten from scratch every time I switch employers. This class is (part of) one of them.

Signal/translations/en.lproj/Localizable.strings

PR #1966, opened 2017-04-10
Stay scrolled to the bottom during in conversation view during updates.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
> To be clear, this is to fix the issue where downloaded images hang below the current scroll, right?

Downloaded and uploaded images (portrait and landscape), incoming and outgoing audio and video attachments, large text messages. There were a lot of ways to end up with "new content under the fold".

PR #1970, opened 2017-04-11
make navbar legible when sharing via iMessage

michaelkirk:
When sharing out attachments, we can't control the style of the MFMessage navbar (AFAICT).

So the work around is to temporarily toggle styles so the buttons become legible. 

The appearance switch is a little jarring, but I'm not sure of a better approach.

signal-2017-03-31-143606

### DIFF
--- a/Signal/src/Models/TSMessageAdapaters/TSMessageAdapter.m
+++ b/Signal/src/Models/TSMessageAdapaters/TSMessageAdapter.m
@@ -312,6 +312,7 @@ NS_ASSUME_NONNULL_BEGIN
         actionString,
         self.interaction.uniqueId,
         [self.mediaItem class]);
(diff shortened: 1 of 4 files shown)


Signal/src/Models/TSMessageAdapaters/TSMessageAdapter.m
```diff
@@ -312,6 +312,7 @@ - (void)performEditingAction:(SEL)action
         actionString,
         self.interaction.uniqueId,
         [self.mediaItem class]);
+    OWSAssert(NO);
```
charlesmchen: clang-format?

michaelkirk:
In 37d408cc743936497c402dd24e05650233268daf

PR #1971, opened 2017-04-11
registration: clarify verification process

michaelkirk:
Two simple changes to make our verification screen a little more intuitive and useful.

5c
screen shot 2017-04-11 at 1 49 52 pm

7+
screen shot 2017-04-11 at 1 48 50 pm

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/CodeVerificationViewController.m
+++ b/Signal/src/ViewControllers/CodeVerificationViewController.m
@@ -126,11 +126,15 @@ NSString *const kCompletedRegistrationSegue = @"CompletedRegistration";
     _phoneNumberLabel = [UILabel new];
     _phoneNumberLabel.textColor = [UIColor ows_darkGrayColor];
     _phoneNumberLabel.font = [UIFont ows_regularFontWithSize:20.f];
+    _phoneNumberLabel.numberOfLines = 2;
+    _phoneNumberLabel.adjustsFontSizeToFitWidth = YES;
     [self.view addSubview:_phoneNumberLabel];
-    [_phoneNumberLabel autoHCenterInSuperview];
-    [_phoneNumberLabel autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:header
-                          withOffset:ScaleFromIPhone5To7Plus(25, 100)];
-    
+    [_phoneNumberLabel autoPinWidthToSuperviewWithMargin:ScaleFromIPhone5(32)];
+    [_phoneNumberLabel autoPinEdge:ALEdgeTop
+                            toEdge:ALEdgeBottom
+                            ofView:header
+                        withOffset:ScaleFromIPhone5To7Plus(30, 100)];
+
     const CGFloat kHMargin = 36;
     
     _challengeTextField = [UITextField new];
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/CodeVerificationViewController.m
```diff
@@ -104,7 +104,7 @@ - (void)createViews {
 
     UILabel *titleLabel = [UILabel new];
     titleLabel.textColor = [UIColor whiteColor];
-    titleLabel.text = NSLocalizedString(@"VERIFICATION_HEADER", @"Navigation title in the registration flow - during the sms code verification process.");
+    titleLabel.text = [self phoneNumberText];
```
charlesmchen: Curious - what motivated this change?

Signal/src/ViewControllers/CodeVerificationViewController.m
```diff
@@ -126,11 +126,15 @@ - (void)createViews {
     _phoneNumberLabel = [UILabel new];
     _phoneNumberLabel.textColor = [UIColor ows_darkGrayColor];
     _phoneNumberLabel.font = [UIFont ows_regularFontWithSize:20.f];
+    _phoneNumberLabel.numberOfLines = 2;
+    _phoneNumberLabel.adjustsFontSizeToFitWidth = YES;
```
charlesmchen: Center-align text of `_phoneNumberLabel`?

Signal/src/ViewControllers/CodeVerificationViewController.m
```diff
@@ -227,12 +231,18 @@ - (void)createViews {
...
 
-- (void)updatePhoneNumberLabel {
-    NSString *phoneNumber = [PhoneNumber bestEffortFormatPartialUserSpecifiedTextToLookLikeAPhoneNumber:[TSAccountManager localNumber]];
+- (NSString *)phoneNumberText
+{
     OWSAssert([TSAccountManager localNumber] != nil);
-    _phoneNumberLabel.text = [NSString stringWithFormat:NSLocalizedString(@"VERIFICATION_PHONE_NUMBER_FORMAT",
-                                                                         @"Label indicating the phone number currently being verified."),
-                             phoneNumber];
+    return [PhoneNumber bestEffortFormatPartialUserSpecifiedTextToLookLikeAPhoneNumber:[TSAccountManager localNumber]];
```
charlesmchen: I think if we use this:
```
        [PhoneNumber bestEffortFormatPartialUserSpecifiedTextToLookLikeAPhoneNumber:[TSAccountManager localNumber]
                                                     withSpecifiedCountryCodeString:countryCode];
```
The phone number format will reflect your region.

Signal/src/ViewControllers/CodeVerificationViewController.m
michaelkirk: I thought the previous header, "Verify" wasn't adding any value. 

If the number was entered incorrectly, I thought putting it near the back button would draw the eye nearer to the next step (going "back" and fixing it.)

Signal/src/ViewControllers/CodeVerificationViewController.m
michaelkirk: Oh, I see what you're saying. If I specified the users country code, we can be sure we aren't just formatting their number in their default locale.

Signal/src/ViewControllers/CodeVerificationViewController.m
michaelkirk: I intentionally left-aligned it since it was more than one line and presented as a sentence (with a ".").

Do you think it would look better centered (and without a period)?

PR #1972, opened 2017-04-11
[JSQMVC] Fixes "floating" edit menu for narrow media items

michaelkirk:
previously, edit menu was only positioned correctly for text bubbles,
and centered for media bubbles.

It wasn't that noticeable for images/videos since those usually extended
to the middle of the conversation view, but with narrower media bubbles
(like arbitrary attachments) this became a bigger problem.

upstream PR: https://github.com/jessesquires/JSQMessagesViewController/pull/2069

PTAL @charlesmchen


michaelkirk:
In 88d99b7a

PR #1973, opened 2017-04-11
use updated localizable strings linter

michaelkirk:
this script checks to make sure the proper number of format characters
appear in our translated strings.

See: https://github.com/WhisperSystems/l10n_lint

PTAL @charlesmchen


michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/1e47f14bddba06eaa2350607e971c847001aa67a

PR #1976, opened 2017-04-12
Charlesmchen/arbitrary attachments

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/SignalAttachment.swift
+++ b/Signal/src/ViewControllers/SignalAttachment.swift
@@ -243,10 +243,7 @@ class SignalAttachment: NSObject {
     }
 
     public class func pasteboardHasPossibleAttachment() -> Bool {
-        guard UIPasteboard.general.numberOfItems >= 1 else {
-            return false
-        }
-        return true
+        return UIPasteboard.general.numberOfItems > 0
     }
 
     // Returns an attachment from the pasteboard, or nil if no attachment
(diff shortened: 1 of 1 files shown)


charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/174

Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -249,13 +216,7 @@ class SignalAttachment: NSObject {
         guard UIPasteboard.general.numberOfItems >= 1 else {
```
michaelkirk: nit: a little simpler if this whole method is replaced by:

```
return UIPasteboard.general.numberOfItems > 0
```

Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -300,9 +261,13 @@ class SignalAttachment: NSObject {
                 return audioAttachment(data : data, dataUTI : dataUTI)
             }
         }
-        // TODO: We could handle generic attachments at this point.
 
-        return nil
+        let dataUTI = pasteboardUTISet[pasteboardUTISet.startIndex]
+        guard let data = dataForFirstPasteboardItem(dataUTI:dataUTI) else {
+            Logger.verbose("\(TAG) Missing expected pasteboard data for UTI: \(dataUTI)")
```
michaelkirk: nit: This *seems* like an error condition? 

If so, I'd say a Logger.error and assertionFailure are in order.

Signal/src/ViewControllers/SignalAttachment.swift
charlesmchen: Done.

Signal/src/ViewControllers/SignalAttachment.swift
charlesmchen: Done.

PR #1977, opened 2017-04-13
Feature/contacts intersection audit

charlesmchen:
Reduces our usage of the contacts intersection endpoint. Should have no impact on the user experience.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/contact/OWSContactsManager.m
+++ b/Signal/src/contact/OWSContactsManager.m
@@ -16,12 +16,13 @@ NSString *const OWSContactsManagerSignalRecipientsDidChangeNotification =
 
 @interface OWSContactsManager ()
 
-@property id addressBookReference;
-@property TOCFuture *futureAddressBook;
-@property ObservableValueController *observableContactsController;
-@property TOCCancelTokenSource *life;
+@property (atomic) id addressBookReference;
+@property (atomic) TOCFuture *futureAddressBook;
+@property (atomic) ObservableValueController *observableContactsController;
+@property (atomic) TOCCancelTokenSource *life;
 @property (atomic) NSDictionary *latestContactsById;
 @property (atomic) NSDictionary<NSString *, Contact *> *contactMap;
+@property (nonatomic) BOOL isContactsUpdateInFlight;
 
 @end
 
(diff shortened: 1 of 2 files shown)


Signal/src/contact/OWSContactsManager.m
```diff
@@ -131,34 +131,70 @@ - (void)handleAddressBookChanged
 
 #pragma mark - Setup
 
-- (void)setupAddressBook {
+- (void)setupAddressBookIfNecessary
+{
     dispatch_async(ADDRESSBOOK_QUEUE, ^{
```
charlesmchen: Both `[application: didFinishLaunchingWithOptions:]` and `[applicationDidBecomeActive:]` both prod `OWSContactsManager` to update its contacts, so on app launch we do this twice, which doubles our bulk usage of the contacts intersection endpoint.  This logic de-bounces that.

Signal/src/contact/OWSContactsManager.m
```diff
@@ -107,9 +109,7 @@ - (void)doAfterEnvironmentInitSetup {
 }
 
 - (void)verifyABPermission {
-    if (!self.addressBookReference) {
```
charlesmchen: We can't debounce here due to concurrency. We need to do it inside `setupAddressBookIfNecessary`, using the `ADDRESSBOOK_QUEUE` to avoid race conditions.

Signal/src/contact/OWSContactsManager.m
```diff
@@ -131,34 +131,70 @@ - (void)handleAddressBookChanged
...
+        [self fireSignalRecipientsDidChange];
+    };
+    void (^failure)(NSError *error) = ^(NSError *error) {
+        DDLogWarn(@"%@ Failed to intersect contacts with error: %@. Rescheduling", self.tag, error);
+
+        // Retry with exponential backoff.
+        //
+        // TODO: Abort if another contact
+        // intersection succeeds in the meantime.
+        dispatch_after(
```
charlesmchen: Only failure we were just hammering the service with retries, even if we hit the rate limit.  Hmm, I should modify this to abort if we get a status code that indicates the rate limit has been hit.

charlesmchen: Done.

charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/175

Signal/src/contact/OWSContactsManager.m
```diff
@@ -131,34 +131,71 @@ - (void)handleAddressBookChanged
...
+
+        if ([error.domain isEqualToString:kContactsUpdaterErrorDomain]
+            && error.code == kContactsUpdaterRateLimitErrorCode) {
+            DDLogError(@"Contact intersection hit rate limit with error: %@", error);
+            return;
+        }
+
+        DDLogWarn(@"%@ Failed to intersect contacts with error: %@. Rescheduling", self.tag, error);
+
+        // Retry with exponential backoff.
```
michaelkirk: This doesn't look exponential?

charlesmchen: Done.

PR #1978, opened 2017-04-13
Charlesmchen/honor attachment filenames

charlesmchen:
* Send filenames with outgoing attachments, if possible.
* Store incoming attachment filenames, if present.
* Write attachment data to disk using the attachment's filename, if possible.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessagesViewController.m
+++ b/Signal/src/ViewControllers/MessagesViewController.m
@@ -2182,8 +2182,8 @@ typedef enum : NSUInteger {
 
     NSURL *referenceURL = [info valueForKey:UIImagePickerControllerReferenceURL];
     if (!referenceURL) {
-        DDLogError(@"Could not retrieve reference URL for picked asset");
-        OWSAssert(0);
+        DDLogVerbose(@"Could not retrieve reference URL for picked asset");
+        [self imagePickerController:picker didFinishPickingMediaWithInfo:info filename:nil];
         return;
     }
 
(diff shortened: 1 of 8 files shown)


charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/177

michaelkirk:
Actually, this doesn't compile for me, but just missing a trivial change in NewGroupViewController

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2193,11 +2194,37 @@ - (void)resetFrame {
...
 {
-    OWSAssert([NSThread isMainThread]);
-    
     [UIUtil modalCompletionBlock]();
     [self resetFrame];
 
+    NSURL *referenceURL = [info valueForKey:UIImagePickerControllerReferenceURL];
+    if (!referenceURL) {
+        DDLogError(@"Could not retrieve reference URL for picked asset");
+        OWSAssert(0);
```
michaelkirk: I hit this assertion failure on iOS9.3 (haven't tested on 10).

**steps:**

- Go to thread
- Paperclip
- Camera
- take a photo 
- use photo

charlesmchen:
> Actually, this doesn't compile for me, but just missing a trivial change in NewGroupViewController

Fixed.

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: Fixed.

PR #1979, opened 2017-04-13
Send oversize text messages as attachments.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessagesViewController.m
+++ b/Signal/src/ViewControllers/MessagesViewController.m
@@ -1098,22 +1098,21 @@ typedef enum : NSUInteger {
 
     text = [text stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
 
-    // Limit outgoing text messages to 16kb.
-    //
-    // We convert large text messages to attachments
-    // which are presented as normal text messages.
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -1101,28 +1101,15 @@ - (void)didPressSendButton:(UIButton *)button
...
-                         completion:nil];
-        return;
-    }
-    
-    if (text.length > 0) {
+    const NSUInteger kOversizeTextMessageSizeThreshold = 16 * 1024;
+    if ([text lengthOfBytesUsingEncoding:NSUTF8StringEncoding] >= kOversizeTextMessageSizeThreshold) {
+        SignalAttachment *attachment = [SignalAttachment oversizeTextAttachmentWithText:text];
+        [ThreadUtil sendMessageWithAttachment:attachment inThread:self.thread messageSender:self.messageSender];
+    } else if (text.length > 0) {
```
michaelkirk: whoops, good catch here. But I think this should be structured a little differently as the soundInForeground, updateKeyboardState, and finishSendingMessage should happen for both oversize and regular text, right?

charlesmchen: Good point!

PR #1980, opened 2017-04-13
Charlesmchen/hardware ringer vs playback

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/AttachmentApprovalViewController.swift
+++ b/Signal/src/ViewControllers/AttachmentApprovalViewController.swift
@@ -43,12 +43,24 @@ class AttachmentApprovalViewController: UIViewController {
         view.backgroundColor = UIColor.black
 
         self.navigationItem.leftBarButtonItem = UIBarButtonItem(barButtonSystemItem:.stop,
-            target:self,
-            action:#selector(donePressed))
+                                                                target:self,
+                                                                action:#selector(donePressed))
         self.navigationItem.title = NSLocalizedString("ATTACHMENT_APPROVAL_DIALOG_TITLE",
                                                       comment: "Title for the 'attachment approval' dialog.")
 
         createViews()
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/AttachmentApprovalViewController.swift
```diff
@@ -42,14 +42,26 @@ class AttachmentApprovalViewController: UIViewController {
...
+                                                                target:self,
+                                                                action:#selector(donePressed))
         self.navigationItem.title = NSLocalizedString("ATTACHMENT_APPROVAL_DIALOG_TITLE",
                                                       comment: "Title for the 'attachment approval' dialog.")
 
         createViews()
     }
 
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
```
michaelkirk: typo: `super.viewWillAppear`

charlesmchen:
> Does this do the right thing WRT voice notes?

I'm not sure how you mean - we can't record voice notes locally yet.  In many cases where we play or record audio, we update the category of `AVAudioSession` before doing so.  If we're missing that anywhere, that isn't a regression - we're already changing the category in a few places.  Nonetheless the changes here - to make sure we're using the right category during video playback - take care to restore the category to a reasonable default after the playback is done.  I'm using `AVAudioSessionCategoryPlayAndRecord` - but I noticed that CallAudioService sets the category to `AVAudioSessionCategoryAmbient` when a call ends.  

> If you get a call while playing a video it will ring loudly right?

That's `CallAudioService`'s responsibility, and it's `startRinging` method makes sure to use AVAudioSessionCategoryAmbient so that we honor the mute switch.

charlesmchen:
Note that this work doesn't do anything around playback of audio - that's a separate task.

Signal/src/ViewControllers/AttachmentApprovalViewController.swift
charlesmchen: Done.

michaelkirk:
> Note that this work doesn't do anything around playback of audio - that's a separate task.

Ok, that explains my confusion.

PR #1983, opened 2017-04-14
local users registration ID in debug log

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -143,6 +143,9 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     [[TSAccountManager sharedInstance]
         ifRegistered:YES
             runAsync:^{
+                DDLogInfo(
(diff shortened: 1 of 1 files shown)


michaelkirk:
> On Apr 14, 2017, at 10:50 AM, Matthew Chen <notifications@github.com> wrote:
> 
> @charlesmchen approved this pull request.
> 
> LGTM
> 
> I've been wondering if we should add a bunch of info to the debug logs, like the device model, iOS version, etc., etc. WDYT?
> 
> 

The top of the submitted log already contains a line like "iPhone Version: iPhone8,1, iOS Version: 10.3”

But if there’s something more you’d like to add, I doubt I’d be opposed.

charlesmchen:
> The top of the submitted log already contains a line like "iPhone Version: iPhone8,1, iOS Version: 10.3”

Oh interesting.  The downside to adding it when we submit the logs is that if some of the logs they're submitting are from before an iOS upgrade, the logs won't reflect that.  And that's useful information.

PR #1986, opened 2017-04-14
Improve thread safety in PushManager and NotificationsManager.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/environment/NotificationsManager.m
+++ b/Signal/src/environment/NotificationsManager.m
@@ -211,31 +211,33 @@
 
 - (void)presentNotification:(UILocalNotification *)notification identifier:(NSString *)identifier
 {
-    AssertIsOnMainThread();
+    dispatch_async(dispatch_get_main_queue(), ^{
+        // Replace any existing notification
+        // e.g. when an "Incoming Call" notification gets replaced with a "Missed Call" notification.
+        if (self.currentNotifications[identifier]) {
+            [self cancelNotificationWithIdentifier:identifier];
+        }
 
-    // Replace any existing notification
-    // e.g. when an "Incoming Call" notification gets replaced with a "Missed Call" notification.
-    if (self.currentNotifications[identifier]) {
-        [self cancelNotificationWithIdentifier:identifier];
-    }
+        [[UIApplication sharedApplication] scheduleLocalNotification:notification];
+        DDLogDebug(@"%@ presenting notification with identifier: %@", self.tag, identifier);
 
-    [[UIApplication sharedApplication] scheduleLocalNotification:notification];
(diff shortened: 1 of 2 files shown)


Signal/src/environment/NotificationsManager.m
```diff
@@ -211,31 +211,33 @@ - (NotificationType)notificationPreviewType
...
-
-    // Replace any existing notification
-    // e.g. when an "Incoming Call" notification gets replaced with a "Missed Call" notification.
-    if (self.currentNotifications[identifier]) {
-        [self cancelNotificationWithIdentifier:identifier];
-    }
+    dispatch_async(dispatch_get_main_queue(), ^{
+        // Replace any existing notification
+        // e.g. when an "Incoming Call" notification gets replaced with a "Missed Call" notification.
+        if (self.currentNotifications[identifier]) {
```
charlesmchen: Only modify `currentNotifications` on the main thread.

Signal/src/network/PushManager.m
```diff
@@ -431,32 +431,34 @@ - (BOOL)applicationIsActive {
...
+            // but the more lag we introduce to notification delivery.
+            const CGFloat kDelaySeconds = 0.3f;
+            notification.fireDate = [NSDate dateWithTimeIntervalSinceNow:kDelaySeconds];
+            notification.timeZone = [NSTimeZone localTimeZone];
+        }
 
-    [[UIApplication sharedApplication] scheduleLocalNotification:notification];
-    [self.currentNotifications addObject:notification];
+        [[UIApplication sharedApplication] scheduleLocalNotification:notification];
+        [self.currentNotifications addObject:notification];
```
charlesmchen: Only modify `currentNotifications` on the main thread.

charlesmchen:
I guess I'm also wondering if there's some redundancy between these two managers that could be de-duplicated.

charlesmchen:
> Is the point of this change that though we're already assert'ing MainThread, that we're not running this code in DEBUG often, so you don't have confidence in their assertion?

These two managers mutate their `currentNotifications` collections in these methods, which isn't safe to do concurrently.  `[PushManager cancelNotificationsWithThreadId:]` in particular iterates over this collection then mutates it which is not safe if another thread is mutating the collection at the same time.

michaelkirk:
But if we're already asserting that we're on the main thread whenever we read or edit, how could it have ever happened concurrently?

charlesmchen:
> But if we're already asserting that we're on the main thread whenever we read or edit, how could it have ever happened concurrently?

I just added those asserts last week, wondering if this is an issue.  Then I hit the asserts, hence this PR.  Gotta love asserts!

michaelkirk:
Well that clears it up. LGTM

PR #1994, opened 2017-04-18
Charlesmchen/call service edge cases

charlesmchen:
While debugging https://github.com/WhisperSystems/Signal-iOS/pull/1993 I often got caller or callee into a bad state.  This PR attempts to improve our handling of edge cases, errors and events from obsolete calls.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -74,8 +74,8 @@ enum CallError: Error {
     case assertionError(description: String)
     case disconnected
     case externalError(underlyingError: Error)
-    case timeout(description: String, call: SignalCall)
-    case obsoleteCall(description: String, call: SignalCall)
+    case timeout(description: String)
+    case obsoleteCall(description: String)
 }
 
 // Should be roughly synced with Android client for consistency
@@ -292,7 +292,7 @@ protocol CallServiceObserver: class {
             Logger.debug("\(self.TAG) got ice servers:\(iceServers)")
 
             guard self.call == call else {
-                return Promise(error: CallError.obsoleteCall(description:"obsolete call in \(#function)", call:call))
+                return Promise(error: CallError.obsoleteCall(description:"obsolete call in \(#function)"))
             }
 
             let useTurnOnly = Environment.getCurrent().preferences.doCallsHideIPAddress()
@@ -306,7 +306,7 @@ protocol CallServiceObserver: class {
             return self.peerConnectionClient!.createOffer()
         }.then { (sessionDescription: HardenedRTCSessionDescription) -> Promise<Void> in
             guard self.call == call else {
-                return Promise(error: CallError.obsoleteCall(description:"obsolete call in \(#function)", call:call))
+                return Promise(error: CallError.obsoleteCall(description:"obsolete call in \(#function)"))
             }
 
             return self.peerConnectionClient!.setLocalSessionDescription(sessionDescription).then {
@@ -316,7 +316,7 @@ protocol CallServiceObserver: class {
             }
         }.then {
             guard self.call == call else {
-                return Promise(error: CallError.obsoleteCall(description:"obsolete call in \(#function)", call:call))
+                return Promise(error: CallError.obsoleteCall(description:"obsolete call in \(#function)"))
             }
 
             let (callConnectedPromise, fulfill, _) = Promise<Void>.pending()
@@ -325,7 +325,7 @@ protocol CallServiceObserver: class {
             // Don't let the outgoing call ring forever. We don't support inbound ringing forever anyway.
             let timeout: Promise<Void> = after(interval: TimeInterval(connectingTimeoutSeconds)).then { () -> Void in
                 // rejecting a promise by throwing is safely a no-op if the promise has already been fulfilled
-                throw CallError.timeout(description: "timed out waiting to receive call answer", call: call)
+                throw CallError.timeout(description: "timed out waiting to receive call answer")
             }
 
             return race(timeout, callConnectedPromise)
@@ -481,7 +481,7 @@ protocol CallServiceObserver: class {
         call = newCall
 
         let backgroundTask = UIApplication.shared.beginBackgroundTask {
-            let timeout = CallError.timeout(description: "background task time ran out before call connected.", call: newCall)
+            let timeout = CallError.timeout(description: "background task time ran out before call connected.")
             DispatchQueue.main.async {
                 guard self.call == newCall else {
                     Logger.warn("\(self.TAG) ignoring obsolete call in \(#function)")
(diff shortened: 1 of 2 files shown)


Signal/src/call/CallService.swift
```diff
@@ -74,7 +74,8 @@ enum CallError: Error {
     case assertionError(description: String)
     case disconnected
     case externalError(underlyingError: Error)
-    case timeout(description: String, call: SignalCall)
```
charlesmchen: Was there are a reason we were parameterizing this error with a call?

Signal/src/call/CallService.swift
```diff
@@ -290,6 +291,10 @@ protocol CallServiceObserver: class {
         return getIceServers().then { iceServers -> Promise<HardenedRTCSessionDescription> in
             Logger.debug("\(self.TAG) got ice servers:\(iceServers)")
 
+            guard self.call == call else {
```
charlesmchen: The incoming call logic was already checking for obsolete calls, but the outgoing call logic wasn't.

Signal/src/call/CallService.swift
```diff
@@ -300,32 +305,49 @@ protocol CallServiceObserver: class {
...
+                throw CallError.obsoleteCall(description: "obsolete outgoing call connected")
+            }
+
             Logger.info("\(self.TAG) outgoing call connected.")
+
+            // TODO: There must be a better way to return a fulfilled promise.
+            let (promise, fulfill, _) = Promise<Void>.pending()
+            fulfill()
+            return promise
```
charlesmchen: What's the right way to return an already-fulfilled void promise?

Signal/src/call/CallService.swift
```diff
@@ -300,32 +305,49 @@ protocol CallServiceObserver: class {
...
         }.catch { error in
             Logger.error("\(self.TAG) placing call failed with error: \(error)")
 
             if let callError = error as? CallError {
-                self.handleFailedCall(error: callError)
+                self.handleFailedCall(failedCall: call, error: callError)
```
charlesmchen: `handleFailedCall()` is now more careful to avoid terminating the current call in response to errors for obsolete calls, so all of its callers must pass the call associated with the error.

Signal/src/call/CallService.swift
```diff
@@ -338,13 +360,12 @@ protocol CallServiceObserver: class {
         AssertIsOnMainThread()
 
         guard let call = self.call else {
-            handleFailedCall(error: .assertionError(description:"call was unexpectedly nil in \(#function)"))
+            Logger.warn("\(self.TAG) ignoring obsolete call in \(#function)")
```
charlesmchen: If `handleReceivedAnswer()` is received for an obsolete call, ignore it.

Signal/src/call/CallService.swift
```diff
@@ -443,7 +470,7 @@ protocol CallServiceObserver: class {
 
         guard call == nil else {
             // TODO on iOS10+ we can use CallKit to swap calls rather than just returning busy immediately.
```
charlesmchen: If `handleRemoteBusy()` is received for an obsolete call, ignore it.

Signal/src/call/CallService.swift
```diff
@@ -540,27 +562,27 @@ protocol CallServiceObserver: class {
         Logger.debug("\(TAG) called \(#function)")
 
         guard self.thread != nil else {
-            handleFailedCall(error: .assertionError(description: "ignoring remote ice update for thread: \(thread.uniqueId) since there is no current thread. Call already ended?"))
+            Logger.warn("ignoring remote ice update for thread: \(thread.uniqueId) since there is no current thread. Call already ended?")
```
charlesmchen: If we receive ice updates for obsolete calls, ignore them.

Signal/src/call/CallService.swift
michaelkirk: When call2 immediately follows call1, we don't want to accidentally cancel call2 when (the already ended) call1 times out.

Signal/src/call/CallService.swift
```diff
@@ -338,13 +353,12 @@ protocol CallServiceObserver: class {
...
 
         guard let call = self.call else {
-            handleFailedCall(error: .assertionError(description:"call was unexpectedly nil in \(#function)"))
+            Logger.warn("\(self.TAG) ignoring obsolete call in \(#function)")
             return
         }
 
         guard call.signalingId == callId else {
-            let description: String = "received answer for call: \(callId) but current call has id: \(call.signalingId)"
-            handleFailedCall(error: .assertionError(description: description))
```
michaelkirk: Good point.

Signal/src/call/CallService.swift
```diff
@@ -453,12 +472,13 @@ protocol CallServiceObserver: class {
         call = newCall
 
         let backgroundTask = UIApplication.shared.beginBackgroundTask {
-            let timeout = CallError.timeout(description: "background task time ran out before call connected.", call: newCall)
```
michaelkirk: As above, I think we want to continue to specify *which* call has timed out, lest we clobber a subsequent call.

Signal/src/call/CallService.swift
```diff
@@ -300,32 +305,42 @@ protocol CallServiceObserver: class {
...
+            }
+
             return self.peerConnectionClient!.setLocalSessionDescription(sessionDescription).then {
                 let offerMessage = OWSCallOfferMessage(callId: call.signalingId, sessionDescription: sessionDescription.sdp)
                 let callMessage = OWSOutgoingCallMessage(thread: thread, offerMessage: offerMessage)
                 return self.messageSender.sendCallMessage(callMessage)
             }
         }.then {
+            guard self.call == call else {
+                return Promise(error: CallError.obsoleteCall(description:"obsolete call in \(#function)"))
```
michaelkirk: nit: idiomatic to throw the error rather than to return a failed promise (here and above).

Signal/src/call/CallService.swift
```diff
@@ -655,7 +687,8 @@ protocol CallServiceObserver: class {
         }
 
         guard let call = self.call else {
-            handleFailedCall(error: .assertionError(description:"\(TAG) call was unexpectedly nil in \(#function)"))
+            // This should never happen; return to a known good state.
```
michaelkirk: nit: assertionFailure for all the "this should never happen"s

Signal/src/call/CallService.swift
michaelkirk: It seems like since you're now similarly pegging most all errors to their `call`, that the new logic ensures the previous timeout behavior. 👍

Signal/src/call/CallService.swift

Signal/src/call/CallService.swift

PR #1996, opened 2017-04-18
Improve group members view.

charlesmchen:
* Use `ContactTableViewCell`.
   * Show avatars
   * Style contact names as they appear elsewhere.
   * Show block status.
* Tapping on a member now offers the following options:
   * Contact info (previously the only option).
   * Block/unblock this user.
   * Send 1:1 message to this user.
   * Call this user.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessagesViewController.h
+++ b/Signal/src/ViewControllers/MessagesViewController.h
@@ -41,7 +41,9 @@ extern NSString *const OWSMessagesViewControllerDidAppearNotification;
 @property (nonatomic, strong) AVAudioPlayer *audioPlayer;
 @property (nonatomic, strong) AVAudioRecorder *audioRecorder;
 
-- (void)configureForThread:(TSThread *)thread keyboardOnViewAppearing:(BOOL)keyboardAppearing;
+- (void)configureForThread:(TSThread *)thread
+    keyboardOnViewAppearing:(BOOL)keyboardAppearing
+        callOnViewAppearing:(BOOL)callOnViewAppearing;
 - (void)popKeyBoard;
 
 #pragma mark 3D Touch Methods
--- a/Signal/src/ViewControllers/ShowGroupMembersViewController.m
+++ b/Signal/src/ViewControllers/ShowGroupMembersViewController.m
@@ -202,7 +202,7 @@ static NSString *const kUnwindToMessagesViewSegue = @"UnwindToMessagesViewSegue"
     [actionSheetController
         addAction:[UIAlertAction actionWithTitle:NSLocalizedString(@"GROUP_MEMBERS_VIEW_CONTACT_INFO",
                                                      @"Button label for the 'show contact info' button")
-                                           style:UIAlertActionStyleDestructive
+                                           style:UIAlertActionStyleDefault
                                          handler:^(UIAlertAction *_Nonnull action) {
                                              [self showContactInfoViewForMember:indexPath];
                                          }]];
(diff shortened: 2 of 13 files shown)


Signal/src/ViewControllers/MessagesViewController.h
```diff
@@ -41,7 +41,9 @@ extern NSString *const OWSMessagesViewControllerDidAppearNotification;
 @property (nonatomic, strong) AVAudioPlayer *audioPlayer;
 @property (nonatomic, strong) AVAudioRecorder *audioRecorder;
 
-- (void)configureForThread:(TSThread *)thread keyboardOnViewAppearing:(BOOL)keyboardAppearing;
+- (void)configureForThread:(TSThread *)thread
+    keyboardOnViewAppearing:(BOOL)keyboardAppearing
+        callOnViewAppearing:(BOOL)callOnViewAppearing;
```
charlesmchen: This isn't great, but I'm not going to change it here.

Signal/src/ViewControllers/ShowGroupMembersViewController.m
```diff
@@ -62,11 +100,15 @@ - (void)viewDidLoad {
 
     self.groupContacts =
         [[GroupContactsResult alloc] initWithMembersId:self.thread.groupModel.groupMemberIds without:nil];
-
+    __weak ShowGroupMembersViewController *weakSelf = self;
```
charlesmchen: I think I've fixed a serious memory leak that would prevent this view from every being deallocated.

Signal/src/Storyboard/Main.storyboard
```diff
@@ -1,11 +1,11 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="11762" systemVersion="15G1217" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" colorMatched="YES" initialViewController="tuk-0x-yCb">
```
michaelkirk: Sorry that I'm crazy and read the storyboard diffs, but it seems like there aren't any meaningful changes here? Did you intend to check this in?

Signal/src/ViewControllers/ShowGroupMembersViewController.m
```diff
@@ -90,44 +134,176 @@ - (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger
...
+    // Adjust index path for the header row.
+    indexPath = [NSIndexPath indexPathForRow:indexPath.row - 1 inSection:indexPath.section];
+
+    UIAlertController *actionSheetController =
+        [UIAlertController alertControllerWithTitle:nil message:nil preferredStyle:UIAlertControllerStyleActionSheet];
+
+    [actionSheetController
+        addAction:[UIAlertAction actionWithTitle:NSLocalizedString(@"GROUP_MEMBERS_VIEW_CONTACT_INFO",
+                                                     @"Button label for the 'show contact info' button")
+                                           style:UIAlertActionStyleDestructive
```
michaelkirk: None of these actions should be styled as "destructive" except "Block"

Signal/src/ViewControllers/ShowGroupMembersViewController.m
```diff
@@ -154,15 +329,58 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
...
 - (Contact *)contactForIndexPath:(NSIndexPath *)indexPath {
     Contact *contact = [self.groupContacts contactForIndexPath:indexPath];
     return contact;
 }
 
+- (BOOL)isContactBlocked:(Contact *)contact
+{
+    if (contact.parsedPhoneNumbers.count < 1) {
+        // Hide contacts without any valid phone numbers.
+        return NO;
```
michaelkirk: I'm not sure what this comment means in this context. It doesn't seem like we're using this method to "hide" anything.

Signal/src/Storyboard/Main.storyboard
charlesmchen: Done.

Signal/src/ViewControllers/ShowGroupMembersViewController.m
charlesmchen: Done.

Signal/src/ViewControllers/ShowGroupMembersViewController.m
charlesmchen: Done.

PR #1997, opened 2017-04-18
Fix paste of text.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessagesViewController.m
+++ b/Signal/src/ViewControllers/MessagesViewController.m
@@ -111,10 +111,24 @@ typedef enum : NSUInteger {
     return YES;
 }
 
+- (BOOL)pasteBoardHasText
+{
+    if ([UIPasteboard generalPasteboard].numberOfItems < 1) {
+        return NO;
+    }
+    NSIndexSet *itemSet = [NSIndexSet indexSetWithIndex:0];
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -111,10 +111,24 @@ - (BOOL)canBecomeFirstResponder {
...
+        ||
+        [utiTypes containsObject:(NSString *)kUTTypeUTF8PlainText] ||
+        [utiTypes containsObject:(NSString *)kUTTypeUTF16PlainText]);
+}
+
 - (BOOL)pasteBoardHasPossibleAttachment {
     // We don't want to load/convert images more than once so we
     // only do a cursory validation pass at this time.
-    return [SignalAttachment pasteboardHasPossibleAttachment];
+    return ([SignalAttachment pasteboardHasPossibleAttachment] && ![self pasteBoardHasText]);
```
michaelkirk: nit: maybe simpler: https://developer.apple.com/reference/uikit/uipasteboard/1829416-hasstrings?language=objc

charlesmchen: I wanted to be precise about considering only the first item if the pasteboard contains multiple items, since that's all we'll use.

In the long run, it'd be nice to support pasting multiple items at once.

PR #2002, opened 2017-04-19
Fix segues in conversation settings view.

charlesmchen:
It turns out that if you delete the only view in a storyboard that uses a segue, the segue is deleted as well.  Switch "conversation settings" view to initiate all navigation programmatically without segues.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
+++ b/Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
@@ -30,11 +30,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-static NSString *const OWSConversationSettingsTableViewControllerSegueUpdateGroup =
-    @"OWSConversationSettingsTableViewControllerSegueUpdateGroup";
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
```diff
@@ -168,31 +163,50 @@ - (void)updateTableContents
     firstSection.customHeaderHeight = @(100.f);
 
     if (!self.isGroupThread && self.thread.hasSafetyNumbers) {
-        [firstSection addItem:[OWSTableItem itemWithCustomCellBlock:^{
-            UITableViewCell *cell = [UITableViewCell new];
-            cell.accessoryType = UITableViewCellAccessoryDisclosureIndicator;
```
charlesmchen: I'm not sure why clang-format reformatted so much of this file.

michaelkirk: I think doing so allowed the "actionBlock" to be brought left a bit.

clang-format is a whimsical beast.

PR #2009, opened 2017-04-20
Callscreen fills buttons to show active state

michaelkirk:
Rather than reflecting the current state, follow the example of the iOS native call screen by using a "filled in" style to indicate when a button is active.

This only affects the audio call screen controls, as the video screen presents some different challenges.

screen shot 2017-04-20 at 12 57 23 pm

PTAL @charlesmchen


TreatyOfRome:
Fixes #1161?

PR #2010, opened 2017-04-20
Charlesmchen/audio player3

charlesmchen:
* Remove waveform from audio message bubbles.
* Remove SCWaveformView

PTAL @michaelkirk 

screen shot 2017-04-20 at 1 48 05 pm


Podfile.lock
```diff
@@ -159,14 +157,13 @@ SPEC CHECKSUMS:
...
   SocketRocket: dbb1554b8fc288ef8ef370d6285aeca7361be31e
   SQLCipher: 43d12c0eb9c57fb438749618fc3ce0065509a559
   TwistedOakCollapsingFutures: f359b90f203e9ab13dfb92c9ff41842a7fe1cd0c
   UnionFind: c33be5adb12983981d6e827ea94fc7f9e370f52d
   YapDatabase: b1e43555a34a5298e23a045be96817a5ef0da58f
 
-PODFILE CHECKSUM: d4f237ca4bd0b71e488d29b1e38dc31d06e45a52
+PODFILE CHECKSUM: cb2cbbe74dab34123e1cb527417ef658aa60bd26
 
-COCOAPODS: 1.2.1
```
michaelkirk: can you update your cocoapods?

Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -141,54 +122,37 @@ - (UIView *)mediaView {
...
-        _durationLabel.text   = [self formatDuration:player.duration];
-        _durationLabel.font   = [UIFont systemFontOfSize:14];
-        [_durationLabel sizeToFit];
-        _durationLabel.frame = CGRectMake((size.width - _durationLabel.frame.size.width) - 10,
-                                          _durationLabel.frame.origin.y,
-                                          _durationLabel.frame.size.width,
-                                          AUDIO_BAR_HEIGHT);
-        _durationLabel.backgroundColor = [UIColor clearColor];
-        _durationLabel.textColor       = [UIColor whiteColor];
+        NSString *audioLabelText = [[MIMETypeUtil fileExtensionForMIMEType:self.contentType] uppercaseString];
```
michaelkirk: nit: on the attachment confirmation page we print "Mp3" not "MP3" (probably the confirmation page is what's wrong, and this should stay the same)

charlesmchen:
PTAL @michaelkirk 

* Show audio filename and file size by default.
* Show audio progress/duration while audio attachments are playing.
* Tweak icons, colors, font sizes, sizing and layout of audio and generic attachment messages.
* Improve formatting of file sizes.

screen shot 2017-04-28 at 6 01 11 pm

charlesmchen:
PTAL @michaelkirk in case you missed it.

Signal/src/Models/TSMessageAdapaters/TSGenericAttachmentAdapter.m
```diff
@@ -100,7 +95,7 @@ - (UIView *)mediaView
 {
     if (_cachedMediaView == nil) {
         CGSize viewSize = [self mediaViewDisplaySize];
-        UIColor *textColor = (self.incoming ? [UIColor blackColor] : [UIColor whiteColor]);
+        UIColor *textColor = (self.incoming ? [UIColor colorWithWhite:0.2 alpha:1.f] : [UIColor whiteColor]);
```
charlesmchen: It's a great question. There's a lot of "ink" within the incoming attachment bubbles and I felt like it was "too much" and "not hierarchical enough" when all of the ink was black.  I thought we should ever-so-slightly de-emphasize the filename and file size labels in order to say "the most important thing here is the play/pause or attachment icon which indicates that this is an attachment" and "also, here are the details."

Notice that I treated incoming and outgoing attachments differently.  IMO White-on-blue doesn't "overwhelm with drabness" in the say way that black-on-gray does, so I only de-emphasized the file size subtitle for outgoing, not the filename.

Obviously this is all subjective.  WDYT?

### Before 

attachment colors before

### After

attachment colors after

charlesmchen: Shoot I deleted your comment while trying to delete mine.  Here's my response.

It's a great question. There's a lot of "ink" within the incoming attachment bubbles and I felt like it was "too much" and "not hierarchical enough" when all of the ink was black.  I thought we should ever-so-slightly de-emphasize the filename and even further de-emphasize the file size labels in order to say "the most important thing here is the play/pause or attachment icon which indicates that this is an attachment" and "also, here is the filename", "lastly, the file size is the least important detail".  I think this is sort of what designers call a hierarchy.

Notice that I treated incoming and outgoing attachments differently.  IMO White-on-blue doesn't "overwhelm with drabness" in the say way that black-on-gray does, so I used full-on white for outgoing filenames and slightly de-emphasized the subtitle.

Obviously this is all subjective.  WDYT?

### Before 

attachment colors before

### After

attachment colors after

Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -81,59 +84,97 @@ - (BOOL)isVideo {
...
+- (CGFloat)bubbleHeight
+{
+    return 35.f;
 }
 
-- (void)removeDurationLabel {
-    [_durationLabel removeFromSuperview];
+- (CGFloat)iconSize
+{
+    return 40.f;
```
charlesmchen: It's really not a big deal, but I find it makes the code more readable to emphasize the floats literals.  For example, you avoid common bugs like mistaking `round(value / 40)` vs. `round(value / 40.f)`.  If it gives you the howling fantods, I'll be happy to abandon this habit.

Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -81,59 +84,97 @@ - (BOOL)isVideo {
...
 
-#pragma mark - JSQMessageMediaData protocol
+- (CGFloat)vMargin
+{
+    return 10.f;
+}
+
+- (UIColor *)audioTextColor
+{
+    return (self.incoming ? [UIColor colorWithWhite:0.2 alpha:1.f] : [UIColor whiteColor]);
```
charlesmchen: See my other comment but I don't intend to use this color elsewhere. The attachment bubbles just seemed way too visually overwhelming with "all of that black ink" and three separate elements crammed into a single bubble.  Ideally we can revisit the colors, type, bubble shapes and size, etc. when we replace JSQ.

Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -144,79 +185,96 @@ - (UIView *)mediaView {
...
+                round(contentFrame.origin.y + (contentFrame.size.height - self.iconSize) * 0.5f),
+                self.iconSize,
+                self.iconSize);
+            _audioPlayPauseButton = [[UIButton alloc] initWithFrame:iconFrame];
+            _audioPlayPauseButton.enabled = NO;
+            [_cachedMediaView addSubview:_audioPlayPauseButton];
+
+            const CGFloat kLabelHSpacing = 3;
+            const CGFloat kLabelVSpacing = 2;
+            NSString *topText =
```
charlesmchen: Yes, I'd love to. So far I have resisted the urge to pull out this logic.

* We're likely to revisit all of this "soon" when we replace JSQ.
* It's not like we'll keep adding more attachment adapters.
* There are slight differences between the attachment adapter's appearance which would complicate this.
* The attachment approval dialog and the "full screen" attachment view are way larger; they won't re-use any of this view/layout logic.

charlesmchen:
> one blocking issue: I can't see the media player on the attachment preview screen, e.g. after picking an mp3 attachment from "Documents"

Oh I was going to do that in a subsequent PR but why not?

img_0275

img_0274

charlesmchen:
PTAL @michaelkirk

Signal/src/Models/TSMessageAdapaters/TSGenericAttachmentAdapter.m
michaelkirk: I see what you're saying about focusing on the play button.  👍

michaelkirk:
LGTM, thanks!

PR #2011, opened 2017-04-20
Fix timer retain cycle.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/util/NSTimer+OWS.m
+++ b/Signal/src/util/NSTimer+OWS.m
@@ -1,0 +1,65 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "NSTimer+OWS.h"
+#import <objc/runtime.h>
+
+@interface NSTimerProxy : NSObject
+
+@property (nonatomic, weak) id target;
+@property (nonatomic) SEL selector;
+
+@end
+
+#pragma mark -
+
+@implementation NSTimerProxy
+
+- (void)timerFired:(NSDictionary *)userInfo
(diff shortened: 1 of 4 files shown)


Signal/src/util/NSTimer+OWS.m
```diff
@@ -0,0 +1,65 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "NSTimer+OWS.h"
+#import <objc/runtime.h>
+
+@interface NSTimerProxy : NSObject
```
michaelkirk: nit: nullability annotations please

Signal/src/util/NSTimer+OWS.m
```diff
@@ -0,0 +1,65 @@
...
+
+@end
+
+#pragma mark -
+
+static void *kNSTimer_OWS_Proxy = &kNSTimer_OWS_Proxy;
+
+@implementation NSTimer (OWS)
+
+- (NSTimerProxy *)proxy
```
michaelkirk: nit: to avoid shooting yourself in the foot, especially with short method names, it's good to prefix category methods, e.g. `ows_proxy`

Signal/src/util/NSTimer+OWS.m
charlesmchen: Done, sorry.

Signal/src/util/NSTimer+OWS.m
charlesmchen: Done.

PR #2012, opened 2017-04-20
Increase local notification delay.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
@michaelkirk oh you mean only delaying notifications like this if we've seen evidence of desktop activity recently?

PR #2018, opened 2017-04-21
Only the last cell associated with a media adapter should clear its v…

charlesmchen:
…iews.

PTAL @michaelkirk


Signal/src/Models/TSMessageAdapaters/TSAnimatedAdapter.m
```diff
@@ -21,6 +21,7 @@ @interface TSAnimatedAdapter ()
 @property (nonatomic) TSAttachmentStream *attachment;
 @property (nonatomic, nullable) AttachmentUploadView *attachmentUploadView;
 @property (nonatomic) BOOL incoming;
+@property (nonatomic, nullable, weak) id lastCell;
```
michaelkirk: nit: maybe "owningCell" or "presentingCell" or something? "last" is sort of ambiguous outside of the context of this PR.

Issue #1749, opened 2017-02-11
Incorrect layout for images with unusual aspect ratios.


charlesmchen:
In conversation view, tall and wide images aren't always laid out properly. This issue seems to affect scroll state mostly.

edit by @michaelkirk

This was related to messages with extreme AR, e.g. really short/wide messages and really tall/skinny.

michaelkirk:
example image attachments or a screenshot would be good to clarify what you mean.

TreatyOfRome:
Screenshot from a photo taken with and received from a Samsung Galaxy S7 EDGE. 
![fullsizerender](https://cloud.githubusercontent.com/assets/26620973/24492319/80d8adae-151a-11e7-84f8-d3a0e231c9a2.jpg)


When selecting another photo from this conversation, closing it and scrolling back to the conversation view with this photo, the aspect ratio is correct again. 
![fullsizerender2](https://cloud.githubusercontent.com/assets/26620973/24492236/2e6806fa-151a-11e7-8073-0bbcc631244c.jpg)

When closing Signal (double tap home button, swipe up) the aspect ratio is back to the wrong aspect ratio as seen in the first screenshot. Opening and closing this particular photo does not correct the aspect ratio. 

I cannot reproduce this with all photo's received from this Samsung. Some have unusual aspect ratio's, some don't. 

Device: iPhone 6S
iOS version: 10.3
Signal version: 2.9.0

big-r81:
Hi,
i can confirm this behavior.

- after uploading the image upload indicator is hiding and sometimes the pictures get squeezed or squashed
- sometimes old pictures getting distored too - after sending a new picture!
- viewing a picture (fullscreen) and closing full screen shows sometimes wrong images aspect ratios
- scrolling the thread will show sometimes wrong image aspect ratios

----------

- leaving a conversation and entering the same conversation can help to show images with correct aspect ratios (not alle images are shown correct)
- long-pressing on an image with wrong aspect ratio shows correct, after showing the menu (copy/delete/save/share)

Until now, i can't give a 100% instruction for reproducing the error(s).

Device IPhone 6
iOS version 10.3
Signal version 2.9.1

charlesmchen:
Fixed in the forthcoming release.

stark226:
[...]

PR #2020, opened 2017-04-21
Charlesmchen/stretched images

charlesmchen:
Fixes the long-standing and highly irritating "portrait images often appear stretched in messages view" bug.

Apparently there's a bug in UIKit and views which a) use contentMode b) reside inside a UIScrollView must have their frame set _after_ their contentMode is set.

The repro was: send or receive a portrait image.  Go to conversation settings, return to messages view _or_  open attachments action sheet -> camera roll view, hit cancel to return to messages view.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/Models/TSMessageAdapaters/TSAnimatedAdapter.m
+++ b/Signal/src/Models/TSMessageAdapaters/TSAnimatedAdapter.m
@@ -88,8 +88,8 @@ NS_ASSUME_NONNULL_BEGIN
         FLAnimatedImageView *imageView = [[FLAnimatedImageView alloc] init];
         imageView.animatedImage        = animatedGif;
         CGSize size                    = [self mediaViewDisplaySize];
-        imageView.frame                = CGRectMake(0.0, 0.0, size.width, size.height);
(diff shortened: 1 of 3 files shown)


Signal/src/Models/TSMessageAdapaters/TSAnimatedAdapter.m
```diff
@@ -88,8 +88,8 @@ - (UIView *)mediaView {
         FLAnimatedImageView *imageView = [[FLAnimatedImageView alloc] init];
         imageView.animatedImage        = animatedGif;
         CGSize size                    = [self mediaViewDisplaySize];
-        imageView.frame                = CGRectMake(0.0, 0.0, size.width, size.height);
         imageView.contentMode          = UIViewContentModeScaleAspectFill;
```
michaelkirk: can you add a comment for the ages since this is so surprisingly sensitive.

TreatyOfRome:
Fixes #1749?

charlesmchen:
Yes, thank you @TreatyOfRome .  I've closed #1749

PR #2022, opened 2017-04-21
Resolve the menu item conflicts between message view and media view.

charlesmchen:
`UIMenuController` has a horrible API.  Which menu items _might_ appear is a function of the last value written to its `menuItems` property, but which menu items you want to appear is contextual, so it's last-writer-wins without any good hook to update the menu items based on the context of the current user action.  

I've made MessagesVC a bit more defensive and made the media view a bit more polite.

PTAL @michaelkirk

### DIFF
--- a/Signal/translations/en.lproj/Localizable.strings
+++ b/Signal/translations/en.lproj/Localizable.strings
@@ -325,6 +325,9 @@
 /* Short name for edit menu item to share contents of media message. */
 "EDIT_ITEM_SHARE_ACTION" = "Share";
 
+/* Short name for edit menu item to share contents of media message. */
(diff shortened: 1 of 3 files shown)


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -325,6 +325,9 @@
 /* Short name for edit menu item to share contents of media message. */
 "EDIT_ITEM_SHARE_ACTION" = "Share";
 
+/* Short name for edit menu item to share contents of media message. */
```
michaelkirk: might want to re-run auto-genstrings since this comment is off.

PR #2024, opened 2017-04-21
Avoid crash when scrolling during a database update.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessagesViewController.m
+++ b/Signal/src/ViewControllers/MessagesViewController.m
@@ -240,6 +240,7 @@ typedef enum : NSUInteger {
 
 @property (nonatomic) NSCache *messageAdapterCache;
 @property (nonatomic) BOOL userHasScrolled;
+@property (nonatomic) BOOL userIsScrolling;
 
 @end
 
@@ -791,6 +792,7 @@ typedef enum : NSUInteger {
     [super viewDidDisappear:animated];
     self.inputToolbar.contentView.textView.editable = NO;
     self.userHasScrolled = NO;
+    self.userIsScrolling = NO;
 }
 
 #pragma mark - Initiliazers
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2598,53 +2602,55 @@ - (void)yapDatabaseModified:(NSNotification *)notification {
...
-      for (YapDatabaseViewRowChange *rowChange in messageRowChanges) {
-          switch (rowChange.type) {
-              case YapDatabaseViewChangeDelete: {
-                  [self.collectionView deleteItemsAtIndexPaths:@[ rowChange.indexPath ]];
-
-                  YapCollectionKey *collectionKey = rowChange.collectionKey;
-                  if (collectionKey.key) {
-                      [self.messageAdapterCache removeObjectForKey:collectionKey.key];
-                  }
-
```
charlesmchen: mostly clang-format; there aren't many changes here.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2598,53 +2602,55 @@ - (void)yapDatabaseModified:(NSNotification *)notification {
...
-              [self scrollToBottomAnimated:shouldAnimateScrollToBottom];
-          }
+            OWSAssert([NSThread isMainThread]);
+
+            if (!success) {
+                [self.collectionView.collectionViewLayout
+                    invalidateLayoutWithContext:[JSQMessagesCollectionViewFlowLayoutInvalidationContext context]];
+                [self.collectionView reloadData];
+            }
+            if (scrollToBottom && self.userIsScrolling) {
```
michaelkirk: I'm not sure how you got here, but want to double check, because I'd assume you'd want this condition inverted.

 e.g. you don't want to scroll *while* the user is *already* scrolling.

PR #2025, opened 2017-04-21
Charlesmchen/attachment action sheet icons

charlesmchen:
PTAL @michaelkirk 

img_0007


charlesmchen:
> take it or leave it, but what do you think of left aligning the action text?

Adding these icons required using undocumented features of `UIActionSheet`, which typically places you in danger of app store review rejection.  But a ton of apps do this, so I went ahead and did it.  Left-aligning the copy (like WhatsApp) will require another hack (crawling the view hierarchy of the alert looking for UILabels).  So I'd rather not unless you feel like it's worth it?

PR #2027, opened 2017-04-22
Skip attachment approval dialog for image and video shot by camera.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
PTAL @michaelkirk

PR #2029, opened 2017-04-24
Charlesmchen/navigation bar appearance

charlesmchen:
* Fix the “navigation titles are black after sharing” issue. 
* Fix the “message view subtitle mis-sized after muting” issue. 

PTAL @michaelkirk


Signal/src/util/UIUtil.m
```diff
@@ -46,14 +51,10 @@ + (void)applySignalAppearence
...
 
-    NSShadow *shadow = [NSShadow new];
-    [shadow setShadowColor:[UIColor clearColor]];
-
-    NSDictionary *navbarTitleTextAttributes = @{
-        NSForegroundColorAttributeName : [UIColor whiteColor],
-        NSShadowAttributeName : shadow,
-    };
-    [[UINavigationBar appearance] setTitleTextAttributes:navbarTitleTextAttributes];
+    // If we set NSShadowAttributeName, the NSForegroundColorAttributeName value is ignored.
```
michaelkirk: And presumably we are OK with just not setting the ShadowAttribute?

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -1013,6 +1013,7 @@ - (void)updateNavigationBarSubtitleLabel
                                            NSForegroundColorAttributeName : [UIColor colorWithWhite:0.9f alpha:1.f],
                                        }]];
     self.navigationBarSubtitleLabel.attributedText = subtitleText;
+    [self.navigationBarSubtitleLabel sizeToFit];
```
michaelkirk: :+1:

Signal/src/util/UIUtil.m
charlesmchen: `setTitleTextAttributes` seems to have no effect if this attribute is included, so I don't see an alternative.

PR #2031, opened 2017-04-24
always send a filename with attachments

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/SignalAttachment.swift
+++ b/Signal/src/ViewControllers/SignalAttachment.swift
@@ -174,11 +174,10 @@ class SignalAttachment: NSObject {
         if dataUTI == SignalAttachment.kUnknownTestAttachmentUTI {
             return OWSMimeTypeUnknownForTests
         }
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -174,11 +174,31 @@ class SignalAttachment: NSObject {
         if dataUTI == SignalAttachment.kUnknownTestAttachmentUTI {
             return OWSMimeTypeUnknownForTests
         }
-        let mimeType = UTTypeCopyPreferredTagWithClass(dataUTI as CFString, kUTTagClassMIMEType)
-        guard mimeType != nil else {
+        guard let mimeType = UTTypeCopyPreferredTagWithClass(dataUTI as CFString, kUTTagClassMIMEType) else {
             return nil
         }
-        return mimeType?.takeRetainedValue() as? String
+        return mimeType.takeRetainedValue() as String
```
michaelkirk: Everything above here is just a refactor to fix a compiler warning.

PR #2032, opened 2017-04-24
Avoid crash in CallService

michaelkirk:
Not for 2.11, but 2.12

Rather than force unwrapping (`!`), use guards/implicit unwrapping to
handle the failure gracefully.

This crash might have already been solved by your changes here @charlesmchen: 4451a3a4, but this adds another level of robustness.

PTAL @charlesmchen


michaelkirk:
in 27e55d29061de63c9b82194b8ca891a2a0c02a36

PR #2037, opened 2017-04-25
Allow paste of URL text

michaelkirk:
The problem surfaces not when pasting a link copied from e.g. the safari
URL bar, but if, when viewing a webpage in safari, when you open the
"share sheet" and choose "copy" - the pasted link was pasted "as an
attachment" from the url clipboard.

PTAL @charlesmchen


michaelkirk:
Closing, as this was incorporated into #2040

PR #2040, opened 2017-04-25
Charlesmchen/bundle document types

charlesmchen:
PTAL @michaelkirk


Signal/src/ViewControllers/OWSTableViewController.m
```diff
@@ -196,6 +197,20 @@ - (void)loadView
...
+    [super viewDidLoad];
+
+    [self.navigationController.navigationBar setTranslucent:NO];
+}
+
+- (void)viewWillAppear:(BOOL)animated
+{
+    [super viewWillAppear:animated];
+
+    self.tableView.tableFooterView = [[UIView alloc] initWithFrame:CGRectZero];
```
michaelkirk: just curious, why do we always want a footer view? I realize it's sized zero so it's not going to appear, but it seems like you did this for a reason. Why?

Signal/src/ViewControllers/SelectThreadViewController.h
```diff
@@ -0,0 +1,21 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+@class TSThread;
+
+@protocol SelectThreadViewControllerDelegate <NSObject>
```
michaelkirk: nullability annotations please

Signal/src/ViewControllers/SelectThreadViewController.m
```diff
@@ -0,0 +1,428 @@
...
+    __weak SelectThreadViewController *weakSelf = self;
+    OWSTableContents *contents = [OWSTableContents new];
+    OWSTableSection *section = [OWSTableSection new];
+
+    // Threads
+    for (TSThread *thread in [self filteredThreadsWithSearchText]) {
+        [section addItem:[OWSTableItem itemWithCustomCellBlock:^{
+            SelectThreadViewController *strongSelf = weakSelf;
+            if (!strongSelf) {
+                return (InboxTableViewCell *)nil;
```
michaelkirk: Why are we casting nil? 

If it's to clobber the "nonnull" constraint declared in:

```
 typedef UITableViewCell *_Nonnull (^OWSTableCustomCellBlock)();
```

Then I think we should just be honest with ourselves and change the UITableViewCell returned by OWSTableCustomCellBlock to be Nullable, since apparently it can be nil.

Signal/src/ViewControllers/SelectThreadViewController.m
```diff
@@ -0,0 +1,428 @@
...
+}
+
+- (BOOL)isContactHidden:(Contact *)contact
+{
+    if (contact.parsedPhoneNumbers.count < 1) {
+        // Hide contacts without any valid phone numbers.
+        return YES;
+    }
+
+    if ([self isCurrentUserContact:contact]) {
```
michaelkirk: Why not? We can compose a message to ourselves. This is used, e.g. to sync a message to desktop.

Signal/src/ViewControllers/SelectThreadViewController.m
```diff
@@ -0,0 +1,428 @@
...
+- (BOOL)isRecipientIdBlocked:(NSString *)recipientId
+{
+    OWSAssert(recipientId.length > 0);
+    
+    return [_blockedPhoneNumberSet containsObject:recipientId];
+}
+
+- (BOOL)isContactHidden:(Contact *)contact
+{
+    if (contact.parsedPhoneNumbers.count < 1) {
```
michaelkirk: nit: this is a little confusing, since `self.contacts` is populated with `contactsManager.signalContacts`, which necessarily have a signal id. Having checks like this muddy the waters as to whose responsibility it is to make sure signalContacts are valid signalContacts.

Signal/src/ViewControllers/SendExternalFileViewController.m
```diff
@@ -0,0 +1,66 @@
...
+//
+
+#import "SendExternalFileViewController.h"
+#import "Environment.h"
+#import "Signal-Swift.h"
+#import "ThreadUtil.h"
+#import <SignalServiceKit/OWSMessageSender.h>
+#import <SignalServiceKit/TSThread.h>
+
+@interface SendExternalFileViewController () <SelectThreadViewControllerDelegate>
```
michaelkirk: nullability annotations here and in the .h please

Signal/src/ViewControllers/SendExternalFileViewController.m
```diff
@@ -0,0 +1,66 @@
...
+- (void)threadWasSelected:(TSThread *)thread
+{
+    OWSAssert(self.attachment);
+    OWSAssert(thread);
+
+    // We should have a valid filename.
+    OWSAssert(self.attachment.filename.length > 0);
+    NSString *fileExtension = [self.attachment.filename pathExtension].lowercaseString;
+    OWSAssert(fileExtension.length > 0);
+    NSSet<NSString *> *textExtensions = [NSSet setWithArray:@[
```
michaelkirk: Take it or leave it, but this UX seems a little off to me.

If Alice is "importing" a .txt file, and sends it to Bob, I think Alice expects Bob to see it presented as a txt "file", not as a message bubble, as if she'd "typed" it in.

I think if Alice wanted it presented as a message bubble that she had *typed* in, she would open the txt file in whatever program and copy/paste the contents of it into the text area.

Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -266,10 +275,26 @@ class SignalAttachment: NSObject {
...
 
+    public var isText: Bool {
+        return SignalAttachment.textUTISet.contains(dataUTI)
+    }
+
     public class func pasteboardHasPossibleAttachment() -> Bool {
         return UIPasteboard.general.numberOfItems > 0
     }
 
+    public class func pasteBoardHasText() -> Bool {
```
michaelkirk: nit: *seems* like we've settled on Pasteboard as one word, we should be consistent.

Signal/src/ViewControllers/ThreadViewHelper.h
```diff
@@ -0,0 +1,22 @@
...
+- (void)threadListDidChange;
+
+@end
+
+#pragma mark -
+
+@class TSThread;
+
+// A helper class
+@interface ThreadViewHelper : NSObject
```
michaelkirk: Nice effort on the comment 😉

Signal/src/ViewControllers/ThreadViewHelper.m
```diff
@@ -0,0 +1,136 @@
...
+            for (NSUInteger item = 0; item < numberOfItems; item++) {
+                TSThread *thread = [[transaction extension:TSThreadDatabaseViewExtensionName]
+                    objectAtIndexPath:[NSIndexPath indexPathForItem:(NSInteger)item inSection:(NSInteger)section]
+                         withMappings:self.threadMappings];
+                [threads addObject:thread];
+            }
+        }
+    }];
+
+    _threads = threads;
```
michaelkirk: nit: [threads copy]

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -113,24 +113,10 @@ - (BOOL)canBecomeFirstResponder {
...
-        ||
-        [utiTypes containsObject:(NSString *)kUTTypeUTF8PlainText] ||
-        [utiTypes containsObject:(NSString *)kUTTypeUTF16PlainText]);
-}
-
 - (BOOL)pasteBoardHasPossibleAttachment {
     // We don't want to load/convert images more than once so we
     // only do a cursory validation pass at this time.
-    return ([SignalAttachment pasteboardHasPossibleAttachment] && ![self pasteBoardHasText]);
+    return ([SignalAttachment pasteboardHasPossibleAttachment] && ![SignalAttachment pasteBoardHasText]);
```
michaelkirk: 👏 👏 👏  This means I can close https://github.com/WhisperSystems/Signal-iOS/pull/2037 right?

Signal/src/ViewControllers/SignalsViewController.m
```diff
@@ -549,6 +548,67 @@ - (void)presentThread:(TSThread *)thread
...
+    OWSAssert([NSThread isMainThread]);
+    OWSAssert(viewController);
+
+    [self presentViewControllerWithBlock:^{
+        [self presentViewController:viewController animated:animatePresentation completion:nil];
+    }
+                        animateDismissal:animateDismissal];
+}
+
+- (void)pushTopLevelViewController:(UIViewController *)viewController
```
michaelkirk: Seems like this is unused?

Signal/src/ViewControllers/OWSTableViewController.m
charlesmchen: I wondered the same thing a couple of months ago when I saw this sprinkled everywhere in our codebase.  I googled a bit and learned on StackOverflow that this is a hack to work around some bug in UITableView.  I can't remember what the bug is, but I've been propagating this fix as I make new UITableViews.

Signal/src/ViewControllers/SelectThreadViewController.h
charlesmchen: Done.

Signal/src/ViewControllers/SelectThreadViewController.m
charlesmchen: If you don't cast to nil, the compiler doesn't correctly deduce the return type of the block and you get a "block has inconsistent return types" build error.

Signal/src/ViewControllers/SelectThreadViewController.m
charlesmchen: True.  I have a huge open branch that pulls out a `ContactsViewHelper` and introduces the `ContactAccount` class which I hope to apply to the 5-6 views that present lists of contacts.  a) it'll let us support contacts with multiple accounts. b) it'll dedupe a huge amount of this kind code. c) it'll fix this particular issue.

Signal/src/ViewControllers/SelectThreadViewController.m
charlesmchen: Great point.

Signal/src/ViewControllers/SendExternalFileViewController.m
charlesmchen: Done.

Signal/src/ViewControllers/SignalAttachment.swift
charlesmchen: Done.

Signal/src/ViewControllers/ThreadViewHelper.h
charlesmchen: I got interrupted. I've elaborated it.

Signal/src/ViewControllers/ThreadViewHelper.m
charlesmchen: Done.

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: Yes, sorry to be redundant but I had to do this to ensure we DRY.

Signal/src/ViewControllers/SignalsViewController.m
charlesmchen: It's not used yet since this is being added so late in the release process, but I'd like to use it to DRY up a bunch of logic inside this class.

Signal/src/ViewControllers/SendExternalFileViewController.m
charlesmchen: Done.

Signal/src/ViewControllers/OWSTableViewController.m
michaelkirk: huh, good to know.

Signal/src/ViewControllers/SendExternalFileViewController.m
```diff
@@ -0,0 +1,146 @@
...
+    return NO;
+}
+
+- (nullable UIView *)createHeader:(UIView *)superview
+{
+    OWSAssert(superview)
+
+        const CGFloat imageSize
+        = ScaleFromIPhone5To7Plus(40, 40);
+    const CGFloat imageLabelSpacing = ScaleFromIPhone5To7Plus(5, 5);
```
michaelkirk: If I had to guess, I'd say this is a convention to communicate that you've considered iPhone5->7+ in your layout without actually changing any behavior?

Signal/src/ViewControllers/SendExternalFileViewController.m
```diff
@@ -0,0 +1,146 @@
...
+    return header;
+}
+
+- (NSString *)formattedFileName
+{
+    OWSAssert(self.attachment) NSString *filename =
+        [self.attachment.filename stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
+    OWSAssert(filename.length > 0);
+    const NSUInteger kMaxFilenameLength = 50;
+    if (filename.length > kMaxFilenameLength) {
```
michaelkirk: What value is gained from this truncation vs. having the label do it (Which I see we're *also* doing)?

Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -266,10 +275,26 @@ class SignalAttachment: NSObject {
...
+    public class func pasteboardHasText() -> Bool {
+        if UIPasteboard.general.numberOfItems < 1 {
+            return false
+        }
+        let itemSet = IndexSet(integer:0)
+        guard let pasteboardUTITypes = UIPasteboard.general.types(forItemSet:itemSet) else {
+            return false
+        }
+        let pasteboardUTISet = Set<String>(pasteboardUTITypes[0])
+        return pasteboardUTISet.intersection(textUTISet).count > 0
```
michaelkirk: nit: just `return pasteboardUTISet.intersects(textUTISet)`

Signal/src/ViewControllers/ThreadViewHelper.m
```diff
@@ -0,0 +1,136 @@
...
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "ThreadViewHelper.h"
+#import <SignalServiceKit/TSDatabaseView.h>
+#import <SignalServiceKit/TSStorageManager.h>
+#import <SignalServiceKit/TSThread.h>
+#import <YapDatabase/YapDatabaseConnection.h>
+#import <YapDatabase/YapDatabaseViewChange.h>
+#import <YapDatabase/YapDatabaseViewConnection.h>
```
michaelkirk: nullability annotations here and in the .h (just in case my previous comment was lost in the rebase)

Signal/src/views/ContactTableViewCell.m
```diff
@@ -50,11 +52,11 @@ - (void)configureProgrammatically
 {
     const CGFloat kAvatarSize = 40.f;
     _avatarView = [UIImageView new];
-    _avatarView.contentMode = UIViewContentModeScaleToFill;
     _avatarView.image = [UIImage imageNamed:@"empty-group-avatar"];
     // applyRoundedBorderToImageView requires the avatar to have
     // the correct size.
     _avatarView.frame = CGRectMake(0, 0, kAvatarSize, kAvatarSize);
+    _avatarView.contentMode = UIViewContentModeScaleToFill;
```
michaelkirk: I thought we found it was better to set the content mode *before* the frame to avoid distortion.

Signal/src/ViewControllers/SendExternalFileViewController.m
charlesmchen: I'm taking out `NSLineBreakByTruncatingMiddle`.  It's because we want to center the filename & the file icon.  We could do this using `NSLineBreakByTruncatingMiddle` and a low-priority constraint on the width of the wrapping view, but I try to avoid using priority in my auto-layout code as much as possible because it _very_ quickly leads to unmaintainable code.

Signal/src/ViewControllers/SignalAttachment.swift
charlesmchen: Would that be a nil-test or empty-test on the set?  I feel like even if its an empty-test, that's the kind of language feature I don't like to use because it's not readable.

charlesmchen: To go on a bit, with complicated math I always try to wrap everything in parens to make the order of operations clear. Sure, you can rely upon operator precedence but a) it's very easy to misread that b) it's easy to make a mistake.

Signal/src/ViewControllers/ThreadViewHelper.m
charlesmchen: thanks.

Signal/src/views/ContactTableViewCell.m
charlesmchen: Whew!  I thought I was applying the same fix here but I got it backwards.

Signal/src/ViewControllers/SendExternalFileViewController.m
charlesmchen: No, I just wanted to defer cleaning these up until I tuned things on iPhone 7+ which I have now done.

Signal/src/ViewControllers/SignalAttachment.swift
michaelkirk: I'm not sure I understand your comments.

Did you notice I suggested `intersect**s**` (which returns a boolen) not `intersect**tion**` which computes the elements of the intersection.

charlesmchen: Oh, yeah I misread your comment.  But I didn't use that method because I didn't think it existed. And trying to use it, I get a build error.

screen shot 2017-04-26 at 4 05 28 pm

michaelkirk: o_0 Guess that one didn't make it to swift from NSSet.

PR #2043, opened 2017-04-26
support arbitrary extensions in UIDocumentPicker

michaelkirk:
1. Make sure we can pick any-ole file extension, like "foo.aslkdjhaskjdhsad"

2. We don't support sending bundles (i.e. directories). Instead of making them unselectable, in the DocumentPicker, let the user pick it, but then show them a message that tells them how to fix the problem (zip the file first).

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/SignalAttachment.swift
+++ b/Signal/src/ViewControllers/SignalAttachment.swift
@@ -12,7 +12,6 @@ enum SignalAttachmentError: Error {
     case couldNotParseImage
     case couldNotConvertToJpeg
     case invalidFileFormat
-    case unknownType
 }
 
 extension SignalAttachmentError: LocalizedError {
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -114,10 +111,6 @@ class SignalAttachment: NSObject {
         self.dataUTI = dataUTI
         self.filename = filename
         super.init()
-
-        if self.mimeType == nil {
```
michaelkirk: mimetype can no longer be nil.

Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -174,15 +167,15 @@ class SignalAttachment: NSObject {
...
+    var mimeType: String {
         if dataUTI == SignalAttachment.kOversizeTextAttachmentUTI {
             return OWSMimeTypeOversizeTextMessage
         }
         if dataUTI == SignalAttachment.kUnknownTestAttachmentUTI {
             return OWSMimeTypeUnknownForTests
         }
         guard let mimeType = UTTypeCopyPreferredTagWithClass(dataUTI as CFString, kUTTagClassMIMEType) else {
-            return nil
+            return OWSMimeTypeApplicationOctetStream
```
michaelkirk: We want to send all attachments. We need *some* mimetype for any attachment we send, so use a generic mimetype when we can't ascertain one from the UTI or extension.

michaelkirk:
in 14f956f3

PR #2044, opened 2017-04-27
Improve errors in the 'export with Signal' feature.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -268,30 +268,64 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
         NSString *filename = url.lastPathComponent;
         if ([filename stringByDeletingPathExtension].length < 1) {
             DDLogError(@"Application opened with URL invalid filename: %@", url);
+            [self showErrorAlertWithTitle:
+                      NSLocalizedString(@"EXPORT_WITH_SIGNAL_ERROR_TITLE",
+                          @"Title for the alert indicating the 'export with signal' attachment had an error.")
+                                  message:NSLocalizedString(@"EXPORT_WITH_SIGNAL_ERROR_MESSAGE_INVALID_FILENAME",
+                                              @"Message for the alert indicating the 'export with signal' file had an "
+                                              @"invalid filename.")];
             return NO;
         }
         NSString *fileExtension = [filename pathExtension];
         if (fileExtension.length < 1) {
             DDLogError(@"Application opened with URL missing file extension: %@", url);
+            [self showErrorAlertWithTitle:
+                      NSLocalizedString(@"EXPORT_WITH_SIGNAL_ERROR_TITLE",
+                          @"Title for the alert indicating the 'export with signal' attachment had an error.")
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -1025,7 +1025,7 @@ - (void)initializeCollectionViewLayout
 {
     [self.collectionView.collectionViewLayout setMessageBubbleFont:[UIFont ows_dynamicTypeBodyFont]];
 
-    self.collectionView.showsVerticalScrollIndicator = NO;
```
michaelkirk: Just sneaking this in? Bold move! 👍

Signal/src/AppDelegate.m
```diff
@@ -268,30 +268,64 @@ - (BOOL)application:(UIApplication *)application
...
+                          @"Title for the alert indicating the 'export with signal' attachment had an error.")
+                                  message:NSLocalizedString(@"EXPORT_WITH_SIGNAL_ERROR_MESSAGE_INVALID_FILENAME",
+                                              @"Message for the alert indicating the 'export with signal' file had an "
+                                              @"invalid filename.")];
             return NO;
         }
         NSString *fileExtension = [filename pathExtension];
         if (fileExtension.length < 1) {
             DDLogError(@"Application opened with URL missing file extension: %@", url);
+            [self showErrorAlertWithTitle:
```
michaelkirk: Having an error is better than failing silently, but we actually want to support this case right?

Signal/src/AppDelegate.m
```diff
@@ -268,30 +268,64 @@ - (BOOL)application:(UIApplication *)application
...
+                          @"Title for the alert indicating the 'export with signal' attachment had an error.")
+                                  message:NSLocalizedString(@"EXPORT_WITH_SIGNAL_ERROR_MESSAGE_UNKNOWN_TYPE",
+                                              @"Message for the alert indicating the 'export with signal' file had "
+                                              @"unknown type.")];
             return NO;
         }
         NSString *utiType = [MIMETypeUtil utiTypeForFileExtension:fileExtension];
         if (utiType.length < 1) {
             DDLogError(@"Application opened with URL of unknown UTI type: %@", url);
+            [self showErrorAlertWithTitle:
```
michaelkirk: Having an error is better than failing silently, but we actually want to support this case right?

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: Sorry that was for debugging. Will remove.

Signal/src/AppDelegate.m
charlesmchen: Unfortunately "open with Signal" isn't presented for files without a file extension - or for files of unknown file extension.

Signal/src/AppDelegate.m
charlesmchen: Unfortunately "open with Signal" isn't presented for files without a file extension - or for files of unknown file extension.

PR #2046, opened 2017-04-27
Ignore “open with Signal” if there is an ongoing call.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -265,6 +265,12 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
             DDLogWarn(@"Application opened with an unknown URL action: %@", url.host);
         }
     } else if ([url.scheme.lowercaseString isEqualToString:@"file"]) {
+
+        if ([Environment getCurrent].callService.call != nil) {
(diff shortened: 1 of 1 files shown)


Signal/src/AppDelegate.m
```diff
@@ -363,6 +369,9 @@ - (void)showErrorAlertWithTitle:(NSString *)title message:(NSString *)message
     [controller addAction:[UIAlertAction actionWithTitle:NSLocalizedString(@"OK", nil)
                                                    style:UIAlertActionStyleDefault
                                                  handler:nil]];
+    [[Environment getCurrent].signalsViewController.presentedViewController presentViewController:controller
```
michaelkirk: We're still in the didLaunchWithOptions at this point. I think we should dispatch_async(dispatch_get_main_queue() this.

michaelkirk: Sorry, that's not right. We're not in didFinishLaunching, but my point stands.

michaelkirk: I'm running in circles over here. The real issue is that presentedViewController is nil.

Here's how I fixed it: https://github.com/WhisperSystems/Signal-iOS/commit/1b61c3b0ce8dc6f083f4ad749be53b8c8509fea4

If that LGTY, feel free to merge away.

PR #2047, opened 2017-04-27
Allow importing of any file that's not a directory

michaelkirk:
PTAL @charlesmchen


michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/48971478dd2d8af38e8adb44a36c2209e7da0a28

PR #2049, opened 2017-04-27
Actually *use* the delay time to enqueu the retry

michaelkirk:
Also, use a consistent type (double) rather than a mishmash of float and
CGFloat.

PTAL @charlesmchen


michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/79a2a7b675808838a15e5406a74bc204b9c54791

PR #2053, opened 2017-04-28
Improve contact-related views.

charlesmchen:
* Add support for contacts with more than one Signal account using ContactAccount class.
* Use `OWSTableViewController` in contact-related views.
* Let users add non-contacts to groups.
* Improve the "new group" and "edit group" views.
* Add utility methods for displaying alerts.
* Warn users before discarding unsaved changes in "edit group" view.
* Pull out "contact view helper" to de-duplicate common logic among contact-related views.
* Pull out "group view helper" to de-duplicate common logic among group-related views.
* Pull out new base class for view used to add accounts to groups or the block list.

What's not included:

* We should also do a similar conversion of "new 1:1 conversation" view, applying the "contact view helper" and `OWSTableViewController` which will let us greatly slim down that view, improve search, make it easier to start conversations with non-contacts, etc.

PTAL @michaelkirk


Signal/src/AppDelegate.m
```diff
@@ -274,23 +275,25 @@ - (BOOL)application:(UIApplication *)application
         NSString *filename = url.lastPathComponent;
         if ([filename stringByDeletingPathExtension].length < 1) {
             DDLogError(@"Application opened with URL invalid filename: %@", url);
-            [self showErrorAlertWithTitle:
```
charlesmchen: I've added a utility function for showing simple alerts.  

* This will DRY up boilerplate.
* We're often not presenting alerts in a safe way, so some alerts are not appearing.  

I've made a card to apply these utility methods in more places.

Signal/src/ViewControllers/AddToBlockListViewController.h
```diff
@@ -2,8 +2,8 @@
 //  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
-#import <UIKit/UIKit.h>
+#import "SelectRecipientViewController.h"
 
-@interface AddToBlockListViewController : UIViewController
+@interface AddToBlockListViewController : SelectRecipientViewController
```
charlesmchen: I've pulled out a common base case `SelectRecipientViewController` from this view.  It can be used in any context where we're selecting a signal account, i.e. adding a user to a block list, adding a member to a new or existing group, etc.

Signal/src/ViewControllers/BlockListUIUtils.h
```diff
@@ -18,32 +19,32 @@ typedef void (^BlockActionCompletionBlock)(BOOL isBlocked);
 
 #pragma mark - Block
 
-+ (void)showBlockContactActionSheet:(Contact *)contact
-                 fromViewController:(UIViewController *)fromViewController
-                    blockingManager:(OWSBlockingManager *)blockingManager
-                    contactsManager:(OWSContactsManager *)contactsManager
-                    completionBlock:(nullable BlockActionCompletionBlock)completionBlock;
```
charlesmchen: Our contact-related views should deal in `ContactAccount` (which specifies which recipientId should be used for contacts with more than one), not `Contact`.

Signal/src/ViewControllers/BlockListUIUtils.m
```diff
@@ -62,6 +32,20 @@ + (void)showBlockPhoneNumberActionSheet:(NSString *)phoneNumber
                            completionBlock:completionBlock];
 }
 
++ (void)showBlockContactAccountActionSheet:(ContactAccount *)contactAccount
+                        fromViewController:(UIViewController *)fromViewController
+                           blockingManager:(OWSBlockingManager *)blockingManager
+                           contactsManager:(OWSContactsManager *)contactsManager
+                           completionBlock:(nullable BlockActionCompletionBlock)completionBlock
+{
+    NSString *displayName = [contactsManager displayNameForContactAccount:contactAccount];
```
charlesmchen: I haven't yet done the work to store the phone number "names/labels" for contacts with more than one account, e.g. "home" or "work".  That should be easy. I've made a card.

Signal/src/ViewControllers/ContactAccount.h
```diff
@@ -0,0 +1,34 @@
...
+// signal account) tuple.
+@interface ContactAccount : NSObject
+
+@property (nonatomic) Contact *contact;
+
+// An E164 value identifying the signal account.
+@property (nonatomic) NSString *recipientId;
+
+//// TODO:
+//@property (nonatomic) NSString *displayName;
```
charlesmchen: I'm leaving this class a bit incomplete since I'm going to revisit it almost immediately.  See above.

Signal/src/ViewControllers/ContactsViewHelper.h
```diff
@@ -0,0 +1,63 @@
...
+- (BOOL)shouldHideLocalNumber;
+
+@end
+
+#pragma mark -
+
+@class OWSContactsManager;
+@class OWSBlockingManager;
+
+@interface ContactsViewHelper : NSObject
```
charlesmchen: This helper reduces the size of contact-related views by almost half.  It could be a shared singleton, but I haven't bothered making it one.

Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -0,0 +1,261 @@
...
+    for (NSString *searchTerm in searchTerms) {
+        if (![self doesContactAccount:contactAccount matchSearchTerm:searchTerm]) {
+            return NO;
+        }
+    }
+
+    return YES;
+}
+
+- (NSArray<ContactAccount *> *)contactAccountsMatchingSearchString:(NSString *)searchText
```
charlesmchen: `OWSContactsSearcher` assumed that there was a single search term.  I'm hoping to remove `OWSContactsSearcher`.

Signal/src/ViewControllers/GroupViewHelper.h
```diff
@@ -0,0 +1,44 @@
...
+
+- (UIViewController *)fromViewController;
+
+@end
+
+#pragma mark -
+
+typedef void (^GroupViewSuccessBlock)();
+
+@interface GroupViewHelper : NSObject
```
charlesmchen: This is another helper which helps us DRY up our view controllers.

Signal/src/ViewControllers/GroupViewHelper.m
```diff
@@ -0,0 +1,204 @@
...
+- (void)imagePickerController:(UIImagePickerController *)picker didFinishPickingMediaWithInfo:(NSDictionary *)info
+{
+    OWSAssert([NSThread isMainThread]);
+    OWSAssert(self.delegate);
+
+    UIImage *rawAvatar = [info objectForKey:UIImagePickerControllerOriginalImage];
+
+    if (rawAvatar) {
+        // TODO: This is busted.
+        UIImage *resizedAvatar = [rawAvatar resizedImageToFitInSize:CGSizeMake(100.00, 100.00) scaleIfSmaller:NO];
```
charlesmchen: I think there's a bug around device scale in this method.

Signal/src/ViewControllers/NewGroupViewController.h
```diff
@@ -2,27 +2,17 @@
...
-@property (nonatomic) IBOutlet UITextField *nameGroupTextField;
-@property (nonatomic) IBOutlet UIButton *groupImageButton;
-@property (nonatomic) IBOutlet UIView *tapToDismissView;
-@property (nonatomic) IBOutlet UILabel *addPeopleLabel;
-@property (nonatomic) UIImage *groupImage;
-@property (nonatomic) TSGroupModel *groupModel;
+@property (nonatomic, weak) id<OWSConversationSettingsViewDelegate> delegate;
 
 @property (nonatomic) BOOL shouldEditGroupNameOnAppear;
 @property (nonatomic) BOOL shouldEditAvatarOnAppear;
```
charlesmchen: I've separated the "new group" and "update group" out into separate classes - the common logic resides in the group view helper.

charlesmchen:
PTAL @michaelkirk 

I've significantly changed this PR.  

* I've renamed `ContactAccount` to `SignalAccount`.  
* I've fixed race conditions in the contact manager and simplified observation of changes to the contact/account state.
* I've added extraction of the phone number labels from the device contacts.

Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -0,0 +1,218 @@
...
+            [signalAccounts addObject:signalAccount];
+        }
+    }
+    self.signalAccountMap = signalAccountMap;
+    self.signalAccounts = signalAccounts;
+
+    [self.delegate contactsViewHelperDidUpdateContacts];
+}
+
+- (BOOL)doesSignalAccount:(SignalAccount *)signalAccount matchSearchTerm:(NSString *)searchTerm
```
michaelkirk: Do these search methods deprecate the OWSContactSearcher?

Signal/src/ViewControllers/GroupViewHelper.m
```diff
@@ -0,0 +1,204 @@
...
+ */
+- (void)imagePickerController:(UIImagePickerController *)picker didFinishPickingMediaWithInfo:(NSDictionary *)info
+{
+    OWSAssert([NSThread isMainThread]);
+    OWSAssert(self.delegate);
+
+    UIImage *rawAvatar = [info objectForKey:UIImagePickerControllerOriginalImage];
+
+    if (rawAvatar) {
+        // TODO: There may be a bug here.
```
michaelkirk: can you expound on this please while it's still (hopefully) fresh in mind?

Signal/src/ViewControllers/GroupViewHelper.m
```diff
@@ -0,0 +1,204 @@
...
+        UIImage *resizedAvatar = [rawAvatar resizedImageToFitInSize:CGSizeMake(100.00, 100.00) scaleIfSmaller:NO];
```
michaelkirk: Does this need to be rebased against the recent group avatar changes? Expecting to see something like 210px IIRC.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -3052,6 +3043,26 @@ - (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
     self.userHasScrolled = YES;
 }
 
+#pragma mark - OWSConversationSettingsViewDelegate
```
michaelkirk: 👏 👏  for the explicit delegate. Seems much clearer than the unwind segue.

Signal/src/ViewControllers/NewGroupViewController.h
michaelkirk: :+1:

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -3,40 +3,53 @@
...
-@property (nonatomic, readonly, strong) OWSMessageSender *messageSender;
-@property (nonatomic, readonly, strong) OWSContactsManager *contactsManager;
+@property (nonatomic, readonly) OWSTableViewController *tableViewController;
+@property (nonatomic, readonly) UIImageView *avatarView;
+@property (nonatomic, readonly) UITextField *groupNameTextField;
 
-@property (nonatomic, readonly) OWSBlockingManager *blockingManager;
-@property (nonatomic, readonly) NSArray<NSString *> *blockedPhoneNumbers;
+@property (nonatomic, nullable) UIImage *groupAvatar;
+@property (nonatomic, nullable) NSMutableSet<NSString *> *memberRecipientIds;
```
michaelkirk: nit: doesn't seem like it's every nil.

Signal/src/ViewControllers/UpdateGroupViewController.m
```diff
@@ -0,0 +1,504 @@
...
+
+    OWSAssert(self.thread);
+    OWSAssert(self.thread.groupModel);
+    OWSAssert(self.thread.groupModel.groupMemberIds);
+
+    [self.memberRecipientIds addObjectsFromArray:self.thread.groupModel.groupMemberIds];
+    self.previousMemberRecipientIds = [NSSet setWithArray:self.thread.groupModel.groupMemberIds];
+
+    self.title = NSLocalizedString(@"EDIT_GROUP_DEFAULT_TITLE", @"The default title for the 'update group' view.");
+    self.navigationItem.leftBarButtonItem =
```
michaelkirk: I *do* like the "confirm before leaving" 👍 

But this breaks the interactive pop gesture (e.g. swipe to go back).  

Maybe "editing a group" should be a modal activity anyway, rather than pushed, since we're not really "drilling down".

That would solve the problem of people thinking they can "swipe back".

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -66,243 +79,353 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder
...
+                                                                                   action:@selector(avatarTouched:)]];
+    avatarView.userInteractionEnabled = YES;
+
+    return firstSectionHeader;
+}
+
+- (void)avatarTouched:(UIGestureRecognizer *)sender
+{
+    if (sender.state == UIGestureRecognizerStateRecognized) {
+        [self showChangeGroupAvatarUI:nil];
```
michaelkirk: nit: please label your params, otherwise I have to go to the caller to have a clue what this `nil` is about.

e.g.  `showChangeGroupAvatarUIWithSender:` or just drop the param since we're not using it.

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -66,243 +79,353 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder
...
+    for (SignalAccount *signalAccount in signalAccounts) {
+        [nonContactMemberRecipientIds removeObject:signalAccount.recipientId];
     }
 
-    if ([self isCurrentUserContact:contact]) {
-        // We never want to add ourselves to a group.
-        return YES;
+    // Non-contact Members
+
+    if (nonContactMemberRecipientIds.count > 0 || signalAccounts.count < 1) {
```
michaelkirk: expected to see something like this in the update-group-controller.

BOOKMARK: This is as far as my review got for now.

michaelkirk: edit: disregard my previous comment - I see now why this is different for edit vs. create. 👍

Signal/src/ViewControllers/SelectThreadViewController.m
```diff
@@ -216,33 +195,31 @@ - (void)updateTableContents
...
+                                                      [weakSelf signalAccountWasSelected:signalAccount];
+                                                  }
+                                              }];
         return;
     }
 
     __block TSThread *thread = nil;
     [[TSStorageManager sharedManager].dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
-        thread = [TSContactThread getOrCreateThreadWithContactId:recipientId transaction:transaction];
+        thread = [TSContactThread getOrCreateThreadWithContactId:signalAccount.recipientId transaction:transaction];
```
michaelkirk: just a rant: 

I'd like to move towards passing SignalAccounts (or something 1:1 with it) in pretty much all the places where we're passing e164 strings around.

e.g. to avoid the "how do I know if this is e164 or not?" problem that you have when you have a parameter named "phoneNumber"

and for future work where we can't assume a recipient id is a e164 phone number.

e.g. in this case `[TSContactThread getOrCreateThreadWithSignalAccount:signalAccount transaction:transaction];

I am *not* advocating that we make any specific changed WRT that in this PR, or all at once, as it will touch a bajillion places, but if you agree, I'd like to keep it in mind for future work.

Signal/src/ViewControllers/ShowGroupMembersViewController.m
```diff
@@ -250,8 +192,8 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
                                                      [BlockListUIUtils
                                                          showUnblockPhoneNumberActionSheet:recipientId
```
michaelkirk: after blocking/unblocking the corresponding "Blocked"/"Unblocked" label doesn't change until you reload the view.

Signal/src/ViewControllers/ShowGroupMembersViewController.m
```diff
@@ -315,70 +261,41 @@ - (void)showContactInfoViewForMember:(NSIndexPath *)indexPath
...
             view.allowsAddingToAddressBook = YES;
             [self.navigationController pushViewController:view animated:YES];
         }
     }
 }
 
-- (void)showConversationViewForMember:(NSIndexPath *)indexPath
+- (void)showConversationViewForRecipientId:(NSString *)recipientId
 {
-    NSString *recipientId;
```
michaelkirk: nice job on all the red.

Signal/src/ViewControllers/SignalAccount.h
```diff
@@ -0,0 +1,52 @@
...
+//
+// This is the key property of this class and it
+// will always be non-null.
+@property (nonatomic, readonly) NSString *recipientId;
+
+// This property is optional and will not be set for
+// non-contact account.
+@property (nonatomic, nullable) Contact *contact;
+
+@property (nonatomic) BOOL isMultipleAccountContact;
```
michaelkirk: small nit: maybe hasMultipleAccountContact?

Signal/src/ViewControllers/SignalAccount.h
```diff
@@ -0,0 +1,52 @@
...
+@property (nonatomic) NSString *multipleAccountLabel;
+
+- (instancetype)init NS_UNAVAILABLE;
+
+- (instancetype)initWithSignalRecipient:(SignalRecipient *)signalRecipient;
+
+- (instancetype)initWithRecipientId:(NSString *)recipientId;
+
+// In most cases this should be non-null. This should only
+// be non-null in the case where the SignalRecipient was
```
michaelkirk: typo: "this should only be null..."

Signal/src/ViewControllers/SignalAccount.m
```diff
@@ -0,0 +1,62 @@
...
+    }
+    return self;
+}
+
+- (nullable SignalRecipient *)signalRecipient
+{
+    OWSAssert([NSThread isMainThread]);
+
+    if (!_signalRecipient) {
+        [[TSStorageManager sharedManager].dbConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
```
michaelkirk: These buried transactions are kind of scary. How much of a pain would it be for the method to be `signalRecipientWithTransaction:`

michaelkirk: BOOKMARK: stopping here for the day. 😪

Signal/src/ViewControllers/AddToGroupViewController.h
```diff
@@ -0,0 +1,21 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "SelectRecipientViewController.h"
+
+@protocol AddToGroupViewControllerDelegate <NSObject>
```
charlesmchen: Done.

Signal/src/ViewControllers/AddToGroupViewController.m
```diff
@@ -0,0 +1,127 @@
...
+        [BlockListUIUtils showUnblockPhoneNumberActionSheet:phoneNumber
+                                         fromViewController:self
+                                            blockingManager:helper.blockingManager
+                                            contactsManager:helper.contactsManager
+                                            completionBlock:^(BOOL isBlocked) {
+                                                if (isBlocked) {
+                                                    [weakSelf addToGroup:phoneNumber];
+                                                }
+                                            }];
+    } else {
```
charlesmchen: > be unselectable

Done. 

> show an alert that says "You cannot remove group members

Well, this dialog is used for adding members to groups, not removing, but you're right, they shouldn't be selectable. Done.  I am however adding the warning you're suggesting to the "edit group" view.

I've applied similar changes to the "add to block list" view. thanks.

Signal/src/ViewControllers/SignalAccount.h
```diff
@@ -0,0 +1,52 @@
...
+
+// This property is optional and will not be set for
+// non-contact account.
+@property (nonatomic, nullable) Contact *contact;
+
+@property (nonatomic) BOOL isMultipleAccountContact;
+
+// For contacts with more than one signal account,
+// this is a label for the account.
+@property (nonatomic) NSString *multipleAccountLabel;
```
michaelkirk: nit, it's not a label, can we call it something else?  maybe `multipleAccountLabelText`?

Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -0,0 +1,218 @@
...
+
+- (instancetype)init
+{
+    self = [super init];
+    if (!self) {
+        return self;
+    }
+
+    _blockingManager = [OWSBlockingManager sharedManager];
+    self.blockedPhoneNumbers = [_blockingManager blockedPhoneNumbers];
```
charlesmchen: Done.

Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -0,0 +1,218 @@
...
+
+    return self.signalAccountMap[recipientId];
+}
+
+- (BOOL)isSignalAccountHidden:(SignalAccount *)signalAccount
+{
+    OWSAssert([NSThread isMainThread]);
+
+    if ([self.delegate shouldHideLocalNumber] && [self isCurrentUser:signalAccount]) {
+        // We never want to add ourselves to a group.
```
charlesmchen: Done.

Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -0,0 +1,218 @@
...
+
+    NSMutableDictionary<NSString *, SignalAccount *> *signalAccountMap = [NSMutableDictionary new];
+    NSMutableArray<SignalAccount *> *signalAccounts = [NSMutableArray new];
+    for (SignalAccount *signalAccount in self.contactsManager.signalAccounts) {
+        if (![self isSignalAccountHidden:signalAccount]) {
+            signalAccountMap[signalAccount.recipientId] = signalAccount;
+            [signalAccounts addObject:signalAccount];
+        }
+    }
+    self.signalAccountMap = signalAccountMap;
```
charlesmchen: Done.

Signal/src/ViewControllers/ContactsViewHelper.m
charlesmchen: Yup.

Signal/src/ViewControllers/ContactsViewHelper.h
```diff
@@ -0,0 +1,59 @@
...
+@property (nonatomic, readonly) OWSBlockingManager *blockingManager;
+
+@property (nonatomic, readonly) NSDictionary<NSString *, SignalAccount *> *signalAccountMap;
+@property (nonatomic, readonly) NSArray<SignalAccount *> *signalAccounts;
+
+@property (nonatomic, readonly) NSArray<NSString *> *blockedPhoneNumbers;
+
+- (nullable SignalAccount *)signalAccountForRecipientId:(NSString *)recipientId;
+
+- (nullable NSArray<NSString *> *)blockedPhoneNumbers;
```
charlesmchen: Done.

Signal/src/contact/OWSContactsManager.m
```diff
@@ -511,6 +556,46 @@ - (NSString *_Nonnull)displayNameForContact:(Contact *)contact
     return displayName;
 }
 
+- (NSString *_Nonnull)displayNameForSignalAccount:(SignalAccount *)signalAccount
```
michaelkirk: There are several methods here which seem like they could be taken out of the contact manager to make them more easily testable and to keep OWSContactsManager from continuing to endlessly bloat.

e.g. all the ones that take a SignalAccount parameter are good candidates to be moved to the SignalAccount model, or if you want a "dumber" model, to a presenter.

Signal/src/ViewControllers/ContactsViewHelper.h
```diff
@@ -0,0 +1,59 @@
...
+
+// This method is faster than OWSBlockingManager but
+// is only safe to be called on the main thread.
+//
+// Returns true if _any_ number associated with this contact
+// is blocked.
+//
+// TODO: Is this obsolete?
```
charlesmchen: Yes, there's a bunch of stuff we'll be able to clean up in https://github.com/WhisperSystems/Signal-iOS/pull/2058 or afterward.

Signal/src/contact/OWSContactsManager.m
michaelkirk: nit: unnecessary `_Nonnull`

since aren't we NS_ASSUME_NONNULL?

Signal/src/ViewControllers/ContactsViewHelper.h
```diff
@@ -0,0 +1,59 @@
...
+- (BOOL)isContactBlocked:(Contact *)contact;
+
+// This method is faster than OWSBlockingManager but
+// is only safe to be called on the main thread.
+- (BOOL)isRecipientIdBlocked:(NSString *)recipientId;
+
+- (NSString *)localNumber;
```
charlesmchen: I've wondered the same thing, but unfortunately what you want is a warning about the transitive closure over the logic used by any given method.  That doesn't lend itself to a naming convention.  I've added a comment though.

Signal/src/contact/OWSContactsManager.h
```diff
@@ -8,39 +8,44 @@
...
 
-@property (nullable, strong) CNContactStore *contactStore;
-@property (nonnull, readonly, strong) NSCache<NSString *, UIImage *> *avatarCache;
+@property (nonnull, readonly) NSCache<NSString *, UIImage *> *avatarCache;
 
-- (nonnull ObservableValue *)getObservableContacts;
+// signalAccountMap and signalAccounts hold the same data.
+// signalAccountMap is for lookup. signalAccounts contains the accounts
+// ordered by display order.
+@property (atomic, readonly) NSDictionary<NSString *, SignalAccount *> *signalAccountMap;
```
michaelkirk: If "signalAccountMap is for lookup", what is `signalAccountForRecipientId:` for? Can we get rid of one of them?

Signal/src/contact/OWSContactsManager.m
```diff
@@ -571,30 +656,30 @@ - (NSAttributedString *)formattedFullNameForRecipientId:(NSString *)recipientId
...
 }
 
-- (Contact * _Nullable)contactForPhoneIdentifier:(NSString * _Nullable)identifier {
-    if (!identifier) {
-        return nil;
-    }
-    return self.contactMap[identifier];
+- (nullable SignalAccount *)signalAccountForRecipientId:(nullable NSString *)recipientId
+{
+    OWSAssert(recipientId.length > 0);
```
michaelkirk: This assert conflicts with the fact that the `recipientId` param  is `nullable`

My hunch is that the param should not be nullable.

Signal/translations/bin/auto-genstrings
```diff
@@ -4,15 +4,6 @@ set -e
 
 SSK_DIR="../SignalServiceKit/src"
 
-pushd $SSK_DIR
```
michaelkirk: I get that this is annoying as is. But how are you going to ensure that you don't push the wrong strings to transifex?

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -316,11 +340,35 @@
...
 
-/* A format for the message of the alert if user tries to add a user to a group who is already in the group.  Embeds {{the blocked user's name or phone number}}. */
-"EDIT_GROUP_VIEW_ALREADY_IN_GROUP_ALERT_MESSAGE_FORMAT" = "%@ is already a member of this group.";
+/* a title for the contacts section of the 'new/update group' view. */
+"EDIT_GROUP_CONTACTS_SECTION_TITLE" = "Contacts";
+
+/* The default title for the 'update group' view. */
+"EDIT_GROUP_DEFAULT_TITLE" = "Edit Group";
+
+/* A label the cell that lets you add a new member to a group. */
```
michaelkirk: comment typo.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -316,11 +340,35 @@
 /* table cell label in conversation settings */
 "EDIT_GROUP_ACTION" = "Edit Group";
 
-/* A format for the message of the alert if user tries to add a user to a group who is already in the group.  Embeds {{the blocked user's name or phone number}}. */
-"EDIT_GROUP_VIEW_ALREADY_IN_GROUP_ALERT_MESSAGE_FORMAT" = "%@ is already a member of this group.";
+/* a title for the contacts section of the 'new/update group' view. */
+"EDIT_GROUP_CONTACTS_SECTION_TITLE" = "Contacts";
+
+/* The default title for the 'update group' view. */
+"EDIT_GROUP_DEFAULT_TITLE" = "Edit Group";
```
michaelkirk: I think it's helpful to provide more context. Is it an alert title? navbar title? Table section title?

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -714,17 +759,29 @@
 /* Text for button to start a new conversation with a non-contact */
 "NEW_CONVERSATION_FOR_NON_CONTACT_FORMAT" = "New conversation with %@";
 
+/* A label the cell that lets you add a new non-contact member to a group. */
```
michaelkirk: comment typo

Signal/src/ViewControllers/UpdateGroupViewController.m
```diff
@@ -0,0 +1,504 @@
...
+    [self updateAvatarView];
+
+    UITextField *groupNameTextField = [UITextField new];
+    _groupNameTextField = groupNameTextField;
+    if (self.thread) {
+        self.groupNameTextField.text = self.thread.groupModel.groupName;
+    }
+    groupNameTextField.textColor = [UIColor blackColor];
+    groupNameTextField.font = [UIFont ows_dynamicTypeTitle2Font];
+    groupNameTextField.placeholder = NSLocalizedString(@"NEW_GROUP_NAMEGROUP_REQUEST_DEFAULT", @"");
```
michaelkirk: can you bring this translation comment back?

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -781,6 +838,33 @@
...
+/* Label for 'Mobile' phone numbers. */
+"PHONE_NUMBER_TYPE_MOBILE" = "Mobile";
+
+/* Label for 'Other FAX' phone numbers. */
+"PHONE_NUMBER_TYPE_OTHER_FAX" = "Other Fax";
+
+/* Label for 'Pager' phone numbers. */
+"PHONE_NUMBER_TYPE_PAGER" = "Pager";
+
+/* Label for 'Unknown' phone numbers. */
```
michaelkirk: In context of all the other types this probably makes sense, but "unknown phone numbers" is kind of vague. Maybe "Label used when we don't what kind of phone number it is (e.g. mobile/work/home)"

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -757,7 +814,7 @@
 "NOTIFICATIONS_SOUND" = "Notification Sounds";
 
 /* No comment provided by engineer. */
-"OK" = "Ok";
+"OK" = "OK";
```

Signal/src/ViewControllers/UpdateGroupViewController.m
charlesmchen: Great idea, done.

Signal/src/ViewControllers/UpdateGroupViewController.m
```diff
@@ -0,0 +1,504 @@
...
+                                    [weakSelf.groupViewHelper
+                                        showRemoveFromGroupAlertForSignalAccount:signalAccount
+                                                              fromViewController:weakSelf
+                                                                 contactsManager:helper.contactsManager
+                                                                    successBlock:^{
+                                                                        [weakSelf removeSignalAccount:signalAccount];
+                                                                    }];
+                                } else {
+                                    [weakSelf.groupViewHelper
+                                        showRemoveFromGroupAlertForRecipientId:recipientId
```
charlesmchen: Good point. I've removed remove confirmation alerts from the "new group" and "update group" views.

Signal/src/ViewControllers/GroupViewHelper.m
charlesmchen: 5

charlesmchen: https://github.com/WhisperSystems/Signal-iOS/pull/2055

Signal/src/ViewControllers/GroupViewHelper.m
charlesmchen: https://github.com/WhisperSystems/Signal-iOS/pull/2055 depends on this PR and will be merged after it.

Signal/src/ViewControllers/NewGroupViewController.m
charlesmchen: Done.

Signal/src/ViewControllers/NewGroupViewController.m
charlesmchen: Done.

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -66,243 +79,353 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder
...
+            [nonContactsSection
+                addItem:[OWSTableItem itemWithCustomCellBlock:^{
+                    NewGroupViewController *strongSelf = weakSelf;
+                    if (!strongSelf) {
+                        return (ContactTableViewCell *)nil;
+                    }
+
+                    ContactTableViewCell *cell = [ContactTableViewCell new];
+                    SignalAccount *signalAccount = [helper signalAccountForRecipientId:recipientId];
+                    BOOL isCurrentMember = [weakSelf.memberRecipientIds containsObject:recipientId];
```
charlesmchen: Done.

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -66,243 +79,353 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder
...
+
+            [nonContactsSection
+                addItem:[OWSTableItem itemWithCustomCellBlock:^{
+                    NewGroupViewController *strongSelf = weakSelf;
+                    if (!strongSelf) {
+                        return (ContactTableViewCell *)nil;
+                    }
+
+                    ContactTableViewCell *cell = [ContactTableViewCell new];
+                    SignalAccount *signalAccount = [helper signalAccountForRecipientId:recipientId];
```
charlesmchen: Done.

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -66,243 +79,353 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder
...
+                        cell.accessoryMessage = NSLocalizedString(
+                            @"NEW_GROUP_MEMBER_LABEL", @"An indicator that a user is a member of the new group.");
+                    } else if (isBlocked) {
+                        cell.accessoryMessage = NSLocalizedString(
+                            @"CONTACT_CELL_IS_BLOCKED", @"An indicator that a contact has been blocked.");
+                    } else {
+                        OWSAssert(cell.accessoryMessage == nil);
+                    }
+
+                    if (signalAccount) {
```
charlesmchen: Great question. When I started this `ContactAccount/SignalAccount` work, I intended it be like a "contact, recipientId" tuple + metadata, i.e. that contact would always be non-nil.  After the work was mostly done I realized that it be nicer if we also use `SignalAccount` to represent non-contacts - to simplify code paths like this one specifically.  

But I'm going to do that cleanup in a subsequent PR.  It was such a struggle to cut this PR off.

charlesmchen: And yes - that they aren't in our contacts.

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -66,243 +79,353 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder
...
-        return NO;
-    }
+    OWSTableSection *signalAccountSection = [OWSTableSection new];
+    signalAccountSection.headerTitle = NSLocalizedString(
+        @"EDIT_GROUP_CONTACTS_SECTION_TITLE", @"a title for the contacts section of the 'new/update group' view.");
+    if (signalAccounts.count > 0) {
+
+        if (nonContactMemberRecipientIds.count < 1) {
+            // We always want to offer a way to add non-contacts.
+            [signalAccountSection addItem:[self createAddNonContactItem]];
```
charlesmchen: I've elaborated the comment greatly.

michaelkirk:
Another UX thought, in master when you update a group you see an "Update" button to save your changes, but not so in this branch.

Also what do you think about replacing the "+" in the new group view with "Create"?

I realize that we now give people the option to save when they hit "back", which I think is good, but I still think we should give people an affirmative "Update" button (top right bar button)

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -66,243 +79,353 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder
...
 
-    for (PhoneNumber *phoneNumber in contact.parsedPhoneNumbers) {
-        if ([_blockedPhoneNumbers containsObject:phoneNumber.toE164]) {
-            return YES;
+        for (SignalAccount *signalAccount in signalAccounts) {
+            [signalAccountSection
+                addItem:[OWSTableItem itemWithCustomCellBlock:^{
+                    NewGroupViewController *strongSelf = weakSelf;
+                    if (!strongSelf) {
+                        return (ContactTableViewCell *)nil;
```
charlesmchen: Instead I'm just going to remove the strongSelf checks since they should never fail.

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -66,243 +79,353 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder
...
+                    }
+
+                    ContactTableViewCell *cell = [ContactTableViewCell new];
+
+                    NSString *recipientId = signalAccount.recipientId;
+                    BOOL isCurrentMember = [weakSelf.memberRecipientIds containsObject:recipientId];
```
charlesmchen: Done.

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -340,254 +463,131 @@ - (void)createGroup
...
-    [mut addObjectsFromArray:_thread.groupModel.groupMemberIds];
-
-    _groupModel = [[TSGroupModel alloc] initWithTitle:_nameGroupTextField.text
-                                            memberIds:[[[NSSet setWithArray:mut] allObjects] mutableCopy]
-                                                image:_thread.groupModel.groupImage
-                                              groupId:_thread.groupModel.groupId];
-
-    [self.nameGroupTextField resignFirstResponder];
+    NSString *title = self.groupNameTextField.text;
+    NSMutableArray *recipientIds = [self.memberRecipientIds.allObjects mutableCopy];
```
charlesmchen: Done.

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -340,254 +463,131 @@ - (void)createGroup
...
+{
+    UIImage *image = (self.groupAvatar ?: [UIImage imageNamed:@"empty-group-avatar"]);
+    OWSAssert(image);
 
-    if ([UIImagePickerController isSourceTypeAvailable:UIImagePickerControllerSourceTypeSavedPhotosAlbum]) {
-        picker.mediaTypes = [[NSArray alloc] initWithObjects:(NSString *)kUTTypeImage, nil];
-        [self presentViewController:picker animated:YES completion:[UIUtil modalCompletionBlock]];
-    }
+    self.avatarView.image = image;
+    self.avatarView.layer.borderColor = [[UIColor lightGrayColor] CGColor];
```
charlesmchen: Yes, I was wondering the same thing. It definitely looks better for avatars with large white portions. I've added a card.

Signal/src/ViewControllers/SelectRecipientViewController.h
```diff
@@ -0,0 +1,37 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+@class SignalAccount;
+
+@protocol SelectRecipientViewControllerDelegate <NSObject>
```
charlesmchen: Done.

Signal/src/ViewControllers/AddToGroupViewController.m
```diff
@@ -0,0 +1,127 @@
...
+}
+
+- (BOOL)shouldHideLocalNumber
+{
+    return YES;
+}
+
+- (BOOL)shouldHideContacts
+{
+    return self.hideContacts;
```
charlesmchen: The idea was this:

* In "update group", you're making a small narrow edit to a group, usually adding one person.  So we depend entirely on the "add to group" view which will let you add by phone number or pick from a list of your contacts.
* In "new group", you're often going to want to pick many users so we include the contacts list right in the "new group" view.  We don't want to show the contacts list again in the "add by phone number" view.

Signal/src/ViewControllers/SelectRecipientViewController.m
```diff
@@ -0,0 +1,553 @@
...
+    phoneNumberLabel.textColor = [UIColor blackColor];
+    phoneNumberLabel.text
+        = NSLocalizedString(@"REGISTRATION_PHONENUMBER_BUTTON", @"Label for the phone number textfield");
+    return phoneNumberLabel;
+}
+
+- (UITextField *)phoneNumberTextField
+{
+    if (!_phoneNumberTextField) {
+        _phoneNumberTextField = [UITextField new];
```
charlesmchen: I actually did this and reverted it.  I couldn't figure out a way to make it look right.  If you pick some arbitrary width, the placeholder text & short group names look wrong. You don't want the underline to be too long or too short.  

Very happy to revisit this- I've added a card.

Signal/src/ViewControllers/SelectRecipientViewController.m
```diff
@@ -0,0 +1,553 @@
...
+    [self tryToSelectPhoneNumber];
+}
+
+- (void)tryToSelectPhoneNumber
+{
+    OWSAssert(self.delegate);
+
+    if (![self hasValidPhoneNumber]) {
+        OWSAssert(0);
+        return;
```
charlesmchen: Added logging, this should never happen.

Signal/src/ViewControllers/SelectRecipientViewController.m
```diff
@@ -0,0 +1,553 @@
...
+    NSString *rawPhoneNumber = [self.callingCode stringByAppendingString:self.phoneNumberTextField.text.digitsOnly];
+
+    NSMutableArray<NSString *> *possiblePhoneNumbers = [NSMutableArray new];
+    for (PhoneNumber *phoneNumber in
+        [PhoneNumber tryParsePhoneNumbersFromsUserSpecifiedText:rawPhoneNumber
+                                              clientPhoneNumber:[TSAccountManager localNumber]]) {
+        [possiblePhoneNumbers addObject:phoneNumber.toE164];
+    }
+    if ([possiblePhoneNumbers count] < 1) {
+        OWSAssert(0);
```
charlesmchen: Ditto added logic. This should never happen since the button should be disabled.

Signal/src/ViewControllers/SelectRecipientViewController.m
```diff
@@ -0,0 +1,553 @@
...
+
+- (void)countryCodeViewController:(CountryCodeViewController *)vc
+             didSelectCountryCode:(NSString *)countryCode
+                      countryName:(NSString *)countryName
+                      callingCode:(NSString *)callingCode
+{
+
+    [self updateCountryWithName:countryName callingCode:callingCode countryCode:countryCode];
+
+    [self textField:self.phoneNumberTextField shouldChangeCharactersInRange:NSMakeRange(0, 0) replacementString:@""];
```
charlesmchen: Yes, I should have added a comment.  The intent is to trigger the formatting logic with a no-op edit, since changing the country code potentially changes the formatting behavior. I've added a comment.

Signal/src/ViewControllers/SelectRecipientViewController.m
```diff
@@ -0,0 +1,553 @@
...
+        [countryCodeLabel autoVCenterInSuperview];
+
+        [countryRow addSubview:self.countryCodeButton];
+        [self.countryCodeButton autoPinEdgeToSuperviewEdge:ALEdgeRight withInset:20.f];
+        [self.countryCodeButton autoVCenterInSuperview];
+
+        // Phone Number Row
+        UIView *phoneNumberRow =
+            [self createRowWithHeight:kPhoneNumberRowHeight previousRow:countryRow superview:cell.contentView];
+        [phoneNumberRow addGestureRecognizer:[[OWSAnyTouchGestureRecognizer alloc]
```
charlesmchen: Okay, I've made this a tap.  We should discuss more generally where to use the "any touch" GR.  The idea is:

* "any touch" recognizes faster - as soon as the touch begins.
* "any touch" is way more permissive and accepts any number of touches / movement.  

I like to use it in places where there's no ambiguity about the user intent in the spirit of "be permissive about what you accept."

Signal/src/ViewControllers/ViewControllerUtils.m
```diff
@@ -96,6 +98,42 @@ + (NSString *)formatFileSize:(unsigned long)fileSize
...
+        [UIAlertController alertControllerWithTitle:title message:message preferredStyle:UIAlertControllerStyleAlert];
+
+    [alert addAction:[UIAlertAction actionWithTitle:buttonLabel style:UIAlertActionStyleDefault handler:nil]];
+
+    [self.topMostController presentViewController:alert animated:YES completion:nil];
+
+    return alert;
+}
+
++ (UIViewController *)topMostController
```
charlesmchen: Done.

Signal/src/ViewControllers/SelectThreadViewController.m
charlesmchen: That's really interesting. Let's discuss on video.

Signal/src/ViewControllers/ShowGroupMembersViewController.m
charlesmchen: Done.

Signal/src/ViewControllers/SignalAccount.h
charlesmchen: Done.

Signal/src/ViewControllers/SignalAccount.h
charlesmchen: Done.

Signal/src/ViewControllers/SignalAccount.m
charlesmchen: Good point. And I'm going to remove the caching of the `SignalRecipient` in this class.

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -66,243 +79,353 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder
...
+    [self.navigationController.navigationBar setTranslucent:NO];
 }
 
 - (void)viewDidAppear:(BOOL)animated
 {
     [super viewDidAppear:animated];
 
     if (self.shouldEditGroupNameOnAppear) {
-        [self.nameGroupTextField becomeFirstResponder];
+        [self.groupNameTextField becomeFirstResponder];
```
charlesmchen: Good point. I think I'm going to leave it auto-selecting the name field the first time we enter this view in order to encourage people to name their groups.  WDYT?

Signal/src/contact/OWSContactsManager.m
```diff
@@ -127,9 +82,8 @@ void onAddressBookChanged(ABAddressBookRef notifyAddressBook, CFDictionaryRef in
 
 - (void)handleAddressBookChanged
 {
-    [self pullLatestAddressBook];
-    [self intersectContacts];
     [self.avatarCache removeAllObjects];
+    [self pullLatestAddressBook];
```
charlesmchen: I've moved the clearing of the avatar cache into `updateWithContacts:`.

Signal/src/contact/OWSContactsManager.m
```diff
@@ -206,30 +160,148 @@ - (void)intersectContactsWithRetryDelay:(double)retryDelaySeconds
...
+            }
+        }
+
+        dispatch_async(dispatch_get_main_queue(), ^{
+            self.allContacts = contacts;
+            self.allContactsMap = [allContactsMap copy];
+
+            [self intersectContacts];
+
+            [self updateSignalAccounts];
```
charlesmchen: Yes, but:

a) the contacts intersection is asynchronous and we'd like to update the views immediately.
b) the contacts intersection may fail.

Updating the views should be cheap enough that it's okay to refresh them twice whenever there's a contacts change, especially since contacts changes should be rare.

Signal/src/contact/OWSContactsManager.m
charlesmchen: Great idea, I've added a card.

charlesmchen: Done.

Signal/src/contact/OWSContactsManager.m
charlesmchen: Done.

Signal/src/ViewControllers/SignalAccount.h
charlesmchen: Done.

Signal/src/contact/OWSContactsManager.h
charlesmchen: You're right, they're redundant. `signalAccountMap` is for bulk lookup (e.g. by `ContactsViewHelper`).  `signalAccountForRecipientId:` is redundant but I'm going to leave it in since it has an assert and I think it makes for more readable code.

Signal/src/ViewControllers/SelectRecipientViewController.m
michaelkirk: I don't think we should use "AnyTouch" for "selecting" items (as in from a list) or "pushing buttons". Users know to tap in this case (with it's tiny delay).

One small reason to use a tap is that it comes with a built in "undo" (e.g. slide away from the the thing you accidentally tapped on before lifting your finger). I don't think many people do this consciously, but whenever you go against the grain of something as fundamentally implemented as a standard tap gesture you run the risk of unexpected consequences.

I do like how we're using it to hide the call controls, but not so much because it's fast, rather it's a place where people actually *might* be swiping to try to "dismiss" the controls.

Signal/src/ViewControllers/NewGroupViewController.m
michaelkirk: sounds good!

Signal/translations/bin/auto-genstrings
charlesmchen: Sorry, I didn't mean to commit this. We should definitely revisit this though - we regularly will want to update the strings off master - and we can hit this check when we're not using the SSK development pod - or a different SSK development pod (I tend to keep around two).  

IMO the right check might be:

* in the "push string" script, not the "update string" script.
* Check that no development pod is in use.
* Check that we only push from master.

Signal/translations/en.lproj/Localizable.strings

Signal/translations/en.lproj/Localizable.strings

Signal/translations/en.lproj/Localizable.strings

Signal/translations/en.lproj/Localizable.strings

Signal/src/ViewControllers/UpdateGroupViewController.m
charlesmchen: Done.

charlesmchen:
> "Update" button (top right bar button).

Done.

> replacing the "+" in the new group view with "Create"?

Done.

Signal/src/ViewControllers/SelectRecipientViewController.m
charlesmchen: Done.

PR #2054, opened 2017-04-29
Don't unwrap with ! in CallService.

charlesmchen:
PTAL @michaelkirk


michaelkirk:
@charlesmchen can you rebase this?

PR #2056, opened 2017-05-01
Fix intermittent crash when placing call

michaelkirk:
requestRecordPermission can return on an arbitrary thread.

PTAL @charlesmchen


michaelkirk:
In c6d253d

PR #2057, opened 2017-05-01
Remove invalid assert in case of legitimately stacking unicode

michaelkirk:
This assert was being falsely triggered by legitimate unicode.

The current state of affairs (unchanged by this PR) is that all
diacriticals will be stripped from a message that also contains e.g. 3
flags in a sequence.

PTAL @charlesmchen


michaelkirk:
In 8a2f1e3e

PR #2058, opened 2017-05-02
Charlesmchen/new conversation view

charlesmchen:
Update “new conversation” view to use OWSTableView, contacts view helper, etc.

// FREEBIE


michaelkirk:
@charlesmchen assuming you'll ping me when you want me to review. It's going to be pretty tough to review anything based on #2053 until it's merged unless you can point me at the individual commits.

charlesmchen:
@michaelkirk PTAL

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -3,435 +3,408 @@
...
+    self.noSignalContactsView.hidden = YES;
+    [self.view addSubview:self.noSignalContactsView];
+    [self.noSignalContactsView autoPinWidthToSuperview];
+    [self.noSignalContactsView autoPinEdgeToSuperviewEdge:ALEdgeTop];
+    [self.noSignalContactsView autoPinToBottomLayoutGuideOfViewController:self withInset:0];
+
+    [self updateTableContents];
+}
+
+- (UIView *)createNoSignalContactsView
```
charlesmchen: @michaelkirk I wonder if we should even have the "no contacts" mode of this view any more. To be discussed.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1025,7 +1025,7 @@
 "SEND_INVITE_SUCCESS" = "You've invited your friend to use Signal!";
 
 /* Text for button to send a Signal invite via SMS. %@ is placeholder for the receipient's phone number. */
-"SEND_INVITE_VIA_SMS_BUTTON_FORMAT" = "Send Invite via SMS to: %@";
+"SEND_INVITE_VIA_SMS_BUTTON_FORMAT" = "Invite via SMS: %@";
```
charlesmchen: Much of the phone number was being truncated on small form factors.

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -3,435 +3,408 @@
...
+    subtitleLabel.font = [UIFont ows_regularFontWithSize:ScaleFromIPhone5To7Plus(12.f, 14.f)];
+    subtitleLabel.textAlignment = NSTextAlignmentCenter;
+    subtitleLabel.lineBreakMode = NSLineBreakByWordWrapping;
+    subtitleLabel.numberOfLines = 0;
+    [contents addSubview:subtitleLabel];
+    [subtitleLabel autoPinWidthToSuperview];
+    [subtitleLabel autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:lastSubview withOffset:15];
+    lastSubview = subtitleLabel;
+
+    UIButton *inviteContactsButton = [UIButton buttonWithType:UIButtonTypeCustom];
```
michaelkirk: We need to show an "invite" cell even when the user has signal contacts.

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -3,435 +3,408 @@
...
-        self.searchController.searchBar.hidden = YES;
-        self.tableView.backgroundView = self.noSignalContactsView;
-        self.tableView.backgroundView.opaque   = YES;
-        self.navigationItem.rightBarButtonItem = nil;
-    } else {
-        self.searchController.searchBar.hidden = NO;
-        self.tableView.backgroundView          = nil;
-        self.navigationItem.rightBarButtonItem = self.createGroupBarButtonItem;
-    }
+    // Group Members
```
michaelkirk: what is this comment?

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -3,435 +3,408 @@
...
 
-    [self.tableView reloadData];
-}
+    OWSTableSection *section = [OWSTableSection new];
 
-#pragma mark - Initializers
+    [section addItem:[OWSTableItem itemWithCustomCellBlock:^{
+        UITableViewCell *cell = [UITableViewCell new];
+        cell.textLabel.text = NSLocalizedString(
+            @"NEW_CONVERSATION_FIND_BY_PHONE_NUMBER", @"A label the cell that lets you add a new member to a group.");
```
michaelkirk: weird comment, I guess because this was copy pasted?

Signal/src/AppDelegate.m
```diff
@@ -445,7 +445,7 @@ - (void)application:(UIApplication *)application
     performActionForShortcutItem:(UIApplicationShortcutItem *)shortcutItem
                completionHandler:(void (^)(BOOL succeeded))completionHandler {
     if ([TSAccountManager isRegistered]) {
-        [[Environment getCurrent].signalsViewController composeNew];
+        [[Environment getCurrent].signalsViewController composeNew:nil];
```
charlesmchen: I've removed it.

Signal/src/ViewControllers/MessageComposeTableViewController.m
charlesmchen: Done.

Signal/src/ViewControllers/MessageComposeTableViewController.m
charlesmchen: Removed and I added proper section comments throughout this method.

Signal/src/ViewControllers/MessageComposeTableViewController.m
charlesmchen: Yeah, I'm trying to reuse strings but also avoid having string comments conflict.

charlesmchen:
PTAL @michaelkirk

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -537,275 +519,197 @@ - (void)messageComposeViewController:(MFMessageComposeViewController *)controlle
...
-    OWSAssert([NSThread isMainThread]);
+- (void)searchTextDidChange
+{
+    [self updateSearchPhoneNumbers];
 
-    self.contacts = [self filteredContacts];
-    [self updateSearchResultsForSearchController:self.searchController];
-    [self.tableView reloadData];
-    // TODO revisit this after https://github.com/WhisperSystems/Signal-iOS/pull/2058 is merged
-    [self showEmptyBackgroundViewIfNecessary];
```
charlesmchen: @michaelkirk I think you added `showEmptyBackgroundViewIfNecessary` recently - I've removed it.  Is it clear why?

Signal/src/Storyboard/Main.storyboard
```diff
@@ -976,36 +976,8 @@
                             </connections>
                         </barButtonItem>
                     </navigationItem>
-                    <connections>
-                        <outlet property="inviteCell" destination="Pgm-Qb-oPu" id="Nf5-sS-IKL"/>
-                        <outlet property="noSignalContactsView" destination="M2F-Tn-Zvq" id="jOQ-hN-f9v"/>
-                    </connections>
                 </tableViewController>
                 <placeholder placeholderIdentifier="IBFirstResponder" id="fUD-iU-Cax" userLabel="First Responder" sceneMemberID="firstResponder"/>
-                <tableViewCell contentMode="scaleToFill" selectionStyle="default" accessoryType="disclosureIndicator" indentationWidth="10" reuseIdentifier="ContactTableInviteCell" textLabel="QGl-1D-W9D" rowHeight="48" style="IBUITableViewCellStyleDefault" id="Pgm-Qb-oPu" userLabel="Invite Cell">
```
michaelkirk: There's now a bar line separating the search bar from the table - was that intentional?

Signal/src/ViewControllers/MessageComposeTableViewController.m
michaelkirk: I think the existing case is a little better flow for the users who have no contacts on Signal. What's the argument for getting rid of it?

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -168,275 +194,229 @@ - (void)viewWillAppear:(BOOL)animated
...
-- (void)setIsNoContactsViewVisible:(BOOL)isNoContactsViewVisible {
-    if (isNoContactsViewVisible == _isNoContactsViewVisible) {
-        return;
+    __weak MessageComposeTableViewController *weakSelf = self;
+    ContactsViewHelper *helper = self.contactsViewHelper;
+
+    OWSTableSection *section = [OWSTableSection new];
+
+    // Find Non-Contacts by Phone Number
+    [section addItem:[OWSTableItem itemWithCustomCellBlock:^{
```
michaelkirk: This is somewhat redundant with the existing search bar, but the "find by phone number" is nice in that it's an easier way to key in the numbers (number pad vs. keyboard) and an easier way to find someone in a different country code.

However, I like the existing search bars better "autocomplete"-like behavior with "Start conversation with #" and the "invite # via sms". The "Find by phone number view" doesn't give any feedback until you "submit".

Also, it's taking up some real prime real estate, especially on smaller devices:

search_invite

Another thing, when I type in part of a phone number, e.g. "555-123-12", "Find by phone number" almost seems like a button to execute the search for accounts based on my incomplete search input.

Weighing those things, I'm not sure we should keep it. WDYT @charlesmchen?

In either case, on a separate design note, I believe the top two items probably *are* the same height as the rest of the table cells, but without the avatars, the top two items feel a little too big.  Do you think it would be unreasonable to make them shorter than the other items in the table?

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -168,275 +194,229 @@ - (void)viewWillAppear:(BOOL)animated
...
+    for (NSString *phoneNumber in searchPhoneNumbers) {
+        OWSAssert(phoneNumber.length > 0);
 
-    [self.tableView reloadData];
-}
+        if ([self.nonContactAccountSet containsObject:phoneNumber]) {
+            [section addItem:[OWSTableItem itemWithCustomCellBlock:^{
+                MessageComposeTableViewController *strongSelf = weakSelf;
+                if (!strongSelf) {
+                    return (ContactTableViewCell *)nil;
```
michaelkirk: nit: This block is typed to not return nil

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -537,275 +519,197 @@ - (void)messageComposeViewController:(MFMessageComposeViewController *)controlle
...
 }
 
-- (BOOL)hasNoContacts
+- (void)newConversationWith:(NSString *)recipientId
 {
-    return self.contacts.count == 0;
-}
+    OWSAssert(recipientId.length > 0);
 
-- (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger)section {
```
michaelkirk: Happy to see this go 🔥 🔥 🔥

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -551,13 +551,13 @@
 "HOURS_TIME_AMOUNT" = "%u hours";
 
 /* Call setup status label */
-"IN_CALL_CONNECTING" = "Connecting...";
+"IN_CALL_CONNECTING" = "Connecting…";
 
 /* Call setup status label */
-"IN_CALL_RINGING" = "Ringing...";
+"IN_CALL_RINGING" = "Ringing…";
```

Signal/src/ViewControllers/MessageComposeTableViewController.m
michaelkirk: Yup, seems like updateTableContents takes care of the issue this was addressing. 👍

Signal/src/ViewControllers/MessageComposeTableViewController.m
charlesmchen: NBD. Just reducing code complexity a bit, and that it's somewhat redundant with the offers to invite contacts already in the view.

Signal/src/Storyboard/Main.storyboard
charlesmchen: I like the visual separation, but I'll go ahead and remove it.

Signal/src/ViewControllers/MessageComposeTableViewController.m
charlesmchen: As you pointed out, the main reason I included "find by phone number" is that it provides a much better experience for finding people in other calling codes.  Very few people are comfortable with e164 formatting / dealing with country codes.  The "find by phone number" view has controls that let surface this, let you specify the calling code by country name, etc. - and it also takes care of the concatenation process.  

I also like that its presence emphasizes that you can communicate with non-contacts - the gaze tends to gloss over placeholder text.

I do think we should leave the ability to search for phone numbers in the search bar.

I'm not worried about the redundancy - offering more than one way to do things seems fine to me.  

> Also, it's taking up some real prime real estate, especially on smaller devices:

Yes, that's an issue.

> Another thing, when I type in part of a phone number, e.g. "555-123-12", "Find by phone number" almost seems like a button to execute the search for accounts based on my incomplete search input.

That's true, but if anyone made that mistake I think they'd figure it out pretty quickly, close the "find by phone number" view and not have lost their search.

> Weighing those things, I'm not sure we should keep it. WDYT @charlesmchen?

It's not perfect, but I think it's the lesser of two evils to keep it in.  The "affordance" of being able to search for non-contacts by phone number isn't clear enough without it and finding users in other calling codes is currently a bad experience.

Let's discuss on a call!

> In either case, on a separate design note, I believe the top two items probably are the same height as the rest of the table cells, but without the avatars, the top two items feel a little too big. Do you think it would be unreasonable to make them shorter than the other items in the table?

I had it that way originally and it felt broken to have them the default size.  Maybe there's a happy medium in between the two cell heights.

Signal/src/ViewControllers/MessageComposeTableViewController.m
charlesmchen: Done.

charlesmchen:
Per our convo I think this is ready for re-review. PTAL @michaelkirk

PR #2064, opened 2017-05-03
delay contact access

michaelkirk:
Most commonly this will be after hitting the "compose" button

But also we'll do it in the SignalViewController once you've received a
message.

- get rid blocking contacts nag
- use Contacts framework simplifies logic
- remove dead AB code

PTAL @charlesmchen

### DIFF
--- a/Signal.xcodeproj/project.pbxproj
+++ b/Signal.xcodeproj/project.pbxproj
@@ -105,6 +105,7 @@
 		453D28B71D32BA5F00D523F0 /* OWSDisplayedMessage.m in Sources */ = {isa = PBXBuildFile; fileRef = 453D28B61D32BA5F00D523F0 /* OWSDisplayedMessage.m */; };
 		453D28BA1D332DB100D523F0 /* OWSMessagesBubblesSizeCalculator.m in Sources */ = {isa = PBXBuildFile; fileRef = 453D28B91D332DB100D523F0 /* OWSMessagesBubblesSizeCalculator.m */; };
 		453D28BB1D332DB100D523F0 /* OWSMessagesBubblesSizeCalculator.m in Sources */ = {isa = PBXBuildFile; fileRef = 453D28B91D332DB100D523F0 /* OWSMessagesBubblesSizeCalculator.m */; };
+		4542F0941EB9372700C7EE92 /* SystemContactsFetcher.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4542F0931EB9372700C7EE92 /* SystemContactsFetcher.swift */; };
 		45464DBC1DFA041F001D3FD6 /* DataChannelMessage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45464DBB1DFA041F001D3FD6 /* DataChannelMessage.swift */; };
 		45666EC61D99483D008FE134 /* OWSAvatarBuilder.m in Sources */ = {isa = PBXBuildFile; fileRef = 45666EC51D99483D008FE134 /* OWSAvatarBuilder.m */; };
 		45666EC91D994C0D008FE134 /* OWSGroupAvatarBuilder.m in Sources */ = {isa = PBXBuildFile; fileRef = 45666EC81D994C0D008FE134 /* OWSGroupAvatarBuilder.m */; };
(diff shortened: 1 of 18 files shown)


Signal.xcodeproj/project.pbxproj
```diff
@@ -2449,7 +2453,7 @@
 					"\"$(SRCROOT)/Libraries\"/**",
 				);
 				INFOPLIST_FILE = "$(SRCROOT)/Signal/Signal-Info.plist";
-				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
```
charlesmchen: Shall we also bump:

* https://github.com/WhisperSystems/SignalServiceKit/blob/master/SignalServiceKit.podspec#L24
* https://github.com/WhisperSystems/SignalProtocolKit/blob/master/Podfile#L1

Strangely, this one is already bumped!

* https://github.com/WhisperSystems/Signal-iOS/blob/master/Podfile#L1

charlesmchen: Huh, I forgot that I included some of this in: https://github.com/WhisperSystems/Signal-iOS/commit/42768294eebda42c1d2f060e0f5ff70770645d0b

Signal/src/ViewControllers/SignalsViewController.m
```diff
@@ -683,6 +651,12 @@ - (void)yapDatabaseModified:(NSNotification *)notification {
     NSArray *sectionChanges = nil;
     NSArray *rowChanges     = nil;
 
+    // If the user hasn't already granted contact access
+    // we don't want to request until they receive a message.
+    if ([TSThread numberOfKeysInCollection] > 0) {
+        [self.contactsManager requestSystemContactsOnce];
+    }
```
charlesmchen: Why is this a trigger to show the contacts permission request?

Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -0,0 +1,120 @@
...
+            assertionFailure("should have called `requestOnce` before this point.")
+            Logger.error("\(TAG) should have called `requestOnce` before checking authorization status.")
+            return false
+        }
+
+        return self.authorizationStatus == .authorized
+    }
+
+    private let contactStore = CNContactStore()
+    private var systemContactsHaveBeenRequestedAtLeastOnce = false
```
charlesmchen: So we nag them once per launch of the app.

Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -0,0 +1,120 @@
...
+        guard authorizationStatus == .authorized else {
+            return
+        }
+
+        updateContacts()
+    }
+
+    private func displayMissingContactsPermissionAlert() {
+        let foo = UIApplication.shared.frontmostViewController
+        Logger.error("TODO")
```
charlesmchen: You might find `[ViewControllerUtils showAlertWithTitle:...]` useful here.

Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -0,0 +1,120 @@
...
+
+    public func requestOnce() {
+        AssertIsOnMainThread()
+
+        guard !systemContactsHaveBeenRequestedAtLeastOnce else {
+            Logger.debug("\(TAG) already requested system contacts")
+            return
+        }
+        systemContactsHaveBeenRequestedAtLeastOnce = true
+        self.startObservingContactChanges()
```
charlesmchen: Why not start observing contact changes immediately?  Will iOS send `CNContactStoreDidChange` notifications if we don't have contacts permission?

Signal/translations/bin/auto-genstrings
```diff
@@ -8,8 +8,13 @@ pushd $SSK_DIR
...
 then
-  echo "[!] Error - SSK must be on master to be sure we're generating up-to-date strings"
-  exit 1
+  if [[ $* == *--non-master* ]]
+  then
+    echo "[!] Note - generating from non-master SSK."
+  else
+    echo "[!] Error - SSK must be on master to be sure we're generating up-to-date strings, or use '--non-master'."
+    exit 1
+  fi
```
charlesmchen: thank you!

Signal/src/contact/SystemContactsFetcher.swift
michaelkirk: nope, it's just a debounce. 

Disregarding updates, we only want to pull in the address book one time. However this could happen at various points in the app. So at each of these checkpoints, we fetch the address book only if it hasn't been previously fetched.

(And then we also start fetching updates at this point with the observer notification)

Signal/src/contact/SystemContactsFetcher.swift
michaelkirk: > Will iOS send CNContactStoreDidChange notifications if we don't have contacts permission?

I don't know, but since the answer to that question isn't obvious, and since we *know* there aren't any meaningful things to observe until we've "requested" contacts permission, this seemed more straight forward.

Signal/src/contact/SystemContactsFetcher.swift
michaelkirk: This is kind of a design TODO. I should have documented my thoughts rather than leaving this TODO. I don't think we actually want to risk enraging hitting the user with a "shame on you" alert immediately after they say "no" to the previous alert. 

I think we will be better served by the contextually relevant reminders (e.g. in the contact list, something like "Your contact list is empty because you haven't enabled contacts access. [Settings]")

WDYT?

Signal.xcodeproj/project.pbxproj
charlesmchen: That's fine.

Signal/src/ViewControllers/SignalsViewController.m
charlesmchen: Gotcha.

Signal/src/contact/SystemContactsFetcher.swift
charlesmchen: Absolutely agree. Probably anywhere we show the "SETTINGS_BLOCK_LIST_NO_CONTACTS" cell we could show another cell "load contacts" cell if they haven't given permission yet.

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/a794fe83588d2cab7ca328c4c51717e302e3b94a

PR #2070, opened 2017-05-05
Charlesmchen/voice memos

charlesmchen:
There's more we could do here:

* We could have a better "you cancelled" animation.
* We could have more help UI.

I'm not sure either is necessary. I'd like to discuss what I've got so far.

PTAL @michaelkirk 

voice memo during

voice memo at rest


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -421,6 +421,9 @@
 /* Generic error used whenver Signal can't contact the server */
 "ERROR_DESCRIPTION_NO_INTERNET" = "Signal was unable to connect to the internet. Please try from another WiFi network or use mobile data.";
 
+/* Error indicating that an outgoing message had no valid recipients. */
+"ERROR_DESCRIPTION_NO_VALID_RECIPIENTS" = "ERROR_DESCRIPTION_NO_VALID_RECIPIENTS";
```
charlesmchen: What happened to this string?

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -3037,6 +3441,61 @@ - (void)showErrorAlertForAttachment:(SignalAttachment * _Nullable)attachment {
...
+- (void)cancelVoiceMemo
+{
+    OWSAssert([NSThread isMainThread]);
+
+    [((OWSMessagesToolbarContentView *)self.inputToolbar.contentView)cancelVoiceMemoIfNecessary];
+    [((OWSMessagesInputToolbar *)self.inputToolbar) hideVoiceMemoUI:NO];
+    [self cancelRecordingVoiceMemo];
+}
+
+- (void)textViewDidChange:(UITextView *)textView
```
charlesmchen: This overrides a JSQ method.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2704,48 +3032,124 @@ - (TSInteraction *)interactionAtIndexPath:(NSIndexPath *)indexPath {
...
-            [self showErrorAlertForAttachment:attachment];
-        } else {
-            [self tryToSendAttachmentIfApproved:attachment];
-        }
+    self.audioRecorder = [[AVAudioRecorder alloc] initWithURL:fileURL
+                                                     settings:@{
+                                                         AVFormatIDKey : @(kAudioFormatMPEG4AAC),
+                                                         AVSampleRateKey : @(44100),
+                                                         AVNumberOfChannelsKey : @(2),
+                                                         AVEncoderBitRateKey: @(128 * 1024),
```
charlesmchen: Take a look at these settings. I think they should be fine.  I considered mono but I've heard that stereo matters in these conditions, especially for the deaf.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2704,48 +3032,124 @@ - (TSInteraction *)interactionAtIndexPath:(NSIndexPath *)indexPath {
...
-                         nil];
-    NSURL *outputFileURL = [NSURL fileURLWithPathComponents:pathComponents];
+- (void)startRecordingVoiceMemo
+{
+    OWSAssert([NSThread isMainThread]);
+
+    DDLogInfo(@"startRecordingVoiceMemo");
+
+    NSString *temporaryDirectory = NSTemporaryDirectory();
+    NSString *filename = [NSString stringWithFormat:@"%lld.m4a", [NSDate ows_millisecondTimeStamp]];
```
charlesmchen: Received voice memos will end up with a .mp4 file extension.

Signal/translations/en.lproj/Localizable.strings
michaelkirk: https://github.com/WhisperSystems/Signal-iOS/commit/800715a5e0246c97d2cc51c0273b688dbec7f26d

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -170,12 +349,144 @@ - (JSQMessagesToolbarContentView *)loadToolbarContentView {
...
+    [self.voiceMemoUI addSubview:cancelLabel];
+
+    [imageView autoVCenterInSuperview];
+    [imageView autoPinEdgeToSuperviewEdge:ALEdgeLeft withInset:10];
+    [self.recordingLabel autoVCenterInSuperview];
+    [self.recordingLabel autoPinEdge:ALEdgeLeft toEdge:ALEdgeRight ofView:imageView withOffset:5.f];
+    [cancelLabel autoVCenterInSuperview];
+    [cancelLabel autoHCenterInSuperview];
+    [self.voiceMemoUI setNeedsLayout];
+    [self.voiceMemoUI layoutSubviews];
```
michaelkirk: Why do we need to layout subviews *now* if we've setNeedsLayout?

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -170,12 +349,144 @@ - (JSQMessagesToolbarContentView *)loadToolbarContentView {
...
+                                                                   selector:@selector(updateVoiceMemo)
+                                                                   userInfo:nil
+                                                                    repeats:YES];
+}
+
+- (void)hideVoiceMemoUI:(BOOL)animated
+{
+    OWSAssert([NSThread isMainThread]);
+
+    UIView *voiceMemoUI = self.voiceMemoUI;
```
michaelkirk: nit: maybe `oldVoiceMemoUI` would be a better name, e.g. to avoid the bug on line 446 where we're removing animations from `nil` :wink:

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2704,48 +3032,125 @@ - (TSInteraction *)interactionAtIndexPath:(NSIndexPath *)indexPath {
...
+        return;
+    }
+
+    NSTimeInterval currentTime = self.audioRecorder.currentTime;
+
+    [self.audioRecorder stop];
+
+    const NSTimeInterval kMinimumRecordingTimeSeconds = 1.f;
+    if (currentTime < kMinimumRecordingTimeSeconds) {
+        DDLogInfo(@"Discarding voice memo; too short.");
```
michaelkirk: Maybe we should prompt the user here, incase they don't understand they have to "hold" to record.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2704,48 +3032,125 @@ - (TSInteraction *)interactionAtIndexPath:(NSIndexPath *)indexPath {
...
 
-- (void)recordAudio {
-    // Define the recorder setting
-    NSArray *pathComponents = [NSArray
-        arrayWithObjects:[NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES) lastObject],
-                         [NSString stringWithFormat:@"%lld.m4a", [NSDate ows_millisecondTimeStamp]],
-                         nil];
-    NSURL *outputFileURL = [NSURL fileURLWithPathComponents:pathComponents];
+- (void)startRecordingVoiceMemo
+{
```
michaelkirk: to test: what happens w/o microphone permissions?

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -2704,48 +3032,125 @@ - (TSInteraction *)interactionAtIndexPath:(NSIndexPath *)indexPath {
...
+        DDLogError(@"%@ Couldn't load audioRecorder data", self.tag);
+        OWSAssert(0);
+        self.audioRecorder = nil;
+        return;
+    }
+
+    self.audioRecorder = nil;
+
+    SignalAttachment *attachment =
+        [SignalAttachment attachmentWithData:audioData dataUTI:(NSString *)kUTTypeMPEG4Audio filename:nil];
```
michaelkirk: Take it or leave it: Until we have a specific "voice note" interface, maybe we ought to have a filename like "Voice Note.m4a" or "Voice Note-[date-stamp].m4a" so that recipients can distinguish it from other audio attachments.

Signal/src/ViewControllers/MessagesViewController.m
michaelkirk: nit: could you mark it as an override with a comment?

michaelkirk:
Also noticed that when "pasting" a "copied" voice note it sends as a .bin

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1273,6 +1273,9 @@
 /* table cell label in conversation settings */
 "VERIFY_PRIVACY" = "Verify Safety Number";
 
+/* Indicates how to cancel a voice memo. */
+"VOICE_MEMO_CANCEL_INSTRUCTIONS" = "Slide To Cancel";
```
RiseT: If you want correct title case here, it's "Tap to Cancel".

Signal/translations/en.lproj/Localizable.strings
charlesmchen: Thanks this was a note to myself that I hadn't followed up on yet.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -130,21 +133,54 @@ - (BOOL)canPerformAction:(SEL)action withSender:(id)sender {
...
+
+- (void)voiceMemoGestureDidChange:(CGFloat)cancelAlpha;
+
+@end
+
+#pragma mark -
+
+@interface OWSMessagesToolbarContentView ()
+
+@property (nonatomic, weak) id<OWSMessagesToolbarContentDelegate> delegate;
```

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -130,21 +133,54 @@ - (BOOL)canPerformAction:(SEL)action withSender:(id)sender {
...
+
+@property (nonatomic) BOOL shouldShowVoiceMemoButton;
+
+@property (nonatomic) UIButton *voiceMemoButton;
```

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -130,21 +133,54 @@ - (BOOL)canPerformAction:(SEL)action withSender:(id)sender {
...
+
+@property (nonatomic) UIButton *sendButton;
```
charlesmchen: In practice, the three properties won't be nil before they're accessed, but since they're lazy populated, I'm adding nullable.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -130,21 +133,54 @@ - (BOOL)canPerformAction:(SEL)action withSender:(id)sender {
...
+
+@property (nonatomic) BOOL isRecordingVoiceMemo;
+
+@property (nonatomic) CGPoint voiceMemoGestureStartLocation;
```
charlesmchen: CGPoint is not a nullable type.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -155,12 +191,155 @@ + (UINib *)nib
...
+    }
+
+    _shouldShowVoiceMemoButton = shouldShowVoiceMemoButton;
+
+    [self ensureVoiceMemoButton];
+}
+
+- (void)ensureVoiceMemoButton
+{
+    if (!self.superview) {
```
charlesmchen: This is no longer necessary, removed.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -155,12 +191,155 @@ + (UINib *)nib
...
+        return;
+    }
+
+    if (!self.sendButton) {
+        OWSAssert(self.rightBarButtonItem);
+
+        self.sendButton = self.rightBarButtonItem;
+    }
+
+    if (self.shouldShowVoiceMemoButton) {
```

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -155,12 +191,155 @@ + (UINib *)nib
...
+
 @end
 
 #pragma mark -
 
 @implementation OWSMessagesInputToolbar
 
+- (void)toggleSendButtonEnabled
+{
+    // Do nothing; disables JSQ's control over send button enabling.
```

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: Because we use the "at rest" layout below when setting up the animations.

Signal/src/ViewControllers/MessagesViewController.m

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: Hadn't tested that yet. I've added logic to handle this.

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: Yes, this is the crux of what I want to discuss with you and Riya before going any further. What "crutches"/"hints" and other polish we want to add around voice memoes, if any.

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: Great idea. The former looks better. I'm not concerned about unique filenames - users will rarely save/export/reshare these.

Signal/src/ViewControllers/MessagesViewController.m

charlesmchen:
> Also noticed that when "pasting" a "copied" voice note it sends as a .bin

I've completely reworked how we copy attachments to the clipboard.  There were a number of serious problems.

Signal/translations/en.lproj/Localizable.strings
charlesmchen: Thanks @RiseT

Signal/src/Models/TSMessageAdapaters/TSAnimatedAdapter.m
```diff
@@ -131,8 +132,14 @@ - (BOOL)canPerformEditingAction:(SEL)action
...
     if (action == @selector(copy:)) {
+        NSString *utiType = [MIMETypeUtil utiTypeForMIMEType:self.attachment.contentType];
+        if (!utiType) {
+            OWSAssert(0);
+            utiType = (NSString *)kUTTypeVideo;
+        }
+
         UIPasteboard *pasteboard = UIPasteboard.generalPasteboard;
-        [pasteboard setData:self.fileData forPasteboardType:(__bridge NSString *)kUTTypeGIF];
+        [pasteboard setData:self.fileData forPasteboardType:utiType];
```
charlesmchen: We don't want to assume animated images are GIFs.

Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -297,9 +297,13 @@ - (void)performEditingAction:(SEL)action
 {
     if ([self isVideo]) {
         if (action == @selector(copy:)) {
+            NSString *utiType = [MIMETypeUtil utiTypeForMIMEType:_contentType];
+            if (!utiType) {
+                OWSAssert(0);
+                utiType = (NSString *)kUTTypeVideo;
+            }
             NSData *data = [NSData dataWithContentsOfURL:self.fileURL];
-            // TODO: This assumes all videos are mp4.
```
charlesmchen: This comment is right.

Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -310,22 +314,30 @@ - (void)performEditingAction:(SEL)action
...
-            } else if ([pasteboardType isEqualToString:@"aiff"]) {
-                [UIPasteboard.generalPasteboard setData:data
-                                      forPasteboardType:(NSString *)kUTTypeAudioInterchangeFileFormat];
-            } else if ([pasteboardType isEqualToString:@"m4a"]) {
-                [UIPasteboard.generalPasteboard setData:data forPasteboardType:(NSString *)kUTTypeMPEG4Audio];
-            } else if ([pasteboardType isEqualToString:@"amr"]) {
-                [UIPasteboard.generalPasteboard setData:data forPasteboardType:@"org.3gpp.adaptive-multi-rate-audio"];
-            } else {
-                [UIPasteboard.generalPasteboard setData:data forPasteboardType:(NSString *)kUTTypeAudio];
+            NSString *utiType = [MIMETypeUtil utiTypeForMIMEType:_contentType];
```
charlesmchen: This should handle most cases.

Signal/src/Models/TSMessageAdapaters/TSAnimatedAdapter.m
```diff
@@ -131,8 +132,14 @@ - (BOOL)canPerformEditingAction:(SEL)action
 - (void)performEditingAction:(SEL)action
 {
     if (action == @selector(copy:)) {
+        NSString *utiType = [MIMETypeUtil utiTypeForMIMEType:self.attachment.contentType];
+        if (!utiType) {
+            OWSAssert(0);
+            utiType = (NSString *)kUTTypeVideo;
```
michaelkirk: Don't we want to default to kUTTypeGIF for animated adapters?

Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -310,22 +314,30 @@ - (void)performEditingAction:(SEL)action
...
+            if (!utiType) {
```
michaelkirk: This logic feels like it belongs in MIMETypeUtil

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -3074,17 +3072,28 @@ - (void)startRecordingVoiceMemo
...
-        [self resetRecordingVoiceMemo];
+        [self cancelVoiceMemo];
         OWSAssert(0);
         return;
     }
+
+    if (session.recordPermission != AVAudioSessionRecordPermissionGranted) {
+        DDLogError(@"%@ we do not have recording permission.", self.tag);
+        [self cancelVoiceMemo];
+        [ViewControllerUtils
```
michaelkirk: I think we want something more actionable like this: https://github.com/WhisperSystems/Signal-iOS/blob/master/Signal/src/call/OutboundCallInitiator.swift#L52

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -3121,8 +3130,11 @@ - (void)endRecordingVoiceMemo
 
     self.audioRecorder = nil;
 
+    NSString *filename = [NSLocalizedString(@"VOICE_MEMO_FILE_NAME", @"Filename for voice memoes.")
```
michaelkirk: comment typo: memos

charlesmchen:
> It's kind of hard to trigger the gesture. I think it's a combination of two things -

> maybe the hit zone is too small?

I've been thinking the same thing.  I've modified how the long press gesture works to make this easier.

https://github.com/WhisperSystems/Signal-iOS/pull/2070/commits/048d3ec6b841300c83cc81fd19062eb4c7d33edd

> I think longpress is too long to activate a button-like interaction.

Done.

> It feels weird not knowing how long an audio file is before we play it. I think the duration is more important than the file size if we can't have both.

I've made a card.

> The duration label wiggles as it progresses - I think we should use a fixed width font.

I've tried a couple of fixed width fonts and they just don't look right in this context.  I've made a card if we want to pursue this.

Signal/src/ViewControllers/MessagesViewController.m

Signal/src/Models/TSMessageAdapaters/TSAnimatedAdapter.m
charlesmchen: Good catch, that's a mistake. But I'm not sure what makes sense as a default here since GIF will always be handled correctly by `[MIMETypeUtil utiTypeForMIMEType:]` and so it'll be wrong if that fails.

Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
charlesmchen: Agreed. Actually, it'd be nice to have quite comprehensive "file extension -> mimetype" and "mimetype -> uti type" mappings as we do now for: `[MIMETypeUtil  genericMIMETypesToExtensionTypes:]`.  I'll file a card.

RiseT:
Don't know if you deem this important, but in Signal Android and other apps (e. g. WA), it's "voice message" rather than "voice memo". When I search for the latter term on the web, it rather seems to get used in the sense of a dictation machine/voice recorder, i. e. voice recording primarily for oneself.

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: Good idea.

I've made a new class where we can hang reusable alerts.

 https://github.com/WhisperSystems/Signal-iOS/pull/2070/commits/11fa9d84f7320d2b55c32f1acded1f766ba6a12a

charlesmchen:
@RiseT 

Thanks; I've renamed these "voice messages."

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: I've added an alert as discussed.

PR #2071, opened 2017-05-05
Show “invite by SMS” offer for matching non-Signal contacts when sear…

charlesmchen:
…ching in “new 1:1: conversation” view.

PTAL @michaelkirk 

img_0281

### DIFF
--- a/Signal/src/ViewControllers/ContactsViewHelper.m
+++ b/Signal/src/ViewControllers/ContactsViewHelper.m
@@ -15,6 +15,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface ContactsViewHelper ()
 
+// This property is a cached value that is lazy-populated.
+@property (nonatomic, nullable) NSArray<Contact *> *nonSignalContacts;
+
 @property (nonatomic) NSDictionary<NSString *, SignalAccount *> *signalAccountMap;
 @property (nonatomic) NSArray<SignalAccount *> *signalAccounts;
 
@@ -143,6 +146,7 @@ NS_ASSUME_NONNULL_BEGIN
     }
     self.signalAccountMap = [signalAccountMap copy];
     self.signalAccounts = [signalAccounts copy];
+    self.nonSignalContacts = nil;
 
     [self.delegate contactsViewHelperDidUpdateContacts];
 }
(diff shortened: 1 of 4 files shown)


charlesmchen:
@moxie0 This was a neat idea from @michaelkirk.  You might want to do something similar on Android.

charlesmchen:
Also including: Slightly reduce the non-contact cell heights in “new 1:1 conversation” view.

img_0032

Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -199,6 +207,76 @@ - (BOOL)doesSignalAccount:(SignalAccount *)signalAccount matchSearchTerms:(NSArr
...
+
+    return NO;
+}
+
+- (BOOL)doesContact:(Contact *)contact matchSearchTerms:(NSArray<NSString *> *)searchTerms
+{
+    OWSAssert(contact);
+    OWSAssert(searchTerms.count > 0);
+
+    for (NSString *searchTerm in searchTerms) {
```
michaelkirk: why does this code feel so familiar?

Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -199,6 +207,76 @@ - (BOOL)doesSignalAccount:(SignalAccount *)signalAccount matchSearchTerms:(NSArr
...
+        return [NSArray new];
+    }
+
+    return [self.nonSignalContacts filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(Contact *contact,
+                                                                   NSDictionary<NSString *, id> *_Nullable bindings) {
+        return [self doesContact:contact matchSearchTerms:searchTerms];
+    }]];
+}
+
+- (nullable NSArray<Contact *> *)nonSignalContacts
```
michaelkirk: this might be an empty array, but it won't be nil, right?

charlesmchen:
> when you hit cancel the searchfield is empty, but the list is still filtered

I can't repro this issue. I've tested iOS 9 and 10.  Can you give me any more details?

> I'm sure this was a pre-existing issue, but SMS view is illegible . We need to "apply default appearance" like we do when messaging someone via the "Invite Friends to Signal" flow (then revert to signal appearance upon completion). You could also try using the InviteFlow.sendSmsTo(phoneNumbers:[String]) which should take care of this for you.

I've filed a card.

Signal/src/ViewControllers/ContactsViewHelper.m
charlesmchen: We do something similar when searching for `SignalContacts`.  There's just enough differences that it doesn't make sense to de-duplicate, IMO.

Signal/src/ViewControllers/ContactsViewHelper.m
charlesmchen: Well, this value is cached and we clear the cache when contacts change by setting the cached value to nil.

charlesmchen:
PTAL @michaelkirk

charlesmchen:
> when you hit cancel the searchfield is empty, but the list is still filtered

I'm going to merge, but can you please re-test this at some point?  I can't repro.  I'll make sure it's in the test plan too.

PR #2072, opened 2017-05-05
Charlesmchen/contacts sync

charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/202

PTAL @michaelkirk


Signal/src/util/OWSContactsSyncing.m
```diff
@@ -119,7 +119,7 @@ - (void)sendSyncContactsMessageIfNecessary
 - (void)sendSyncContactsMessageIfPossible
 {
     AssertIsOnMainThread();
-    if (self.contactsManager.signalContacts.count == 0) {
+    if (self.contactsManager.signalAccounts.count == 0) {
```
michaelkirk: oh, I should have changed this. Thanks for catching it.

PR #2073, opened 2017-05-05
no contacts banner -> system settings

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/views/NotificationView.swift
+++ b/Signal/src/views/NotificationView.swift
@@ -1,0 +1,92 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+class NotificationView: UIView {
+
+    let TAG = "[NotificationView]"
+    let label = UILabel()
+
+    let defaultTapAction = {
+        Logger.debug("[NotificationView] tapped.")
+    }
+
+    var tapAction: () -> Void
+
+    var text: String? {
+        get {
+            return label.text
+        }
+
+        set(newText) {
+            label.text = newText
+        }
+    }
+
+    required init?(coder: NSCoder) {
+        self.tapAction = defaultTapAction
+
+        super.init(coder: coder)
+
+        setupSubviews()
+    }
+
+    override init(frame: CGRect) {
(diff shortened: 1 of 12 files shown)


Signal/src/views/NotificationView.swift
```diff
@@ -0,0 +1,92 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+class NotificationView: UIView {
```
charlesmchen: I'm wondering if there's a better name for this - `NotificationView` is probably fine but makes me think of NSNotifications, local notifications, remote push notifications.  Maybe: `WarningView`, `NagView`, etc., etc.

charlesmchen: `ReminderView`,

Signal/src/views/NotificationView.swift
```diff
@@ -0,0 +1,92 @@
...
+        setupSubviews()
+    }
+
+    convenience init(tapAction: @escaping () -> Void) {
+        self.init(frame: .zero)
+        self.tapAction = tapAction
+    }
+
+    func setupSubviews() {
+        self.backgroundColor = UIColor.ows_infoMessageBorder().withAlphaComponent(0.20)
```
charlesmchen: `ows_infoMessageBackground()`?

Signal/src/views/NotificationView.swift
charlesmchen: `InfoMessageView`

michaelkirk: I hear this is one of those "hard" problems. I've settled on "ReminderView" since InfoMessage might get conflated with the "Info Messages" that occur in the MessagesView.

Signal/src/views/NotificationView.swift
michaelkirk: I've added a new color "reminderYellow"

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/be504a7e46e6e6a65b587d304f74d0ac20820455

PR #2075, opened 2017-05-05
Fix assert in conversation settings view.

charlesmchen:
The problem with the `commonInit` pattern is that if a superclass and subclass both use the pattern, the subclass's implementation won't know to call superclass's implementation since it's a private implementation detail.

PTAL @michaelkirk


michaelkirk:
Damn dynamic dispatch. LGTM

PR #2078, opened 2017-05-05
compose w/o contacts

michaelkirk:
screen shot 2017-05-05 at 6 22 22 pm

As well as adding the banner, removed the irrelevant bits of the interface (e.g. search bar an Invite) which don't make sense without contact access.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MessageComposeTableViewController.m
+++ b/Signal/src/ViewControllers/MessageComposeTableViewController.m
@@ -34,6 +34,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) OWSTableViewController *tableViewController;
 
 @property (nonatomic, readonly) UISearchBar *searchBar;
+@property (nonatomic, readonly) NSLayoutConstraint *hideContactsPermissionReminderViewConstraint;
 
 // A list of possible phone numbers parsed from the search text as
 // E164 values.
@@ -58,6 +59,18 @@ NS_ASSUME_NONNULL_BEGIN
     _contactsViewHelper = [ContactsViewHelper new];
     _contactsViewHelper.delegate = self;
     _nonContactAccountSet = [NSMutableSet set];
+
+    ReminderView *contactsPermissionReminderView = [[ReminderView alloc]
+        initWithText:NSLocalizedString(@"COMPOSE_SCREEN_MISSING_CONTACTS_PERMISSION",
+                         @"Multiline label explaining why compose-screen contact picker is empty.")
+           tapAction:^{
+               [[UIApplication sharedApplication] openSystemSettings];
+           }];
+    [self.view addSubview:contactsPermissionReminderView];
+    [contactsPermissionReminderView autoPinWidthToSuperview];
+    [contactsPermissionReminderView autoPinEdgeToSuperviewMargin:ALEdgeTop];
+    _hideContactsPermissionReminderViewConstraint =
+        [contactsPermissionReminderView autoSetDimension:ALDimensionHeight toSize:0];
 
     self.navigationItem.leftBarButtonItem =
         [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemStop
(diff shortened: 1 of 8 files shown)


michaelkirk:
rebased and more easily reviewable now. PTAL @charlesmchen

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -59,6 +60,18 @@ - (void)loadView
...
+        initWithText:NSLocalizedString(@"COMPOSE_SCREEN_MISSING_CONTACTS_PERMISSION",
+                         @"Multiline label explaining why compose-screen contact picker is empty.")
+           tapAction:^{
+               [[UIApplication sharedApplication] openSystemSettings];
+           }];
+    [self.view addSubview:contactsPermissionReminderView];
+    [contactsPermissionReminderView autoPinWidthToSuperview];
+    [contactsPermissionReminderView autoPinEdgeToSuperviewMargin:ALEdgeTop];
+    _hideContactsPermissionReminderViewConstraint =
+        [contactsPermissionReminderView autoSetDimension:ALDimensionHeight toSize:0];
```
charlesmchen: I wonder if we should make this a cell in the table. That way it could scroll offscreen, which would be especially nice on smaller form factors.

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -101,6 +115,20 @@ - (void)loadView
...
+{
+    _hideContactsPermissionReminderViewConstraint.active = !isVisible;
+}
+
+- (void)showSearchBar:(BOOL)isVisible
+{
+    if (isVisible) {
+        self.tableViewController.tableView.tableHeaderView = self.searchBar;
+    } else {
+        self.tableViewController.tableView.tableHeaderView = nil;
```
charlesmchen: If we hide the search bar, users can't "invite non-contacts via SMS."  It might also be nice to modify `SelectRecipientViewController` to offer to "invite via SMS" when it's `shouldValidatePhoneNumbers` delegate property is true.

Signal/src/ViewControllers/MessageComposeTableViewController.m
michaelkirk: What would you be scrolling if there are no contacts?

Signal/src/ViewControllers/MessageComposeTableViewController.m
michaelkirk: Good point. I think ultimately the find by phone number screen should incorporate the faster feedback and "invite" functionality currently available on the search bar. 

I think in the immediate I'll replace the search bar and add a card.

Signal/src/ViewControllers/MessageComposeTableViewController.m
charlesmchen: Well, if we show the search bar (see other comment), you might have:

* This header.
* Search bar.
* "find by phone number".
* "no contacts"?  (I think you suppress that when this header is visible?)
* two different possible "invite by sms" cells.

michaelkirk:
I ended up making some additional changes to avoid a flickering repaint when viewing the compose controller for the first time (after install)

PTAL @charlesmchen: https://github.com/WhisperSystems/Signal-iOS/compare/b2d96d99f7653459e02e94a9a944db671a145543...mkirk/missing-contacts-compose?expand=1

charlesmchen:
Will there still be a flicker if you enter "new 1:1 conversation view", edit your contacts, return to Signal?

michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/bfc4ff6f05b718d0c5bd374e96f43c472040baec

michaelkirk:
> Will there still be a flicker if you enter "new 1:1 conversation view", edit your contacts, return to Signal?

@charlesmchen - can you clarify the steps to reproduce the flickering you're talking about?

charlesmchen:
> @charlesmchen - can you clarify the steps to reproduce the flickering you're talking about?

Huh, I thought there was a flicker but there I can't repro anymore. Please ignore.

PR #2080, opened 2017-05-08
Try to deduce attachment MIME type from the file extension if possible.

charlesmchen:
Requires: https://github.com/WhisperSystems/SignalServiceKit/pull/206

Also includes some small fixes.

PTAL @michaelkirk


Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -316,23 +316,8 @@ - (void)performEditingAction:(SEL)action
...
-                    utiType = (NSString *)kUTTypeMPEG4Audio;
-                } else if ([_contentType isEqualToString:@"audio/aiff"] ||
-                    [_contentType isEqualToString:@"audio/x-aiff"]) {
-                    utiType = (NSString *)kUTTypeAudioInterchangeFileFormat;
-                } else {
-                    OWSAssert(0);
-                    utiType = (NSString *)kUTTypeAudio;
-                }
+                OWSAssert(0);
+                utiType = (NSString *)kUTTypeAudio;
```
charlesmchen: Moved to `MIMETypeUtil`.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -325,6 +325,10 @@ - (void)cancelVoiceMemoIfNecessary
 
 - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceiveTouch:(UITouch *)touch
 {
+    if (self.rightBarButtonItem != self.voiceMemoButton) {
+        return NO;
```
charlesmchen: Not strictly necessary but nice to be defensive.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -3158,7 +3162,7 @@ - (void)endRecordingVoiceMemo
     self.audioRecorder = nil;
 
     NSString *filename = [NSLocalizedString(@"VOICE_MESSAGE_FILE_NAME", @"Filename for voice messages.")
-        stringByAppendingPathExtension:[MIMETypeUtil fileExtensionForUTIType:(NSString *)kUTTypeMPEG4Audio]];
+        stringByAppendingPathExtension:@".m4a"];
```
charlesmchen: I'm pretty sure we should .m4a, not .aac.  .mp4 is the container format and .m4a is the right extension for a .mp4 that contains .aac (as opposed to video).  But I think Android is using .aac, which indicates a raw aac-encoded a bitstream. I'm not sure, but I doubt that's what Android is sending. I'm going to follow up with Moxie.

charlesmchen: Actually, it occurs to me now that it's not sufficient to do this fix - we should stop using `fileExtensionForUTIType:` entirely.  Incoming .m4a files without a filename would be wrongly saved as .mp4 which, when shared, will be wrongly treated as video.  I'll do this now if possible.

Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -204,6 +216,12 @@ class SignalAttachment: NSObject {
     // Returns the file extension for this attachment or nil if no file extension
     // can be identified.
     var fileExtension: String? {
+        if let filename = filename {
+            let fileExtension = (filename as NSString).pathExtension
+            if fileExtension.characters.count > 0 {
+                return fileExtension
+            }
+        }
```
charlesmchen: A small but important fix - prefer to determine the file extension from the filename than the mimetype.  This will avoid some cases where attachments will be relabeled.  We should trust the file extension we have.

Signal/src/ViewControllers/MessagesViewController.m
michaelkirk: does this indicate an error? i.e. should we have an assertion failure?

Signal/src/ViewControllers/MessagesViewController.m
michaelkirk: Ah, this is what https://github.com/WhisperSystems/SignalServiceKit/pull/206 is about, right?

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: Not strictly, so an assert wouldn't be appropriate.

Signal/src/ViewControllers/MessagesViewController.m
charlesmchen: Yup, exactly.

PR #2083, opened 2017-05-09
Flag voice messages as such in protos.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/207

PR #2084, opened 2017-05-09
Always sync and log push tokens. Apply OWSAlerts in more places.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/Signal-Prefix.pch
+++ b/Signal/Signal-Prefix.pch
@@ -20,6 +20,4 @@
     #import <SignalServiceKit/OWSAnalytics.h>
     #import <SignalServiceKit/OWSDispatch.h>
     #import <SignalServiceKit/iOSVersions.h>
-
-    #define SignalAlertView(title,msg) [[[UIAlertView alloc] initWithTitle:title message:msg delegate:nil cancelButtonTitle:NSLocalizedString(@"OK", @"") otherButtonTitles:nil, nil] show]    
 #endif
--- a/Signal/src/ViewControllers/AdvancedSettingsTableViewController.m
+++ b/Signal/src/ViewControllers/AdvancedSettingsTableViewController.m
@@ -131,7 +131,8 @@ typedef NS_ENUM(NSInteger, AdvancedSettingsTableViewControllerSection) {
     } else if ([tableView cellForRowAtIndexPath:indexPath] == self.registerPushCell) {
         [OWSSyncPushTokensJob runWithPushManager:[PushManager sharedManager]
                                   accountManager:[Environment getCurrent].accountManager
-                                     preferences:[Environment preferences]];
+                                     preferences:[Environment preferences]
+                                      showAlerts:YES];
     } else {
         DDLogDebug(@"%@ Ignoring cell selection at indexPath: %@", self.tag, indexPath);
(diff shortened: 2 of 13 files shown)


Signal/src/ViewControllers/AdvancedSettingsTableViewController.m
```diff
@@ -130,21 +129,9 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
...
-                                               accountManager:[Environment getCurrent].accountManager
-                                                  preferences:[Environment preferences]];
-        syncJob.uploadOnlyIfStale = NO;
-        [syncJob run]
-            .then(^{
-                DDLogDebug(@"%@ Successfully ran syncPushTokensJob.", self.tag);
-                SignalAlertView(NSLocalizedString(@"PUSH_REGISTER_SUCCESS", @"Alert title"), nil);
-            })
-            .catch(^(NSError *error) {
-                DDLogError(@"%@ Failed to run syncPushTokensJob with error: %@", self.tag, error);
```
michaelkirk: What you have now is consolidating code nicely, but I liked giving the user visual feedback here, while just logging it elsewhere. Reason being that this is in response to an *explicit* action the user takes.

e.g. How does the user know "Did it work?"

Signal/Signal-Prefix.pch
```diff
@@ -20,6 +20,4 @@
     #import <SignalServiceKit/OWSAnalytics.h>
     #import <SignalServiceKit/OWSDispatch.h>
     #import <SignalServiceKit/iOSVersions.h>
-
-    #define SignalAlertView(title,msg) [[[UIAlertView alloc] initWithTitle:title message:msg delegate:nil cancelButtonTitle:NSLocalizedString(@"OK", @"") otherButtonTitles:nil, nil] show]
```
michaelkirk: so long deprecated code 👋

Signal/src/views/OWSAlerts.swift
```diff
@@ -23,15 +23,18 @@ import Foundation
         UIApplication.shared.frontmostViewController?.present(alertController, animated: true, completion: nil)
     }
 
+    public class func showAlert(withTitle title: String) {
+        self.showAlert(withTitle: title, message: nil, buttonTitle: NSLocalizedString("OK", comment: ""))
+    }
+
     public class func showAlert(withTitle title: String, message: String) {
```
michaelkirk: nit: default params might dedupe this helpfully

https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Functions.html#//apple_ref/doc/uid/TP40014097-CH10-ID169

Signal/src/ViewControllers/AdvancedSettingsTableViewController.m
charlesmchen: Restored!

Signal/src/views/OWSAlerts.swift
charlesmchen: Well, I've done this but we're almost entirely using this from Objective-C and you can't use default params from Objective-C - at least not to my knowledge.

Issue #2082, opened 2017-05-09
Sent messages pop up in the message editing field


jhellen:
- [x] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description
Previously sent messages keep popping up in the message editing field.

### Steps to reproduce
- Start writing to a contact

**Actual result:**  Previously sent message has popped up in the message editing field but it does not happen every time. So you need to erase the message editing field every time you send a message. Sometimes it has been working fine for a couple of months but then the problem reappears.
**Expected result:** The message editing field should be empty or contain previously unsent text.

### Device info

**Device**: iPhone 6
**iOS version**: 10.3.1
**Signal version:** 2.11.3

TreatyOfRome:
This is a duplicate of #1537.

michaelkirk:
closing as dupe of https://github.com/WhisperSystems/Signal-iOS/issues/1537, fix coming in #2085

charlesmchen:
This is fixed and will ship with v2.12.

Issue #1537, opened 2016-12-09
Message draft reappearing 


riyapenn:
- [X] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description
Message draft reappears in message input field even if message was sent in the past

### Steps to reproduce -- intermittent, not consistent
- Compose message
- send message
- leave/close out of Signal at some point
- open Signal and view a conversation thread
- see draft of a message

**Actual result:** See draft of a message, this happens for some threads and is not consistent
**Expected result:** No draft appears if message was sent

### Device info

**Device**: iPhone 6S, 6
**iOS version**: 10.0.2, 10.1.1
**Signal version:** 2.6.7

### Link to debug log
User 1: https://gist.github.com/anonymous/c41f100e7868310d917cc2edf13b8f10
User 2: https://gist.github.com/anonymous/2e6a42d1aa29c2f10e224aeeb874ddb2

**User 1 potentially relevant information here:** 
**Original message written:**
```
2016/12/07 22:13:39:295  Error in websocket heartbeat: Invalid State: Cannot call `sendPing:error:` until connection is open.
2016/12/07 22:14:04:670  [TSAttachmentStream] Created file at /var/mobile/Containers/Data/Application/1150D4C0-75C1-4F49-A23A-5892CF925BF7/Documents/Attachments/6BC5D6BA-D26D-4D07-B98D-E80CFD8DD555.jpeg
2016/12/07 22:14:04:699  [TSNetworkManager] Making request: <TSAllocAttachmentRequest: 0x170637ac0> { URL: v1/attachments }
2016/12/07 22:14:06:695  [OWSUploadingService] Uploaded attachment.
2016/12/07 22:14:06:774  [TSNetworkManager] Making request: <TSSubmitMessageRequest: 0x17443f740> { URL: v1/messages/[ REDACTED_PHONE_NUMBER ] }
2016/12/07 22:14:07:164  [TSNetworkManager] Making request: <TSSubmitMessageRequest: 0x17443eb00> { URL: v1/messages/[ REDACTED_PHONE_NUMBER ] }
2016/12/07 22:14:07:289  Succesfully sent sync transcript.
2016/12/07 22:14:09:833  [TSNetworkManager] Making request: <TSSubmitMessageRequest: 0x1708250a0> { URL: v1/messages/[ REDACTED_PHONE_NUMBER ] }
[...]

PR #2085, opened 2017-05-09
Always clear message drafts after sending a text message.

charlesmchen:
PTAL @michaelkirk


michaelkirk:
fixes 
https://github.com/WhisperSystems/Signal-iOS/issues/2082
https://github.com/WhisperSystems/Signal-iOS/issues/1537

PR #2087, opened 2017-05-09
Rework the “disappearing messages” logic.

charlesmchen:
* Convert `OWSDisappearingMessagesJob` to a singleton since its work is expensive.
* Ensure `OWSDisappearingMessagesJob` does all of its work on a private queue, off the main thread.
* Rework scheduling of `OWSDisappearingMessagesJob` runs to use a NSTimer, so that we can cancel redundant scheduled runs.
* Move NSTimer+OWS category to SSK.
* Fix bug wherein some disappearing messages never got deleted.

PTAL @michaelkirk 

See: https://github.com/WhisperSystems/SignalServiceKit/pull/208


Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -1646,6 +1644,12 @@ - (UICollectionViewCell *)collectionView:(JSQMessagesCollectionView *)collection
     }
     cell.delegate = collectionView;
 
+    //    @property (nonatomic) uint64_t expiresAtSeconds;
+    //    @property (nonatomic) uint32_t expiresInSeconds;
+    //
+    //    @property (nonatomic, copy) NSDate *messageDate;
+    //    @property (nonatomic, retain) NSString *messageBody;
+
```
michaelkirk: Did you intend to leave this comment in? If so can you preface it with an explanation?

charlesmchen: Removed!

PR #2088, opened 2017-05-09
Fix edge case where screen protection is not removed.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -417,7 +417,10 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
       if ([TSAccountManager isRegistered]) {
           dispatch_sync(dispatch_get_main_queue(), ^{
-              [self protectScreen];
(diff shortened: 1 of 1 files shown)


Signal/src/AppDelegate.m
```diff
@@ -417,7 +417,10 @@ - (void)applicationWillResignActive:(UIApplication *)application {
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
       if ([TSAccountManager isRegistered]) {
           dispatch_sync(dispatch_get_main_queue(), ^{
-              [self protectScreen];
+              if ([UIApplication sharedApplication].applicationState != UIApplicationStateActive) {
```
michaelkirk: Was this necessary for the fix? Or just an additional nice-to-have?

charlesmchen: It's part of the fix. We should never show the screen protection if the app is active.

michaelkirk: Got it - I guess with the async dispatches it would be possible to toggle in/out of the app and hit this race condition.

Separately, and this predates your changes, but do we really want to dispatch_**sync** here? It freaks me out a little bit.

And if we change it to `async`, I suppose the `[application endBackgroundTask:` should be moved within this block.

PR #2091, opened 2017-05-09
Mkirk/edit 1 1 contact

michaelkirk:
Add functionality to edit 1:1 contact details.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ContactsViewHelper.h
+++ b/Signal/src/ViewControllers/ContactsViewHelper.h
@@ -7,12 +7,21 @@ NS_ASSUME_NONNULL_BEGIN
 @class ContactsViewHelper;
 @class Contact;
 @class SignalAccount;
+@protocol CNContactViewControllerDelegate;
 
 @protocol ContactsViewHelperDelegate <NSObject>
 
 - (void)contactsViewHelperDidUpdateContacts;
 
+@optional
+
 - (BOOL)shouldHideLocalNumber;
+
+@end
+
+@protocol ContactEditingDelegate <CNContactViewControllerDelegate>
+
+- (void)didFinishEditingContact;
 
 @end
 
@@ -45,6 +54,10 @@ NS_ASSUME_NONNULL_BEGIN
 - (NSArray<SignalAccount *> *)signalAccountsMatchingSearchString:(NSString *)searchText;
 
 - (NSArray<Contact *> *)nonSignalContactsMatchingSearchString:(NSString *)searchText;
+
+- (void)presentContactViewControllerForRecipientId:(NSString *)recipientId
+                                fromViewController:(UIViewController<ContactEditingDelegate> *)fromViewController
+                                   editImmediately:(BOOL)shouldEditImmediately;
 
 @end
 
--- a/Signal/src/ViewControllers/ContactsViewHelper.h
+++ b/Signal/src/ViewControllers/ContactsViewHelper.h
@@ -55,6 +55,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSArray<Contact *> *)nonSignalContactsMatchingSearchString:(NSString *)searchText;
 
+/**
+ * NOTE: This method calls `[UIUtil applyDefaultSystemAppearence]`.
+ * When using this method, you must call `[UIUtil applySignalAppearence]` once contact editing is   finished;
+ */
 - (void)presentContactViewControllerForRecipientId:(NSString *)recipientId
                                 fromViewController:(UIViewController<ContactEditingDelegate> *)fromViewController
                                    editImmediately:(BOOL)shouldEditImmediately;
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/ContactsViewHelper.h
```diff
@@ -7,15 +7,24 @@ NS_ASSUME_NONNULL_BEGIN
...
 
+@optional
+
 - (BOOL)shouldHideLocalNumber;
 
 @end
 
+@protocol ContactEditingDelegate <CNContactViewControllerDelegate>
+
+- (void)didFinishEditingContact;
```
charlesmchen: Could just be @optional method of `ContactsViewHelperDelegate`, but this works too.

charlesmchen: I take it back, this makes sense.

Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -277,6 +283,91 @@ - (BOOL)doesContact:(Contact *)contact matchSearchTerms:(NSArray<NSString *> *)s
...
+        [fromViewController presentViewController:alertController animated:YES completion:nil];
+        return;
+    }
+
+    CNContactViewController *_Nullable contactViewController;
+    if (signalAccount) {
+        CNContact *_Nullable cnContact = signalAccount.contact.cnContact;
+        if (cnContact) {
+            if (shouldEditImmediately) {
+                // Not acutally a "new" contact, but this brings up the edit form rather than the "Read" form
```
charlesmchen: s/acutally/actually/

Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -277,6 +283,91 @@ - (BOOL)doesContact:(Contact *)contact matchSearchTerms:(NSArray<NSString *> *)s
...
+
+    // We want the presentation to imply a "replacement" in this case.
+    if (shouldEditImmediately) {
+        navigationController.modalTransitionStyle = UIModalTransitionStyleCrossDissolve;
+    }
+    [fromViewController presentViewController:navigationController animated:YES completion:nil];
+
+    // HACK otherwise CNContactViewController Navbar is shown as black.
+    // RADAR rdar://28433898 http://www.openradar.me/28433898
+    // CNContactViewController incompatible with opaque navigation bar
```
charlesmchen: Maybe add a note in the header for this method that the caller should call the "opposite" `UIUtil` method in view will appear?

Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
```diff
@@ -115,6 +120,58 @@ - (BOOL)isGroupThread
...
+    if (![self.thread isKindOfClass:[TSContactThread class]]) {
+        DDLogError(@"%@ unexpected thread: %@ in %s", self.tag, self.thread, __PRETTY_FUNCTION__);
+        OWSAssert(NO);
+        return;
+    }
+
+    TSContactThread *contactThread = (TSContactThread *)self.thread;
+    [self.contactsViewHelper presentContactViewControllerForRecipientId:contactThread.contactIdentifier
+                                                     fromViewController:self
+                                                        editImmediately:YES];
```
charlesmchen: In "group" mode, this view shows the edit view if the thread name or avatar is tapped. Shall we do the same here?

Signal/src/ViewControllers/ContactsViewHelper.m
michaelkirk: done

Signal/src/ViewControllers/ContactsViewHelper.m
michaelkirk: done

Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
michaelkirk: I'd rather add an explicit "Edit" right bar button to the group settings and remove the "tap-to-edit" functionality from the group settings. 

Motivation:
1. the right bar button "Edit" item is standard - users know to look for it.
2. tap to edit seems more likley to be misfired.

Separately, what do you think about using a "fade" transition for the "group edit" modal instead of the modal pop-over?

charlesmchen: Why use a fade transition?  I tend to favor them only in contents where there is strong visual relationship between the before and after content, which I don't think is true here?

michaelkirk: >  I tend to favor them only in contents where there is strong visual relationship between the before and after content, which I don't think is true here?

img_0663
img_0664

Well, I'd argue that there is a pretty strong relationship between the editable portion of the before/after content, but I can table this until such a time as it makes more sense.

michaelkirk: > In "group" mode, this view shows the edit view if the thread name or avatar is tapped. Shall we do the same here?

I had a long hard look in the mirror and have added this.

PR #2094, opened 2017-05-10
Show example phone numbers.

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/SignalServiceKit/pull/209

### Example: US

screen shot 2017-05-10 at 10 57 15 am

### Example: Mexico

screen shot 2017-05-10 at 10 57 22 am

### DIFF
--- a/Signal/src/ViewControllers/SelectRecipientViewController.m
+++ b/Signal/src/ViewControllers/SelectRecipientViewController.m
@@ -234,6 +234,13 @@ NSString *const kSelectRecipientViewControllerCellIdentifier = @"kSelectRecipien
     NSString *examplePhoneNumber = [PhoneNumberUtil examplePhoneNumberForCountryCode:countryCode];
     OWSAssert(!examplePhoneNumber || [examplePhoneNumber hasPrefix:callingCode]);
     if (examplePhoneNumber && [examplePhoneNumber hasPrefix:callingCode]) {
+        NSString *formattedPhoneNumber =
+            [PhoneNumber bestEffortFormatPartialUserSpecifiedTextToLookLikeAPhoneNumber:examplePhoneNumber
+                                                         withSpecifiedCountryCodeString:countryCode];
+        if (formattedPhoneNumber.length > 0) {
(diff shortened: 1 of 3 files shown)


michaelkirk:
The examples are kind of hard to parse, since their just long strings of numbers. 

Is it possible/Do you think it would be clearer to include formatting in the example?

charlesmchen:
> Is it possible relatively easy to include formatting in the example?

Done.

Signal/src/ViewControllers/SelectRecipientViewController.m
```diff
@@ -442,9 +481,17 @@ - (void)updateTableContents
         [self.phoneNumberTextField autoPinEdge:ALEdgeLeft toEdge:ALEdgeRight ofView:phoneNumberLabel withOffset:0];
         [self.phoneNumberTextField autoVCenterInSuperview];
 
+        // Example row.
+        UIView *examplePhoneNumberRow = [self createRowWithHeight:examplePhoneNumberRowHeight
+                                                      previousRow:phoneNumberRow
+                                                        superview:cell.contentView];
+        [examplePhoneNumberRow addSubview:self.examplePhoneNumberLabel];
+        [self.examplePhoneNumberLabel autoVCenterInSuperview];
+        [self.examplePhoneNumberLabel autoPinEdgeToSuperviewEdge:ALEdgeRight withInset:20.f];
```
michaelkirk: nit: Where's the left constraint? 

Probably not an issue for the example numbers which will be relatively short, but I think it's good practice to fully constrain your view, and if you are going for a right alignment, do that on the label.

charlesmchen: Sorry I just saw this. Interesting, I've usually taken a different approach. I'll think this over!

michaelkirk: see e.g. https://github.com/WhisperSystems/Signal-iOS/commit/9dc9813deaee34ab4aa6dbe85bcab57cabdbf7d8

PR #2095, opened 2017-05-10
Always honor filtering in contact-related views.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
> By "filtering" you mean respecting the isSignalAccountHidden right?

Yup.

PR #2101, opened 2017-05-11
Mkirk/polite intersection

michaelkirk:
Be more careful about when we do contact intersections.

**Always:** whenever the system sends a contact-changed notification

**Sometimes:** when the app becomes active

1. If the contacts changed since we last checked - we won't always get a notification when contacts change, e.g. if the app was suspended

2. If 12 hours has passed (in line with Android)

PTAL @charlesmchen

### DIFF
--- a/Signal/src/contact/SystemContactsFetcher.swift
+++ b/Signal/src/contact/SystemContactsFetcher.swift
@@ -14,6 +14,7 @@ import ContactsUI
 class SystemContactsFetcher: NSObject {
 
     private let TAG = "[SystemContactsFetcher]"
+    var lastContactUpdateHash: Int?
 
     public weak var delegate: SystemContactsFetcherDelegate?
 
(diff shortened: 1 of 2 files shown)


Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -145,3 +186,22 @@ class SystemContactsFetcher: NSObject {
     }
 
 }
+
+struct HashableArray<Element: Hashable>: Hashable {
```
charlesmchen: Why a collection?  Why not just a function that computes the hash for a collection of Hashables?

michaelkirk: Just style. The idiomatic way to get this kind of a hash for something is to call `.hash` (or `.hashValue` in swift).

PR #2102, opened 2017-05-11
Don’t add blocked users to new groups and handle block alert results …

charlesmchen:
…correctly.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/NewGroupViewController.m
+++ b/Signal/src/ViewControllers/NewGroupViewController.m
@@ -4,6 +4,7 @@
 
 #import "NewGroupViewController.h"
 #import "AddToGroupViewController.h"
+#import "BlockListUIUtils.h"
 #import "ContactTableViewCell.h"
 #import "ContactsViewHelper.h"
 #import "Environment.h"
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -205,37 +206,56 @@ - (void)updateTableContents
...
-                    // In the "contacts" section, we label members as such when editing an existing group.
-                    cell.accessoryMessage = NSLocalizedString(
-                        @"NEW_GROUP_MEMBER_LABEL", @"An indicator that a user is a member of the new group.");
-                } else if (isBlocked) {
-                    cell.accessoryMessage = NSLocalizedString(
-                        @"CONTACT_CELL_IS_BLOCKED", @"An indicator that a contact has been blocked.");
-                } else {
-                    OWSAssert(cell.accessoryMessage == nil);
-                }
+            [nonContactsSection
```
michaelkirk: looks like the `customCellBlock` part is just clang-format having a party?

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -205,37 +206,56 @@ - (void)updateTableContents
...
-                return cell;
-            }
-                                            customRowHeight:[ContactTableViewCell rowHeight]
-                                            actionBlock:^{
-                                                [weakSelf removeRecipientId:recipientId];
-                                            }]];
+                    return cell;
+                }
+                            customRowHeight:[ContactTableViewCell rowHeight]
+                            actionBlock:^{
```
michaelkirk: ...and that the substantive change is here. (this part LGTM)

Signal/src/ViewControllers/NewGroupViewController.m
charlesmchen: Yeah ¯\_(ツ)_/¯

charlesmchen: I'm inclined to let clang-format have its way.

PR #2107, opened 2017-05-12
Ensure contact delegate helper is set during initialization.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ContactsViewHelper.m
+++ b/Signal/src/ViewControllers/ContactsViewHelper.m
@@ -26,18 +26,25 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic) NSArray<NSString *> *blockedPhoneNumbers;
 
+@property (nonatomic) BOOL suppressDelegateEvents;
+
 @end
 
 #pragma mark -
 
 @implementation ContactsViewHelper
 
-- (instancetype)init
+- (instancetype)initWithDelegate:(id<ContactsViewHelperDelegate>)delegate
 {
     self = [super init];
     if (!self) {
         return self;
     }
+
+    OWSAssert(delegate);
+    _delegate = delegate;
+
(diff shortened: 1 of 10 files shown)


Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -26,19 +26,26 @@ @interface ContactsViewHelper ()
...
 {
     self = [super init];
     if (!self) {
         return self;
     }
 
+    OWSAssert(delegate);
+    _delegate = delegate;
+
+    self.suppressDelegateEvents = YES;
```
charlesmchen: We don't want to fire delegate events during initialization, but we do want to consult the "should hide local number" delegate method.

michaelkirk: Why do we want the "shouldHideLocalNumber" method, it's not obvious that we're using it during init?

And there are only two delegate methods at the moment, so when you say we don't want to fire delegate events except using the shouldHideLocalNumber, you mean we don't want to fire the other one, which is just contactsViewHelperDidUpdateContacts, right?

I imagine this makes sense to you after digging a bit, but at first glance I don't know what behavior you're trying to fix. Can you leave a code comment that outlines the rationale?

michaelkirk: Ah, sorry, with the GH split diff I missed the call to `[self updateContacts];` which uses the shouldHideLocalNumber (via another method call).

This makes sense, but a comment would still be helpful.

michaelkirk: maybe the variable could be called "shouldNotifyDelegateOfUpdatedContacts`

charlesmchen: Done. PTAL @michaelkirk

PR #2109, opened 2017-05-12
Align photo and audio bubble widths.

charlesmchen:
PTAL @michaelkirk 

img_0566


Signal/src/Models/TSMessageAdapaters/TSVideoAttachmentAdapter.m
```diff
@@ -355,6 +356,9 @@ - (void)ensureAudioDurationSeconds
 - (CGSize)mediaViewDisplaySize {
     CGSize size = [super mediaViewDisplaySize];
     if ([self isAudio]) {
+        if ([[UIDevice currentDevice] isiPhoneVersionSixOrMore]) {
```
michaelkirk: nit: It would be nice to use the existing logic rather than duplicate this 1.2 constant from `ows_adjustBubbleSize` in case it changes in the future.

charlesmchen: Done.

PR #2117, opened 2017-05-15
Charlesmchen/bluetooth audio

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessagesViewController.m
+++ b/Signal/src/ViewControllers/MessagesViewController.m
@@ -3148,7 +3148,9 @@ typedef enum : NSUInteger {
     OWSAssert(session.recordPermission == AVAudioSessionRecordPermissionGranted);
 
     NSError *error;
-    [session setCategory:AVAudioSessionCategoryRecord error:&error];
+    [session setCategory:AVAudioSessionCategoryRecord
+             withOptions:AVAudioSessionCategoryOptionAllowBluetooth
+                   error:&error];
     if (error) {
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -3123,7 +3123,9 @@ - (void)startRecordingVoiceMemo
     // Setup audio session
     AVAudioSession *session = [AVAudioSession sharedInstance];
     NSError *error;
-    [session setCategory:AVAudioSessionCategoryRecord error:&error];
+    [session setCategory:AVAudioSessionCategoryRecord
+             withOptions:AVAudioSessionCategoryOptionAllowBluetooth
```
charlesmchen: We need "allow bluetooth" when recording voice messages or calls...

Signal/src/call/CallAudioService.swift
```diff
@@ -98,14 +98,15 @@ import AVFoundation
...
                             mode: AVAudioSessionModeVideoChat,
-                            options: .defaultToSpeaker)
+                            options: [.defaultToSpeaker])
         } else if call.isSpeakerphoneEnabled {
             setAudioSession(category: AVAudioSessionCategoryPlayAndRecord,
                             mode: AVAudioSessionModeVoiceChat,
-                            options: .defaultToSpeaker)
+                            options: [.defaultToSpeaker])
         } else {
             setAudioSession(category: AVAudioSessionCategoryPlayAndRecord,
```
charlesmchen: ...but not when speakerphone is enabled.

PR #2119, opened 2017-05-15
Mkirk/bordered avatar

michaelkirk:
On a wild tear this weekend...

- use shared avatar building code for concise and consistent presentation
- use higher resolution generated avatar for call screen. Current one is blurry.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/views/AvatarImageView.swift
+++ b/Signal/src/views/AvatarImageView.swift
@@ -4,7 +4,6 @@
 
 import UIKit
 
-@IBDesignable
 class AvatarImageView: UIImageView {
 
     init() {


Signal/src/Models/OWSAvatarBuilder.m
```diff
@@ -13,10 +14,15 @@ @implementation OWSAvatarBuilder
 
 + (UIImage *)buildImageForThread:(TSThread *)thread contactsManager:(OWSContactsManager *)contactsManager
 {
+    const CGFloat kDefaultAvatarDiameter = 100;
+    return [self buildImageForThread:thread contactsManager:contactsManager diameter:kDefaultAvatarDiameter];
```
charlesmchen: You do not have to address this in this PR, but:

A) We should unify this "avatar" sizing with the "custom avatar" sizing at 210px.

See:

* iOS https://github.com/WhisperSystems/Signal-iOS/blob/5b9978b07fcaef170d0ae09dc6f2ae94e2b4450e/Signal/src/ViewControllers/GroupViewHelper.m#L121
* Android https://github.com/WhisperSystems/Signal-Android/blob/4112f23f335ca7fe767edae52bb3b264d9e51c92/src/org/thoughtcrime/securesms/GroupCreateActivity.java#L105

B) _Or_ we should make sure that these avatar sizes reflect the actual view size _in pixels not points_.  In `ContactTableViewCell`, these are shown at 40pt, so 40, 80, 120px. I'm not sure about `InboxTableViewCell`, etc.  But there would be a real quality advantage to rendering to the exact view size.

Signal/src/views/AvatarImageView.swift
```diff
@@ -1,14 +1,45 @@
...
+    }
+
+    func configureView() {
+        self.layer.minificationFilter = kCAFilterTrilinear
+        self.layer.magnificationFilter = kCAFilterTrilinear
+        self.layer.borderWidth = 0.5
         self.layer.masksToBounds = true
-        self.layer.cornerRadius = self.frame.size.width / 2
+        if self.image == nil {
+            self.image = #imageLiteral(resourceName: "empty-group-avatar")
```
charlesmchen: So, shouldn't we use a different default for non-groups or will this always be overwritten in the 1:1 case?

Signal/src/views/AvatarImageView.swift
```diff
@@ -1,14 +1,45 @@
-//  Created by Michael Kirk on 12/11/16.
-//  Copyright © 2016 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 import UIKit
 
 @IBDesignable
 class AvatarImageView: UIImageView {
```
charlesmchen: Can we use `AvatarImageView` in the (new group, update group, conversation settings) views?

Signal/src/views/AvatarImageView.swift
michaelkirk: It's always overwritten in the 1:1 case, but there's no real reason for this default since we're always setting it. I've removed it.

Signal/src/views/AvatarImageView.swift
michaelkirk: done

michaelkirk:
I rebased and made a couple additions.

always require explicit dimensions when building avatars: https://github.com/WhisperSystems/Signal-iOS/pull/2119/commits/5d587666c04aa29ed85c99ad8bcfa46b09a92c26

use avatar view in new/edit group views:
https://github.com/WhisperSystems/Signal-iOS/pull/2119/commits/67a7eaf7d21009909d6f5e46f60d5b34dad58752

PTAL @charlesmchen

Signal/src/ViewControllers/InboxTableViewCell.xib
```diff
@@ -16,15 +16,16 @@
...
-                        <rect key="frame" x="0.0" y="0.0" width="400" height="72"/>
+                        <rect key="frame" x="0.0" y="0.0" width="400" height="47"/>
                         <subviews>
-                            <imageView userInteractionEnabled="NO" contentMode="scaleAspectFill" horizontalHuggingPriority="251" verticalHuggingPriority="251" translatesAutoresizingMaskIntoConstraints="NO" id="p9o-x6-nT5">
-                                <rect key="frame" x="10" y="10" width="52" height="53"/>
+                            <imageView userInteractionEnabled="NO" contentMode="scaleAspectFill" horizontalHuggingPriority="251" verticalHuggingPriority="251" translatesAutoresizingMaskIntoConstraints="NO" id="p9o-x6-nT5" customClass="AvatarImageView" customModule="Signal" customModuleProvider="target">
+                                <rect key="frame" x="10" y="10" width="52" height="52"/>
                                 <constraints>
+                                    <constraint firstAttribute="width" secondItem="p9o-x6-nT5" secondAttribute="height" multiplier="1:1" id="IjZ-Ac-GiV"/>
```
michaelkirk: only real change is enforcing a 1:1 aspect ratio on inbox table view avatar

charlesmchen:
LGTM

PR #2120, opened 2017-05-15
Don't show "no signal accounts" until contact intersection has completed

michaelkirk:
at least once

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MessageComposeTableViewController.m
+++ b/Signal/src/ViewControllers/MessageComposeTableViewController.m
@@ -396,7 +396,8 @@ NS_ASSUME_NONNULL_BEGIN
     if (!hasSearchText && helper.signalAccounts.count < 1) {
         // No Contacts
 
-        if (self.contactsViewHelper.contactsManager.isSystemContactsAuthorized) {
+        if (self.contactsViewHelper.contactsManager.isSystemContactsAuthorized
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -462,10 +463,15 @@ - (void)presentInviteFlow
...
-        [self showContactsPermissionReminder:NO];
+        if (self.contactsViewHelper.contactsManager.hasFetchedSignalAccountsAtLeastOnce
+            && self.contactsViewHelper.signalAccounts.count < 1
+            && ![[Environment preferences] hasDeclinedNoContactsView]) {
+            self.isNoContactsModeActive = YES;
+        } else {
+            self.isNoContactsModeActive = NO;
+        }
 
+        [self showContactsPermissionReminder:NO];
```
charlesmchen: Shouldn't this read:

```
[self showContactsPermissionReminder:! self.isNoContactsModeActive];
```

charlesmchen: Sorry this is wrong; ignore;

michaelkirk: Well, you're not entirely wrong in that this method sucks.

michaelkirk:
In 3ca2d0cbe0ee73637a69be0aa7ad9b86a5832edd

PR #2122, opened 2017-05-15
Mkirk/mic perms

michaelkirk:
Currently new users are hit with a confusing double whammy microphone permission request if they try to record a voice message before making a call.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MessagesViewController.m
+++ b/Signal/src/ViewControllers/MessagesViewController.m
@@ -3105,6 +3105,22 @@ typedef enum : NSUInteger {
 
 #pragma mark - Audio
 
+- (void)requestRecordingVoiceMemo
+{
+    OWSAssert([NSThread isMainThread]);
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -3105,6 +3105,22 @@ - (TSInteraction *)interactionAtIndexPath:(NSIndexPath *)indexPath {
 
 #pragma mark - Audio
 
+- (void)requestRecordingVoiceMemo
+{
+    OWSAssert([NSThread isMainThread]);
+    [[AVAudioSession sharedInstance] requestRecordPermission:^(BOOL granted) {
```
charlesmchen: Probably should use weakSelf in case messages view is dismissed while "recording permission" alert is presented, e.g. if app receives an "export with Signal" event.

michaelkirk: done.

PR #2124, opened 2017-05-16
Charlesmchen/unread indicator

charlesmchen:
* I could have may the unread indicator a new kind of `TSInfoMessage`, but it has a very different appearance. 

PTAL @michaelkirk

### DIFF
--- a/Signal/src/UIView+OWS.m
+++ b/Signal/src/UIView+OWS.m
@@ -131,6 +131,47 @@ CGFloat ScaleFromIPhone5(CGFloat iPhone5Value)
     [self setContentCompressionResistancePriority:UILayoutPriorityRequired forAxis:UILayoutConstraintAxisVertical];
 }
 
+#pragma mark - Manual Layout
+
+- (CGFloat)left
+{
+    return self.frame.origin.x;
+}
+
+- (CGFloat)right
+{
+    return self.frame.origin.x + self.frame.size.width;
+}
+
+- (CGFloat)top
+{
+    return self.frame.origin.y;
+}
+
+- (CGFloat)bottom
+{
+    return self.frame.origin.x + self.frame.size.height;
+}
+
+- (CGFloat)width
+{
+    return self.frame.size.width;
+}
+
+- (CGFloat)height
+{
+    return self.frame.size.height;
+}
+
+- (void)centerOnSuperview
+{
+    self.frame = CGRectMake(
+        round(self.superview.bounds.origin.x + (self.superview.bounds.size.width - self.frame.size.width) * 0.5f),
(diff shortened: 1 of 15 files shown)


Signal/src/views/OWSUnreadIndicatorCell.m
```diff
@@ -0,0 +1,72 @@
...
+        [self.contentView addSubview:self.rightPathView];
+    }
+}
+
+- (void)layoutSubviews
+{
+    [self.label sizeToFit];
+    [self.label centerOnSuperview];
+    self.leftPathView.frame = CGRectMake(0, 0, self.label.left, self.height);
+    self.rightPathView.frame = CGRectMake(self.label.right, 0, self.width - self.label.right, self.height);
```
charlesmchen: I've added convenience methods to UIView that makes writing this kind of layout code more readable.

charlesmchen:
screen shot 2017-05-16 at 12 16 53 pm

charlesmchen:
> I'm afraid that enumerating every interaction in the conversation on every thread load is going to be unreasonably expensive. Have you measured performance when a conversation has many interactions?

> Sketching out an alternative approach:

> Build a new view that includes only unread messages, grouped by thread, sorted by timestamp.
Then we could always ask in constant time "What is the oldest unread message in this thread?"
Rather than having a new "unread indicator" interaction, use the information from 2. to decorate that interaction.
> For me the biggest unknown in this approach would be how easy/hard is it going to be get the layout to play nice with JSQMVC, but we can and have customized the message views before.

I had the same question and assumed we'd need to do something like that, but my testing so far was encouraging.  I tested on my slowest device (iPhone 5c) in a conversation with many messages and perf was reasonable.  Yes, we should test more thoroughly - I've made a card to do so.

Signal/src/UIView+OWS.m
```diff
@@ -131,6 +131,47 @@ - (void)setCompressionResistanceVerticalHigh
...
+}
+
+- (CGFloat)top
+{
+    return self.frame.origin.y;
+}
+
+- (CGFloat)bottom
+{
+    return self.frame.origin.x + self.frame.size.height;
```
charlesmchen: Good catch.

Signal/src/UIView+OWS.m
```diff
@@ -131,6 +131,47 @@ - (void)setCompressionResistanceVerticalHigh
...
+}
+
+- (CGFloat)height
+{
+    return self.frame.size.height;
+}
+
+- (void)centerOnSuperview
+{
+    self.frame = CGRectMake(
```
charlesmchen: Done.

charlesmchen:
PTAL @michaelkirk

PR #2129, opened 2017-05-17
Honor the local calling code in select recipient view.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/222

PR #2130, opened 2017-05-17
Charlesmchen/missing local notifications

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/environment/NotificationsManager.m
+++ b/Signal/src/environment/NotificationsManager.m
@@ -14,6 +14,7 @@
 #import <SignalServiceKit/TSErrorMessage.h>
 #import <SignalServiceKit/TSIncomingMessage.h>
 #import <SignalServiceKit/TextSecureKitEnv.h>
+#import <SignalServiceKit/Threading.h>
 
 @interface NotificationsManager ()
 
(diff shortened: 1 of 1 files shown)


Signal/src/environment/NotificationsManager.m
```diff
@@ -215,7 +216,7 @@ - (NotificationType)notificationPreviewType
 
 - (void)presentNotification:(UILocalNotification *)notification identifier:(NSString *)identifier
 {
-    dispatch_async(dispatch_get_main_queue(), ^{
+    DispatchMainThreadSafe(^{
         // Replace any existing notification
         // e.g. when an "Incoming Call" notification gets replaced with a "Missed Call" notification.
         if (self.currentNotifications[identifier]) {
```
charlesmchen: The next line is the bug: we call `cancelNotificationWithIdentifier` which is async and which will always cancel the notification we're trying to present. Bug only affects iOS 9.  Looks like on iOS 10 `dispatch_async(dispatch_get_main_queue...` will execute synchronously if you're already on the main thread.  I thought that same behavior was in earlier versions of iOS; I was wrong.

PR #2131, opened 2017-05-17
Charlesmchen/voice messages edge cases

charlesmchen:
Fixes multiple issues around voice messages.

// FREEBIE

### DIFF
--- a/Signal/src/ViewControllers/MessagesViewController.m
+++ b/Signal/src/ViewControllers/MessagesViewController.m
@@ -3210,6 +3210,8 @@ typedef enum : NSUInteger {
         DDLogInfo(@"Discarding voice message; too short.");
         self.audioRecorder = nil;
 
+        [self dismissKeyBoard];
+
         [OWSAlerts
             showAlertWithTitle:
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -3210,6 +3226,8 @@ - (void)endRecordingVoiceMemo
         DDLogInfo(@"Discarding voice message; too short.");
         self.audioRecorder = nil;
 
+        [self dismissKeyBoard];
```
charlesmchen: This alert can overlap the keyboard on small form factors.  I'm not sure if we need to solve problem in other places as well.

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -3573,6 +3592,17 @@ - (void)voiceMemoGestureDidStart
 
     DDLogInfo(@"voiceMemoGestureDidStart");
 
+    const CGFloat kIgnoreMessageSendDoubleTapDurationSeconds = 2.f;
+    if (self.lastMessageSentDate &&
```
michaelkirk: sneaking this one in! Nice! 😎

PR #2134, opened 2017-05-18
iOS8 contact handling 

michaelkirk:
Hide  Contacts.framework from ContactFetcher behind a ContactStoreAdapter and implement an iOS8 compatible ContactStoreAdaptee.

PTAL @charlesmchen


Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -33,15 +328,27 @@ class SystemContactsFetcher: NSObject {
...
+
+    override init() {
+        self.contactStoreAdapter = ContactStoreAdapter()
+    }
+
+    var supportsContactEditing: Bool {
+        return self.contactStoreAdapter.supportsContactEditing
+    }
+
+    private func setupObservationIfNecessary() {
```
michaelkirk: This is actually an ancillary change. We don't want to start observing on init (because the user may not have authorized access yet)

We want to start observing at "first use" which could be either on launch via `fetchIfAlreadyAuthorized` or a couple different places that potentially call `requestOnce`

Signal.xcodeproj/project.pbxproj
```diff
@@ -2534,7 +2534,7 @@
 					"\"$(SRCROOT)/Libraries\"/**",
 				);
 				INFOPLIST_FILE = "$(SRCROOT)/Signal/Signal-Info.plist";
-				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
```
charlesmchen: I think there's slightly more project changes to revert, but they're in my PR.

Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
```diff
@@ -677,6 +677,10 @@ - (void)showUpdateGroupView:(UpdateGroupMode)mode
 
 - (void)presentContactViewController
 {
+    if (!self.contactsManager.supportsContactEditing) {
+        DDLogWarn(@"%@ Contact editing not supported", self.tag);
+        return;
```
charlesmchen: `OWSAssert(0);`?

Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -6,6 +6,300 @@ import Foundation
...
+        // should only call once
+        assert(self.changeHandler == nil)
+        self.changeHandler = changeHandler
+
+        NotificationCenter.default.addObserver(self, selector: #selector(runChangeHandler), name: kAddressBookContactyStoreDidChangeNotificationName, object: nil)
+
+        let callback: ABExternalChangeCallback = { (_, _, _) in
+            // Ideally we'd just call the changeHandler here, but because this is a C style callback in swift, 
+            // we can't capture any state in the closure, so we use a notification as a trampoline
+            NotificationCenter.default.post(name: kAddressBookContactyStoreDidChangeNotificationName, object: nil)
```
charlesmchen: I think that's what the "context" is intended for, but I'm not objecting.

Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -6,6 +6,300 @@ import Foundation
...
+        ABAddressBookRegisterExternalChangeCallback(addressBook, callback, nil)
+    }
+
+    func requestAccess(completionHandler: @escaping (Bool, Error?) -> Void) {
+        ABAddressBookRequestAccessWithCompletion(addressBook, completionHandler)
+    }
+
+    func fetchContacts() -> Result<[Contact], Error> {
+        // Changes are not reflected unless we create a new address book
+        self.addressBook = ABAddressBookCreateWithOptions(nil, nil).takeRetainedValue()
```
charlesmchen: Isn't think redundant?  The `addressBook` is initialized where it is declared?

Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -6,6 +6,300 @@ import Foundation
...
+            }
+        }
+
+        return Contact(contactWithFirstName: firstName,
+                       andLastName: lastName,
+                       andUserTextPhoneNumbers: phoneNumbers,
+                       andImage: addressBookRecord.image,
+                       andContactID: addressBookRecord.recordId)
+    }
+
```
charlesmchen: nit: extra newline.

Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -6,6 +6,300 @@ import Foundation
...
+            if let companyName = addressBookRecord.companyName {
+                firstName = companyName
+            } else {
+                firstName = phoneNumbers.first
+            }
+        }
+
+        return Contact(contactWithFirstName: firstName,
+                       andLastName: lastName,
+                       andUserTextPhoneNumbers: phoneNumbers,
```
charlesmchen: So "work" and "home" phone number will be unlabeled on iOS 8? Fine.

Signal/src/contact/SystemContactsFetcher.swift
charlesmchen: Nice

Signal/src/contact/SystemContactsFetcher.swift
michaelkirk: What would the context be in this case? 

We can't capture any state in this closure per compiler restriction, so e.g. `self` or `callbackHandler` are off limits.

Signal/src/contact/SystemContactsFetcher.swift
michaelkirk: Yes, it looks redundant. I initially did not have this.

But then I saw that changes to the address book were not reflected in the app. I peaked at our old AB integration and saw we were doing something like this every time contacts were updated.

Hence the comment

Signal/src/contact/SystemContactsFetcher.swift
michaelkirk: Yup.

Signal/src/contact/SystemContactsFetcher.swift
michaelkirk: (The compiler restriction comes into play because we are passing a closure as a C function pointer) It's just a wart from writing this code in swift.

Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
michaelkirk: done

Signal/src/contact/SystemContactsFetcher.swift
michaelkirk: done 👀

michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/92466aaf8d2dae0aacb8e79572cb009fa96f301a

PR #2139, opened 2017-05-19
Cache display names for accounts.

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/SignalServiceKit/pull/225


Signal/src/environment/NotificationsManager.m
```diff
@@ -155,6 +154,11 @@ - (void)notifyUserForIncomingMessage:(TSIncomingMessage *)message
     }
 
     NSString *messageDescription = message.description;
+    NSString *senderName = [contactsManager displayNameForPhoneIdentifier:message.authorId];
+    NSString *groupName = [thread.name stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
+    if (groupName.length < 1) {
+        groupName = NSLocalizedString(@"UNNAMED_GROUP_NAME", @"Named used for groups without a name.");
```
charlesmchen: It's quite easy to create groups without a name, so I've made sure we handle that case here.

Signal/src/contact/OWSContactsManager.m
```diff
@@ -191,11 +194,44 @@ - (void)updateSignalAccounts
...
+
+    NSMutableDictionary<NSString *, NSString *> *accountNameMap = [NSMutableDictionary new];
+    for (SignalAccount *signalAccount in self.signalAccounts) {
+        NSString *displayName = [self displayNameForSignalAccount:signalAccount];
+        if (![displayName isEqualToString:signalAccount.recipientId]) {
+            accountNameMap[signalAccount.recipientId] = displayName;
+        }
+    }
+
+    dispatch_async(dispatch_get_main_queue(), ^{
```
michaelkirk: We're on the main thread, what's the motivation for dispatching async to the main queue?

Signal/src/environment/NotificationsManager.m
```diff
@@ -155,6 +154,11 @@ - (void)notifyUserForIncomingMessage:(TSIncomingMessage *)message
     }
 
     NSString *messageDescription = message.description;
+    NSString *senderName = [contactsManager displayNameForPhoneIdentifier:message.authorId];
+    NSString *groupName = [thread.name stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
```
michaelkirk: This name is a little confusing, because it will be populated with the contact's name for ContactThreads, right? 

Since we never want dangling white space in a thread name, maybe this whitespace trimming belongs at a higher level, since apparently we're not doing a good job of trimming group names when their set.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1255,6 +1255,9 @@
 /* Label text in device manager for a device with no name */
 "UNNAMED_DEVICE" = "Unnamed Device";
 
+/* Named used for groups without a name. */
+"UNNAMED_GROUP_NAME" = "Unnamed Group";
```
michaelkirk: Should we use this instead of "New Group" everywhere we display a group name, outside of the "group creation" view? Otherwise it might be confusing to see "Unnamed Group" in the notification which corresponds to some "New Group" in the inbox.

Signal/src/contact/OWSContactsManager.m
```diff
@@ -245,16 +281,23 @@ - (NSString *)unknownContactName
                              @"Displayed if for some reason we can't determine a contacts phone number *or* name");
 }
 
-- (NSString * _Nonnull)displayNameForPhoneIdentifier:(NSString * _Nullable)identifier {
-    if (!identifier) {
+- (NSString *_Nonnull)displayNameForPhoneIdentifier:(NSString *_Nullable)recipientId
+{
```
michaelkirk: Hopefully we're not using this in a transaction somewhere 😬

charlesmchen: I just rewrote this entire PR.  

* It'll now be transaction-safe; we no longer use transaction inside the "get name" methods.
* Results will now be consistent between "unattributed lookup by signal account" and "unattributed lookup by recipient id".
* Results will now be consistent between "attributed lookup by signal account" and "attributed lookup by recipient id".
* Results will now be more consistent between attributed and non-attributed names.
* There are small improvements around formatting of spaces, parentheses, etc.

Signal/src/environment/NotificationsManager.m
charlesmchen: > This name is a little confusing, because it will be populated with the contact's name for ContactThreads, right?

True, but we only use the groupName when displaying notifications for groups.

Signal/translations/en.lproj/Localizable.strings
charlesmchen: I'm going to revert this and keep using our existing default name "New Group" until we implement your idea of "Alice, Bob, etc."

charlesmchen:
PTAL @michaelkirk

Signal/src/contact/OWSContactsManager.m
```diff
@@ -188,14 +200,127 @@ - (void)updateSignalAccounts
...
+
+        self.cachedAccountNameMap = [cachedAccountNameMap copy];
+        self.cachedFirstNameMap = [cachedFirstNameMap copy];
+        self.cachedLastNameMap = [cachedLastNameMap copy];
+    });
+}
+
+- (NSString *_Nullable)cachedDisplayNameForRecipientId:(NSString *)recipientId
+{
+    OWSAssert(recipientId.lastPathComponent > 0);
```
michaelkirk: Why `lastPathComponent`? I was expecting `recipientId.length`

Signal/src/contact/OWSContactsManager.m
```diff
@@ -188,14 +200,127 @@ - (void)updateSignalAccounts
...
 }
 
+- (void)updateCachedDisplayNames
+{
+    OWSAssert([NSThread isMainThread]);
+
+    NSMutableDictionary<NSString *, NSString *> *cachedAccountNameMap = [NSMutableDictionary new];
+    NSMutableDictionary<NSString *, NSString *> *cachedFirstNameMap = [NSMutableDictionary new];
+    NSMutableDictionary<NSString *, NSString *> *cachedLastNameMap = [NSMutableDictionary new];
+    for (SignalAccount *signalAccount in self.signalAccounts) {
```
michaelkirk: If we're backing our displayNameForRecipientId with this cache, and we're only populating the cache with SignalAccount's then it seems like we'd bring back the bug where we don't show the name of ex-signal users who have unregistered.

Signal/src/environment/NotificationsManager.m
michaelkirk: fair enough!

Signal/src/contact/OWSContactsManager.m
charlesmchen: Fixed.

Signal/src/contact/OWSContactsManager.m
charlesmchen: Wow, bad auto-complete. Should have been `length`.

PR #2144, opened 2017-05-22
Add debug functions for sending media messages.

charlesmchen:
PTAL @michaelkirk


Signal/src/ViewControllers/DebugUI/DebugUIContacts.h
```diff
@@ -0,0 +1,15 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "OWSTableViewController.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface DebugUIContacts : NSObject
```
charlesmchen: I've broken out the debug code by topic.  Eventually, we'll want to break them out by topic in the debug UI itself.

PR #2146, opened 2017-05-23
Replace app settings segue.

charlesmchen:
A bunch of unrelated cleanup.

PTAL @michaelkirk


charlesmchen:
> Maybe take the opportunity to clean up the call-settings nag which also uses a segue to present the settings view controller. Or at least verify it still works - seems like it using a separate code path, but I didn't run it.

Done - and I went farther and:

* Cleaned up a few other transitions.
* Culled view controllers from the storyboards which we only instantiate programmatically now.

There's only a few places left where we still use the storyboards:

* Some of the segues during registration.
* Fingerprint view.
* Maybe one or two other views in app settings.

PR #2147, opened 2017-05-23
Fix caching of message bubble sizes.

charlesmchen:
The cache for message bubble sizes has used hash values which were often zero, causing collisions.  I've switched to hashing the underlying interaction's uniqueId.

PTAL @michaelkirk


Signal/src/Models/OWSDisplayedMessage.h
```diff
@@ -1,38 +0,0 @@
-//  Created by Dylan Bourgeois on 29/11/14.
-//  Copyright (c) 2014 Hexed Bits. All rights reserved.
-//  Portions Copyright (c) 2016 Open Whisper Systems. All rights reserved.
-
-#import "JSQMessageData.h"
```
charlesmchen: This class wasn't used anywhere.

Signal/src/Models/TSMessageAdapaters/TSMessageAdapter.m
```diff
@@ -77,9 +77,8 @@ - (instancetype)initWithInteraction:(TSInteraction *)interaction
 
     _interaction = interaction;
     _messageDate = interaction.date;
-    // TODO casting a string to an integer? At least need a comment here explaining why we are doing this.
```
michaelkirk: ah yes, I remember that TODO. =(

Signal/src/Models/TSMessageAdapaters/TSMessageAdapter.m
```diff
@@ -354,11 +353,9 @@ - (NSString *)text {
 
 - (NSUInteger)messageHash
 {
-    if (self.isMediaMessage) {
-        return [self.mediaItem mediaHash];
```
michaelkirk: Is this going to properly invalidate the bubble size when a media message changes state?

e.g. when it goes from from "downloading" -> "downloaded"?

charlesmchen: Great point. Fixed.

PR #2149, opened 2017-05-24
Rename audio duration and image size methods in TSAttachmentStream.

charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/228

PTAL @michaelkirk


charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/229

PR #2153, opened 2017-05-25
Fetch safety number upon entering thread

michaelkirk:
see also: https://github.com/WhisperSystems/SignalServiceKit/pull/231

Incidentally:
* uniform method to get recipients in group/contact thread
  * use it to clean up markAsSeenJob

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ProfileFetcherJob.swift
+++ b/Signal/src/ProfileFetcherJob.swift
@@ -1,0 +1,113 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+@objc
+class ProfileFetcherJob: NSObject {
+
+    let TAG = "[ProfileFetcherJob]"
+
+    let networkManager: TSNetworkManager
+    let storageManager: TSStorageManager
+
+    let thread: TSThread
+
+    public class func run(thread: TSThread, networkManager: TSNetworkManager) {
+        ProfileFetcherJob(thread: thread, networkManager: networkManager).run()
+    }
+
+    init(thread: TSThread, networkManager: TSNetworkManager) {
+        self.networkManager = networkManager
+        self.storageManager = TSStorageManager.shared()
+
+        self.thread = thread
+    }
+
+    public func run() {
+        for recipientId in self.thread.recipientIdentifiers {
+            let request = OWSGetProfileRequest(recipientId: recipientId)
+
+            self.networkManager.makeRequest(
+                request,
+                success: { (_: URLSessionDataTask?, responseObject: Any?) -> Void in
+                    guard let profileResponse = SignalServiceProfile(recipientId: recipientId, rawResponse: responseObject) else {
+                        Logger.error("\(self.TAG) response object had unexpected content")
+                        assertionFailure("\(self.TAG) response object had unexpected content")
+                        return
+                    }
+
+                    self.processResponse(signalServiceProfile: profileResponse)
+            },
+                failure: { (_: URLSessionDataTask?, error: Error?) in
+                    guard let error = error else {
+                        Logger.error("\(self.TAG) error in \(#function) was surpringly nil. sheesh rough day.")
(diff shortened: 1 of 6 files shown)


Signal/src/ProfileFetcherJob.swift
```diff
@@ -0,0 +1,113 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+@objc
+class ProfileFetcherJob: NSObject {
```
charlesmchen: I'm curious why you made this a `NSObject`.

charlesmchen: To be clear, as a Swift neophyte my understanding is that the Swift-y thing to do is to not make classes NSObjects unless you know you're going to retain references to that class in Obj-C.

Signal/src/ProfileFetcherJob.swift
```diff
@@ -0,0 +1,113 @@
...
+                Logger.info("\(self.TAG) updated identity key in fetched profile for recipient: \(recipientId)")
+                self.storageManager.deleteAllSessions(forContact: recipientId)
+            } else {
+                // no change in identity.
+            }
+        }
+    }
+}
+
+struct SignalServiceProfile {
```
charlesmchen: I'm wondering if `SignalServiceProfileResponse` would be a better class name?

Signal/src/ProfileFetcherJob.swift
```diff
@@ -0,0 +1,113 @@
...
+    let TAG = "[ProfileResponse]"
```
charlesmchen: Shouldn't the TAG always agree with the class name?

Signal/src/ProfileFetcherJob.swift
charlesmchen: `OWSProfileResponse`?

Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -1212,6 +1212,7 @@ - (void)viewDidAppear:(BOOL)animated {
     }
     [self updateNavigationBarSubtitleLabel];
     [MarkIdentityAsSeenJob runWithThread:self.thread];
+    [ProfileFetcherJob runWithThread:self.thread networkManager:self.networkManager];
```
charlesmchen: Brainstorming - would we also want do this...

* ...When we add a member to a group?
* ...When we "export with Signal" to a thread? Or I guess that's redundant, since we'll learn of the identity change when we do the send?

charlesmchen: Not objecting, but is "viewDidAppear" the right place?  I think it means we'll re-fetch whenever we close conversation settings or the app returns from the background in this conversation.  The latter seems right but maybe not the former?  But the former should be rare so... fine.

Oh, the former covers the "...When we add a member to a group?" case.  Makes sense.

Signal/src/ProfileFetcherJob.swift
michaelkirk: Ha. You see I too struggled and iterated with the name.

Signal/src/ProfileFetcherJob.swift
michaelkirk: not just retained from objc, but any interaction from objc requires being marked as @objc and descending from NSObject. I believe under the hood a little, this has to do with being compatible with dynamic dispatch.

Signal/src/ViewControllers/MessagesViewController.m
michaelkirk: I made this choice mainly to account for the fact that you might not leave a thread for a long time. 

e.g. you may background the app and come back to the conversation. 

Maybe I should move it to viewDidLoad and then again on UIApplicationWillEnterForegroundNotification

> ...When we add a member to a group?

That requires sending a message to that group member, so any SN changes have to be resolved *before* they can be added. Not really addressed by this IMO.

charlesmchen: Current arrangement sounds good, thanks.

PR #2158, opened 2017-05-27
Tweak appearance of registration views.

charlesmchen:
PTAL @michaelkirk


Signal/src/AppDelegate.m
```diff
@@ -122,6 +122,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
         RegistrationViewController *viewController = [RegistrationViewController new];
         UINavigationController *navigationController =
             [[UINavigationController alloc] initWithRootViewController:viewController];
+        navigationController.navigationBarHidden = YES;
```
charlesmchen: This got lost in the conversion from storyboard.

PR #2159, opened 2017-05-27
Hide the socket status view when censorship circumvention is active.

charlesmchen:
PTAL @michaelkirk 

Part two would be to show an indicator that censorship circumvention is active.

### DIFF
--- a/Signal/src/ViewControllers/SignalsNavigationController.h
+++ b/Signal/src/ViewControllers/SignalsNavigationController.h
@@ -1,15 +1,9 @@
 //
-//  SignalsNavigationController.h
-//  Signal
-//
-//  Created by Dylan Bourgeois on 18/11/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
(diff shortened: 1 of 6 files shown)


Signal/src/ViewControllers/SignalsNavigationController.h
```diff
@@ -1,15 +1,9 @@
 //
-//  SignalsNavigationController.h
-//  Signal
-//
-//  Created by Dylan Bourgeois on 18/11/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
 #import <UIKit/UIKit.h>
```
michaelkirk: since you're cleaning up, UIKit is in the prefix header ;)

michaelkirk:
> Part two would be to show an indicator that censorship circumvention is active.

Apparently I couldn't be bothered to read your 1 sentence PR description. Sorry. LGTM

charlesmchen:
> Apparently I couldn't be bothered to read your 1 sentence PR description. Sorry. LGTM

Yup. My take was that doing part two (a censorship circumvention indicator) would be non-trivial and we didn't want to wait until part two to do part one.

PR #2160, opened 2017-05-27
Interstitial SN confirmation alert

michaelkirk:
Following the BlockUI utils pattern, I attempted to catch any point in the app that would result in sending to a known changed safety number, and instead of failing with a retry dialog, you present a "confirm and continue"

This now works for:

- message sending
- attachments (including voice notes)
- paper-clip picker
- calls
- add to group (new group and group edit, both by picking and phone number)

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/SafetyNumberConfirmationAlert.swift
+++ b/Signal/src/ViewControllers/SafetyNumberConfirmationAlert.swift
@@ -1,0 +1,119 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+class SafetyNumberConfirmationAlert: NSObject {
+
+    let TAG = "[SafetyNumberConfirmationAlert]"
+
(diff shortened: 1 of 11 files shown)


Signal/src/ViewControllers/SafetyNumberConfirmationAlert.swift
```diff
@@ -0,0 +1,119 @@
...
+                                                   theirDisplayName: displayName,
+                                                   completion: { completion(false) })
+
+        }
+        actionSheetController.addAction(showSafetyNumberAction)
+
+        let dismissAction = UIAlertAction(title: NSLocalizedString("TXT_CANCEL_TITLE", comment: "generic cancel text"), style: .cancel)
+        actionSheetController.addAction(dismissAction)
+
+        fromViewController.present(actionSheetController, animated: true)
```
charlesmchen: If you use `UIApplication.shared.frontmostViewController?.present(...)` you can remove a parameter. As you like.

PR #2163, opened 2017-05-30
Reset the unread indicator state if possible while app is in the back…

charlesmchen:
…ground.

* Clean up timer usage.

Fixes an edge case around the unread indicator.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/OWSLinkedDevicesTableViewController.m
+++ b/Signal/src/ViewControllers/OWSLinkedDevicesTableViewController.m
@@ -6,6 +6,7 @@
 #import "OWSDeviceTableViewCell.h"
 #import "OWSLinkDeviceViewController.h"
 #import "UIViewController+CameraPermissions.h"
+#import <SignalServiceKit/NSTimer+OWS.h>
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/OWSLinkedDevicesTableViewController.m
```diff
@@ -100,11 +101,11 @@ - (void)expectMoreDevices
     self.editing = NO;
 
     __weak typeof(self) wself = self;
-    self.pollingRefreshTimer = [NSTimer scheduledTimerWithTimeInterval:(10.0)
-                                                                target:wself
-                                                              selector:@selector(refreshDevices)
-                                                              userInfo:nil
-                                                               repeats:YES];
+    [self.pollingRefreshTimer invalidate];
+    self.pollingRefreshTimer = [NSTimer weakScheduledTimerWithTimeInterval:(10.0)target:wself
```
michaelkirk: nit: format parameters 1 per line here.

charlesmchen: Strangely enough clang-format is insisting.

PR #2165, opened 2017-05-30
Fix issues around initial layout of messages view.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessagesViewController.m
+++ b/Signal/src/ViewControllers/MessagesViewController.m
@@ -911,23 +911,6 @@ typedef enum : NSUInteger {
     [self initializeToolbars];
 }
 
-- (void)viewDidLayoutSubviews
-{
-    [super viewDidLayoutSubviews];
-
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/MessagesViewController.m
```diff
@@ -853,23 +853,6 @@ - (void)viewDidLoad
...
-- (void)viewDidLayoutSubviews
-{
-    [super viewDidLayoutSubviews];
-
-    // JSQMVC width is initially 375px on iphone6/ios9 (as specified by the xib), which causes
-    // our initial bubble calculations to be off since they happen before the containing
-    // view is layed out. https://github.com/jessesquires/JSQMessagesViewController/issues/1257
-    if (CGRectGetWidth(self.collectionView.frame) != self.previousCollectionViewFrameWidth) {
-        // save frame value from next comparison
-        self.previousCollectionViewFrameWidth = CGRectGetWidth(self.collectionView.frame);
```
michaelkirk: Maybe we can now get rid of previousCollectionViewFrameWidth property?

PR #2171, opened 2017-06-01
Handle incoming/outgoing calls from an unseen changed identity

michaelkirk:
- Reject incoming calls

- Confirm outgoing calls with unseen identity changes (e.g. from the contacts app)

screen shot 2017-05-31 at 5 00 06 pm

PTAL @charlesmchen


Signal/src/Jobs/MarkIdentityAsSeenJob.swift
```diff
@@ -8,18 +8,24 @@ import Foundation
...
 
-    init(thread: TSThread) {
-        self.thread = thread
+    init(recipientIds: [String]) {
+        self.recipientIds = recipientIds
     }
 
     public func run() {
-        for recipientId in self.thread.recipientIdentifiers {
+        for recipientId in self.recipientIds {
```
charlesmchen: I wonder if we should modify all of the "jobs" to dispatch async to the global worker thread to ensure this kind of thing is never done on the main thread.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1258,6 +1267,9 @@
 /* Action sheet item */
 "SHOW_SAFETY_NUMBER_ACTION" = "Show new safety number";
```
charlesmchen: Should we make sure "Safety Number" is always capitalized throughout this file?

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1258,6 +1267,9 @@
...
 
+/* notification action */
+"SHOW_THREAD_BUTTON_TITLE" = "Show Thread";
```
charlesmchen: I suggest we use "thread" only in the code, and exclusively use "conversation" in our copy.

Signal/src/ViewControllers/DebugUI/DebugUITableViewController.m
```diff
@@ -84,6 +84,51 @@ + (void)presentDebugUIForThread:(TSThread *)thread fromViewController:(UIViewCon
...
+                                                dispatch_after(dispatch_time(DISPATCH_TIME_NOW,
+                                                                   (int64_t)(notificationDelay * NSEC_PER_SEC)),
+                                                    dispatch_get_main_queue(),
+                                                    ^{
+                                                        [[Environment getCurrent].callService.notificationsAdapter
+                                                            presentRejectedCallWithUnseenIdentityChange:call
+                                                                                             callerName:thread.name];
+                                                    });
+                                            }],
+                                  ]]];
```
charlesmchen: Nice.

charlesmchen: I intend to do the work to make this UI hierarchical sometime this month.

Signal/src/ViewControllers/SafetyNumberConfirmationAlert.swift
```diff
@@ -87,33 +95,31 @@ class SafetyNumberConfirmationAlert: NSObject {
         let dismissAction = UIAlertAction(title: NSLocalizedString("TXT_CANCEL_TITLE", comment: "generic cancel text"), style: .cancel)
         actionSheetController.addAction(dismissAction)
 
-        fromViewController.present(actionSheetController, animated: true)
+        UIApplication.shared.frontmostViewController?.present(actionSheetController, animated: true)
```
charlesmchen: NBD but I've been wondering if we should have an `OWSAlerts.present(alert, animated)` method to DRY up a tiny bit.

Signal/src/network/PushManager.h
```diff
@@ -10,22 +10,25 @@ NS_ASSUME_NONNULL_BEGIN
 
 @class UILocalNotification;
 
-#define Signal_Thread_UserInfo_Key @"Signal_Thread_Id"
-#define Signal_Message_UserInfo_Key @"Signal_Message_Id"
+FOUNDATION_EXPORT NSString *const Signal_Thread_UserInfo_Key;
+FOUNDATION_EXPORT NSString *const Signal_Message_UserInfo_Key;
```
charlesmchen: Why are we using `FOUNDATION_EXPORT`?

Signal/src/network/PushManager.m
```diff
@@ -379,12 +402,38 @@ - (UIUserNotificationCategory *)signalMissedCallCategory
...
+    UIMutableUserNotificationCategory *missedCallCategory = [UIMutableUserNotificationCategory new];
+    missedCallCategory.identifier = PushManagerCategoriesMissedCall;
+    [missedCallCategory setActions:@[ callBackAction ] forContext:UIUserNotificationActionContextMinimal];
+    [missedCallCategory setActions:@[ callBackAction ] forContext:UIUserNotificationActionContextDefault];
 
-    return callCategory;
+    return missedCallCategory;
+}
+
+- (UIUserNotificationCategory *)signalRejectedCallWithUnseenIdentityChangeCategory
```
charlesmchen: Is this redundant with logic in `UserNotificationsAdaptee`?

Signal/src/network/PushManager.h
michaelkirk: I don't feel strongly about it. I guess you prefer extern?

Signal/src/network/PushManager.m
michaelkirk: Well - sort of. 

UserNotificationsAdaptee is currently unused code, but it's iOS10 only. I've been trying to keep it up do date as we add new actions so that the eventual transition to UserNotifications is easier. But even then, we'll still need to have this pre iOS10 implementation.

Signal/translations/en.lproj/Localizable.strings
michaelkirk: I think so? And generally the hig wants buttons to be title cased.

Signal/translations/en.lproj/Localizable.strings
michaelkirk: Good call.

Signal/translations/en.lproj/Localizable.strings
michaelkirk: I've actually gone the other direction. We mostly lowercase it, as do the other clients.

The exception is button text which we always (should) title case.

Signal/src/Jobs/MarkIdentityAsSeenJob.swift
michaelkirk: As written this job does run sync to allow inline "confirm and continue" type behavior. This job potentially entails some db writes, shouldn't take too long.

Maybe some kind of naming convention like `run`/`runAsync` is sufficient? If you agree I'll go through the existing jobs to make sure that convention is honored.

WDYT @charlesmchen?

Signal/src/network/PushManager.h
michaelkirk: I've changed these to externs

PR #2172, opened 2017-06-01
Charlesmchen/paste vs multiple uti types

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/SignalAttachment.swift
+++ b/Signal/src/ViewControllers/SignalAttachment.swift
@@ -308,6 +308,10 @@ class SignalAttachment: NSObject {
             return false
         }
         let pasteboardUTISet = Set<String>(pasteboardUTITypes[0])
+        
+        if pasteboardUTISet.contains(kUTTypeURL as String) {
+            return true
+        }
 
         // The pasteboard can be populated with multiple UTI types
         // with different payloads.  iMessage for example will copy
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -279,16 +279,6 @@ class SignalAttachment: NSObject {
...
-            kUTTypeText as String,
-            kUTTypePlainText as String,
-            kUTTypeUTF8PlainText as String,
-            kUTTypeUTF16PlainText as String,
-            kUTTypeURL as String
-        ]
-    }
-
     public var isImage: Bool {
         return SignalAttachment.outputImageUTISet.contains(dataUTI)
```
charlesmchen: We could also use `UTTypeConformsTo()` elsewhere. I'm not sure it would be so valuable.

Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -279,16 +279,6 @@ class SignalAttachment: NSObject {
         return MIMETypeUtil.supportedAudioUTITypes()
     }
 
-    public class var textUTISet: Set<String> {
-        return [
-            kUTTypeText as String,
-            kUTTypePlainText as String,
-            kUTTypeUTF8PlainText as String,
-            kUTTypeUTF16PlainText as String,
-            kUTTypeURL as String
```
michaelkirk: IIRC pasting a URL was considered an attachment before, just wanted to make sure you tested that pasting attachments still worked with the new "conforms" check.

charlesmchen: Good catch!  Fixed.

PR #2176, opened 2017-06-02
only run migration once

michaelkirk:
PTAL @charlesmchen


michaelkirk:
PTAL @charlesmchen, this one's not controversial. =)

PR #2177, opened 2017-06-02
Mkirk/block external send

michaelkirk:
Disable reply-from-lockscreen and external file share with unseen identities.

PTAL @charlesmchen 

uses https://github.com/WhisperSystems/SignalServiceKit/pull/243


michaelkirk:
closing for now.

PR #2178, opened 2017-06-04
Fixed Icons + Optimized PNGs [FREEBIE]

majouji:
• Remade some of the PNG assets that had pixel clipping to get them to perfectly fit on the pixel grid
• Optimized PNG assets with ImageOptim (see attachment below)
• Added @1x assets for error_white and warning_white


majouji:
screen shot 2017-06-04 at 2 26 48 pm

dermaaarkus:
Good work! I guess i won't be merged if you do not solve the errored checks, right?

majouji:
Honestly not sure why one of the checks failed since all I did was crunch static asset… kinda puzzled.

michaelkirk:
Tests are broken at the moment, but they should be fixed soon.

Also, @majouji at the time you opened this PR we revamped a lot of our icons for the new status message style that came out in 2.13. I wonder how many of these are still relevant?

majouji:
@michaelkirk You’re probably right. In that case maybe think about a way to optimize assets automatically at build time with SVGO and ImageOptim or something?

PR #2182, opened 2017-06-06
Indicate if censorship circumvention is active in the main app settin…

charlesmchen:
…gs view.

PTAL @michaelkirk .
screen shot 2017-06-06 at 10 39 29 am

### DIFF
--- a/Signal/translations/en.lproj/Localizable.strings
+++ b/Signal/translations/en.lproj/Localizable.strings
@@ -792,6 +792,9 @@
 /* No comment provided by engineer. */
 "NETWORK_ERROR_RECOVERY" = "Please check you're online and try again.";
 
+/* Indicates to the user that censorship circumvention has been activated. */
+"NETWORK_STATUS_CENSORSHIP_CIRCUMVENTION_ACTIVE" = "Censorship Circumvention: On";
+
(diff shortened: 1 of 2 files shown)


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -783,6 +783,9 @@
 /* No comment provided by engineer. */
 "NETWORK_ERROR_RECOVERY" = "Please check you're online and try again.";
 
+/* Indicates to the user that censorship circumvention has been activated. */
+"NETWORK_STATUS_CENSORSHIP_CIRCUMVENTION_ACTIVE" = "Censorship Circumvention: On";
```
charlesmchen: I'm not wild about the copy but as you can see in the screenshot it was hard to find copy that would fit.

NicolasCGN: what about 'bypass censorship'?

charlesmchen: @NicolasCGN Hmm, well it's not a button, it's a state indicator.  Also I'd like to be consistent and unfortunately the language we're using elsewhere is "censorship circumvention".

PR #2183, opened 2017-06-06
identity manager

michaelkirk:
uses https://github.com/WhisperSystems/SignalServiceKit/pull/248

PTAL @charlesmchen


Signal/src/network/PushManager.m
```diff
@@ -216,9 +216,6 @@ - (void)application:(UIApplication *)application
             return;
         }
 
-        TSContactThread *thread = [TSContactThread getOrCreateThreadWithContactId:recipientId];
-        [MarkIdentityAsSeenJob runWithThread:thread];
```
michaelkirk: 1. We should either do something like`OWSIdentityManager.shared().setVerificationState(.default, identityKey: noLongerVerifiedIdentity.identityKey, recipientId: noLongerVerifiedIdentity.recipientId, sendSyncMessage: true)

2. Or maybe just remove the option to call back from the lockscreen in this case, since we don't have a reference to the identity here.

I'm leaning towards 2.

charlesmchen: Leaning towards 2 too.

michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/34f59d661ab28993553cfeccf5e4e4d1198bf3d9

PR #2184, opened 2017-06-06
identity manager

michaelkirk:
You *can* callback users with changed identities if they were not previously verified.

uses: https://github.com/WhisperSystems/SignalServiceKit/pull/249
Intended to supercede: https://github.com/WhisperSystems/Signal-iOS/pull/2183


Signal/src/UserInterface/Strings.swift
```diff
@@ -15,16 +15,13 @@ import Foundation
 
     // MARK: Notification actions
     static let callBackButtonTitle = NSLocalizedString("CALLBACK_BUTTON_TITLE", comment: "notification action")
-    static let confirmIdentityAndCallBackButtonTitle = NSLocalizedString("CONFIRM_IDENTITY_AND_CALLBACK_BUTTON_TITLE", comment: "notification action, confirming that it's OK to proceed calling after a caller's Safety Number has changed")
     static let showThreadButtonTitle = NSLocalizedString("SHOW_THREAD_BUTTON_TITLE", comment: "notification action")
 
     // MARK: Missed Call Notification
-    static let missedCallNotificationBody = NSLocalizedString("MISSED_CALL", comment: "notification title")
-    static let missedCallNotificationBodyWithCallerName = NSLocalizedString("MSGVIEW_MISSED_CALL_WITH_NAME", comment: "notification title. Embeds {{Caller's Name}}")
```
michaelkirk: Got rid of a little duplication here. I didn't think it was worth distinguishing between copy which used *no* contact name, and that which used the "Signal User" generic name.

Signal/src/UserInterface/Strings.swift
```diff
@@ -15,16 +15,13 @@ import Foundation
```
charlesmchen: Can we rename this class to `Strings` or rename the source file to `CallStrings.swift`?

charlesmchen:
> Intended to supercede: #2183

I don't understand - aren't there changes in #2183 that aren't in this PR that we still need?

michaelkirk:
>> Intended to supercede: #2183
> I don't understand - aren't there changes in #2183 that aren't in this PR that we still need?

Note that this PR didn't target master, it targets #2183, so that you can see the new diff.

When I merge it I'll merge 2184 directly to master and close 2183.

Signal/src/UserInterface/Strings.swift
michaelkirk: IMO we should try to duplicate strings less and move all common usage of a string to this file, in which case a namespace like CallStrings is arguably useful. We just haven't yet de-duped anything besides call strings.

charlesmchen: Oh, so you see multiple classes being defined in this single source file. Got it.

michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/34f59d661ab28993553cfeccf5e4e4d1198bf3d9

PR #2188, opened 2017-06-07
block file sharing to no-longer-verified identity

michaelkirk:
I first tried and failed to deduplicate by integrating these checks into ThreadUtil, but decided against it after following the consequences of the many presentation context in which the ThreadUtil methods are used

e.g. sometimes we need to dismiss a keyboard, sometimes we need to dismiss a view. 

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/SendExternalFileViewController.m
+++ b/Signal/src/ViewControllers/SendExternalFileViewController.m
@@ -15,6 +15,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface SendExternalFileViewController () <SelectThreadViewControllerDelegate>
 
+@property (nonatomic, readonly) OWSContactsManager *contactsManager;
 @property (nonatomic, readonly) OWSMessageSender *messageSender;
 
 @end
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/SendExternalFileViewController.m
```diff
@@ -47,6 +49,19 @@ - (void)threadWasSelected:(TSThread *)thread
     OWSAssert(self.attachment);
     OWSAssert(thread);
 
+    BOOL didShowSNAlert =
+        [SafetyNumberConfirmationAlert presentAlertIfNecessaryWithRecipientIds:thread.recipientIdentifiers
+                                                              confirmationText:[SafetyNumberStrings confirmSendButton]
+                                                               contactsManager:self.contactsManager
+                                                                    completion:^(BOOL didConfirm) {
+                                                                        if (didConfirm) {
+                                                                            [self threadWasSelected:thread];
```
charlesmchen: weakself.

michaelkirk: done.

PR #2189, opened 2017-06-07
Fix paste of text fragments.

charlesmchen:
Taking a third run at this.  I keep finding unexpected behavior in the pasteboard.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/SignalAttachment.swift
+++ b/Signal/src/ViewControllers/SignalAttachment.swift
@@ -279,6 +279,12 @@ class SignalAttachment: NSObject {
         return MIMETypeUtil.supportedAudioUTITypes()
     }
 
+    // Returns the set of UTIs that correspond to valid image, video and audio formats
+    // for Signal attachments.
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -332,7 +338,7 @@ class SignalAttachment: NSObject {
         for utiType in pasteboardUTISet {
             if UTTypeConformsTo(utiType as CFString, kUTTypeText) {
                 hasTextUTIType = true
-            } else {
+            } else if mediaUTISet.contains(utiType) {
                 hasNonTextUTIType = true
```
michaelkirk: take it or leave it: would it be simpler to just short circuit and return false here, getting rid of the `hasNonTextUTIType` var?

charlesmchen: I thought about it but I thought this was slightly more readable and I keep coming back in here and doing logging.

PR #2192, opened 2017-06-08
Show no redundant error when failing to send due to changed identity

michaelkirk:
The old "blocking" style SN alerts are deprecated. We maintain the
functionality to decrypt incoming messages from legacy SN alerts so that
users have time to recover their undecrypted messages.

uses: https://github.com/WhisperSystems/SignalServiceKit/pull/254

PTAL @charlesmchen


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -704,7 +704,7 @@
 "MESSAGE_STATUS_DELIVERED" = "Delivered";
 
 /* message footer for failed messages */
-"MESSAGE_STATUS_FAILED" = "Sending failed. Tap to retry.";
+"MESSAGE_STATUS_FAILED" = "Sending failed. Tap for info.";
```
michaelkirk: WDYT of this copy change @charlesmchen / @riyapenn?

I didn't want people to hesitate to tap, thinking it would automatically retry. Also might help surface that you can actually get more info by tapping.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -518,7 +518,7 @@
 "FAILED_SENDING_BECAUSE_RATE_LIMIT" = "Too many failures with this contact. Please try again shortly.";
 
 /* action sheet header when re-sending message which failed because of untrusted identity keys */
-"FAILED_SENDING_BECAUSE_UNTRUSTED_IDENTITY_KEY" = "You must accept the new safety number before you will be able to send.";
+"FAILED_SENDING_BECAUSE_UNTRUSTED_IDENTITY_KEY" = "Recipient's safety number recently changed. You may wish to verify before resending.";
```
charlesmchen: "Recipient's safety number _has_ recently changed."?

Also I wonder if we should format this to replace `Recipient` with the "display name" of the recipient (especially in groups).

Signal/translations/en.lproj/Localizable.strings
charlesmchen: It's better!

Signal/translations/en.lproj/Localizable.strings
michaelkirk: I've updated this to "Your safety number with %@ has recently changed. You may wish to verify before resending."

michaelkirk:
in 7335fc4

Signal/translations/en.lproj/Localizable.strings
charlesmchen: I like to end the string's key with _FORMAT as a reminder to format the string anywhere it is used.

PR #2193, opened 2017-06-08
Prefer archiving, to deleting old sessions for better handling of out of order decryption.

michaelkirk:
PTAL @charlesmchen 

See also:

SSK: https://github.com/WhisperSystems/SignalServiceKit/pull/255
SPK: https://github.com/WhisperSystems/SignalProtocolKit/pull/27


Signal/src/ProfileFetcherJob.swift
```diff
@@ -68,8 +68,8 @@ class ProfileFetcherJob: NSObject {
             }
 
             if OWSIdentityManager.shared().saveRemoteIdentity(latestIdentityKey, recipientId: recipientId) {
-                Logger.info("\(self.TAG) updated identity key in fetched profile for recipient: \(recipientId)")
-                self.storageManager.deleteAllSessions(forContact: recipientId)
```
charlesmchen: Can we get rid of `deleteAllSessions`?

Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
```diff
@@ -49,13 +49,20 @@ + (OWSTableSection *)sectionForContactThread:(TSContactThread *)contactThread
                                        actionBlock:^{
                                            [self presentVerificationStatePickerForContactThread:contactThread];
                                        }],
-                       [OWSTableItem itemWithTitle:@"Delete session"
+                       [OWSTableItem itemWithTitle:@"Delete all sessions"
                                        actionBlock:^{
                                            dispatch_async([OWSDispatch sessionStoreQueue], ^{
                                                [[TSStorageManager sharedManager]
                                                    deleteAllSessionsForContact:contactThread.contactIdentifier];
```
charlesmchen: If we're only going to use deleteAllSessionsForContact for debugging perhaps we can `#ifdef DEBUG`?

PR #2197, opened 2017-06-09
Charlesmchen/fingerprint view3

charlesmchen:
* DRY up safety number success and failure handling.
* Add “mark as verified” option to success case.
* Add “learn more” link. 
* Move “scan” button into QR code.
* Show verification state in fingerprint view.
* Let user change verification state.

PTAL @michaelkirk 

### iPhone 7+

img_1482
img_1481

### iPhone SE

screen shot 2017-06-09 at 11 29 09 am
screen shot 2017-06-09 at 11 29 05 am


Signal/src/ViewControllers/FingerprintViewController.m
```diff
@@ -97,6 +129,7 @@ - (void)configureWithRecipientId:(NSString *)recipientId
     OWSRecipientIdentity *_Nullable recipientIdentity =
         [[OWSIdentityManager sharedManager] recipientIdentityForRecipientId:recipientId];
     OWSAssert(recipientIdentity);
+    self.identityKey = recipientIdentity.identityKey;
```
charlesmchen: By capturing the identity key when we enter these views, we prevent the edge case where the user verifies a key that we learned about while this view was open.

charlesmchen: I've added a comment to that effect.

Signal/src/ViewControllers/FingerprintViewController.m
```diff
@@ -346,10 +425,13 @@ - (void)showScanner
...
 }
 
-- (void)scanButtonTapped:(UIGestureRecognizer *)gestureRecognizer
+- (void)learnMoreButtonTapped:(UIGestureRecognizer *)gestureRecognizer
 {
     if (gestureRecognizer.state == UIGestureRecognizerStateRecognized) {
-        [self showScanner];
+        NSString *learnMoreURL = @"https://support.whispersystems.org/hc/en-us/articles/"
+                                 @"213134107-How-do-I-verify-the-person-I-m-sending-messages-to-is-who-they-say-they-"
+                                 @"are-";
```
charlesmchen: Do we link directly to support articles or use redirection?

charlesmchen: Riya says remove everything after the numbers. Done.

PR #2199, opened 2017-06-09
Show verification state banner. Show verification state in conversati…

charlesmchen:
…on settings view.

* Show verification state banner. 
* Show verification state in conversation settings view.
* Check for identity key before presenting fingerprint view.
* Show verification state in a separate subtitle in conversation view.
* Let users verify from group members view.

PTAL @michaelkirk 

### Let users verify from group members view.

screen shot 2017-06-09 at 4 55 06 pm

### Show verification state in conversation settings view (contact has name).

screen shot 2017-06-09 at 4 52 22 pm

### Show verification state in conversation settings view (contact has no name).

screen shot 2017-06-09 at 4 16 23 pm

### Verification banner (1:1 conversation).

screen shot 2017-06-09 at 4 07 40 pm

### Verification banner (group, 1 no longer verified group member).

screen shot 2017-06-09 at 4 08 06 pm

### Verification banner (group, more than one no longer verified group member).

screen shot 2017-06-09 at 4 07 33 pm

### Show verification state in group members view.

screen shot 2017-06-09 at 4 07 24 pm

Note: Some of the copy changed after I took these screenshots.

### DIFF
--- a/Signal/src/ViewControllers/ContactsViewHelper.m
+++ b/Signal/src/ViewControllers/ContactsViewHelper.m
@@ -79,18 +79,18 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)signalAccountsDidChange:(NSNotification *)notification
 {
-    dispatch_async(dispatch_get_main_queue(), ^{
-        [self updateContacts];
-    });
+    OWSAssert([NSThread isMainThread]);
+
+    [self updateContacts];
 }
 
 - (void)blockedPhoneNumbersDidChange:(id)notification
 {
-    dispatch_async(dispatch_get_main_queue(), ^{
-        self.blockedPhoneNumbers = [_blockingManager blockedPhoneNumbers];
+    OWSAssert([NSThread isMainThread]);
 
-        [self updateContacts];
-    });
+    self.blockedPhoneNumbers = [_blockingManager blockedPhoneNumbers];
+
+    [self updateContacts];
 }
 
 #pragma mark - Contacts
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -1273,11 +1273,7 @@ typedef enum : NSUInteger {
     // return from FingerprintViewController.
     [self dismissKeyBoard];
 
-    FingerprintViewController *fingerprintViewController = [FingerprintViewController new];
-    [fingerprintViewController configureWithRecipientId:recipientId];
-    UINavigationController *navigationController =
-        [[UINavigationController alloc] initWithRootViewController:fingerprintViewController];
-    [self presentViewController:navigationController animated:YES completion:nil];
+    [FingerprintViewController showVerificationViewFromViewController:self recipientId:recipientId];
 }
 
 #pragma mark - Calls
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -182,7 +182,7 @@ typedef enum : NSUInteger {
 @property (nonatomic) UILabel *navigationBarTitleLabel;
 @property (nonatomic) UILabel *navigationBarSubtitleLabel;
 @property (nonatomic) UIButton *attachButton;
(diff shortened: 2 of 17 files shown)


Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -79,18 +79,18 @@ - (void)dealloc
 
 - (void)signalAccountsDidChange:(NSNotification *)notification
 {
-    dispatch_async(dispatch_get_main_queue(), ^{
-        [self updateContacts];
-    });
+    OWSAssert([NSThread isMainThread]);
+
+    [self updateContacts];
```
charlesmchen: We already take care to fire these notifications on the main thread.

Signal/src/ViewControllers/DebugUI/DebugUITableViewController.m
```diff
@@ -38,41 +39,43 @@ - (void)pushPageWithSection:(OWSTableSection *)section
...
-    [contents addSection:[OWSTableSection
-                             sectionWithTitle:messagesSection.headerTitle
-                                        items:@[
-                                            [OWSTableItem disclosureItemWithText:messagesSection.headerTitle
-                                                                     actionBlock:^{
-                                                                         [weakSelf pushPageWithSection:messagesSection];
-                                                                     }],
-                                        ]]];
-
+    NSMutableArray<OWSTableItem *> *subsectionItems = [NSMutableArray new];
```
charlesmchen: I compacted the Debug UI a bit.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -943,6 +958,9 @@
 /* Label indicating that the user is not verified. Embeds {{the user's name or phone number}}. */
 "PRIVACY_IDENTITY_IS_NOT_VERIFIED_FORMAT" = "%@ is not verified.";
```
michaelkirk: Maybe better copy would be: "You have not verified Bob." or  even "You have not marked Bob as verified."

The confusion I want to avoid is the perception that "verified" is some service level attribute. Rather "verification" is a tool for Alice to classify each of her contacts.

Contrast this with the completely different twitter concept of "verified". When Bob has a "verified" twitter account, it means Bob took some action, then  appears "verified" to *all* twitter users. Alice had no role in that. I think the recommended copy better distinguishes what we're doing from a user's potentially previous experiences.

WDYT @charlesmchen / @riyapenn

Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -79,18 +79,18 @@ - (void)dealloc
 
 - (void)signalAccountsDidChange:(NSNotification *)notification
 {
-    dispatch_async(dispatch_get_main_queue(), ^{
-        [self updateContacts];
-    });
+    OWSAssert([NSThread isMainThread]);
```
charlesmchen: What I like about OWSAssert() is that it completely compiles out in production. 

We could probably use preprocessor directives to achieve the same thing with AssertIsOnMainThread(), while remaining swift-compatible.  I can't figure out how to detect if when an objective-c header is being used from swift.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -1213,11 +1273,7 @@ - (void)showFingerprintWithRecipientId:(NSString *)recipientId
     // return from FingerprintViewController.
     [self dismissKeyBoard];
 
-    FingerprintViewController *fingerprintViewController = [FingerprintViewController new];
-    [fingerprintViewController configureWithRecipientId:recipientId];
-    UINavigationController *navigationController =
-        [[UINavigationController alloc] initWithRootViewController:fingerprintViewController];
-    [self presentViewController:navigationController animated:YES completion:nil];
+    [FingerprintViewController showVerificationViewFromViewController:self recipientId:recipientId];
```
charlesmchen: Done.

Signal/translations/en.lproj/Localizable.strings
charlesmchen: Done. I've made a couple of other similar copy changes.

PR #2200, opened 2017-06-10
Charlesmchen/fingerprints6

charlesmchen:
* Tweak verification state messages.
* Use shield instead of checkmark in conversation settings view when users is not verified.

PTAL @michaelkirk


Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
```diff
@@ -240,12 +240,16 @@ - (void)updateTableContents
...
+        BOOL isVerified = [[OWSIdentityManager sharedManager] verificationStateForRecipientId:recipientId]
+            == OWSVerificationStateVerified;
+
         [firstSection addItem:[OWSTableItem itemWithCustomCellBlock:^{
             return [weakSelf
                 disclosureCellWithName:
                     NSLocalizedString(@"VERIFY_PRIVACY",
                         @"Label for button or row which allows users to verify the safety number of another user.")
-                              iconName:@"table_ic_verify"];
+                              iconName:(isVerified ? @"table_ic_verify" : @"table_ic_not_verified")];
```
michaelkirk: Hm... I see no difference in the settings controller for this icon in running the 2.13 beta. 

Maybe it's since been reverted (still going through backlog of PRs).

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1430,16 +1430,16 @@
 "VERIFICATION_PHONE_NUMBER_FORMAT" = "Enter the verification code we sent to %@.";
 
 /* Format for info message indicating that the verification state was unverified on this device. Embeds {{user's name or phone number}}. */
-"VERIFICATION_STATE_CHANGE_FORMAT_NOT_VERIFIED_LOCAL" = "You unverified %@.";
+"VERIFICATION_STATE_CHANGE_FORMAT_NOT_VERIFIED_LOCAL" = "You marked %@ as not verified.";
```

Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
charlesmchen: As noted elsewhere, this behavior was later changed again.

PR #2203, opened 2017-06-12
Throttle profile fetch and retry on failure.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ProfileFetcherJob.swift
+++ b/Signal/src/ProfileFetcherJob.swift
@@ -14,6 +14,9 @@ class ProfileFetcherJob: NSObject {
 
     let thread: TSThread
 
+    // This property is only accessed on the default global queue.  
+    static var fetchDateMap = [String: Date]()
+
     public class func run(thread: TSThread, networkManager: TSNetworkManager) {
         ProfileFetcherJob(thread: thread, networkManager: networkManager).run()
     }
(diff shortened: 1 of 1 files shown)


Signal/src/ProfileFetcherJob.swift
```diff
@@ -26,30 +29,60 @@ class ProfileFetcherJob: NSObject {
...
+
+    public func getProfile(recipientId: String, remainingRetries: Int = 3) {
 
-                    Logger.error("\(self.TAG) failed to fetch profile for recipient: \(recipientId) with error: \(error)")
-            })
+        if let lastDate = ProfileFetcherJob.fetchDateMap[recipientId] {
+            let lastTimeInterval = fabs(lastDate.timeIntervalSinceNow)
+            // Don't check a profile more often than every N minutes.
+            let kGetProfileMaxFrequencySeconds = 60.0 * 5.0
+            if lastTimeInterval < kGetProfileMaxFrequencySeconds {
```
michaelkirk: style nit: use guard to enforce early-return pattern.

Signal/src/ProfileFetcherJob.swift
```diff
@@ -26,30 +29,60 @@ class ProfileFetcherJob: NSObject {
...
+                return
+            }
         }
+        ProfileFetcherJob.fetchDateMap[recipientId] = Date()
+
+        Logger.error("\(self.TAG) getProfile: \(recipientId)")
+
+        let request = OWSGetProfileRequest(recipientId: recipientId)
+
+        // We don't need to retainUntilComplete() since the success and failure
```
michaelkirk: not sure I understand this comment - `retainUntilComplete` is a PromiseKit extension - but we're not using promises here.

Signal/src/ProfileFetcherJob.swift
```diff
@@ -14,6 +14,9 @@ class ProfileFetcherJob: NSObject {
 
     let thread: TSThread
 
+    // This property is only accessed on the default global queue.  
+    static var fetchDateMap = [String: Date]()
```
charlesmchen: You're right. I've moved to main.

Signal/src/ProfileFetcherJob.swift
charlesmchen: Done.

Signal/src/ProfileFetcherJob.swift
charlesmchen: My thought process was that we tend to use promises in jobs and so will often want to retainUntilComplete() in jobs to the point where we should either a) use retainUntilComplete() or b) add a comment noting that it's not necessary.

But I guess it's excessive.  I'll remove it.

PR #2207, opened 2017-06-12
Throttle notification sounds.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/environment/NotificationsManager.m
+++ b/Signal/src/environment/NotificationsManager.m
@@ -22,6 +22,8 @@
 @property (nonatomic, readonly) NSMutableDictionary<NSString *, UILocalNotification *> *currentNotifications;
 @property (nonatomic, readonly) NotificationType notificationPreviewType;
 
+@property (nonatomic, readonly) NSMutableArray<NSDate *> *notificationHistory;
+
 @end
 
 #pragma mark -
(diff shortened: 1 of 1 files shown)


Signal/src/environment/NotificationsManager.m
```diff
@@ -290,12 +302,47 @@ - (void)notifyUserForIncomingMessage:(TSIncomingMessage *)message
...
+    // five-second window.
+    const CGFloat kNotificationWindowSeconds = 5.f;
+    const NSUInteger kMaxNotificationRate = 2;
+
+    // Cull obsolete notification timestamps from the thread's notification history.
+    while (self.notificationHistory.count > 0) {
+        NSDate *notificationTimestamp = self.notificationHistory[0];
+        CGFloat notificationAgeSeconds = fabs(notificationTimestamp.timeIntervalSinceNow);
+        if (notificationAgeSeconds > kNotificationWindowSeconds) {
+            [self.notificationHistory removeObjectAtIndex:0];
```
michaelkirk: Since repeatedly removing the first element of an array is expensive, it's not ideal, but it seems like there's no issue because there will never be more than 2 items in this array. Right?

charlesmchen: That's about right. We might tune the constant here from 2 to 3 or whatever, but the array won't ever be that large.

PR #2209, opened 2017-06-13
Charlesmchen/use reference cells for measurement

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/Models/OWSMessagesBubblesSizeCalculator.m
+++ b/Signal/src/Models/OWSMessagesBubblesSizeCalculator.m
@@ -31,7 +31,19 @@ NS_ASSUME_NONNULL_BEGIN
 - (CGSize)jsq_avatarSizeForMessageData:(id<JSQMessageData>)messageData
                             withLayout:(JSQMessagesCollectionViewFlowLayout *)layout;
 - (CGFloat)textBubbleWidthForLayout:(JSQMessagesCollectionViewFlowLayout *)layout;
+
 @end
+
+#pragma mark -
+
+@interface OWSMessagesBubblesSizeCalculator ()
+
+@property (nonatomic) OWSSystemMessageCell *referenceSystemMessageCell;
+@property (nonatomic) OWSUnreadIndicatorCell *referenceUnreadIndicatorCell;
+
+@end
+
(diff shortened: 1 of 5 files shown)


Signal/src/Models/OWSMessagesBubblesSizeCalculator.m
```diff
@@ -106,8 +119,12 @@ - (CGSize)sizeForSystemMessage:(TSInteraction *)interaction
         return [cachedSize CGSizeValue];
     }
 
-    CGSize result =
-        [OWSSystemMessageCell cellSizeForInteraction:interaction collectionViewWidth:layout.collectionView.width];
+    if (!self.referenceSystemMessageCell) {
+        self.referenceSystemMessageCell = [OWSSystemMessageCell new];
```
michaelkirk: take it or leave it: Maybe a readonly lazily populated getter so we don't have to worry about this being nil, should we ever decide to use it elsewhere?

Signal/src/views/OWSSystemMessageCell.m
```diff
@@ -51,7 +51,10 @@ - (instancetype)init
 
 - (void)commontInit
 {
-    OWSAssert(!self.imageView);
```
michaelkirk: nit: Are we calling `commonInit` twice? If so it's probably not an `init` method. If we're not wittingly calling it twice, I prefer the (previous) failing asserts so we could fix a bug.

charlesmchen:
> This is intended as a performance improvement, so we're not constantly initing a new label - is that right?

Yup. This is a perf hotspot for us, so recycling cell contents (i.e. the label(s)) can be significant.  I think it also ends up making the cell code more readable since you don't have to intermix class and instance methods.

Signal/src/Models/OWSMessagesBubblesSizeCalculator.m
charlesmchen: Done.

Signal/src/views/OWSSystemMessageCell.m
charlesmchen: Okay, I'll slice this a different way.

PR #2210, opened 2017-06-13
Charlesmchen/calls stuck on connecting

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -470,9 +470,20 @@ protocol CallServiceObserver: class {
 
         let newCall = SignalCall.incomingCall(localId: UUID(), remotePhoneNumber: thread.contactIdentifier(), signalingId: callId)
 
+        guard self.call == nil else {
+            // TODO on iOS10+ we can use CallKit to swap calls rather than just returning busy immediately.
+            Logger.info("\(TAG) receivedCallOffer for thread: \(thread) but we're already in call: \(call!)")
+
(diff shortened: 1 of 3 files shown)


Signal/src/call/CallService.swift
```diff
@@ -494,19 +505,15 @@ protocol CallServiceObserver: class {
             newCall.callRecord = callRecord
             callRecord.save()
 
-            return
-        }
-
-        guard self.call == nil else {
-            // TODO on iOS10+ we can use CallKit to swap calls rather than just returning busy immediately.
-            Logger.verbose("\(TAG) receivedCallOffer for thread: \(thread) but we're already in call: \(call!)")
+            terminateCall()
```
michaelkirk: This is a good catch.

Signal/src/call/CallService.swift
```diff
@@ -470,9 +470,20 @@ protocol CallServiceObserver: class {
 
         let newCall = SignalCall.incomingCall(localId: UUID(), remotePhoneNumber: thread.contactIdentifier(), signalingId: callId)
 
+        guard self.call == nil else {
```
michaelkirk: Why this reordering? I think we should prefer *not* to send anything (even a busy signal) in this scenario.

charlesmchen: Reverted in https://github.com/WhisperSystems/Signal-iOS/pull/2220

charlesmchen:
The core goal here is to improve logging around the "call stuck in connecting" issue.

PR #2215, opened 2017-06-14
Reduce number of database views.

charlesmchen:
First startup times are still slow for users with large numbers of messages, but these changes help (for a user with ~13,000 messages across 4 threads, startup times that were ~80 seconds are now ~50 seconds).

PTAL @michaelkirk

https://github.com/WhisperSystems/SignalServiceKit/pull/265


charlesmchen:
@michaelkirk It's worth noting that the first startup time measurements mentioned above are on our worst-case device, the iPhone 5c.

PR #2219, opened 2017-06-15
Avoid crashing on startup due to database view creation.

charlesmchen:
* Substitute unread view for unseen view until unseen view is ready.
* Register as many views as possible async.
* Perform blocking, safe migrations before async registration of views.
* Add debug settings view for internal app stats.
* Show blocking overlay until async registration of database views is complete.
* Convert version migrations to use AppVersions.

See: https://github.com/WhisperSystems/SignalServiceKit/pull/267

PTAL @michaelkirk


Signal/src/environment/VersionMigrations.m
```diff
@@ -12,6 +12,7 @@
 #import "SignalKeyingStorage.h"
```
michaelkirk: FYI I'm hoping to nuke this class in favor of just the MigrationRunner at some point. The last migration we added to this class was 2.4.1 - 10 months ago.

Signal/src/util/ThreadUtil.m
```diff
@@ -123,7 +123,7 @@ + (ThreadDynamicInteractions *)ensureDynamicInteractionsForThread:(TSThread *)th
         NSMutableArray<TSInvalidIdentityKeyErrorMessage *> *blockingSafetyNumberChanges = [NSMutableArray new];
         NSMutableArray<TSInteraction *> *nonBlockingSafetyNumberChanges = [NSMutableArray new];
         // We use different views for performance reasons.
-        [[transaction ext:TSThreadSpecialMessagesDatabaseViewExtensionName]
+        [[TSDatabaseView threadSpecialMessagesDatabaseView:transaction]
```
michaelkirk: I like these accessors 👍

Signal/src/environment/VersionMigrations.m
charlesmchen: +1

PR #2225, opened 2017-06-16
ignore tap on legacy non-blocking SN change message

michaelkirk:
PTAL @charlesmchen 

// FREEBIE


charlesmchen:
Do you think we should show an alert?

michaelkirk:
> Do you think we should show an alert?

Actually in 1:1 threads, maybe we should just do the right thing and show that users fingerprint screen. I'm going to make that change.

For group threads, I'm not sure what we could helpfully show in the alert.

PR #2226, opened 2017-06-16
Charlesmchen/holiday code review omnibus

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/SignalServiceKit/pull/272


Signal/src/Models/OWSMessagesBubblesSizeCalculator.m
```diff
@@ -120,7 +120,7 @@ - (CGSize)sizeForSystemMessage:(TSInteraction *)interaction
     }
 
     if (!self.referenceSystemMessageCell) {
```
michaelkirk: nit: If this change is from my feedback I did not make myself clear. I was suggesting you don't create these reference cells as a side effect of calling the `sizeForSystemMessage` (the name of which does not imply side effects), but rather we'd continue to use the property accessor as before, and implement the getter something like this:

```
@property (nonatomic, readonly) OWSSystemMessageCell *referenceSystemMessageCell;
...
@synthesize referenceSystemMessageCell = _referenceSystemMessageCell;
- (OWSSystemMessageCell *)referenceSystemMessageCell
{
  if(_referenceSystemMessageCell == nil) {
     _referenceSystemMessageCell = [OWSSystemMessageCell new];
  }
  return _referenceSystemMessageCell;
}
```

No biggie, but I think it's more resilient to future assumptions if properties never return nil (when reasonable).

Signal/src/views/OWSSystemMessageCell.m
```diff
@@ -40,21 +40,9 @@ - (instancetype)initWithFrame:(CGRect)frame
     return self;
 }
 
-- (instancetype)init
-{
```
michaelkirk: To be clear, I didn't have any qualms with this extra `init` method being defined as it was. In fact, it seems like it should *remain• as it was to return a useful object.

My confusion was the comment "don't init twice". 

How would that ever happen?

I guess if you wrote code like:

`[[[OWSSystemMessageCell alloc] init] init]`

But I guess the same could be said for *any* objc class. I just thought I was missing something.

No strong feelings either way, but I was just trying to understand what was being communicated.

Signal/src/Models/OWSMessagesBubblesSizeCalculator.m
charlesmchen: Eh simpler to just make them in the initializer.

Signal/src/views/OWSSystemMessageCell.m
charlesmchen: `[UIView init]` invokes `[self initWithFrame:...]`.  I don't like to bake that assumption into the code since a lot of people don't know that.  Anyway, I guess I'll leave things as they stand but add comments.

PR #2227, opened 2017-06-16
Fix “Invalid media attachments” crash.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/Models/TSMessageAdapaters/TSAnimatedAdapter.m
+++ b/Signal/src/Models/TSMessageAdapaters/TSAnimatedAdapter.m
@@ -103,8 +103,7 @@ NS_ASSUME_NONNULL_BEGIN
         NSData *fileData = [NSData dataWithContentsOfURL:[self.attachment mediaURL]];
         if (!fileData) {
             DDLogError(@"%@ Could not load image: %@", [self tag], [self.attachment mediaURL]);
-            OWSAssert(0);
-            return nil;
+            return [UIView new];
         }
         FLAnimatedImage *animatedGif = [FLAnimatedImage animatedImageWithGIFData:fileData];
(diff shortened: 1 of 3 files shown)


Signal/src/Models/TSMessageAdapaters/TSAnimatedAdapter.m
```diff
@@ -103,8 +103,7 @@ - (UIView *)mediaView {
         NSData *fileData = [NSData dataWithContentsOfURL:[self.attachment mediaURL]];
         if (!fileData) {
             DDLogError(@"%@ Could not load image: %@", [self tag], [self.attachment mediaURL]);
-            OWSAssert(0);
-            return nil;
+            return [UIView new];
```
charlesmchen: Turns out these methods need to return a valid UIView.

charlesmchen: Hmm, maybe I'l convert these to have a gray background so they show up against the white background.

Signal/src/Models/TSMessageAdapaters/TSPhotoAdapter.m
```diff
@@ -75,8 +75,7 @@ - (UIView *)mediaView {
         UIImage *image = self.attachment.image;
         if (!image) {
             DDLogError(@"%@ Could not load image: %@", [self tag], [self.attachment mediaURL]);
-            OWSAssert(0);
```
michaelkirk: Don't we still want to assert(0)? How could this happen and not be an error?

charlesmchen: If the image or video is corrupt or invalid.  That should be rare but possible in production.  I removed this debug assert since the attachments created by debug UI's "create tiny attachments are all invalid.  I feel like its worth it - this assert was buying us very little.

PR #2229, opened 2017-06-16
Don’t auto-scroll after “loading more messages” unless we have “more …

charlesmchen:
…unseen messages”.

Otherwise, tapping on "load more messages" autoscrolls you _downward_ which is completely wrong.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -2184,6 +2184,9 @@ typedef enum : NSUInteger {
                              header:(JSQMessagesLoadEarlierHeaderView *)headerView
     didTapLoadEarlierMessagesButton:(UIButton *)sender
 {
+    OWSAssert(!self.isUserScrolling);
+
+    BOOL hasMoreUnseenMessages = self.dynamicInteractions.hasMoreUnseenMessages;
 
     // We want to restore the current scroll state after we update the range, update
     // the dynamic interactions and re-layout.  Here we take a "before" snapshot.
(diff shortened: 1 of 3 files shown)


Signal/src/util/ThreadUtil.h
```diff
@@ -36,6 +36,8 @@ NS_ASSUME_NONNULL_BEGIN
 // YES in ensureDynamicInteractionsForThread:...
 @property (nonatomic, nullable) NSNumber *firstUnseenInteractionTimestamp;
 
+@property (nonatomic) BOOL hasMoreUnseenMessages;
```
michaelkirk: nit: read only in the .h?

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -2229,13 +2232,16 @@ - (void)collectionView:(JSQMessagesCollectionView *)collectionView
...
     self.collectionView.contentOffset = CGPointMake(0, self.collectionView.contentSize.height - scrollDistanceToBottom);
+
     [self.scrollLaterTimer invalidate];
-    // We want to scroll to the bottom _after_ the layout has been updated.
-    self.scrollLaterTimer = [NSTimer weakScheduledTimerWithTimeInterval:0.001f
-                                                                 target:self
-                                                               selector:@selector(scrollToUnreadIndicatorAnimated)
-                                                               userInfo:nil
-                                                                repeats:NO];
+    if (hasMoreUnseenMessages) {
```
michaelkirk: I had to read the PR comment a second time to understand what this was doing, but it totally makes sense. 

Take it or leave it, but since this re:'loadEarlierMessages' maybe calling this `hasEarlierUnseenMessages` would be clearer.

Signal/src/util/ThreadUtil.h
charlesmchen: Done.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
charlesmchen: Improved the comments in the code.

charlesmchen: Also took your renaming suggestion.

PR #2230, opened 2017-06-16
Tweak font size on registration view.

charlesmchen:
Increase font size of phone number to match size of country code.

PTAL @michaelkirk 

img_0025


michaelkirk:
LGTM

PR #2231, opened 2017-06-16
System messsages can show timestamp

michaelkirk:
added the JSQCollectionViewCell cellTopLabel to the system message
layout

screen shot 2017-06-16 at 4 10 19 pm

(note I modified the kTSMessageSentDateShowTimeInterval for the screenshot to show smaller time intervals)

I followed the existing pattern of manual layout - but is there a reason we're not using autoLayout?

PTAL @charlesmchen


Signal/src/views/OWSSystemMessageCell.m
```diff
@@ -60,6 +64,12 @@ - (void)commontInit
 
     self.backgroundColor = [UIColor whiteColor];
 
+    self.cellTopLabel = [UILabel new];
+    self.cellTopLabel.textAlignment = NSTextAlignmentCenter;
+    self.cellTopLabel.font = self.topLabelFont;
+    self.cellTopLabel.textColor = [UIColor lightGrayColor];
```
charlesmchen: Can these ever line wrap?

Signal/src/views/OWSSystemMessageCell.m
```diff
@@ -265,9 +293,12 @@ - (CGSize)cellSizeForInteraction:(TSInteraction *)interaction collectionViewWidt
     [self applyTitleForInteraction:interaction label:self.titleLabel];
     CGFloat maxTitleWidth = (collectionViewWidth - ([self hMargin] * 2.f + [self hSpacing] + [self iconSize]));
     CGSize titleSize = [self.titleLabel sizeThatFits:CGSizeMake(maxTitleWidth, CGFLOAT_MAX)];
+
     CGFloat contentHeight = ceil(MAX([self iconSize], titleSize.height));
     result.height += contentHeight;
 
+    result.height += [self topLabelBottomMargin];
```
charlesmchen: Won't this affect all cells - even if they have no top label content?

charlesmchen:
> I followed the existing pattern of manual layout - but is there a reason we're not using autoLayout?

I've read conflicting things about whether iOS auto layout can cause perf problems in collection view cells.  I also wanted to have fine-grained control over the layout and measurement.

Signal/src/views/OWSSystemMessageCell.m
michaelkirk: Even on our smallest device this takes up ~30% of the view width, so I think we're safe.

Signal/src/views/OWSSystemMessageCell.m
michaelkirk: You're right. I think I've pushed up a better approach here:
https://github.com/WhisperSystems/Signal-iOS/pull/2231/commits/5a0d6cba581cbe248540eb0f11535179cacfce5a

PTAL @charlesmchen

michaelkirk: updated screenshot:

screen shot 2017-06-16 at 5 21 13 pm

PR #2232, opened 2017-06-16
Ensure message mapping is up-to-date when app returns from background.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -561,6 +561,13 @@ typedef enum : NSUInteger {
     // or on another device.
     [self hideInputIfNeeded];
 
+    // We need to `beginLongLivedReadTransaction` before we update our
+    // mapping in order to jump to the most recent commit.
+    [self.uiDatabaseConnection beginLongLivedReadTransaction];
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -561,6 +561,13 @@ - (void)viewWillAppear:(BOOL)animated
     // or on another device.
     [self hideInputIfNeeded];
 
+    // We need to `beginLongLivedReadTransaction` before we update our
+    // mapping in order to jump to the most recent commit.
+    [self.uiDatabaseConnection beginLongLivedReadTransaction];
```
michaelkirk: This seems like a good thing to do, but I'm not sure if it explains the misplaced interactions (if it was intended to).

So long as we're always *reading* from our uiDatabaseConnection to present the UI - it might be stale, but shouldn't be inconsistent. right?

charlesmchen: Right - it's the `messageMappings` that is inconsistent, and before we update it below, we want to make sure that uiDatabaseConnection has latest.  You'll notice we do this before `toggleObservers`, where we begin to observe database modified events.  

I don't have a repro for this transient issue, but I think this is the right approach.

PR #2240, opened 2017-06-19
Fix: tapping input obscures last messages

michaelkirk:
Previously tapping the input would pop the keyboard, which would hide the last couple messages, until you tapped a key, which would scroll to bottom.

Instead we should scroll to bottom when the keyboard is popped.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/OWSMessagesComposerTextView.m
+++ b/Signal/src/ViewControllers/ConversationView/OWSMessagesComposerTextView.m
@@ -19,7 +19,7 @@ NS_ASSUME_NONNULL_BEGIN
     BOOL becameFirstResponder = [super becomeFirstResponder];
     if (becameFirstResponder) {
         // Intercept to scroll to bottom when text view is tapped.
-        [self.textViewPasteDelegate textViewDidChangeSize];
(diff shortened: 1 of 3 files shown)


charlesmchen:
Should we override `- (BOOL)becomeFirstResponder;` instead?

michaelkirk:
done, PTAL @charlesmchen

Signal/src/ViewControllers/ConversationView/OWSMessagesComposerTextView.m
```diff
@@ -14,6 +14,16 @@ - (BOOL)canBecomeFirstResponder
     return YES;
 }
 
+- (BOOL)becomeFirstResponder
+{
+    BOOL becameFirstResponder = [super becomeFirstResponder];
+    if (becameFirstResponder) {
+        // Intercept to scroll to bottom when text view is tapped.
+        [self.textViewPasteDelegate textViewDidChangeSize];
```
charlesmchen: Should we rename the delegate method?  Just asking.

michaelkirk: done.

Issue #2154, opened 2017-05-25
Weird contacts ordering and display


hodgesmr:
- [X] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description
I _think_ this started after upgrading to 2.12.2, but not completely sure. I noticed that contacts are now sorted by last name, but have the firs name bolded. This is not how iOS sorts its contacts (which, you can configure). For example, I have my iOS contacts sorted by first name, and the first name is bolded.

### Steps to reproduce
- Tap the compose button, which takes you to your Signal contacts

**Actual result:** Contacts are sorted by last name, but have first name in bold
**Expected result:** I personally like my contacts sorted by first name (and to have first name bolded), but others may prefer sorting by last name (and have last name bolded)

### Device info
**Signal version:** 2.12.2

michaelkirk:
What version of iOS are you running?

Can you attach a screenshot of your contacts settings?

Open the Settings app and go to Contacts

hodgesmr:
iOS version: 10.3.2

![img_4303](https://cloud.githubusercontent.com/assets/1208575/26476723/979cb634-418e-11e7-84c6-291cc589d2fc.JPG)

denver-s:
I think this can be closed now.

michaelkirk-signal:
Please re-open if you're still able to reproduce this.

PR #2245, opened 2017-06-20
respect system sort order for contacts

michaelkirk:
Fixes https://github.com/WhisperSystems/Signal-iOS/issues/2154

### DIFF
--- a/Signal/src/contact/SystemContactsFetcher.swift
+++ b/Signal/src/contact/SystemContactsFetcher.swift
@@ -25,6 +25,8 @@ class ContactsFrameworkContactStoreAdaptee: ContactStoreAdaptee {
     private let contactStore = CNContactStore()
     private var changeHandler: (() -> Void)?
     private var initializedObserver = false
+    private var lastSortOrder: CNContactSortOrder?
+
     let supportsContactEditing = true
 
     private let allowedContactKeys: [CNKeyDescriptor] = [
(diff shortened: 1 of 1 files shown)


charlesmchen:
So, the question is where should we honor this?

* "new 1:1 conversation view" seems like an obvious yes.
* The group views are more ambiguous since they could be a mix of contacts and non-contacts.
*  The Block list view is likely to be mostly non-contacts.

Also, via profiles we'll soon have more "display names", and our profile names don't have separate first/last components.  It'd be weird if we showed a mixture of phone numbers, contact names, profile names with inconsistent sorting between them.

charlesmchen:
In lots of the views (like the group views) we sort by recipient id right now FYI.

michaelkirk:
> So, the question is where should we honor this?
> 
> "new 1:1 conversation view" seems like an obvious yes.
> The group views are more ambiguous since they could be a mix of contacts and non-contacts.
> The Block list view is likely to be mostly non-contacts.
> Also, via profiles we'll soon have more "display names", and our profile names don't have separate > first/last components. It'd be weird if we showed a mixture of phone numbers, contact names, profile names with inconsistent sorting between them.

This PR fixes sorting the results returned by the contactFetcher. So anything that respects that sort order (anything using `contactsHelper.allContacts`) benefits from this.

for example:
- picking contact when starting new 1:1 conversation
- selecting contacts when creating a new group
- adding a contact to group
- blocking a contact

Since it seems like all forward movement (which I think covers 99% of the common use case). Are you OK with merging this as is, and building it out as necessary for future feature work?

charlesmchen:
> Since it seems like all forward movement (which I think covers 99% of the common use case). Are you OK with merging this as is, and building it out as necessary for future feature work?

Totally - just wanted to remember that we might not get this for free everywhere.  We should also make sure that we're still honoring sort order when searching, for example.

charlesmchen:
And I do think there are some difficult questions about consistency is certain places.

michaelkirk:
thanks for the feedback.

Rebased and merged in: https://github.com/WhisperSystems/Signal-iOS/commit/3162160b5ef55522f6bf92cd381139ec091d5534

PR #2246, opened 2017-06-20
Max out the unread count at 99.

charlesmchen:
PTAL @michaelkirk 

img_0028

img_0027

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -3812,7 +3812,8 @@ typedef enum : NSUInteger {
 - (void)updateBackButtonUnreadCount
 {
     AssertIsOnMainThread();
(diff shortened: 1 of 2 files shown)


charlesmchen:
Those screenshots are "before", demonstrating the issue this PR intends to address.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -3812,7 +3812,8 @@ - (void)clearDraft
 - (void)updateBackButtonUnreadCount
 {
     AssertIsOnMainThread();
-    self.backButtonUnreadCount = [self.messagesManager unreadMessagesCountExcept:self.thread];
```
michaelkirk: Can we fit in a "99+"? Or would that look bad.

charlesmchen: Good question.

charlesmchen: I tried, it doesn't fit.

PR #2249, opened 2017-06-20
delete system messages

michaelkirk:
The reason this doesn't "just work" with our collection view, is JSQ does a hit test based on the "messageBubbleContainerView" which we don't use in the system message. 

I considered adding one, just to satisfy the hit test in a consistent way, but I think I actually prefer this approach.

Plus doing layout with a container view when we're not using autolayout is a bunch of lines of code I'd rather not mess up.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -2682,8 +2682,9 @@ typedef enum : NSUInteger {
     [systemMessageCell becomeFirstResponder];
 
     UIMenuController *theMenu = [UIMenuController sharedMenuController];
-
-    [theMenu setTargetRect:systemMessageCell.titleLabel.frame inView:systemMessageCell];
+    CGRect targetRect = [systemMessageCell.titleLabel.superview convertRect:systemMessageCell.titleLabel.frame
+                                                                     toView:systemMessageCell];
+    [theMenu setTargetRect:targetRect inView:systemMessageCell];
     [theMenu setMenuVisible:YES animated:YES];
 }
 
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -1531,6 +1531,24 @@ typedef enum : NSUInteger {
     return YES;
 }
 
+- (BOOL)collectionView:(UICollectionView *)collectionView
+      canPerformAction:(SEL)action
+    forItemAtIndexPath:(NSIndexPath *)indexPath
+            withSender:(id)sender
+{
+    id<OWSMessageData> messageData = [self messageAtIndexPath:indexPath];
+    return [messageData canPerformEditingAction:action];
+}
+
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -1531,6 +1531,24 @@ - (BOOL)collectionView:(JSQMessagesCollectionView *)collectionView
     return YES;
 }
 
+- (BOOL)collectionView:(UICollectionView *)collectionView
```
michaelkirk: This was just a straight move - colocated with the other CollectionViewDelegate methods.

Signal/src/views/OWSSystemMessageCell.h
```diff
@@ -21,6 +23,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, weak) id<OWSSystemMessageCellDelegate> systemMessageCellDelegate;
 
+@property (nonatomic, readonly) UILabel *titleLabel;
```
michaelkirk: using this to set the target rect for the menu, otherwise the contentCell might be a bit too high when we're showing a date stamp.

Signal/src/views/OWSSystemMessageCell.m
```diff
@@ -72,6 +72,10 @@ - (void)commontInit
     UITapGestureRecognizer *tap =
         [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(handleTapGesture:)];
     [self addGestureRecognizer:tap];
+
+    UILongPressGestureRecognizer *longpress =
```
charlesmchen: s/longpress/longPress/

Signal/src/views/OWSSystemMessageCell.m
```diff
@@ -299,6 +318,26 @@ - (void)handleTapGesture:(UITapGestureRecognizer *)tap
     [self.systemMessageCellDelegate didTapSystemMessageWithInteraction:self.interaction];
 }
 
+- (void)handleLongpressGesture:(UILongPressGestureRecognizer *)longpress
```
charlesmchen: s/handleLongpressGesture/handleLongPressGesture/

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -2653,6 +2671,22 @@ - (void)didTapSystemMessageWithInteraction:(TSInteraction *)interaction
...
+    OWSAssert(systemMessageCell);
+    OWSAssert(systemMessageCell.interaction);
+
+    DDLogDebug(@"%@ long pressed system message cell: %@", self.tag, systemMessageCell);
+
+    [systemMessageCell becomeFirstResponder];
+
+    UIMenuController *theMenu = [UIMenuController sharedMenuController];
+
+    [theMenu setTargetRect:systemMessageCell.titleLabel.frame inView:systemMessageCell];
```
charlesmchen: This assumes that titleLabel's frame is in the coordinate system of systemMessageCell. Probably fine, but we could also convert coordinate systems.

PR #2250, opened 2017-06-20
Invalid message adapter cache when app becomes visible.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
> I think we're on the same page, but I either this or #2251 should be merged, but not both. Right?

Right. And since #2251 is not safe to merge yet, I'll take this one for this release.  Perf hit. =/

PR #2251, opened 2017-06-20
Always observe notifications in conversation view.

charlesmchen:
An alternative to https://github.com/WhisperSystems/Signal-iOS/pull/2250

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -215,6 +215,7 @@ typedef enum : NSUInteger {
 
 @property (nonatomic, readonly) BOOL isGroupConversation;
 @property (nonatomic) BOOL isUserScrolling;
+@property (nonatomic) BOOL hasAppeared;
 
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -456,48 +457,32 @@ - (void)registerCustomMessageNibs
...
-                                                      object:nil];
-        [[NSNotificationCenter defaultCenter] removeObserver:self
-                                                        name:UIApplicationWillResignActiveNotification
-                                                      object:nil];
-        [[NSNotificationCenter defaultCenter] removeObserver:self
-                                                        name:UIApplicationDidEnterBackgroundNotification
-                                                      object:nil];
-    }
+- (void)addObservers
+{
```
michaelkirk: Can we combine this with addNotificationListeners?

charlesmchen:
Definitely don't want to do this now. I realize there's also so unfinished business in the "load more" logic.

charlesmchen:
Obsolete.

PR #2254, opened 2017-06-20
Fix upgrade launch screen.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -199,13 +199,22 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
         [[UIStoryboard storyboardWithName:@"Launch Screen" bundle:nil] instantiateInitialViewController];
 
     BOOL shouldShowUpgradeLabel = NO;
-    NSString *previousVersion = AppVersion.instance.lastCompletedLaunchAppVersion;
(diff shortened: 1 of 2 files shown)


Signal/src/AppDelegate.m
```diff
@@ -199,13 +199,22 @@ - (UIViewController *)loadingRootViewController
         [[UIStoryboard storyboardWithName:@"Launch Screen" bundle:nil] instantiateInitialViewController];
 
     BOOL shouldShowUpgradeLabel = NO;
-    NSString *previousVersion = AppVersion.instance.lastCompletedLaunchAppVersion;
+    NSString *lastAppVersion = AppVersion.instance.lastAppVersion;
+    NSString *lastCompletedLaunchAppVersion = AppVersion.instance.lastCompletedLaunchAppVersion;
+    BOOL mayNeedUpgrade = ([TSAccountManager isRegistered] &&
```
michaelkirk: Is the long term strategy to bump these constants as we add new migrations?

charlesmchen: Yup,

michaelkirk: I'd be more concerned, but it only affects the appearance of the label. Expect an outrageous fury of indignation if we start re-inventing the "one shot" approach to migrations taken in VersionMigrations

PR #2256, opened 2017-06-20
Charlesmchen/attachment stream upgrade perf

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/SignalServiceKit/pull/277

### DIFF
--- a/Signal/src/environment/Migrations/OWS105AttachmentFilePaths.m
+++ b/Signal/src/environment/Migrations/OWS105AttachmentFilePaths.m
@@ -29,9 +29,7 @@ static NSString *const OWS105AttachmentFilePathsMigrationId = @"105";
                                                   return;
                                               }
                                               TSAttachmentStream *attachmentStream = (TSAttachmentStream *)attachment;
-                                              if (attachmentStream.hasUnsavedFilePath) {
-                                                  [attachmentStreams addObject:attachmentStream];
-                                              }
+                                              [attachmentStreams addObject:attachmentStream];
                                           }];
 
     DDLogInfo(@"Saving %zd attachment streams.", attachmentStreams.count);
--- a/Signal/src/environment/Migrations/OWS105AttachmentFilePaths.m
+++ b/Signal/src/environment/Migrations/OWS105AttachmentFilePaths.m
@@ -34,6 +34,9 @@ static NSString *const OWS105AttachmentFilePathsMigrationId = @"105";
 
     DDLogInfo(@"Saving %zd attachment streams.", attachmentStreams.count);
 
+    // Persist the new localRelativeFilePath property of TSAttachmentStream.
+    // For performance, we want to upgrade all existing attachment streams in
+    // a single transaction.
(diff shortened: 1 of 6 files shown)


charlesmchen:
@michaelkirk It's a _massive_ perf and responsiveness win to save these all in a single transaction.

Signal/src/environment/Migrations/OWS105AttachmentFilePaths.m
```diff
@@ -0,0 +1,58 @@
...
+    OWSAssert(transaction);
+
+    NSMutableArray<TSAttachmentStream *> *attachmentStreams = [NSMutableArray new];
+    [transaction enumerateKeysAndObjectsInCollection:TSAttachmentStream.collection
+                                          usingBlock:^(NSString *key, TSAttachment *attachment, BOOL *stop) {
+                                              if (![attachment isKindOfClass:[TSAttachmentStream class]]) {
+                                                  return;
+                                              }
+                                              TSAttachmentStream *attachmentStream = (TSAttachmentStream *)attachment;
+                                              if (attachmentStream.hasUnsavedFilePath) {
```
michaelkirk: Proposal:

In ssk, rename `ensureFilePathAndPersist:` to just `ensureFilePath`, and remove any of the persistence logic.

Then change this migration to:
```
[attachmentStream ensureFilePath];
[attachmentStream saveWithTransaction:transaction];
```

We continue to call `ensureFilePath` in initWithCoder, which will be fast because of this migration.

That would let us get rid of tracking this new `hasUnsavedFilePath` property which is only used here, in this migration and allows us to simplify the `ensureFilePath` method.

WDYT @charlesmchen?

charlesmchen: Not a bad idea, although the migration doesn't need to `ensureFilePath`; (all of) the initializers have to do that.

charlesmchen:
PTAL @michaelkirk

Signal/src/environment/Migrations/OWS105AttachmentFilePaths.m
```diff
@@ -0,0 +1,56 @@
...
+                                              if (![attachment isKindOfClass:[TSAttachmentStream class]]) {
+                                                  return;
+                                              }
+                                              TSAttachmentStream *attachmentStream = (TSAttachmentStream *)attachment;
+                                              [attachmentStreams addObject:attachmentStream];
+                                          }];
+
+    DDLogInfo(@"Saving %zd attachment streams.", attachmentStreams.count);
+
+    for (TSAttachmentStream *attachmentStream in attachmentStreams) {
```
michaelkirk: maybe just a comment for future me

```
// For performance, persist the file path which we've newly added to the initWithCoder initializer
```

Signal/src/environment/Migrations/OWS105AttachmentFilePaths.m
michaelkirk: ah, you're right!

PR #2258, opened 2017-06-21
Improve logging in CallService.

charlesmchen:
Comments in-line.

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -462,7 +462,7 @@ protocol CallServiceObserver: class {
 
             let callerName = self.contactsManager.displayName(forPhoneIdentifier: thread.contactIdentifier())
 
-            switch(untrustedIdentity!.verificationState) {
+            switch untrustedIdentity!.verificationState {
             case .verified:
                 Logger.error("\(TAG) shouldn't have missed a call due to untrusted identity if the identity is verified")
                 assertionFailure("shouldn't have missed a call due to untrusted identity if the identity is verified")
@@ -491,6 +491,16 @@ protocol CallServiceObserver: class {
             Logger.info("\(TAG) receivedCallOffer: \(newCall.debugString) but we're already in call: \(call!.debugString)")
 
             handleLocalBusyCall(newCall, thread: thread)
+
+            if self.call?.remotePhoneNumber == newCall.remotePhoneNumber {
+                // If we're receiving a new call offer from the user we think we have a call with, terminate
+                // our current call to get back to a known good state.  If they call back, we'll be ready.
+                // 
+                // TODO: Auto-accept this incoming call if our current call was either a) outgoing or 
+                // b) ever connected.  There will be a bit of complexity around making sure that two
+                // parties that call each other at the same time end up connected.
+                terminateCall()
+            }
 
             return
         }
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -130,8 +130,6 @@ protocol CallServiceObserver: class {
         }
     }
 
-    // TODO code cleanup: move thread into SignalCall? Or refactor messageSender to take SignalRecipient identifier.
-    var thread: TSContactThread?
     var call: SignalCall? {
         didSet {
             AssertIsOnMainThread()
@@ -142,7 +140,7 @@ protocol CallServiceObserver: class {
             updateIsVideoEnabled()
             updateLockTimerEnabling()
 
-            Logger.debug("\(self.TAG) .call setter: \(oldValue != nil) -> \(call != nil) \(String(describing: call))")
+            Logger.debug("\(self.TAG) .call setter: \(oldValue?.debugString as Optional) -> \(call?.debugString as Optional)")
 
             for observer in observers {
                 observer.value?.didUpdateCall(call:call)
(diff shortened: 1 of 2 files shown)


Signal/src/call/CallService.swift
```diff
@@ -130,8 +130,6 @@ protocol CallServiceObserver: class {
         }
     }
 
-    // TODO code cleanup: move thread into SignalCall? Or refactor messageSender to take SignalRecipient identifier.
-    var thread: TSContactThread?
```
charlesmchen: Moved this property into `SignalCall`.

Signal/src/call/CallService.swift
```diff
@@ -161,9 +159,6 @@ protocol CallServiceObserver: class {
     var sendIceUpdatesImmediately = true
     var pendingIceUpdateMessages = [OWSCallIceUpdateMessage]()
 
-    // ensure the incoming call promise isn't dealloc'd prematurely
-    var incomingCallPromise: Promise<Void>?
```
charlesmchen: I removed this property in favor of using `retainUntilComplete()`.

Signal/src/call/CallService.swift
```diff
@@ -338,22 +330,24 @@ protocol CallServiceObserver: class {
                 self.handleFailedCall(failedCall: call, error: externalError)
             }
         }
+        promise.retainUntilComplete()
```
charlesmchen: I made sure we're using `retainUntilComplete()`, even though its not necessary.

Signal/src/call/CallService.swift
```diff
@@ -142,7 +140,7 @@ protocol CallServiceObserver: class {
             updateIsVideoEnabled()
             updateLockTimerEnabling()
 
-            Logger.debug("\(self.TAG) .call setter: \(oldValue != nil) -> \(call != nil) \(String(describing: call))")
+            Logger.debug("\(self.TAG) .call setter: \(oldValue?.debugString as Optional) -> \(call?.debugString as Optional)")
```
charlesmchen: `SignalCall.debugString()` returns the triplet "remote phone number, local id, signaling id" which will make these logs more usable.

Signal/src/call/CallService.swift
```diff
@@ -1198,34 +1163,35 @@ protocol CallServiceObserver: class {
...
+        handleFailedCall(failedCall: self.call, error: error)
     }
 
     // This method should be called when a fatal error occurred for a call.
     //
     // * If we know which call it was, we should update that call's state
     //   to reflect the error.
     // * IFF that call is the current call, we want to terminate it.
-    public func handleFailedCall(failedCall: SignalCall?, error: CallError, forceTerminate: Bool = false) {
+    public func handleFailedCall(failedCall: SignalCall?, error: CallError) {
```
charlesmchen: I've modified `handleFailedCall()` to always `terminateCall()` when the failed call is nil (i.e. the `handleFailedCurrentCall()` case), which it wasn't before.  That was a bug.

charlesmchen:
PTAL @michaelkirk

charlesmchen:
I pushed one more commit to: "Hang up current call if we receive a call offer from the same user."

Signal/src/call/CallService.swift
```diff
@@ -494,16 +488,25 @@ protocol CallServiceObserver: class {
...
             handleLocalBusyCall(newCall, thread: thread)
 
+            if self.call?.remotePhoneNumber == newCall.remotePhoneNumber {
+                // If we're receiving a new call offer from the user we think we have a call with, terminate
+                // our current call to get back to a known good state.  If they call back, we'll be ready.
+                // 
+                // TODO: Auto-accept this incoming call if our current call was either a) outgoing or 
+                // b) ever connected.  There will be a bit of complexity around making sure that two
+                // parties that call each other at the same time end up connected.
+                terminateCall()
```
michaelkirk: I haven't run this - what does this user see locally when they terminate their call? 

Presumably if both users are calling each other at the same time, they'll *both* hit this logic, and *both* terminate their local call?

Signal/src/call/CallService.swift
michaelkirk: :+1:

charlesmchen:
> nit: does it make sense to override/use the existing debugDescription property?

I could see logging different ways in different contexts.  I wanted this to be a very specific "triplet" or tuple of values that would be used to identify calls but as tersely as possible, i.e. not include call state, etc.   I could see debugDescription being much more verbose.  Maybe I should rename it `logIdentifiers` or something like that.

charlesmchen:
> Actually, this might be a significant change for the better - we were returning, but then not retaining this in the non-callkit adapter.

It's so easy to make this kind error - we've done it before - that I'm tempted to always use retainUntilComplete() for all promises that have more than element.

Signal/src/call/CallService.swift
```diff
@@ -597,28 +601,23 @@ protocol CallServiceObserver: class {
         AssertIsOnMainThread()
         Logger.info("\(TAG) called \(#function)")
 
-        guard let currentThread = self.thread else {
+        guard let call = self.call else {
             Logger.warn("ignoring remote ice update for thread: \(thread.uniqueId) since there is no current thread. Call already ended?")
```

charlesmchen:
> Presumably if both users are calling each other at the same time, they'll both hit this logic, and both terminate their local call?

One caller will see "busy" and one will see "call failed".

PR #2259, opened 2017-06-21
Persist registration values in the keychain for debug builds only.

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/SignalServiceKit/pull/278

### DIFF
--- a/Signal/src/ViewControllers/RegistrationViewController.m
+++ b/Signal/src/ViewControllers/RegistrationViewController.m
@@ -18,9 +18,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 #ifdef DEBUG
 
-NSString *const kKeychainServiceDebug = @"kKeychainServiceDebug";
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/RegistrationViewController.m
```diff
@@ -396,9 +427,65 @@ - (BOOL)textField:(UITextField *)textField
...
+
+#ifdef DEBUG
+
+- (NSString *_Nullable)debugKeychainValueForKey:(NSString *)key
+{
+    OWSCAssert([NSThread isMainThread]);
+    OWSCAssert(key.length > 0);
+
+    NSError *error;
+    NSString *value = [SAMKeychain passwordForService:kKeychainServiceDebug account:key error:&error];
```
michaelkirk: Take it or leave it:

Why store this in the keychain vs user defaults?

I  feel like the system keychain is a beast to be avoided whenever reasonable.

charlesmchen: Done.

PR #2261, opened 2017-06-22
If two users call each other at the same time, ensure they both see “…

charlesmchen:
…busy”.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -505,9 +505,8 @@ protocol CallServiceObserver: class {
                 // be ready.
                 // 
                 // TODO: Auto-accept this incoming call if our current call was either a) outgoing or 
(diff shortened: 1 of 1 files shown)


charlesmchen:
PTAL @michaelkirk

charlesmchen:
PTAL @michaelkirk

Signal/src/call/CallService.swift
```diff
@@ -492,14 +492,27 @@ protocol CallServiceObserver: class {
...
+
+                // If we're receiving a new call offer from the user we already think we have a call with,
+                // terminate our current call to get back to a known good state.  If they call back, we'll 
+                // be ready.
                 // 
                 // TODO: Auto-accept this incoming call if our current call was either a) outgoing or 
                 // b) ever connected.  There will be a bit of complexity around making sure that two
-                // parties that call each other at the same time end up connected.
-                terminateCall()
+                // parties that call each other at
```

PR #2264, opened 2017-06-22
Fix crash in manual censorship circumvention logic on iOS 9.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/OWSCountryMetadata.m
+++ b/Signal/src/ViewControllers/OWSCountryMetadata.m
@@ -24,7 +24,7 @@ NS_ASSUME_NONNULL_BEGIN
     instance.googleDomain = googleDomain;
     instance.countryCode = countryCode;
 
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/OWSCountryMetadata.m
```diff
@@ -24,7 +24,7 @@ + (OWSCountryMetadata *)countryMetadataWithName:(NSString *)name
     instance.googleDomain = googleDomain;
     instance.countryCode = countryCode;
 
-    NSString *localizedCountryName = [[NSLocale currentLocale] localizedStringForCountryCode:countryCode];
+    NSString *localizedCountryName = [[NSLocale currentLocale] displayNameForKey:NSLocaleCountryCode value:countryCode];
```
charlesmchen: This method is only available on iOS 10.

PR #2267, opened 2017-06-22
Add script to bump build numbers.

charlesmchen:
It's not the most polished script in the world, but it's safe and it'll fail with messages that will be useful for resolving any issue that comes up.

PTAL @michaelkirk

### DIFF
--- a/Scripts/bump_build_number.py
+++ b/Scripts/bump_build_number.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python
+import sys
+import os
+import re
+import commands
+import subprocess
+
+def fail(message):
+    print message
+    sys.exit(1)
+
+def find_project_root():
+    path = os.path.abspath(os.curdir)
+    
+    while True:
+        # print 'path', path
+        if not os.path.exists(path):
+            break
+        git_path = os.path.join(path, '.git')
+        if os.path.exists(git_path):
+            return path
+        new_path = os.path.abspath(os.path.dirname(path))
+        if not new_path or new_path == path:
+            break
+        path = new_path
+    
+    fail('Could not find project root path')
+
+if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


charlesmchen:
Sample output on success:
```
matthews-iMac:Signal-iOS-4 matthew$ ./Scripts/bump_build_number.py
old_build_number: 2.13.0.18
new_build_number: 2.13.0.19
```
Sample output on failure:
```
matthews-iMac:Signal-iOS-4 matthew$ ./Scripts/bump_build_number.py
AM Scripts/bump_build_number.py
M  Signal/Signal-Info.plist

Git repository has untracked files.
```

Scripts/bump_build_number.py
```diff
@@ -0,0 +1,81 @@
...
+    build_number = str(1 + int(build_number))
+    new_build_number = old_build_number[:build_number_match.start(1)] + build_number
+    print 'new_build_number:', new_build_number
+    
+    text = text[:file_match.start(1)] + new_build_number + text[file_match.end(1):]
+    with open(plist_path, 'wt') as f:
+        f.write(text)
+        
+    output = subprocess.check_output(['git', 'add', '.'])
+    output = subprocess.check_output(['git', 'commit', '-m', 'Bump build from to %s.\n\n// FREEBIE' % new_build_number])
```
michaelkirk: "Bump build from **%s** to %s" ?

or just 

"Bump build to %s" ?

charlesmchen: I prefer the latter but let me know if you want to flip it.

charlesmchen:
> Out of the box they do an integer (non-decimal) build number, which is also fine with me.

Let's stick with this for now, since we use the long version numbers in various places and I can't be bothered to change them.

PR #2268, opened 2017-06-23
Charlesmchen/disk usage

charlesmchen:
PTAL @michaelkirk


Signal/src/ViewControllers/DebugUI/DebugUIDiskUsage.m
```diff
@@ -0,0 +1,299 @@
...
+    return @"Orphans & Disk Usage";
+}
+
+- (nullable OWSTableSection *)sectionForThread:(nullable TSThread *)thread
+{
+    return [OWSTableSection sectionWithTitle:self.name
+                                       items:@[
+                                           [OWSTableItem itemWithTitle:@"Audit & Log"
+                                                           actionBlock:^{
+                                                               [DebugUIDiskUsage auditWithoutCleanup];
```
charlesmchen: You may find some missing attachment files. I think all of mine correspond to attachments made during development of the "arbitrary attachments" and "oversize text attachments" work.  Please send me the output of this command on your primary.

Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
```diff
@@ -13,10 +13,19 @@
 
 @implementation DebugUISessionState
 
-+ (OWSTableSection *)sectionForContactThread:(TSContactThread *)contactThread
```
michaelkirk: Why did you get rid of this? It was intended to indicate that this could only be used for ContactThreads.

Most of these actions don't make sense for group threads. 

e.g. line 54

```
+                                           NSString *recipientId = [thread contactIdentifier];
```

Signal/src/ViewControllers/DebugUI/DebugUIDiskUsage.m
```diff
@@ -0,0 +1,299 @@
...
+}
+
++ (void)auditAndCleanup:(BOOL)shouldCleanup
+{
+    NSString *attachmentsFolder = [TSAttachmentStream attachmentsFolder];
+    DDLogError(@"attachmentsFolder: %@", attachmentsFolder);
+
+    __block int fileCount = 0;
+    __block long long totalFileSize = 0;
+    NSMutableSet *diskFilePaths = [NSMutableSet new];
```
michaelkirk: Could I convince you to move this implementation to the orphan file cleaner?

motivation:

1. That class is broken now that we have  subdirectories in our attachment path.
2. It seems like they'll be a day when we use this outside of debugUI.

Signal/src/ViewControllers/DebugUI/DebugUIDiskUsage.m
```diff
@@ -0,0 +1,299 @@
...
+                                                              *stop = YES;
+                                                              return;
+                                                          }
+                                                          [interactionsToDelete addObject:interaction];
+                                                      }];
+        }
+
+        DDLogInfo(@"Deleting %zd interactions.", interactionsToDelete.count);
+
+        for (TSInteraction *interaction in interactionsToDelete) {
```
michaelkirk: If this ever becomes public facing we may want to batch this to avoid loading a billion interactions into memory at once.

I think it's completely reasonable as is for debug though.

Signal/src/ViewControllers/DebugUI/DebugUIDiskUsage.m
```diff
@@ -0,0 +1,299 @@
...
+    [storageManager.newDatabaseConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
+
+        NSMutableArray<NSString *> *threadIds = [NSMutableArray new];
+        YapDatabaseViewTransaction *interactionsByThread = [transaction ext:TSMessageDatabaseViewExtensionName];
+        [interactionsByThread enumerateGroupsUsingBlock:^(NSString *group, BOOL *stop) {
+            [threadIds addObject:group];
+        }];
+        NSMutableArray<TSInteraction *> *interactionsToDelete = [NSMutableArray new];
+        for (NSString *threadId in threadIds) {
+            [interactionsByThread enumerateKeysAndObjectsInGroup:threadId
```
michaelkirk: Do we really need to do this per thread? Can we more simply enumerate all keys and objects in the tsinteraction collection?

Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
charlesmchen: It was _much_ simpler to give DebugUIPage a simple, common interface - even for pages that don't require a thread or require a contact thread, etc.  There's lots of ways we could model this more elegantly, but I'm trying to be pragmatic since this is just Debug UI.  So right now the responsibility falls on the creator of the page to make sure it's used appropriately.  We also use asserts.

Signal/src/ViewControllers/DebugUI/DebugUIDiskUsage.m
charlesmchen: Yes, I'd like to do that.  I'd like to get this merged first.

* This PR is already big enough.
* I'd like to collect some stats from your primary device before modifying the cleaner in prod.
* We might want to rework the cleaner to run regularly as a job.

Signal/src/ViewControllers/DebugUI/DebugUIDiskUsage.m
charlesmchen: The advantage to doing this per-thread is that we can `stop` when we find a "new" message, since this view is sorted.  

This functionality too is something that'd be nice to do in a job started on launch, if/when we add a "cull messages after N days" setting.  If/when we do that in prod, this kind of perf improvement will be key.

Signal/src/ViewControllers/DebugUI/DebugUIDiskUsage.m
charlesmchen: Yup, there's a few things that'll want to do if/when we production-ize this. Thanks.

Signal/src/ViewControllers/DebugUI/DebugUIDiskUsage.m
michaelkirk: Makes sense - good call!

PR #2269, opened 2017-06-23
Don’t reply to “no longer verified”.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/environment/NotificationsManager.m
+++ b/Signal/src/environment/NotificationsManager.m
@@ -264,12 +264,20 @@
 
         switch (self.notificationPreviewType) {
             case NotificationNamePreview: {
-                BOOL senderIsNoLongerVerified =
-                    [OWSIdentityManager.sharedManager verificationStateForRecipientId:message.authorId]
-                    == OWSVerificationStateNoLongerVerified;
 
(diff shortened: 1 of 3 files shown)


Signal/src/environment/NotificationsManager.m
```diff
@@ -263,8 +263,13 @@ - (void)notifyUserForIncomingMessage:(TSIncomingMessage *)message
         }
 
         switch (self.notificationPreviewType) {
-            case NotificationNamePreview:
-                notification.category = Signal_Full_New_Message_Category;
+            case NotificationNamePreview: {
+                BOOL senderIsNoLongerVerified =
+                    [OWSIdentityManager.sharedManager verificationStateForRecipientId:message.authorId]
+                    == OWSVerificationStateNoLongerVerified;
```
michaelkirk: What if it's an incoming group message, and someone *else* is no longer verified?

charlesmchen:
PTAL @michaelkirk

Signal/src/environment/NotificationsManager.m
charlesmchen: Good point. Done.

PR #2277, opened 2017-06-30
update badge count when app is in background

michaelkirk:
see: https://github.com/WhisperSystems/SignalServiceKit/pull/288

PTAL @charlesmchen


Signal/src/ViewControllers/SignalsViewController.m
```diff
@@ -603,7 +603,6 @@ - (NSNumber *)updateInboxCountLabel {
     [_segmentedControl setTitle:unreadString forSegmentAtIndex:0];
     [_segmentedControl.superview setNeedsLayout];
     [_segmentedControl reloadInputViews];
-    [[UIApplication sharedApplication] setApplicationIconBadgeNumber:badgeNumber.integerValue];
```
michaelkirk: Since messagesManager is *always* observing and keeping applicationIconBadgeNumber up to date, we don't need to do any such thing here.

PR #2279, opened 2017-07-01
Play audible alert for missed call

michaelkirk:
Previously we weren't playing alert sounds for any of the following:

- missed call due to no longer verified
- missed call due to recent SN change
- regular ole' missed call

PTAL @charlesmchen

### DIFF
--- a/Signal/src/environment/NotificationsManager.m
+++ b/Signal/src/environment/NotificationsManager.m
@@ -96,6 +96,9 @@
         PushManagerUserInfoKeysCallBackSignalRecipientId : call.remotePhoneNumber
     };
 
+    if ([self shouldPlaySoundForNotification]) {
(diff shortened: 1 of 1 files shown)


Signal/src/environment/NotificationsManager.m
```diff
@@ -96,6 +96,9 @@ - (void)presentMissedCall:(SignalCall *)call callerName:(NSString *)callerName
         PushManagerUserInfoKeysCallBackSignalRecipientId : call.remotePhoneNumber
     };
 
+    if ([self shouldPlaySoundForNotification]) {
+        notification.soundName = @"NewMessage.aifc";
```
charlesmchen: Constantize?

michaelkirk: done!

PR #2280, opened 2017-07-01
show timestamp after unread indicator

michaelkirk:
PTAL @charlesmchen 

hotfix 2.13.2 I think.


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -1835,6 +1835,13 @@ - (BOOL)showDateAtIndexPath:(NSIndexPath *)indexPath
         id<OWSMessageData> previousMessage =
             [self messageAtIndexPath:[NSIndexPath indexPathForItem:indexPath.row - 1 inSection:indexPath.section]];
 
+        if ([previousMessage.interaction isKindOfClass:[TSUnreadIndicatorInteraction class]]) {
```
michaelkirk: This is the fix. The rest is just for fun.

PR #2281, opened 2017-07-01
remove unneccessary hack

michaelkirk:
No change in behavior, but  now that we have a nice system message design, we don't need this less-than-righteous hack.

Also, fixes assertion check added here: https://github.com/WhisperSystems/Signal-iOS/pull/2280/files#diff-7bb3586101621d594a0d0a80cf3b173bR1843

PTAL @charlesmchen

### DIFF
--- a/Signal/src/Models/OWSCall.h
+++ b/Signal/src/Models/OWSCall.h
@@ -15,9 +15,9 @@ typedef enum : NSUInteger {
     kCallMissed = 3,
     kCallOutgoingIncomplete = 4,
     kCallIncomingIncomplete = 5,
-    // kGroupUpdateJoin has been deprecated.
-    kGroupUpdateLeft = 7,
-    kGroupUpdate = 8,
+    // kGroupUpdateJoin = 6, has been deprecated.
+    // kGroupUpdateLeft = 7, has been deprecated.
+    // kGroupUpdate = 8, has been deprecated.
     kCallMissedBecauseOfChangedIdentity = 9,
 } CallStatus;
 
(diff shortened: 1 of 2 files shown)


Signal/src/Models/OWSCall.h
```diff
@@ -15,9 +15,9 @@ typedef enum : NSUInteger {
     kCallMissed = 3,
     kCallOutgoingIncomplete = 4,
     kCallIncomingIncomplete = 5,
-    // kGroupUpdateJoin has been deprecated.
-    kGroupUpdateLeft = 7,
-    kGroupUpdate = 8,
+    // kGroupUpdateJoin = 6, has been deprecated.
+    // kGroupUpdateLeft = 7, has been deprecated.
+    // kGroupUpdate = 8, has been deprecated.
```
charlesmchen: My inclination is to _not_ comment them out but to rename them instead to something like kGroupUpdateJoin_OBSOLETE.  We want to make sure that they're handled properly, and by leaving these values in the enum we'll get valuable compile-time warnings about incomplete switch statements.

Signal/src/Models/TSMessageAdapaters/TSMessageAdapter.m
```diff
@@ -224,28 +224,6 @@ + (NSCache *)displayableTextCache
...
-            // TODO: I suspect that we'll want a separate model
-            //       that conforms to <OWSMessageData> for info
-            //       messages.
-            CallStatus status = 0;
-            if (adapter.infoMessageType == TSInfoMessageTypeGroupQuit) {
-                status = kGroupUpdateLeft;
-            } else if (adapter.infoMessageType == TSInfoMessageTypeGroupUpdate) {
-                status = kGroupUpdate;
-            }
-            OWSCall *call = [[OWSCall alloc] initWithInteraction:interaction
```
charlesmchen: Aww yeah I forgot to excise this!

Signal/src/Models/OWSCall.h
charlesmchen: Sorry, I thought these were being persisted but I realize now that they're not.

charlesmchen:
LGTM

Signal/src/Models/OWSCall.h
michaelkirk: I'm OK either way - I'm interpreting your last comment to mean that you're OK with the current implementation. If not LMK and I can open a new PR.

charlesmchen:
Yup, I'm okay with current implementation.

Issue #1902, opened 2017-03-28
Bluetooth headset/car not working if you initiate the call


Deadlocklucifer:
Since the call kit integration, I'm only able to patch the call through the car if I'm receiving the call. 

Otherwise there is no way I can ascertain to connect headsets/cars.

simpowich:
I have the same problem! Latest signal and ios, using the bluetooth handsfree in an VW passat -14.

alimakki:
Similar issue with bluetooth headsets:

When initiating or receiving calls, audio isn't routed correctly through the headset.

TreatyOfRome:
A lot of bluetooth issues have been addressed in Signal 2.14.0. Issue could be closed?

michaelkirk:
Should be working as of 2.14.0.

PR #2283, opened 2017-07-03
Non-Callkit adapter plays audio from notification when app is in background

michaelkirk:
A couple of changes here - I'll annotate in line

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/CallAudioService.swift
+++ b/Signal/src/call/CallAudioService.swift
@@ -253,13 +253,13 @@ import AVFoundation
     private func play(sound: Sound) {
         guard let newPlayer = sound.player else {
             Logger.error("\(self.TAG) unable to build player")
+            assertionFailure()
             return
         }
         Logger.info("\(self.TAG) playing sound: \(sound.filePath)")
 
-        newPlayer.play()
         self.currentPlayer?.stop()
-
+        newPlayer.play()
         self.currentPlayer = newPlayer
     }
 
--- a/Signal/src/call/CallAudioService.swift
+++ b/Signal/src/call/CallAudioService.swift
@@ -258,6 +258,9 @@ import AVFoundation
         }
         Logger.info("\(self.TAG) playing sound: \(sound.filePath)")
 
+        // It's important to stop the current player **before** starting the new player. In the case that 
+        // we're playing the same sound, since the player is memoized on the sound instance, we'd otherwise 
+        // stop the sound we just started.
         self.currentPlayer?.stop()
         newPlayer.play()
         self.currentPlayer = newPlayer
--- a/Signal/src/call/CallAudioService.swift
+++ b/Signal/src/call/CallAudioService.swift
@@ -78,24 +78,27 @@ import AVFoundation
     internal func speakerphoneDidChange(call: SignalCall, isEnabled: Bool) {
         AssertIsOnMainThread()
 
-        ensureIsEnabled(call: call)
+        ensureProperAudioSession(call: call)
     }
 
     internal func hasLocalVideoDidChange(call: SignalCall, hasLocalVideo: Bool) {
         AssertIsOnMainThread()
 
-        ensureIsEnabled(call: call)
+        ensureProperAudioSession(call: call)
     }
 
-    private func ensureIsEnabled(call: SignalCall?) {
+    private func ensureProperAudioSession(call: SignalCall?) {
         guard let call = call else {
             setAudioSession(category: AVAudioSessionCategoryPlayback,
                             mode: AVAudioSessionModeDefault)
             return
         }
 
-        // Auto-enable speakerphone when local video is enabled.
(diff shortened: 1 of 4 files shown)


Signal/src/call/CallAudioService.swift
```diff
@@ -238,16 +250,16 @@ import AVFoundation
...
             Logger.error("\(self.TAG) unable to build player")
+            assertionFailure()
             return
         }
         Logger.info("\(self.TAG) playing sound: \(sound.filePath)")
 
-        newPlayer.play()
         self.currentPlayer?.stop()
-
+        newPlayer.play()
```
michaelkirk: This re-ordering is apparently critical. I'm actually not sure why.

I could imagine why it would fail if we were calling play(sound: Sound.incomingRing) multiple times - but we aren't. 

Without it, on iOS9, with the app in the foreground, the ringer would only be audible 2 in 5 times. With this re-ordering I heard proper audio 5 in 5 times.

Signal/src/call/UserInterface/CallUIAdapter.swift
```diff
@@ -63,6 +63,7 @@ extension CallUIAdaptee {
 
         guard self.callService.call == nil else {
             Logger.info("unexpectedly found an existing call when trying to start outgoing call: \(recipientId)")
+            //TODO terminate existing call.
```
michaelkirk: At some point in all my call testing, I found I was unable to place an outgoing call until the "current" call (which was invisible to the user) timed out ~2 minutes later.

So I'd hit the "call" button, and nothing would happen. I think if the user is attempting to place an outbound call we should end any current call so that the new one can go through.

@charlesmchen - if you think this is a good idea I can make a PR.

Signal/src/call/CallService.swift
```diff
@@ -857,6 +857,8 @@ protocol CallServiceObserver: class {
 
         Logger.info("\(TAG) in \(#function): \(call.identifiersForLogs).")
 
+        // TODO make call record.
```
michaelkirk: I don't think this is controversial, an upcoming PR will insert a "declined call" record.

Signal/src/call/CallAudioService.swift
```diff
@@ -142,68 +146,73 @@ import AVFoundation
 
     private func handleDialing(call: SignalCall) {
         Logger.debug("\(TAG) \(#function)")
+        AssertIsOnMainThread()
 
-        if call.isSpeakerphoneEnabled {
-            setAudioSession(category: AVAudioSessionCategoryPlayAndRecord,
-                            mode: AVAudioSessionModeVoiceChat,
-                            options: [.defaultToSpeaker, .mixWithOthers])
```
charlesmchen: `mixWithOthers` - sounds good.

Signal/src/call/CallAudioService.swift
charlesmchen: Can you leave a comment to that effect please?

Signal/src/call/CallAudioService.swift
```diff
@@ -299,12 +309,22 @@ import AVFoundation
     private func setAudioSession(category: String,
                                  mode: String? = nil,
                                  options: AVAudioSessionCategoryOptions = AVAudioSessionCategoryOptions(rawValue: 0)) {
+        
+        let session = AVAudioSession.sharedInstance()
         do {
             if #available(iOS 10.0, *), let mode = mode {
-                try AVAudioSession.sharedInstance().setCategory(category, mode: mode, options: options)
+                if session.category == category, session.mode == mode, session.categoryOptions == options {
```
charlesmchen: Nice. I've also read that calling `setCategory()` is more expensive than you'd think, so that's a nice win too.

Signal/src/call/UserInterface/CallUIAdapter.swift
charlesmchen: I think that's a great idea, but this fits into the larger pattern where CallService often ends up with a stale call.  +1 to the proposed fix, but that won't help us where incoming calls are rejected due to stale calls.  I'd really like to look at the root cause(s) of this issue.

charlesmchen:
@michaelkirk We should probably test this PR re: https://github.com/WhisperSystems/Signal-iOS/issues/1902#issuecomment-312874572

Signal/src/call/CallAudioService.swift
michaelkirk: done

michaelkirk:
> @michaelkirk We should probably test this PR re: #1902 (comment)

Updated the test doc - though our current BT integration has some issues. I'll be looking into those today and be able to more thoroughly flesh out the test plan.

charlesmchen:
Please let me know whenever you want me to run through the BT test plan on
a branch - I now have AirPods and my other BT device.

PR #2285, opened 2017-07-03
print call record when declining a call

michaelkirk:
depends on: https://github.com/WhisperSystems/SignalServiceKit/pull/290

PTAL @charlesmchen


michaelkirk:
In bf59348

PR #2286, opened 2017-07-03
network status is not tappable

michaelkirk:
PTAL @charlesmchen


Signal/src/ViewControllers/SettingsTableViewController.m
```diff
@@ -143,7 +143,7 @@ - (void)updateTableContents
             cell.textLabel.text = NSLocalizedString(@"NETWORK_STATUS_HEADER", @"");
             cell.textLabel.font = [UIFont ows_regularFontWithSize:18.f];
             cell.textLabel.textColor = [UIColor blackColor];
-
+            cell.userInteractionEnabled = NO;
```
charlesmchen: I think the right way to do this is `cell.selectionStyle = UITableViewCellSelectionStyleNone`.

michaelkirk:
PTAL @charlesmchen

PR #2291, opened 2017-07-04
Run orphan cleanup on startup.

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/SignalServiceKit/pull/291


charlesmchen:
See: WhisperSystems/SignalServiceKit#291

PR #2292, opened 2017-07-04
Charlesmchen/app update nag

charlesmchen:
PTAL @michaelkirk


Signal/src/UIApplication+OWS.swift
```diff
@@ -9,8 +9,20 @@ extension UIApplication {
     var frontmostViewController: UIViewController? {
         let window = UIApplication.shared.keyWindow
         var viewController = window!.rootViewController
-        while viewController?.presentedViewController != nil {
-            viewController = viewController?.presentedViewController
+        
+        while true {
```
charlesmchen: This slightly changes the semantics of `frontmostViewController` by returning the top-most view in a `UINavigationController` rather than the `UINavigationController` itself.  

We mostly use `frontmostViewController` to present alerts and modals, in which case this change will have no effect.  This change however allows us to discriminate the current user-visible view in cases where we need to.

Signal/src/UIApplication+OWS.swift
```diff
@@ -9,8 +9,20 @@ extension UIApplication {
...
+            if let nextViewController = viewController?.presentedViewController {
+                viewController = nextViewController
+            } else if viewController is UINavigationController {
```
michaelkirk: style nit: conditional casting removes a line and puts Optional checking in logical place.

```
 // ...
} else if let navigationViewController = viewController as? UINavigationController {
    if let nextViewController = navigationController.topViewController {
      // ...
    }
}
```

Signal/src/util/AppUpdateNag.m
```diff
@@ -0,0 +1,130 @@
...
+    const NSTimeInterval kMinute = 60.f;
+    const NSTimeInterval kHour = 60 * kMinute;
+    const NSTimeInterval kDay = 24 * kHour;
+    const NSTimeInterval kNagFrequency = kDay * 14;
+    BOOL canNag = (!lastNagDate || fabs(lastNagDate.timeIntervalSinceNow) > kNagFrequency);
+    if (!canNag) {
+        return;
+    }
+
+    // NOTE: The iTunes app store API exposes "short" version numbers, so
```
michaelkirk: Surprised to hear this - Do you have any more details? Even if they don't know our build number, we are required to increment the version with every release.

Doing a little snooping - the output of`curl https://itunes.apple.com/lookup?id=874139669`

includes the string:

```
"version":"2.13.1
```

So it seems like sufficient information exists to determine our hotfix release - but I haven't dug into ATAppUpdater to verify they are using that.

However, it seems like this might not work to notify our testflight users that there is a new build available for their currently Version (but testflight notifications mostly take care of that anyway).

charlesmchen: Dur, this actually makes sense.  Yes, the iTunes API exposes 3 (not 4) components of the version number.  I was thinking it would be an issue that the fourth (build number) component was missing, but I forgot that iTunes release tracks will only release a single build.  Please ignore!

charlesmchen: > So it seems like sufficient information exists to determine our hotfix release - but I haven't dug into ATAppUpdater to verify they are using that.

I have, its fine.

charlesmchen: > However, it seems like this might not work to notify our testflight users that there is a new build available for their currently Version

Yes, this won't work for TF. TF takes care of those notifications for us.

Signal/src/UIApplication+OWS.swift
charlesmchen: Done.

PR #2293, opened 2017-07-04
Add “new contact” and “add to existing contact” buttons in 1:1 conver…

charlesmchen:
…sation settings view.

screen shot 2017-07-04 at 4 34 14 pm

PTAL @michaelkirk


Signal/Images.xcassets/table_ic_new_contact.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
...
+    {
+      "idiom" : "universal",
+      "filename" : "table_ic_new_contact@2x.png",
+      "scale" : "2x"
+    },
+    {
+      "idiom" : "universal",
+      "filename" : "table_ic_new_contact@3x.png",
+      "scale" : "3x"
+    }
```
charlesmchen: I don't like these icons, but I spent a couple hours trying to arrive at something better and couldn't.  Happy to show alternatives if you're interested.

Signal/src/ViewControllers/OWSAddToContactViewController.m
```diff
@@ -0,0 +1,250 @@
...
+#import "ContactsViewHelper.h"
+#import "Environment.h"
+#import "OWSContactsManager.h"
+#import "UIUtil.h"
+
+@import ContactsUI;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface NamedContact : NSObject
```
charlesmchen: We use these tuples to sort by display name.

Signal/src/ViewControllers/OWSAddToContactViewController.m
```diff
@@ -0,0 +1,250 @@
...
+    [self updateTableContents];
+}
+
+- (nullable NSString *)displayNameForContact:(Contact *)contact
+{
+    OWSAssert(contact);
+
+    if (contact.fullName.length > 0) {
+        return contact.fullName;
+    }
```
charlesmchen: We only let users add a phone number to a contact that already has name, phone number or email.

charlesmchen:
screen shot 2017-07-04 at 4 59 27 pm

Signal/src/contact/OWSContactsManager.m
```diff
@@ -215,14 +215,10 @@ - (void)updateCachedDisplayNames
...
     OWSAssert([NSThread isMainThread]);
 
-    // Preserve any existing values, so that contacts that have been removed
-    // from system contacts still show up properly in the app.
-    NSMutableDictionary<NSString *, NSString *> *cachedAccountNameMap
-        = (self.cachedAccountNameMap ? [self.cachedAccountNameMap mutableCopy] : [NSMutableDictionary new]);
-    NSMutableDictionary<NSString *, NSString *> *cachedFirstNameMap
-        = (self.cachedFirstNameMap ? [self.cachedFirstNameMap mutableCopy] : [NSMutableDictionary new]);
-    NSMutableDictionary<NSString *, NSString *> *cachedLastNameMap
-        = (self.cachedLastNameMap ? [self.cachedLastNameMap mutableCopy] : [NSMutableDictionary new]);
```
charlesmchen: As I played with these views - editing contacts, adding and removing numbers to contacts - it didn't make sense at all that we didn't update our display name cache to reflect these changes.  Happy to discuss, since I suspect our instincts here about the correct behavior may be different.

charlesmchen:
PTAL @michaelkirk

Signal/src/contact/OWSContactsManager.m
michaelkirk: For testing I often want to remove someone from my contacts - and the previous caching logic prevents that. I agree that's weird. So 👍 for addressing this.

However, I believe your change introduces a regression where no-longer-registered users will show up as phone numbers instead of using their name from the address book.

I have a feeling that the deeper problem is that we should be populating these caches not by iterating SignalAccounts, but rather by iterating Contacts - since we use them to display Contact's display names (some of whom do not have a SignalAccount).

The pickle is that disambiguating multi-number contacts is done on the SignalAccount. Maybe **that** disambiguation could be moved to the Contact?

Seems like a non-trivial refactor. FWIW I have a slight preference for the former behavior, than to what's proposed here, even if it makes testing harder, because I think it will seem less broken for more users.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -319,6 +322,9 @@
 /* Indicates that this thread is muted until a given date or time. Embeds {{The date or time which the thread is muted until}}. */
 "CONVERSATION_SETTINGS_MUTED_UNTIL_FORMAT" = "until %@";
 
+/* Label for 'new contact' button in conversation settings view. */
+"CONVERSATION_SETTINGS_NEW_CONTACT" = "New Contact";
```
charlesmchen: Better.

Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -340,21 +351,49 @@ - (void)presentContactViewControllerForRecipientId:(NSString *)recipientId
...
+        CNMutableContact *updatedContact = [addToExistingCnContact mutableCopy];
+        NSMutableArray<CNLabeledValue *> *phoneNumbers
+            = (updatedContact.phoneNumbers ? [updatedContact.phoneNumbers mutableCopy] : [NSMutableArray new]);
+        // Only add recipientId as a phone number for the existing contact
+        // if its not already present.
+        BOOL hasPhoneNumber = NO;
+        for (CNLabeledValue *existingPhoneNumber in phoneNumbers) {
+            CNPhoneNumber *phoneNumber = existingPhoneNumber.value;
+            if ([phoneNumber.stringValue isEqualToString:recipientId]) {
+                hasPhoneNumber = YES;
```
charlesmchen: Done. Yeah, this was mostly about future-proofing this code since it's foreseeable that we could eventually add a "add to existing contact" action somewhere else.

Signal/src/ViewControllers/OWSAddToContactViewController.m
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        NamedContact *namedContact = [NamedContact new];
+        namedContact.contact = contact;
+        namedContact.displayName = displayName;
+        [namedContacts addObject:namedContact];
+    }
+
+    [namedContacts sortUsingComparator:^NSComparisonResult(NamedContact *_Nonnull left, NamedContact *_Nonnull right) {
+        return [left.displayName caseInsensitiveCompare:right.displayName];
```
charlesmchen: You're right.  I've taken out the logic to do local sorting here.  

> ContactsManager.allContacts is already sorted in this order, but it don't think it accounts for sorting unnamed contacts by email/phone number.

The sorting done in `ContactsFrameworkContactStoreAdaptee` works pretty well with "un-named" contacts.

Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
```diff
@@ -158,10 +160,22 @@ - (void)updateEditButton
...
 }
 
+- (BOOL)hasExistingContact
+{
+    OWSAssert([self.thread isKindOfClass:[TSContactThread class]]);
+
+    TSContactThread *contactThread = (TSContactThread *)self.thread;
+    NSString *recipientId = contactThread.contactIdentifier;
+    SignalAccount *signalAccount = [self.contactsViewHelper signalAccountForRecipientId:recipientId];
+    return signalAccount.contact;
```
charlesmchen: Done, but we test id/NSObject pointers for nil without explicitly casting them to BOOL in about a million places.

charlesmchen:
> However, I believe your change introduces a regression where no-longer-registered users will show up as phone numbers instead of using their name from the address book.

I've modified updateCachedDisplayNames to fallback to using `allContacts` to populate the the `cachedAccountNameMap` if there is no SignalAccount for a given e164 phone number.

charlesmchen:
PTAL @michaelkirk

PR #2294, opened 2017-07-05
Block device from sleeping during certain activities.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/util/DeviceSleepManager.swift
+++ b/Signal/src/util/DeviceSleepManager.swift
@@ -1,0 +1,81 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+// This entity has responsibility for blocking the device from sleeping if
+// certain behaviors (e.g. recording or playing voice messages) are in progress.
+// 
+// Sleep blocking is keyed using "block objects" whose lifetime correponds to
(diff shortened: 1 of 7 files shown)


michaelkirk:
I haven't looked closely yet, but can we also prevent sleep when running DB migrations? Or would you prefer that be apart of another PR?

charlesmchen:
PTAL @michaelkirk 

I pushed another commit: "Block device from sleeping while Debug UI is visible and during database upgrades."

Signal/src/util/DeviceSleepManager.swift
```diff
@@ -0,0 +1,81 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+// This entity has responsibility for blocking the device from sleeping if
+// certain behaviors (e.g. recording or playing voice messages) are in progress.
+// 
+// Sleep blocking is keyed using "block objects" whose lifetime correponds to
```
michaelkirk: > // Sleep blocking is keyed using "block objects" whose lifetime correponds to
> // Sleep blocking is keyed using "block objects" whose lifetime corre__s__ponds to

charlesmchen: Done.

PR #2296, opened 2017-07-05
Modify TSStorageManager  to use separate shared read and write connec…

charlesmchen:
PTAL @michaelkirk


charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/292

PR #2297, opened 2017-07-05
Move more initialization until after didBecomeActive is complete to a…

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -136,50 +136,6 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
     [self prepareScreenProtection];
 
-    // At this point, potentially lengthy DB locking migrations could be running.
-    // Avoid blocking app launch by putting all further possible DB access in async thread.
-    [[TSAccountManager sharedInstance]
-        ifRegistered:YES
-            runAsync:^{
-                DDLogInfo(
-                    @"%@ running post launch block for registered user: %@", self.tag, [TSAccountManager localNumber]);
-
-                [TSSocketManager requestSocketOpen];
-
-                RTCInitializeSSL();
-
-                [OWSSyncPushTokensJob runWithPushManager:[PushManager sharedManager]
-                                          accountManager:[Environment getCurrent].accountManager
-                                             preferences:[Environment preferences]
-                                              showAlerts:NO];
-
-                // Clean up any messages that expired since last launch immediately
-                // and continue cleaning in the background.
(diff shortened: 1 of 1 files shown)


Signal/src/AppDelegate.m
```diff
@@ -505,6 +461,56 @@ - (void)applicationDidBecomeActive:(UIApplication *)application {
...
+        // At this point, potentially lengthy DB locking migrations could be running.
+        // Avoid blocking app launch by putting all further possible DB access in async thread.
+        [[TSAccountManager sharedInstance]
+            ifRegistered:YES
+                runAsync:^{
+                    DDLogInfo(@"%@ running post launch block for registered user: %@",
+                        self.tag,
+                        [TSAccountManager localNumber]);
+
+                    [TSSocketManager requestSocketOpen];
```
michaelkirk: Can we remove this requestSocketOpen call since it's already called *every* time we didBecomeActive?

Signal/src/AppDelegate.m
```diff
@@ -505,6 +461,56 @@ - (void)applicationDidBecomeActive:(UIApplication *)application {
...
+
+                    [AppStoreRating setupRatingLibrary];
+                }];
+
+        [[TSAccountManager sharedInstance]
+            ifRegistered:NO
+                runAsync:^{
+                    dispatch_async(dispatch_get_main_queue(), ^{
+                        DDLogInfo(@"%@ running post launch block for unregistered user.", self.tag);
+                        [TSSocketManager requestSocketOpen];
```
michaelkirk: Huh, I realize you just copy/pasted this, but it seems weird to be trying to open the websocket when we're not registered. Maybe it's just to get the retry-timer going for when we complete registration.

charlesmchen: Yes, I think it also prods socket manager to observe notifications.  Clearly the method could have a better name. Open to suggestions.

Signal/src/AppDelegate.m
charlesmchen: Done.

charlesmchen:
PTAL @michaelkirk  I also reordered some operations.

Signal/src/AppDelegate.m
michaelkirk: I think you removed the wrong one. We want to requestSocketOpen every time we become active, right?

charlesmchen: You're right. Pushing a commit.

PR #2299, opened 2017-07-06
Fix missing “database upgrade” label on launch screen.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -158,10 +158,11 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     BOOL shouldShowUpgradeLabel = NO;
     NSString *lastCompletedLaunchAppVersion = AppVersion.instance.lastCompletedLaunchAppVersion;
     NSString *kLastVersionWithDatabaseViewChange = @"2.13.0";
-    BOOL mayNeedUpgrade = ([TSAccountManager isRegistered] && lastCompletedLaunchAppVersion &&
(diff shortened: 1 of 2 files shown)


Signal/src/AppDelegate.m
```diff
@@ -156,12 +156,15 @@ - (UIViewController *)loadingRootViewController
...
     NSString *lastCompletedLaunchAppVersion = AppVersion.instance.lastCompletedLaunchAppVersion;
-    BOOL mayNeedUpgrade = ([TSAccountManager isRegistered] &&
-        [VersionMigrations isVersion:lastAppVersion atLeast:@"2.0.0" andLessThan:@"2.13.0"]);
+    NSString *kLastVersionWithDatabaseViewChange = @"2.13.0";
+    BOOL mayNeedUpgrade = ([TSAccountManager isRegistered]
+        && (!lastCompletedLaunchAppVersion ||
+               [VersionMigrations isVersion:lastCompletedLaunchAppVersion
+                                    atLeast:@"2.0.0"
+                                andLessThan:kLastVersionWithDatabaseViewChange]));
     BOOL hasCompletedUpgrade = (lastCompletedLaunchAppVersion &&
```
michaelkirk: nit: Can't we remove `hasCompletedUpgrade` all together since there is already a `andLessThan` check in `mayNeedUpgrade`?

If so it might make sense to rename `mayNeedUpgrade` -> `needsUpgrade`

charlesmchen: Done.

PR #2301, opened 2017-07-06
Fix bluetooth audio for calls

michaelkirk:
With appropriate logging and now that we've consolidated AudioSession mgmt into a single method, this was pretty straight forward.

PTAL @charlesmchen


charlesmchen:
Let me know when it's merged and I'll test my two BT devices.

PR #2306, opened 2017-07-07
Mkirk/trickle ice

michaelkirk:
Based on https://github.com/WhisperSystems/Signal-iOS/pull/2305 so you can merge that one first.

Also this will likely have some merge conflicts with https://github.com/WhisperSystems/Signal-iOS/pull/2261

I don't think they are fundamentally incompatible or anything - they just touched some of the same code.

Now that incoming messages are not blocked on SN change, the caller no longer has to wait to receive a call answer before sending their ICE updates. This let's us connect calls faster.

This ended up being not quite straight forward, see my comments in line.

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -620,7 +620,9 @@ protocol CallServiceObserver: class {
                 self.peerConnectionClient = peerConnectionClient
             }
 
-            peerConnectionClient.addIceCandidate(RTCIceCandidate(sdp: sdp, sdpMLineIndex: lineIndex, sdpMid: mid))
+            let iceCandidate = RTCIceCandidate(sdp: sdp, sdpMLineIndex: lineIndex, sdpMid: mid)
+            Logger.info("\(self.TAG) peerConnectionClient added candidate with url: \(String(describing: iceCandidate.serverUrl))")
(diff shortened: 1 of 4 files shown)


Signal/src/call/CallService.swift
```diff
@@ -495,58 +470,52 @@ protocol CallServiceObserver: class {
...
         }
 
         let incomingCallPromise = firstly {
-            return getIceServers()
-        }.then { (iceServers: [RTCIceServer]) -> Promise<HardenedRTCSessionDescription> in
-            // FIXME for first time call recipients I think we'll see mic/camera permission requests here,
-            // even though, from the users perspective, no incoming call is yet visible.
-            guard self.call == newCall else {
-                throw CallError.obsoleteCall(description: "getIceServers() response for obsolete call")
+            return getPeerConnectionClientForIncomingCall(incomingCall)
```
michaelkirk: Because we can't guarantee message ordering, it's possible the callee would receive an important ICE update **before** receiving this CallOffer. Because both `handleCallOffer` and `handleRemoteAddedIceCandidate` need a peerConnectionClient to be in place,  I extracted the build-peer-connection logic behind a shared promise.

michaelkirk:
Closing this for now - I'm seeing some connectivity issues. I'll re-open when I've figure it out.

PR #2309, opened 2017-07-08
Log/Code Cleanup

michaelkirk:
- make some logging clearer
- remove noisy unhelpful logging
- clearer method names

While debugging some call connectivity issues, I tried to clean up some of our logging and code. I wanted to separate this from the actual connectivity changes so that PR can be more focused.

The log statements I removed were kind of arbitrary - don't hesitate to veto any that you'd like to keep in. Comments in line...

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/CallAudioService.swift
+++ b/Signal/src/call/CallAudioService.swift
@@ -321,8 +321,7 @@ import AVFoundation
                 let oldMode = session.mode
                 let oldOptions = session.categoryOptions
 
-                if oldCategory == category, oldMode == mode, oldOptions == options {
-                    Logger.debug("\(self.TAG) in \(#function) doing nothing, since audio session is unchanged.")
+                guard oldCategory != category || oldMode != mode || oldOptions != options else {
                     return
                 }
 
@@ -333,17 +332,15 @@ import AVFoundation
                     Logger.debug("\(self.TAG) audio session changed mode: \(oldMode) -> \(mode) ")
                 }
                 if oldOptions != options {
-                    Logger.debug("\(self.TAG) audio session changed category: \(oldOptions) -> \(options) ")
+                    Logger.debug("\(self.TAG) audio session changed options: \(oldOptions) -> \(options) ")
                 }
-
-                Logger.debug("\(self.TAG) setting new category: \(category) mode: \(mode) options: \(options)")
                 try session.setCategory(category, mode: mode, options: options)
 
             } else {
                 let oldCategory = session.category
                 let oldOptions = session.categoryOptions
-                if session.category == category, session.categoryOptions == options {
-                    Logger.debug("\(self.TAG) in \(#function) doing nothing, since audio session is unchanged.")
+
+                guard session.category != category || session.categoryOptions != options else {
                     return
                 }
 
@@ -351,10 +348,8 @@ import AVFoundation
                     Logger.debug("\(self.TAG) audio session changed category: \(oldCategory) -> \(category) ")
                 }
                 if oldOptions != options {
-                    Logger.debug("\(self.TAG) audio session changed category: \(oldOptions) -> \(options) ")
+                    Logger.debug("\(self.TAG) audio session changed options: \(oldOptions) -> \(options) ")
                 }
-
-                Logger.debug("\(self.TAG) setting new category: \(category) options: \(options)")
                 try session.setCategory(category, with: options)
 
             }
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -227,15 +227,11 @@ protocol CallServiceObserver: class {
 
     func didEnterBackground() {
         AssertIsOnMainThread()
-        Logger.info("\(self.TAG) \(#function)")
-
         self.updateIsVideoEnabled()
     }
 
     func didBecomeActive() {
         AssertIsOnMainThread()
-        Logger.info("\(self.TAG) \(#function)")
-
         self.updateIsVideoEnabled()
     }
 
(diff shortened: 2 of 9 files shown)


Signal/src/call/PeerConnectionClient.swift
```diff
@@ -444,13 +444,13 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
     }
 
-    public func addIceCandidate(_ candidate: RTCIceCandidate) {
+    public func addRemoteIceCandidate(_ candidate: RTCIceCandidate) {
         PeerConnectionClient.signalingQueue.async {
             guard self.peerConnection != nil else {
                 Logger.debug("\(self.TAG) \(#function) Ignoring obsolete event in terminated client")
                 return
             }
-            Logger.debug("\(self.TAG) adding candidate")
```
michaelkirk: Because **so** much is happening in CallService, I thought it would be useful to put all the ICE connectivity logging **just** in PeerConnectionClient. Filtering by "[PeerConnectionClient]" then gives you a succinct view of the ICE handshake without lots of noise.

Signal/src/call/WebRTCCallMessageHandler.swift
```diff
@@ -29,23 +29,20 @@ class WebRTCCallMessageHandler: NSObject, OWSCallMessageHandler {
 
     public func receivedOffer(_ offer: OWSSignalServiceProtosCallMessageOffer, from callerId: String) {
         AssertIsOnMainThread()
-        Logger.verbose("\(TAG) handling offer from caller:\(callerId)")
```
michaelkirk: These log entries are redundant with what's already in CallService.

e.g. it's not that helpful to have logs like this:
```
[CallHandler] Called Foo
[CallService] Called Foo
[CallHandler] Called Bar
[CallService] Called Bar
```

Let's just log in CallService only.

Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -471,8 +471,6 @@ class SystemContactsFetcher: NSObject {
                     return
                 }
 
-                Logger.debug("\(self.TAG) Notifying delegate that system contacts did change. hash:\(contactsHash)")
-
```
michaelkirk: There are other lines that already communicate this.

Signal/src/util/DeviceSleepManager.swift
```diff
@@ -71,7 +71,7 @@ import Foundation
 
         if UIApplication.shared.isIdleTimerDisabled != shouldBlock {
             if shouldBlock {
-                Logger.info("\(self.TAG) \(#function): Blocking sleep")
```
michaelkirk: If there's ever an issue, I imagine it could be helpful to know *what* is blocking the device from going to sleep.

Signal/src/call/CallService.swift
```diff
@@ -1205,6 +1202,10 @@ protocol CallServiceObserver: class {
     public func handleFailedCall(failedCall: SignalCall?, error: CallError) {
         AssertIsOnMainThread()
 
+        if case .assertionError(let description) = error {
+            assertionFailure(description)
```
michaelkirk: I wanted to make sure you saw this. 

We've taken a lot of care to have the CallService return to a good state when it encounters errors, but we might inadvertently be ignoring some programming errors. I'm proposing we actually crash when we hit assertion errors in the call service for DEBUG builds.

It's likely that we'll iadvertently hit a couple of debug crashes at first which may require us to reclassify the error (e.g. like I did on line 715.)

Signal/src/call/CallAudioService.swift
```diff
@@ -321,8 +321,7 @@ import AVFoundation
                 let oldMode = session.mode
                 let oldOptions = session.categoryOptions
 
-                if oldCategory == category, oldMode == mode, oldOptions == options {
-                    Logger.debug("\(self.TAG) in \(#function) doing nothing, since audio session is unchanged.")
+                guard oldCategory != category || oldMode != mode || oldOptions != options else {
```
charlesmchen: Don't you want to also make the analogous change below in the other "doing nothing, since audio session is unchanged" check?

Signal/src/call/CallService.swift
charlesmchen: +1

I'd like to add a swift free function analogous to OWSFail() that `assertionFailure` in DEBUG and DDLogError() in production.

charlesmchen: I'll add a card.

Signal/src/call/PeerConnectionClient.swift
charlesmchen: +1

Signal/src/call/WebRTCCallMessageHandler.swift
charlesmchen: +1

Signal/src/util/DeviceSleepManager.swift
charlesmchen: Sure, just bear in mind that:

* there could be multiple block objects.
* For robustness, `DeviceSleepManager` uses weak references but we still need to be very careful to remove block objects ASAP.  Weak references are only a half-measure since their expiration can't be observed.

michaelkirk: > there could be multiple block objects.

I've added a little clearer logging message in this case: `Blocking sleep because of: FOO and 2 others`

Though in practice it's unlikely to hit this. We only log when sleep-blocking changes - so adding a second blocking object will not log unless both of those objects were added at the same time. 

I've also added some asserts to make sure we're only interacting with the `blocks` state on the main thread.

PR #2310, opened 2017-07-08
Mkirk/trickle ice 2

michaelkirk:
Based on #2305 and #2309 

Feel free to review those first.

But all the changes are in a single commit: https://github.com/WhisperSystems/Signal-iOS/commit/b4700226076d9a7ae8356d436a0bea665f28fbe3

If you're feeling bold.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -157,19 +157,14 @@ protocol CallServiceObserver: class {
         }
     }
 
-    /**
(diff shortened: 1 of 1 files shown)


charlesmchen:
LGTM  I'm wondering how we can improve the "regression" section of our test plan to be more rigorous about calls until we're more confident that we're bug-free.

PR #2313, opened 2017-07-10
Add owsFail free function for swift.

charlesmchen:
PTAL @michaelkirk


Signal/src/util/OWSAsserts.swift
```diff
@@ -0,0 +1,9 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+func owsFail(_ message: String) {
+    Logger.error(message)
+    Logger.flush()
+    owsFail(message)
```
michaelkirk: search and replace fail. =P

PR #2314, opened 2017-07-10
Charlesmchen/archive view indicator

charlesmchen:
* Convert home view to programmatic layout.
* Add a reminder that you are in archive view. 
* Fix bug wherein reminder view's hot area was too small.
* Fix bug wherein "current" thread wasn't always visible when you returned to home view.

PTAL @michaelkirk 

screen shot 2017-07-10 at 4 38 24 pm

### DIFF
--- a/Signal/src/ViewControllers/SignalsViewController.m
+++ b/Signal/src/ViewControllers/SignalsViewController.m
@@ -29,9 +29,11 @@
 #import <YapDatabase/YapDatabaseViewConnection.h>
 
 #define CELL_HEIGHT 72.0f
-#define HEADER_HEIGHT 44.0f
 
-@interface SignalsViewController ()
+@interface SignalsViewController () <UITableViewDelegate, UITableViewDataSource, UIViewControllerPreviewingDelegate>
+
+@property (nonatomic) UITableView *tableView;
+@property (nonatomic) UILabel *emptyBoxLabel;
 
 @property (nonatomic) YapDatabaseConnection *editingDbConnection;
 @property (nonatomic) YapDatabaseConnection *uiDatabaseConnection;
@@ -57,8 +59,10 @@
 
 // Views
 
-@property (weak, nonatomic) IBOutlet ReminderView *missingContactsPermissionView;
-@property (weak, nonatomic) IBOutlet NSLayoutConstraint *hideMissingContactsPermissionViewConstraint;
+@property (weak, nonatomic) ReminderView *missingContactsPermissionView;
+@property (weak, nonatomic) NSLayoutConstraint *hideMissingContactsPermissionViewConstraint;
+
+@property (nonatomic) TSThread *lastThread;
 
 @end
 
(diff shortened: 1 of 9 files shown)


Signal/src/views/ReminderView.swift
```diff
@@ -51,10 +51,10 @@ class ReminderView: UIView {
         self.backgroundColor = UIColor.ows_reminderYellow()
         self.clipsToBounds = true
 
-        let container = UIView()
-
         let tapGesture = UITapGestureRecognizer(target: self, action: #selector(handleTap(gestureRecognizer:)))
-        container.addGestureRecognizer(tapGesture)
```
charlesmchen: The container is significantly smaller than the entire reminder view.

Signal/src/network/PushManager.m
```diff
@@ -155,7 +155,6 @@ - (void)application:(UIApplication *)application
                 success:^{
                     // TODO do we really want to mark them all as read?
                     [self markAllInThreadAsRead:notification.userInfo completionHandler:completionHandler];
-                    [[[[Environment getCurrent] signalsViewController] tableView] reloadData];
```
charlesmchen: This shouldn't be necessary; marking messages as read "touches" the messages' threads so the home view will get a "yap database modified" event.

charlesmchen:
PTAL @michaelkirk I'd like to do some RTL work that on top of this PR.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -643,6 +643,9 @@
 /* Call setup status label */
 "IN_CALL_TERMINATED" = "Call Ended.";
 
+/* Label reminding the user that they are in archive mode. */
+"INBOX_VIEW_ARCHIVE_MODE_REMINDER" = "You're in archive mode.  Tap to return to your Inbox.";
```

Signal/src/ViewControllers/SignalsViewController.m
```diff
@@ -299,6 +365,30 @@ - (void)viewWillAppear:(BOOL)animated {
 
     self.isViewVisible = YES;
     [self checkIfEmptyView];
+
+    // When returning to home view, try to ensure that the "last" thread is still
+    // visible.  The threads often change ordering while in conversation view due
+    // to incoming & outgoing messages.
+    if (self.lastThread) {
```
charlesmchen: Better, thanks.

Signal/src/ViewControllers/SignalsViewController.m
```diff
@@ -695,14 +787,18 @@ - (void)presentViewControllerWithBlock:(void (^)())presentationBlock animateDism
     }
 }
 
-#pragma mark - IBAction
+#pragma mark - Groupings
 
-- (IBAction)selectedInbox:(id)sender {
+- (void)selectedInbox
+{
+    self.segmentedControl.selectedSegmentIndex = 0;
```
charlesmchen: Previously the only way to switch between these "groups" was through the segmented control, so it wasn't necessary to sync it with these changes. Now we can switch from the "reminder" that I've just added.

michaelkirk: ah right. Makes sense.

PR #2316, opened 2017-07-11
fix tests

michaelkirk:
### Fix build failures

- Update old method signatures

### Fix failing test

- Use known interaction type for message bubble tests

### Disable singleton asserts in tests

- Introduce new build configuration ("Test") for Signal.xcproject based on Debug, which defines the SSK_BUILD_FOR_TESTS preprocessor definition.
- Use that new build configuration when building tests.
- define the SSK_BUILD_FOR_TESTS for the SSK when using the Test configuration.

Standard disclaimer on introducing a new build configuration - any time we make a change to our Debug configuration, we'll probably want to duplicate that change to the Test configuration.

### update CI / use fastlane

- get on xcode 8.3
- run tests with fastlane scan, which I find to be more concise and legible than native xctest and easier to keep build scripts up to date (e.g. picking which device to run on).

PTAL @charlesmchen


michaelkirk:
merged in https://github.com/WhisperSystems/Signal-iOS/commit/195f124afe4a83df38fb7da3214df2167d7b900e

PR #2317, opened 2017-07-11
Partial revert of: Send ICE updates immediately - 2dcfb4e3b8b07bb2b54…

michaelkirk:
Partial revert of: Send ICE updates immediately - 2dcfb4e3b8b07bb2b5419cad6fa324e5e585c3da

That commit made it more likely that the remote client will receive an
ICE update **before** the call offer. That commit also made clients robust to receiving them, but none of the clients in the wild have that robustness fix.

We want to ship the robustness fix (waitForPeerConnection in
handleRemoteAddedIceCandidate) before we have the caller start sending
ICE updates immediately. 

In a future release, when we're ready to have the caller ship ICE updates immediately, we can simply revert this commit.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -160,6 +160,17 @@ protocol CallServiceObserver: class {
     // Used to coordinate promises across delegate methods
     private var fulfillCallConnectedPromise: (() -> Void)?
 
+    /**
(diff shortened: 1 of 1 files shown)


michaelkirk:
Going to make one more change. Closing this for now.

PR #2318, opened 2017-07-11
call connectivity fixes

michaelkirk:
replaces https://github.com/WhisperSystems/Signal-iOS/pull/2317

Make sure we don't send ICE updates until *after* sending the session description (call offer / call answer).

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -160,6 +160,17 @@ protocol CallServiceObserver: class {
     // Used to coordinate promises across delegate methods
     private var fulfillCallConnectedPromise: (() -> Void)?
 
+    /**
+     * In the process of establishing a connection between the clients (ICE process) we must exchange ICE updates.
+     * Because this happens via Signal Service it's possible the callee user has not accepted any change in the caller's 
+     * identity. In which case *each* ICE update would cause an "identity change" warning on the callee's device. Since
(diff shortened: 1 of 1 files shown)


Signal/src/call/CallService.swift
```diff
@@ -304,6 +324,10 @@ protocol CallServiceObserver: class {
                 throw CallError.obsoleteCall(description:"obsolete call in \(#function)")
             }
 
+            // For outgoing calls, wait until call offer is sent before we send any ICE updates, to ensure message ordering for
+            // clients that don't support receiving ICE updates before receiving the call offer.
+            self.readyToSendIceUpdates(call: call)
+
```
michaelkirk: This is partially redundant with the re-introduced "wait for CallAnswer before sending Ice updates" but we'll want to keep this part of the solution in place even when we get rid of the "wait for CallAnswer" behavior to ensure message ordering.

Signal/src/call/CallService.swift
```diff
@@ -260,6 +277,9 @@ protocol CallServiceObserver: class {
 
         self.call = call
 
+        sendIceUpdatesImmediately = false
+        pendingIceUpdateMessages = []
```
michaelkirk: These were re-introduced to maintain compatibility with legacy clients.

michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/55ddce79bf54de963733eed568a34f0a89f53faf

PR #2320, opened 2017-07-12
Charlesmchen/right to left

charlesmchen:
* Fixes the most glaring problems with the Arabic localization (l10n).  

See: https://github.com/WhisperSystems/SignalServiceKit/pull/300

What is done:

* Doing a pass through the common states of most views and fixing the obvious issues.

An incomplete list of what's _not_ yet done:

* Hebrew l10n and other RTL languages.
* Fixing icons - especially those with asymmetry, e.g. padding for toolbar buttons.
* Time duration formatting.
* Auditing every single piece of code / every single state of every view.  
* Message bubble tails point the wrong way on RTL.

PTAL @michaelkirk 

I had to change a bunch of l10n strings, so I suggest we _not_ include this in the forthcoming v2.14 release.  By waiting for v2.15 we'll give our translators a couple of weeks to fix the l10n regressions.

### Conversation View

screen shot 2017-07-12 at 10 57 59 am

### Conversation Settings View

img_0023

### Import With Signal View

img_0022 1

### Home View

screen shot 2017-07-12 at 10 58 04 am

### Verification View

screen shot 2017-07-11 at 10 45 48 am

### Country View

screen shot 2017-07-11 at 10 42 50 am

### Registration View

reg

### Select Recipient View

screen shot 2017-07-11 at 10 14 50 am

### Settings Views

screen shot 2017-07-11 at 10 18 57 am
screen shot 2017-07-11 at 10 18 21 am


Signal.xcodeproj/project.pbxproj
```diff
@@ -443,7 +443,6 @@
 		34B3F8481E8DF1700035BE1A /* FullImageViewController.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = FullImageViewController.m; sourceTree = "<group>"; };
 		34B3F8491E8DF1700035BE1A /* InboxTableViewCell.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = InboxTableViewCell.h; sourceTree = "<group>"; };
 		34B3F84A1E8DF1700035BE1A /* InboxTableViewCell.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = InboxTableViewCell.m; sourceTree = "<group>"; };
-		34B3F84B1E8DF1700035BE1A /* InboxTableViewCell.xib */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = file.xib; path = InboxTableViewCell.xib; sourceTree = "<group>"; };
```
charlesmchen: Eliminated one .xib.

Signal/src/UIView+OWS.h
```diff
@@ -60,6 +60,31 @@ CGFloat ScaleFromIPhone5(CGFloat iPhone5Value);
...
+- (NSLayoutConstraint *)autoPinLeadingToView:(UIView *)view;
+- (NSLayoutConstraint *)autoPinLeadingToView:(UIView *)view margin:(CGFloat)margin;
+- (NSLayoutConstraint *)autoPinTrailingToView:(UIView *)view;
+- (NSLayoutConstraint *)autoPinTrailingToView:(UIView *)view margin:(CGFloat)margin;
+// Return Right on LTR and Right on RTL.
+- (NSTextAlignment)textAlignmentUnnatural;
+// Leading and trailing anchors honor layout margins.
+// When using a UIView as a "div" to structure layout, we don't want it to have margins.
++ (UIView *)containerView;
+- (void)setHLayoutMargins:(CGFloat)value;
```
charlesmchen: Added a slew of helper methods to make RTL-enabled layout easier, more concise and more readable.

There's a tricky thing here.  I've tried to write these methods so that the "margin" constant makes sense in normal, left-to-right layout.  In some cases that requires these methods to invert the actual value .

Signal/src/UIView+OWS.h
```diff
@@ -60,6 +60,31 @@ CGFloat ScaleFromIPhone5(CGFloat iPhone5Value);
...
+- (NSLayoutConstraint *)autoPinTrailingToSuperView;
+- (NSLayoutConstraint *)autoPinTrailingToSuperViewWithMargin:(CGFloat)margin;
+- (NSLayoutConstraint *)autoPinLeadingToTrailingOfView:(UIView *)view;
+- (NSLayoutConstraint *)autoPinLeadingToTrailingOfView:(UIView *)view margin:(CGFloat)margin;
+- (NSLayoutConstraint *)autoPinLeadingToView:(UIView *)view;
+- (NSLayoutConstraint *)autoPinLeadingToView:(UIView *)view margin:(CGFloat)margin;
+- (NSLayoutConstraint *)autoPinTrailingToView:(UIView *)view;
+- (NSLayoutConstraint *)autoPinTrailingToView:(UIView *)view margin:(CGFloat)margin;
+// Return Right on LTR and Right on RTL.
+- (NSTextAlignment)textAlignmentUnnatural;
```
charlesmchen: The "default" text alignment is left in LTR and right in RTL.  In some cases you want the opposite.  I've called that "Unnatural" - I couldn't think of a better name.

Signal/src/UIView+OWS.h
```diff
@@ -60,6 +60,31 @@ CGFloat ScaleFromIPhone5(CGFloat iPhone5Value);
 
 - (void)centerOnSuperview;
 
+#pragma mark - RTL
+
+// For correct right-to-left layout behavior, use "leading" and "trailing",
+// not "left" and "right".
+//
+// NOTE: the margin values are inverted in RTL layouts.
+- (BOOL)isRTL;
```
charlesmchen: Ideally, we'll use the `isRTL` method as little as possible.

Signal/src/ViewControllers/AdvancedSettingsTableViewController.m
```diff
@@ -204,25 +204,18 @@ - (void)updateTableContents
                                                          actionBlock:nil]];
 
     if (OWSSignalService.sharedInstance.isCensorshipCircumventionManuallyActivated) {
-        [censorshipSection addItem:[OWSTableItem itemWithCustomCellBlock:^{
```
charlesmchen: I've used this PR to consolidate around the `OWSTableItem` factory methods which DRY up a ton of code.

charlesmchen:
@michaelkirk There's a lot of whitespace changes, so I suggest you review ignoring whitespace:

https://github.com/WhisperSystems/Signal-iOS/pull/2320/files?w=1

Signal/src/ViewControllers/ViewControllerUtils.m
```diff
@@ -86,6 +86,44 @@ + (void)setAudioIgnoresHardwareMuteSwitch:(BOOL)shouldIgnore
     }
 }
 
++ (NSString *)examplePhoneNumberForCountryCode:(NSString *)countryCode callingCode:(NSString *)callingCode
```
charlesmchen: Moved the method from below to group the "formatting" methods together.

Signal/src/ViewControllers/ViewControllerUtils.m
```diff
@@ -86,6 +86,44 @@ + (void)setAudioIgnoresHardwareMuteSwitch:(BOOL)shouldIgnore
...
+                    @"A format for a label showing an example phone number. Embeds {{the example phone number}}."),
+            [examplePhoneNumber substringFromIndex:callingCode.length]];
+    } else {
+        return @"";
+    }
+}
+
+#pragma mark - Formatting
+
++ (NSString *)formatInt:(int)value
```
charlesmchen: printf-style formatting (e.g. %d) is not localized.  There's still probably a ton of places I haven't fixed yet.

Signal/src/ViewControllers/OWSTableViewController.h
```diff
@@ -72,6 +72,10 @@ typedef UITableViewCell *_Nonnull (^OWSTableCustomCellBlock)();
 
 + (OWSTableItem *)actionItemWithText:(NSString *)text actionBlock:(nullable OWSTableActionBlock)actionBlock;
 
++ (OWSTableItem *)softCenterLabelItemWithText:(NSString *)text;
```
charlesmchen: Perhaps not a great name, but these "soft, centered label" items are a common part of our design language in our tables.

Signal/src/ViewControllers/InboxTableViewCell.m
```diff
@@ -184,68 +272,11 @@ - (NSAttributedString *)dateAttributedString:(NSDate *)date {
     return attributedString;
 }
 
-- (void)setUnreadMsgCount:(NSUInteger)unreadMessages {
```
charlesmchen: We were using a pretty elaborate method to render the unread badges in the home view cells.  I replaced this with something much simpler (see above).

Signal/src/ViewControllers/InboxTableViewCell.m
```diff
@@ -3,18 +3,14 @@
 //
 
 #import "InboxTableViewCell.h"
-#import "Environment.h"
 #import "OWSAvatarBuilder.h"
-#import "OWSContactAvatarBuilder.h"
-#import "PropertyListPreferences.h"
 #import "Signal-Swift.h"
```
charlesmchen: There was a lot of stale cruft in this source file and its header.

Signal/src/contact/OWSContactsManager.m
```diff
@@ -360,12 +361,13 @@ + (NSString *)accountLabelForContact:(Contact *)contact recipientId:(NSString *)
     if (phoneNumbersWithTheSameName.count > 1) {
         NSUInteger index =
             [[phoneNumbersWithTheSameName sortedArrayUsingSelector:@selector(compare:)] indexOfObject:recipientId];
+        NSString *indexText = [ViewControllerUtils formatInt:(int)index + 1];
```
charlesmchen: Fixed a tiny bug here.  If a user had two phone numbers, both "home", they would show up as "home 0" and "home 1".  These indexes should start at 1.

michaelkirk:
> I suggest we not include this in the forthcoming v2.14 release. By waiting for v2.15 we'll give our translators a couple of weeks to fix the l10n regressions.

Sounds good. As this is a jumbo PR (understandably) I'm going to wait to review until my 2.14 work is done.

charlesmchen:
PTAL @michaelkirk

Signal/src/ViewControllers/CallViewController.swift
```diff
@@ -450,18 +451,18 @@ class CallViewController: UIViewController, CallObserver, CallServiceObserver, R
...
 
-            localVideoView.autoPinEdge(toSuperviewEdge:.right, withInset:videoPreviewHMargin)
+            localVideoView.autoPinTrailingToSuperView(withMargin: videoPreviewHMargin)
             localVideoView.autoPinEdge(toSuperviewEdge:.top, withInset:topMargin)
             let localVideoSize = ScaleFromIPhone5To7Plus(80, 100)
             localVideoView.autoSetDimension(.width, toSize:localVideoSize)
             localVideoView.autoSetDimension(.height, toSize:localVideoSize)
 
             contactNameLabel.autoPinEdge(toSuperviewEdge:.top, withInset:topMargin)
-            contactNameLabel.autoPinEdge(toSuperviewEdge:.left, withInset:contactHMargin)
```
michaelkirk: Caller name and duration label seem a little too close to the left edge now. (in LTR)

screen shot 2017-07-20 at 11 59 26 am

Signal/src/views/ContactTableViewCell.m
```diff
@@ -54,24 +54,26 @@ + (CGFloat)rowHeight
 
 - (void)configureProgrammatically
 {
+    self.preservesSuperviewLayoutMargins = YES;
```
michaelkirk: Appreciate the numerous little cleanups like this. 👍

Signal/src/views/ContactTableViewCell.m
```diff
@@ -192,10 +194,11 @@ - (void)layoutSubviews
         OWSAssert(self.nameLabel.superview == self.contentView);
         const CGFloat kSubtitleVMargin
             = ((self.contentView.height - self.nameLabel.font.lineHeight) * 0.5f - self.subtitle.height) * 0.5f;
-        self.subtitle.frame = CGRectMake(self.nameLabel.left,
-            round((self.contentView.height - self.subtitle.height) - kSubtitleVMargin),
-            self.subtitle.width,
-            self.subtitle.height);
+        self.subtitle.frame
```
michaelkirk: Would it be simpler to align `subtitle.leading` with the `nameLabel.leading`?

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -794,17 +794,17 @@ - (void)createBannerWithTitle:(NSString *)title bannerColor:(UIColor *)bannerCol
     [bannerView addSubview:closeButton];
     const CGFloat kBannerCloseButtonPadding = 8.f;
     [closeButton autoPinEdgeToSuperviewEdge:ALEdgeTop withInset:kBannerCloseButtonPadding];
-    [closeButton autoPinEdgeToSuperviewEdge:ALEdgeRight withInset:kBannerCloseButtonPadding];
```
michaelkirk: "Close" button is outside the banner in LTR

screen shot 2017-07-20 at 12 25 25 pm

Both LTR and RTL look fine if you take the original and only change `ALEdgeRight` -> `ALEdgeTrailing`:

e.g. 
```
[closeButton autoPinEdgeToSuperviewEdge:ALEdgeTrailing withInset:kBannerCloseButtonPadding];
```

Signal/src/UIView+OWS.h
michaelkirk: How is the behavior of this method different from the similar existing API?

e.g. 

```
callStatusLabel.autoPinLeadingToSuperViewWithMargin(contactHMargin)
```

vs. the seemingly similar

```
callStatusLabel.autoPinEdge(toSuperViewEdge:.leading, withInset:contactHMargin)
```

I'm not convinced yet that we need these...  Can you point me to a piece of code where we use these new helpers where it can't be reasonably expressed with the existing autoLayout api? 

I'm not against adding helpers if they solve new problems or are much more readable, but my preference is to rally around a smaller consistent API if reasonable. 

> There's a tricky thing here. I've tried to write these methods so that the "margin" constant makes sense in normal, left-to-right layout. In some cases that requires these methods to invert the actual value .

In fact, do we actually need to handle flipping the constant? When using leading/trailing it seems like autoLayout is mostly doing the right thing already. At least in a couple places I checked, when we manually handle flipping the constant it seems like things go wrong.

e.g. see my comments re: the "close" button on the block banner and (and looks like the same thing with the spinner being off on the "submit code" button during registration.)

It's possible that I just hit a couple of bugs, but I haven't yet seen the case where these are useful.

Signal/src/ViewControllers/OWSTableViewController.h
michaelkirk: nonsense, `tis a fine name.

Signal/src/UIView+OWS.h
charlesmchen: > I'm not convinced yet that we need these... Can you point me to a piece of code where we use these new helpers where it can't be reasonably expressed with the existing autoLayout api?

Our custom methods are more concise - easier to read and easier to auto-complete than existing methods - and I _love_ small improvements in readability of layout code which is so hard to read.  

But the real difference is that our methods use `layoutMarginsGuide` anchors, which behave differently than the PureLayout alternatives you indicated.  Honoring `layoutMargins` is particularly important in cell layouts, where it lets us align with the complicated built-in behavior of table and collection view cells' default contents.

charlesmchen: I've added a comment to that effect.

Signal/src/ViewControllers/CallViewController.swift
charlesmchen: Okay, I've widened this up a bit.  I made a few small changes like this as I converted things - the idea was by honoring layoutMargins instead of our arbitrary margin/spacing values, we could feel a little bit more idiomatic to iOS.  But I agree, in this case, it's too much.

Signal/src/ViewControllers/CallViewController.swift
```diff
@@ -552,13 +553,13 @@ class CallViewController: UIViewController, CallObserver, CallServiceObserver, R
         var constraints: [NSLayoutConstraint] = []
 
         if localVideoView.isHidden {
-            let contactHMargin = CGFloat(30)
-            constraints.append(contactNameLabel.autoPinEdge(toSuperviewEdge:.right, withInset:contactHMargin))
-            constraints.append(callStatusLabel.autoPinEdge(toSuperviewEdge:.right, withInset:contactHMargin))
+            let contactHMargin = CGFloat(0)
```
charlesmchen: Changed.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
charlesmchen: This was a bug in the custom leading/trailing methods.  Trailing margin constants need to be inverted.

Signal/src/UIView+OWS.h
charlesmchen: > In fact, do we actually need to handle flipping the constant? When using leading/trailing it seems like autoLayout is mostly doing the right thing already

You're right!  We do need to invert the constants in some cases, but we don't need to special case RTL.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -1081,8 +1081,9 @@ - (void)setBarButtonItemsForDisappearingMessagesConfiguration:
     // This method gets called multiple times, so it's important we re-layout the unread badge
     // with respect to the new backItem.
     [backItem.customView addSubview:_backButtonUnreadCountView];
+    // TODO:
```
charlesmchen: Added a comment.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -1163,7 +1164,9 @@ - (void)setBarButtonItemsForDisappearingMessagesConfiguration:
             + kTitleVSpacing);
     self.navigationBarTitleLabel.frame
         = CGRectMake(0, 0, titleViewWidth, self.navigationBarTitleLabel.frame.size.height);
-    self.navigationBarSubtitleLabel.frame = CGRectMake(0,
+    self.navigationBarSubtitleLabel.frame = CGRectMake((self.view.isRTL ? self.navigationBarTitleView.frame.size.width
```
charlesmchen: Yes it would but iOS auto layout doesn't work properly inside custom bar button items.

Signal/src/ViewControllers/ConversationView/OWSMessagesInputToolbar.m
```diff
@@ -126,9 +143,9 @@ - (void)showVoiceMemoUI
     [whiteIconView autoCenterInSuperview];
 
     [imageView autoVCenterInSuperview];
-    [imageView autoPinEdgeToSuperviewEdge:ALEdgeLeft withInset:10];
+    [imageView autoPinLeadingToSuperViewWithMargin:10.f];
     [self.recordingLabel autoVCenterInSuperview];
-    [self.recordingLabel autoPinEdge:ALEdgeLeft toEdge:ALEdgeRight ofView:imageView withOffset:5.f];
+    [self.recordingLabel autoPinLeadingToTrailingOfView:imageView margin:5.f];
```
charlesmchen: See above - we want to honor layout margins, etc.

Signal/src/ViewControllers/ConversationView/OWSMessagesToolbarContentView.m
```diff
@@ -142,7 +142,7 @@ - (void)handleLongPress:(UIGestureRecognizer *)sender
             if (self.isRecordingVoiceMemo) {
                 // Check for "slide to cancel" gesture.
                 CGPoint location = [sender locationInView:self];
-                CGFloat offset = MAX(0, self.voiceMemoGestureStartLocation.x - location.x);
+                CGFloat offset = fabs(self.voiceMemoGestureStartLocation.x - location.x);
```
charlesmchen: Done.

Signal/src/ViewControllers/InboxTableViewCell.m
```diff
@@ -3,18 +3,14 @@
 //
```
charlesmchen: Fixed.

Signal/src/ViewControllers/MessageComposeTableViewController.m
```diff
@@ -76,6 +76,7 @@ - (void)loadView
         [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemStop
                                                       target:self
                                                       action:@selector(dismissPressed)];
+    // TODO:
```
charlesmchen: Added a comment.

Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
```diff
@@ -255,18 +255,20 @@ - (void)updateTableContents
...
             UITableViewCell *cell = [UITableViewCell new];
+            cell.preservesSuperviewLayoutMargins = YES;
+            cell.contentView.preservesSuperviewLayoutMargins = YES;
             cell.selectionStyle = UITableViewCellSelectionStyleNone;
 
-            UIView *topView = [UIView new];
+            UIView *topView = [UIView containerView];
             [cell.contentView addSubview:topView];
-            [topView autoPinWidthToSuperview];
+            [topView autoPinLeadingAndTrailingToSuperview];
```
charlesmchen: Yes - it honors layout margins.

Signal/src/UIView+OWS.h
michaelkirk: > But the real difference is that our methods use layoutMarginsGuide anchors, which behave differently than the PureLayout alternatives you indicated. Honoring layoutMargins is particularly important in cell layouts, where it lets us align with the complicated built-in behavior of table and collection view cells' default contents.

PureLayout also offers a flavor of functions that honors margins rather than edges. 

e.g.  `- autoPinEdgeToSuperviewMargin:(withInset:)`

How are these new methods different?

My basis for skepticism is that RTL is not some fringe thing and PureLayout is not some fringe thing. I'd be surprised if they didn't have built in the tools for solving these problems - but of course it's possible! Thanks for bearing with me. I'm less familiar with the problem space and want to make sure we don't go all in on a "Weird Signal Way" of doing things if there's already an established convention.

Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
```diff
@@ -255,18 +255,20 @@ - (void)updateTableContents
...
             [topView autoPinEdgeToSuperviewEdge:ALEdgeTop];
             [topView autoSetDimension:ALDimensionHeight toSize:kOWSTable_DefaultCellHeight];
 
             UIImageView *iconView = [self viewForIconWithName:@"table_ic_hourglass"];
             [topView addSubview:iconView];
             [iconView autoVCenterInSuperview];
-            [iconView autoPinEdgeToSuperviewEdge:ALEdgeLeft withInset:16.f];
+            [iconView autoPinLeadingToSuperView];
```
charlesmchen: Yes, this is the glory of using layout margins - by honoring the cell's default layout margins, we align with default cell contents which is something we were doing a terrible job of before.  e.g. this view was misaligned on large "plus" devices.

Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
```diff
@@ -443,7 +450,8 @@ - (void)updateTableContents
         statusLabel.text = muteStatus;
         [cell.contentView addSubview:statusLabel];
         [statusLabel autoVCenterInSuperview];
-        [statusLabel autoPinEdgeToSuperviewEdge:ALEdgeRight];
+        //        [statusLabel autoPinLeadingToTrailingOfView:rowLabel margin:weakSelf.iconSpacing];
```
charlesmchen: Removed. And I recently learned that there's a better way to do this (have about an accessory type _AND_ an accessory label.  You use `UITableViewCellStyleValue11 and `detailTextLabel`.

Signal/src/ViewControllers/OWSConversationSettingsTableViewController.m
```diff
@@ -530,15 +546,15 @@ - (UIView *)firstSectionHeader
     _avatarView = avatarView;
     [threadInfoView addSubview:avatarView];
     [avatarView autoVCenterInSuperview];
```
charlesmchen: Done.

Signal/src/ViewControllers/SelectRecipientViewController.m
```diff
@@ -463,11 +465,11 @@ - (void)updateTableContents
 
         UILabel *countryCodeLabel = self.countryCodeLabel;
         [countryRow addSubview:countryCodeLabel];
-        [countryCodeLabel autoPinEdgeToSuperviewEdge:ALEdgeLeft withInset:20.f];
```
charlesmchen: See above - we're now using layout margins to properly align cell contents.  And it's easier, as you can see - way less constants.  "Cutting with the grain vs. cutting against the grain," so to speak.

Podfile
```diff
@@ -9,8 +9,8 @@ target 'Signal' do
     #pod 'JSQMessagesViewController',   path: '../JSQMessagesViewController'
     pod 'PureLayout'
     pod 'Reachability'
-    pod 'SignalServiceKit',           git: 'https://github.com/WhisperSystems/SignalServiceKit.git'
```
charlesmchen: See: WhisperSystems/SignalServiceKit#300

Signal/src/ViewControllers/CodeVerificationViewController.m
```diff
@@ -180,8 +181,8 @@ - (void)createViews {
     [_submitCodeSpinner autoSetDimension:ALDimensionWidth toSize:kSpinnerSize];
     [_submitCodeSpinner autoSetDimension:ALDimensionHeight toSize:kSpinnerSize];
     [_submitCodeSpinner autoVCenterInSuperview];
-    [_submitCodeSpinner autoPinEdgeToSuperviewEdge:ALEdgeRight withInset:kSpinnerSpacing];
-    
+    [_submitCodeSpinner autoPinTrailingToSuperViewWithMargin:kSpinnerSpacing];
```
charlesmchen: Fixed.  See above - you were right, the trailing methods needs to invert, not invert-only-if-RTL their constants.

Signal/src/views/ContactTableViewCell.m
charlesmchen: I was doing custom layout to get the vertical positioning exactly right.  But you're right, there was a simpler way.  Thanks.

charlesmchen:
PTAL @michaelkirk

Signal/src/UIView+OWS.h
charlesmchen: > - autoPinEdgeToSuperviewMargin:(withInset:)

Where do you see this method?  I see:

> - (NSLayoutConstraint *)autoPinEdgeToSuperviewMargin:(ALEdge)edge;

> - (NSLayoutConstraint *)autoPinEdgeToSuperviewMargin:(ALEdge)edge relation:(NSLayoutRelation)relation;

michaelkirk: Ah, you're right, the margin flavor doesn't exist. I was confused by the documentation which says:

```
- autoPinEdgeToSuperview(Edge:|Margin:)(withInset:) // Margin variant iOS 8.0+ only
```

It seems like "usually" you don't want an offset from the margin, as the whole "glory of margins" is that containers specify where their inner content should go. But I see where we are doing weird things like making our own back button, it's quick to be able to say "fudge this by an extra 10 points".

Thanks for answering my questions on this. If you are aware of what's available in PureLayout and still think your approach is superior to, for example, just adding one new method that parameterizes margin inset, then I trust it's a reasonable decision.

PR #2323, opened 2017-07-13
Charlesmchen/refine call service

charlesmchen:
* Converted `call` and `thread` properties of `CallViewController` from var to let.
* Cleared fulfill/reject properties in `CallService` as they are fulfilled or rejected.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -304,8 +304,6 @@ protocol CallServiceObserver: class {
             Logger.debug("\(self.TAG) setting peerConnectionClient in \(#function) for call: \(call.identifiersForLogs)")
             self.peerConnectionClient = peerConnectionClient
             self.fulfillPeerConnectionClientPromise?()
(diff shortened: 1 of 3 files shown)


Signal/src/call/CallService.swift
```diff
@@ -376,6 +378,8 @@ protocol CallServiceObserver: class {
         }
 
         fulfillReadyToSendIceUpdatesPromise()
+        self.fulfillReadyToSendIceUpdatesPromise = nil
```
michaelkirk: Why do we bother to nil all these promise methods out? I don't *think* there's harm in removing them, but we already have a central place to `reset` state - in `terminate`.

What's the danger in these lingering?

charlesmchen: Yeah, perhaps its a waste of time.  PromiseKit should ensure that its safe to:

* fulfill a promise twice.
* fulfill and reject a promise, in any order.

Removed!

PR #2324, opened 2017-07-13
[SSK] Try applying national prefixes when parsing phone numbers.

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/SignalServiceKit/pull/286

### DIFF
--- a/Signal/src/ViewControllers/SelectRecipientViewController.m
+++ b/Signal/src/ViewControllers/SelectRecipientViewController.m
@@ -376,11 +376,17 @@ NSString *const kSelectRecipientViewControllerCellIdentifier = @"kSelectRecipien
     }
     NSString *possiblePhoneNumber =
         [self.callingCode stringByAppendingString:self.phoneNumberTextField.text.digitsOnly];
-    PhoneNumber *parsedPhoneNumber = [PhoneNumber tryParsePhoneNumberFromUserSpecifiedText:possiblePhoneNumber];
+    NSArray<PhoneNumber *> *parsePhoneNumbers =
+        [PhoneNumber tryParsePhoneNumbersFromsUserSpecifiedText:possiblePhoneNumber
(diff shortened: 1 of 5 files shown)


Signal/src/call/OutboundCallInitiator.swift
```diff
@@ -26,7 +26,7 @@ import Foundation
     public func initiateCall(handle: String) -> Bool {
         Logger.info("\(TAG) in \(#function) with handle: \(handle)")
 
-        guard let recipientId = PhoneNumber(fromUserSpecifiedText: handle)?.toE164() else {
```
michaelkirk: I believe this could be user specified text when calling from the contacts app. We could push that parsing farther up the chain though.

It would be nice to use a type system to distinguish between "a bunch of text" and "a valid signal identitifer"

Signal/src/ViewControllers/SelectRecipientViewController.m
```diff
@@ -376,11 +376,17 @@ - (BOOL)hasValidPhoneNumber
     }
     NSString *possiblePhoneNumber =
         [self.callingCode stringByAppendingString:self.phoneNumberTextField.text.digitsOnly];
-    PhoneNumber *parsedPhoneNumber = [PhoneNumber tryParsePhoneNumberFromUserSpecifiedText:possiblePhoneNumber];
+    NSArray<PhoneNumber *> *parsePhoneNumbers =
+        [PhoneNumber tryParsePhoneNumbersFromsUserSpecifiedText:possiblePhoneNumber
+                                              clientPhoneNumber:[TSAccountManager localNumber]];
```
michaelkirk: This is one of those situations where an innocuous sounding helper method is going to create a db transaction (readonly).

I know it's not elegant, but memoizing `[StorageManager localNumber]` in `TSAccountManager` might be some good diligence.

take it or leave it.

PR #2327, opened 2017-07-13
Use recommended approach for speakerphone mode

michaelkirk:
From: https://developer.apple.com/library/content/qa/qa1754/_index.html

> Q:  Can you explain the difference between calling the AVAudioSession method overrideOutputAudioPort: with the value AVAudioSessionPortOverrideSpeaker and using the category option AVAudioSessionCategoryOptionDefaultToSpeaker with setCategory:withOptions:error:.
>
> A: The difference is that setting the AVAudioSessionPortOverride by calling overrideOutputAudioPort: is more transient than using the category option AVAudioSessionCategoryOptionDefaultToSpeaker.
>
> Important: The use of both AVAudioSessionPortOverrideSpeaker and AVAudioSessionCategoryOptionDefaultToSpeaker are only applicable to the AVAudioSessionCategoryPlayAndRecord category.
Calling overrideOutputAudioPort: and setting the AVAudioSessionPortOverride to AVAudioSessionPortOverrideSpeaker is a way of temporarily overriding the output to play to the speaker. Any route change or interruption will cause the audio to be routed back to its normal route, following the last-in wins rule. Think of using overrideOutputAudioPort: in terms of what you might use to implement a Speakerphone button where you want to be able to toggle between the speaker (AVAudioSessionPortOverrideSpeaker) and the normal output route (AVAudioSessionPortOverrideNone).
>
> Note: This property is intended to allow 3rd party applications to mimic the behavior of a Speakerphone button and therefore may change the input route as well as output route. For example, setting the AVAudioSessionPortOverride to AVAudioSessionPortOverrideSpeaker while a headset is plugged in will cause the route to change to built-in mic / built-in speaker.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/CallAudioService.swift
+++ b/Signal/src/call/CallAudioService.swift
@@ -145,7 +145,7 @@ struct AudioSource: Hashable {
         ensureProperAudioSession(call: call)
 
         // It's importent to set preferred input *after* ensuring properAudioSession
-        // because some sources are only valid for certain categories.
+        // because some sources are only valid for certain category/option combinations.
         let session = AVAudioSession.sharedInstance()
         do {
             try session.setPreferredInput(audioSource?.portDescription)
@@ -161,6 +161,8 @@ struct AudioSource: Hashable {
     }
 
     private func ensureProperAudioSession(call: SignalCall?) {
+        AssertIsOnMainThread()
+
         guard let call = call else {
             setAudioSession(category: AVAudioSessionCategoryPlayback,
                             mode: AVAudioSessionModeDefault)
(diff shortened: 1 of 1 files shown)


Signal/src/call/CallAudioService.swift
```diff
@@ -178,16 +180,22 @@ struct AudioSource: Hashable {
...
         } else {
             setAudioSession(category: AVAudioSessionCategoryPlayAndRecord,
                             mode: AVAudioSessionModeVoiceChat,
                             options: [.allowBluetooth])
         }
+
+        let session = AVAudioSession.sharedInstance()
+        do {
+            if call.isSpeakerphoneEnabled {
+                try session.overrideOutputAudioPort(.speaker)
```
charlesmchen: Is this cheap?  `setAudioSession` ignored redundant changes - can/should we do something like that here?

Signal/src/call/CallAudioService.swift
```diff
@@ -178,16 +180,22 @@ struct AudioSource: Hashable {
...
+            } else {
+                try session.overrideOutputAudioPort(.none)
+            }
+        } catch {
+            owsFail("\(TAG) failed overrideing output audio. isSpeakerPhoneEnabled: \(call.isSpeakerphoneEnabled)")
```
charlesmchen: s/overrideing/overriding/.

michaelkirk:
in 914d2936

Issue #2284, opened 2017-07-03
Disappearing message timer automatically turned off


TreatyOfRome:
- [X ] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description
The disappearing message timer setting is automatically turned off.

### Steps to reproduce
- I'm (iOS) having a chat with my chatpartner (Android);
- I have set the dissapearing message timer to 1 week;
- System message received that I have set the dissapearing message timer to 1 week;
- We exchange some messages;
- After a while, I receive a system message that my chatpartner has disabled the disappearing message timer;
- My chatpartner receives a system message that I have disabled the dissapearing message timer;
- We both did not disabled the disappearing message timer.

**Actual result:** Disappearing message status changes automatically.
**Expected result:** Disappearing message status remains unchanged.

### Device info
**Device**: iPhone 6S
**iOS version**: 10.3.2
**Signal version:** 2.13.0.16

### Link to debug log
https://gist.github.com/1a6aaaca50aa6b44f9798b12a6b88b08
Debug log is produced 1-5 minutes after automatical disabling of disappearing message timer.

TreatyOfRome:
Additional info: 
- This bug appeared since my update to Signal 2.13.0;
- I have verified safety numbers with my chatpartner (camera+scan);
- Deleting the chat and starting an new chat with disappearing message timer does not solve this bug.

michaelkirk:
Did your friend reinstall? Are either of you using a linked desktop?

TreatyOfRome:
Neither party reinstalled. No linked desktops involved.
Friend has Signal 4.6.1 (Android).

michaelkirk:
Thanks for the details.

From your log, here is where it happened:

https://gist.github.com/anonymous/1a6aaaca50aa6b44f9798b12a6b88b08#file-org-whispersystems-signal-2017-07-03-10-19-log-L800

    2017/07/03 15:57:52:149  [OWSDisappearingMessagesJob] Received remote message which had no expiration set, disabling our expiration to become consistent.

Any idea which message in your conversation history that corresponds to (15:57:52 local time)? Can you provide any details? Any way to get a debug log from your friend at this time?

michaelkirk:
Also, could you ask your friend to update to the latest Signal-Android?

TreatyOfRome:
Friends debug log did not have any data from that time (15:57) anymore. 
Friend has updated Signal-Android to 4.7.4. 
[...]

PR #2328, opened 2017-07-14
maintain disappearing state when replying from lock screen

michaelkirk:
fixes https://github.com/WhisperSystems/Signal-iOS/issues/2284#issuecomment-315256803

PTAL @charlesmchen

### DIFF
--- a/Signal/src/network/PushManager.m
+++ b/Signal/src/network/PushManager.m
@@ -8,6 +8,7 @@
 #import "OWSContactsManager.h"
 #import "PropertyListPreferences.h"
 #import "Signal-Swift.h"
+#import "ThreadUtil.h"
 #import <SignalServiceKit/NSDate+millisecondTimeStamp.h>
 #import <SignalServiceKit/OWSMessageSender.h>
 #import <SignalServiceKit/OWSSignalService.h>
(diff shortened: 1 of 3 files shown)


Signal/src/network/PushManager.m
```diff
@@ -147,16 +148,16 @@ - (void)application:(UIApplication *)application
...
-            TSOutgoingMessage *message =
-                [[TSOutgoingMessage alloc] initWithTimestamp:[NSDate ows_millisecondTimeStamp]
-                                                    inThread:thread
-                                                 messageBody:responseInfo[UIUserNotificationActionResponseTypedTextKey]];
-            [self.messageSender sendMessage:message
+            NSString *replyText = responseInfo[UIUserNotificationActionResponseTypedTextKey];
+
+            [ThreadUtil sendMessageWithText:replyText
+                inThread:thread
+                messageSender:self.messageSender
```
charlesmchen: There's a lot of places we call `sendMessage:`, e.g. in many places we send "group update" and "leave group" messages.  Should these be moved to using some `ThreadUtil` method that preserves this policy or does that not matter in these cases?

michaelkirk: Leaving/adding people to a group does not affect disappearing time on Desktop/iOS, furthermore we intentionally don't automatically delete those status messages, so this seems sufficient, if not a bit tenuous.

michaelkirk:
In 847a0ff8

PR #2330, opened 2017-07-14
Charlesmchen/empty home view

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/SignalsViewController.m
+++ b/Signal/src/ViewControllers/SignalsViewController.m
@@ -125,6 +125,10 @@
                                              selector:@selector(applicationDidEnterBackground:)
                                                  name:UIApplicationDidEnterBackgroundNotification
                                                object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(yapDatabaseModified:)
+                                                 name:YapDatabaseModifiedNotification
+                                               object:nil];
 }
 
 - (void)dealloc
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/SignalsViewController.m
```diff
@@ -384,6 +383,8 @@ - (void)viewWillAppear:(BOOL)animated {
                                           animated:NO];
         }
     }
+
+    [self checkIfEmptyView];
```
charlesmchen: Let's do this last.

Signal/src/ViewControllers/SignalsViewController.m
```diff
@@ -846,6 +849,7 @@ - (void)yapDatabaseModified:(NSNotification *)notification {
     [self updateInboxCountLabel];
 
     if ([sectionChanges count] == 0 && [rowChanges count] == 0) {
+        [self checkIfEmptyView];
```
charlesmchen: This could explain it.

Signal/src/ViewControllers/SignalsViewController.m
```diff
@@ -902,6 +906,13 @@ - (void)yapDatabaseModified:(NSNotification *)notification {
     [self checkIfEmptyView];
 }
 
+- (void)setViewingThreadsIn:(CellState)viewingThreadsIn
+{
+    _viewingThreadsIn = viewingThreadsIn;
+
+    [self checkIfEmptyView];
```
charlesmchen: `viewingThreadsIn` affects `checkIfEmptyView`; we should call `checkIfEmptyView` every time `viewingThreadsIn` changes - especially because the default value is `kArchiveState` not `kInboxState`.

Signal/src/ViewControllers/SignalsViewController.m
```diff
@@ -414,7 +421,24 @@ - (void)updateShouldObserveDBModifications
 
 - (void)setShouldObserveDBModifications:(BOOL)shouldObserveDBModifications
 {
-    if (!_shouldObserveDBModifications && shouldObserveDBModifications && self.threadMappings != nil) {
```
michaelkirk: nit: Can we rename this to reflect it's new behavior? Maybe: `shouldHandleDBModifications`

charlesmchen: Done.

PR #2332, opened 2017-07-14
audio route cleanup

michaelkirk:
- Nicer labels for devices in route picker
- Fix issue with audio not being routed to receiver if picked too early

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/CallAudioService.swift
+++ b/Signal/src/call/CallAudioService.swift
@@ -151,15 +151,6 @@ struct AudioSource: Hashable {
         AssertIsOnMainThread()
 
         ensureProperAudioSession(call: call)
-
-        // It's importent to set preferred input *after* ensuring properAudioSession
-        // because some sources are only valid for certain category/option combinations.
-        let session = AVAudioSession.sharedInstance()
-        do {
-            try session.setPreferredInput(audioSource?.portDescription)
-        } catch {
-            owsFail("\(TAG) setPreferredInput in \(#function) failed with error: \(error)")
-        }
     }
 
     internal func hasLocalVideoDidChange(call: SignalCall, hasLocalVideo: Bool) {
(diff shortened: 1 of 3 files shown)


Signal/src/call/CallAudioService.swift
```diff
@@ -169,32 +176,52 @@ struct AudioSource: Hashable {
             return
         }
 
+        // Disallow bluetooth when user has explicitly chosen the built in receiver.
+        // I'm actually not sure why this is required - it seems like we should just be able
+        // to setPreferredInput to call.audioSource.portDescription in this case,
+        // but in practice I'm seeing the call revert to the bluetooth headset.
+        // Presumably something else (in WebRTC?) is touching our shared AudioSession.
+        let options: AVAudioSessionCategoryOptions = call.audioSource?.isBuiltInEarPiece == true ? [] : [.allowBluetooth]
```
michaelkirk: This is pretty ham fisted, but it works. If you have a bluetooth in your ear, but for some reason, have opted to not use it, you may get a little blip of ringtone as the call connects.

Signal/src/call/CallAudioService.swift
```diff
@@ -12,27 +12,43 @@ struct AudioSource: Hashable {
     let image: UIImage
     let localizedName: String
     let portDescription: AVAudioSessionPortDescription?
+
+    // The built in loud speaker / aka speakerphone
     let isBuiltInSpeaker: Bool
 
-    init(localizedName: String, image: UIImage, isBuiltInSpeaker: Bool, portDescription: AVAudioSessionPortDescription? = nil) {
+    // The build in quiet speaker, aka the normal phone handset receiver earpiece
```
charlesmchen: s/build in/built-in/

Signal/src/call/CallAudioService.swift
charlesmchen: ` == true` is redundant and can be omitted?

charlesmchen: Will this interfere with the user connecting their bluetooth _after_ the call is connected?

Signal/src/call/CallAudioService.swift
```diff
@@ -169,32 +176,52 @@ struct AudioSource: Hashable {
...
+            // does not include my linked bluetooth device
             setAudioSession(category: AVAudioSessionCategoryPlayAndRecord,
                             mode: AVAudioSessionModeVoiceChat,
-                            options: [.allowBluetooth])
+                            options: options)
         }
 
         let session = AVAudioSession.sharedInstance()
         do {
+            // It's importent to set preferred input *after* ensuring properAudioSession
```
charlesmchen: s/importent/important/

Signal/src/call/CallAudioService.swift
michaelkirk: `== true` is redundant and can be omitted?

Not in Swift - it's comparing to an optional boolean, so `== true` means it's not false **and** it's not nil.

> Will this interfere with the user connecting their bluetooth after the call is connected?

Should be fine. This only disallows bluetooth when the user has explicitly chosen "iPhone" output. If, after that, they explicitly choose a bluetooth source, then this restriction goes away. I'll try to clarify the comment.

PR #2336, opened 2017-07-18
Adding to existing contact requires contact access

michaelkirk:
Previously we'd show an empty "list" of existing contacts.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ContactsViewHelper.m
+++ b/Signal/src/ViewControllers/ContactsViewHelper.m
@@ -297,7 +297,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - Editing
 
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ContactsViewHelper.m
```diff
@@ -297,6 +297,35 @@ - (BOOL)doesContact:(Contact *)contact matchSearchTerms:(NSArray<NSString *> *)s
 
 #pragma mark - Editing
 
+- (UIAlertController *)missingContactAccessAlertController
```
charlesmchen: Why not `showMissingContactAccessAlertController:(UIViewController *)fromViewController` and present this vc as well?

michaelkirk: Done!

PR #2338, opened 2017-07-18
Remove verification debug UI

michaelkirk:
There is a more true-to-life way to set verification state - mark
someone as verified or default via the SN UI.

Once they are marked as verified, you can mark them as no-longer-verified
by toggling a keychange in the debug UI

PTAL @charlesmchen


michaelkirk:
> LGTM - but I don't understand why these actions left users in a bad state. They twiddled the state in the the identity manager, which should be the single source of truth. What was missing?

See this line:
https://github.com/WhisperSystems/SignalServiceKit/blob/master/src/Messages/OWSIdentityManager.m#L376

PR #2340, opened 2017-07-19
Always reload home view table when changing grouping.

charlesmchen:
Might be redundant in some cases but `reloadData` should be cheap. Might help with the "empty home view" issue and also future-proofs us when refactoring home view in the future.

PTAL @michaelkirk


charlesmchen:
PTAL @michaelkirk

michaelkirk:
Will this induce an extra flicker when switching groupings?

charlesmchen:
> Will this induce an extra flicker when switching groupings?

It shouldn't.  In that case we have to `reloadData`.  Best case, this case ensures we do in edge cases were we weren't before. Worst case, it causes us to call it twice - but in immediate succession so its perf/ux impact should be negligible.

michaelkirk:
LGTM

PR #2351, opened 2017-07-24
Address yet another edge cases around pasteboards that contain both t…

charlesmchen:
…extual and non-textual content, e.g. a gif and the URL of that gif.

PTAL @michaelkirk


Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -341,6 +337,9 @@ class SignalAttachment: NSObject {
                 hasNonTextUTIType = true
             }
         }
+        if pasteboardUTISet.contains(kUTTypeURL as String) {
```
michaelkirk: maybe a comment about why this ordering is critical so we don't lose this knowledge in a "harmless" refactor.

PR #2352, opened 2017-07-24
atomic registration

michaelkirk:
Three related changes here - each with their own commit.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ProfileFetcherJob.swift
+++ b/Signal/src/ProfileFetcherJob.swift
@@ -3,6 +3,7 @@
 //
 
 import Foundation
+import PromiseKit
 
 @objc
 class ProfileFetcherJob: NSObject {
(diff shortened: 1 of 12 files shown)


Signal/src/ProfileFetcherJob.swift
```diff
@@ -12,34 +13,54 @@ class ProfileFetcherJob: NSObject {
...
     // This property is only accessed on the main queue.
     static var fetchDateMap = [String: Date]()
 
     public class func run(thread: TSThread, networkManager: TSNetworkManager) {
-        ProfileFetcherJob(thread: thread, networkManager: networkManager).run()
+        ProfileFetcherJob(networkManager: networkManager).run(thread: thread)
     }
 
-    init(thread: TSThread, networkManager: TSNetworkManager) {
+    init(networkManager: TSNetworkManager) {
```
michaelkirk: Reused the ProfileFetcherJob, but now it doesn't *always* take a thread.

SignalServiceKit/src/Messages/PreKeyBundle+jsonDict.m
```diff
@@ -5,17 +5,27 @@
...
++ (nullable PreKeyBundle *)preKeyBundleFromDictionary:(NSDictionary *)dictionary forDeviceNumber:(NSNumber *)number
+{
     PreKeyBundle *bundle        = nil;
-    NSString *identityKeyString = [dictionary objectForKey:@"identityKey"];
-    NSArray *devicesArray       = [dictionary objectForKey:@"devices"];
 
-    if (!(identityKeyString && [devicesArray isKindOfClass:[NSArray class]])) {
-        DDLogError(@"Failed to get identity key or messages array from server request");
+    id identityKeyObject = [dictionary objectForKey:@"identityKey"];
+    if (![identityKeyObject isKindOfClass:[NSString class]]) {
```
michaelkirk: prevent crash when messaging someone with an incomplete profile

SignalServiceKit/src/Account/TSAccountManager.m
```diff
@@ -234,6 +234,11 @@ - (void)verifyAccountWithCode:(NSString *)verificationCode
                                                                                forNumber:phoneNumber
                                                                             signalingKey:signalingKey
                                                                                  authKey:authToken];
+    void (^completedRegistrationBlock)() = ^{
```
michaelkirk: Don't consider registration complete until prekeys are uploaded.

Signal/src/environment/OWS106EnsureProfileComplete.swift
```diff
@@ -0,0 +1,110 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+import PromiseKit
+
+class OWS106EnsureProfileComplete: OWSDatabaseMigration {
```
michaelkirk: Fixup any half-registered users in the wild.

Signal/src/environment/OWS106EnsureProfileComplete.swift
```diff
@@ -0,0 +1,110 @@
...
+
+                var isCompleted = false
+                strongSelf.ensureProfileComplete().then { _ -> Void in
+                    guard isCompleted == false else {
+                        Logger.info("Already saved. Skipping redundant call.")
+                        return
+                    }
+                    Logger.info("\(strongSelf.TAG) complete. Canceling timer and saving.")
+                    isCompleted = true
+                    timer.invalidate()
```
charlesmchen: Will this work?  I don't think the local `timer` variable is initialized at the time its captured, so I'd expect this not to work.  Swift might be different in this regard that Obj-C.

Signal/src/environment/OWS106EnsureProfileComplete.swift
```diff
@@ -0,0 +1,110 @@
...
+                    strongSelf.completionHandler()
+                }.catch { error in
+                    Logger.error("\(strongSelf.TAG) failed with \(error). We'll try again in \(CompleteRegistrationFixerJob.kRetryInterval) seconds.")
+                }.retainUntilComplete()
+            }
+            self.timer = timer
+
+            timer.fire()
```
charlesmchen: Why is this necessary?

Signal/src/environment/OWS106EnsureProfileComplete.swift
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    // Overriding runUp since we have some specific completion criteria which
+    // is more likely to fail since it involves network requests.
+    override func runUp() {
+        CompleteRegistrationFixerJob(completionHandler: {
+            Logger.info("\(self.TAG) Completed. Saving.")
+            self.save()
+        }).start()
+    }
```
charlesmchen: Won't CompleteRegistrationFixerJob get deallocated as soon as it starts?  Who maintains a strong reference to it while it "runs"?

SignalServiceKit/src/Messages/PreKeyBundle+jsonDict.m
charlesmchen: Are we making the same fix in android and desktop?

Signal/src/environment/OWS106EnsureProfileComplete.swift
michaelkirk: Not strictly necessary, just makes the initial run *now* instead of in **n** seconds.

SignalServiceKit/src/Messages/PreKeyBundle+jsonDict.m
michaelkirk: In my testing Android and Desktop didn't crash in this scenario, so presumably they already have more robust parsing logic than we did.

Signal/src/environment/OWS106EnsureProfileComplete.swift
michaelkirk: sorry this was convoluted. This timer is not `self.timer` it's the block param of the same name. I'm going to clean this up a bit to be a little clearer.

charlesmchen: Even still, there may be an issue here.

charlesmchen: LGTM with these changes.

michaelkirk:
in bf82d2e2e03ecd1747bbee60592c98fd9f59bc68

PR #2353, opened 2017-07-24
Charlesmchen/analytics5

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Storage/TSStorageManager.m
+++ b/SignalServiceKit/src/Storage/TSStorageManager.m
@@ -121,14 +121,16 @@ static NSString *keychainDBPassAccount    = @"TSDatabasePass";
         // The best we can try to do is to discard the current database
         // and behave like a clean install.
 
-        OWSProdError(@"storage_error_could_not_load_database");
+        // NOTE: This analytics event will never be delivered, since the database isn't working.
+        OWSProdCritical(@"storage_error_could_not_load_database");
 
         // Try to reset app by deleting database.
         // Disabled resetting storage until we have better data on why this happens.
         // [self resetSignalStorage];
 
         if (![self tryToLoadDatabase]) {
-            OWSProdError(@"storage_error_could_not_load_database_second_attempt");
+            // NOTE: This analytics event will never be delivered, since the database isn't working.
+            OWSProdCritical(@"storage_error_could_not_load_database_second_attempt");
 
             [NSException raise:TSStorageManagerExceptionNameNoDatabase format:@"Failed to initialize database."];
         }
@@ -187,9 +189,9 @@ static NSString *keychainDBPassAccount    = @"TSDatabasePass";
             OWSProdErrorWParams(@"storage_error_deserialization", ^{
                 return (@{
                     @"collection" : collection,
-                    kOWSProdAssertParameterNSExceptionName : exception.name,
-                    kOWSProdAssertParameterNSExceptionReason : exception.reason,
-                    kOWSProdAssertParameterNSExceptionClassName : NSStringFromClass([exception class]),
+                    kOWSAnalyticsParameterNSExceptionName : exception.name,
+                    kOWSAnalyticsParameterNSExceptionReason : exception.reason,
+                    kOWSAnalyticsParameterNSExceptionClassName : NSStringFromClass([exception class]),
                 });
             });
             @throw exception;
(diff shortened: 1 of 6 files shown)


SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -121,14 +121,16 @@ - (instancetype)initDefault
         // The best we can try to do is to discard the current database
         // and behave like a clean install.
 
-        OWSProdError(@"storage_error_could_not_load_database");
+        // NOTE: This analytics event will never be delivered, since the database isn't working.
+        OWSProdCritical(@"storage_error_could_not_load_database");
```
michaelkirk: I guess it's a TODO. Hopefully we could fire it off immediately without needing to persist anything.

SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -187,9 +189,9 @@ + (YapDatabaseDeserializer)logOnFailureDeserializer
             OWSProdErrorWParams(@"storage_error_deserialization", ^{
                 return (@{
```
michaelkirk: It seems like we're trying to add one extra parameter to the standard AnalyticsParametersFromNSException. You don't have to do it now, but it would be nice to be able to specify "additionalParams" which would mix with the default params.

SignalServiceKit/src/Util/OWSAnalytics.m
```diff
@@ -157,6 +157,12 @@ - (void)tryToSyncEvents
 
         DDLogDebug(@"%@ trying to deliver event: %@", self.tag, firstEventKey);
         self.hasRequestInFlight = YES;
+
+        __block UIBackgroundTaskIdentifier task;
+        task = [UIApplication.sharedApplication beginBackgroundTaskWithExpirationHandler:^{
+            [UIApplication.sharedApplication endBackgroundTask:task];
```
michaelkirk: do we need to reset `hasRequestInFlight = NO` in our expiration handler? Any other state?

SignalServiceKit/src/Util/OWSAnalytics.m
```diff
@@ -157,6 +157,12 @@ - (void)tryToSyncEvents
 
         DDLogDebug(@"%@ trying to deliver event: %@", self.tag, firstEventKey);
         self.hasRequestInFlight = YES;
+
+        __block UIBackgroundTaskIdentifier task;
```
michaelkirk: why is this `__block`?

SignalServiceKit/src/Util/OWSAnalytics.m
```diff
@@ -277,18 +291,11 @@ - (void)logEvent:(NSString *)eventName
     DDLogFlag logFlag;
     BOOL async = YES;
     switch (severity) {
-        case OWSAnalyticsSeverityDebug:
```
michaelkirk: 👍

SignalServiceKit/src/Storage/TSStorageManager.m
charlesmchen: Yeah, I've been thinking about it, but it would probably require a whole new set of database-specific macros.  I'm wondering if we'll use the "critical-severity" macros anywhere else.

SignalServiceKit/src/Storage/TSStorageManager.m
charlesmchen: Yup, it's the old tension between making the "reusable" flexible "enough" - able to handle the common use cases without making it crazy complicated to handle the edge cases.  Right now, there's only one place we need an extra param.

SignalServiceKit/src/Util/OWSAnalytics.m
charlesmchen: This goes back to what we were discussing this morning.  Objective-C captures by value by default, Swift captures by reference.  `task` isn't set yet when the block is defined, so we can't capture it's value.  `__block` has two side effects: it causes this variable to be captured by reference and allocates it on the heap instead of the stack so that it outlives the scope in which its defined.

screen shot 2017-07-24 at 5 15 24 pm

SignalServiceKit/src/Util/OWSAnalytics.m
charlesmchen: Good catch!  Just `hasRequestInFlight`.

PR #2354, opened 2017-07-24
Instrument CallService.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -81,6 +81,18 @@ enum CallError: Error {
 // Should be roughly synced with Android client for consistency
 fileprivate let connectingTimeoutSeconds = 120
 
+// Example: OWSProdError("blah", #file, #function, #line)
+func OWSProdError(_ __eventName: String, file __file: String, function __function: String, line __line: Int32) {
+    let location = "\((__file as NSString).lastPathComponent):\(__function)"
+    OWSAnalytics
+        .logEvent(__eventName, severity: .error, parameters: nil, location: (location as NSString).utf8String!, line:__line)
+}
+
+func OWSProdCallAssertionError(description: String) -> CallError {
+    OWSProdError(description, file:#file, function:#function, line:#line)
+    return .assertionError(description: description)
+}
+
 // All Observer methods will be invoked from the main thread.
 protocol CallServiceObserver: class {
     /**
@@ -273,7 +285,7 @@ protocol CallServiceObserver: class {
             let errorDescription = "\(TAG) call was unexpectedly already set."
             Logger.error(errorDescription)
             call.state = .localFailure
-            return Promise(error: CallError.assertionError(description: errorDescription))
+            return Promise(error: OWSProdCallAssertionError(description: errorDescription))
         }
 
         self.call = call
@@ -295,7 +307,7 @@ protocol CallServiceObserver: class {
             guard self.peerConnectionClient == nil else {
                 let errorDescription = "\(self.TAG) peerconnection was unexpectedly already set."
                 Logger.error(errorDescription)
-                throw CallError.assertionError(description: errorDescription)
+                throw OWSProdCallAssertionError(description: errorDescription)
             }
 
             let useTurnOnly = Environment.getCurrent().preferences.doCallsHideIPAddress()
@@ -371,7 +383,7 @@ protocol CallServiceObserver: class {
         }
 
         guard let fulfillReadyToSendIceUpdatesPromise = self.fulfillReadyToSendIceUpdatesPromise else {
-            self.handleFailedCall(failedCall: call, error: .assertionError(description: "failed to create fulfillReadyToSendIceUpdatesPromise"))
+            self.handleFailedCall(failedCall: call, error: OWSProdCallAssertionError(description: "failed to create fulfillReadyToSendIceUpdatesPromise"))
             return
         }
 
@@ -409,7 +421,7 @@ protocol CallServiceObserver: class {
         }
 
         guard let peerConnectionClient = self.peerConnectionClient else {
-            handleFailedCall(failedCall: call, error: CallError.assertionError(description: "peerConnectionClient was unexpectedly nil in \(#function)"))
+            handleFailedCall(failedCall: call, error: OWSProdCallAssertionError(description: "peerConnectionClient was unexpectedly nil in \(#function)"))
             return
         }
 
@@ -697,7 +709,7 @@ protocol CallServiceObserver: class {
         AssertIsOnMainThread()
 
         guard let call = self.call else {
-            self.handleFailedCurrentCall(error: .assertionError(description: "ignoring local ice candidate, since there is no current call."))
+            self.handleFailedCurrentCall(error: OWSProdCallAssertionError(description: "ignoring local ice candidate, since there is no current call."))
             return
         }
 
@@ -712,7 +724,7 @@ protocol CallServiceObserver: class {
             guard call.state != .idle else {
                 // This will only be called for the current peerConnectionClient, so
                 // fail the current call.
-                self.handleFailedCurrentCall(error: .assertionError(description: "ignoring local ice candidate, since call is now idle."))
+                self.handleFailedCurrentCall(error: OWSProdCallAssertionError(description: "ignoring local ice candidate, since call is now idle."))
                 return
             }
 
@@ -748,7 +760,7 @@ protocol CallServiceObserver: class {
         guard let call = self.call else {
             // This will only be called for the current peerConnectionClient, so
             // fail the current call.
-            handleFailedCurrentCall(error: .assertionError(description:"\(TAG) ignoring \(#function) since there is no current call."))
+            handleFailedCurrentCall(error: OWSProdCallAssertionError(description:"\(TAG) ignoring \(#function) since there is no current call."))
             return
         }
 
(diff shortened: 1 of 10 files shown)


Signal/src/call/CallService.swift
```diff
@@ -81,6 +81,18 @@ enum CallError: Error {
 // Should be roughly synced with Android client for consistency
 fileprivate let connectingTimeoutSeconds = 120
 
+// Example: OWSProdError("blah", #file, #function, #line)
+func OWSProdError(_ __eventName: String, file __file: String, function __function: String, line __line: Int32) {
+    let location = "\((__file as NSString).lastPathComponent):\(__function)"
+    OWSAnalytics
+        .logEvent(__eventName, severity: .error, parameters: nil, location: (location as NSString).utf8String!, line:__line)
+}
```
charlesmchen: So far, I've only instrumented a single .swift file.  If we instrument anywhere else, I'll pull out a new OWSAnalytics.swift file.  

Is there any problem (at link time?) with having objective-c and swift modules with the same name, e.g. Alice.m and Alice.swift?

Signal/src/call/CallService.swift
```diff
@@ -81,6 +81,18 @@ enum CallError: Error {
 // Should be roughly synced with Android client for consistency
 fileprivate let connectingTimeoutSeconds = 120
 
+// Example: OWSProdError("blah", #file, #function, #line)
+func OWSProdError(_ __eventName: String, file __file: String, function __function: String, line __line: Int32) {
+    let location = "\((__file as NSString).lastPathComponent):\(__function)"
```
charlesmchen: The poor man's `__PRETTY_FUNCTION__`.

Signal/src/call/CallService.swift
```diff
@@ -81,6 +81,18 @@ enum CallError: Error {
...
+    OWSAnalytics
+        .logEvent(__eventName, severity: .error, parameters: nil, location: (location as NSString).utf8String!, line:__line)
+}
+
+func OWSProdCallAssertionError(description: String) -> CallError {
+    OWSProdError(description, file:#file, function:#function, line:#line)
```
charlesmchen: Crap this is wrong.

charlesmchen: Okay, I've pushed another commit that fixes this. =(

Signal/src/call/CallService.swift
```diff
@@ -81,6 +81,18 @@ enum CallError: Error {
 // Should be roughly synced with Android client for consistency
 fileprivate let connectingTimeoutSeconds = 120
 
+// Example: OWSProdError("blah", #file, #function, #line)
+func OWSProdError(_ __eventName: String, file: String, function: String, line: Int32) {
+    let location = "\((file as NSString).lastPathComponent):\(function)"
```
charlesmchen: This is the poor man's `__PRETTY_FUNCTION_`.

SignalServiceKit/src/Util/OWSAnalytics.h
```diff
@@ -83,12 +83,12 @@ typedef NSDictionary<NSString *, id> *_Nonnull (^OWSProdAssertParametersBlock)()
 // parametersBlock is of type OWSProdAssertParametersBlock.
 // The "C" variants (e.g. OWSProdAssert() vs. OWSProdCAssert() should be used in free functions,
 // where there is no self. They can also be used in blocks to avoid capturing a reference to self.
-#define OWSProdAssertWParamsTemplate(__value, __analyticsEventName, __parametersBlock, __assertMacro)                  \
```
charlesmchen: Simple rename: `__analyticsEventName` -> `__eventName`

SignalServiceKit/src/Util/OWSAnalytics.m
```diff
@@ -145,46 +162,103 @@ - (void)tryToSyncEvents
...
+}
 
-        DDLogDebug(@"%@ trying to deliver event: %@", self.tag, firstEventKey);
+- (void)sendEvent:(NSDictionary *)eventDictionary eventKey:(NSString *)eventKey isCritical:(BOOL)isCritical
+{
+    OWSAssert(eventDictionary);
+    OWSAssert(eventKey);
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+    OWSAssert(dispatch_get_current_queue() == self.serialQueue);
```
michaelkirk: We achiever a similar end through different means in PeerConnectionClient and SessionCipher. It would be nice if we had just one approach to this. I understand the drawbacks in the approach used elesewhere to be:

1. only runs on iOS10+
2. Since check actually crashes the app, we only run it in debug, meaning no sign of assertion failure in Production.

I'm less clear on the drawbacks to this approach based on the reasons that this method was deprecated - do we always get the correct answer here?

Anyway, it'd be nice to centralize on the "one true way" and use it consistently.

SignalServiceKit/src/Util/OWSAnalytics.m
```diff
@@ -145,46 +162,103 @@ - (void)tryToSyncEvents
...
+#pragma clang diagnostic pop
+
+    if (isCritical) {
+        [self submitEvent:eventDictionary
+            eventKey:eventKey
+            success:^{
+                DDLogDebug(@"%@ sendEvent[critical] succeeded: %@", self.tag, eventKey);
```
michaelkirk: Though I think in all the current cases we'll be crashing shortly after this point, but that's not necessarily intuitive. Should we reset hasRequestInFlight here as well in case there's a situation where multiple critical events are being fired?

SignalServiceKit/src/Util/OWSAnalytics.m
```diff
@@ -145,46 +162,103 @@ - (void)tryToSyncEvents
...
+- (void)sendEvent:(NSDictionary *)eventDictionary eventKey:(NSString *)eventKey isCritical:(BOOL)isCritical
+{
+    OWSAssert(eventDictionary);
+    OWSAssert(eventKey);
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+    OWSAssert(dispatch_get_current_queue() == self.serialQueue);
+#pragma clang diagnostic pop
+
+    if (isCritical) {
```
michaelkirk: take it or leave it: based on our earlier discussion - I agree that "critical" is a little bit opaque. It might make sense to label it as `shouldSubmitImmediately = true` or `shouldPersist = false` or something else more to the point.

SignalServiceKit/src/Util/OWSAnalytics.m
```diff
@@ -145,46 +162,103 @@ - (void)tryToSyncEvents
...
-            self.hasRequestInFlight = NO;
+- (void)submitEvent:(NSDictionary *)eventDictionary
+           eventKey:(NSString *)eventKey
+            success:(void (^_Nonnull)())successBlock
+            failure:(void (^_Nonnull)())failureBlock
+{
+    OWSAssert(eventDictionary);
+    OWSAssert(eventKey);
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
michaelkirk: take it or leave it: Should we pull this into a new `OWSAssertCurrentQueue(self.seriaQueue)` ?

SignalServiceKit/src/Util/OWSAnalytics.m
```diff
@@ -145,46 +162,103 @@ - (void)tryToSyncEvents
...
-            [UIApplication.sharedApplication endBackgroundTask:task];
-        }];
+    DDLogDebug(@"%@ submitting: %@", self.tag, eventKey);
 
-        // Until we integrate with an analytics platform, behave as though all event delivery succeeds.
-        dispatch_async(dispatch_get_main_queue(), ^{
-            self.hasRequestInFlight = NO;
+    __block UIBackgroundTaskIdentifier task;
+    task = [UIApplication.sharedApplication beginBackgroundTaskWithExpirationHandler:^{
+        failureBlock();
```
michaelkirk: Does this run the risk of calling multiple completion handlers? e.g. either failure twice or failure once and success once?

Should we inspect/set a `hasCompleted` boolean before calling any of the failure/success blocks?

Signal/src/call/CallService.swift
```diff
@@ -81,6 +81,18 @@ enum CallError: Error {
 // Should be roughly synced with Android client for consistency
 fileprivate let connectingTimeoutSeconds = 120
 
+// Example: OWSProdError("blah", #file, #function, #line)
+func OWSProdError(_ __eventName: String, file: String, function: String, line: Int32) {
```
charlesmchen: Yeah, you're right.  I was trying to be consistent with the preprocessor macros, but it just doesn't make sense.

SignalServiceKit/src/Util/OWSAnalytics.m
charlesmchen: Good point.  I've created a `AssertOnDispatchQueue()` macro for obj-c and a `assertOnQueue()` free function for swift.  I've also reorganized our util code a tiny bit.

SignalServiceKit/src/Util/OWSAnalytics.m
charlesmchen: Yup, still thinking about this.

SignalServiceKit/src/Util/OWSAnalytics.m
charlesmchen: `hasRequestInFlight` only applies to "non-critical" events.  "Critical" events ignore this flag - don't honor it, set it or clear it.  So I think the current arrangement is right.

SignalServiceKit/src/Util/OWSAnalytics.m
charlesmchen: Yup, done.

SignalServiceKit/src/Util/OWSAnalytics.m
charlesmchen: Good idea, done.

PR #2355, opened 2017-07-25
Rework database observation in home and message views.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -219,6 +219,10 @@ typedef enum : NSUInteger {
 
 @property (nonatomic) UIView *scrollDownButton;
 
+@property (nonatomic) BOOL isViewVisible;
+@property (nonatomic) BOOL isAppInBackground;
+@property (nonatomic) BOOL shouldObserveDBModifications;
+
 @end
 
 #pragma mark -
@@ -294,6 +298,30 @@ typedef enum : NSUInteger {
                                              selector:@selector(identityStateDidChange:)
                                                  name:kNSNotificationName_IdentityStateDidChange
                                                object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(didChangePreferredContentSize:)
+                                                 name:UIContentSizeCategoryDidChangeNotification
+                                               object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(yapDatabaseModified:)
+                                                 name:YapDatabaseModifiedNotification
+                                               object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(applicationWillEnterForeground:)
+                                                 name:UIApplicationWillEnterForegroundNotification
+                                               object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(applicationDidEnterBackground:)
+                                                 name:UIApplicationDidEnterBackgroundNotification
+                                               object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(applicationWillResignActive:)
+                                                 name:UIApplicationWillResignActiveNotification
+                                               object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(cancelReadTimer)
+                                                 name:UIApplicationDidEnterBackgroundNotification
+                                               object:nil];
 }
 
 - (void)blockedPhoneNumbersDidChange:(id)notification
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -1886,8 +1853,8 @@ - (TSOutgoingMessage *)nextOutgoingMessage:(NSIndexPath *)indexPath
 {
     NSInteger rowCount = [self.collectionView numberOfItemsInSection:indexPath.section];
     for (NSInteger row = indexPath.row + 1; row < rowCount; row++) {
-        id<OWSMessageData> nextMessage =
-            [self messageAtIndexPath:[NSIndexPath indexPathForRow:row inSection:indexPath.section]];
+        NSIndexPath *nextIndexPath = [NSIndexPath indexPathForRow:row inSection:indexPath.section];
+        TSInteraction *nextMessage = [self interactionAtIndexPath:nextIndexPath];
```
charlesmchen: This fixes an unrelated and embarrassing bug - we need to check the type of the _message (interaction)_, not the _message adapter_.

Signal/src/ViewControllers/SignalsViewController.m
```diff
@@ -861,20 +862,28 @@ - (YapDatabaseConnection *)uiDatabaseConnection {
...
 - (void)yapDatabaseModified:(NSNotification *)notification {
+    if (!self.shouldObserveDBModifications) {
+        return;
+    }
+
     NSArray *notifications  = [self.uiDatabaseConnection beginLongLivedReadTransaction];
-    NSArray *sectionChanges = nil;
-    NSArray *rowChanges     = nil;
+
+    if (![[self.uiDatabaseConnection ext:TSThreadDatabaseViewExtensionName] hasChangesForGroup:self.currentGrouping
```
charlesmchen: This could be a significant optimization.

charlesmchen:
Yay, this PR resolves the repro I found for the "last known layout glitch" in messages view.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -2267,41 +2234,10 @@ - (void)collectionView:(JSQMessagesCollectionView *)collectionView
...
-    // the mappings will be updated by yapDatabaseModified.
-    // This will leave the mapping range in a bad state.
-    // Therefore we temporarily disable observation of YapDatabaseModifiedNotification
-    // while updating the range and the dynamic interactions.
-    [[NSNotificationCenter defaultCenter] removeObserver:self name:YapDatabaseModifiedNotification object:nil];
-
-    // We need to update the dynamic interactions after loading earlier messages,
-    // since the unseen indicator may need to move or change.
-    [self ensureDynamicInteractions];
+    self.shouldObserveDBModifications = NO;
```
michaelkirk: These two new lines feel a little underhanded to me. A cursory read actually looks like a no-op. But looking deeper in the setter for shouldObserveDBModifications, there's a handful of lines to reset of all our mappings/layout/range. I think that's ultimately what you're trying to call here.

What if we extracted all the stuff at the bottom of `setShouldObserveDBModifications` to `reloadMessagesViewController` or whatever and just called that here.

Or is there something else going on here?

Signal/src/ViewControllers/SignalsViewController.m
michaelkirk: 👍 

Will this solve the problem of requesting contact access when receiving your first message while the app is in the background?

I wonder if there's an edge case where we won't properly request contact access if you are viewing your archive when you receive your first message.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
charlesmchen: It seems to work correctly for me with these changes.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -333,16 +361,14 @@ - (void)configureForThread:(TSThread *)thread
...
     _callOnOpen = callOnViewAppearing;
 
-    // We need to create the "unread indicator" before we mark
-    // all messages as read.
-    [self ensureDynamicInteractions];
-
     [self.uiDatabaseConnection beginLongLivedReadTransaction];
     self.messageMappings =
         [[YapDatabaseViewMappings alloc] initWithGroups:@[ thread.uniqueId ] view:TSMessageDatabaseViewExtensionName];
+    [self updateMessageMappingRangeOptions];
```
charlesmchen: Added a comment. This avoids a perf issue.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
charlesmchen: Good point.  I've pulled out `resetMappings` methods in both VCs to make the intent more clear.

Signal/src/ViewControllers/SignalsViewController.m
charlesmchen: Although `hasChangesForGroup:` filters the modifications by group, the most important filtering it does is to ignore all database changes that don't affect `TSThreadDatabaseViewExtensionName`.  E.g. writing a received message to the durable queue, saving a prekey change, etc.

> Will this solve the problem of requesting contact access when receiving your first message while the app is in the background?

It might.  We should test that.

> I wonder if there's an edge case where we won't properly request contact access if you are viewing your archive when you receive your first message.

...or if the app was in the background when you received your first message.  Sounds like we should add a check for that in the new `resetMappings` method.  I'll do so.

PR #2356, opened 2017-07-25
Charlesmchen/leaked view controllers

charlesmchen:
This fixes an absurdly large number of leaks in our UI.

// FREEBIE

### DIFF
--- a/Signal/src/ViewControllers/AttachmentApprovalViewController.swift
+++ b/Signal/src/ViewControllers/AttachmentApprovalViewController.swift
@@ -5,7 +5,7 @@
 import Foundation
 import MediaPlayer
 
-class AttachmentApprovalViewController: UIViewController, OWSAudioAttachmentPlayerDelegate {
+class AttachmentApprovalViewController: OWSViewController, OWSAudioAttachmentPlayerDelegate {
 
     let TAG = "[AttachmentApprovalViewController]"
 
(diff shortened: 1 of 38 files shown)


Signal/src/ViewControllers/AttachmentApprovalViewController.swift
```diff
@@ -5,7 +5,7 @@
 import Foundation
 import MediaPlayer
 
-class AttachmentApprovalViewController: UIViewController, OWSAudioAttachmentPlayerDelegate {
+class AttachmentApprovalViewController: OWSViewController, OWSAudioAttachmentPlayerDelegate {
```
charlesmchen: I added a `OWSViewController` base class that logs view deallocation.  It helps expose these leaks, which are easy to do.

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -206,7 +206,7 @@ - (void)updateTableContents
             [nonContactsSection
                 addItem:[OWSTableItem itemWithCustomCellBlock:^{
                     NewGroupViewController *strongSelf = weakSelf;
-                    OWSAssert(strongSelf);
+                    OWSCAssert(strongSelf);
```
charlesmchen: We need to use the "C" flavor of the Assert and Fail macros inside blocks since the non-C flavors capture self.

PR #2357, opened 2017-07-25
Store undecrypted envelopes before doing any processing

michaelkirk:
PTAL @charlesmchen 

// FREEBIE

### DIFF
--- a/SignalServiceKit/src/Messages/OWSMessageReceiver.m
+++ b/SignalServiceKit/src/Messages/OWSMessageReceiver.m
@@ -1,0 +1,338 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "OWSMessageReceiver.h"
+#import "OWSSignalServiceProtos.pb.h"
+#import "TSDatabaseView.h"
+#import "TSMessagesManager.h"
+#import "TSStorageManager.h"
+#import "TSYapDatabaseObject.h"
+#import <YapDatabase/YapDatabaseConnection.h>
+#import <YapDatabase/YapDatabaseTransaction.h>
+#import <YapDatabase/YapDatabaseView.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+#pragma mark - Persisted data model
+@class OWSSignalServiceProtosEnvelope;
+
+@interface OWSMessageProcessingJob : TSYapDatabaseObject
+
+@property (nonatomic, readonly) OWSSignalServiceProtosEnvelope *envelopeProto;
+@property (nonatomic, readonly) NSDate *createdAt;
+
+- (instancetype)initWithEnvelope:(OWSSignalServiceProtosEnvelope *)envelope NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithUniqueId:(NSString *)uniqueId NS_UNAVAILABLE;
+
+@end
+
+@interface OWSMessageProcessingJob ()
+
+@property (nonatomic, readonly) NSData *envelopeData;
+
+@end
+
+// TODO rename?
+@implementation OWSMessageProcessingJob
+
+- (instancetype)initWithEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
+{
+    self = [super initWithUniqueId:[NSUUID new].UUIDString];
+    if (!self) {
+        return self;
+    }
+
+    _envelopeData = envelope.data;
+    _createdAt = [NSDate new];
+
+    return self;
+}
+
+- (OWSSignalServiceProtosEnvelope *)envelopeProto
+{
+    return [OWSSignalServiceProtosEnvelope parseFromData:self.envelopeData];
+}
+
+@end
+
+#pragma mark - Finder
+
+NSString *const OWSMessageProcessingJobFinderExtensionName = @"OWSMessageProcessingJobFinderExtensionName";
+NSString *const OWSMessageProcessingJobFinderExtensionGroup = @"OWSMessageProcessingJobFinderExtensionGroup";
+
+@interface OWSMessageProcessingJobFinder : NSObject
+
+- (nullable OWSMessageProcessingJob *)nextJob;
+- (void)addJobForEnvelope:(OWSSignalServiceProtosEnvelope *)envelope;
+- (void)removeJobWithId:(NSString *)uniqueId;
+
+@end
+
+@interface OWSMessageProcessingJobFinder ()
+
+@property (nonatomic, readonly) YapDatabaseConnection *dbConnection;
+
+@end
+
+@implementation OWSMessageProcessingJobFinder
+
+- (instancetype)initWithDBConnection:(YapDatabaseConnection *)dbConnection
+{
+    OWSSingletonAssert();
+
+    self = [super init];
+    if (!self) {
+        return self;
+    }
+
+    _dbConnection = dbConnection;
+
+    return self;
+}
+
+- (nullable OWSMessageProcessingJob *)nextJob
+{
+    __block OWSMessageProcessingJob *_Nullable job;
+    [self.dbConnection readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
+        YapDatabaseViewTransaction *viewTransaction = [transaction ext:OWSMessageProcessingJobFinderExtensionName];
+        OWSAssert(viewTransaction != nil);
+        job = [viewTransaction firstObjectInGroup:OWSMessageProcessingJobFinderExtensionGroup];
+    }];
+
+    return job;
+}
+
+- (void)addJobForEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
+{
+    [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
+        [[[OWSMessageProcessingJob alloc] initWithEnvelope:envelope] saveWithTransaction:transaction];
+    }];
+}
+
+- (void)removeJobWithId:(NSString *)uniqueId
+{
+    [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
+        [transaction removeObjectForKey:uniqueId inCollection:[OWSMessageProcessingJob collection]];
+    }];
+}
+
++ (YapDatabaseView *)databaseExension
+{
+
+    YapDatabaseViewSorting *sorting =
+        [YapDatabaseViewSorting withObjectBlock:^NSComparisonResult(YapDatabaseReadTransaction *transaction,
+            NSString *group,
+            NSString *collection1,
+            NSString *key1,
+            id object1,
+            NSString *collection2,
+            NSString *key2,
+            id object2) {
+
+            if (![object1 isKindOfClass:[OWSMessageProcessingJob class]]) {
+                OWSFail(@"Unexpected object: %@ in collection: %@", object1, collection1) return NSOrderedSame;
+            }
+            OWSMessageProcessingJob *job1 = (OWSMessageProcessingJob *)object1;
+
+            if (![object2 isKindOfClass:[OWSMessageProcessingJob class]]) {
+                OWSFail(@"Unexpected object: %@ in collection: %@", object2, collection2) return NSOrderedSame;
+            }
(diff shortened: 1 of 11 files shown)


SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -0,0 +1,338 @@
...
+#import <YapDatabase/YapDatabaseView.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+#pragma mark - Persisted data model
+@class OWSSignalServiceProtosEnvelope;
+
+@interface OWSMessageProcessingJob : TSYapDatabaseObject
+
+@property (nonatomic, readonly) OWSSignalServiceProtosEnvelope *envelopeProto;
```
charlesmchen: If we make `envelopeProto` a property, will Mantle try to serialize it?

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -0,0 +1,338 @@
...
+@property (nonatomic, readonly) NSDate *createdAt;
```
charlesmchen: Should we use this value to initialize `TSMessage.receivedAtTimestamp`?  

* That'd make the "received" time a bit more accurate.
* It'd preserve ordering for messages we decrypt "later", e.g. after receiving a response to a group info update.  That might be desirable or not.
* It might have some surprisingly side effects in other cases where we process the queue out of received order.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -0,0 +1,338 @@
...
+
+@end
+
+@interface OWSMessageProcessingJob ()
+
+@property (nonatomic, readonly) NSData *envelopeData;
+
+@end
+
+// TODO rename?
```
charlesmchen: `OWSReceivedMessage`?  `OWSIncomingMessageData`?

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -0,0 +1,338 @@
...
+
+    _envelopeData = envelope.data;
+    _createdAt = [NSDate new];
+
+    return self;
+}
+
+- (OWSSignalServiceProtosEnvelope *)envelopeProto
+{
+    return [OWSSignalServiceProtosEnvelope parseFromData:self.envelopeData];
```
charlesmchen: If we received an invalid envelope, would this return null?  Should this have a nullable return value?

charlesmchen: I guess this class is initialized with a valid envelope so it should be valid unless we broke our proto schema in the meantime.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -0,0 +1,338 @@
...
+@implementation OWSMessageProcessingJob
+
+- (instancetype)initWithEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
+{
+    self = [super initWithUniqueId:[NSUUID new].UUIDString];
+    if (!self) {
+        return self;
+    }
+
+    _envelopeData = envelope.data;
```
charlesmchen: OWSAssert(envelope);

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -0,0 +1,338 @@
...
+- (void)removeJobWithId:(NSString *)uniqueId
+{
+    [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
+        [transaction removeObjectForKey:uniqueId inCollection:[OWSMessageProcessingJob collection]];
+    }];
+}
+
++ (YapDatabaseView *)databaseExension
+{
+
```
charlesmchen: Empty line

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -0,0 +1,338 @@
...
+            NSString *group,
+            NSString *collection1,
+            NSString *key1,
+            id object1,
+            NSString *collection2,
+            NSString *key2,
+            id object2) {
+
+            if (![object1 isKindOfClass:[OWSMessageProcessingJob class]]) {
+                OWSFail(@"Unexpected object: %@ in collection: %@", object1, collection1) return NSOrderedSame;
```
charlesmchen: missing semicolon before return statement, here and below.

charlesmchen: I suggest you log `[object1 class]`, not `object1`.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -0,0 +1,338 @@
...
+            return [job1.createdAt compare:job2.createdAt];
+        }];
+
+    YapDatabaseViewGrouping *grouping =
+        [YapDatabaseViewGrouping withObjectBlock:^NSString *_Nullable(YapDatabaseReadTransaction *_Nonnull transaction,
+            NSString *_Nonnull collection,
+            NSString *_Nonnull key,
+            id _Nonnull object) {
+            if (![object isKindOfClass:[OWSMessageProcessingJob class]]) {
+                OWSFail(@"Unexpected object: %@ in collection: %@", object, collection) return nil;
```
charlesmchen: ;

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -0,0 +1,338 @@
...
+        [[YapWhitelistBlacklist alloc] initWithWhitelist:[NSSet setWithObject:[OWSMessageProcessingJob collection]]];
+
+    return [[YapDatabaseView alloc] initWithGrouping:grouping sorting:sorting versionTag:@"1" options:options];
+}
+
+
++ (void)syncRegisterDatabaseExtension:(YapDatabase *)database
+{
+    YapDatabaseView *existingView = [database registeredExtension:OWSMessageProcessingJobFinderExtensionName];
+    if (existingView) {
```
charlesmchen: Don't we expect to be registered exactly once?  Can we assert or fail if its already registered?  That would indicate, for example, that perhaps two views have the same name or something.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -0,0 +1,338 @@
...
+}
+
+#pragma mark - instance methods
+
+- (void)enqueueEnvelopeForProcessing:(OWSSignalServiceProtosEnvelope *)envelope
+{
+    [self.finder addJobForEnvelope:envelope];
+}
+
+- (void)drainQueue
```
charlesmchen: This method name feels misleading, since this method won't drain the queue if there is more than one job queued.  `processQueue` or `processNextJob`, etc.?

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -0,0 +1,338 @@
...
+              completion:^{
+                  [self drainQueue];
+              }];
+    });
+}
+
+- (void)processJob:(OWSMessageProcessingJob *)job completion:(void (^)())completion
+{
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [self.messagesManager processEnvelope:job.envelopeProto
```
charlesmchen: So if we were going to build upon this work to retain jobs that couldn't be processed (e.g. messages for unknown groups), I assume this completion would have a success parameter, and we'd add a "failed" flag and a "thread id" to jobs and we'd modify the view to ignore failed jobs, and when we got a group update we'd clear the failed flag for all jobs in that thread?

SignalServiceKit/src/Messages/OWSMessageReceiver.m
michaelkirk: I thought about this for a while, and it's not clear to me what's best.

You bring up so good points, but my conclusions was that even if it means the resultant received history is "out of order" with respect to when it was sent, we never want to insert a new message "up" in the conversation, because the recipient might miss the message.

In practice, we currently process the messages in the order in which they were received due to the YapDBView which sorts by createdAt. So to be clear, currently we're processing messages in the same order before and after this PR.

I'm inclined to leave it as is unless we can make a change this is (at least mostly) uncontroversial.

And specifically in the case of the group info update, we probably want the ordering to reflect *processing* time, so the "group update" would appear, *and then* the message. Right?

I think one thing at the root of this is our distinction between sort order and display date. To get in line with android, we short sort by received date, but display the sending users timestamp in the UI. There is a problem when the senders clock is busted, but in practice I guess this seems to have provided more value than always showing the "Received at" time. 

I think a similar approach (sorting by received at, but displaying senders time) would better serve our purposes in the long run - but it probably only makes sense once we have a visible per message timestamp. Otherwise with only the current date headers, it would be easy to have an **even more** confusing situation when the sender's clock is off e.g. in a group.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
michaelkirk: done

SignalServiceKit/src/Messages/OWSMessageReceiver.m
michaelkirk: Done and done!

SignalServiceKit/src/Messages/OWSMessageReceiver.m
michaelkirk: done.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
michaelkirk: Good point - added an OWSFail

SignalServiceKit/src/Messages/OWSMessageReceiver.m
michaelkirk: done!

SignalServiceKit/src/Messages/OWSMessageReceiver.m
michaelkirk: Hm... did you see that it's recursive? 

We call it every time we enqueue a job, so in practice it will usually only process one job, but if there are any more, it'll run them.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
michaelkirk: In the case of "unknown group" messages, I think that's problematic. We can't re-decrypt a message. Once we decrypt it, we throw away the key, so we'd have to persist the decrypted portion.

Maybe we ultimately need to split this into a DecryptEnvelopeJob and a ProcessDecryptedEnvelopeJob (names are a WIP 😉 ). Doing that would require a non-trivial refactor of TSMessagesManager, but I think it would be appropriate. We have way too much Signal-iOS specific app logic in TSMessagesManager as is.

Does that seem like the right approach?

And if so, does merging this as is move us closer, farther, laterally from that goal?

I feel like it moves us a little closer, in that now we have a persistent queue of encrypted envelopes. The next step would be splitting out some functionality and creating a persistent store of decrypted envelopes (waiting to be processed into something user visible).

SignalServiceKit/src/Messages/OWSMessageReceiver.m
charlesmchen: Yeah, it would make sense as a method name if its was purely recursive, but since it dispatches async, I tend to think the name should reflect the effect or work done by a single invocation.  Having said that, I was just asking.  Fine to leave as it is.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
charlesmchen: I agree.  I just wanted to raise the question(s).

SignalServiceKit/src/Messages/OWSMessageReceiver.m
charlesmchen: Yup, let's merge this.  You're right, if we did that we'd want a separate "decrypted" queue.

michaelkirk:
In 8b724c5d

PR #2358, opened 2017-07-26
Make it easier to tap-to-edit group names in new group and update gro…

charlesmchen:
…up views.

PTAL @michaelkirk


darrenchaker:
Thank you for the updates! Darren Chaker

PR #2360, opened 2017-07-26
Instrument registration happy path with analytics.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/CodeVerificationViewController.m
+++ b/Signal/src/ViewControllers/CodeVerificationViewController.m
@@ -260,8 +260,11 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)verifyChallengeAction:(nullable id)sender
 {
     [self startActivityIndicator];
+    OWSProdInfo(@"registration_registering_code");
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/CodeVerificationViewController.m
```diff
@@ -260,8 +260,11 @@ - (void)stopActivityIndicator
 - (void)verifyChallengeAction:(nullable id)sender
 {
     [self startActivityIndicator];
+    OWSProdInfo(@"registration_registering_code");
```
michaelkirk: maybe: `registration_registering_submitted_code`
And separately maybe: `registration_registering_requested_new_code`

charlesmchen: Done.

charlesmchen:
> Do you think the number of people who go "back" to update their phone number is negligible?

I've added an explicit event.  It's an "info" message in an edge case of an infrequent workflow, so I'm not worried about the volume.

PR #2362, opened 2017-07-26
Properly handle too-large messages

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Messages/OWSMessageReceiver.m
+++ b/SignalServiceKit/src/Messages/OWSMessageReceiver.m
@@ -333,6 +333,20 @@ NSString *const OWSMessageProcessingJobFinderExtensionGroup = @"OWSMessageProces
 
 - (void)handleReceivedEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
 {
+    // Drop any too-large messages on the floor. Well behaving clients should never send them.
+    NSUInteger kMaxEnvelopeByteCount = 250 * 1024;
+    if (envelope.serializedSize > kMaxEnvelopeByteCount) {
+        OWSProdError(@"message_receiver_error_oversize_message");
+        return;
+    }
+
+    // Take note of any messages larger than we expect, but still process them.
+    // This likely indicates a misbehaving sending client.
+    NSUInteger kLargeEnvelopeWarningByteCount = 25 * 1024;
+    if (envelope.serializedSize > kLargeEnvelopeWarningByteCount) {
+        OWSProdError(@"message_receiver_error_large_message");
+    }
+
     [self.processingQueue enqueueEnvelopeForProcessing:envelope];
     [self.processingQueue drainQueue];
 }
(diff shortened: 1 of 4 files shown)


SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -333,6 +333,28 @@ + (void)syncRegisterDatabaseExtension:(YapDatabase *)database
...
+            return (@{
+                @"message_size" : @([OWSAnalytics orderOfMagnitudeOf:(long)envelope.serializedSize]),
+            });
+        });
+        return;
+    }
+
+    // Take note of any messages larger than we expect, but still process them.
+    // This likely indicates a misbehaving sending client.
+    NSUInteger kLargeEnvelopeWarningByteCount = 25 * 1024;
```
michaelkirk: I picked this randomly. We could probably be even more conservative.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -333,6 +333,28 @@ + (void)syncRegisterDatabaseExtension:(YapDatabase *)database
 
 - (void)handleReceivedEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
 {
+    // Drop any too-large messages on the floor. Well behaving clients should never send them.
+    NSUInteger kMaxEnvelopeByteCount = 250 * 1024;
+    if (envelope.serializedSize > kMaxEnvelopeByteCount) {
+        OWSProdErrorWParams(@"message_receiver_error_oversize_message", ^{
+            return (@{
+                @"message_size" : @([OWSAnalytics orderOfMagnitudeOf:(long)envelope.serializedSize]),
```
charlesmchen: Per our privacy review, please just use `OWSProdError()` without the params.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -333,6 +333,28 @@ + (void)syncRegisterDatabaseExtension:(YapDatabase *)database
...
+            });
+        });
+        return;
+    }
+
+    // Take note of any messages larger than we expect, but still process them.
+    // This likely indicates a misbehaving sending client.
+    NSUInteger kLargeEnvelopeWarningByteCount = 25 * 1024;
+    if (envelope.serializedSize > kLargeEnvelopeWarningByteCount) {
+        OWSProdErrorWParams(@"message_receiver_error_large_message", ^{
```
charlesmchen: Ditto.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
charlesmchen: I'm guessing that the largest valid envelope should be some small multiple of the largest possible text message.

```
        const NSUInteger kOversizeTextMessageSizeThreshold = 16 * 1024;
```

I'm guessing that a 16k text message would yield an envelope smaller than 25k, but we should verify that?

Issue #2346, opened 2017-07-21
Signal crashes when trying to open a conversation (but not all of them)


ejpcmac:
- [x] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description

My mother cannot open her conversation with me on Signal: when she tries to open it, Signal crashes. She can open other conversations though, even a group I am in. She also sees that she have one unread message from me, and its content in the conversations list.

### Steps to reproduce

I don’t know how to reproduce this bug in another conversation, neither on her phone nor on mine. It is specific to her conversation with me, on her iPhone.

**Actual result:** Signal crashes.
**Expected result:** The conversation should be opened.

### Device info

**Device**: iPhone 4S
**iOS version**: 9.3.5
**Signal version:** 2.14.1 (I just did the upgrade, but the bug was there in the previous version)

### Link to debug log

https://gist.github.com/anonymous/b880dd4a55f459c3633e07ceabae2d9c

Note: I just activated the loggin, then opened a conversation that opens correctly, then opend her conversation with me, which crashes.

Strangly enough, there is no recent Signal crash log in the diagnostics data.

PR #2363, opened 2017-07-26
ensure blocking keychange message has identityKey before proceeding

michaelkirk:
This is increasingly irrelevant due to changes around safety number handling, but older clients with legacy messages can still run into this.

FIXES: https://github.com/WhisperSystems/Signal-iOS/issues/2346

PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Messages/InvalidKeyMessages/TSInvalidIdentityKeySendingErrorMessage.m
+++ b/SignalServiceKit/src/Messages/InvalidKeyMessages/TSInvalidIdentityKeySendingErrorMessage.m
@@ -50,12 +50,18 @@ NSString *TSInvalidRecipientKey = @"TSInvalidRecipientKey";
     OWSFail(@"accepting new identity key is deprecated.");
 
     // Saving a new identity mutates the session store so it must happen on the sessionStoreQueue
+    NSData *_Nullable newIdentityKey = self.newIdentityKey;
+    if (!newIdentityKey) {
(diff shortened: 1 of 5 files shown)


SignalServiceKit/src/Messages/InvalidKeyMessages/TSInvalidIdentityKeySendingErrorMessage.m
```diff
@@ -50,12 +50,18 @@ - (void)acceptNewIdentityKey
     OWSFail(@"accepting new identity key is deprecated.");
 
     // Saving a new identity mutates the session store so it must happen on the sessionStoreQueue
+    NSData *_Nullable newIdentityKey = self.newIdentityKey;
+    if (!newIdentityKey) {
+        OWSFail(@"newIdentityKey is unexpectedly nil. Bad Prekey bundle?: %@", self.preKeyBundle);
```
charlesmchen: Is it safe to log prekey bundles?

michaelkirk:
> LGTM I don't see a corresponding crash in XCode crash organizer.

Came from the GH report #2346. Sounds like they don't have automatic crash reporting enabled

SignalServiceKit/src/Messages/InvalidKeyMessages/TSInvalidIdentityKeySendingErrorMessage.m
michaelkirk: No privacy concerns.

PR #2365, opened 2017-07-27
Charlesmchen/gather analytics event names

charlesmchen:
PTAL @michaelkirk


charlesmchen:
@michaelkirk 95% of the changes in this PR were code generated by a python script.  I didn't bother cleaning up that python script much since I'm not sure if we're going to use it going forward.  We could - like the l10n scripts it normalizes, sorts, and ensures a consistent mapping between the "event name" and the method name used to access it.

SignalServiceKit/src/Util/OWSAnalyticsEvents.h
```diff
@@ -0,0 +1,139 @@
...
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OWSAnalyticsEvents : NSObject
+
+- (instancetype)init NS_UNAVAILABLE;
+
+#pragma mark - Code Generation Marker
```
michaelkirk: Maybe a comment to tie this back to the python script to talk about how this code was generated.

SignalServiceKit/Utilities/extract_analytics_event_names.py
```diff
@@ -0,0 +1,368 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import os
```
michaelkirk: maybe a line or two comment explaining what this was used for.

michaelkirk:
This will conflict with https://github.com/WhisperSystems/Signal-iOS/commit/898f122d2b02d8eb1332a62f9965b4b0cb65aa90 which modified some event names - so hopefully you just need to re-run that script.

SignalServiceKit/src/Util/OWSAnalyticsEvents.h
charlesmchen: Done.

SignalServiceKit/Utilities/extract_analytics_event_names.py
charlesmchen: Done. On reflection, I think we should continue to use this script as we add more analytics events.

charlesmchen:
> This will conflict with 898f122 which modified some event names - so hopefully you just need to re-run that script.

Yup, no problem.  Re-ran the script.

PR #2366, opened 2017-07-27
Fix assert in conversation view around nil title.

charlesmchen:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -545,6 +545,8 @@ - (void)viewWillAppear:(BOOL)animated
     // or on another device.
     [self hideInputIfNeeded];
 
+    [self setNavigationTitle];
```
michaelkirk: We're also calling this 14 lines down. If we really need to call it twice, can you leave an explanatory comment for this precarious situation?

charlesmchen:
PTAL @michaelkirk I found a better solution.

PR #2367, opened 2017-07-27
Hide empty conversations in home view.

charlesmchen:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -527,9 +527,6 @@ - (void)viewWillAppear:(BOOL)animated
     // We need to update the dynamic interactions before we do any layout.
     [self ensureDynamicInteractions];
 
-    // Triggering modified notification renders "call notification" when leaving full screen call view
-    [self.thread touch];
```
charlesmchen: This shouldn't be necessary - we're observing yap database changes and saving interactions touches the interaction's thread.

PR #2368, opened 2017-07-27
Charlesmchen/hide empty conversations

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -171,6 +171,9 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
     NSString *lastLaunchedAppVersion = AppVersion.instance.lastAppVersion;
     NSString *lastCompletedLaunchAppVersion = AppVersion.instance.lastCompletedLaunchAppVersion;
+    // Every time we change a database view in such a way that might cause a delay on launch,
+    // we need to bump this constant.
+    //
     // We added a number of database views in v2.13.0.
     NSString *kLastVersionWithDatabaseViewChange = @"2.13.0";
     BOOL mayNeedUpgrade = ([TSAccountManager isRegistered] && lastLaunchedAppVersion
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -528,9 +528,6 @@ typedef enum : NSUInteger {
     // We need to update the dynamic interactions before we do any layout.
     [self ensureDynamicInteractions];
 
-    // Triggering modified notification renders "call notification" when leaving full screen call view
-    [self.thread touch];
-
     [self ensureBannerState];
 
     [super viewWillAppear:animated];
--- a/SignalServiceKit/src/Messages/Interactions/TSInteraction.m
+++ b/SignalServiceKit/src/Messages/Interactions/TSInteraction.m
@@ -114,6 +114,13 @@ NS_ASSUME_NONNULL_BEGIN
     [fetchedThread updateWithLastMessage:self transaction:transaction];
 }
 
+- (void)removeWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
+{
+    [super removeWithTransaction:transaction];
+
+    [self touchThreadWithTransaction:transaction];
+}
+
 - (BOOL)isDynamicInteraction
 {
     return NO;
--- a/SignalServiceKit/src/Storage/TSDatabaseView.h
+++ b/SignalServiceKit/src/Storage/TSDatabaseView.h
@@ -25,7 +25,8 @@ extern NSString *const TSSecondaryDevicesDatabaseViewExtensionName;
 
 + (BOOL)hasPendingViewRegistrations;
 
-+ (void)registerThreadDatabaseView;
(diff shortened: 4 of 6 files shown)


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -527,9 +527,6 @@ - (void)viewWillAppear:(BOOL)animated
     // We need to update the dynamic interactions before we do any layout.
     [self ensureDynamicInteractions];
 
-    // Triggering modified notification renders "call notification" when leaving full screen call view
-    [self.thread touch];
```
charlesmchen: This shouldn't be necessary - we're observing yap database changes and saving interactions touches the interaction's thread.

Signal/src/AppDelegate.m
```diff
@@ -172,10 +172,13 @@ - (UIViewController *)loadingRootViewController
...
     NSString *lastCompletedLaunchAppVersion = AppVersion.instance.lastCompletedLaunchAppVersion;
+    // Every time we change a database view in such a way that might cause a delay on launch,
+    // we need to bump this constant.
+    NSString *lastVersionWithDatabaseViewChange = @"2.13.0";
     BOOL mayNeedUpgrade = ([TSAccountManager isRegistered] &&
-        [VersionMigrations isVersion:lastAppVersion atLeast:@"2.0.0" andLessThan:@"2.13.0"]);
+        [VersionMigrations isVersion:lastAppVersion atLeast:@"2.0.0" andLessThan:lastVersionWithDatabaseViewChange]);
     BOOL hasCompletedUpgrade = (lastCompletedLaunchAppVersion &&
-        [VersionMigrations isVersion:lastCompletedLaunchAppVersion atLeast:@"2.13.0"]);
+        [VersionMigrations isVersion:lastCompletedLaunchAppVersion atLeast:lastVersionWithDatabaseViewChange]);
```
charlesmchen: I just wanted to constant-ize this and add a comment, but we don't need to bump the constant with this PR.

SignalServiceKit/src/Messages/Interactions/TSInteraction.m
```diff
@@ -114,6 +114,13 @@ - (void)saveWithTransaction:(YapDatabaseReadWriteTransaction *)transaction {
     [fetchedThread updateWithLastMessage:self transaction:transaction];
 }
 
+- (void)removeWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
+{
+    [super removeWithTransaction:transaction];
+
+    [self touchThreadWithTransaction:transaction];
```
charlesmchen: We need to touch a message's thread when deleting that message.

Podfile.lock
```diff
@@ -125,7 +125,7 @@ EXTERNAL SOURCES:
     :branch: signal-master
     :git: https://github.com/WhisperSystems/JSQMessagesViewController.git
   SignalServiceKit:
-    :path: "."
```
michaelkirk: What version of cocoapods are you using? Can you use the version specified in the Gemfile.lock?

 (or just run `bundle` to install the correct version, then invoke with `bundle exec pod install`)

Signal/src/AppDelegate.m
```diff
@@ -172,10 +172,13 @@ - (UIViewController *)loadingRootViewController
...
 
     // We added a number of database views in v2.13.0.
     if (mayNeedUpgrade && !hasCompletedUpgrade) {
```
michaelkirk: I may be missing some of the nuance here, but can we simplify this to something like:

```
BOOL shouldShowUpgradeLabel = [TSAccountManager isRegistered] && lastCompletedLaunchAppVersion && ![VersionMigrations isVersion:lastCompletedLaunchAppVersion atLeast:lastVersionWithDatabaseViewChange]
```

I believe the only affect of:
```
[VersionMigrations isVersion:lastAppVersion atLeast:@"2.0.0" andLessThan:lastVersionWithDatabaseViewChange]);
```

would be to omit the label if the app didn't complete upgrading the last time it launched this version. Which seems undesirable.

Podfile.lock
charlesmchen: `Gemfile.lock` says:
```
    cocoapods (1.2.1)
```

```pod --version``` says:

```
1.2.1
```
So I appear to be on the right version.  What's your output for `pod --version`?

File: Signal/src/AppDelegate.m (hunk 5 above)
Comment (charlesmchen) [inline Signal/src/AppDelegate.m]: Fixed this by merging in the unmerged hotfixes =/.

File: Podfile.lock (hunk 4 above)
Comment (michaelkirk) [inline Podfile.lock]: oh geeze. Seems like it's a ruby thing: https://github.com/CocoaPods/CocoaPods/issues/4771

We can either:
1. track a `.ruby-version` file with the project and use it.
2. you can tell me version you're running and I'll meet you there.

FYI - I'm at 2.41
```
→ ruby --version
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
```

Comment (charlesmchen) [inline Podfile.lock]: Weird, I'm already at the same ruby?
```
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin15]
```

michaelkirk: 🏳️

PR #2369, opened 2017-07-27
Avoid crash when deleting threads - and improve perf.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
https://trello.com/c/yr3Oeouh/1130-2140-crash-when-trying-to-delete-thread

I'm guessing that the root cause would be that a failure while trying to instantiate `TSInteraction` -
 perhaps because we deleted or renamed the class so Mantle/YapDatabase can't deserialize it? - but in the absence of knowing more, this will allow users to delete all threads.

PR #2377, opened 2017-07-28
Tweaks to registration views.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/CodeVerificationViewController.m
+++ b/Signal/src/ViewControllers/CodeVerificationViewController.m
@@ -262,13 +262,14 @@ NS_ASSUME_NONNULL_BEGIN
 {
     [self startActivityIndicator];
     OWSProdInfo([OWSAnalyticsEvents registrationRegisteringCode]);
+    __weak CodeVerificationViewController *weakSelf = self;
     [self.accountManager registerWithVerificationCode:[self validationCodeFromTextField]]
         .then(^{
             OWSProdInfo([OWSAnalyticsEvents registrationRegisteringSubmittedCode]);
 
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/CodeVerificationViewController.m
```diff
@@ -262,13 +262,14 @@ - (void)verifyChallengeAction:(nullable id)sender
 {
     [self startActivityIndicator];
     OWSProdInfo([OWSAnalyticsEvents registrationRegisteringCode]);
+    __weak CodeVerificationViewController *weakSelf = self;
```
charlesmchen: NBD. Using weak self will let this view be cleaned up a bit faster if user hits back during a long network request.

Signal/src/ViewControllers/RegistrationViewController.m
```diff
@@ -61,6 +61,9 @@ - (void)viewDidLoad {
 - (void)createViews
 {
     self.view.backgroundColor = [UIColor ows_signalBrandBlueColor];
+    self.view.userInteractionEnabled = YES;
+    [self.view
+        addGestureRecognizer:[[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(backgroundTapped:)]];
```
charlesmchen: On "verification code" view, tapping the background _deselects_ the text field, since there's a ton of cramped UI in that view and it might help access/view the bottom elements in some languages.  

On "registration" view we want to do the opposite since its less cramped - errant taps should select the text field, to help nudge the users through the registration process.

charlesmchen:
@michaelkirk I'm going to merge this after the release.

PR #2379, opened 2017-07-28
Mkirk/drain queue perf

michaelkirk:
Two things:

1. Don't let our dispatch queue amplify
2. Kick the queue when the app becomes active

PTAL @charlesmchen

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -517,6 +517,9 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
                 // This will fetch new messages, if we're using domain
                 // fronting.
                 [[PushManager sharedManager] applicationDidBecomeActive];
+
+                // If there were any messages in our local queue which we hadn't yet processed.
+                [[OWSMessageReceiver sharedInstance] handleAnyUnprocessedEnvelopes];
             }];
 
     DDLogInfo(@"%@ applicationDidBecomeActive completed.", self.tag);
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -517,9 +517,6 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
                 // This will fetch new messages, if we're using domain
                 // fronting.
                 [[PushManager sharedManager] applicationDidBecomeActive];
-
-                // If there were any messages in our local queue which we hadn't yet processed.
-                [[OWSMessageReceiver sharedInstance] handleAnyUnprocessedEnvelopes];
             }];
 
     DDLogInfo(@"%@ applicationDidBecomeActive completed.", self.tag);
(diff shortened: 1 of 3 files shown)


Signal/src/AppDelegate.m
```diff
@@ -517,6 +517,9 @@ - (void)applicationDidBecomeActive:(UIApplication *)application {
                 // This will fetch new messages, if we're using domain
                 // fronting.
                 [[PushManager sharedManager] applicationDidBecomeActive];
+
+                // If there were any messages in our local queue which we hadn't yet processed.
+                [[OWSMessageReceiver sharedInstance] handleAnyUnprocessedEnvelopes];
```
charlesmchen: Why not have `OWSMessageReceiver` observe `UIApplicationDidBecomeActiveNotification`?  I prefer that approach so that everything is independent as possible.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -220,16 +222,34 @@ - (void)enqueueEnvelopeForProcessing:(OWSSignalServiceProtosEnvelope *)envelope
...
 {
-    dispatch_async(self.class.serialGCDQueue, ^{
-        OWSMessageProcessingJob *_Nullable job = [self.finder nextJob];
-        if (job == nil) {
-            DDLogVerbose(@"%@ Queue is drained", self.tag);
-            return;
-        }
+    AssertIsOnMainThread();
+
+    if (self.isDrainingQueue) {
```
charlesmchen: I'm not clear on why this was necessary. `serialGCDQueue` already ensured that only one job was being parsed at a time - is that we were queuing up a ton of work on `serialGCDQueue`?  Shouldn't that be cheap?

Put another way: can we now get rid of `serialGCDQueue` if we're going to use a flag to ensure that only one job is processing at a time?

Signal/src/AppDelegate.m
```diff
@@ -793,6 +793,9 @@ - (void)databaseViewRegistrationComplete
     [AppVersion.instance appLaunchDidComplete];
 
     [self ensureRootViewController];
+
+    // If there were any messages in our local queue which we hadn't yet processed.
+    [[OWSMessageReceiver sharedInstance] handleAnyUnprocessedEnvelopesAsync];
```
charlesmchen: FYI we could observe kNSNotificationName_DatabaseViewRegistrationComplete but this is fine too.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -220,41 +222,43 @@ - (void)enqueueEnvelopeForProcessing:(OWSSignalServiceProtosEnvelope *)envelope
...
-+ (dispatch_queue_t)serialGCDQueue
+- (void)processJob:(OWSMessageProcessingJob *)job completion:(void (^)())completion
 {
-    static dispatch_once_t onceToken;
-    static dispatch_queue_t queue;
-    dispatch_once(&onceToken, ^{
-        queue = dispatch_queue_create("org.whispersystems.signal.messageProcessingQueue", NULL);
-    });
-    return queue;
+    [self.messagesManager processEnvelope:job.envelopeProto
```
charlesmchen: I'd love to move processing of envelopes off the main queue at some point. Thoughts?

michaelkirk: Yes, that'd be swell.

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/2a6df19e0e320cb113f8fba8337ef9f7a0fae8ba

PR #2383, opened 2017-07-31
Throttle prekey checks.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -248,6 +248,8 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
     self.staleNotificationObserver = [OWSStaleNotificationObserver new];
     [self.staleNotificationObserver startObserving];
+
+    [TSPreKeyManager sharedManager];
 }
 
 - (void)application:(UIApplication *)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken
@@ -448,9 +450,6 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     [self removeScreenProtection];
 
     [self ensureRootViewController];
-
-    // Always check prekeys after app launches, and sometimes check on app activation.
-    [TSPreKeyManager checkPreKeysIfNecessary];
 
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
--- a/SignalServiceKit/src/Account/TSPreKeyManager.m
+++ b/SignalServiceKit/src/Account/TSPreKeyManager.m
@@ -236,16 +236,12 @@ static const NSTimeInterval kSignedPreKeyUpdateFailureMaxFailureDuration = 10 * 
 // This method is called whenever the app activates and is less insistent.
 //
 // Under normal circumstances, we don't need to check that often.
-- (void)checkPreKeysOnActivation
++ (void)checkPreKeysOnActivation
 {
+    OWSAssert([NSThread isMainThread]);
     OWSAssert([UIApplication sharedApplication].applicationState == UIApplicationStateActive);
 
-    [[TSAccountManager sharedInstance] ifRegistered:YES
-                                           runAsync:^{
-                                               dispatch_async(TSPreKeyManager.serialQueue, ^{
-                                                   [self checkPreKeysWithMaxFrequency:kPreKeyCheckSlowFrequencySeconds];
-                                               });
-                                           }];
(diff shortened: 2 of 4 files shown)


Signal/src/AppDelegate.m
```diff
@@ -248,6 +248,8 @@ - (void)setupEnvironment
 
     self.staleNotificationObserver = [OWSStaleNotificationObserver new];
     [self.staleNotificationObserver startObserving];
+
+    [TSPreKeyManager sharedManager];
```
charlesmchen: I moved observation of the activation event into `TSPreKeyManager` in order to gather concerns.

SignalServiceKit/src/Account/TSPreKeyManager.m
```diff
@@ -24,15 +27,17 @@
 // How often we check prekey state on app activation.
 //
 // Currently we check prekey state every 12 hours.
-static const NSTimeInterval kPreKeyCheckFrequencySeconds = 12 * kHourInterval;
+static const NSTimeInterval kPreKeyCheckSlowFrequencySeconds = 12 * kHourInterval;
+
+// How often we check prekey state on receiving a PreKeyWhisperMessage.
+//
+// Currently we check prekey state every 2 minutes.
+static const NSTimeInterval kPreKeyCheckFastFrequencySeconds = 2 * kMinuteInterval;
```
charlesmchen: I've added a second threshold that applies for `PreKeyWhisperMessage`.

SignalServiceKit/src/Account/TSPreKeyManager.m
```diff
@@ -45,8 +50,53 @@
...
+// This property should only be accessed on serialQueue.
+@property (atomic) NSDate *lastPreKeyCheckTimestamp;
+
+@end
+
+#pragma mark -
+
 @implementation TSPreKeyManager
 
++ (instancetype)sharedManager
```
charlesmchen: I made this a singleton to be consistent and so that we could enable NSNotifications.

SignalServiceKit/src/Messages/TSMessagesManager.m
```diff
@@ -447,7 +447,7 @@ - (void)handlePreKeyBundleAsync:(OWSSignalServiceProtosEnvelope *)preKeyEnvelope
             NSData *plaintextData;
             @try {
                 // Check whether we need to refresh our PreKeys every time we receive a PreKeyWhisperMessage.
-                [TSPreKeyManager checkPreKeys];
+                [TSPreKeyManager didReceivePreKeyWhisperMessage];
```
charlesmchen: I made a bunch of small changes to make things (hopefully) more readable.

SignalServiceKit/src/Account/TSPreKeyManager.m
```diff
@@ -45,8 +50,53 @@
 
 #pragma mark -
 
+@interface TSPreKeyManager ()
+
+// This property should only be accessed on serialQueue.
+@property (atomic) NSDate *lastPreKeyCheckTimestamp;
```
michaelkirk: nit: `nullable`

Signal/src/AppDelegate.m
michaelkirk: Since we're using this line of code (a getter) only for it's side effects, I think we need a comment.

But at that point, it might be more concise to move the `observation` out of the init, and just say `[[TSPreKeyManager sharedManager] observeNotifications]` which speaks for itself.

SignalServiceKit/src/Messages/TSMessagesManager.m
michaelkirk: 👍

michaelkirk: The time window is probably sufficiently small for almost all users, but I'm wondering about public signal numbers (e.g. hotlines) where it's likely they get a deluge of messages from first-time-senders in a short time. What if instead we parameterized this with the PreKeyWhisperMessage on line 452, like this:

```
PreKeyWhisperMessage *message = [[PreKeyWhisperMessage alloc] initWithData:encryptedData];
[TSPreKeyManager didReceivePreKeyWhisperMessageWithPrekeyID:message.preKeyId];
```

Then PreKeyManager could keep an in-memory cache to check only once per preKeyId (per launch), rather than dealing with timeouts.

One complication, we want to make sure we **always** check prekey count when receiving a message without a  one time PreKey. Though that's probably true of the existing PR too.

Imperiopolis-Signal:

PR #2384, opened 2017-07-31
Sketch out profile manager.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Profiles/OWSProfilesManager.m
+++ b/SignalServiceKit/src/Profiles/OWSProfilesManager.m
@@ -1,0 +1,144 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "OWSProfilesManager.h"
+#import "OWSMessageSender.h"
+#import "SecurityUtils.h"
+#import "TSStorageManager.h"
+#import "TextSecureKitEnv.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+NSString *const kOWSProfilesManager_Collection = @"kOWSProfilesManager_Collection";
+// This key is used to persist the local user's profile key.
+NSString *const kOWSProfilesManager_LocalProfileKey = @"kOWSProfilesManager_LocalProfileKey";
+
+// TODO:
+static const NSInteger kProfileKeyLength = 16;
+
+@interface OWSProfilesManager ()
+
+@property (nonatomic, readonly) TSStorageManager *storageManager;
+@property (nonatomic, readonly) OWSMessageSender *messageSender;
+
+@property (nonatomic, readonly, nullable) NSData *localProfileKey;
+
+@end
+
+#pragma mark -
+
+@implementation OWSProfilesManager
+
+@synthesize localProfileKey = _localProfileKey;
+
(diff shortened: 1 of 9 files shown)


SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -0,0 +1,144 @@
...
+#pragma mark - Local Profile Key
+
++ (NSData *)generateLocalProfileKey
+{
+    // TODO:
+    OWSFail(@"Profile key generation is not yet implemented.");
+    return [SecurityUtils generateRandomBytes:kProfileKeyLength];
+}
+
+- (nullable NSData *)localProfileKey
```
michaelkirk: Is this really nullable? If so, our assert is going to fail.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -0,0 +1,144 @@
...
+    _messageSender = messageSender;
+
+    OWSSingletonAssert();
+
+    // Register this manager with the message sender.
+    // This is a circular dependency.
+    [messageSender setProfilesManager:self];
+
+    // Try to load.
+    _localProfileKey = [self.storageManager objectForKey:kOWSProfilesManager_LocalProfileKey
```
michaelkirk: As a pattern, I'm starting to resent using the `storageManager`'s internal `dbConnection` for anything. 

Instead, I think I prefer the general approach that each "manager"-like-thing has it's own dbConnection, which we use to read/write any of the objects "owned" by that manager. 

WDYT @charlesmchen?

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: Yup, this is just a first draft of a first draft.  It's already been slightly tweaked in the next PR.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: The key-value accessor methods in TSStorageManager are really useful.  I propose that we move them to a category on `YapDatabaseReadTransaction` and `YapDatabaseReadWriteTransaction`.  

I'm not sure that I support eliminating `TSStorageManager.dbReadConnection` and `TSStorageManager.dbReadWriteConnection` entirely, without reviewing all of the places that they're used. 

I'm making cards "below the line"?

michaelkirk: To clarify, are you proposing leaving this as is and making a card to maybe remove dbReadConnection/dbReadWriteConnection entirely?

charlesmchen: Yup.

* https://trello.com/c/urYDlexK/1264-consider-eliminating-tsstoragemanagerdbreadconnection-and-tsstoragemanagerdbreadwriteconnection
* https://trello.com/c/ndz0SEY7/1263-move-key-value-accessor-methods-from-tsstoragemanager-to-a-category-on-yapdatabasereadtransaction-and-yapdatabasereadwritetransa

PR #2386, opened 2017-07-31
fix avatar flash

michaelkirk:
Fixed the avatar flicker in the inbox and added debug actions to create a bunch of contact threads.

### DIFF
--- a/Signal/src/Models/OWSContactAvatarBuilder.m
+++ b/Signal/src/Models/OWSContactAvatarBuilder.m
@@ -95,7 +95,7 @@ NS_ASSUME_NONNULL_BEGIN
                                                                        textColor:[UIColor whiteColor]
                                                                             font:[UIFont ows_boldFontWithSize:fontSize]
                                                                         diameter:self.diameter] avatarImage];
-    [self.contactsManager.avatarCache setObject:image forKey:self.signalId];
+    [self.contactsManager.avatarCache setObject:image forKey:cacheKey];
     return image;
 }
 
--- a/Signal/src/ViewControllers/InboxTableViewCell.m
+++ b/Signal/src/ViewControllers/InboxTableViewCell.m
@@ -20,7 +20,7 @@ NS_ASSUME_NONNULL_BEGIN
 #define DATE_LABEL_SIZE 13
 #define SWIPE_ARCHIVE_OFFSET -50
 
-const NSUInteger kavatarViewDiameter = 52;
+const NSUInteger kAvatarViewDiameter = 52;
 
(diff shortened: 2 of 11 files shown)


Signal/src/ViewControllers/InboxTableViewCell.m
```diff
@@ -227,16 +227,8 @@ - (void)configureWithThread:(TSThread *)thread
...
-    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        UIImage *avatar =
-            [OWSAvatarBuilder buildImageForThread:thread contactsManager:contactsManager diameter:kavatarViewDiameter];
-        dispatch_async(dispatch_get_main_queue(), ^{
-            if ([_threadId isEqualToString:threadIdCopy]) {
-                self.avatarView.image = avatar;
-            }
-        });
-    });
+    self.avatarView.image =
```
michaelkirk: We initially talked about having this being async in the event of a cache miss, but in my testing, after fixing the caching, even on my iphone5 with 1k threads, this scrolls smoothly without egregious CPU (4% in generating avatars - when scrolling wildly, but it trends to zero when you are scrolling back and forth due to the cache hits).

I decided it wasn't worth the complexity.

Signal/src/Models/OWSContactAvatarBuilder.m
```diff
@@ -95,7 +95,7 @@ - (UIImage *)buildDefaultImage
                                                                        textColor:[UIColor whiteColor]
                                                                             font:[UIFont ows_boldFontWithSize:fontSize]
                                                                         diameter:self.diameter] avatarImage];
-    [self.contactsManager.avatarCache setObject:image forKey:self.signalId];
+    [self.contactsManager.avatarCache setObject:image forKey:cacheKey];
```
michaelkirk: We were effectively never caching generated avatars =(

charlesmchen: =(

Signal/src/ViewControllers/InboxTableViewCell.m
charlesmchen: Okay.  Please make sure that perf testing is part of the test plan.

charlesmchen: Specifically, I'm assuming that perf is a function of input avatar size, so maybe we should modify the "create fake contacts" logic to include a randomly generated image, whose file size/content sizes vary.

PR #2389, opened 2017-08-01
Charlesmchen/yap database connection category

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Storage/YapDatabaseConnection+OWS.m
+++ b/SignalServiceKit/src/Storage/YapDatabaseConnection+OWS.m
@@ -3,15 +3,21 @@
 //
 
 #import "YapDatabaseConnection+OWS.h"
+#import <25519/Curve25519.h>
+#import <AxolotlKit/PreKeyRecord.h>
+#import <AxolotlKit/SignedPrekeyRecord.h>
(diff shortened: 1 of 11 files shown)


charlesmchen:
PTAL @michaelkirk

charlesmchen:
This method only applies the new category in a few places, but hopefully it demonstrates how easy it will be to migrate more - and then we can get rid of the convenience methods on `TSStorageManager`.

SignalServiceKit/src/Storage/YapDatabaseConnection+OWS.m
```diff
@@ -0,0 +1,138 @@
...
+//
+
+#import "YapDatabaseConnection+OWS.h"
+#import <YapDatabase/YapDatabaseTransaction.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@implementation YapDatabaseConnection (OWS)
+
+- (id)objectForKey:(NSString *)key inCollection:(NSString *)collection
```
michaelkirk: nit: `(nullable id)`

charlesmchen: Done. In fact, I went you one better and significantly cleaned up these methods - DRY'ing up code, adding asserts, etc.

PR #2393, opened 2017-08-01
Remove iOS 9-only APIs from layout code.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/UIView+OWS.m
+++ b/Signal/src/UIView+OWS.m
@@ -217,8 +217,13 @@ CGFloat ScaleFromIPhone5(CGFloat iPhone5Value)
 
 - (BOOL)isRTL
 {
-    return ([UIView userInterfaceLayoutDirectionForSemanticContentAttribute:self.semanticContentAttribute]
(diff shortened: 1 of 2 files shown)


Signal/src/UIView+OWS.m
```diff
@@ -228,10 +233,16 @@ - (NSLayoutConstraint *)autoPinLeadingToSuperView
...
-    constraint.active = YES;
-    return constraint;
+    if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(9, 0)) {
+        NSLayoutConstraint *constraint =
+            [self.leadingAnchor constraintEqualToAnchor:self.superview.layoutMarginsGuide.leadingAnchor
+                                               constant:margin];
+        constraint.active = YES;
+        return constraint;
+    } else {
+        margin += (self.isRTL ? self.superview.layoutMargins.right : self.superview.layoutMargins.left);
```
charlesmchen: The key thing is we have to bake-in the layout margins in some cases.

PR #2397, opened 2017-08-02
Charlesmchen/profiles6

charlesmchen:
* Rework how user profiles are updated and persisted. 
* Persist other user’s profiles. 
* Load and cache other user’s profile avatars.
* Add accessor for other users’ profile names.
* DRY up and refine the logic to attach the local profile key to outgoing messages.
* Add “share profile” row to conversation settings.
* Add support for adding groups to the profile whitelist.
* Add profile key to proto schema. 
* Send and receive profile keys. 
* Cache profile manager state.
* Persist profile whitelist and known profile keys.

PTAL @michaelkirk


Signal/Images.xcassets/table_ic_share_profile.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
...
+    {
+      "idiom" : "universal",
+      "filename" : "table_ic_share_profile@3x.png",
+      "scale" : "3x"
+    }
+  ],
+  "info" : {
+    "version" : 1,
+    "author" : "xcode"
+  }
```
charlesmchen: This icon doesn't look great in the context of conversation settings. I intend to change it.

charlesmchen: screen shot 2017-08-02 at 10 55 10 am

Signal/src/ViewControllers/OWSConversationSettingsViewController.m
```diff
@@ -728,6 +783,45 @@ - (void)viewWillDisappear:(BOOL)animated
 
 #pragma mark - Actions
 
+- (void)showShareProfileAlert
+{
+    UIAlertController *alertController =
+        [UIAlertController alertControllerWithTitle:nil message:nil preferredStyle:UIAlertControllerStyleActionSheet];
+
+    UIAlertAction *leaveAction = [UIAlertAction
```
charlesmchen: The copy here is provisional.

SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
```diff
@@ -455,11 +458,52 @@ - (OWSSignalServiceProtosDataMessage *)buildDataMessage
...
+        // For Group threads, we want to include the profile key IFF the
+        // recipient OR the group is in the whitelist.
+        if ([OWSProfilesManager.sharedManager isUserInProfileWhitelist:recipient.recipientId]) {
+            shouldIncludeProfileKey = YES;
+        } else if (self.thread.isGroupThread) {
+            TSGroupThread *groupThread = (TSGroupThread *)self.thread;
+            NSData *groupId = groupThread.groupModel.groupId;
+
+            if ([OWSProfilesManager.sharedManager isGroupIdInProfileWhitelist:groupId]) {
+                [contentBuilder setProfileKey:OWSProfilesManager.sharedManager.localProfileKey];
```
michaelkirk: based on the earlier code, I assume this should be:
```
shouldIncludeProfileKey = YES;
```

And then we only actually set the profile key in one place at the end.

SignalServiceKit/src/Messages/OWSOutgoingNullMessage.m
```diff
@@ -35,7 +36,7 @@ - (instancetype)initWithContactThread:(TSContactThread *)contactThread
 
 #pragma mark - override TSOutgoingMessage
 
-- (NSData *)buildPlainTextData
+- (NSData *)buildPlainTextData:(SignalRecipient *)recipient
 {
     OWSSignalServiceProtosContentBuilder *contentBuilder = [OWSSignalServiceProtosContentBuilder new];
```
michaelkirk: Rather than remembering to add the `[self addLocalProfileKeyIfNecessary:recipient]` to every new content type, maybe we should consider wrapping the Content builder.

SignalServiceKit/src/Messages/OWSSignalServiceProtos.pb.m
```diff
@@ -594,12 +593,20 @@ - (void) setHasNullMessage:(BOOL) _value_ {
   hasNullMessage_ = !!_value_;
 }
 @synthesize nullMessage;
+- (BOOL) hasProfileKey {
```
michaelkirk: Thank you for *not* formatting this file =)

SignalServiceKit/src/Messages/TSMessagesManager.m
```diff
@@ -494,6 +495,14 @@ - (void)handleEnvelope:(OWSSignalServiceProtosEnvelope *)envelope plaintextData:
     if (envelope.hasContent) {
         OWSSignalServiceProtosContent *content = [OWSSignalServiceProtosContent parseFromData:plaintextData];
         DDLogInfo(@"%@ handling content: <Content: %@>", self.tag, [self descriptionForContent:content]);
+        
+        if ([content hasProfileKey]) {
+            NSData *profileKey = [content profileKey];
+            NSString *recipientId = envelope.source;
+            [OWSProfilesManager.sharedManager setProfileKey:profileKey
```
michaelkirk: Does this need to be done on the main thread? (just thinking about our future work to move message processing off the main thread)

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -3,84 +3,100 @@
...
-@property (nonatomic, readonly) NSString *fileName;
-@property (nonatomic, readonly) NSString *avatarUrl;
-@property (nonatomic, readonly) NSString *avatarDigest;
+@property (nonatomic, nullable) NSString *avatarFileName;
+
+// This should reflect when either:
+//
+// * The last successful update was started.
+// * The in-flight update was started.
+@property (nonatomic, nullable) NSDate *lastUpdateDate;
```
michaelkirk: Maybe `updateStartedAt` ? otherwise IMO it reads as though update was completed.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -3,84 +3,100 @@
...
     _avatarUrl = avatarUrl;
     _avatarDigest = avatarDigest;
+    _avatarFileName = avatarFileName;
 
     return self;
 }
 
++ (NSString *)collection
+{
+    return @"UserProfile";
```
michaelkirk: FYI: TSYapDatabaseObject's already use the class name by default. The only time this needs to be used is to override that behavior, 

e.g. We want `TSGroupThread` and `TSContactThread` to share a collection.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -90,22 +106,29 @@ @interface OWSProfilesManager ()
...
 
-@property (atomic, readonly, nullable) NSData *localProfileKey;
+// This property should only be mutated on the main thread,
+//
+// NOTE: Do not access this property directly; use getOrCreateLocalUserProfile instead.
+@property (nonatomic, nullable) UserProfile *localUserProfile;
+// This property should only be mutated on the main thread,
+@property (nonatomic, nullable) UIImage *localCachedAvatarImage;
+
+// These caches are lazy-populated.  The single point truth is the database.
```
michaelkirk: single point **of** truth

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -375,32 +436,164 @@ - (void)updateProfileOnService:(nullable NSString *)localProfileName
...
 
-- (void)loadLocalProfileAsync
+#pragma mark - Profile Whitelist
+
+- (void)addUserToProfileWhitelist:(NSString *)recipientId
+{
+    OWSAssert(recipientId.length > 0);
+
+    [self.dbConnection setObject:@(1) forKey:recipientId inCollection:kOWSProfilesManager_UserWhitelistCollection];
+    self.userProfileWhitelistCache[recipientId] = @(YES);
```
michaelkirk: nit: Would it be simpler to have `userProfileWhitelistCache` be an NSSet of recipientIds, and maybe use `- (BOOL)hasObjectForKey:(NSString *)key inCollection:(nullable NSString *)collection;` rather than worrying about the value?

edit: I guess the current implementation makes it's easier to build a lazily populated cache. Makes sense as is!

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -375,32 +436,164 @@ - (void)updateProfileOnService:(nullable NSString *)localProfileName
...
-            }
+    OWSAssert(recipientId.length > 0);
+
+    NSNumber *_Nullable value = self.userProfileWhitelistCache[recipientId];
+    if (value) {
+        return [value boolValue];
+    }
+
+    value =
+        @(nil != [self.dbConnection objectForKey:recipientId inCollection:kOWSProfilesManager_UserWhitelistCollection]);
```
michaelkirk: readability nit: maybe just `value = @([self.dbConnection hasObjectForKey:recipientId inCollection:...])`

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -375,32 +436,164 @@ - (void)updateProfileOnService:(nullable NSString *)localProfileName
...
 }
 
-- (void)loadLocalProfileAsync
+#pragma mark - Profile Whitelist
+
+- (void)addUserToProfileWhitelist:(NSString *)recipientId
+{
+    OWSAssert(recipientId.length > 0);
+
+    [self.dbConnection setObject:@(1) forKey:recipientId inCollection:kOWSProfilesManager_UserWhitelistCollection];
```
michaelkirk: `setBoolValue:YES`?

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -375,32 +436,164 @@ - (void)updateProfileOnService:(nullable NSString *)localProfileName
...
+{
+    OWSAssert(contactRecipientIds);
+    
+    // TODO: The persisted whitelist could either be:
+    //
+    // * Just users manually added to the whitelist.
+    // * Also include users auto-added by, for example, being in the user's
+    //   contacts or when the user initiates a 1:1 conversation with them, etc.
+    for (NSString *recipientId in contactRecipientIds) {
+        [self addUserToProfileWhitelist:recipientId];
```
michaelkirk: As written, this will create a one-write transaction for every one of my contacts (on the main thread), right? 

A couple of perf thoughts:

1. We should be able to do this in a single transaction
2. We should only write to the db when the value differs from what's in our cache (maybe?)

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -375,32 +436,164 @@ - (void)updateProfileOnService:(nullable NSString *)localProfileName
...
+    }
+
+    NSData *_Nullable existingProfileKey = [self profileKeyForRecipientId:recipientId];
+    if (existingProfileKey &&
+        [existingProfileKey isEqual:profileKey]) {
+        // Ignore redundant update.
+        return;
+    }
+
+    [self.dbConnection setObject:profileKey
```
michaelkirk: `setData:`?

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -375,32 +436,164 @@ - (void)updateProfileOnService:(nullable NSString *)localProfileName
...
+{
+    OWSAssert(recipientId.length > 0);
+
+    NSData *_Nullable profileKey = self.otherUsersProfileKeyCache[recipientId];
+    if (profileKey.length > 0) {
+        return profileKey;
+    }
+
+    profileKey =
+        [self.dbConnection objectForKey:recipientId inCollection:kOWSProfilesManager_OtherUsersProfileKeysCollection];
```
michaelkirk: `dataForKey:`?

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -375,32 +436,164 @@ - (void)updateProfileOnService:(nullable NSString *)localProfileName
...
-            self.localProfileAvatarMetadata = localProfileAvatarMetadata;
+    return image;
+}
 
-            [[NSNotificationCenter defaultCenter] postNotificationName:kNSNotificationName_LocalProfileDidChange
-                                                                object:nil
-                                                              userInfo:nil];
-        });
-    });
+- (void)fetchProfileForRecipientId:(NSString *)recipientId
```
michaelkirk: I'm not sure why we need this method. 

My thought was the ProfileFetcherJob is run whenever a thread becomes visible (It's already throttled).
Whenever it fetches it tells the ProfileManager the latest details for a profile.
ProfileManager can expire cache immediately if the profile has changed.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -375,32 +436,164 @@ - (void)updateProfileOnService:(nullable NSString *)localProfileName
...
+    for (NSString *recipientId in contactRecipientIds) {
+        [self addUserToProfileWhitelist:recipientId];
+    }
+}
+
+#pragma mark - Other User's Profiles
+
+- (void)setProfileKey:(NSData *)profileKey forRecipientId:(NSString *)recipientId
+{
+    OWSAssert(profileKey.length == kProfileKeyLength);
```
michaelkirk: assert on main thread?

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -375,32 +436,164 @@ - (void)updateProfileOnService:(nullable NSString *)localProfileName
...
+
+    value =
+        @(nil != [self.dbConnection objectForKey:groupIdKey inCollection:kOWSProfilesManager_GroupWhitelistCollection]);
+    self.groupProfileWhitelistCache[groupIdKey] = value;
+    return [value boolValue];
+}
+
+- (void)setContactRecipientIds:(NSArray<NSString *> *)contactRecipientIds
+{
+    OWSAssert(contactRecipientIds);
```
michaelkirk: assert on main thread?

SignalServiceKit/src/Profiles/OWSProfilesManager.m
```diff
@@ -375,32 +436,164 @@ - (void)updateProfileOnService:(nullable NSString *)localProfileName
     });
 }
 
-- (void)loadLocalProfileAsync
+#pragma mark - Profile Whitelist
+
+- (void)addUserToProfileWhitelist:(NSString *)recipientId
+{
+    OWSAssert(recipientId.length > 0);
```
michaelkirk: assert on main thread?

charlesmchen:
> Except our profile fetching strategy is still up in the air, but we can address that as we go.

I've made some progress on this in my next PR which I'll open soon.

Signal/src/ViewControllers/OWSConversationSettingsViewController.m
```diff
@@ -292,10 +294,50 @@ - (void)updateTableContents
                                  }]];
     }
 
+    BOOL isOnProfileWhitelist = NO;
```
charlesmchen: Done.

SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
charlesmchen: Fixed.

SignalServiceKit/src/Messages/OWSOutgoingNullMessage.m
charlesmchen: Hopefully we'll add content types pretty rarely.  I'm not sure it's worth the complexity, since we could just as easily forget to use the wrapped content builder.

SignalServiceKit/src/Messages/TSMessagesManager.m
charlesmchen: Yeah, I've been thinking that and over the course of the PRs I've been consolidating around "profile state can be read on any thread but can only be written on the main thread".  But I've been tempted to move all writes to a serial queue.  Reads could still happen synchronously on any thread.  I haven't done it yet because I didn't want to prematurely optimize.  I'll add a card so we follow up.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: Sorry, I clarified this in the next PR.  The problem is that we don't initiate all updates - other components of the system might also launch a profile fetch job that will end up updating profile state.  The comment is improved in the next PR; let's discuss there.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: Oh thanks! TIL

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: Done.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: Done, great idea.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: Nice.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: +1

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: You're right, our connection category is a real opportunity to make this code more readable.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: Done.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: I've done a bunch of work around thread safety that's in the next PR.  Let's continue there.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: Why bother?

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: Good points. Done.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: Done.

SignalServiceKit/src/Profiles/OWSProfilesManager.m
charlesmchen: Yup, let's revisit this in the context of the next PR which fleshes this out.  Essentially, I think there's now going to be many more triggers for refreshing a profile.

PR #2402, opened 2017-08-03
Terminate call if call view presentation is delayed.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -144,10 +144,10 @@ protocol CallServiceObserver: class {
                 if let oldValue = oldValue {
                     DeviceSleepManager.sharedInstance.removeBlock(blockObject: oldValue)
                 }
-                stopAnyActiveCallTimer()
+                stopAnyCallTimer()
                 if let call = call {
                     DeviceSleepManager.sharedInstance.addBlock(blockObject: call)
-                    self.startActiveCallTimer()
+                    self.startCallTimer()
                 }
             }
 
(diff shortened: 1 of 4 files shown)


Signal/src/call/CallService.swift
```diff
@@ -144,8 +144,10 @@ protocol CallServiceObserver: class {
                 if let oldValue = oldValue {
                     DeviceSleepManager.sharedInstance.removeBlock(blockObject: oldValue)
                 }
+                stopAnyActiveCallTimer()
                 if let call = call {
                     DeviceSleepManager.sharedInstance.addBlock(blockObject: call)
+                    self.startActiveCallTimer()
```
charlesmchen: I feel like this slightly more strongly ensure the invariant "there is a timer IFF there is a call."

Signal/src/call/CallService.swift
```diff
@@ -1574,9 +1573,8 @@ protocol CallServiceObserver: class {
     // MARK: CallViewController Timer
 
     var activeCallTimer: Timer?
```
charlesmchen: We could rename this to `callTimer`.

Signal/src/call/CallService.swift
michaelkirk: Sounds reasonable.  We might want to rename some of the associated methods to give a hint as to their new roles.

maybe: 
- `startCallTimer` 
- `stopAnyCallTimer`
- `ensureUndisplayedCallTerminated`

charlesmchen: Renamed.

PR #2404, opened 2017-08-03
Charlesmchen/profiles7

charlesmchen:
PTAL @michaelkirk


charlesmchen:
Just realized that I need to push another commit to reflect discussion on last PR.

Signal/src/Profiles/ProfileFetcherJob.swift
```diff
@@ -105,7 +114,10 @@ class ProfileFetcherJob: NSObject {
     private func updateProfile(signalServiceProfile: SignalServiceProfile) {
         verifyIdentityUpToDateAsync(recipientId: signalServiceProfile.recipientId, latestIdentityKey: signalServiceProfile.identityKey)
 
-        // Eventually we'll want to do more things with new SignalServiceProfile fields here.
+        OWSProfileManager.shared().updateProfile(forRecipientId : signalServiceProfile.recipientId,
```
michaelkirk: Take it or leave it:

This works, but it's the circular dependency is a little goofy wherein the OWSProfileManager calls the Fetcher Job which then calls back to the OWSProfileManager. Maybe this method should take a block or return a Promise<SignalServiceProfile>

...and then we move the verifyIdentityUpToDate logic into ProfileManager.

All that seems compatible with my earlier mentioned plan of having only *one* block of throttling logic.

This works though.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -623,11 +637,154 @@ - (void)fetchProfileForRecipientId:(NSString *)recipientId
...
+    [self saveUserProfile:userProfile];
+
+    [ProfileFetcherJob runWithRecipientId:recipientId
+                           networkManager:self.networkManager
+                         ignoreThrottling:ignoreThrottling];
+}
+
+- (void)updateProfileForRecipientId:(NSString *)recipientId
+               profileNameEncrypted:(NSData *_Nullable)profileNameEncrypted
+                 avatarUrlEncrypted:(NSData *_Nullable)avatarUrlEncrypted
```
michaelkirk: As discussed I think the URL/digest will be plain text, but the blob will be encrypted. Though we can address that as part of the decryption PR.

Signal/src/Profiles/ProfileFetcherJob.swift
```diff
@@ -126,32 +138,65 @@ struct SignalServiceProfile {
...
 
+        var profileNameEncrypted: Data? = nil
+        if let profileNameString = responseDict["name"] as? String {
+            guard let data = Data(base64Encoded: profileNameString) else {
+                throw ValidationError.invalidProfileName(description: "\(TAG) unable to parse profile name: \(profileNameString)")
+            }
+            profileNameEncrypted = data
+        }
+
+        var avatarUrlEncrypted: Data? = nil
```
michaelkirk: ditto: I think this is plain text string.

Signal/src/Profiles/ProfileFetcherJob.swift
```diff
@@ -126,32 +138,65 @@ struct SignalServiceProfile {
...
+        if let avatarUrlString = responseDict["avatar"] as? String {
+            guard let data = Data(base64Encoded: avatarUrlString) else {
+                throw ValidationError.invalidAvatarUrl(description: "\(TAG) unable to parse avatar URL: \(avatarUrlString)")
+            }
+            avatarUrlEncrypted = data
+        }
+
+        var avatarDigestEncrypted: Data? = nil
+        if let avatarDigestString = responseDict["avatarDigest"] as? String {
```
michaelkirk: ditto: I think this is plain text string.

SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
```diff
@@ -478,15 +482,15 @@ - (void)addLocalProfileKeyIfNecessary:(OWSSignalServiceProtosContentBuilder *)co
...
             shouldIncludeProfileKey = YES;
-        } else if ([OWSProfilesManager.sharedManager isThreadInProfileWhitelist:self.thread]) {
+        } else if ([profileManager isThreadInProfileWhitelist:self.thread]) {
             shouldIncludeProfileKey = YES;
         }
     }
 
     if (shouldIncludeProfileKey) {
-        [contentBuilder setProfileKey:OWSProfilesManager.sharedManager.localProfileKey];
+        [contentBuilder setProfileKey:localProfileKey];
```
michaelkirk: per earlier discussion this will be in the DataMessage, but we can do that in a followup.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -82,31 +82,24 @@ - (NSUInteger)hash
 
 #pragma mark -
 
+NSString *const kNSNotificationName_LocalProfileUniqueId = @"kNSNotificationName_LocalProfileUniqueId";
```
charlesmchen: You mean that prefix?  I've used that (many places) because they are posted with the method `postNotificationName:`.  If you want we can just use the prefix `kNSNotification_`, NBD.

charlesmchen: Oh duh, now I see what you're talking about.  Fixed.

michaelkirk: No, I'm not worried about the prefix convention, which I think is good.

But doesn't look like we're using this as a notification name. Isn't this constant used only as the uniqueId of a DB object?

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -156,32 +150,30 @@ - (instancetype)initWithStorageManager:(TSStorageManager *)storageManager
...
-    _localProfileKey = [self.dbConnection objectForKey:kOWSProfilesManager_LocalProfileSecretKey
-                                            inCollection:kOWSProfilesManager_Collection];
-    if (!_localProfileKey) {
-        // Generate
-        _localProfileKey = [OWSProfilesManager generateLocalProfileKey];
-        // Persist
-        [self.dbConnection setObject:_localProfileKey
-                                forKey:kOWSProfilesManager_LocalProfileSecretKey
-                          inCollection:kOWSProfilesManager_Collection];
+    self.localUserProfile = [self getOrCreateUserProfileForRecipientId:kNSNotificationName_LocalProfileUniqueId];
```
charlesmchen: Wait... this should supercede the old way that you're referring to, where the key was the actual local phone number.  I wanted to make sure that we could, for example, safely generate this key _before_ the user registered, in the interest of decoupling.

On the other hand, I haven't special-cased "self" (at least not yet) so if we have a conversation with ourself, we'll load our profile.  It wouldn't be too hard to special-case if you want to do that.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -156,32 +150,30 @@ - (instancetype)initWithStorageManager:(TSStorageManager *)storageManager
...
+    OWSAssert(self.localUserProfile);
+    if (!self.localUserProfile.profileKey) {
```
charlesmchen: I had it that way originally (for those reasons), but moved to this approach because I wanted to guarantee that this property is a) populated before it is accessed b) always populated on the main thread c) can be accessed synchronously from any thread.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -199,51 +191,48 @@ - (void)observeNotifications
...
 
-- (void)appLaunchDidBegin
-{
-    // Do nothing; we only want to make sure this singleton is created on startup.
-}
-
 #pragma mark - User Profile Accessor
 
+// This method can be safely called from any thread.
 - (UserProfile *)getOrCreateUserProfileForRecipientId:(NSString *)recipientId
```
charlesmchen: Done.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -199,51 +191,48 @@ - (void)observeNotifications
...
+    }
+
     OWSAssert(instance);
 
     return instance;
 }
 
-- (nullable UserProfile *)getOrCreateLocalUserProfile
+// All writes to user profiles should occur on the main thread.
+- (void)saveUserProfile:(UserProfile *)userProfile
```
charlesmchen: Yes, my concern was that mutations have two steps: a) modifying the properties b) writing to database. I didn't want to allow one thread to be modifying a property while another thread was saving the profile, especially since some of our profile properties (like avatar url + digest, and lastUpdateDate + all other properties) should be modifies in lockstep.  

As discussed, I think we can allow writes from any thread if we actually do the writing & saving on a serial queue.  I think it'll work, I just haven't gotten there and haven't committed to doing it yet since I haven't seen a key reason to.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -199,51 +191,48 @@ - (void)observeNotifications
...
+                                                          userInfo:nil];
+    } else {
+        [[NSNotificationCenter defaultCenter] postNotificationName:kNSNotificationName_OtherUsersProfileDidChange
+                                                            object:nil
+                                                          userInfo:nil];
     }
-    return self.localUserProfile;
 }
 
 #pragma mark - Local Profile Key
```
charlesmchen: Done.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -257,14 +246,21 @@ + (NSData *)generateLocalProfileKey
...
     OWSAssert([NSThread isMainThread]);
 
-    return [self getOrCreateLocalUserProfile].profileName;
+    return self.localUserProfile.profileName;
 }
 
-- (UIImage *)localProfileAvatarImage
+- (nullable UIImage *)localProfileAvatarImage
 {
     OWSAssert([NSThread isMainThread]);
```
charlesmchen: You're right. I'll make a card and try to do it today.

charlesmchen: I think there's going to be lots of small bugs like this, since I haven't been able to test yet.  Moxie suggested testing by modifying the encrypt/decrypt methods to return inputs for now.  I'm going to make that change today - hopefully it'll help flush out the worst issues.

Signal/src/Profiles/OWSProfileManager.m
michaelkirk: Sounds good.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -623,11 +637,154 @@ - (void)fetchProfileForRecipientId:(NSString *)recipientId
...
-    [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
-        [userProfile saveWithTransaction:transaction];
-    }];
 
-    // TODO: Actually update the profile.
+    [self saveUserProfile:userProfile];
+
+    [ProfileFetcherJob runWithRecipientId:recipientId
+                           networkManager:self.networkManager
+                         ignoreThrottling:ignoreThrottling];
```
charlesmchen: > It's a little weird to have one timeout clobber another.

First of all, this `ignoreThrottling` might be excessive.  The intention was to ensure that we refetched whenever a profile key changed, but since profile key changes (i.e. re-registration) should be rare, this shouldn't be a big deal.  It'll help with testing and be a slightly more crisp experience when a contact re-registers.

> An interesting consequence of this thought exercise is I realize we'll now that scrolling through the inbox triggers profile fetches, and triggers any learned SN changes.

That's absolutely right - there was now _way more_ triggers for profile fetches.  This will work in the "new conversation" views too, which is nice.  

>  It's a little weird to have one timeout clobber another.

I'm not sure what you mean - it's not two separate timeouts, it's an event (profile key change) that triggers an immediate, unthrottled fetch.

Signal/src/Profiles/OWSProfileManager.m
charlesmchen: Yup, I just learned that in sprint planning yesterday and haven't updated yet.

Signal/src/Profiles/ProfileFetcherJob.swift
charlesmchen: Yeah, I thought about it but:

a) Didn't yet because this is already working.
b) This is swift and the manager is obj-c (since it was born in SSK).
c) Throttling is all done in one place - but we sometimes (may) want to make an exception to it.

If you want to migrate it, +1.

Signal/src/Profiles/ProfileFetcherJob.swift
charlesmchen: Yup, I have a card.

SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
charlesmchen: I've done that work in the PR I'm about to open.

Signal/src/Profiles/OWSProfileManager.m
michaelkirk: Sorry my comment wasn't clear. I think what you've proposed here is reasonable as is.

Signal/src/Profiles/OWSProfileManager.m
michaelkirk: What I mean is ProfileFetcherJob has a throttling mechanism.
This method has a separate throttling mechanism.

This methods throttling mechanism disregards the ProfileFetcherJob's throttling mechanism.

I'm hopeful that there's an opportunity to consolidate the two throttling mechanisms and make the code easier to reason about.

charlesmchen: Yep yep yep. I'll make a card.

PR #2416, opened 2017-08-07
Use profile name/image when available.

michaelkirk:
In the process, refactored the ContactAvatarBuilder to clarify intent of the two
methods. One is only used for non-signal contacts in the Invite flow.
The other should be used for any signal contacts so we have a single
consistent way of generating the avatar initials.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/Models/OWSContactAvatarBuilder.m
+++ b/Signal/src/Models/OWSContactAvatarBuilder.m
@@ -24,10 +24,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 @implementation OWSContactAvatarBuilder
 
+#pragma mark - Initializers
+
 - (instancetype)initWithContactId:(NSString *)contactId
                              name:(NSString *)name
-                  contactsManager:(OWSContactsManager *)contactsManager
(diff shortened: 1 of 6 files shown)


Signal/src/Models/OWSContactAvatarBuilder.m
```diff
@@ -36,19 +38,30 @@ - (instancetype)initWithContactId:(NSString *)contactId
...
+    NSString *name = [contactsManager displayNameForPhoneIdentifier:signalId];
+    return [self initWithContactId:signalId name:name diameter:diameter contactsManager:contactsManager];
 }
 
+- (instancetype)initWithNonSignalName:(NSString *)nonSignalName
+                            colorSeed:(NSString *)colorSeed
+                             diameter:(NSUInteger)diameter
+                      contactsManager:(OWSContactsManager *)contactsManager
+{
+    return [self initWithContactId:colorSeed name:nonSignalName diameter:diameter contactsManager:contactsManager];
```
charlesmchen: s/initWithContactId/initWithColorSeed/

PR #2419, opened 2017-08-09
upload profile avatar

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/Profiles/OWSProfileManager.m
+++ b/Signal/src/Profiles/OWSProfileManager.m
@@ -336,12 +336,15 @@ static const NSInteger kProfileKeyLength = 16;
     UserProfile *userProfile = self.localUserProfile;
     OWSAssert(userProfile);
 
-    // If we have a new avatar image, we must first:
-    //
-    // * Encode it to JPEG.
-    // * Write it to disk.
-    // * Upload it to service.
     if (avatarImage) {
+
+        // If we have a new avatar image, we must first:
+        //
+        // * Encode it to JPEG.
+        // * Write it to disk.
+        // * Encrypt it
+        // * Upload it to asset service
+        // * Send asset service info to Signal Service
         if (self.localCachedAvatarImage == avatarImage) {
             OWSAssert(userProfile.avatarUrl.length > 0);
             OWSAssert(userProfile.avatarDigest.length > 0);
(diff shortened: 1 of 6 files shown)


Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -400,21 +403,33 @@ - (void)writeAvatarToDisk:(UIImage *)avatar
...
-- (void)uploadAvatarToService:(NSData *)data
-                     fileName:(NSString *)fileName
+- (NSData *)encryptedAvatarData:(NSData *)plainTextData outDigest:(NSData **)outDigest
+{
+    DDLogError(@"TODO: Profile encryption scheme not yet settled.");
+
+    // server accepts up to 14 base64 chars for digest
+    // 14 <= 4 * ceil(n/3)
+    NSUInteger kAvatarDigestByteLength = 9;
+    *outDigest = [Cryptography computeSHA256Digest:plainTextData truncatedToBytes:kAvatarDigestByteLength];
```
charlesmchen: We don't need avatar digests any more, right?

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -400,21 +403,33 @@ - (void)writeAvatarToDisk:(UIImage *)avatar
     });
 }
 
-// TODO: The exact API & encryption scheme for avatars is not yet settled.
-- (void)uploadAvatarToService:(NSData *)data
-                     fileName:(NSString *)fileName
+- (NSData *)encryptedAvatarData:(NSData *)plainTextData outDigest:(NSData **)outDigest
```
charlesmchen: I know this code is going away, but I'll mention it anyway.

First, I'm not a big fan of the ** "writable param to return secondary result" pattern.

Secondly, when using that pattern, I like to make sure that the first thing that the method/function does is nil that param so if we ever exit early, that writable param has a predictable value.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -487,87 +493,66 @@ - (void)uploadAvatarToService:(NSData *)data
...
+                profileHttpManager.responseSerializer.acceptableContentTypes = nil;
+
+                [profileHttpManager POST:@""
+                    parameters:nil
+                    constructingBodyWithBlock:^(id<AFMultipartFormData> _Nonnull formData) {
+                        NSData * (^formDataForString)(NSString *formString) = ^(NSString *formString) {
+                            return [formString dataUsingEncoding:NSUTF8StringEncoding];
+                        };
+
+                        // We have to build up the form manually vs. simply passing in a paramaters dict
```
charlesmchen: s/paramaters/parameters/

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -487,87 +493,66 @@ - (void)uploadAvatarToService:(NSData *)data
...
-                        if (uploadError) {
-                            DDLogError(@"%@ Avatar upload failed: %@", self.tag, uploadError);
+                    success:^(NSURLSessionDataTask *_Nonnull uploadTask, id _Nullable responseObject) {
+                        OWSAssert([uploadTask.response isKindOfClass:[NSHTTPURLResponse class]]);
+                        NSHTTPURLResponse *response = (NSHTTPURLResponse *)uploadTask.response;
+
+                        // We could also construct this URL locally from manager.baseUrl + formKey
+                        // but the approach of getting it from the remote provider seems a more
+                        // robust way to ensure we've actually created the resource where we
+                        // think we have.
```
charlesmchen: Should we assert that the avatarURL has the expected manager.baseUrl prefix?

charlesmchen: Oh, that won't work in the censorship circumvention case, right?

SignalServiceKit/src/Network/OWSSignalService.m
```diff
@@ -204,6 +206,25 @@ - (AFHTTPSessionManager *)reflectorHTTPSessionManager
...
 }
 
+#pragma mark - Profile Uploading
+
+- (AFHTTPSessionManager *)profileUploadingSessionManagerWithHostname:(NSString *)hostname
+{
+    OWSAssert(hostname.length > 0);
+    if (self.isCensorshipCircumventionActive) {
+        DDLogInfo(@"%@ Profile uploading may not work when under censorship.", self.tag);
+    }
```
charlesmchen: `owsfail` for now?

PR #2421, opened 2017-08-10
[JSQMVC] Fix scrolling crashes

michaelkirk:
- First occurs when the "Load More Messages" is shown, while scrolling and
  receiving new messages
- Second occurs when scrolling while popping/dismissing the keyboard
  (Not sure if receiving messages was a precondition for this crash)

PTAL @charlesmchen


michaelkirk:
In 93d2baa0

PR #2425, opened 2017-08-11
Don't consider registration complete until user registers their push notification tokens.

michaelkirk:
Previously we let registration complete and then immediately uploaded push tokens. But in the event of a failure-to-upload for whatever reason, this would mean the user can send, but not receive, messages.

PTAL @charlesmchen


michaelkirk:
In 92ecf0cdc14f57e5ef6dcff18277166d69b7ed8c

PR #2427, opened 2017-08-14
new profile service scheme

michaelkirk:
Crypto Scheme:

- Name (un)padding
- WIP AES-GCM (functioning, but need to verify against android
  implementation, and tag functionality)

Changes to avatar API:

- hard code avatar domain (cdn.signal.org)
- avatar form hands out new avatar key, invalidating old avatar
- preliminary aes-gcm integration

Also:

- New type to represent AES128 keys, rather than passing around opaque
  data blobs everywhere, we can use the compiler to help us make sure
  we're passing compliant keying material.

- Started using factory pattern for API requests. This is intended to be
  a lighter weight way to implement new API requests, rather than the
  current 1-method class ceremony.

PTAL @charlesmchen


Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -24,12 +25,14 @@ @interface UserProfile : TSYapDatabaseObject
...
 @property (atomic, readonly) NSString *recipientId;
-@property (atomic, nullable) NSData *profileKey;
+@property (atomic, nullable) OWSAES128Key *profileKey;
 
 // These properties may be accessed only from the main thread.
 @property (nonatomic, nullable) NSString *profileName;
+
+// TODO This isn't really a URL, since it doesn't contain the host.
+//      rename to "avatarPath" or "avatarKey"
 @property (nonatomic, nullable) NSString *avatarUrl;
```
charlesmchen: avatarUrlPath

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -403,37 +400,25 @@ - (void)writeAvatarToDisk:(UIImage *)avatar
...
+        NSData *encryptedAvatarData = [self encryptProfileData:avatarData];
         OWSAssert(encryptedAvatarData.length > 0);
-        OWSAssert(encryptedAvatarDigest.length > 0);
 
         // See: https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-UsingHTTPPOST.html
         TSProfileAvatarUploadFormRequest *formRequest = [TSProfileAvatarUploadFormRequest new];
 
+        // TODO: Since this form request causes the server to reset my avatar URL, if the update fails
+        // at some point from here on out, we want the user to understand they probably no longer have
+        // a profile avatar on the server.
```
charlesmchen: Please add another TODO to clear local profile state for local user once upload begins?

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -987,76 +923,80 @@ - (BOOL)isNullableStringEqual:(NSString *_Nullable)left toString:(NSString *_Nul
...
+    const char *bytes = decryptedData.bytes;
 
-    if (value) {
-        NSData *_Nullable data = [value dataUsingEncoding:NSUTF8StringEncoding];
-        if (data) {
-            NSData *_Nullable encryptedData = [self encryptProfileData:data profileKey:profileKey];
-            return encryptedData;
+    // Work through the bytes until we encounter our first
+    // padding byte (our padding scheme is NULL bytes)
+    for (NSUInteger i = 0; i < decryptedData.length; i++) {
```
charlesmchen:
```
for (; unpaddedLength < decryptedData.length; unpaddedLength ++) {
 if (bytes[unpaddedLength] == 0) {
```

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -987,76 +923,80 @@ - (BOOL)isNullableStringEqual:(NSString *_Nullable)left toString:(NSString *_Nul
...
+        return nil;
+    }
 
-- (NSData *_Nullable)encryptProfileString:(NSString *_Nullable)value
-{
-    return [OWSProfileManager encryptProfileString:value profileKey:self.localProfileKey];
+    NSUInteger paddingByteCount = kOWSProfileManager_NameDataLength - nameData.length;
+
+    NSMutableData *paddedNameData = [nameData mutableCopy];
+    [paddedNameData increaseLengthBy:paddingByteCount];
```
charlesmchen: I would add a comment that `increaseLengthBy:` sets all of the added bytes to zero.

SignalServiceKit/src/Network/API/Requests/TSRequest.h
```diff
@@ -14,4 +10,9 @@
 
 - (void)makeAuthenticatedRequest;
 
+#pragma mark - Factory methods
+
+// move to builder class/header
++ (instancetype)setProfileNameRequestWithProfileName:(NSString *)encryptedName;
```
charlesmchen: I don't see this method implemented anywhere?

Signal/src/Profiles/OWSProfileManager.m
michaelkirk: done.

Signal/src/Profiles/OWSProfileManager.m
michaelkirk: I'm not sure I understand. 

My position was that since it's the `GET v1/profile/form` that effectively removes the avatar, maybe we should do it here (which is strictly earlier than the upload that you're suggesting).

Signal/src/Profiles/OWSProfileManager.m
michaelkirk: heh, I actually had this, but thought you'd find it too clever.

Signal/src/Profiles/OWSProfileManager.m
michaelkirk: done.

SignalServiceKit/src/Network/API/Requests/TSRequest.h
michaelkirk: whoops - good catch. I opted for a separate Builder class instead. I'll remove this declaration.

Signal/src/Profiles/OWSProfileManager.m
charlesmchen: Sorry, we're agreeing.

PR #2430, opened 2017-08-14
Sketch out profile header in app settings view.

charlesmchen:
### This PR is ready for review.

screen shot 2017-08-14 at 4 18 42 pm
screen shot 2017-08-14 at 4 18 24 pm
screen shot 2017-08-14 at 4 18 17 pm
screen shot 2017-08-14 at 4 18 05 pm
screen shot 2017-08-14 at 4 17 58 pm

PTAL @michaelkirk


charlesmchen:
PTAL @michaelkirk

charlesmchen:
I used a camera icon & our material blue color to prompt the user to set their profile avatar and name.  If we like this, I'll probably do something similar in our group avatar views.

Signal/src/ViewControllers/AppSettingsViewController.m
```diff
@@ -238,6 +211,90 @@ - (void)updateTableContents
...
+        UIImage *cameraImage = [UIImage imageNamed:@"settings-avatar-camera"];
+        cameraImage = [cameraImage imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
+        UIImageView *cameraImageView = [[UIImageView alloc] initWithImage:cameraImage];
+        cameraImageView.tintColor = [UIColor ows_materialBlueColor];
+        [cell.contentView addSubview:cameraImageView];
+        [cameraImageView autoPinTrailingToView:avatarView];
+        [cameraImageView autoPinEdge:ALEdgeBottom toEdge:ALEdgeBottom ofView:avatarView];
+    }
+
+    UIView *threadNameView = [UIView containerView];
```
michaelkirk: several variables in here called "thread" when the should probably be "profile"

Signal/src/ViewControllers/AppSettingsViewController.m
```diff
@@ -238,6 +211,90 @@ - (void)updateTableContents
...
+    }
+    threadTitleLabel.lineBreakMode = NSLineBreakByTruncatingTail;
+    [threadNameView addSubview:threadTitleLabel];
+    [threadTitleLabel autoPinEdgeToSuperviewEdge:ALEdgeTop];
+    [threadTitleLabel autoPinWidthToSuperview];
+
+    __block UIView *lastTitleView = threadTitleLabel;
+
+    const CGFloat kSubtitlePointSize = 12.f;
+    void (^addSubtitle)(NSAttributedString *) = ^(NSAttributedString *subtitle) {
```
michaelkirk: I'm unclear on the purpose of this block. Can you comment about why it's a block vs just inlining it?

Signal/src/util/UIFont+OWS.m
```diff
@@ -87,4 +87,17 @@ + (UIFont *)ows_dynamicTypeTitle2Font {
...
++ (UIFont *)ows_dynamicTypeTitle3Font
+{
+    if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(9, 0)) {
+        return [UIFont preferredFontForTextStyle:UIFontTextStyleTitle3];
+    } else {
+        // Dynamic title font for ios8 defaults to bold 12.0 pt, whereas ios9+ it's 22.0pt regular weight.
+        // Here we chose to break dynamic font, in order to have uniform style across versions.
+        // It's already huge, so it's unlikely to present a usability issue.
+        // Handy font translations: http://swiftiostutorials.com/comparison-of-system-fonts-on-ios-8-and-ios-9/
+        return [self ows_regularFontWithSize:22.0];
```
michaelkirk: Is this intentionally the same size as the Title2Font?

charlesmchen:
> LGTM, but I think we can get rid of the "Profile" cell in the settings table cells, right?

I'm not sure.  First of all, I think we want to include a bunch of UI elements to explain the avatar & name - at the very least linking to the support article about their visibility, etc.  And I think we'd also benefit from other in-app copy explaining exactly what they are.

Signal/src/ViewControllers/AppSettingsViewController.m
charlesmchen: We are now showing a "avatar + name + optional subtitle" header in multiple places: profile view, conversation settings, new group, update group.  This code is cribbed from one of those other locations.  I'd like to eventually pull out a widget that we can hopefully use in all of those locations, so my intent was to try to keep the code as similar as possible for now. Reasonable or should I rename?

Signal/src/ViewControllers/AppSettingsViewController.m
charlesmchen: See above.

Signal/src/util/UIFont+OWS.m
charlesmchen: Just pushed another commit.

Signal/src/ViewControllers/AppSettingsViewController.m
michaelkirk: I'd prefer to not have to know the origin of this copy and pasted code to be able to make sense of it. I'd say either give them reasonable enough names so they can stand on their own or extract something re-usable before you merge it in, in case the "I'll clean this up later" doesn't get priority for a while.

michaelkirk:
A second-look design thought - can we add a disclosure indicator on the right side to always make it clear that this is actionable?

Signal/src/ViewControllers/AppSettingsViewController.m
charlesmchen: Done.

charlesmchen:
> A second-look design thought - can we add a disclosure indicator on the right side to always make it clear that this is actionable?

Great idea; now we can get rid of the "profile" row.

PR #2431, opened 2017-08-15
resize profile avatar

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/Profiles/OWSProfileManager.h
+++ b/Signal/src/Profiles/OWSProfileManager.h
@@ -9,6 +9,8 @@ NS_ASSUME_NONNULL_BEGIN
 extern NSString *const kNSNotificationName_LocalProfileDidChange;
 extern NSString *const kNSNotificationName_OtherUsersProfileDidChange;
 
+extern const NSUInteger kOWSProfileManager_MaxAvatarWidth;
+
 @class TSThread;
 @class OWSAES128Key;
 
--- a/Signal/src/Profiles/OWSProfileManager.h
+++ b/Signal/src/Profiles/OWSProfileManager.h
@@ -9,7 +9,7 @@ NS_ASSUME_NONNULL_BEGIN
 extern NSString *const kNSNotificationName_LocalProfileDidChange;
 extern NSString *const kNSNotificationName_OtherUsersProfileDidChange;
 
-extern const NSUInteger kOWSProfileManager_MaxAvatarWidth;
+extern const NSUInteger kOWSProfileManager_MaxAvatarDiameter;
 
 @class TSThread;
 @class OWSAES128Key;
--- a/Signal/src/Profiles/OWSProfileManager.m
+++ b/Signal/src/Profiles/OWSProfileManager.m
@@ -413,17 +413,10 @@ const NSUInteger kOWSProfileManager_MaxAvatarDiameter = 640;
     }
 
     NSData *_Nullable data;
-    for (NSUInteger attempts = 0; attempts < 5; attempts++) {
-        CGFloat quality = (CGFloat)0.95 - attempts * (CGFloat)0.1;
-        data = UIImageJPEGRepresentation(image, quality);
-        if (data.length <= kMaxAvatarBytes) {
-            return data;
-        } else {
-            // This for-loop is really just paranoia. Our avatar dimensions are so small that
-            // it's incredibly unlikely we wouldn't be able to fit our profile photo with even
-            // our highest quality.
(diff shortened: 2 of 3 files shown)


Signal/src/Profiles/OWSProfileManager.h
```diff
@@ -9,6 +9,8 @@ NS_ASSUME_NONNULL_BEGIN
 extern NSString *const kNSNotificationName_LocalProfileDidChange;
 extern NSString *const kNSNotificationName_OtherUsersProfileDidChange;
 
+extern const NSUInteger kOWSProfileManager_MaxAvatarWidth;
```
charlesmchen: I think we should use `Size` (i.e. height or width), not `Width`.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -398,6 +399,36 @@ - (void)writeAvatarToDisk:(UIImage *)avatar
     });
 }
 
+- (NSData *)processedImageDataForRawAvatar:(UIImage *)image
+{
+    NSUInteger kMaxAvatarBytes = 5 * 1000 * 1000;
+
+    if (image.size.width != kOWSProfileManager_MaxAvatarWidth
```
charlesmchen: I agree we should downsample oversize avatars, but do we want to upsample small avatars?  It seems like it can only cause harm.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -94,6 +94,7 @@ - (NSUInteger)hash
 /// The max bytes for a user's profile name, encoded in UTF8.
 /// Before encrypting and submitting we NULL pad the name data to this length.
 static const NSUInteger kOWSProfileManager_NameDataLength = 26;
+const NSUInteger kOWSProfileManager_MaxAvatarWidth = 640;
```
charlesmchen: Did android and desktop teams sign off on 640/5mb?

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -398,6 +399,36 @@ - (void)writeAvatarToDisk:(UIImage *)avatar
...
+    NSData *_Nullable data;
+    for (NSUInteger attempts = 0; attempts < 5; attempts++) {
+        CGFloat quality = (CGFloat)0.95 - attempts * (CGFloat)0.1;
+        data = UIImageJPEGRepresentation(image, quality);
+        if (data.length <= kMaxAvatarBytes) {
+            return data;
+        } else {
+            // This for-loop is really just paranoia. Our avatar dimensions are so small that
+            // it's incredibly unlikely we wouldn't be able to fit our profile photo with even
+            // our highest quality.
```
charlesmchen: The stress test will be to use as input rgb noise, since I think that'll be worst-case for jPEG compression.  If even that compresses under 5mb, let's take out this loop since a) all clients should be robust to oversize avatars b) simplicity!

Signal/src/Profiles/OWSProfileManager.m
michaelkirk: Yup.

Signal/src/Profiles/OWSProfileManager.h
michaelkirk: Just to be difficult, I went with Diameter since `size` usually refers to an instance of CGSize(x, y)

Signal/src/Profiles/OWSProfileManager.m
michaelkirk: That's a good point. 

A couple thoughts:

1. We should handle group avatars the same way since it's fundamentally the same problem.
2. We still want to maintain a 1:1 AR. So e.g. if you upload a 100x200px avatar, the uploaded avatar will be 100x100 max.

charlesmchen: Yup.  So,

100x200 input = 100x100.
100x100 input = not resized or resampled.

Signal/src/Profiles/OWSProfileManager.h
charlesmchen: Okay!

Signal/src/Profiles/OWSProfileManager.m
michaelkirk: done.

(The generated noise compressed to less than 200k.)

Signal/src/Profiles/OWSProfileManager.m
michaelkirk: Since any scaling already induces a recompression, I decided this didn't warrant additional complexity. We might revisit this when doing manual cropping/scaling.

PR #2432, opened 2017-08-15
Use profile name/avatar when saving new contact

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/Profiles/OWSProfileManager.h
+++ b/Signal/src/Profiles/OWSProfileManager.h
@@ -57,6 +57,10 @@ extern NSString *const kNSNotificationName_OtherUsersProfileDidChange;
 
 - (nullable UIImage *)profileAvatarForRecipientId:(NSString *)recipientId;
 
+// Reads raw avatar data from disk if available. Uncached, so shouldn't be used frequently,
+// but useful to get the raw image data for populating cnContact.imageData without lossily re-encoding.
+- (nullable NSData *)profileAvatarDataForRecipientId:(NSString *)recipientId;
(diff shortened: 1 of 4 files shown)


Signal/src/contact/OWSContactsManager.m
```diff
@@ -419,12 +420,23 @@ - (NSString *_Nonnull)displayNameForPhoneIdentifier:(NSString *_Nullable)recipie
...
 
-    // Else fall back to just using their recipientId
-    if (displayName.length < 1) {
-        displayName = recipientId;
+    return displayName;
+}
+
+- (NSString *_Nonnull)displayNameForPhoneIdentifier:(NSString *_Nullable)recipientId
+{
+    if (!recipientId) {
```
charlesmchen: When would this be nil?

Signal/src/contact/OWSContactsManager.h
```diff
@@ -51,8 +51,13 @@ extern NSString *const OWSContactsManagerSignalAccountsDidChangeNotification;
 
 #pragma mark - Util
 
+// returns a saved contact name if available, else a saved profile name, else nil
+- (nullable NSString *)nonPhoneNameForPhoneIdentifier:(NSString *_Nullable)recipientId;
```
charlesmchen: Why is this method exposed?  I don't see it being used outside this class.

michaelkirk: I started down a different path and failed to remove the unused code after a refactor. Good catch!

PR #2433, opened 2017-08-15
Rework concurrency in the profile manager.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/Profiles/OWSProfileManager.m
+++ b/Signal/src/Profiles/OWSProfileManager.m
@@ -947,9 +947,11 @@ const NSUInteger kOWSProfileManager_MaxAvatarDiameter = 640;
 
             [self saveUserProfile:userProfile];
 
-            [ProfileFetcherJob runWithRecipientId:recipientId
-                                   networkManager:self.networkManager
-                                 ignoreThrottling:ignoreThrottling];
+            dispatch_async(dispatch_get_main_queue(), ^{
+                [ProfileFetcherJob runWithRecipientId:recipientId
+                                       networkManager:self.networkManager
+                                     ignoreThrottling:ignoreThrottling];
+            });
         }
     });
 }
--- a/Signal/src/Profiles/OWSProfileManager.m
+++ b/Signal/src/Profiles/OWSProfileManager.m
@@ -20,18 +20,14 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-// UserProfile properties should only be mutated on the main thread.
+// UserProfile properties may be read from any thread, but should
+// only be mutated when synchronized on the profile manager.
 @interface UserProfile : TSYapDatabaseObject
 
-// These properties may be accessed from any thread.
 @property (atomic, readonly) NSString *recipientId;
 @property (atomic, nullable) OWSAES128Key *profileKey;
-
-// These properties may be accessed only from the main thread.
 @property (nonatomic, nullable) NSString *profileName;
-
 @property (nonatomic, nullable) NSString *avatarUrlPath;
-
 // This filename is relative to OWSProfileManager.profileAvatarsDirPath.
 @property (nonatomic, nullable) NSString *avatarFileName;
 
@@ -39,8 +35,6 @@ NS_ASSUME_NONNULL_BEGIN
 //
 // * The last successful update finished.
 // * The current in-flight update began.
-//
-// This property may be accessed from any thread.
 @property (nonatomic, nullable) NSDate *lastUpdateDate;
 
 - (instancetype)init NS_UNAVAILABLE;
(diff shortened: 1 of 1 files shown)


charlesmchen:
Ignoring whitespace will clarify: https://github.com/WhisperSystems/Signal-iOS/pull/2433/files?w=1

charlesmchen:
There's a lot more `synchronized` now which could block some reads by the UI but profile-related writes should be rare and narrow, so I'm not too worried about it.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -101,26 +95,26 @@ @interface OWSProfileManager ()
 @property (nonatomic, readonly) YapDatabaseConnection *dbConnection;
 @property (nonatomic, readonly) TSNetworkManager *networkManager;
 
+// These properties can be accessed on any thread, while synchronized on self.
```
michaelkirk: nit: Previously while reading comments making statements about "these properties" I've been confused as to what was actually included in "these properties".  Consider either making a line comment per property or use the specific property names in the comment.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -225,150 +219,170 @@ - (UserProfile *)getOrBuildUserProfileForRecipientId:(NSString *)recipientId
...
+        BOOL isLocalUserProfile = userProfile == self.localUserProfile;
+
+        dispatch_async(dispatch_get_main_queue(), ^{
+            if (isLocalUserProfile) {
+                [[NSNotificationCenter defaultCenter] postNotificationName:kNSNotificationName_LocalProfileDidChange
+                                                                    object:nil
+                                                                  userInfo:nil];
+            } else {
+                [[NSNotificationCenter defaultCenter]
+                    postNotificationName:kNSNotificationName_OtherUsersProfileDidChange
```
michaelkirk: seems likely we'll want to include the recipientId or the userProfile in the notification.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -225,150 +219,170 @@ - (UserProfile *)getOrBuildUserProfileForRecipientId:(NSString *)recipientId
...
-        NSString *_Nullable avatarUrlPath, NSString *_Nullable avatarFileName) {
-        [self updateServiceWithProfileName:profileName
-            success:^{
-                // All reads and writes to user profiles should happen on the main thread.
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        @synchronized(self)
+        {
+            // Ensure that the success and failure blocks are called on the main thread.
+            void (^failureBlock)() = ^{
                 dispatch_async(dispatch_get_main_queue(), ^{
```
michaelkirk: For my own edification, do you have a preference for using this style vs the `DispatchMainThreadSafe` ?

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -671,156 +704,181 @@ - (void)setContactRecipientIds:(NSArray<NSString *> *)contactRecipientIds
...
+                            } else if (error) {
+                                DDLogError(@"%@ avatar download failed: %@", self.tag, error);
+                            } else if (!encryptedData) {
+                                DDLogError(@"%@ avatar encrypted data could not be read.", self.tag);
+                            } else if (!decryptedData) {
+                                DDLogError(@"%@ avatar data could not be decrypted.", self.tag);
+                            } else if (!image) {
+                                DDLogError(@"%@ avatar image could not be loaded: %@", self.tag, error);
+                            } else {
+                                // TODO: Verify that the avatar URL hasn't changed since the download began.
```
michaelkirk: re: TODO, it seems like you've addressed this with your additions a few lines up?

```
 } else if (![avatarUrlPathAtStart isEqualToString:currentUserProfile.avatarUrlPath]) {
...
```

Also to nit: `currentUserProfile` isn't the best name. For a second I thought it referred to the "local" user. Maybe `latestUserProfile`? Names are hard. Take it or leave it.

Signal/src/Profiles/OWSProfileManager.m
charlesmchen: Yup. I've tried to make things clearer by adopting a single policy: "only access the mutable state while synchronized on self".  Anyway, I've added per-property comments, but they're all the same now.

Signal/src/Profiles/OWSProfileManager.m
charlesmchen: Yup!  For whoever does the "update views to reflect changes" work.

Signal/src/Profiles/OWSProfileManager.m
charlesmchen: DispatchMainThreadSafe() is an optimization - it avoids dispatching async if we're already on the main thread.  It comes at the cost of being slightly harder to read/understand.  I try to only use it where we need that optimization (e.g. to update a view synchronously if possible).

Signal/src/Profiles/OWSProfileManager.m
charlesmchen: Done and done.

PR #2434, opened 2017-08-15
Fix layout glitch in messages view.

charlesmchen:
I found and fixed a repro for yet another form of the "stale mappings" bug.  Namely, we need to `beginLongLivedReadTransaction` and `update mappings` in that order and avoid doing any layout after the first step but before the second step is done.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -1645,7 +1645,7 @@ typedef enum : NSUInteger {
             break;
         }
         default: {
-            DDLogWarn(@"using default cell constructor for message: %@", message);
+            OWSFail(@"using default cell constructor for message: %@", message);
             cell = (JSQMessagesCollectionViewCell *)[super collectionView:collectionView
                                                    cellForItemAtIndexPath:indexPath];
             break;
@@ -1750,7 +1750,6 @@ typedef enum : NSUInteger {
 }
 
 #pragma mark - Adjusting cell label heights
-
 
 /**
  Due to the usage of JSQMessagesViewController, and it non-conformity to Dynamyc Type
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -1645,7 +1645,7 @@ - (UICollectionViewCell *)collectionView:(JSQMessagesCollectionView *)collection
             break;
         }
         default: {
-            DDLogWarn(@"using default cell constructor for message: %@", message);
+            OWSFail(@"using default cell constructor for message: %@", message);
```
charlesmchen: A small, unrelated change.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -3260,6 +3259,18 @@ - (void)yapDatabaseModified:(NSNotification *)notification
...
+    // HACK to work around radar #28167779
+    // "UICollectionView performBatchUpdates can trigger a crash if the collection view is flagged for layout"
+    // more: https://github.com/PSPDFKit-labs/radar.apple.com/tree/master/28167779%20-%20CollectionViewBatchingIssue
+    // This was our #2 crash, and much exacerbated by the refactoring somewhere between 2.6.2.0-2.6.3.8
+    //
+    // NOTE: It's critical we do this before beginLongLivedReadTransaction.
+    //       layoutIfNeeded triggers layout (obviously) which will update our cells using the current mappings
+    //       but loading cells using interactionAtIndexPath: and messageAtIndexPath:, which will return the
+    //       wrong results if the db connection has been updated but the mappings haven't.
+    [self.collectionView layoutIfNeeded];
```
michaelkirk: 💩

charlesmchen:
> Great catch. What was the repro?

Weird, I could have sworn it was in the card.

* Alice creates a new group with Bob and sends one message in that group.
* Bob navigates to that new group and sends back a single message.

PR #2437, opened 2017-08-16
Charlesmchen/profile for new and old users

charlesmchen:
Add profile view to registration workflow and upgrade/nag scenario.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -11,6 +11,7 @@
 #import "NotificationsManager.h"
 #import "OWSContactsManager.h"
 #import "OWSContactsSyncing.h"
+#import "OWSNavigationController.h"
 #import "OWSProfileManager.h"
 #import "OWSStaleNotificationObserver.h"
 #import "Pastelog.h"
@@ -802,8 +803,8 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
         self.window.rootViewController = navigationController;
     } else {
         RegistrationViewController *viewController = [RegistrationViewController new];
-        UINavigationController *navigationController =
-            [[UINavigationController alloc] initWithRootViewController:viewController];
+        OWSNavigationController *navigationController =
+            [[OWSNavigationController alloc] initWithRootViewController:viewController];
         navigationController.navigationBarHidden = YES;
         self.window.rootViewController = navigationController;
     }
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -158,7 +158,6 @@ typedef enum : NSUInteger {
     OWSVoiceMemoGestureDelegate,
     UIDocumentMenuDelegate,
     UIDocumentPickerDelegate,
-    UIGestureRecognizerDelegate,
     UIImagePickerControllerDelegate,
     UINavigationControllerDelegate,
     UITextViewDelegate>
@@ -548,10 +547,6 @@ typedef enum : NSUInteger {
 
     // In case we're dismissing a CNContactViewController which requires default system appearance
     [UIUtil applySignalAppearence];
-
-    // Since we're using a custom back button, we have to do some extra work to manage the
-    // interactivePopGestureRecognizer
-    self.navigationController.interactivePopGestureRecognizer.delegate = self;
 
     // We need to recheck on every appearance, since the user may have left the group in the settings VC,
     // or on another device.
@@ -991,10 +986,6 @@ typedef enum : NSUInteger {
     [super viewWillDisappear:animated];
 
     self.isViewVisible = NO;
-
-    // Since we're using a custom back button, we have to do some extra work to manage the
-    // interactivePopGestureRecognizer
-    self.navigationController.interactivePopGestureRecognizer.delegate = nil;
 
     [self.audioAttachmentPlayer stop];
     self.audioAttachmentPlayer = nil;
--- a/Signal/src/ViewControllers/NewGroupViewController.m
+++ b/Signal/src/ViewControllers/NewGroupViewController.m
@@ -50,7 +50,6 @@ const NSUInteger kNewGroupViewControllerAvatarWidth = 68;
 @property (nonatomic) NSMutableSet<NSString *> *memberRecipientIds;
 
 @property (nonatomic) BOOL hasUnsavedChanges;
-@property (nonatomic) BOOL shouldIgnoreSavedChanges;
 @property (nonatomic) BOOL hasAppeared;
 
 @end
@@ -552,7 +551,7 @@ const NSUInteger kNewGroupViewControllerAvatarWidth = 68;
 {
     [self.groupNameTextField resignFirstResponder];
 
-    if (!self.hasUnsavedChanges || self.shouldIgnoreSavedChanges) {
+    if (!self.hasUnsavedChanges) {
         // If user made no changes, return to conversation settings view.
         [self.navigationController popViewControllerAnimated:YES];
         return;
@@ -571,7 +570,6 @@ const NSUInteger kNewGroupViewControllerAvatarWidth = 68;
                                                      @"The label for the 'discard' button in alerts and action sheets.")
                                            style:UIAlertActionStyleDestructive
                                          handler:^(UIAlertAction *action) {
-                                             self.shouldIgnoreSavedChanges = YES;
                                              [self.navigationController popViewControllerAnimated:YES];
                                          }]];
     [controller addAction:[UIAlertAction actionWithTitle:NSLocalizedString(@"TXT_CANCEL_TITLE", nil)
@@ -655,8 +653,8 @@ const NSUInteger kNewGroupViewControllerAvatarWidth = 68;
 
 - (BOOL)shouldCancelNavigationBack
 {
-    BOOL result = self.hasUnsavedChanges && !self.shouldIgnoreSavedChanges;
-    if (result) {
+    BOOL result = self.hasUnsavedChanges;
+    if (self.hasUnsavedChanges) {
         [self backButtonPressed];
     }
     return result;
--- a/Signal/src/ViewControllers/NewGroupViewController.m
+++ b/Signal/src/ViewControllers/NewGroupViewController.m
@@ -50,6 +50,7 @@ const NSUInteger kNewGroupViewControllerAvatarWidth = 68;
 @property (nonatomic) NSMutableSet<NSString *> *memberRecipientIds;
 
 @property (nonatomic) BOOL hasUnsavedChanges;
+@property (nonatomic) BOOL shouldIgnoreSavedChanges;
 @property (nonatomic) BOOL hasAppeared;
 
 @end
@@ -551,7 +552,7 @@ const NSUInteger kNewGroupViewControllerAvatarWidth = 68;
 {
     [self.groupNameTextField resignFirstResponder];
 
-    if (!self.hasUnsavedChanges) {
+    if (!self.hasUnsavedChanges || self.shouldIgnoreSavedChanges) {
         // If user made no changes, return to conversation settings view.
         [self.navigationController popViewControllerAnimated:YES];
         return;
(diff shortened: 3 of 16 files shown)


Signal/src/ViewControllers/ProfileViewController.m
```diff
@@ -29,28 +43,49 @@ @interface ProfileViewController () <UITextFieldDelegate, AvatarViewHelperDelega
 
 @property (nonatomic) BOOL hasUnsavedChanges;
 
+@property (nonatomic) ProfileViewMode profileViewMode;
+
+@property (nonatomic) YapDatabaseConnection *databaseConnection;
```
charlesmchen: This isn't strictly necessary; we only use it in the initializer so far.

Signal/src/ViewControllers/SignalsViewController.m
```diff
@@ -530,7 +532,11 @@ - (void)displayAnyUnseenUpgradeExperience
                          completion:^{
                              [self markAllUpgradeExperiencesAsSeen];
                          }];
+    } else if (!self.hasBeenPresented && [ProfileViewController shouldDisplayProfileViewOnLaunch]) {
+        [ProfileViewController presentForUpgradeOrNag:self];
```
charlesmchen: So, this does the bare minimum to arbitrate between "on launch" conflicts. Basically we only show the profile nag if there is no "upgrade experience" to be seen.

Signal/src/ViewControllers/ProfileViewController.m
```diff
@@ -285,6 +387,50 @@ - (void)updateAvatarView
...
+{
+    // Only nag until the user sets a profile _name_.  Profile names are
+    // recommended; profile avatars are optional.
+    if ([OWSProfileManager sharedManager].localProfileName.length > 0) {
+        return NO;
+    }
+
+    NSTimeInterval kProfileNagFrequency = kDayInterval * 30;
+    NSDate *_Nullable lastPresentedDate =
+        [[[TSStorageManager sharedManager] dbReadConnection] dateForKey:kProfileView_LastPresentedDate
```
charlesmchen: Nope, this is a class method.

Signal/src/ViewControllers/ProfileViewController.m
```diff
@@ -204,14 +244,44 @@ - (void)setHasUnsavedChanges:(BOOL)hasUnsavedChanges
...
+        case ProfileViewMode_Registration:
+            self.navigationItem.leftBarButtonItem = [[UIBarButtonItem alloc]
+                initWithTitle:NSLocalizedString(@"NAVIGATION_ITEM_SKIP_BUTTON", @"A button to skip a view.")
+                        style:UIBarButtonItemStylePlain
+                       target:self
+                       action:@selector(backOrSkipButtonPressed)];
+            break;
+    }
+    if (self.hasUnsavedChanges) {
+        // If we have a unsaved changes, right item should be a "save" button.
```
charlesmchen: Our registration flow already has the problem of wanting to use big, bottom-aligned buttons for "next/done", which is already tricky since these views have text input, and when the keyboard is up there's very little screen real estate left.  

I didn't see how to solve this problem in the profile view - we want to show the avatar large, for example.  And we haven't even added yet:

* A brief description of when your profile is visible and to who, with a link to more info.
* Any text about whether or your profile name should be your real name, whether or not it needs to be unique, etc.

screen shot 2017-08-17 at 11 39 36 am
screen shot 2017-08-17 at 11 38 46 am

charlesmchen: Let's revisit this in another PR.

Signal/src/ViewControllers/ProfileViewController.m
michaelkirk: My point was not about if it will compile, but to double down on what I believe to be best practice when working with yap DB connections. Specifically - a concurrency concern (in this case, coherent read/write to when the nag was last presented) should be put through a single dbConnection.

Signal/src/ViewControllers/ProfileViewController.m
```diff
@@ -204,14 +244,44 @@ - (void)setHasUnsavedChanges:(BOOL)hasUnsavedChanges
...
+        [self.navigationController setNavigationBarHidden:NO animated:YES];
+    }
+
+    // Always display a left item to leave the view without making changes.
+    // This might be a "back", "skip" or "cancel" button depending on the
+    // context.
+    switch (self.profileViewMode) {
+        case ProfileViewMode_AppSettings:
+            self.navigationItem.leftBarButtonItem =
+                [self createOWSBackButtonWithTarget:self selector:@selector(backOrSkipButtonPressed)];
```
charlesmchen: Okay, I found a better approach on stackoverflow that works pretty well.  It eliminates the "custom" back buttons everywhere except conversation view.  It's a big change, so PTAL @michaelkirk

charlesmchen: Please note that the "back gesture" currently skips past the "are you sure" check. I think we could go the other way if we wanted to.

Signal/src/ViewControllers/ProfileViewController.m
charlesmchen: Sorry, I don't get it.  Are you saying we should make a new db connection here?

Signal/src/ViewControllers/ProfileViewController.m
charlesmchen: Okay, it was a bit of a pain, but I've reworked how "cancel navigate back" works throughout the app.  This required extensive changes.

PTAL @michaelkirk

Signal/src/AppDelegate.m
```diff
@@ -802,8 +803,8 @@ - (void)ensureRootViewController
         self.window.rootViewController = navigationController;
     } else {
         RegistrationViewController *viewController = [RegistrationViewController new];
-        UINavigationController *navigationController =
-            [[UINavigationController alloc] initWithRootViewController:viewController];
+        OWSNavigationController *navigationController =
+            [[OWSNavigationController alloc] initWithRootViewController:viewController];
```
charlesmchen: We need to use OWSNavigationController (or a subclass thereof) to support views that can cancel "navigate back".

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -549,10 +548,6 @@ - (void)viewWillAppear:(BOOL)animated
     // In case we're dismissing a CNContactViewController which requires default system appearance
     [UIUtil applySignalAppearence];
 
-    // Since we're using a custom back button, we have to do some extra work to manage the
-    // interactivePopGestureRecognizer
-    self.navigationController.interactivePopGestureRecognizer.delegate = self;
```
charlesmchen: I'm not clear why we were doing this since this class and its JSQ superclass don't actually implement any `UIGestureRecognizerDelegate` methods.  It was complicating and/or interfering with `OWSNavigationController`, so I've removed it.

Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -97,8 +99,6 @@ - (void)loadView
     [super loadView];
 
     self.title = NSLocalizedString(@"NEW_GROUP_DEFAULT_TITLE", @"The navbar title for the 'new group' view.");
-    self.navigationItem.leftBarButtonItem =
-        [self createOWSBackButtonWithTarget:self selector:@selector(backButtonPressed:)];
```
charlesmchen: I was mostly able to remove the custom back buttons - except in conversation view where we're deliberately changing its appearance.

Signal/src/ViewControllers/OWSConversationSettingsViewController.m
```diff
@@ -827,10 +827,7 @@ - (void)showUpdateGroupView:(UpdateGroupMode)mode
     updateGroupViewController.conversationSettingsViewDelegate = self.conversationSettingsViewDelegate;
     updateGroupViewController.thread = (TSGroupThread *)self.thread;
     updateGroupViewController.mode = mode;
-
-    UINavigationController *navigationController =
-        [[UINavigationController alloc] initWithRootViewController:updateGroupViewController];
-    [self presentViewController:navigationController animated:YES completion:nil];
+    [self.navigationController pushViewController:updateGroupViewController animated:YES];
```
charlesmchen: I don't know why we were presenting the "edit group" view as a new modal instead of pushing it.

Signal/src/ViewControllers/OWSNavigationController.h
```diff
@@ -0,0 +1,24 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+// Any view controller which wants to be able cancel back button
+// presses and back gestures should implement this protocol.
+@protocol OWSNavigationView <NSObject>
```
charlesmchen: There's a bunch of subtlety here which I've tried to capture in comments.

Signal/src/ViewControllers/ProfileViewController.m
charlesmchen: Oh, I see, you're saying that we should make a new db connection that hangs on the class for consistency.  

I'd rather not keep the connection dangling around forever when we'll use it very rarely, especially since we read/write from this particular state very rarely and a race condition is pretty unlikely. 

Hmm, point taken though, I guess I'll just use the storage manager's shared read/write connection.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
michaelkirk: Removing this is fine, as long as the user can still swipe to go back.

Signal/src/ViewControllers/OWSConversationSettingsViewController.m
michaelkirk: The way I choose a transition is to consider if the new screen is more akin to "starting a new activity" or "drilling down" or "progressing" with the current activity. From a strict utility sense, they're equivalent, i.e. "going back" in a pushed view has the same end result as "canceling" a modal. As a rough heuristic (there are exceptions of course), I find that things which can be "canceled" are better expressed as modals.

I admit that this is subjective, so if I haven't persuaded you, I'm fine with this change.

Signal/src/ViewControllers/OWSNavigationController.m
```diff
@@ -0,0 +1,76 @@
...
+
+#pragma mark - UIGestureRecognizerDelegate
+
+// We serve as the UIGestureRecognizerDelegate of the interactivePopGestureRecognizer
+// in order to cancel some "back" gestures - for example,
+// if a view has unsaved changes.
+- (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
+{
+    UIViewController *topViewController = self.topViewController;
+    if ([topViewController conformsToProtocol:@protocol(OWSNavigationView)]) {
```
michaelkirk: This reads as though it would have the effect of blocking all gesture recognizers in the view. Do we need to hone this in to only target the interactivePopGestureRecognizer?

What's the end result here - that swiping back does nothing when there are unsaved changes?

Signal/src/ViewControllers/ProfileViewController.m
```diff
@@ -29,28 +44,47 @@ @interface ProfileViewController () <UITextFieldDelegate, AvatarViewHelperDelega
...
+    self = [super init];
+
+    if (!self) {
+        return self;
+    }
+
+    self.profileViewMode = profileViewMode;
+
+    // Use the TSStorageManager.dbReadWriteConnection for consistency with the reads below.
+    [[[TSStorageManager sharedManager] dbReadWriteConnection] setDate:[NSDate new]
```
michaelkirk: In my mind using the globally shared dbConnection's make things more complicated to reason about WRT to unexpected blocking (so they should be avoided), and it's simpler to have a dedicated connection for independent actors.

Can you tell me how you're thinking about it - e.g. When is it preferable to use the global dbConnections and when is it preferable to build a dedicated dbConnection?

Signal/src/ViewControllers/OWSNavigationController.m
charlesmchen: A GRDelegate only applies to GRs for which it is set as the delegate.  We're only setting this vc as delegate of the `interactivePopGestureRecognizer`.

Signal/src/ViewControllers/ProfileViewController.m
charlesmchen: I agree that we should be moving away from the global db connections, hence the way I had it before.  But I don't want to open a new db connection _that we never close_ everywhere we do a write.  Why don't we talk at standup.

Signal/src/ViewControllers/ProfileViewController.m
charlesmchen: See: https://github.com/WhisperSystems/Signal-iOS/pull/2452

PR #2438, opened 2017-08-16
Show activity indicator during profile update; trim whitespace from p…

charlesmchen:
…rofile names.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ProfileViewController.m
+++ b/Signal/src/ViewControllers/ProfileViewController.m
@@ -223,22 +223,40 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)updateProfile
 {
     __weak ProfileViewController *weakSelf = self;
-    [OWSProfileManager.sharedManager updateLocalProfileName:self.nameTextField.text
-        avatarImage:self.avatar
-        success:^{
-            [weakSelf.navigationController popViewControllerAnimated:YES];
-        }
-        failure:^{
-            [OWSAlerts showAlertWithTitle:NSLocalizedString(@"ALERT_ERROR_TITLE", @"")
-                                  message:NSLocalizedString(@"PROFILE_VIEW_ERROR_UPDATE_FAILED",
-                                              @"Error message shown when a profile update fails.")];
-        }];
+
+    // Show an activity indicator to block the UI during the profile upload.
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ProfileViewController.m
```diff
@@ -223,22 +223,40 @@ - (void)updatePressed
...
-        }
-        failure:^{
-            [OWSAlerts showAlertWithTitle:NSLocalizedString(@"ALERT_ERROR_TITLE", @"")
-                                  message:NSLocalizedString(@"PROFILE_VIEW_ERROR_UPDATE_FAILED",
-                                              @"Error message shown when a profile update fails.")];
-        }];
+
+    // Show an activity indicator to block the UI during the profile upload.
+    UIAlertController *alertController =
+        [UIAlertController alertControllerWithTitle:NSLocalizedString(@"PROFILE_VIEW_SAVING",
```
michaelkirk: nit: try to include the UI context for translators e.g. "Alert title that indicates the user's profile view is being saved"

Signal/src/ViewControllers/ProfileViewController.m
```diff
@@ -223,22 +223,40 @@ - (void)updatePressed
...
+                                                        @"Indicates that the user's profile view is being saved.")
+                                            message:nil
+                                     preferredStyle:UIAlertControllerStyleAlert];
+
+    [self presentViewController:alertController
+                       animated:YES
+                     completion:^{
+                         [alertController dismissViewControllerAnimated:NO completion:nil];
```
michaelkirk: Am I understanding that we present the "saving" alert only momentarily, then immediately dismiss it. Then start uploading.

I would have thought the `OWSProfileManager.sharedManager updateLocalProfileName` completion handler would dismiss the alert, otherwise it will just be a quick flash right?

Signal/src/ViewControllers/ProfileViewController.m
charlesmchen: Better, done.

Signal/src/ViewControllers/ProfileViewController.m
charlesmchen: Thanks, fixed.

PR #2439, opened 2017-08-16
Charlesmchen/new contact conversations vs profile whitelist

charlesmchen:
* Add debug UI tools for clearing and logging the profile whitelist.
* Auto-add new contact threads to profile whitelist when local user sends first message to that thread.
* Ensure dynamic interactions have a non-negative timestamp even if the conversation was empty.
* Only call updateMessageMappingRangeOptions _after_ beginLongLivedReadTransaction and updating messageMappings.
* Improve documentation around how to avoid corrupt mappings in conversation view.
* Fix edge cases around large initial range sizes.
* Always treat dynamic interactions as read.
* Rebuild the “unseen” database views to remove dynamic interactions (see above).

PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -176,7 +176,7 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     // we need to bump this constant.
     //
     // We added a number of database views in v2.13.0.
-    NSString *kLastVersionWithDatabaseViewChange = @"2.13.0";
+    NSString *kLastVersionWithDatabaseViewChange = @"2.16.0";
     BOOL mayNeedUpgrade = ([TSAccountManager isRegistered] && lastLaunchedAppVersion
         && (!lastCompletedLaunchAppVersion ||
                [VersionMigrations isVersion:lastCompletedLaunchAppVersion
--- a/Signal/src/Profiles/OWSProfileManager.m
+++ b/Signal/src/Profiles/OWSProfileManager.m
@@ -598,6 +598,8 @@ const NSUInteger kOWSProfileManager_MaxAvatarDiameter = 640;
 #ifdef DEBUG
 - (void)clearProfileWhitelist
 {
+    DDLogWarn(@"%@ Clearing the profile whitelist.", self.tag);
+
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
         @synchronized(self)
         {
@@ -621,8 +623,20 @@ const NSUInteger kOWSProfileManager_MaxAvatarDiameter = 640;
             DDLogError(@"groupProfileWhitelistCache: %zd", self.groupProfileWhitelistCache.count);
             DDLogError(@"kOWSProfileManager_UserWhitelistCollection: %zd",
                 [self.dbConnection numberOfKeysInCollection:kOWSProfileManager_UserWhitelistCollection]);
+            [self.dbConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
+                [transaction enumerateKeysInCollection:kOWSProfileManager_UserWhitelistCollection
+                                            usingBlock:^(NSString *_Nonnull key, BOOL *_Nonnull stop) {
+                                                DDLogError(@"\t profile whitelist user: %@", key);
+                                            }];
+            }];
             DDLogError(@"kOWSProfileManager_GroupWhitelistCollection: %zd",
                 [self.dbConnection numberOfKeysInCollection:kOWSProfileManager_GroupWhitelistCollection]);
+            [self.dbConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
+                [transaction enumerateKeysInCollection:kOWSProfileManager_GroupWhitelistCollection
+                                            usingBlock:^(NSString *_Nonnull key, BOOL *_Nonnull stop) {
+                                                DDLogError(@"\t profile whitelist group: %@", key);
+                                            }];
+            }];
         }
     });
 }
--- a/Signal/src/Profiles/OWSProfileManager.m
+++ b/Signal/src/Profiles/OWSProfileManager.m
@@ -595,6 +595,39 @@ const NSUInteger kOWSProfileManager_MaxAvatarDiameter = 640;
 
 #pragma mark - Profile Whitelist
 
+#ifdef DEBUG
+- (void)clearProfileWhitelist
+{
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        @synchronized(self)
+        {
+            [self.userProfileWhitelistCache removeAllObjects];
+            [self.groupProfileWhitelistCache removeAllObjects];
+
+            [self.dbConnection purgeCollection:kOWSProfileManager_UserWhitelistCollection];
+            [self.dbConnection purgeCollection:kOWSProfileManager_GroupWhitelistCollection];
+            OWSAssert(0 == [self.dbConnection numberOfKeysInCollection:kOWSProfileManager_UserWhitelistCollection]);
+            OWSAssert(0 == [self.dbConnection numberOfKeysInCollection:kOWSProfileManager_GroupWhitelistCollection]);
+        }
+    });
+}
+
+- (void)logProfileWhitelist
+{
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        @synchronized(self)
+        {
+            DDLogError(@"userProfileWhitelistCache: %zd", self.userProfileWhitelistCache.count);
+            DDLogError(@"groupProfileWhitelistCache: %zd", self.groupProfileWhitelistCache.count);
+            DDLogError(@"kOWSProfileManager_UserWhitelistCollection: %zd",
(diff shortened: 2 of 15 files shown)


Signal/src/AppDelegate.m
```diff
@@ -176,7 +176,7 @@ - (UIViewController *)loadingRootViewController
     // we need to bump this constant.
     //
     // We added a number of database views in v2.13.0.
-    NSString *kLastVersionWithDatabaseViewChange = @"2.13.0";
+    NSString *kLastVersionWithDatabaseViewChange = @"2.16.0";
```
charlesmchen: Not psyched about this; let's discuss.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -166,6 +166,19 @@ @interface MessagesViewController () <AVAudioPlayerDelegate,
...
+// * The first (required) step is to update uiDatabaseConnection using beginLongLivedReadTransaction.
+// * The second (required) step is to update messageMappings.
+// * The third (optional) step is to update the messageMappings range using
+//   updateMessageMappingRangeOptions.
+// * The steps must be done in strict order.
+// * If we do any of the steps, we must do all of the required steps.
+// * We can't use messageMappings in between the first and second steps; e.g.
+//   we can't do any layout, since that uses numberOfItemsInSection: and
+//   interactionAtIndexPath: which use the messageMappings.
+// * If we do the third step, we must call resetContentAndLayout afterward.
```
charlesmchen: It took so much blood sweat and tears to learn this that I figured I would document it.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -370,10 +383,10 @@ - (void)configureForThread:(TSThread *)thread
         [[YapDatabaseViewMappings alloc] initWithGroups:@[ thread.uniqueId ] view:TSMessageDatabaseViewExtensionName];
     // We need to impose the range restrictions on the mappings immediately to avoid
     // doing a great deal of unnecessary work and causing a perf hotspot.
-    [self updateMessageMappingRangeOptions];
     [self.uiDatabaseConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
         [self.messageMappings updateWithTransaction:transaction];
     }];
+    [self updateMessageMappingRangeOptions];
```
charlesmchen: Yet another fix around the mappings.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -1481,6 +1492,9 @@ - (void)didPressSendButton:(UIButton *)button
         [self clearDraft];
         [self finishSendingMessage];
         [((OWSMessagesToolbarContentView *)self.inputToolbar.contentView)ensureSubviews];
+        if (didAddToProfileWhitelist) {
+            [self ensureDynamicInteractions];
```
charlesmchen: If we send a message in a new (empty) contact thread, we've just created it and we can auto-add this user to the profile whitelist.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -2285,17 +2299,23 @@ - (void)updateLoadEarlierVisible
...
+    // The "page-based" range length may have been increased by loading "prev" pages at the
+    // top of the window.
+    NSUInteger rangeLength;
+    while (YES) {
+        rangeLength = kYapDatabasePageSize * (self.page + 1);
+        if (rangeLength >= oldLength) {
+            break;
+        }
+        self.page = self.page + 1;
+    }
```
charlesmchen: A small fix to improve behavior of the "load more" button in large conversations.

charlesmchen: That should read: A small fix to improve behavior of the "load more" button in conversations with a large amount of unread messages.

Signal/src/ViewControllers/SendExternalFileViewController.m
```diff
@@ -64,6 +64,7 @@ - (void)threadWasSelected:(TSThread *)thread
         return;
     }
 
+    [ThreadUtil addThreadToProfileWhitelistIfEmptyContactThread:thread];
```
charlesmchen: It's not great that we have to remember to call `addThreadToProfileWhitelistIfEmptyContactThread` anywhere that we might be sending a message to a thread that we've just created, but I was a bit wary of pulling this isn't ThreadUtil or MessageSender, not least because we only want to do this when the user sends a "normal" message - a text message or attachment - not any other message type.  We could probably identify "normal" messages... tough call.

Signal/src/util/ThreadUtil.h
```diff
@@ -38,6 +38,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, readonly) BOOL hasMoreUnseenMessages;
 
+@property (nonatomic, readonly) BOOL didInsertDynamicInteractions;
```
charlesmchen: I might remove this property in my next PR.

SignalServiceKit/src/Storage/TSDatabaseView.m
```diff
@@ -144,7 +144,7 @@ + (void)asyncRegisterUnseenDatabaseView
 
     [self registerMessageDatabaseViewWithName:TSUnseenDatabaseViewExtensionName
                                  viewGrouping:viewGrouping
-                                      version:@"1"
+                                      version:@"2"
```
charlesmchen: I don't think we need to bother with rebuilding the unread database view, since its just used for badge counts.

michaelkirk: As long as we're bumping the DB view, should the DB view have logic to discard anything which `isDynamicInteraction` - then we wouldn't have to worry about the "read"/"unread" status of the interactions.

Signal/src/util/ThreadUtil.h
charlesmchen: Removed it for now.

SignalServiceKit/src/Storage/TSDatabaseView.m
charlesmchen: That was the other option - they'd have the same effect.  I don't see why we'd do both.  

I felt this approach was better so that we were consistent about the state of the `read` property and the `OWSReadTracking` protocol.  If we only ignored dynamic interactions, we might be surprised to find that some interactions in a thread still had `read` false when the "unseen" view was empty, etc.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -595,6 +595,39 @@ - (void)updateServiceWithProfileName:(nullable NSString *)localProfileName
 
 #pragma mark - Profile Whitelist
 
+#ifdef DEBUG
+- (void)clearProfileWhitelist
+{
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        @synchronized(self)
+        {
+            [self.userProfileWhitelistCache removeAllObjects];
```
charlesmchen: Done.

SignalServiceKit/src/Storage/TSDatabaseView.m
michaelkirk: I agree it's good to keep the read status updated as you have. I was just thinking for paranoia incase we add a new dynamic interaction and forget that we have to have a line in each initializer to mark it as read - but no biggie.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -595,6 +595,39 @@ - (void)updateServiceWithProfileName:(nullable NSString *)localProfileName
...
+{
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        @synchronized(self)
+        {
+            DDLogError(@"userProfileWhitelistCache: %zd", self.userProfileWhitelistCache.count);
+            DDLogError(@"groupProfileWhitelistCache: %zd", self.groupProfileWhitelistCache.count);
+            DDLogError(@"kOWSProfileManager_UserWhitelistCollection: %zd",
+                [self.dbConnection numberOfKeysInCollection:kOWSProfileManager_UserWhitelistCollection]);
+            DDLogError(@"kOWSProfileManager_GroupWhitelistCollection: %zd",
+                [self.dbConnection numberOfKeysInCollection:kOWSProfileManager_GroupWhitelistCollection]);
```
charlesmchen: Good idea, done.

Signal/src/util/ThreadUtil.m
```diff
@@ -437,14 +445,14 @@ + (ThreadDynamicInteractions *)ensureDynamicInteractionsForThread:(TSThread *)th
             // We want the offer to be the first interaction in their
```
charlesmchen: Reworked and moved this comment.

SignalServiceKit/src/Storage/YapDatabaseConnection+OWS.m
```diff
@@ -94,6 +94,15 @@ - (nullable NSDate *)dateForKey:(NSString *)key inCollection:(NSString *)collect
 
 #pragma mark -
 
+- (NSUInteger)numberOfKeysInCollection:(NSString *)collection
+{
+    __block NSUInteger result;
+    [self readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
```
charlesmchen: Good catch

PR #2441, opened 2017-08-16
Clarify where we use the profile name

michaelkirk:
Use profile name in less places, and add some hints to differentiate profile name usage from a name pulled from the contacts.

image

image

PTAL @charlesmchen

### DIFF
--- a/Signal/src/contact/OWSContactsManager.m
+++ b/Signal/src/contact/OWSContactsManager.m
@@ -413,10 +413,14 @@ NSString *const kTSStorageManager_AccountLastNames = @"kTSStorageManager_Account
 - (nullable NSString *)formattedProfileNameForRecipientId:(NSString *)recipientId
 {
     NSString *_Nullable profileName = [self.profileManager profileNameForRecipientId:recipientId];
-    if (profileName == nil) {
+    if (profileName.length == 0) {
         return nil;
     }
-    return [@"~" stringByAppendingString:profileName];
+
+    NSString *profileNameFormatString = NSLocalizedString(@"PROFILE_NAME_LABEL_FORMAT",
+        @"Prepend a simple marker to differentiate the profile name, embeds the contact's {{profile name}}.");
+
+    return [NSString stringWithFormat:profileNameFormatString, profileName];
 }
 
 - (NSString *_Nonnull)displayNameForPhoneIdentifier:(NSString *_Nullable)recipientId
@@ -530,9 +534,15 @@ NSString *const kTSStorageManager_AccountLastNames = @"kTSStorageManager_Account
         return [[NSAttributedString alloc] initWithString:savedContactName];
     }
 
-    NSString *_Nullable formattedProfileName = [self formattedProfileNameForRecipientId:recipientId];
-    if (formattedProfileName.length > 0) {
-        NSString *numberAndProfileName = [NSString stringWithFormat:@"%@ %@", recipientId, formattedProfileName];
+    NSString *_Nullable profileName = [self.profileManager profileNameForRecipientId:recipientId];
+    if (profileName.length > 0) {
+        NSString *numberAndProfileNameFormat = NSLocalizedString(@"PROFILE_NAME_AND_PHONE_NUMBER_LABEL_FORMAT",
+            @"Label text combining the phone number and profile name separated by a simple demarcation character. "
+            @"Phone number should be most prominent. '%1$@' is replaced with {{phone number}} and '%2$@' is replaced "
+            @"with {{profile name}}");
+
+        NSString *numberAndProfileName =
+            [NSString stringWithFormat:numberAndProfileNameFormat, recipientId, profileName];
         return [[NSAttributedString alloc] initWithString:numberAndProfileName];
     }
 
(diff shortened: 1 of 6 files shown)


charlesmchen:
"this is my profile name" feels unwieldy.  Can we use something shorted like "(profile name)" in parens?

Signal/src/contact/OWSContactsManager.m
```diff
@@ -399,12 +399,26 @@ - (BOOL)phoneNumber:(PhoneNumber *)phoneNumber1 matchesNumber:(PhoneNumber *)pho
...
                              @"Displayed if for some reason we can't determine a contacts phone number *or* name");
 }
 
+- (nullable NSString *)formattedProfileNameForRecipientId:(NSString *)recipientId
+{
+    NSString *_Nullable profileName = [self.profileManager profileNameForRecipientId:recipientId];
+    if (profileName == nil) {
+        return nil;
+    }
+    return [@"~" stringByAppendingString:profileName];
```
charlesmchen: What about RTL?

Signal/src/contact/OWSContactsManager.m
```diff
@@ -399,12 +399,26 @@ - (BOOL)phoneNumber:(PhoneNumber *)phoneNumber1 matchesNumber:(PhoneNumber *)pho
...
 - (NSString *)unknownContactName
 {
     return NSLocalizedString(@"UNKNOWN_CONTACT_NAME",
                              @"Displayed if for some reason we can't determine a contacts phone number *or* name");
 }
 
+- (nullable NSString *)formattedProfileNameForRecipientId:(NSString *)recipientId
+{
+    NSString *_Nullable profileName = [self.profileManager profileNameForRecipientId:recipientId];
+    if (profileName == nil) {
```
charlesmchen: a) let's make sure we're stripping leading and trailing whitespace from profile names when we decrypt them. I already do so for local user's profile name changes.
b) I'm thinking we should also write and use a function to filter profile names for invalid characters and apply it to incoming and outgoing profile names.  I'm sure we'll add more and more rules to that filter with time, e.g. the unicode character the changes the orientation of text, etc.
c) I'm using `profileName.length > or == 0` instead of nil checks so that we handle empty names appropriately.

Signal/src/contact/OWSContactsManager.m
```diff
@@ -521,6 +522,24 @@ - (NSAttributedString *)formattedFullNameForRecipientId:(NSString *)recipientId
...
+{
+    // Prefer a saved name from system contacts, if available
+    NSString *_Nullable savedContactName = [self cachedDisplayNameForRecipientId:recipientId];
+    if (savedContactName.length > 0) {
+        return [[NSAttributedString alloc] initWithString:savedContactName];
+    }
+
+    NSString *_Nullable formattedProfileName = [self formattedProfileNameForRecipientId:recipientId];
+    if (formattedProfileName.length > 0) {
+        NSString *numberAndProfileName = [NSString stringWithFormat:@"%@ %@", recipientId, formattedProfileName];
```
charlesmchen: RTL

michaelkirk:
> "this is my profile name" feels unwieldy. Can we use something shorted like "(profile name)" in parens?

Hah, sorry, I literally put "This is my profile name." as my profile name - it's not a label. It could have said "Jonny" or whatever, then the text would have read "+15555555555 ~Jonny"

Signal/src/views/ContactTableViewCell.m
```diff
@@ -60,19 +61,37 @@ - (void)configureProgrammatically
...
+
+    // profileNameLabel can be zero sized, in which case nameLabel essentially occupies the totality of
+    // nameContainerView's frame.
+    [_profileNameLabel autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:_nameLabel];
+    [_profileNameLabel autoPinEdgeToSuperviewEdge:ALEdgeBottom];
+    [_profileNameLabel autoPinWidthToSuperview];
+
+    [nameContainerView autoVCenterInSuperview];
+    [nameContainerView autoPinLeadingToTrailingOfView:_avatarView margin:12.f];
+    [nameContainerView autoPinTrailingToSuperView];
```
charlesmchen: Doesn't `ContactTableViewCell` sometimes use accessory views, e.g. "not verified" in your screenshots?  Shouldn't we use a "manual" accessory view and pin trailing to that, with margin?

Signal/src/contact/OWSContactsManager.m
michaelkirk: Hm. Any insight into what this should be? 

We could localize the format, but it will be a bit unwieldy for the translators. Do you think this is reasonable?

/* Format string for displaying a profile name below a message bubble.  "%1$@ embeds the contact's {{phone number}} and  %2$@ embeds their {{profile name}} */
"PROFILE_NAME_FORMAT" = @"%1$@ ~%2$@"

Or should we just flip it when RTL?

Signal/src/contact/OWSContactsManager.m
michaelkirk: a) done
b) made a card
c) done

Signal/src/views/ContactTableViewCell.m
michaelkirk: Note that the superview we're pinning `nameContainerView` to is the `self.contentView`. When there is an accessory view, the contentView is resized to accommodate it, so I believe this is correct as written.

Signal/src/contact/OWSContactsManager.m
michaelkirk: FYI I ended up going with a localized format string.

Signal/src/contact/OWSContactsManager.m
michaelkirk: done.

PR #2450, opened 2017-08-21
Don't fail when profile decryption fails.

charlesmchen:
PTAL @michaelkirk


michaelkirk:
LGTM

michaelkirk:
I think we can close this, as it's now redundant with https://github.com/WhisperSystems/Signal-iOS/commit/7ede899a74007fabd9036ea3c0e73064b8c56ea9

PR #2452, opened 2017-08-21
Charlesmchen/rework profile view

charlesmchen:
screen shot 2017-08-21 at 12 22 00 pm
screen shot 2017-08-21 at 12 21 44 pm

* We prefer a layout that won't be hidden on small form factor devices.
* We could add a large "Save/Skip" button in the view when it is used in the onboarding workflow. 
* I added a first pass at the "profile info" copy and the link to more info.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ProfileViewController.h
+++ b/Signal/src/ViewControllers/ProfileViewController.h
@@ -2,13 +2,11 @@
 //  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
-#import "OWSTableViewController.h"
-
 NS_ASSUME_NONNULL_BEGIN
 
 @class SignalsViewController;
 
-@interface ProfileViewController : OWSTableViewController
+@interface ProfileViewController : UIViewController
 
 - (instancetype)init NS_UNAVAILABLE;
 
--- a/Signal/src/ViewControllers/ProfileViewController.m
+++ b/Signal/src/ViewControllers/ProfileViewController.m
@@ -38,8 +38,6 @@ NSString *const kProfileView_LastPresentedDate = @"kProfileView_LastPresentedDat
 
 @property (nonatomic) UIImageView *cameraImageView;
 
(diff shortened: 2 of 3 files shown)


charlesmchen:
I reworked the profile view again in light of our conversation.

screen shot 2017-08-21 at 3 09 39 pm
screen shot 2017-08-21 at 3 09 25 pm
screen shot 2017-08-21 at 3 06 24 pm
screen shot 2017-08-21 at 3 03 33 pm

charlesmchen:
This PR is now ready for review. PTAL @michaelkirk

michaelkirk:
> I reworked the profile view again in light of our conversation.

That second screenshot seems off. Is the big blue button really supposed to say "Skip"? Maybe it should say "Save" but be disabled until they input some text.

charlesmchen:
> That second screenshot seems off. Is the big blue button really supposed to say "Skip"? Maybe it should say "Save" but be disabled until they input some text.

I thought the profile was optional.  I was going to show "Skip" until they changed one of the fields.  Feels odd to "save" when you have no changes, and I want to _heavily_ emphasize that this is optional - not least to prevent users misunderstanding that this is now required and becoming upset.

michaelkirk:
> I thought the profile was optional.

I'm not advocating to make the profile mandatory, I'm advocating for making the primary action (the big button) in pursuit of people completing their profile. So it would be "Save" even before they can successfully save, but be disabled, until they enter content. The navbar button "Skip" would remain usable the entire time. They can skip their profile.

To your point, this does put more design weight behind *completing* the profile vs. *skipping* it -  which is my intention. 

> Feels odd to "save" when you have no changes

This is a pretty ubiquitous UX pattern. A control is disabled until the user has provided the appropriate inputs. e.g. see the "Send" button in the messages view controller.

If you want to keep what you've proposed as is that's OK, but I wanted to make sure you understood my proposal since it seems like I wasn't clear enough.

PR #2456, opened 2017-08-22
Add assert to ensure that we don't use write transactions before sync…

charlesmchen:
… database view registration is complete.

PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Storage/TSStorageManager.m
+++ b/SignalServiceKit/src/Storage/TSStorageManager.m
@@ -33,6 +33,91 @@ static const NSString *const databaseName = @"Signal.sqlite";
 static NSString *keychainService          = @"TSKeyChainService";
 static NSString *keychainDBPassAccount    = @"TSDatabasePass";
 
+static BOOL isDatabaseInitializedFlag = NO;
+
+NSObject *isDatabaseInitializedFlagLock()
+{
+    static NSObject *instance = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        instance = [NSObject new];
+    });
+    return instance;
+}
+
+BOOL isDatabaseInitialized()
+{
+    @synchronized(isDatabaseInitializedFlagLock())
+    {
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -198,6 +291,8 @@ - (void)setupDatabaseWithSafeBlockingMigrations:(void (^_Nonnull)())safeBlocking
     [self.database registerExtension:[TSDatabaseSecondaryIndexes registerTimeStampIndex] withName:@"idx"];
     [OWSMessageReceiver syncRegisterDatabaseExtension:self.database];
 
+    setDatabaseInitialized();
```
michaelkirk: Can you leave a comment as to why we're doing this here after the sync migrations but before the migrations and async registrations?

SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -33,6 +33,91 @@
...
+
+@interface OWSDatabase : YapDatabase
+
+@end
+
+#pragma mark -
+
+@implementation OWSDatabase
+
+- (YapDatabaseConnection *)newConnection
```
michaelkirk: suggested comment:

This clobbers the superclass implementation to include instrumentation which ensures the database is in a ready state before creating transactions.

charlesmchen: Good point.  I've added a bunch of comments.

SignalServiceKit/src/Storage/TSStorageManager.m
charlesmchen: Well, in the absence of finding documentation that can shed light on the issue we've been seeing, this issue only seems to affect sync and not async registrations.  We've always been opening write transactions before the async registrations complete without negative consequences.

michaelkirk: Right - I just want to capture what we believe in comments so when we revisit this later or get hit by a bus or something...

PR #2458, opened 2017-08-22
Fix slow start crash after upgrade to 2.16

michaelkirk:
We were opening a write transaction before our sync extensions were
registered. This seems to have caused our views to rebuild themselves
once they did register, which in turn can cause device timeout.

Instead of opening transcations in `init` we should explicitly setup any
state once we know the db to be in a ready state.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/Profiles/OWSProfileManager.m
+++ b/Signal/src/Profiles/OWSProfileManager.m
@@ -106,7 +106,7 @@ const NSUInteger kOWSProfileManager_MaxAvatarDiameter = 640;
 @property (nonatomic, readonly) TSNetworkManager *networkManager;
 
 // This property can be accessed on any thread, while synchronized on self.
-@property (atomic, nullable) UserProfile *localUserProfile;
+@property (nonatomic, readonly) UserProfile *localUserProfile;
 // This property can be accessed on any thread, while synchronized on self.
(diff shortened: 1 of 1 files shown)


michaelkirk:
@charlesmchen I've rolled this back to a smaller change, which, when paired with your https://github.com/WhisperSystems/Signal-iOS/pull/2456 should cover us.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -264,6 +252,44 @@ - (void)saveUserProfile:(UserProfile *)userProfile
...
+    {
+        if (_localUserProfile == nil) {
+            _localUserProfile = [self getOrCreateLocalProfile];
+        }
+
+        return _localUserProfile;
+    }
+}
+
+- (UserProfile *)getOrCreateLocalProfile
```
charlesmchen: Why bother pulling this out into another method?  Please at least leave a comment that `getOrCreateLocalProfile` should only be called by `localUserProfile`, but IMO better to combine them and avoid the possibility of it being called accidentally.

michaelkirk: Good point. I'll make that change.

PR #2460, opened 2017-08-23
Ensure message view range is properly truncated with view is configured.

charlesmchen:
Fixes a serious perf bug I introduced with earlier work in this release.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -2422,6 +2422,7 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
     NSUInteger targetLength = oldLength;
     if (mode == MessagesRangeSizeMode_Truncate) {
         // During the initial configuration of the view, we want to truncate the
+        // range size.
         targetLength = MIN(targetLength, (NSUInteger)(kYapDatabasePageSize * kYapDatabaseMaxInitialPageCount));
     }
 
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -112,6 +112,12 @@
     kMediaTypeVideo,
 } kMediaTypes;
 
+typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
```
charlesmchen: I'm using an enum instead of a BOOL param to be a bit more readable.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -2386,17 +2411,24 @@ - (void)updateLoadEarlierVisible
...
-- (void)updateMessageMappingRangeOptions
+- (void)updateMessageMappingRangeOptions:(MessagesRangeSizeMode)mode
 {
     // The "old" range length may have been increased by insertions of new messages
     // at the bottom of the window.
     NSUInteger oldLength = [self.messageMappings numberOfItemsInGroup:self.thread.uniqueId];
+
+    NSUInteger targetLength = oldLength;
+    if (mode == MessagesRangeSizeMode_Truncate) {
+        // During the initial configuration of the view, we want to truncate the
```
michaelkirk: Comment cropped?

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
michaelkirk: :+1:

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
charlesmchen: Fixed.

PR #2462, opened 2017-08-23
More performant unread indicator search

michaelkirk:
It occurred to me there was a better way to do this lookup - but I couldn't reproduce the code path.

When does this code get executed @charlesmchen?

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -664,16 +664,16 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
         return nil;
     }
 
-    // TODO: We could do binary search.
-    int numberOfMessages = (int)[self.messageMappings numberOfItemsInGroup:self.thread.uniqueId];
-    for (int i = 0; i < numberOfMessages; i++) {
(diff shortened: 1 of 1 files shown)


charlesmchen:
Just search for `indexPathOfUnreadMessagesIndicator`.  It's used in `scrollToDefaultPosition` and `scrollToUnreadIndicatorAnimated`.

michaelkirk:
> Just search for indexPathOfUnreadMessagesIndicator. It's used in scrollToDefaultPosition and scrollToUnreadIndicatorAnimated.

Sorry - I should have been clearer. I am aware of when this method is called, but in my testing it always hit the early return before getting to the code in this diff, seemingly even when there was an unreadIndicator. I'll try to reproduce.

charlesmchen:
> Sorry - I should have been clearer. I am aware of when this method is called, but in my testing it always hit the early return before getting to the code in this diff, seemingly even when there was an unreadIndicator. I'll try to reproduce.

Well that ain't great.  That "early exit" check I recently added may have a bug.

charlesmchen:
> Well that ain't great. That "early exit" check I recently added may have a bug.

@michaelkirk this works fine for me.  I just passed the early exit in group and 1:1 conversations with unread messages.

PR #2466, opened 2017-08-23
Mkirk/more name

michaelkirk:
PTAL @charlesmchen 

We might end up revisiting this per discussed design shortcomings.

### DIFF
--- a/Signal/src/ViewControllers/InboxTableViewCell.m
+++ b/Signal/src/ViewControllers/InboxTableViewCell.m
@@ -163,7 +163,7 @@ const NSUInteger kAvatarViewDiameter = 52;
         NSString *contactIdentifier = thread.contactIdentifier;
         isBlocked = [blockedPhoneNumberSet containsObject:contactIdentifier];
     }
-    
+
     self.threadId = thread.uniqueId;
     NSMutableAttributedString *snippetText = [NSMutableAttributedString new];
     if (isBlocked) {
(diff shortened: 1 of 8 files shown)


Signal/src/ViewControllers/InboxTableViewCell.m
```diff
@@ -271,9 +262,78 @@ - (NSAttributedString *)dateAttributedString:(NSDate *)date {
...
+        } else {
+            name = [[NSAttributedString alloc] initWithString:thread.name];
+        }
+    } else {
+        name = [contactsManager attributedStringForConversationTitleWithPhoneIdentifier:thread.contactIdentifier
+                                                                            primaryFont:self.nameLabel.font
+                                                                          secondaryFont:[UIFont ows_footnoteFont]];
+    }
+    
+    self.nameLabel.attributedText = name;
```
charlesmchen: Do we need to call `setNeedsLayout` here?

Signal/src/ViewControllers/OWSConversationSettingsViewController.m
```diff
@@ -1124,6 +1133,19 @@ - (void)identityStateDidChange:(NSNotification *)notification
...
 
+- (void)otherUsersProfileDidChange:(NSNotification *)notification
+{
+    OWSAssert([NSThread isMainThread]);
+
+    NSString *recipientId = notification.userInfo[kNSNotificationKey_ProfileRecipientId];
+    OWSAssert(recipientId.length > 0);
+
+    if (recipientId.length > 0 && [self.thread isKindOfClass:[TSContactThread class]] &&
+        [self.thread.contactIdentifier isEqualToString:recipientId]) {
```
charlesmchen: If we're going to do this a lot, we could create a category on `NSNotification+Profiles` with a `appliesToGroup:`, `appliesToUser:` and/or `appliesToThread:` methods.

PR #2470, opened 2017-08-24
Show an error when profile name is too long.

charlesmchen:
PTAL @mi


charlesmchen:
> We may want to revisit the UI, e.g. update the field with an "error" state and print the message in line vs. an alert, but this is functional.

Yup, I went with the "simplest thing that could work".  It's not great but I held off on doing something more complicated because screen real estate is at a premium and I'd really prefer for the "save" button to not even be partially hidden by the keyboard if possible, and it's already close on EN (other languages are more verbose) using the built-in keyboard (other keyboards can be taller).

PR #2471, opened 2017-08-24
profile name for avatar

michaelkirk:
If the contact is not in your address book, and they don't have a profile avatar, but they do have a profile name, use that name to generate the default (initials) avatar.

This was only a little trickier than anticipated due to the way we cache avatars, also the current solution has a race condition. Because clearing the avatarCache *and* redrawing the table cell are driven by the same `kNSNotificationName_OtherUsersProfileDidChange` notification, it's possible the cell will be redrawn before the cache is cleared. *or* we could have *another* notification to redraw the contact cell whenever the avatar cache clears, but it seems like a wasteful road to go down. WDYT @charlesmchen?

PTAL @charlesmchen


charlesmchen:
> Because clearing the avatarCache and redrawing the table cell are driven by the same kNSNotificationName_OtherUsersProfileDidChange notification

What if we fire two notifications "profile will change" and "profile did change"?  We clear the avatar cache using the former and we update the cell using the latter.

Signal/src/Models/OWSContactAvatarBuilder.m
```diff
@@ -48,7 +49,14 @@ - (instancetype)initWithSignalId:(NSString *)signalId
                         diameter:(NSUInteger)diameter
                  contactsManager:(OWSContactsManager *)contactsManager
 {
-    NSString *name = [contactsManager displayNameForPhoneIdentifier:signalId];
+    // Name for avatar initialis.
```
charlesmchen: s/initialis/initials.

Signal/src/util/ImageCache.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+/**
+ * A two dimensional hash, allowing you to store variations under a single key.
+ * This is useful because we generate multiple diameters of an image, but when we 
+ * want to clear out the images for a key we want to clear out *all* variations.
+ */
+@objc
+class ImageCache: NSObject {
```
charlesmchen: I'd love to eventually a) write avatars to disk and b) evacuate the cache only when its contents change (E.g. if bob's avatar in device contacts change, only evacuate his avatar, not Alice's - and if alice's profile avatar changed _not_ evacuate her avatar if we're using an avatar from our device contacts for her).

Signal/src/Models/OWSContactAvatarBuilder.m
```diff
@@ -48,7 +49,14 @@ - (instancetype)initWithSignalId:(NSString *)signalId
...
+    NSString *_Nullable name = [contactsManager nameFromSystemContactsForRecipientId:signalId];
+    if (name.length == 0) {
+        name = [contactsManager profileNameForRecipientId:signalId];
+    }
+    if (name.length == 0) {
+        name = signalId;
```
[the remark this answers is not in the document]
michaelkirk: This will generate an avatar of "#"


> IMO we should change that to be a user icon.

Like a generic user icon you mean? We talked about this once months ago, but opted for consistency with Signal-Desktop/Android.

PR #2473, opened 2017-08-25
profiles vs. censorship circumvention

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/Profiles/OWSProfileManager.m
+++ b/Signal/src/Profiles/OWSProfileManager.m
@@ -195,7 +195,7 @@ const NSUInteger kOWSProfileManager_MaxAvatarDiameter = 640;
 
 - (AFHTTPSessionManager *)avatarHTTPManager
 {
-    return [OWSSignalService sharedInstance].cdnSessionManager;
+    return [OWSSignalService sharedInstance].CDNSessionManager;
 }
 
 #pragma mark - User Profile Accessor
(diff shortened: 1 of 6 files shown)


michaelkirk:
Actually - I want to verify something.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -564,22 +564,8 @@ - (void)uploadAvatarToService:(NSData *)avatarData
...
                     }
                     success:^(NSURLSessionDataTask *_Nonnull uploadTask, id _Nullable responseObject) {
-                        OWSAssert([uploadTask.response isKindOfClass:[NSHTTPURLResponse class]]);
-                        NSHTTPURLResponse *response = (NSHTTPURLResponse *)uploadTask.response;
-
-                        // We could also construct this URL locally from manager.baseUrl + formKey
-                        // but the approach of getting it from the remote provider seems a more
-                        // robust way to ensure we've actually created the resource where we
-                        // think we have.
-                        NSString *avatarUrlPath = response.allHeaderFields[@"Location"];
```
michaelkirk: We only store the object key, and not the fully qualified url, since the host can change per censorship circumvention.

SignalServiceKit/src/TSConstants.h
```diff
@@ -24,15 +24,21 @@ typedef enum { kSMSVerification, kPhoneNumberVerification } VerificationTranspor
...
+#define textSecureCDNReflectorHost @"signal-cdn-reflector.appspot.com"
 
 #else
 
+// Staging
 #define textSecureWebSocketAPI @"wss://textsecure-service-staging.whispersystems.org/v1/websocket/"
 #define textSecureServerURL @"https://textsecure-service-staging.whispersystems.org/"
 #define textSecureCDNServerURL @"https://cdn-staging.signal.org"
+#define textSecureServiceReflectorHost @"meek-signal-service-staging.appspot.com";
+#define textSecureCDNReflectorHost @"meek-signal-cdn-staging.appspot.com";
```
michaelkirk: note that because I set up the staging reflectors, they might not behave exactly the same as the production ones. I'll follow up with moxie to see if we can get them more in sync.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -564,22 +564,8 @@ - (void)uploadAvatarToService:(NSData *)avatarData
...
-                        if (avatarUrlPath.length == 0) {
-                            OWSProdFail(@"profile_manager_error_avatar_upload_no_location_in_response");
```
charlesmchen: I know you eliminated this prod fail, but please use `SignalServiceKit/Utilities/extract_analytics_event_names.py` to convert analytics keys into code-generated analytics event name methods.

michaelkirk: Done!

michaelkirk:
in https://github.com/WhisperSystems/Signal-iOS/commit/5ec9b40f101cc5afce99609df98f6e361f6e53fd

PR #2474, opened 2017-08-25
Charlesmchen/message view tweaks

charlesmchen:
* Unbound message bubble cache size and DRY up cache logic. 
* Remove obsolete "scrollLaterTimer" hack in messages view.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/Models/OWSMessagesBubblesSizeCalculator.m
+++ b/Signal/src/Models/OWSMessagesBubblesSizeCalculator.m
@@ -56,6 +56,10 @@ NS_ASSUME_NONNULL_BEGIN
         _referenceSystemMessageCell = [OWSSystemMessageCell new];
         _referenceUnreadIndicatorCell = [OWSUnreadIndicatorCell new];
         _referenceContactOffersCell = [OWSContactOffersCell new];
+
+        // Calculating message size is relatively expensive, so unbound the size of the cache.
+        self.cache.countLimit = 0;
+        self.cache.totalCostLimit = 0;
     }
     return self;
 }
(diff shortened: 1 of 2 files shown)


Signal/src/Models/OWSMessagesBubblesSizeCalculator.m
```diff
@@ -75,28 +79,40 @@ - (CGSize)messageBubbleSizeForMessageData:(id<JSQMessageData>)messageData
                               atIndexPath:(NSIndexPath *)indexPath
                                withLayout:(JSQMessagesCollectionViewFlowLayout *)layout
 {
+    id cacheKey = [self cacheKeyForMessageData:messageData];
```
charlesmchen: DRY.  The base class will do some redundant work updating the same cache for incoming and outgoing messages, but this is harmless and cheap.

Signal/src/Models/OWSMessagesBubblesSizeCalculator.m
```diff
@@ -56,6 +56,10 @@ - (instancetype)init
         _referenceSystemMessageCell = [OWSSystemMessageCell new];
         _referenceUnreadIndicatorCell = [OWSUnreadIndicatorCell new];
         _referenceContactOffersCell = [OWSContactOffersCell new];
+
+        // Calculating message size is relatively expensive, so unbound the size of the cache.
+        self.cache.countLimit = 0;
+        self.cache.totalCostLimit = 0;
```
charlesmchen: I don't foresee this "hash-to-NSValues containing CGSizes" cache becoming that large in memory and its a huge perf hotspot.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -228,7 +228,6 @@ @interface MessagesViewController () <AVAudioPlayerDelegate,
 @property (nonatomic) NSCache *messageAdapterCache;
 @property (nonatomic) BOOL userHasScrolled;
 @property (nonatomic) NSDate *lastMessageSentDate;
-@property (nonatomic) NSTimer *scrollLaterTimer;
```
charlesmchen: This was an earlier hack to work around https://github.com/WhisperSystems/JSQMessagesViewController/pull/17.  Now that we've addressed the root cause, this timer can go away!

Signal/src/Models/OWSMessagesBubblesSizeCalculator.m
michaelkirk: Nice.

PR #2475, opened 2017-08-28
Clear the local profile avatar immediately when we request upload form.

charlesmchen:
PTAL @michaelkirk


Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -580,6 +605,10 @@ - (void)uploadAvatarToService:(NSData *)avatarData
                     }];
             }
             failure:^(NSURLSessionDataTask *task, NSError *error) {
+                if (task.response != nil) {
```
michaelkirk: What is implied by this if statement? Can you leave a comment?

charlesmchen: Done. Please check out the comment.

PR #2477, opened 2017-08-28
profile key sync

michaelkirk:
Syncing both our own and any contact profile keys we know about.

More comments inline.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/MultiDeviceProfileKeyUpdateJob.swift
+++ b/Signal/src/MultiDeviceProfileKeyUpdateJob.swift
@@ -1,0 +1,55 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+import PromiseKit
+
+@objc class MultiDeviceProfileKeyUpdateJob: NSObject {
+
+    let TAG = "[MultiDeviceProfileKeyUpdateJob]"
+
+    let profileKey: OWSAES256Key
+    let identityManager: OWSIdentityManager
+    let messageSender: MessageSender
+    let profileManager: OWSProfileManager
+
+    required init(profileKey: OWSAES256Key, identityManager: OWSIdentityManager, messageSender: MessageSender, profileManager: OWSProfileManager) {
+        self.profileKey = profileKey
+
+        self.identityManager = identityManager
+        self.messageSender = messageSender
+        self.profileManager = profileManager
+    }
+
+    class func run(profileKey: OWSAES256Key, identityManager: OWSIdentityManager, messageSender: MessageSender, profileManager: OWSProfileManager) {
+        return self.init(profileKey: profileKey, identityManager: identityManager, messageSender: messageSender, profileManager: profileManager).run()
+    }
+
+    func run(retryDelay: TimeInterval = 1) {
+        guard let localNumber = TSAccountManager.localNumber() else {
+            owsFail("\(self.TAG) localNumber was unexpectedly nil")
+            return
+        }
+
+        let localSignalAccount = SignalAccount(recipientId: localNumber)
+        localSignalAccount.contact = Contact()
+        let syncContactsMessage = OWSSyncContactsMessage(signalAccounts: [localSignalAccount],
+                                                        identityManager: self.identityManager,
+                                                        profileManager: self.profileManager)
+
+        self.messageSender.sendTemporaryAttachmentData(syncContactsMessage.buildPlainTextAttachmentData(),
+                                                       contentType: OWSMimeTypeApplicationOctetStream,
+                                                       in: syncContactsMessage,
+                                                       success: {
+                                                        Logger.info("\(self.TAG) Successfully synced profile key")
+
+        },
+                                                       failure: { error in
+                                                        Logger.error("\(self.TAG) in \(#function) failed with error: \(error) retrying in \(retryDelay)s.")
+                                                        after(seconds: retryDelay).then {
+                                                            self.run(retryDelay: retryDelay * 2)
+                                                            }.retainUntilComplete()
+        })
+    }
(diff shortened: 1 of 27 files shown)


Signal/src/MultiDeviceProfileKeyUpdateJob.swift
```diff
@@ -0,0 +1,58 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+import PromiseKit
+
+/**
```
michaelkirk: Following Signal-Android's lead - instead of attaching our profile key to *every* sync message, we send it explicitly when updating our profile.

This is less fool-proof, but it did seem like overkill to transmit an extra 30-some bytes on every sync message, when we ultimately only need to transmit it once. Especially given how chatty sync messages are.

SignalServiceKit/protobuf/OWSSignalServiceProtos.proto
```diff
@@ -37,35 +37,18 @@ message Content {
   optional NullMessage nullMessage = 4;
 }
```
michaelkirk: Most of the changes in the proto files were in an effort to minimize the diff with upstream (signal-android)

SignalServiceKit/protobuf/OWSSignalServiceProtos.proto
```diff
@@ -143,15 +146,14 @@ message SyncMessage {
...
 
-  optional Sent     sent       = 1;
-  optional Contacts contacts   = 2;
-  optional Groups   groups     = 3;
-  optional Request  request    = 4;
-  repeated Read     read       = 5;
-  optional Blocked  blocked    = 6;
-  optional Verified verified   = 7;
-  optional bytes    padding    = 8;
-  optional bytes    profileKey = 9;
```
michaelkirk: Most of this block is reformatting to align with Android's protos, but I did also remove the profileKey field from SyncMessages.

Signal/src/MultiDeviceProfileKeyUpdateJob.swift
charlesmchen: Hmm, but we could link to desktop at any time and years could pass after linking before users update their profile again.  What if we also include it in the "bulk contact sync" messages?

michaelkirk: Profile key is sent as part of provisioning. This job just exists for legacy clients.

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -236,6 +242,11 @@ - (void)saveUserProfile:(UserProfile *)userProfile
 
         dispatch_async(dispatch_get_main_queue(), ^{
             if (isLocalUserProfile) {
+                [MultiDeviceProfileKeyUpdateJob runWithProfileKey:userProfile.profileKey
```
charlesmchen: Hmm. `saveUserProfile` isn't called by `localUserProfile` when `localUserProfile` generates the local profile for the first time.  

Nor should it, or this line above would become unsafe:

```
 BOOL isLocalUserProfile = userProfile == self.localUserProfile;
```
So for existing users who upgrade to a "post-profile world" for the first time, their profile key _won't_ be sync to desktop when it is generated.  

I guess this is okay now since only your primary can edit your profile but eventually won't we want to let desktop edit your profile?  It seems to me like we should sync this sooner if we can.

Signal/src/MultiDeviceProfileKeyUpdateJob.swift
charlesmchen: Gotcha thanks.

Signal/src/Profiles/OWSProfileManager.m
michaelkirk: We sync the profile key as soon as the user submits profile data. The value of distributing a profile key not yet associated with any profile data seems pretty negligible.

charlesmchen: If desktop wants to ship a profile editing, it won't be able to if primary doesn't share the profile key until it has edited the profile.  e.g. "Desktop can't edit profile until you've edited profile on primary".  

I'm trying to think about the simplest thing we could do.  It might be to replace this:

```
            if (_localUserProfile == nil) {
                DDLogInfo(@"%@ Building local profile.", self.tag);
                _localUserProfile = [[UserProfile alloc] initWithRecipientId:kLocalProfileUniqueId];
                _localUserProfile.profileKey = [OWSAES256Key generateRandomKey];
                [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
                    [_localUserProfile saveWithTransaction:transaction];
                }];
            }
```

with this:
```
            if (_localUserProfile == nil) {
                DDLogInfo(@"%@ Building local profile.", self.tag);
                _localUserProfile = [[UserProfile alloc] initWithRecipientId:kLocalProfileUniqueId];
                _localUserProfile.profileKey = [OWSAES256Key generateRandomKey];
                [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
                    [_localUserProfile saveWithTransaction:transaction];
                }];
                
                    @synchronized(self)
                    {
                        [self saveUserProfile:_localUserProfile];
                    }
                });
            }
```

michaelkirk: > If desktop wants to ship a profile editing, it won't be able to if primary doesn't share the profile key until it has edited the profile. e.g. "Desktop can't edit profile until you've edited profile on primary".

Ok - in case that's part of the initial release, I'll make a PR with this change.

PR #2479, opened 2017-08-28
profile key message

michaelkirk:
1. Send an empty message upon whitelisting someone so they get our key right away.
2. Handle said messages.
3. Do not create a text message if there is no message body.

At present (1.) is implemented but disabled behind a feature flag so that we can let handling these messages roll out first.

(3.) exists so that we don't need to do this staged release dance in the future. 

PTAL @charlesmchen

### DIFF
--- a/Signal/translations/en.lproj/Localizable.strings
+++ b/Signal/translations/en.lproj/Localizable.strings
@@ -1417,6 +1417,9 @@
 /* action sheet item */
 "SHARE_ACTION_TWEET" = "Twitter";
 
+/* alert title */
+"SHARE_PROFILE_FAILED_TITLE" = "Failed to Share Your Profile";
+
 /* Action sheet item */
 "SHOW_SAFETY_NUMBER_ACTION" = "Show New Safety Number";
 
--- a/Signal/translations/en.lproj/Localizable.strings
+++ b/Signal/translations/en.lproj/Localizable.strings
@@ -1417,9 +1417,6 @@
 /* action sheet item */
 "SHARE_ACTION_TWEET" = "Twitter";
 
-/* alert title */
-"SHARE_PROFILE_FAILED_TITLE" = "Failed to Share Your Profile";
-
 /* Action sheet item */
 "SHOW_SAFETY_NUMBER_ACTION" = "Show New Safety Number";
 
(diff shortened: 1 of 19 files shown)


charlesmchen:
> At present (1.) is implemented but disabled behind a feature flag so that we can let handling these messages roll out first.

What's our trigger to flip this flag?  Should we add a card for our quarterly planning?

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1417,6 +1417,9 @@
 /* action sheet item */
 "SHARE_ACTION_TWEET" = "Twitter";
 
+/* alert title */
+"SHARE_PROFILE_FAILED_TITLE" = "Failed to Share Your Profile";
```
charlesmchen: I wonder if we should bother the user with this.  We send our profile key with every message, right?  So the other user will see your profile as soon as you have any contact with them. "Don't make me think."

SignalServiceKit/src/Messages/TSMessagesManager.m
```diff
@@ -828,6 +831,29 @@ - (void)handleExpirationTimerUpdateMessageWithEnvelope:(OWSSignalServiceProtosEn
...
+    NSString *recipientId = incomingEnvelope.source;
+    if (!dataMessage.hasProfileKey) {
+        OWSFail(@"%@ recevied profile key message without profile key from: %@",
+            self.tag,
+            envelopeAddress(incomingEnvelope));
+        return;
+    }
+    NSData *profileKey = dataMessage.profileKey;
+    if (profileKey.length != kAES256_KeyByteLength) {
+        OWSFail(@"%@ recevied profile key of unexpected length:%lu from:%@",
```
charlesmchen: s/recevied/received/

charlesmchen: And elsewhere in this PR.

michaelkirk:
>> At present (1.) is implemented but disabled behind a feature flag so that we can let handling these messages roll out first.
> What's our trigger to flip this flag? Should we add a card for our quarterly planning?

I don't think there's anything to discuss that would warrant bringing it up in a meeting. Once we're confident the number of affected clients is sufficiently small (~2 weeks) we can flip the switch. I'll put a card in to make sure it's not forgotten.

Signal/translations/en.lproj/Localizable.strings
michaelkirk: To be clear, are you suggesting we get rid of the retry logic as well as the alert?

SignalServiceKit/src/Messages/TSMessagesManager.m
michaelkirk: done!

PR #2481, opened 2017-08-29
infer when group has been whitelisted on linked device

michaelkirk:
Since clients only include the `profileKey` for synced group messages when that group is whitelisted, it's presence is sufficient to deduce the group is whitelisted.

PTAL @charlesmchen


SignalServiceKit/src/Protocols/ProfileManagerProtocol.h
```diff
@@ -16,5 +16,6 @@
 - (BOOL)isThreadInProfileWhitelist:(TSThread *)thread;
 
 - (void)addUserToProfileWhitelist:(NSString *)recipientId;
+- (void)addGroupIdToProfileWhitelist:(NSData *)groupId;
```
charlesmchen: This is fine but we could also use `addThreadToProfileWhitelist:` if its more convenient/concise.

PR #2482, opened 2017-08-29
sync local profile as soon as it's created

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/Profiles/OWSProfileManager.m
+++ b/Signal/src/Profiles/OWSProfileManager.m
@@ -287,6 +287,13 @@ const NSUInteger kOWSProfileManager_MaxAvatarDiameter = 640;
                 [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
                     [_localUserProfile saveWithTransaction:transaction];
                 }];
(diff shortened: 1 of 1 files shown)


charlesmchen:
Actually it occurs to me that the problem is worse.  Now that we're not sending our profile key with every sync message, if iOS ships its desktop support before desktop (or users upgrade iOS before desktop), we again are in a situation where desktop doesn't know our profile key and has no way to request it.  If user never edits their profile again on iOS, we won't sync it.  I wonder if we should run your "sync profile key" job every time we upgrade or something.

michaelkirk:
> Actually it occurs to me that the problem is worse. Now that we're not sending our profile key with every sync message, if iOS ships its desktop support before desktop (or users upgrade iOS before desktop), we again are in a situation where desktop doesn't know our profile key and has no way to request it. If user never edits their profile again on iOS, we won't sync it. I wonder if we should run your "sync profile key" job every time we upgrade or something.

Admittedly it's not a 100% solution, but it seems like it would work reasonably soon, and the affected group will be diminishing.

Desktop could also grab it off any synced transcript to a whitelisted member if we want to be more comprehensive.

PR #2483, opened 2017-08-29
Fix glitches in the scroll state in message view.

charlesmchen:
* Revisit the [UIScrollView _adjustContentOffsetIfNecessary] issue. 
  * See: https://github.com/WhisperSystems/JSQMessagesViewController/pull/18/files
* Fix glitches in the initial scroll state in message view. 
  * We can't (and don't need to) determine whether we're "scrolled to the bottom" until the collection view has determined its content size.
* Don't reset scroll state when returning to the message view.
   * We only want to set the scroll state the first time we enter the view.

PTAL @michaelkirk


Podfile.lock
```diff
@@ -171,6 +167,6 @@ SPEC CHECKSUMS:
   UnionFind: c33be5adb12983981d6e827ea94fc7f9e370f52d
   YapDatabase: cd911121580ff16675f65ad742a9eb0ab4d9e266
 
-PODFILE CHECKSUM: 2f847bb25e70d1d376f38cf21ae08624fa6ed67d
+PODFILE CHECKSUM: 4a6db311736b5ab36b660af813dcdf0af62da48c
 
-COCOAPODS: 1.2.1
```
michaelkirk: Can you please also update the Gemfile.lock when you update cocoapods? That way you and me and travis have half a chance of being on the same build tools.

charlesmchen: I've downgraded cocoapods locally.

PR #2484, opened 2017-08-29
Re-sync local profile state with service if necessary.

charlesmchen:
PTAL @michaelkirk


Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -673,6 +674,17 @@ - (void)updateServiceWithProfileName:(nullable NSString *)localProfileName
     });
 }
 
+- (void)syncLocalProfile
```
michaelkirk: nit: `fetchLocalUsersProfile` or `pullLocalProfile` might better indicate the direction of data flow here, vs sync which could be either push or pull.

charlesmchen: Good point!

PR #2490, opened 2017-08-30
Remain scrolled to bottom after presenting the keyboard in messages v…

charlesmchen:
…iew.

PTAL @michaelkirk


charlesmchen:
> For my understanding, are these changes just intended to affect users with custom keyboards? E.g. users with the default keyboard will see no change in behavior from this?

This will benefit all users but it was a little bit of extra work to handle custom keyboards properly.

michaelkirk:
>> For my understanding, are these changes just intended to affect users with custom keyboards? E.g. users with the default keyboard will see no change in behavior from this?
>
> This will benefit all users but it was a little bit of extra work to handle custom keyboards properly.

Then I guess my question is why does this happen sometimes but not always. If there's a solution that doesn't involve adding more notification listeners I'd be happy.

PR #2491, opened 2017-08-30
Refine the "is scrolled to bottom" logic to better handle new convers…

charlesmchen:
…ations.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -3579,9 +3579,27 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
         return NO;
     }
 
+    // This is a bit subtle.
+    //
+    // The _wrong_ way to determine if we're scrolled to the bottom is to
+    // measure whether the collection view's content is "near" the bottom edge
+    // of the collection view.  This is wrong because the collection view
+    // might not have enough content to fill the collection view's bounds
+    // _under certain conditions_ (e.g. with the keyboard dismissed).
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -3521,9 +3521,27 @@ - (BOOL)isScrolledToBottom
...
+    //
+    // To determine that, we find the appropriate "content offset y" if
+    // the scroll view were scrolled down as far as possible.  IFF the
+    // actual "content offset y" is "near" that value, we return YES.
     const CGFloat kIsAtBottomTolerancePts = 5;
-    return (self.collectionView.contentOffset.y + self.collectionView.bounds.size.height + kIsAtBottomTolerancePts
-        >= self.collectionView.contentSize.height);
+    // Note the usage of MAX() to handle the case where there isn't enough
+    // content to fill the collection view at its current size.
+    CGFloat contentOffsetYBottom
```
michaelkirk: Would it be equally correct to leave the existing code, but add a guard like this before it?

```
if (self.collectionView.contentSize.height <= self.collectionView.bounds.size.height + kIsAtBottomTolerancePts) {
    return YES;
}
```

IMO it'd be a little easier to follow, and might obviate the need for the verbose comment.

charlesmchen: I think so but it will be significantly harder to test this condition, since its easy to change scroll state in a fine-grained way but incredibly difficult to manipulate content size in the same way.  I think I'll leave this as-is.

PR #2492, opened 2017-08-31
fix production crash on addObject, presumably it's nil.

michaelkirk:
The root cause of this crash is still unclear, but this will prevent the proximate cause, and give us additional diagnostic data in the cases where it was happening. 

IMO, as a rule we should not be writing code which blindly casts.

PTAL @charlesmchen


michaelkirk:
Soon to be superseded.

PR #2493, opened 2017-08-31
Mkirk/prod crashes

michaelkirk:
A handful of small crash fixes.

Each commit fixes a different crash, see the individual commits for comments.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/CallViewController.swift
+++ b/Signal/src/ViewControllers/CallViewController.swift
@@ -25,6 +25,7 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver, 
 
     // MARK: Views
 
+    var isViewLoaded = false
     var hasConstraints = false
     var blurView: UIVisualEffectView!
     var dateFormatter: DateFormatter?
@@ -158,7 +159,9 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver, 
     }
 
     func didBecomeActive() {
-        shouldRemoteVideoControlsBeHidden = false
+        if (self.isViewLoaded) {
+            shouldRemoteVideoControlsBeHidden = false
+        }
     }
 
     // MARK: View Lifecycle
@@ -195,6 +198,7 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver, 
         call.addObserverAndSyncState(observer: self)
 
         Environment.getCurrent().callService.addObserverAndSyncState(observer: self)
+        self.isViewLoaded = true
     }
 
     // MARK: - Create Views
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -3952,20 +3952,30 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
 
     TSThread *thread = self.thread;
     uint64_t lastVisibleTimestamp = self.lastVisibleTimestamp;
+
     [self.editingDatabaseConnection asyncReadWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
         NSMutableArray<id<OWSReadTracking>> *interactions = [NSMutableArray new];
+        
         [[TSDatabaseView unseenDatabaseViewExtension:transaction]
             enumerateRowsInGroup:thread.uniqueId
                       usingBlock:^(
                           NSString *collection, NSString *key, id object, id metadata, NSUInteger index, BOOL *stop) {
 
-                          TSInteraction *interaction = object;
-                          if (interaction.timestampForSorting > lastVisibleTimestamp) {
+                          if (![object conformsToProtocol:@protocol(OWSReadTracking)]) {
+                              OWSFail(@"Expected to conform to OWSReadTracking: object with class: %@ collection: %@ "
+                                      @"key: %@",
+                                  [object class],
+                                  collection,
+                                  key);
+                              return;
+                          }
+                          id<OWSReadTracking> possiblyRead = (id<OWSReadTracking>)object;
+
+                          if (possiblyRead.timestampForSorting > lastVisibleTimestamp) {
                               *stop = YES;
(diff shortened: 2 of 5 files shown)


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -2140,33 +2149,38 @@ - (void)collectionView:(JSQMessagesCollectionView *)collectionView
...
                     if (tappedImage == nil) {
                         DDLogWarn(@"tapped TSPhotoAdapter with nil image");
-                    } else {
-                        UIWindow *window = [UIApplication sharedApplication].keyWindow;
-                        JSQMessagesCollectionViewCell *cell
-                            = (JSQMessagesCollectionViewCell *)[collectionView cellForItemAtIndexPath:indexPath];
-                        OWSAssert([cell isKindOfClass:[JSQMessagesCollectionViewCell class]]);
-                        CGRect convertedRect = [cell.mediaView convertRect:cell.mediaView.bounds toView:window];
+                        return;
+                    }
```
michaelkirk: The rest of this block is just whitespace changes due to using an early return instead of if/else.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -3876,20 +3890,30 @@ - (void)markVisibleMessagesAsRead
...
         NSMutableArray<id<OWSReadTracking>> *interactions = [NSMutableArray new];
+        
         [[TSDatabaseView unseenDatabaseViewExtension:transaction]
             enumerateRowsInGroup:thread.uniqueId
                       usingBlock:^(
                           NSString *collection, NSString *key, id object, id metadata, NSUInteger index, BOOL *stop) {
 
-                          TSInteraction *interaction = object;
-                          if (interaction.timestampForSorting > lastVisibleTimestamp) {
+                          if (![object conformsToProtocol:@protocol(OWSReadTracking)]) {
```
michaelkirk: We should avoid doing blind casts.

Signal/src/ViewControllers/CallViewController.swift
```diff
@@ -158,7 +159,9 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver,
     }
 
     func didBecomeActive() {
-        shouldRemoteVideoControlsBeHidden = false
+        if (self.isViewLoaded) {
+            shouldRemoteVideoControlsBeHidden = false
```
charlesmchen: We can become active before we load?  That's very unexpected.  Are you sure?

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -485,8 +485,17 @@ - (void)configureForThread:(TSThread *)thread
...
-        [[YapDatabaseViewMappings alloc] initWithGroups:@[ thread.uniqueId ] view:TSMessageDatabaseViewExtensionName];
+
+    if (thread.uniqueId.length > 0) {
+        self.messageMappings = [[YapDatabaseViewMappings alloc] initWithGroups:@[ thread.uniqueId ]
+                                                                          view:TSMessageDatabaseViewExtensionName];
+    } else {
+        OWSFail(@"uniqueId unexpectedly empty for thread: %@", thread);
+        self.messageMappings =
+            [[YapDatabaseViewMappings alloc] initWithGroups:@[] view:TSMessageDatabaseViewExtensionName];
+        return;
```
charlesmchen: This is happening in production!?  That is, someone is somehow navigating into conversation view for an unsaved thread?

charlesmchen: Maybe a better fix is to save the thread if its unsaved (e.g. has not uniqueId)?

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
charlesmchen: So does that mean that in prod there's something in the `unseenDatabaseViewExtension` that doesn't conform to `OWSReadTracking`?

Signal/src/ViewControllers/CallViewController.swift
charlesmchen: Okay, this makes sense.  Sorry!

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
charlesmchen: I'm definitely leaning towards the "save the thread" approach - want to discuss? There's other places in this class that assume the thread is saved / has a uniqueId.

Signal/src/ViewControllers/CallViewController.swift
michaelkirk: It was my most likely interpretation for accessing a nil view which is created in viewDidLoad.

If you'd like to look into it yourself, it's our top crash.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
michaelkirk: > This is happening in production!?

Not often, but yes. 

> That is, someone is somehow navigating into conversation view for an unsaved thread?

No, I don't think that's the cause, since thread id's are assigned in init, not on save. Honestly not sure what the issue is, but hopefully the new logging will make it clearer if people find themselves in this state.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
michaelkirk: The actual crash was a few lines down:

```
[interactions addObject:possiblyRead];
```

Presumably `possiblyRead` is nil - the root cause of that I don't know.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
charlesmchen: I wonder if thread itself is nil?

michaelkirk:
In 7f5975f0

PR #2495, opened 2017-08-31
Charlesmchen/layout glitches

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/JSQMessagesViewController/pull/19


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -404,7 +406,7 @@ - (void)keyboardDidShow:(NSNotification *)notification
     // We only want to update scroll state for non-empty show events.
     BOOL isPresenting = endValue.size.height > 0.f;
 
-    if (isPresenting) {
+    if (isPresenting && self.wasScrolledToBottomBeforeKeyboardShow) {
```
charlesmchen: Small fix for change done earlier this week.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -543,6 +545,8 @@ - (void)hideInputIfNeeded
 
 - (void)viewDidLoad
 {
+    self.collectionView.layoutDelegate = self;
```
michaelkirk: nit: Rather than further diverging JSQMVC with this delegate protocol only to call back to delegate methods implemented in our own class, it seems to me like it would be simpler to skip all that and just override `setFrame` and `setBounds` in our own class.

Is there a reason that you did it this way?

charlesmchen: Yes, we need to override `setFrame` and `setBounds` - but they hang on `JSQMessagesCollectionView`, not `JSQMessagesViewController`, so we can't override them in `MessagesViewController`.  We're not using a custom subclass of `JSQMessagesCollectionView`.  I suppose we could, but it'd be a bunch of work and we're already diverging from JSQ, and JSQ is defunct, so...

michaelkirk: Whoops, my mistake. Thanks for the clarification.

PR #2497, opened 2017-09-01
Revamp avatar cropper to show mask

michaelkirk:
- black masking style for avatar cropper
- cancel/done buttons
- no navbar for modal edit
- vertically center image

PTAL @charlesmchen 

## portrait

avatar3

## landscape

avatar1

## zoomed in

avatar2

### DIFF
--- a/Signal/src/ViewControllers/CropScaleImageViewController.swift
+++ b/Signal/src/ViewControllers/CropScaleImageViewController.swift
@@ -204,13 +204,16 @@ class CropScaleImageViewController: OWSViewController {
         let maskingView = OWSBezierPathView()
         contentView.addSubview(maskingView)
 
-        maskingView.configureShapeLayerBlock = { layer, bounds in
+        maskingView.configureShapeLayerBlock = { [weak self] layer, bounds in
+            guard let strongSelf = self else {
+                return
+            }
             let path = UIBezierPath(rect: bounds)
 
-            let radius = min(bounds.size.width, bounds.size.height) * 0.5 - self.maskMargin
-            // Center the circle's bounding rectangle
-            let circleRect = CGRect(x: bounds.size.width * 0.5 - radius, y: bounds.size.height * 0.5 - radius, width: radius * 2, height: radius * 2)
+            let circleRect = strongSelf.cropFrame(forBounds:bounds)
+            let radius = circleRect.size.width * 0.5
             let circlePath = UIBezierPath(roundedRect: circleRect, cornerRadius: radius)
+
             path.append(circlePath)
             path.usesEvenOddFillRule = true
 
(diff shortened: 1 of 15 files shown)


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -389,7 +389,7 @@
 "CREATE_NEW_GROUP" = "Create new group";
 
 /* Title for the 'crop/scale image' dialog. */
-"CROP_SCALE_IMAGE_VIEW_TITLE" = "Crop Image";
+"CROP_SCALE_IMAGE_VIEW_TITLE" = "Crop and Scale";
```
charlesmchen: I don't think most users know what "Scale" means.  I prefer the old title.

Signal/src/ViewControllers/CropScaleImageViewController.swift
```diff
@@ -294,34 +307,35 @@ class CropScaleImageViewController: OWSViewController {
...
-        let imageViewFrame = imageRenderRect(forDstSize:viewSizePoints)
+        var imageViewFrame = imageRenderRect(forDstSize: viewSizePoints)
+
+        // offset to vertically center image in view (aspect ratio?)
+        let srcToViewRatio = viewSizePoints.width / srcCropSizePoints.width / imageScale
+        let heightOfImageLayer = srcDefaultCropSizePoints.height * srcToViewRatio
+        let yOffset = imageView.frame.height * 0.5 - heightOfImageLayer * 0.5
+        imageViewFrame = imageViewFrame.offsetBy(dx: 0, dy: yOffset)
+
+        // inset frame by the with of the masking margin so the user can pan to the very edge of the image
```
charlesmchen: s/with/width/

Signal/translations/en.lproj/Localizable.strings
michaelkirk: Both WhatsApp and iOS System contacts editing go with "Move and Scale" so I'm actually now inclined to go this third way.

Signal/src/ViewControllers/CropScaleImageViewController.swift
```diff
@@ -174,52 +175,64 @@ class CropScaleImageViewController: OWSViewController {
...
+            let circlePath = UIBezierPath(roundedRect: circleRect, cornerRadius: radius)
+            path.append(circlePath)
+            path.usesEvenOddFillRule = true
+
+            layer.path = path.cgPath
+            layer.fillRule = kCAFillRuleEvenOdd
+            layer.fillColor = UIColor.black.cgColor
+            layer.opacity = 0.7
+        }
+        maskingView.autoPinEdgesToSuperviewEdges()
```
charlesmchen: This is a big change, if I'm reading it properly.  I'm pretty sure this changes the coordinate system of the crop state and will affect the output of `generateDstImage()` so that this is no longer WYSIWIG.  I did a little testing and it looks close to but a little off.  

To Repro:

* Take a screenshot of a conversation with a lot of messages - conversation view should have a lot of features with sharp edges, like message bubbles.
* Crop to the edges of a message bubble that is near a _corner_ of the screen, so that the edges of the bubble are near but inside the edges of the crop region.  
* Exit view. Look at the cropped output avatar.  The crop will be off.

### Before Adjusting Crop

screen shot 2017-09-01 at 3 17 05 pm

### After Adjusting Crop

screen shot 2017-09-01 at 3 17 18 pm

### Cropped output

screen shot 2017-09-01 at 3 17 24 pm

Might I suggest instead:

* Keep the current coordinate system, where the crop region is represented as a translation and scale relative to source image coordinates, using `srcTranslation` and `imageScale` properties.
* Modify maskingView's bounds to correspond to just the crop circle's frame.  
* Use separate OWSBezierPathView to overlay the rest of the background image with the semi-transparent dark overlay.

charlesmchen: If you don't want to change the maskingView's layout, let me know - I can take a crack at fixing the image rendering so that the crop is WYSIWYG.

PR #2499, opened 2017-09-01
Charlesmchen/fix fake contacts

charlesmchen:
A few fixes to avoid out-of-memory crashes when creating fake contacts.

PTAL @michaelkirk


Signal/src/ViewControllers/AppSettingsViewController.m
```diff
@@ -172,7 +172,7 @@ - (void)updateTableContents
                                                   [weakSelf showAbout];
                                               }]];
 
-#ifdef DEBUG
+#ifdef USE_DEBUG_UI
```
michaelkirk: 👍

PR #2501, opened 2017-09-01
enforce name length limit

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/util/TextFieldHelper.swift
+++ b/Signal/src/util/TextFieldHelper.swift
@@ -1,0 +1,56 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+@objc class TextFieldHelper: NSObject {
+
+    // Used to implement the UITextFieldDelegate method: `textField:shouldChangeCharactersInRange:replacementString`
+    // Takes advantage of Swift's superior unicode handling to append partial pasted text without splitting multi-byte characters.
+    class func textField(_ textField: UITextField, shouldChangeCharactersInRange editingRange: NSRange, replacementString: String, byteLimit: UInt) -> Bool {
+
+        let byteLength = { (string: String) -> UInt in
+            return UInt(string.utf8.count)
+        }
+
+        let existingString = textField.text ?? ""
+
+        // Given an NSRange, we need to interact with the NS flavor of substring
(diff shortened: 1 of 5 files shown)


Signal/src/util/TextFieldHelper.swift
```diff
@@ -0,0 +1,56 @@
...
+//
+
+import Foundation
+import UIKit
+
+@objc class TextFieldHelper: NSObject {
+
+    // Used to implement the UITextFieldDelegate method: `textField:shouldChangeCharactersInRange:replacementString` but takes advantage of
+    // Swifts superior unicode handling to append partial pasted text without splitting multi-byte characters.
+    class func textField(_ textField: UITextField, shouldChangeCharactersInRange editingRange: NSRange, replacementString: String, byteLimit: UInt) -> Bool {
```
charlesmchen: nit: comments and code have long line length making it harder to review.

Signal/src/util/TextFieldHelper.swift
```diff
@@ -0,0 +1,56 @@
...
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+@objc class TextFieldHelper: NSObject {
+
+    // Used to implement the UITextFieldDelegate method: `textField:shouldChangeCharactersInRange:replacementString` but takes advantage of
+    // Swifts superior unicode handling to append partial pasted text without splitting multi-byte characters.
```
charlesmchen: s/Swifts/Swift's/

Signal/src/util/TextFieldHelper.swift
```diff
@@ -0,0 +1,56 @@
...
+        let lengthOfRemainingExistingString = byteLength(existingString) - byteLength(removedString)
+
+        let newLength = lengthOfRemainingExistingString + byteLength(replacementString)
+
+        if (newLength <= byteLimit) {
+            return true
+        }
+
+        // Don't allow any change if inserting a single char is already over the limit (typically this means typing)
+        if (replacementString.characters.count < 2) {
```
charlesmchen: Why special case this?

michaelkirk: Agreed that the behavior should be equivalent without this check, but this simple check for the common case (when a user is typing in their name) avoids the more complex logic/additional allocations required for the more complex case.

PR #2502, opened 2017-09-01
Add group members individual to profile whitelist when adding group.

charlesmchen:
This shouldn't matter in most cases, but it can only help profiles sync faster in case of delivery problems or unreliable networks.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/Profiles/OWSProfileManager.m
+++ b/Signal/src/Profiles/OWSProfileManager.m
@@ -868,6 +868,13 @@ const NSUInteger kOWSProfileManager_MaxAvatarDiameter = 640;
         TSGroupThread *groupThread = (TSGroupThread *)thread;
         NSData *groupId = groupThread.groupModel.groupId;
         [self addGroupIdToProfileWhitelist:groupId];
+
+        // When we add a group to the profile whitelist,
+        // also add all current members to the profile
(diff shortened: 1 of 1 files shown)


michaelkirk:
What's a scenario this is intended to address?

charlesmchen:
> What's a scenario this is intended to address?

You share your profile with a group and delivery fails, so not everyone in the group gets your profile key. Then you send a message to a group member in a 1:1 conversation.  

It's certainly an edge case, but it can't hurt.  NBD

michaelkirk:
LGTM

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -868,6 +868,13 @@ - (void)addThreadToProfileWhitelist:(TSThread *)thread
         TSGroupThread *groupThread = (TSGroupThread *)thread;
         NSData *groupId = groupThread.groupModel.groupId;
         [self addGroupIdToProfileWhitelist:groupId];
+
+        // When we add a group to the profile whitelist,
```
michaelkirk: comment nit: it's usually more helpful to explain why you're doing a thing rather than just restating what the code does. The code is pretty legible as is.

PR #2504, opened 2017-09-01
Refine message view's initial range size.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -78,7 +78,6 @@
 #import <SignalServiceKit/OWSDisappearingMessagesConfiguration.h>
 #import <SignalServiceKit/OWSIdentityManager.h>
 #import <SignalServiceKit/OWSMessageSender.h>
-#import <SignalServiceKit/OWSProfileKeyMessage.h>
 #import <SignalServiceKit/OWSVerificationStateChangeMessage.h>
 #import <SignalServiceKit/SignalRecipient.h>
 #import <SignalServiceKit/TSAccountManager.h>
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -488,6 +488,9 @@ - (void)configureForThread:(TSThread *)thread
 
     [self.uiDatabaseConnection beginLongLivedReadTransaction];
 
+    self.page = 0;
+    [self ensureDynamicInteractions];
```
charlesmchen: We need to update the "unread indicator" _before_ we determine the initial range size, since it depends on where the unread indicator is placed.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -663,9 +648,6 @@ - (void)viewWillAppear:(BOOL)animated
 {
     DDLogDebug(@"%@ viewWillAppear", self.tag);
 
-    // We need to update the dynamic interactions before we do any layout.
-    [self ensureDynamicInteractions];
```
charlesmchen: This is too late to reliably place the unread indicator for the first time.

Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -2487,9 +2469,25 @@ - (void)updateMessageMappingRangeOptions:(MessagesRangeSizeMode)mode
...
+                = (NSUInteger)[self.dynamicInteractions.unreadIndicatorPosition longValue];
+            // If there is an unread indicator, increase the initial load window
+            // to include it.
+            OWSAssert(unreadIndicatorPosition > 0);
+            OWSAssert(unreadIndicatorPosition <= kYapDatabaseRangeMaxLength);
+
+            // We'd like to include at least N seen messages,
+            // to give the user the context of where they left off the conversation.
+            const NSUInteger kPreferredSeenMessageCount = 1;
+            targetLength = unreadIndicatorPosition + kPreferredSeenMessageCount;
```
charlesmchen: It's really quite simple - use the unread indicator position to determine the initial range size.

michaelkirk: This makes sense.

PR #2507, opened 2017-09-05
profile upgrade experience

michaelkirk:
PTAL @charlesmchen
 
profile-iphone5c


Signal/src/ViewControllers/ExperienceUpgradesPageViewController.swift
```diff
@@ -4,6 +4,94 @@
...
+
+        // Title label
+        let titleLabel = UILabel()
+        view.addSubview(titleLabel)
+        titleLabel.text = header
+        titleLabel.textAlignment = .center
+        titleLabel.font = UIFont.ows_regularFont(withSize: ScaleFromIPhone5(24))
+        titleLabel.textColor = UIColor.white
+        titleLabel.minimumScaleFactor = 0.5
+        titleLabel.adjustsFontSizeToFitWidth = true
```
charlesmchen: Can you please attach screenshots on plus size device?

Signal/src/environment/ExperienceUpgradeFinder.swift
```diff
@@ -6,27 +6,45 @@ import Foundation
...
-        }
-
-        return upgrades
+        return [
+            // Disable old experience upgrades. Most people have seen them by now, and accomodating multiple makes layout harder.
+            // Note if we ever want to show multiple experience upgrades again
+            // we'll have to update the layout in ExperienceUpgradesPageViewController
+            //
+            // videoCalling,
+            // (UIDevice.current.supportsCallKit ? callKit : nil),
```
charlesmchen: `ExperienceUpgradesPageViewController` still has paging... is that for future scenarios where we might want to have two pages?

michaelkirk: Correct.

Signal/src/ViewControllers/ExperienceUpgradesPageViewController.swift
michaelkirk: screen shot 2017-09-05 at 4 28 28 pm

PR #2508, opened 2017-09-05
Don't "scroll to bottom" when messages are sent from desktop.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -1704,6 +1704,13 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
     }
 }
 
+- (void)finishSendingMessageAnimated:(BOOL)animated
+{
+    [super finishSendingMessageAnimated:animated];
+
+    [self scrollToBottomAnimated:animated];
+}
+
 - (void)toggleDefaultKeyboard
 {
     // Primary language is nil for the emoji keyboard & we want to stay on it after sending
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/ConversationView/MessagesViewController.m
```diff
@@ -1642,6 +1642,13 @@ - (void)didPressSendButton:(UIButton *)button
     }
 }
 
+- (void)finishSendingMessageAnimated:(BOOL)animated
+{
+    [super finishSendingMessageAnimated:animated];
```
michaelkirk: Though this is more concise, I believe the reason the scroll logic was in yapDatabaseModified was because not until then does the new cell get inserted into the collectionView. 

How can we be sure the interaction has been inserted into the collectionView at this point?

michaelkirk:
I'm not 100% sure, but I actually think the desired behavior would be to stay scrolled to bottom iff we are already at the bottom. But if the user is scrolled up, to *not* scroll to bottom. In other words, I think the rules governing incoming messages vs scroll equally apply to synced outgoing transcripts.

charlesmchen:
> I'm not 100% sure, but I actually think the desired behavior would be to stay scrolled to bottom iff we are already at the bottom. But if the user is scrolled up, to not scroll to bottom. In other words, I think the rules governing incoming messages vs scroll equally apply to synced outgoing transcripts.

Great point. 

PTAL @michaelkirk

charlesmchen:
I've updated the PR, PTAL @michaelkirk

PR #2510, opened 2017-09-05
Ensure [UIApplication registerUserNotificationSettings] and [UIApplic…

charlesmchen:
…ation registerForRemoteNotifications] are called on the main thread.

### I consider this PR a work in progress.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/Models/SyncPushTokensJob.swift
+++ b/Signal/src/Models/SyncPushTokensJob.swift
@@ -27,41 +27,55 @@ class SyncPushTokensJob: NSObject {
     func run() -> Promise<Void> {
         Logger.info("\(TAG) Starting.")
 
-        // Required to potentially prompt user for notifications settings
-        // before `requestPushTokens` will return.
-        self.pushManager.validateUserNotificationSettings()
+        let runPromise: Promise<Void> =
+            Promise { fulfill, _ in
+                // TODO: PromiseKit only lets you specify the queue in `then`,
+                //       so use a noop promise here so that we can specify the
+                //       queues to use below.
+                fulfill()
+                } .then(on:DispatchQueue.main) {
+                    return Promise { fulfill, _ in
+                        AssertIsOnMainThread()
+                        
+                        // Required to potentially prompt user for notifications settings
+                        // before `requestPushTokens` will return.
+                        self.pushManager.validateUserNotificationSettings()
+                        
+                        fulfill()
+                    }
+                } .then(on:DispatchQueue.global()) {
+                    return self.requestPushTokens()
+                } .then(on:DispatchQueue.global()) { (pushToken: String, voipToken: String) in
+                    var shouldUploadTokens = false
 
-        let runPromise: Promise<Void> = self.requestPushTokens().then { (pushToken: String, voipToken: String) in
-            var shouldUploadTokens = false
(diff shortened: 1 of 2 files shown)


Signal/src/Models/SyncPushTokensJob.swift
```diff
@@ -27,41 +27,55 @@ class SyncPushTokensJob: NSObject {
...
-                Logger.debug("\(self.TAG) Successfully ran syncPushTokensJob.")
-            }.catch { error in
-                Logger.error("\(self.TAG) Failed to run syncPushTokensJob with error: \(error).")
-            }
+        let runPromise: Promise<Void> =
+            Promise { fulfill, _ in
+                // TODO: PromiseKit only lets you specify the queue in `then`,
+                //       so use a noop promise here so that we can specify the
+                //       queues to use below.
+                fulfill()
```
charlesmchen: Doesn't PromiseKit have a simpler way to create a new promise that is immediately fulfilled?

Signal/src/Models/SyncPushTokensJob.swift
```diff
@@ -27,41 +27,55 @@ class SyncPushTokensJob: NSObject {
...
+                } .then(on:DispatchQueue.main) {
```
charlesmchen: The `then(on:` changes are the critical thing here.

Signal/src/Models/SyncPushTokensJob.swift
```diff
@@ -27,41 +27,55 @@ class SyncPushTokensJob: NSObject {
...
+                    return Promise { fulfill, _ in
+                        AssertIsOnMainThread()
+                        
+                        // Required to potentially prompt user for notifications settings
+                        // before `requestPushTokens` will return.
+                        self.pushManager.validateUserNotificationSettings()
+                        
+                        fulfill()
+                    }
+                } .then(on:DispatchQueue.global()) {
```
charlesmchen: I've noticed we sometimes use `DispatchQueue.default`.  Should we use `global` instead?

Signal/src/Models/SyncPushTokensJob.swift
michaelkirk: PromiseKit dispatches to the main thread by default (unless you configure PromiseKit otherwise).

IMO it's better to stick with this default as we already generally coordinate things on the main queue across the app and it's more readable to only be explicit when we *break* the convention e.g. in peerConnectionClient when we want to call PromiseKit Futures on a specific (non-main) thread.

What do you think of this alternative which is a smaller change, and IMO the result is a little more legible: https://github.com/WhisperSystems/Signal-iOS/compare/mkirk/validate-settings-on-main-thread?expand=1

Signal/src/Models/SyncPushTokensJob.swift
charlesmchen: Done, thanks.

charlesmchen:
PTAL @michaelkirk

Signal/src/Models/SyncPushTokensJob.swift
charlesmchen: Done.

michaelkirk:
LGTM.

Issue #2202, opened 2017-06-11
HEVC and HEIF are not supported


jabs7:
In ios11 apple introduced all videos and pictures will use HEVC and HEIF formats. Signal seems to not be able to send and receive media messages currently on the ios11 developer beta.

charlesmchen:
Thanks @jabs7 - we'll look into it.

michaelkirk:
@jabs7 

To be clear about your reproduction here:

Take a photo using the native iOS11 camera app.
Open a conversation in Signal, choose media library, and pick the image you just snapped.
This sends as a "file" rather than an image. 

Is that right?

If so, that's fixed in https://github.com/WhisperSystems/Signal-iOS/pull/2511

However you also mentioned having trouble with HEVC. I was not able to reproduce any problems sending video taken on iOS11. If you are, can you provide some more explicit steps to reproduce?

michaelkirk:
Fixed in #2511 

Please re-open if you are able to reproduce an issue with v2.16.0 or greater.

Issue #2148, opened 2017-05-24
Add share extension for Photos


metbril:
- [x] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description
Describe here the issue that you are experiencing.

There is a Share extension for the recently added attachments feature. Unfortunately I still can't share a photo from the iOS Photos app. Could you please add a share extension for photos?

### Steps to reproduce
- using hyphens as bullet points
- list the steps
- that reproduce the bug

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)
**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)

### Device info

**Device**: iDevice X
**iOS version**: X.Y.Z
**Signal version:** Z.Y

NicolasCGN:
I can also confirm it. Sharing photos is possible from Signal, but Signal doesn't show when trying to share a photo form the Photos app.

macuser666:
I can also confirm the absence of share extensions.
This is unfortunate, since without these extensions sharing files from within other apps is very cumbersome. 
Threema or Wire or Whattsapp support share extensions ...

TreatyOfRome:
This is a duplicate of open issue #650. This one can be closed?

michaelkirk:
closed as dupe of #650

PR #2511, opened 2017-09-06
Fix sending images taken on iOS11 camera.

michaelkirk:
fixes https://github.com/WhisperSystems/Signal-iOS/issues/2202

Convert .heic to .jpg upon sending

I punted on displaying HEIC natively in Signal, not so much for technical reasons, but moreso that I'm not sure what the right move is here. iOS seems to want to treat HEIC as an internal optimization that's invisible to the user.  e.g. sharing out an HEIC image from Photos converts it to JPG. 

So at this point, you'd have to be trying pretty hard to actually get an HEIC image into Signal. If you do - maybe you intend for it to *stay* an HEIC? But we couldn't do that if a user e.g. copy/pasted an HEIC image to another thread, since we can't assume the other client understand what HEIC's are. FWIW this is inline with what whatsapp does.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/SignalAttachment.swift
+++ b/Signal/src/ViewControllers/SignalAttachment.swift
@@ -251,7 +251,12 @@ class SignalAttachment: NSObject {
     // Image attachments may be converted to another image format before 
     // being uploaded.
     private class var inputImageUTISet: Set<String> {
-        return MIMETypeUtil.supportedImageUTITypes().union(animatedImageUTISet)
+         // HEIC is valid input, but not valid output. Non-iOS11 clients do not support it.
+        let heicSet: Set<String> = Set(["public.heic"])
+
+        return MIMETypeUtil.supportedImageUTITypes()
+            .union(animatedImageUTISet)
+            .union(heicSet)
     }
 
     // Returns the set of UTIs that correspond to valid _output_ image formats
@@ -441,7 +446,7 @@ class SignalAttachment: NSObject {
         }
 
         guard imageData.count > 0 else {
-            assert(imageData.count > 0)
+            owsFail("\(self.TAG) in \(#function) imageData was empty")
             attachment.error = .invalidData
             return attachment
         }
--- a/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/MessagesViewController.m
@@ -3311,6 +3311,8 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
     } else {
         // Non-Video image picked from library
 
+        // To avoid re-encoding GIF and PNG's as JPEG we have to get the raw data of
+        // the selected item vs. using the UIImagePickerControllerOriginalImage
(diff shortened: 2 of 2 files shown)


charlesmchen:
> I punted on displaying HEIC natively in Signal, not so much for technical reasons, but moreso that I'm not sure what the right move is here.

Makes sense.

> e.g. sharing out an HEIC image from Photos converts it to JPG.

That includes copying and pasting from the camera roll?  And importing from camera roll in Signal?

Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -251,7 +251,12 @@ class SignalAttachment: NSObject {
     // Image attachments may be converted to another image format before 
     // being uploaded.
     private class var inputImageUTISet: Set<String> {
-        return MIMETypeUtil.supportedImageUTITypes().union(animatedImageUTISet)
+         // HEIC is valid input, but not valid output. Non-iOS11 clients do not support it.
+        let heicSet: Set<String> = Set(["public.heic"])
```
charlesmchen: In my cursory reading of the docs, it seems like there a couple of container formats (.heic and .heif) which can contain a variety of encodings.  Will this single UTI cover them all?

Signal/src/ViewControllers/SignalAttachment.swift
```diff
@@ -251,7 +251,12 @@ class SignalAttachment: NSObject {
...
+
+        return MIMETypeUtil.supportedImageUTITypes()
```
charlesmchen: My understanding is that animated .heifs/heics will be used for "live images".  I'm wondering if we should support sharing those (as .gifs?)?

charlesmchen:
So if I send a .heif to you from desktop, you could a) save it to camera roll & share it to someone else .jpg?  b) copy and paste my .heif and send it to someone else as .heif?

Signal/src/ViewControllers/SignalAttachment.swift
michaelkirk: That might be a good idea. I've made a card.

Signal/src/ViewControllers/SignalAttachment.swift
michaelkirk: In all my testing Photos are captured (static, live, pano) as image.heic. However the scenario you outlined below, receiving a .HEIF from elsewhere and saving it into Photos, does make it possible to import a .HEIF.

I've added the public.heif type and verified you can select such a file now.  Thanks!

michaelkirk:
> That includes copying and pasting from the camera roll?

Copy/paste yup.

>  And importing from camera roll in Signal?

Yes, you can select heif/heic images from the photo library from Signal, and they will be encoded as JPG. 

Currently you cannot import-to-signal **from** the Photos app at all. But that's another problem. #2148 

> So if I send a .heif to you from desktop, you could a) save it to camera roll & share it to someone else .jpg? b) copy and paste my .heif and send it to someone else as .heif?

That's right.

michaelkirk:
In https://github.com/WhisperSystems/Signal-iOS/commit/77760b8adb58ad9efbda46b1949be0e92e393b69

PR #2514, opened 2017-09-06
Charlesmchen/pull to refresh contacts

charlesmchen:
### This PR is now ready for review.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/NewContactThreadViewController.m
+++ b/Signal/src/ViewControllers/NewContactThreadViewController.m
@@ -127,7 +127,7 @@ NS_ASSUME_NONNULL_BEGIN
 {
     OWSAssert([NSThread isMainThread]);
 
-    [self.contactsViewHelper.contactsManager requestSystemContactsOnce];
+    [self.contactsViewHelper.contactsManager fetchSystemContactsIfAlreadyAuthorizedAndIgnoreDebounce];
 
     [refreshControl endRefreshing];
 }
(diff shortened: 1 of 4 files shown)


charlesmchen:

PTAL @michaelkirk

Signal/src/ViewControllers/NewContactThreadViewController.m
```diff
@@ -113,9 +113,25 @@ - (void)loadView
...
+    [self.tableViewController.tableView insertSubview:pullToRefreshView atIndex:0];
+
     [self updateTableContents];
 }
 
+- (void)pullToRefreshPerformed:(UIRefreshControl *)refreshControl
+{
+    OWSAssert([NSThread isMainThread]);
+
+    [self.contactsViewHelper.contactsManager requestSystemContactsOnce];
```
michaelkirk: As I understand it, the primary use case for this feature is when Alice has had Bob in her contacts for a long time, but Bob *just* signed up for Signal. Alice won't see Bob listed as a valid recipient (with HEAD of master) unless:
1.  she knows to restart her app
2. she makes some edit to her contacts
3. the contact-refresh timeout expires (12 hours).

As I understand it the purpose of this feature is to give a user a somewhat intuitive way to override that 12 hour timeout. As written, I don't think this change would actually accomplish anything since we already `requestSystemContactsOnce` when this VC appears.

Instead I think we want this to perform a contact intersection **every** time someone pulls to refresh, i.e. call `updateContacts` directly. Though maybe we need to ensure we're doing the right thing WRT contact access.

charlesmchen: Great point.  I've updated the PR.  

PTAL @michaelkirk

PR #2516, opened 2017-09-07
Charlesmchen/conversation back button hot area

charlesmchen:
* Moves the conversation view's title header from being a "left bar button item" to the `titleView`.
* Increases the hot area for the back button which has been hard to hit.
* Takes full advantage of the navigation bar width, which ensures that the "tap for settings" interaction works anywhere in the title area.
* Future-proofs the app against changes in iOS since we're no longer "cutting against the grain" of how navigation bar contents are supposed to work.
* Simplifies this logic.
* Simplifies RTL.

### 1:1 Conversation with disappearing messages (Worst Case)

screen shot 2017-09-07 at 10 46 45 am

### 1:1 Conversation 

screen shot 2017-09-07 at 10 46 38 am

### Group Conversation with disappearing messages (Worst Case)

screen shot 2017-09-07 at 10 44 50 am

### Group Conversation

screen shot 2017-09-07 at 10 44 43 am

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -170,7 +170,7 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
     // We need to manually resize and position the title views;
     // iOS AutoLayout doesn't work inside navigation bar items.
     const int kTitleVSpacing = 0.f;
-    const int kTitleHMargin = 5.f;
+    const int kTitleHMargin = 0.f;
     CGFloat titleHeight = ceil([self.titleLabel sizeThatFits:CGSizeZero].height);
     CGFloat subtitleHeight = ceil([self.subtitleLabel sizeThatFits:CGSizeZero].height);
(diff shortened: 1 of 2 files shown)


charlesmchen:
The downside here is that there's now more apparent & hot space between the back button and the title, but I'd argue that that's an improvement - the back button was too hard to hit before and it's a core interaction.  

Bear in mind that nearly all of the space between the back button and the title is "hot" for the back button, not the title - as it should be.

michaelkirk:
Actually, I ran this on an iPhone7 sim and it feels to close to the the back button.

screen shot 2017-09-07 at 12 46 02 pm

Are you seeing this?

charlesmchen:
> Actually, I ran this on an iPhone7 sim and it feels too close to the the back button.

Great catch.  This affects iPhone 7 / iOS 11 simulator on XCode 9 beta, but not simulators or devices with XCode 8.  And it doesn't appear to affect us before this PR on iPhone 7 / iOS 11 simulator on XCode 9 beta, so I'm not going to try to slip this into v2.16.0.  

I've pushed a fix.

PR #2517, opened 2017-09-07
Always present settings from OWSNavigationController

michaelkirk:
For 2.16

We were using UINavigationController in a couple of places.

PTAL @charlesmchen


charlesmchen:
Can you please re-open against `release/2.16.0` instead of `master`?

PR #2518, opened 2017-09-07
Fix one-time crash when opening thread without having a local profile

michaelkirk:
I saw this crash locally, but am not able to reproduce it from a fresh install.

[symbolicated-incoming_call .txt](https://github.com/WhisperSystems/Signal-iOS/files/1285266/symbolicated-incoming_call.txt)

ProfileManager creates a transaction while saving the localProfile (which must not exist yet) during an existing transaction from `ensureDynamicInteractions`.

"Normally" we create the local profile when the profile view is displayed so this shouldn't an issue.
However, I think it's probably better to be explicit about this creation to catch whatever edge case exists.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/SignalsViewController.m
+++ b/Signal/src/ViewControllers/SignalsViewController.m
@@ -11,6 +11,7 @@
 #import "NSDate+millisecondTimeStamp.h"
 #import "OWSContactsManager.h"
 #import "OWSNavigationController.h"
+#import "OWSProfileManager.h"
 #import "ProfileViewController.h"
 #import "PropertyListPreferences.h"
 #import "PushManager.h"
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/SignalsViewController.m
```diff
@@ -491,6 +492,7 @@ - (void)viewDidAppear:(BOOL)animated {
         // Start running the disappearing messages job in case the newly registered user
         // enables this feature
         [[OWSDisappearingMessagesJob sharedJob] startIfNecessary];
+        [[OWSProfileManager sharedManager] ensureLocalProfileCached];
```
charlesmchen: I feel like this doesn't belong in a VC - it belongs in a "user did register" method in the app delegate.  Thoughts?  Not for this PR of course.

michaelkirk: I agree that we should extract the app setup process. maybe the appDelegate is the right spot.

charlesmchen:
@michaelkirk Okay if I close this?

michaelkirk:
Yup, thanks!

PR #2520, opened 2017-09-07
callscreen uses profile name

michaelkirk:
A handful of related changes:

- Prefer formatted phone number over e164 in call screens

- Show the callers (formatted) phone number on the callkit screen, even if they're not in our contacts (unless callkit privacy is enabled).

- Use profile name on both the call kit and in-app call screens.

Because using the profile name in the in-app call screen uses a lot more space, we attempt to mimic the marquee behavior of the native call screen when the name doesn't fit within the label.

PTAL @charlesmchen


michaelkirk:
@riyapenn and I had a look at this and think this is good to go.

@charlesmchen can you PTAL?

Signal/src/views/MarqueeLabel.swift
```diff
@@ -0,0 +1,1855 @@
+//  Grabbed from: https://github.com/cbpowell/MarqueeLabel-Swift/blob/cd331f3cfc3f9d7114ffa5aa4f243f1d5eda9d0d/Classes/MarqueeLabel.swift
+//  License: MIT License
+//
+//  MarqueeLabel.swift
+//
+//  Created by Charles Powell on 8/6/14.
+//  Copyright (c) 2015 Charles Powell. All rights reserved.
+//
```
charlesmchen: Won't this header get rewritten by the precommit script?

charlesmchen: Our precommit script already ignores source inside directories named "ThirdParty".  We should move this to the ThirdParty dir.

charlesmchen:
Merged into `hotfix/2.16.1`.

PR #2521, opened 2017-09-08
audit/fix up supported l10ns (added km and lt)

michaelkirk:
As policy (enforced by the sync-translations script) we don't include a language in Signal until it has at least
80% of strings translated. We don't currently remove languages that fall
below that threshold.

There are two things that must happen to add a localization in Signal.

1. Pull the Localizable.strings from transifex for each language, add
   them to git.
2. Add the localization to the Xcode project file

These four languages were in the repository (step 1) but not properly
added to the project file (step 2):

- khmer (km)
- lithuanian (lt)
- hebrew (he)
- persian (fa)

Of these, only khmer and lithuanian were sufficiently translated, so I
added them properly to the Xcode project file, making them available to
our users.

To make this process slightly easier to track in the future, I've
removed the insufficiently complete `he` and `fa` files from the
repository. Once they fall above 80% translated they will be
automatically downloaded by our sync-translations script, appearing as
new files, being the cue to the developer to add those new translations
to the project (step 2).

Note: the fa/he translations were never available to users, so we're not
going removing anyones existing functionality.

PTAL @charlesmchen


charlesmchen:
Merged in hotfix/2.16.1.

PR #2523, opened 2017-09-08
Charlesmchen/lazy load attachments

charlesmchen:
Sorry, this PR got large because I fixed some style issues in the course of this work.

* Use `DataSource` to:
  * Lazy-load attachment data.
  * Deduplicate attachment data reads & writes.
  * Avoid unnecessary attachment data reads & writes.
* Use `DataSource` in `SignalAttachment` which is applies to the "preparation" and "approval" steps of sending attachments.
* Use `DataSource` in `OWSMessageSender` which is applies to the sending attachments.
* Apply `OWSFail()` in more places.

More comments in-line.

PTAL @michaelkirk


Signal/src/Models/SignalAttachment.swift
```diff
@@ -60,21 +60,26 @@ class SignalAttachment: NSObject {
...
+    }
+    public var dataLength: UInt {
+        return dataSource.dataLength()
+    }
+    public var dataUrl: URL? {
+        return dataSource.dataUrl()
+    }
+    public var sourceFilename: String? {
+        return dataSource.sourceFilename()
+    }
```
charlesmchen: These moved into `DataSource`.

Signal/src/Models/SignalAttachment.swift
```diff
@@ -60,21 +60,26 @@ class SignalAttachment: NSObject {
...
     // Attachment types are identified using UTIs.
     //
     // See: https://developer.apple.com/library/content/documentation/Miscellaneous/Reference/UTIRef/Articles/System-DeclaredUniformTypeIdentifiers.html
     let dataUTI: String
 
-    // An optional field that indicates the filename, if known, for this attachment.
-    let filename: String?
-
-    static let kOversizeTextAttachmentUTI = "org.whispersystems.oversize-text-attachment"
-    static let kUnknownTestAttachmentUTI = "org.whispersystems.unknown"
```
charlesmchen: These constants moved to `MIMETypeUtils`.

Signal/src/Models/SignalAttachment.swift
```diff
@@ -98,46 +103,24 @@ class SignalAttachment: NSObject {
...
-            do {
-                try data.write(to: fileUrl)
-            } catch {
-                owsFail("\(SignalAttachment.TAG) Could not write data to disk: \(dataUTI)")
-                return nil
-            }
-            temporaryDataUrl = fileUrl
-        }
-        return temporaryDataUrl
-    }
```
charlesmchen: This de-duplication moved into `DataSource`.

Signal/src/Models/SignalAttachment.swift
```diff
@@ -60,21 +60,26 @@ class SignalAttachment: NSObject {
...
+        return dataSource.sourceFilename()
+    }
 
     // Attachment types are identified using UTIs.
     //
     // See: https://developer.apple.com/library/content/documentation/Miscellaneous/Reference/UTIRef/Articles/System-DeclaredUniformTypeIdentifiers.html
     let dataUTI: String
 
-    // An optional field that indicates the filename, if known, for this attachment.
-    let filename: String?
```
charlesmchen: These moved into `DataSource`. We used this property to preserve filenames, e.g. when sending documents.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3440,54 +3428,46 @@ - (void)sendMessageAttachment:(SignalAttachment *)attachment
 
 - (NSURL *)videoTempFolder
 {
-    NSArray *paths = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
-    NSString *basePath = ([paths count] > 0) ? [paths objectAtIndex:0] : nil;
-    basePath = [basePath stringByAppendingPathComponent:@"videos"];
-    if (![[NSFileManager defaultManager] fileExistsAtPath:basePath]) {
-        [[NSFileManager defaultManager] createDirectoryAtPath:basePath
+    NSString *temporaryDirectory = NSTemporaryDirectory();
```
charlesmchen: A small but key change: by creating video temp files in the "temp" folder rather than the "cache" folder, we cue iOS to clean them up sooner.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3440,54 +3428,46 @@ - (void)sendMessageAttachment:(SignalAttachment *)attachment
...
                     self.tag,
                     __PRETTY_FUNCTION__,
                     attachment ? [attachment errorName] : @"Missing data");
                 [self showErrorAlertForAttachment:attachment];
             } else {
                 [self tryToSendAttachmentIfApproved:attachment skipApprovalDialog:skipApprovalDialog];
             }
-
-            NSError *error;
-            [[NSFileManager defaultManager] removeItemAtURL:compressedVideoUrl error:&error];
```
charlesmchen: We don't need to clean up our temp files ourselves, now that we are creating them in the "temp" folder instead of the "cache" folder.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4292,15 +4267,9 @@ - (void)showErrorAlertForAttachment:(SignalAttachment *_Nullable)attachment
...
-                                                        @"The title of the 'attachment error' alert.")
-                                            message:errorMessage
-                                     preferredStyle:UIAlertControllerStyleAlert];
-    [controller addAction:[UIAlertAction actionWithTitle:NSLocalizedString(@"OK", nil)
-                                                   style:UIAlertActionStyleDefault
-                                                 handler:nil]];
-    [self presentViewController:controller animated:YES completion:nil];
+    [OWSAlerts showAlertWithTitle:NSLocalizedString(
+                                      @"ATTACHMENT_ERROR_ALERT_TITLE", @"The title of the 'attachment error' alert.")
+                          message:errorMessage];
```
charlesmchen: Continuing to DRY up code with `OWSAlerts` as well.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -506,27 +508,22 @@ - (void)sendTemporaryAttachmentData:(NSData *)attachmentData
...
             sourceFilename:(nullable NSString *)sourceFilename
                  inMessage:(TSOutgoingMessage *)message
                    success:(void (^)())successHandler
                    failure:(void (^)(NSError *error))failureHandler
 {
-    // There's an odd bug wherein instances of NSData/Data created in Swift
-    // code reliably crash on iOS 9 when calling [NSData writeToFile:...].
-    // We can avoid these crashes by simply copying the Data.
-    //
-    // TODO: Move the iOSVersion header to SSK.
```
charlesmchen: This fix moved into `DataSource`.

SignalServiceKit/src/Util/DataSource.h
```diff
@@ -0,0 +1,71 @@
...
+//
+
+NS_ASSUME_NONNULL_BEGIN
+
+// A protocol that abstracts away a source of NSData
+// and allows us to:
+//
+// * Lazy-load if possible.
+// * Avoid duplicate reads & writes.
+@protocol DataSource
```
charlesmchen: "All problems in computer science can be solved by another level of indirection"

https://en.wikipedia.org/wiki/Indirection

charlesmchen:
One more thing that's worth noting: one of my original motives for eagerly loading attachment data was to "fail quickly", e.g. surface errors to the user before they moved on, while they were still in the context of their activity.  

We paid a heavy price for this and it's not worth it, especially since these types of failures should be very rare and probably indicative of other catastrophic problems - e.g. a path or URL not being readable.  

 This PR has complicated consequences, since we send attachment from many different sources / through many different interactions.  We should only see perf improvements, and in some cases these perf improvements are very large.  e.g. when sending video taken from the camera, we now _never_ load that video into memory which is significant.

SignalServiceKit/src/Util/DataSource.m
```diff
@@ -0,0 +1,280 @@
...
+
++ (nullable id<DataSource>)dataSourceWithData:(NSData *)data utiType:(NSString *)utiType
+{
+    NSString *fileExtension = [MIMETypeUtil fileExtensionForUTIType:utiType];
+    return [self dataSourceWithData:data fileExtension:fileExtension];
+}
+
++ (nullable id<DataSource>)dataSourceWithOversizeText:(NSString *_Nullable)text
+{
+    OWSAssert(text);
```
michaelkirk: nit: This assert seems to contradict the nullability annotation of the `text` param. In other words, should this declaration be: 

```
+ (nullable id<DataSource>)dataSourceWithOversizeText:(NSString *)text
```

So that the compiler can catch places where we might pass in a nil `text`?

SignalServiceKit/src/Util/DataSource.m
```diff
@@ -0,0 +1,280 @@
...
+{
+    OWSAssert(self.dataValue);
+
+    @synchronized(self)
+    {
+        if (!self.cachedFilePath) {
+            NSString *dirPath = NSTemporaryDirectory();
+            NSString *fileName = [[[NSUUID UUID] UUIDString] stringByAppendingPathExtension:self.fileExtension];
+            NSString *filePath = [dirPath stringByAppendingPathComponent:fileName];
+            DDLogError(@"%@ ---- writing data", self.tag);
```
michaelkirk: This seems normal - why is this an error log?

SignalServiceKit/src/Util/DataSource.m
```diff
@@ -0,0 +1,280 @@
...
+}
+
+- (NSData *)data
+{
+    OWSAssert(self.filePath);
+
+    @synchronized(self)
+    {
+        if (!self.cachedData) {
+            DDLogError(@"%@ ---- reading data", self.tag);
```
michaelkirk: Similarly, why is this an error log - it seems normal?

SignalServiceKit/src/Util/DataSource.m
```diff
@@ -0,0 +1,280 @@
...
+    OWSAssert(self.filePath);
+
+    @synchronized(self)
+    {
+        if (!self.cachedDataLength) {
+            NSError *error;
+            NSDictionary<NSFileAttributeKey, id> *_Nullable attributes =
+                [[NSFileManager defaultManager] attributesOfItemAtPath:self.filePath error:&error];
+            if (!attributes || error) {
+                OWSFail(@"%@ Could not read data length from disk: %@", self.tag, self.filePath);
```
michaelkirk: nit: log the error?

SignalServiceKit/src/Util/DataSource.m
```diff
@@ -0,0 +1,280 @@
...
+}
+
+- (BOOL)writeToPath:(NSString *)dstFilePath
+{
+    OWSAssert(self.filePath);
+
+    NSError *error;
+    BOOL success = [[NSFileManager defaultManager] copyItemAtPath:self.filePath toPath:dstFilePath error:&error];
+    if (!success || error) {
+        OWSFail(@"%@ Could not write data from path: %@, to path: %@", self.tag, self.filePath, dstFilePath);
```
michaelkirk: nit: log the error?

SignalServiceKit/src/Util/OWSError.h
```diff
@@ -25,6 +25,7 @@ typedef NS_ENUM(NSInteger, OWSErrorCode) {
     OWSErrorCodeMessageSendFailedToBlockList = 777406,
     OWSErrorCodeMessageSendNoValidRecipients = 777407,
     OWSErrorCodeContactsUpdaterRateLimit = 777407,
+    OWSErrorCodeCouldNotWriteAttachmentData = 777408,
```
michaelkirk: huh, there are two `777407`s

Signal/src/ViewControllers/ConversationView/ConversationViewController.m

Signal/src/Models/SignalAttachment.swift
```diff
@@ -305,6 +288,18 @@ class SignalAttachment: NSObject {
         return SignalAttachment.audioUTISet.contains(dataUTI)
     }
 
+    public var isValidImage: Bool {
```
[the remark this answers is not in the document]
charlesmchen: Done.

Signal/src/Models/SignalAttachment.swift
```diff
@@ -543,8 +546,14 @@ class SignalAttachment: NSObject {
                                                                 return attachment
             }
 
-            if jpgImageData.count <= kMaxFileSizeImage {
-                let recompressedAttachment = SignalAttachment(data : jpgImageData, dataUTI: kUTTypeJPEG as String, filename: filename)
+            guard let dataSource = DataSourceValue.dataSource(with:jpgImageData, fileExtension:"jpg") else {
```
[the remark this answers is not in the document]
charlesmchen: This is a bit subtle, but `DataSource`'s have a `fileExtension` property (because they may need to be written to disk and their path/url should have the write file extension) whereas `SignalAttachment` has a UTI property.

SignalServiceKit/src/Util/DataSource.m
charlesmchen: I've made the text non-null.

SignalServiceKit/src/Util/DataSource.m
charlesmchen: Removed.

SignalServiceKit/src/Util/DataSource.m
charlesmchen: Removed.

SignalServiceKit/src/Util/OWSError.h
charlesmchen: Fixed.

charlesmchen:
https://github.com/WhisperSystems/Signal-iOS/pull/2546

PR #2525, opened 2017-09-11
Add modal activity indicator view.

charlesmchen:
Add a reusable "modal activity indicator" view that can be used during long-running blocking operations to:

* Block user interactions.
* Give visual indication to the user of the app state.
* Optionally offer a cancel button.

Note:

* The modal blocks UI immediately but hides for a second to avoid showing up during short-lived operations.
* The modal fades into view.
* We could add a label indicating what the app is doing, but this should be clear from context.

PTAL @michaelkirk 

screen shot 2017-09-11 at 12 41 08 pm

### DIFF
--- a/Signal/src/ViewControllers/AppSettingsViewController.m
+++ b/Signal/src/ViewControllers/AppSettingsViewController.m
@@ -374,12 +374,24 @@
 
 - (void)proceedToUnregistration
 {
-    [TSAccountManager unregisterTextSecureWithSuccess:^{
-        [Environment resetAppData];
-    }
-        failure:^(NSError *error) {
-            [OWSAlerts showAlertWithTitle:NSLocalizedString(@"UNREGISTER_SIGNAL_FAIL", @"")];
-        }];
+    [ModalActivityIndicatorViewController
+        presentFromViewController:self
+                        canCancel:NO
+                presentCompletion:^(ModalActivityIndicatorViewController *modalActivityIndicator) {
+                    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+                        [TSAccountManager unregisterTextSecureWithSuccess:^{
+                            [Environment resetAppData];
+                        }
+                            failure:^(NSError *error) {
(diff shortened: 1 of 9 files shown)


charlesmchen:
We currently show the modal:

* When creating a group.
* When re-encoding video attachments.
* When deleting your signal account.
* When looking up users by phone number.

Signal/src/ViewControllers/AppSettingsViewController.m
```diff
@@ -374,12 +374,24 @@ - (void)unregisterUser
 
 - (void)proceedToUnregistration
 {
-    [TSAccountManager unregisterTextSecureWithSuccess:^{
-        [Environment resetAppData];
-    }
-        failure:^(NSError *error) {
-            [OWSAlerts showAlertWithTitle:NSLocalizedString(@"UNREGISTER_SIGNAL_FAIL", @"")];
-        }];
+    [ModalActivityIndicatorViewController
```
charlesmchen: The downside of the modal's API is that you end up with JS-style "callback hell."

michaelkirk: Proposal for a little readability improvement...

As I understand it, this modal is used to block the user from interacting with the app while a (relatively) long running background task occurs. As such, I think it would be reasonable to have the block passed execute on a background queue by default - saving the caller from having to dispatch.

In that case I'd also probably rename the param for clarity:

```
[ModalActivityIndicatorViewController presentFromViewController:self
                                                      canCancel:NO
                                                backgroundBlock:^(ModalActivityIndicatorViewController *modalActivityIndicator) {

    [TSAccountManager unregisterTextSecureWithSuccess:^{
                             [Environment resetAppData];
                         }
                                               failure:^(NSError *error) {
                                     [modalActivityIndicator dismissWithCompletion:^{
                                         [OWSAlerts
                                             showAlertWithTitle:NSLocalizedString(@"UNREGISTER_SIGNAL_FAIL", @"")];
                                     }];
                                 });
                             }];
}];
```

charlesmchen: Done. https://github.com/WhisperSystems/Signal-iOS/pull/2548

PR #2526, opened 2017-09-12
Charlesmchen/flat button

charlesmchen:
This button is a step towards a "design language":

* Consistent corner rounding.
* Consistent "down state" (usually derived from "up state").
* Consistent font sizes / weights.
* DRY up view code.

PTAL @michaelkirk 

screen shot 2017-09-12 at 10 36 13 am
screen shot 2017-09-12 at 10 30 56 am
screen shot 2017-09-12 at 10 30 52 am
screen shot 2017-09-12 at 10 20 22 am
screen shot 2017-09-12 at 10 17 03 am
screen shot 2017-09-12 at 10 13 06 am
screen shot 2017-09-12 at 10 10 26 am
screen shot 2017-09-12 at 9 40 58 am

### DIFF
--- a/Signal/src/util/UIImage+OWS.m
+++ b/Signal/src/util/UIImage+OWS.m
@@ -2,26 +2,28 @@
 //  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
-#import "UIImage+normalizeImage.h"
+#import "UIImage+OWS.h"
 
 @implementation UIImage (normalizeImage)
 
-- (UIImage *)normalizedImage {
+- (UIImage *)normalizedImage
+{
     if (self.imageOrientation == UIImageOrientationUp) {
         return self;
     }
 
     UIGraphicsBeginImageContextWithOptions(self.size, NO, self.scale);
-    [self drawInRect:(CGRect){{0, 0}, self.size}];
+    [self drawInRect:(CGRect){ { 0, 0 }, self.size }];
     UIImage *normalizedImage = UIGraphicsGetImageFromCurrentImageContext();
     UIGraphicsEndImageContext();
     return normalizedImage;
 }
 
-- (UIImage *)resizedWithQuality:(CGInterpolationQuality)quality rate:(CGFloat)rate {
+- (UIImage *)resizedWithQuality:(CGInterpolationQuality)quality rate:(CGFloat)rate
+{
     UIImage *resized = nil;
-    CGFloat width    = self.size.width * rate;
-    CGFloat height   = self.size.height * rate;
+    CGFloat width = self.size.width * rate;
+    CGFloat height = self.size.height * rate;
 
     UIGraphicsBeginImageContext(CGSizeMake(width, height));
     CGContextRef context = UIGraphicsGetCurrentContext();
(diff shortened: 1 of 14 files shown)


Signal/src/util/UIImage+OWS.m
```diff
@@ -166,4 +168,28 @@ - (UIImage *)resizedImageToFillPixelSize:(CGSize)dstSize
...
+
+    return [self imageWithColor:color size:CGSizeMake(1.f, 1.f)];
+}
+
++ (UIImage *)imageWithColor:(UIColor *)color size:(CGSize)size
+{
+    OWSAssert(color);
+
+    CGRect rect = CGRectMake(0.0f, 0.0f, size.width, size.height);
+    UIGraphicsBeginImageContextWithOptions(rect.size, NO, 1.f);
```
michaelkirk: Can we assert main thread here?

Signal/src/views/OWSFlatButton.swift
```diff
@@ -0,0 +1,176 @@
...
+    public func setPressedBlock(_ pressedBlock: @escaping () -> Void) {
+        guard self.pressedBlock == nil else {
+            owsFail("Button already has pressed block.")
+            return
+        }
+        self.pressedBlock = pressedBlock
+    }
+
+    internal func buttonPressed() {
+        guard let pressedBlock = pressedBlock else {
```
michaelkirk: nit: This method could just be `pressedBlock?()`

Signal/src/views/OWSFlatButton.swift
```diff
@@ -0,0 +1,176 @@
...
+    public class func fontForHeight(_ height: CGFloat) -> UIFont {
+        let fontPointSize = round(height * 0.45)
+        return UIFont.ows_mediumFont(withSize:fontPointSize)!
+    }
+
+    // MARK: Methods
+
+    public func setTitle(title: String, font: UIFont,
+                         titleColor: UIColor ) {
+        guard let button = self.button else {
```
michaelkirk: nit: Is there a reason `self.button` isn't assigned in init so we can avoid these guards?

Signal/src/views/OWSFlatButton.swift
```diff
@@ -0,0 +1,176 @@
...
+                                    downColor: UIColor ) {
+        guard let button = self.button else {
+            owsFail("Missing button")
+            return
+        }
+        button.setBackgroundImage(UIImage(color:upColor), for: .normal)
+        button.setBackgroundImage(UIImage(color:downColor), for: .highlighted)
+    }
+
+    public func setBackgroundColors(_ backgroundColor: UIColor ) {
```
michaelkirk: At the call sites above, I didn't notice the pluralization of this method, and assumed we were just calling UIView's built in `setBackgrondColor:`. 

How about renaming to `setBackgroundColors(upColor:)` /`setBackgroundColorsWithUpColor:` to disambiguate.

Signal/src/util/UIImage+OWS.m
charlesmchen: Done.

Signal/src/views/OWSFlatButton.swift
charlesmchen: Done.  I wanted to gather "creation & configuration of views" logic in `createContent()` but pragmatism > purity.

Signal/src/views/OWSFlatButton.swift
charlesmchen: Nice idea.  I also added a `owsFail()` in the `setBackgroundColor:` property setter to catch slip ups.

Signal/src/views/OWSFlatButton.swift
charlesmchen: Done.

charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2549

PR #2527, opened 2017-09-12
Charlesmchen/calls vs permissions crashes

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -1622,7 +1622,8 @@ protocol CallServiceObserver: class {
         }
 
         let frontmostViewController = UIApplication.shared.frontmostViewControllerIgnoringAlerts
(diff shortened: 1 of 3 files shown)


Signal/src/call/CallService.swift
```diff
@@ -1621,9 +1621,10 @@ protocol CallServiceObserver: class {
             return
         }
 
-        guard nil != UIApplication.shared.frontmostViewController as? CallViewController else {
+        let frontmostViewController = UIApplication.shared.frontmostViewControllerIgnoringAlerts
+            guard nil != frontmostViewController as? CallViewController else {
```
michaelkirk: weird indent here.

charlesmchen: Fixed.

PR #2530, opened 2017-09-13
Fix NPEs when profiles change while entering conversation view.

charlesmchen:
We don't want to register our notification listeners until the conversation view loads, otherwise profile change events in particular can fire before the view is loaded and NPEs are easy to hit.

// FREEBIE


michaelkirk:
LGTM

PR #2533, opened 2017-09-14
Startup logging.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
> It might be helpful to consolidate all the general purpose diagnostics (e.g. this and whats in AppVersion) and prefix them with a grep-able TAG.

Interesting.  Please feel free to do that!

PR #2534, opened 2017-09-14
Move post-registration work from view to app delegate.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
@michaelkirk I know you expressed some uncertainty about whether or not the `AppDelegate` was the best place to hang this logic.  I'm merging this PR since it's blocking the "Read receipts" work which will pile up quickly tomorrow.  I'm very happy to revisit the question of where this logic should hang.

PR #2537, opened 2017-09-15
Don't send sync messages to self if no linked devices.

charlesmchen:
This should dramatically reduce the number of requests we make for most users.

PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSMessageSender.m
+++ b/SignalServiceKit/src/Messages/OWSMessageSender.m
@@ -964,22 +964,37 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
 
     NSString *localNumber = [TSAccountManager localNumber];
     if ([localNumber isEqualToString:recipient.uniqueId]) {
-        if (deviceMessages.count < 1) {
-            DDLogInfo(@"Ignoring sync message without linked devices: %@", [message class]);
+        __block BOOL hasSecondaryDevices;
+        [self.dbConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
+            hasSecondaryDevices = [OWSDevice hasSecondaryDevicesWithTransaction:transaction];
+        }];
+
+        BOOL hasDeviceMessages = deviceMessages.count > 0;
+
+        if (!hasSecondaryDevices && !hasDeviceMessages) {
+            DDLogInfo(@"%@ Ignoring sync message without secondary devices: %@", self.tag, [message class]);
             OWSAssert([message isKindOfClass:[OWSOutgoingSyncMessage class]]);
 
             dispatch_async([OWSDispatch sendingQueue], ^{
+                // This emulates the completion logic of an actual successful save (see below).
                 [recipient save];
-                [self handleMessageSentLocally:message];
                 successHandler();
             });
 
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -939,6 +939,25 @@ - (void)sendMessage:(TSOutgoingMessage *)message
         }
     }
 
+    NSString *localNumber = [TSAccountManager localNumber];
+    if ([localNumber isEqualToString:recipient.uniqueId]) {
+        if (deviceMessages.count < 1) {
```
michaelkirk: nit: This seems like kind of a weird place for this logic - I'd expect it to be in or near `sendSyncTranscriptForMessage`.

i.e. In the case of sending an attachment to ourself we'd get as far as uploading the attachment before bailing out here.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -939,6 +939,25 @@ - (void)sendMessage:(TSOutgoingMessage *)message
...
+            DDLogInfo(@"Ignoring sync message without linked devices: %@", [message class]);
+            OWSAssert([message isKindOfClass:[OWSOutgoingSyncMessage class]]);
+
+            dispatch_async([OWSDispatch sendingQueue], ^{
+                [recipient save];
```
michaelkirk: nit: Can you leave a comment as to why we need to save the recipient here?

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -939,6 +939,25 @@ - (void)sendMessage:(TSOutgoingMessage *)message
...
+                [self handleMessageSentLocally:message];
```
michaelkirk: If this code is about preventing sync-message sending - presumably it was called via `sendSyncTranscriptForMessage` which in turn is only ever called from within `handleMessageSentLocally`.

Currently  this shouldn't cause an infinite loop because `handleMessageSentLocally` has a short-circuit to make sure it only syncs once, but why call it at all here?

SignalServiceKit/src/Messages/OWSMessageSender.m
charlesmchen: `sendSyncTranscriptForMessage` is only used for sending "sync transcripts".  We want this logic to apply to all sync messages.  Search for ` : OWSOutgoingSyncMessage` for a complete list.

SignalServiceKit/src/Messages/OWSMessageSender.m
charlesmchen: Done.  I'm trying to emulate the behavior of successful sends.

SignalServiceKit/src/Messages/OWSMessageSender.m
charlesmchen: I think you're right - this should be harmless but always redundant.  Removed.

charlesmchen:
PTAL @michaelkirk I significantly reworked this to be more cautious, improving logging, etc.  I'm going to merge since I think this will be uncontroversial, but I suggest you TAL.

PR #2538, opened 2017-09-15
Modify read receipt protos to support sending read receipts to both l…

charlesmchen:
…inked devices and senders.

PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/protobuf/OWSSignalServiceProtos.proto
+++ b/SignalServiceKit/protobuf/OWSSignalServiceProtos.proto
@@ -46,7 +46,6 @@ message ReceiptMessage {
 
   optional Type   type       = 1;
   repeated uint64 timestamps = 2;
-  optional uint64 when       = 3;
 }
 
 message CallMessage {
--- a/SignalServiceKit/protobuf/OWSSignalServiceProtos.proto
+++ b/SignalServiceKit/protobuf/OWSSignalServiceProtos.proto
@@ -44,8 +44,9 @@ message ReceiptMessage {
     READ     = 1;
   }
 
-  optional Type   type      = 1;
-  repeated uint64 timestamp = 2;
+  optional Type   type       = 1;
+  repeated uint64 timestamp  = 2;
+  optional bytes  profileKey = 3;
 }
 
 message CallMessage {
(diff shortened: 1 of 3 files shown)


SignalServiceKit/protobuf/OWSSignalServiceProtos.proto
```diff
@@ -142,19 +142,25 @@ message SyncMessage {
     optional Type type = 1;
   }
 
-  message Read {
+  message ReadLinkedDevices {
```
charlesmchen: Renaming the `Read` message to `ReadLinkedDevices` to disambiguate from `ReadSender`.

SignalServiceKit/protobuf/OWSSignalServiceProtos.proto
```diff
@@ -142,19 +142,25 @@ message SyncMessage {
...
-  optional Sent     sent     = 1;
-  optional Contacts contacts = 2;
-  optional Groups   groups   = 3;
-  optional Request  request  = 4;
-  repeated Read     read     = 5;
-  optional Blocked  blocked  = 6;
-  optional Verified verified = 7;
-  optional bytes    padding  = 8;
+  message ReadSender {
+    optional bytes  groupId   = 1;
```
charlesmchen: groupId should not be set for messages sent in 1:1 conversations.

SignalServiceKit/protobuf/OWSSignalServiceProtos.proto
```diff
@@ -31,10 +31,22 @@ message Envelope {
...
+
+message ReceiptMessage {
+  enum Type {
+    DELIVERY = 0;
+    READ     = 1;
+  }
+
+  optional Type   type       = 1;
+  repeated uint64 timestamp  = 2;
+  optional bytes  profileKey = 3;
```
michaelkirk: We're gonna be sending a lot of these - do we really want to include the profile key every time? It would seemingly dwarf the size of the rest of the message.

charlesmchen: Good point - and Moxie already asked me to remove this I just forgot to push the commit. Sorry!

PR #2539, opened 2017-09-15
Charlesmchen/read receipts manager

charlesmchen:

Add read receipts manager.
* Simplify read receipts plumbing.
* Rework incoming read receipts handling.
* Rework outgoing read receipts handling.
* Make "database view registration complete" check thread-safe.
* Don't send sync messages to self if no linked devices.

PTAL @michaelkirk


charlesmchen:
Note this PR was at d4a29fb

charlesmchen:

PTAL @michaelkirk

SignalServiceKit/src/Devices/OWSReadReceiptsProcessor.h
```diff
@@ -12,6 +12,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 extern NSString *const OWSReadReceiptsProcessorMarkedMessageAsReadNotification;
 
+// TODO:
```
charlesmchen: As mentioned, I plan to rework this class later this week.

michaelkirk:
> Note this PR was at d4a29fb

Sorry, I'm not sure what you mean. Do you mean you force pushed? Can I safely ignore d4a29fb?

charlesmchen:
> Sorry, I'm not sure what you mean. Do you mean you force pushed? Can I safely ignore d4a29fb?

@michaelkirk Sorry, please ignore that comment.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -949,6 +949,25 @@ - (void)sendMessage:(TSOutgoingMessage *)message
         }
     }
```
michaelkirk: Not sure if you intended to include this change in this PR, but see my comments here: https://github.com/WhisperSystems/Signal-iOS/pull/2537

SignalServiceKit/src/Messages/OWSReadReceiptManager.m
```diff
@@ -0,0 +1,191 @@
...
+}
+
+- (void)dealloc
+{
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
+}
+
+- (void)databaseViewRegistrationComplete
+{
+    [self scheduleProcessing];
```
michaelkirk: Do we need to manually kick this off if db view registration completes before this class is `init`?

Eventually we'll read a message locally, which also kicks it off, but is there a case where we'll still want this to run?

SignalServiceKit/src/Messages/OWSReadReceiptManager.m
```diff
@@ -0,0 +1,191 @@
...
+    });
+}
+
+- (void)process
+{
+    @synchronized(self)
+    {
+        self.isProcessing = NO;
+
+        NSArray<OWSReadReceipt *> *readReceiptsToSend = [[self.toLinkedDevicesReadReceiptMap allValues] copy];
```
michaelkirk: nit: seems like this `copy` isn't doing anything.

SignalServiceKit/src/Messages/OWSReadReceiptManager.m
```diff
@@ -0,0 +1,191 @@
...
+            });
+        }
+    }
+}
+
+- (void)messageWasReadLocally:(TSIncomingMessage *)message;
+{
+    @synchronized(self)
+    {
+        NSString *threadUniqueId = message.thread.uniqueId;
```
michaelkirk: significant: this creates a transaction, instead we already have the thread's unqiqueId at message.uniqueThreadId

SignalServiceKit/src/Messages/OWSReadReceiptManager.m
```diff
@@ -0,0 +1,191 @@
...
+        OWSAssert(threadUniqueId.length > 0);
+
+        // Only groupthread sets authorId, thus this crappy code.
```
michaelkirk: nit: update comments per previous discussion. Only legacy 1:1 incoming messages will be lacking an author id. Should we move the logic into the message class?

SignalServiceKit/src/Messages/OWSReadReceiptManager.m
```diff
@@ -0,0 +1,191 @@
...
+            // Group Thread
+            messageAuthorId = message.authorId;
+        } else {
+            // Contact Thread
+            messageAuthorId = [TSContactThread contactIdFromThreadId:message.uniqueThreadId];
+        }
+        OWSAssert(messageAuthorId.length > 0);
+
+        OWSReadReceipt *newReadReceipt =
+            [[OWSReadReceipt alloc] initWithSenderId:messageAuthorId timestamp:message.timestamp];
```
michaelkirk: Clarifying question: 

So, at the individual OWSReadReceipt model level:

e.g. `[[OWSReadReceipt alloc] initWithSenderId:@"AlicePhoneNumber" timestamp:123]`

Could represent either:
- I read a message from Alice, with timestamp 123, on my linked device
- or that Alice read a message from me sent at 123

Presumably the distinction will be made just by whether the read receipt is in the `toLinkedDevicesReadReceiptMap` and a forthcoming `fromOtherUsersReadReceiptMap` (e.g.).

Is that right?

Technically that's enough information, but for sanity we may want to include an explicit type (incoming vs. outgoing) so that we can make asserts.

Take it or leave it - we can revisit it later if it seems like it could be an issue.

SignalServiceKit/src/Messages/Interactions/TSIncomingMessage.m
```diff
@@ -141,10 +140,7 @@ - (void)markAsReadWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
     if (sendReadReceipt) {
         // Notification must happen outside of the transaction, else we'll likely crash when the notification receiver
```
michaelkirk: nit: This comment is now incorrect. 

Previously it was easier to see that nothing fancy was happening at this point within this method, which is nice because we know there is an open transaction (per the transaction param). 

Is there anything you can think of to help ensure future us do something dumb like open a new db transaction within the `messageWasReadLocally` code path? Maybe the answer is just "be careful" 😬 

FWIW I prefer the old async notification mechanism for this reason, but it's probably not going to be as performant as what you've proposed here.

This isn't a blocking concern - just an ongoing quest to get better at reasoning about our db transaction adjacent code.

SignalServiceKit/src/Messages/OWSMessageSender.m
charlesmchen: Removed!

SignalServiceKit/src/Messages/OWSReadReceiptManager.m
charlesmchen: Great point!

SignalServiceKit/src/Messages/OWSReadReceiptManager.m
charlesmchen: Removed.

SignalServiceKit/src/Messages/OWSReadReceiptManager.m
charlesmchen: Good catch.

SignalServiceKit/src/Messages/OWSReadReceiptManager.m
charlesmchen: Done.  I DRY'd this up and applied the new method in one new place.

SignalServiceKit/src/Messages/OWSReadReceiptManager.m
charlesmchen: There's four kind of read receipts:

* Sent to linked device.
* Received from linked device.
* Sent to message sender.
* Received from message recipient.

First, `OWSReadReceipt` is used to represent the first two, not the last two.  

Secondly, `OWSReadReceipt` isn't a new class - I've just streamlined the logic that uses it.  I was tempted to rename/rework it, but I didn't want to bother with the work/possible bugs that would come from that kind of migration.

SignalServiceKit/src/Messages/Interactions/TSIncomingMessage.m
charlesmchen: Yup, I have the same questions.  I can tell you that I've become much more careful when touching code paths that are hard to exercise in a test plan.  A code path like this that is touched by every read message I'm less worried about.

michaelkirk: fair point!

PR #2556, opened 2017-09-20
Fix 1-time crash when launching 2.16 from notification

michaelkirk:
didBecomeActive is too late in the case of launching from a
notification.

Also, start tracking when app setup is complete, and prevent certain
actions from occurring in that case. Eventually we'll enqueue these
actions rather than ignoring them, but we'll want to do more testing
before releasing that. In the meanwhile, if the environment isn't setup
at this point, a crash would be eminent anyway.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/AppDelegate.h
+++ b/Signal/src/AppDelegate.h
@@ -10,5 +10,6 @@ extern NSString *const AppDelegateStoryboardMain;
 
 @interface AppDelegate : UIResponder <UIApplicationDelegate>
 
+@property (atomic) BOOL isEnvironmentSetup;
 
 @end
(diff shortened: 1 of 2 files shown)


charlesmchen:
* Please add a card to add a DispatchWhenEnvironmentReady() (with last-writer-wins) method.  
* Please add a card to add a NSNotification that components can use to observe environment becoming ready.

charlesmchen:
https://trello.com/c/Capp2l73/1565-add-a-dispatchwhenenvironmentready-with-last-writer-wins-method

charlesmchen:
https://trello.com/c/a9GhSKhS/1566-add-a-nsnotification-that-components-can-use-to-observe-environment-becoming-ready

PR #2557, opened 2017-09-20
Charlesmchen/precommit includes and classes

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2,18 +2,18 @@
 //  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
-#import "ConversationViewController.h"
 #import "AppDelegate.h"
 #import "AttachmentSharing.h"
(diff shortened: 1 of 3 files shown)


SignalServiceKit/src/Devices/OWSReadReceiptsProcessor.h
```diff
@@ -4,8 +4,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class OWSSignalServiceProtosSyncMessageRead;
```
charlesmchen: It sorts and deduplicates `@class` declarations.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2,18 +2,18 @@
 //  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
-#import "ConversationViewController.h"
```
charlesmchen: It sorts, deduplicates and normalizes includes and imports.

PR #2559, opened 2017-09-21
Charlesmchen/fix message processing edge cases

charlesmchen:

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/DebugUI/DebugUIMisc.m
+++ b/Signal/src/ViewControllers/DebugUI/DebugUIMisc.m
@@ -61,6 +61,13 @@ NS_ASSUME_NONNULL_BEGIN
                                      actionBlock:^{
                                          [DebugUIMisc clearHasDismissedOffers];
                                      }]];
+    if ([thread isKindOfClass:[TSGroupThread class]]) {
+        TSGroupThread *groupThread = (TSGroupThread *)thread;
+        [items addObject:[OWSTableItem itemWithTitle:@"Hallucinate twin group"
+                                         actionBlock:^{
+                                             [DebugUIMisc hallucinateTwinGroup:groupThread];
+                                         }]];
+    }
     return [OWSTableSection sectionWithTitle:self.name items:items];
 }
 
@@ -113,6 +120,28 @@ NS_ASSUME_NONNULL_BEGIN
                 }
             }
         }];
+}
+
+// Creates a new group (by cloning the current group) without informing the,
+// other members. This can be used to test "group info requests", etc.
++ (void)hallucinateTwinGroup:(TSGroupThread *)groupThread
+{
+    __block TSGroupThread *thread;
+    [[TSStorageManager sharedManager].dbReadWriteConnection
+        readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
+            TSGroupModel *groupModel =
+                [[TSGroupModel alloc] initWithTitle:[groupThread.groupModel.groupName stringByAppendingString:@" Copy"]
+                                          memberIds:[groupThread.groupModel.groupMemberIds mutableCopy]
+                                              image:groupThread.groupModel.groupImage
+                                            groupId:[SecurityUtils generateRandomBytes:16]];
+            thread = [TSGroupThread getOrCreateThreadWithGroupModel:groupModel transaction:transaction];
+        }];
+    OWSAssert(thread);
+
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [Environment presentConversationForThread:thread];
+
+    });
 }
 
 @end
--- a/SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
+++ b/SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
@@ -102,7 +102,7 @@ NS_ASSUME_NONNULL_BEGIN
 
     // If there is an attachment + text, render the text here, as Signal-iOS renders two messages.
     if (attachmentsProcessor.hasSupportedAttachments && transcript.body && ![transcript.body isEqualToString:@""]) {
-        // render text *after* the attachment
+        // We want the text to appear after the attachment.
         uint64_t textMessageTimestamp = transcript.timestamp + 1;
         TSOutgoingMessage *textMessage = [[TSOutgoingMessage alloc] initWithTimestamp:textMessageTimestamp
                                                                              inThread:thread
--- a/SignalServiceKit/src/Messages/Interactions/TSInteraction.m
+++ b/SignalServiceKit/src/Messages/Interactions/TSInteraction.m
@@ -14,6 +14,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 + (instancetype)interactionForTimestamp:(uint64_t)timestamp
                         withTransaction:(YapDatabaseReadWriteTransaction *)transaction {
+    OWSAssert(timestamp > 0);
+
     __block int counter = 0;
     __block TSInteraction *interaction;
 
--- a/SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
+++ b/SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
@@ -132,6 +132,7 @@ NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSBatchMessageProc
 
 - (void)addJobWithEnvelopeData:(NSData *)envelopeData plaintextData:(NSData *_Nullable)plaintextData
 {
+    // We need to persist the decrypted envelope data ASAP to prevent data loss.
     [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
         OWSMessageContentJob *job =
             [[OWSMessageContentJob alloc] initWithEnvelopeData:envelopeData plaintextData:plaintextData];
@@ -280,6 +281,7 @@ NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSBatchMessageProc
 {
     OWSAssert(envelopeData);
 
+    // We need to persist the decrypted envelope data ASAP to prevent data loss.
     [self.finder addJobWithEnvelopeData:envelopeData plaintextData:plaintextData];
 }
 
@@ -322,7 +324,15 @@ NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSBatchMessageProc
                        self.tag,
                        jobs.count,
                        [OWSMessageContentJob numberOfKeysInCollection]);
-                   [self drainQueueWorkStep];
+
+                   // Wait a bit in hopes of increasing the batch size.
+                   // This delay won't affect the first message to arrive when this queue is idle,
+                   // so by definition we're receiving more than one message and can benefit from
+                   // batching.
+                   dispatch_after(
+                       dispatch_time(DISPATCH_TIME_NOW, (int64_t)0.1f * NSEC_PER_SEC), dispatch_get_main_queue(), ^{
+                           [self drainQueueWorkStep];
+                       });
                });
            }];
 }
@@ -429,6 +439,7 @@ NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSBatchMessageProc
 {
     OWSAssert(envelopeData);
 
+    // We need to persist the decrypted envelope data ASAP to prevent data loss.
     [self.processingQueue enqueueEnvelopeData:envelopeData plaintextData:plaintextData];
     [self.processingQueue drainQueue];
 }
--- a/SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
+++ b/SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
@@ -5,6 +5,7 @@
 #import "OWSBatchMessageProcessor.h"
 #import "NSArray+OWS.h"
 #import "OWSMessageManager.h"
+#import "OWSQueues.h"
 #import "OWSSignalServiceProtos.pb.h"
 #import "TSDatabaseView.h"
 #import "TSStorageManager.h"
@@ -277,6 +278,16 @@ NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSBatchMessageProc
 
 #pragma mark - instance methods
 
+- (dispatch_queue_t)serialQueue
+{
+    static dispatch_queue_t queue = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        queue = dispatch_queue_create("org.whispersystems.message.process", DISPATCH_QUEUE_SERIAL);
+    });
+    return queue;
+}
+
 - (void)enqueueEnvelopeData:(NSData *)envelopeData plaintextData:(NSData *_Nullable)plaintextData
 {
     OWSAssert(envelopeData);
@@ -287,7 +298,7 @@ NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSBatchMessageProc
 
 - (void)drainQueue
 {
-    DispatchMainThreadSafe(^{
+    dispatch_async(self.serialQueue, ^{
         if ([TSDatabaseView hasPendingViewRegistrations]) {
             // We don't want to process incoming messages until database
             // view registration is complete.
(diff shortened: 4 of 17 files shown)


Signal/src/ViewControllers/DebugUI/DebugUIMisc.m
```diff
@@ -115,6 +122,28 @@ + (void)clearHasDismissedOffers
         }];
 }
 
+// Creates a new group (by cloning the current group) without informing the,
+// other members. This can be used to test "group info requests", etc.
```
charlesmchen: The first of many "stress test" debug UI methods.

SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
```diff
@@ -102,7 +102,7 @@ - (void)runWithAttachmentHandler:(void (^)(TSAttachmentStream *attachmentStream)
 
     // If there is an attachment + text, render the text here, as Signal-iOS renders two messages.
     if (attachmentsProcessor.hasSupportedAttachments && transcript.body && ![transcript.body isEqualToString:@""]) {
-        // render text *after* the attachment
+        // We want the text to appear after the attachment.
         uint64_t textMessageTimestamp = transcript.timestamp + 1;
```
charlesmchen: We often used `timestamp + 1` to place a new interaction directly after another to which it is related.

* Sometimes we forgot to add 1, I've tried to address that in this PR since we sometimes look up interactions by timestamp.

SignalServiceKit/src/Messages/Interactions/TSInteraction.m
```diff
@@ -14,6 +14,8 @@ @implementation TSInteraction
 
 + (instancetype)interactionForTimestamp:(uint64_t)timestamp
                         withTransaction:(YapDatabaseReadWriteTransaction *)transaction {
+    OWSAssert(timestamp > 0);
```
charlesmchen: Some message types have a timestamp of zero (like group updates IIRC).

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
```diff
@@ -263,16 +278,27 @@ - (void)databaseViewRegistrationComplete
 
 #pragma mark - instance methods
 
+- (dispatch_queue_t)serialQueue
+{
+    static dispatch_queue_t queue = nil;
```
charlesmchen: It turns out that using dedicated serial queues in the message decrypter and message batch processor queues is necessary.  It allows us to skip a `dispatch_async()` when processing many messages and takes load off the main thread.

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
```diff
@@ -300,32 +326,35 @@ - (void)drainQueueWorkStep
...
+    DDLogVerbose(@"%@ completed %zd jobs. %zd jobs left.",
+        self.tag,
+        jobs.count,
+        [OWSMessageContentJob numberOfKeysInCollection]);
+
+    // Wait a bit in hopes of increasing the batch size.
+    // This delay won't affect the first message to arrive when this queue is idle,
+    // so by definition we're receiving more than one message and can benefit from
+    // batching.
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.1f * NSEC_PER_SEC)), self.serialQueue, ^{
```
charlesmchen: With a delay of 0.1 seconds, we tend to get batch sizes of 1-2.  Larger batches would be better, but decryption is much more expensive than message processing, so the message processing pipeline never accumulates a larger batch.

SignalServiceKit/src/Messages/OWSMessageDecrypter.m
```diff
@@ -247,7 +247,7 @@ - (void)decryptEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
             NSData *plaintextData = [[cipher decrypt:cipherMessage] removePadding];
             successBlock(plaintextData);
         } @catch (NSException *exception) {
-            dispatch_async(dispatch_get_main_queue(), ^{
+            dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
charlesmchen: This shouldn't be done on the main thread.

SignalServiceKit/src/Messages/OWSMessageDecrypter.m
```diff
@@ -306,7 +305,10 @@ - (void)processException:(NSException *)exception envelope:(OWSSignalServiceProt
 - (void)notifyForErrorMessage:(TSErrorMessage *)errorMessage withEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
 {
     TSThread *contactThread = [TSContactThread getOrCreateThreadWithContactId:envelope.source];
-    [[TextSecureKitEnv sharedEnv].notificationsManager notifyUserForErrorMessage:errorMessage inThread:contactThread];
+    dispatch_async(dispatch_get_main_queue(), ^{
```
charlesmchen: This should be done on the main thread.

SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -292,15 +294,16 @@ - (void)handleIncomingEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
...
                 [[OWSSyncGroupsRequestMessage alloc] initWithThread:thread groupId:groupId];
             [self.messageSender sendMessage:syncGroupsRequestMessage
-                transaction:transaction
                 success:^{
                     DDLogWarn(@"%@ Successfully sent Request Group Info message.", self.tag);
                 }
                 failure:^(NSError *error) {
                     DDLogError(@"%@ Failed to send Request Group Info message with error: %@", self.tag, error);
                 }];
+            return;
```
charlesmchen: We shouldn't try to process group message for unknown groups below.

SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -752,7 +752,6 @@ - (TSIncomingMessage *_Nullable)handleReceivedEnvelope:(OWSSignalServiceProtosEn
                                                             image:nil
                                                           groupId:dataMessage.group.id];
         TSGroupThread *gThread = [TSGroupThread getOrCreateThreadWithGroupModel:model transaction:transaction];
-        [gThread saveWithTransaction:transaction];
```
charlesmchen: No need to `saveWithTransaction:` after `getOrCreateThreadWithGroupModel:`

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -105,27 +106,16 @@ - (instancetype)initWithDBConnection:(YapDatabaseConnection *)dbConnection
     return self;
 }
 
-- (NSArray<OWSMessageDecryptJob *> *)nextJobsForBatchSize:(NSUInteger)maxBatchSize
+- (OWSMessageDecryptJob *_Nullable)nextJob
 {
-    NSMutableArray<OWSMessageDecryptJob *> *jobs = [NSMutableArray new];
+    __block OWSMessageDecryptJob *_Nullable job = nil;
```
charlesmchen: As discussed offline, we can't batch decryption of messages - it risks data loss.  After decryption we have to persist the output ASAP to avoid data loss.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -260,14 +250,24 @@ - (void)databaseViewRegistrationComplete
 
 #pragma mark - instance methods
 
+- (dispatch_queue_t)serialQueue
```
charlesmchen: Here too there are real perf wins to using a dedicated serial queue instead of the main queue.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -285,75 +285,48 @@ - (void)drainQueue
...
-    OWSAssert(unprocessedJobs.count <= jobs.count);
+    OWSSignalServiceProtosEnvelope *envelope = job.envelopeProto;
+    [self.messageDecrypter decryptEnvelope:envelope
+        successBlock:^(NSData *_Nullable plaintextData) {
 
-    if (unprocessedJobs.count < 1) {
-        for (OWSMessageDecryptJob *job in jobs) {
-            NSData *_Nullable plaintextData = plaintextDataMap[job.uniqueId];
+            // We can't decrypt the same message twice, so we need to persist
+            // the decrypted envelope data ASAP to prevent data loss.
```
charlesmchen: This code got a lot simpler.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -425,32 +426,35 @@ - (void)sendMessage:(TSOutgoingMessage *)message
...
-        transaction:(YapDatabaseReadWriteTransaction *_Nullable)transaction
-            success:(void (^)())successHandler
-            failure:(void (^)(NSError *error))failureHandler
-{
     OWSAssert(message);
 
-    if (transaction) {
-        [message updateWithMessageState:TSOutgoingMessageStateAttemptingOut transaction:transaction];
-    } else {
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
charlesmchen: There's a few steps to sending a message - try to do as many of them as possible off the main thread.  They need to be performed serially, so we get a bit of "callback hell".

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -425,32 +426,35 @@ - (void)sendMessage:(TSOutgoingMessage *)message
...
+        // This method will use a read/write transaction. This transaction
+        // will block until any open read/write transactions are complete.
+        //
+        // That's key - we don't want to send any messages in response
+        // to an incoming message until processing of that batch of messages
+        // is complete.
```
charlesmchen: This comment is important.

SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -266,12 +266,14 @@ - (void)becomeConsistentWithConfigurationForMessage:(TSMessage *)message
         TSIncomingMessage *incomingMessage = (TSIncomingMessage *)message;
         NSString *contactName = [contactsManager displayNameForPhoneIdentifier:incomingMessage.authorId];
 
-        [[[OWSDisappearingConfigurationUpdateInfoMessage alloc] initWithTimestamp:message.timestamp
+        // We want the info message to appear after the message.
+        [[[OWSDisappearingConfigurationUpdateInfoMessage alloc] initWithTimestamp:message.timestamp + 1
```
michaelkirk: TIOLI: I think it makes more sense for the info message to come before the text

e.g. if Alice:
- sends "Foo" 
- goes into settings and enables Disappearing messages
- sends "Bar"
- sends "Baz"

Bob will see:

- "Foo"
- "Alice enabled Disappearing Messages"
- "Bar (timer)"
- "Baz (timer)

To me it seems  a little weird that we'd build a scenario where Bob sees:

- "Foo"
- "Bar (timer)"
- "Alice enabled Disappearing Messages"
- "Baz (timer)

SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -191,7 +191,9 @@ - (void)handleDeliveryReceipt:(OWSSignalServiceProtosEnvelope *)envelope
         TSOutgoingMessage *outgoingMessage = (TSOutgoingMessage *)interaction;
         [outgoingMessage updateWithWasDeliveredWithTransaction:transaction];
     } else {
-        OWSFail(@"%@ Unexpected message with timestamp: %llu", self.tag, envelope.timestamp);
+        // Desktop currently sends delivery receipts for "unpersisted" messages
+        // like group updates, so these errors are expected to a certain extent.
+        DDLogError(@"%@ Unexpected message with timestamp: %llu", self.tag, envelope.timestamp);
```
michaelkirk: Oh right - I suppose all clients do, since that's currently a server thing. Maybe DDLogInfo or even Debug.

SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -292,15 +294,16 @@ - (void)handleIncomingEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
             OWSSyncGroupsRequestMessage *syncGroupsRequestMessage =
                 [[OWSSyncGroupsRequestMessage alloc] initWithThread:thread groupId:groupId];
             [self.messageSender sendMessage:syncGroupsRequestMessage
-                transaction:transaction
```
michaelkirk: why not use the same transaction?

SignalServiceKit/src/Messages/OWSMessageSender.m
michaelkirk: So we're using the side effect of yap's one-write-transaction-per-db as a semaphor here?

I can imagine, generally speaking, the problems this intends to address, but do you have a specific example? This comment is more "what and how" than "why".

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -1324,12 +1328,14 @@ - (void)saveGroupMessage:(TSOutgoingMessage *)message inThread:(TSThread *)threa
         // TODO: Why is this necessary?
         [message save];
     } else if (message.groupMetaMessage == TSGroupMessageQuit) {
-        [[[TSInfoMessage alloc] initWithTimestamp:message.timestamp
+        // We want the info message to appear after the message.
```
michaelkirk: I don't think these two offsets are necessary, since, if it's not TSGroupMessageDeliver, there is no corresponding visible "message" to come "after". 

No immediate problems with offsetting these timestamps by 1ms. The only thing I could think of is it might make debugging trickier down the road when trying to cross reference something stored in the DB with a log. No biggie if you'd rather keep them like this for some reason.

SignalServiceKit/src/Messages/Interactions/TSInteraction.m
charlesmchen: Yup, this assert is meant to start catching those.

charlesmchen: I'm going to add another similar assert in the `TSInteraction` initializer to determine if this is indeed an issue.

SignalServiceKit/src/Messages/OWSMessageSender.m
charlesmchen: Right. I've elaborated the comment!

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
```diff
@@ -263,16 +278,27 @@ - (void)databaseViewRegistrationComplete
 
 #pragma mark - instance methods
 
+- (dispatch_queue_t)serialQueue
+{
+    static dispatch_queue_t queue = nil;
+    static dispatch_once_t onceToken;
```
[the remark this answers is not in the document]
charlesmchen: Making it a class method wouldn't change the behavior?  It'd still be a single queue shared between all instances of this (usually singleton) class, which is the intention.

SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
charlesmchen: Good point! Done.

SignalServiceKit/src/Messages/OWSMessageManager.m
charlesmchen: Done.

SignalServiceKit/src/Messages/OWSMessageManager.m
charlesmchen: See the big huge comment (that we discussed above) in `OWSMessageSender`.

SignalServiceKit/src/Messages/OWSMessageSender.m

charlesmchen:
https://github.com/WhisperSystems/Signal-iOS/commit/77e0c9664c413a1a44ebea0a3bf2e0d1e1e90576

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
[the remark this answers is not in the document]
michaelkirk: Right, no change in behavior. Making it a class method would just make this limitation clear.

PR #2563, opened 2017-09-22
Hide all read receipts in UI if not enabled.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
> Since this is called all the time and on the main thread, anything we can easily do to avoid having the underlying accessor synchronized on the profile manager? I guess we can wait and see if its problematic.

Great point. I think we can safely remove this synchronization.  I can't imagine a scenario in which a race would occur.

charlesmchen:
PTAL @michaelkirk

michaelkirk:
LGTM

PR #2570, opened 2017-09-27
Fix class rename.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
+++ b/SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
@@ -24,20 +24,14 @@ NS_ASSUME_NONNULL_BEGIN
 @interface OWSMessageContentJob : TSYapDatabaseObject
 
 @property (nonatomic, readonly) NSDate *createdAt;
+@property (nonatomic, readonly) NSData *envelopeData;
+@property (nonatomic, readonly, nullable) NSData *plaintextData;
 
 - (instancetype)initWithEnvelopeData:(NSData *)envelopeData
                        plaintextData:(NSData *_Nullable)plaintextData NS_DESIGNATED_INITIALIZER;
+- (nullable instancetype)initWithCoder:(NSCoder *)coder NS_DESIGNATED_INITIALIZER;
 - (instancetype)initWithUniqueId:(NSString *)uniqueId NS_UNAVAILABLE;
 - (OWSSignalServiceProtosEnvelope *)envelopeProto;
-
-@end
-
-#pragma mark -
-
-@interface OWSMessageContentJob ()
-
-@property (nonatomic, readonly) NSData *envelopeData;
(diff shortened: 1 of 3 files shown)


SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -103,6 +98,8 @@ - (instancetype)initWithDBConnection:(YapDatabaseConnection *)dbConnection
 
     _dbConnection = dbConnection;
 
+    [NSKeyedUnarchiver setClass:[OWSMessageDecryptJob class] forClassName:[OWSMessageDecryptJob collection]];
```
michaelkirk: Can you leave a comment in this code?

SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -187,10 +187,23 @@ @implementation OWSUnarchiverDelegate
 
 - (nullable Class)unarchiver:(NSKeyedUnarchiver *)unarchiver cannotDecodeObjectOfClassName:(NSString *)name originalClasses:(NSArray<NSString *> *)classNames
 {
+    DDLogVerbose(@"%@ Could not decode object: %@", self.tag, name);
```
michaelkirk: Might as well make this DDLogError, as the details would be helpful in production.

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
```diff
@@ -75,8 +74,8 @@ - (OWSSignalServiceProtosEnvelope *)envelopeProto
 
 #pragma mark - Finder
 
-NSString *const OWSMessageContentJobFinderExtensionName = @"OWSBatchMessageProcessingFinderExtensionName";
-NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSBatchMessageProcessingFinderExtensionGroup";
+NSString *const OWSMessageContentJobFinderExtensionName = @"OWSMessageContentJobFinderExtensionName";
```
michaelkirk: This will trigger rebuilding the extension, but it should be pretty quick, just the pending jobs, right?

charlesmchen:
> Is this relevant to clients in the wild, or is this to fix up your debug devices?

It's relevant to clients in the wild - we've renamed a class, so we need to ensure that YDB knows how to decrypt them properly.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
charlesmchen: Done.

SignalServiceKit/src/Storage/TSStorageManager.m
charlesmchen: Done.

charlesmchen:
PTAL @michaelkirk

michaelkirk:
LGTM

PR #2572, opened 2017-09-27
Modify message metadata view to observe DB changes.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessageMetadataViewController.swift
+++ b/Signal/src/ViewControllers/MessageMetadataViewController.swift
@@ -6,11 +6,14 @@ import Foundation
 
 class MessageMetadataViewController: OWSViewController {
 
+    static let TAG = "[MessageMetadataViewController]"
     let TAG = "[MessageMetadataViewController]"
 
     // MARK: Properties
 
-    let message: TSMessage
+    let databaseConnection: YapDatabaseConnection
+
+    var message: TSMessage
 
     var mediaMessageView: MediaMessageView?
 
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/MessageMetadataViewController.swift
```diff
@@ -403,4 +428,39 @@ class MessageMetadataViewController: OWSViewController {
...
+    }
+
+    internal func yapDatabaseModified(notification: NSNotification) {
+//        Logger.info("\(TAG) in \(#function)")
+        AssertIsOnMainThread()
+
+        let notifications = self.databaseConnection.beginLongLivedReadTransaction()
+
+// TODO: I can't figure out what the !@#$%&^* auto-generated Swift interface to this method is.
+//        guard self.databaseConnection.hasChange(forKey:message.uniqueId,
```
charlesmchen: @michaelkirk can you please try to autocomplete this statement?  I can't generate the swift interface for YapDBConnection.

michaelkirk:
```
    open func hasChange(forKey key: String, inCollection collection: String, in notifications: [Notification]) -> Bool
```

I put the rest of that class here if you need it: https://gist.github.com/michaelkirk/224d32429358abf996e58b7eb31d03f8

Xcode **did** crash the first 3 times I tried it =(

charlesmchen: THANK YOU

PR #2578, opened 2017-09-28
Simplify processing of messages from linked devices.

charlesmchen:
PTAL @michaelkirk


SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
```diff
@@ -34,14 +36,16 @@ - (instancetype)initWithIncomingSentMessageTranscript:(OWSIncomingSentMessageTra
                                          messageSender:[TextSecureKitEnv sharedEnv].messageSender
```
michaelkirk: can we remove the message sender dependency now?

charlesmchen: Great idea.

PR #2581, opened 2017-09-28
Fix assert for empty thread

michaelkirk:
PTAL @charlesmchen


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4075,6 +4075,10 @@ - (void)markVisibleMessagesAsRead
 
     uint64_t lastVisibleTimestamp = self.lastVisibleTimestamp;
```
charlesmchen: Should we make lastVisibleTimestamp a `NSNumber *`?

michaelkirk: We could - but making this change would clutter up a few places where we (correctly) rely on the nil == 0 for concise assignment. I'm not sure it's worth it.

PR #2582, opened 2017-09-28
Don't mark messages sent from linked devices as delivered.

charlesmchen:
IMO we shouldn't be doing this. Thoughts?

PTAL @michaelkirk


michaelkirk:
Now that we persist early delivery receipts there's no downside, right?

LGTM

michaelkirk:
Oh - wait we don't yet persist early delivery receipts...

I'd say this is of dubious value without that. I get that the current implementation has problems, but I think removing it would be worse in the average case (messages are usually delivered).

michaelkirk-signal:
Looks like this was effectively merged with the recipient status refactor, but with dubious value.

PR #2584, opened 2017-09-28
Show proper thread from notification

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/environment/Environment.m
+++ b/Signal/src/environment/Environment.m
@@ -215,6 +215,19 @@ static Environment *environment = nil;
     });
 }
 
++ (void)presentConversationForThreadId:(NSString *)threadId
+{
+    OWSAssert(threadId.length > 0);
+
+    TSThread *thread = [TSThread fetchObjectWithUniqueID:threadId];
+    if (thread == nil) {
+        OWSFail(@"%@ unable to find thread with id: %@", self.tag, threadId);
+        return;
+    }
+
+    [self presentConversationForThread:thread];
+}
+
 + (void)presentConversationForThread:(TSThread *)thread
 {
     [self presentConversationForThread:thread keyboardOnViewAppearing:YES callOnViewAppearing:NO];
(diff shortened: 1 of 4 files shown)


Signal/src/environment/Environment.m
```diff
@@ -215,6 +215,24 @@ + (void)presentConversationForRecipientId:(NSString *)recipientId
...
 }
 
++ (void)presentConversationForThreadId:(NSString *)threadId
+{
+    OWSAssert(threadId.length > 0);
+
+    __block TSThread *thread;
+    [[TSStorageManager sharedManager].dbReadWriteConnection
+        readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
+            thread = [TSThread fetchObjectWithUniqueID:threadId];
```
charlesmchen: What?  Shouldn't this be `[fetchObjectWithUniqueID: transaction:]` or ... why not just use `[fetchObjectWithUniqueID:]` without the wrapping transaction?

Signal/src/network/PushManager.m
```diff
@@ -231,12 +234,21 @@ - (void)application:(UIApplication *)application
...
+        }
         completionHandler();
     } else {
         OWSFail(@"%@ Unhandled action with identifier: %@", self.tag, identifier);
         NSString *threadId = notification.userInfo[Signal_Thread_UserInfo_Key];
-        [Environment presentConversationForRecipientId:threadId];
+        if (threadId) {
+            [Environment presentConversationForThreadId:threadId];
+        } else {
+            OWSFail(@"%@ threadId was unexpectedly nil in action with identifier: %@", self.tag, identifier);
```
charlesmchen: Why not DRY up this nil test into `presentConversationForThreadId:`?

michaelkirk: A failure here likely indicates malformed notification params, which, being a K-V bag is easy to mess up, so I wanted to log the action as debug context.

Signal/src/environment/Environment.m
michaelkirk: thank you!

Signal/src/network/PushManager.m
charlesmchen: Makes sense.

PR #2585, opened 2017-09-28
Fix "mark as read on linked device".

charlesmchen:
PTAL @michaelkirk


SignalServiceKit/src/Messages/OWSReadReceiptManager.m
```diff
@@ -473,8 +473,8 @@ - (void)markAsReadOnLinkedDevice:(TSIncomingMessage *)message
 {
     OWSAssert(message);
     OWSAssert(transaction);
-    
-    [self markAsReadBeforeTimestamp:message.timestamp
```
michaelkirk: Since this was not obvious to me, can you leave a comment to the affect of:

// use the timestamp reflected by local sorting, rather than the local users timestamp at send-time.

charlesmchen: Done.

PR #2586, opened 2017-09-28
Mkirk/bubble metadata view

michaelkirk:
This doesn't change the presentation of media views. PTAL @charlesmchen 
screen shot 2017-09-28 at 2 34 28 pm

screen shot 2017-09-28 at 2 34 58 pm

screen shot 2017-09-28 at 2 37 23 pm
screen shot 2017-09-28 at 2 37 31 pm

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -1541,7 +1541,8 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
 
 + (JSQMessagesBubbleImageFactory *)sharedBubbleImageFactory
 {
-
+    AssertIsOnMainThread();
+    
     static JSQMessagesBubbleImageFactory *bubbleImageFactory;
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
@@ -1553,6 +1554,8 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
 
 + (JSQMessagesBubbleImage *)outgoingBubbleImageData
 {
+    AssertIsOnMainThread();
+    
     static JSQMessagesBubbleImage *bubbleImage;
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
@@ -1565,6 +1568,8 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
 
 + (JSQMessagesBubbleImage *)incomingBubbleImageData
 {
+    AssertIsOnMainThread();
+    
     static JSQMessagesBubbleImage *bubbleImage;
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
@@ -1577,6 +1582,8 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
 
 + (JSQMessagesBubbleImage *)currentlyOutgoingBubbleImageData
 {
+    AssertIsOnMainThread();
+    
     static JSQMessagesBubbleImage *bubbleImage;
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1541,15 +1536,66 @@ - (void)initializeCollectionViewLayout
...
-        [bubbleFactory incomingMessagesBubbleImageWithColor:[UIColor jsq_messageBubbleLightGrayColor]];
-    self.outgoingBubbleImageData = [bubbleFactory outgoingMessagesBubbleImageWithColor:[UIColor ows_materialBlueColor]];
-    self.currentlyOutgoingBubbleImageData =
-        [bubbleFactory outgoingMessagesBubbleImageWithColor:[UIColor ows_fadedBlueColor]];
-    self.outgoingMessageFailedImageData = [bubbleFactory outgoingMessagesBubbleImageWithColor:[UIColor grayColor]];
+}
+
++ (JSQMessagesBubbleImageFactory *)sharedBubbleImageFactory
+{
+
```
charlesmchen: Should we assert that we're on the main thread in these accessors?  I _think_ we only need them on the main thread while configuring views and these factories might not be thread-safe.

Signal/src/ViewControllers/MessageMetadataViewController.swift
```diff
@@ -280,13 +280,24 @@ class MessageMetadataViewController: OWSViewController {
...
+                bodyLabel.autoPinLeadingToSuperView(withMargin: leadingMargin)
+                bodyLabel.autoPinTrailingToSuperView(withMargin: trailingMargin)
+                bodyLabel.autoPinHeightToSuperview(withMargin: 10)
+
+                // Try to hug content both horizontally and vertically, but *prefer* wide and short, to narrow and tall.
+                // While never exceeding max width, and never cropping content.
+                bodyLabel.setContentHuggingPriority(UILayoutPriorityDefaultLow, for: .horizontal)
+                bodyLabel.setContentHuggingPriority(UILayoutPriorityDefaultHigh, for: .vertical)
+                bodyLabel.setContentCompressionResistancePriority(UILayoutPriorityRequired, for: .vertical)
+                bodyLabel.autoSetDimension(.width, toSize:ScaleFromIPhone5(210), relation:.lessThanOrEqual)
```
charlesmchen: Is 210 a JSQ constant?

Signal/src/ViewControllers/MessageMetadataViewController.swift
```diff
@@ -280,13 +280,24 @@ class MessageMetadataViewController: OWSViewController {
                 bodyLabel.numberOfLines = 10
                 bodyLabel.lineBreakMode = .byWordWrapping
 
-                let bubbleView = UIView()
-                bubbleView.backgroundColor = isIncoming ? UIColor.jsq_messageBubbleLightGray() : UIColor.ows_materialBlue()
+                let bubbleImageData = isIncoming ? ConversationViewController.incomingBubbleImageData() : ConversationViewController.outgoingBubbleImageData()
+                let leadingMargin: CGFloat = isIncoming ? 15 : 10
+                let trailingMargin: CGFloat = isIncoming ? 10 : 15
```
charlesmchen: Will this be right for RTL?

Signal/src/ViewControllers/MessageMetadataViewController.swift
michaelkirk: It is only a crude approximate thereof. Actual message bubble width is determined based on the collection view, WRT the collection view insets and device width.

Signal/src/ViewControllers/MessageMetadataViewController.swift
michaelkirk: Good catch. It was in line with what we were doing in conversation view, but our bubbles in conversation view were already weird for RTL. I've addressed that in 
https://github.com/WhisperSystems/Signal-iOS/pull/2586/commits/d202b7ce3fae3b9089868bd3304425285fdbdd2e

PTAL @charlesmchen

michaelkirk: screen shot 2017-09-28 at 3 24 14 pm
screen shot 2017-09-28 at 3 24 08 pm

michaelkirk: Actually, I'm going to merge this as it was, with your CR, but then open a new pr for the RTL bubble fix.

PR #2588, opened 2017-09-28
Check all l10n preconditions up front

michaelkirk:
PTAL @charlesmchen


michaelkirk:
In b2a38323b10cede4f63507bc37e35a3d0e256d66

PR #2593, opened 2017-09-29
Charlesmchen/bigger message batches

charlesmchen:
Processing still noticeable on slower devices; this should help.

PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
+++ b/SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
@@ -338,7 +338,7 @@ NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSMessageContentJo
     // This delay won't affect the first message to arrive when this queue is idle,
     // so by definition we're receiving more than one message and can benefit from
     // batching.
-    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.1f * NSEC_PER_SEC)), self.serialQueue, ^{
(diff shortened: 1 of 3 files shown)


SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
```diff
@@ -338,7 +341,7 @@ - (void)drainQueueWorkStep
     // This delay won't affect the first message to arrive when this queue is idle,
     // so by definition we're receiving more than one message and can benefit from
     // batching.
-    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.1f * NSEC_PER_SEC)), self.serialQueue, ^{
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(1.f * NSEC_PER_SEC)), self.serialQueue, ^{
```
michaelkirk: Doesn't this max us out at 16 messages/second vs the previous 100 messages/second? That seems unnecessarily slow e.g. for startup with a big backlog.

charlesmchen: How about 32/0.5 seconds = 64 messages/sec.?  That's 4x faster.  Will churn through 1,000 messages in 15 seconds.

michaelkirk: That sounds more reasonable - we can continue to hone this as we go.

PR #2594, opened 2017-09-29
Refine read receipt processing cycle.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSReadReceiptManager.m
+++ b/SignalServiceKit/src/Messages/OWSReadReceiptManager.m
@@ -221,7 +221,6 @@ NSString *const OWSReadReceiptManagerAreReadReceiptsEnabled = @"areReadReceiptsE
     {
         DDLogVerbose(@"%@ Processing read receipts.", self.tag);
 
-        // OWSReadReceipt
         NSArray<OWSLinkedDeviceReadReceipt *> *readReceiptsForLinkedDevices =
             [self.toLinkedDevicesReadReceiptMap allValues];
         [self.toLinkedDevicesReadReceiptMap removeAllObjects];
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Messages/OWSReadReceiptManager.m
```diff
@@ -210,17 +210,7 @@ - (void)scheduleProcessing
...
-            // We want a value high enough to allow us to effectively de-duplicate,
-            // read receipts without being so high that we risk not sending read
-            // receipts due to app exit.
-            const CGFloat kProcessingFrequencySeconds = 3.f;
-            dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(kProcessingFrequencySeconds * NSEC_PER_SEC)),
-                dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0),
-                ^{
-                    [self process];
-                });
+            [self process];
```
charlesmchen: We shouldn't delay "initial" read receipt processing, so that read receipts are emitted ASAP when user is interacting with a conversation.

We do want to wait between processing "passes" in cases where the user is reading a lot of messages in order to facilitate batching.

SignalServiceKit/src/Messages/OWSReadReceiptManager.m
```diff
@@ -231,9 +221,9 @@ - (void)process
     {
         DDLogVerbose(@"%@ Processing read receipts.", self.tag);
 
-        self.isProcessing = NO;
-
-        NSArray<OWSReadReceipt *> *readReceiptsForLinkedDevices = [self.toLinkedDevicesReadReceiptMap allValues];
```
charlesmchen: Fix rebase breakage here.

Issue #1812, opened 2017-02-26
Reauthencate device


Cj-Malone:
- [x] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description
I should say I'm unauthorized because I thought "Link devices" meant I could use the same phone number across devices, but it means a web browser. So I used a verification code on another device.

Now "network status" loops "connecting" and "offline." It seems like the app has never been designed to deal with this situation, I think I just need to get another verification prompt but it's never been brought up and I can't find a way to prompt it.

[Docs](https://support.whispersystems.org/hc/en-us/articles/213240927--Can-I-use-multiple-devices-Why-is-my-device-offline-) say I'd have to wipe the data to re register, but I don't want to lose my messages and because you can't [backup](https://github.com/WhisperSystems/Signal-iOS/issues/967) I was hoping re authorizing would be enough.

### Steps to reproduce
- verify on anther device
- original device is now unauthorized ( can be checked by going to settings > linked devices and a dialog will saying "401 unauthorized " )

**Actual result:** no prompt to reauthencate
**Expected result:** a prompt like a fresh install to be able to get a code and reauthencate

Cj-Malone:
You can force reauthincation within the Android app by switching Signal to just be the SMS app, then back to the Signal protocol. I know none of that is possible on iOS, but I do think Signal-iOS should be able to handle the server responding with a 401status code.

basnijholt:
I am experiencing the exact same issue.

I would like to preserve my chat history but cannot see how to do this without deleting the app.

I would be willing to work on this, would this (a re-authenticate button) be an easy feature to implement?

basnijholt:
I think this issue might be fixed by https://github.com/WhisperSystems/Signal-iOS/pull/2598, @michaelkirk could you comment?

alexforencich:
Does #2598 fix this?  If so, how is this feature accessed?

shiitake:
I have the same question. I use signal on my iPhone and tried to add it to my iPad and now I'm pretty much locked out from using it on the iPhone.

alexforencich:
Signal seems to have some serious user interface and architectural bugs with multi-device support.  If I had known about that earlier, I might have decided to use a different platform.  As the saying goes, security at the expense of usability is at the expense of security.

basnijholt:
[...]

PR #2598, opened 2017-09-29
Re-register without losing your messages in Debug-UI

michaelkirk:
As we talked about previously - eventually it would be good to expose this functionality to users when the server considers them "deauthed".

Though I'd probably want to smooth some things out before then - especially making the process abortable. Currently once you proceed, you're locked out until you re-register.

PTAL @charlesmchen


SignalServiceKit/src/Account/TSAccountManager.m
```diff
@@ -76,6 +77,17 @@ - (void)setPhoneNumberAwaitingVerification:(NSString *_Nullable)phoneNumberAwait
                                                            userInfo:nil];
 }
 
+- (void)resetForRegistration
+{
```
charlesmchen: Do we need to call this in `[AppSettingsViewController proceedToUnregistration]`?

charlesmchen: Or from `[TSAccountManager unregisterTextSecureWithSuccess]`?

charlesmchen:
Do we need to clear the identity key?

michaelkirk:
> I wonder if this will break any of our existing systems - e.g. where do we assume "once user is registered they will remain registered for the duration of this session."?

I couldn't think of anywhere else - but if it's any consolation:
1. The registration view blocks the user from interacting with the app until it's complete.
2. Part of the reason I wanted this as debug-only at this point was to ferret out these kinds of problems.

> Do we need to clear the identity key?

We specifically *don't want* to clear the identity key. This will allow Alice to recover from the scenario where she accidentally registers a second device temporarily.

basnijholt:
Would this PR solve this https://github.com/WhisperSystems/Signal-iOS/issues/1812 issue?

If so, that would be absolutely great and in that case, I will have to wait with using Signal again 😢 until this is merged.

michaelkirk:
> Would this PR solve this #1812 issue?

Not completely, but it's in pursuit of that. Currently this is only available for debug builds. Getting this ready for de-authed users is on the roadmap. Updates will be in #1812.

michaelkirk:
@charlesmchen  can you take another look?

I've added https://github.com/WhisperSystems/Signal-iOS/pull/2598/commits/e53459c029396e2e1aae43b06242589a355eb592 to ensure other people know to start a new session when contacting us post-re-register.

SignalServiceKit/src/Account/TSAccountManager.m
michaelkirk: Since those code paths take the more drastic step of removing the entire database, calling this would be redundant. 

The point of this feature is to remove the minimal amount of data necessary to get your device into shape for receiving/sending messages again, while retaining all your messages/attachments/identity store.

We reset the session store because it's possible our local session store is out of sync. This would happen if we'd activated another device while *this* device was deactivated.

michaelkirk:
> @charlesmchen can you take another look?

Ping. I've just rebased against master.

michaelkirk:
In ef80b53f

PR #2602, opened 2017-10-02
Charlesmchen/fix is contact test

charlesmchen:
PTAL @michaelkirk


charlesmchen:
My question is where else we should apply this.  I guess the conversation settings view?

michaelkirk:
In 2292f20c9341a365614360ba392cf4fb67b3b7d2

PR #2605, opened 2017-10-03
Sync read receipt configuration upon set

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Messages/OWSReadReceiptManager.m
+++ b/SignalServiceKit/src/Messages/OWSReadReceiptManager.m
@@ -560,8 +560,22 @@ NSString *const OWSReadReceiptManagerAreReadReceiptsEnabled = @"areReadReceiptsE
     // We don't need to worry about races around this cached value.
     if (!self.areReadReceiptsEnabledCached) {
         // Default to NO.
-        self.areReadReceiptsEnabledCached = @([self.dbConnection boolForKey:OWSReadReceiptManagerAreReadReceiptsEnabled
-                                                               inCollection:OWSReadReceiptManagerCollection]);
+        self.areReadReceiptsEnabledCached =
+            @([self.dbConnection boolForKey:OWSReadReceiptManagerAreReadReceiptsEnabled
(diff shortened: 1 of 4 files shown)


SignalServiceKit/src/Messages/OWSReadReceiptManager.m
```diff
@@ -568,11 +569,22 @@ - (BOOL)areReadReceiptsEnabled
 
 - (void)setAreReadReceiptsEnabled:(BOOL)value
 {
-    DDLogInfo(@"%@ areReadReceiptsEnabled: %d.", self.tag, value);
```
charlesmchen: Looks like `areReadReceiptsEnabled` can open a transaction inside a transaction. See how `OWSMessageManager` uses it, for example.  I think we may need to load this cached value explicitly on app launch.

charlesmchen: I've also wary of introducing deadlocks in this manager as we did with the portfolio manager.

michaelkirk:
>  This change LGTM but we need to address these other issues.

Addressed in  https://github.com/WhisperSystems/Signal-iOS/pull/2605/commits/2125dbe72ca793375a9aca64a29f06eea4e7294e

SignalServiceKit/src/Messages/OWSReadReceiptManager.m
```diff
@@ -559,20 +560,45 @@ - (BOOL)areReadReceiptsEnabled
...
+            @([self.dbConnection boolForKey:OWSReadReceiptManagerAreReadReceiptsEnabled
+                               inCollection:OWSReadReceiptManagerAreReadReceiptsEnabled]);
+    }
+
+    return [self.areReadReceiptsEnabledCached boolValue];
+}
+
+- (BOOL)areReadReceiptsEnabledWithTransaction:(YapDatabaseReadTransaction *)transaction
+{
+    if (!self.areReadReceiptsEnabledCached) {
```
charlesmchen: Maybe add a comment about how we don't need to synchronize access to this property since, this state can only be edited locally by the user, edits will be very rare, and races don't have serious side effects.

PR #2610, opened 2017-10-04
Ignore group info requests if sender and recipient aren't both known …

charlesmchen:
…group members.

PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSMessageManager.m
+++ b/SignalServiceKit/src/Messages/OWSMessageManager.m
@@ -811,11 +811,21 @@ NS_ASSUME_NONNULL_BEGIN
         return;
     }
 
+    // Ensure sender is in the group.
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -811,11 +811,21 @@ - (void)handleGroupInfoRequest:(OWSSignalServiceProtosEnvelope *)envelope
         return;
     }
 
+    // Ensure sender is in the group.
     if (![gThread.groupModel.groupMemberIds containsObject:envelope.source]) {
         DDLogWarn(@"%@ Ignoring 'Request Group Info' message for non-member of group. %@ not in %@",
             self.tag,
             envelope.source,
             gThread.groupModel.groupMemberIds);
+        return;
```
charlesmchen: Missing a return =(

PR #2615, opened 2017-10-04
webrtc mtl view

michaelkirk:
PTAL @charlesmchen 

- Take advantage of metalkit to drive remote video view where available.
- disable remote video on iOS8

**iOS9 32bit (using legacy):**
callscreen-ios932bit

**iOS8 64bit (using legacy):**
callscreen-ios8-64bit

**iOS11 64bit (using MetalKit)**
callscreen-ios11-64bit

**iOS10 32bit (using legacy):**
ios10-32bit


Signal/src/UIView+OWS.m
```diff
@@ -100,14 +100,19 @@ - (void)autoPinHeightToHeightOfView:(UIView *)view
 
 - (NSLayoutConstraint *)autoPinToSquareAspectRatio
 {
+    return [self autoPinToAspectRatio:1.0];
+}
+
+- (NSLayoutConstraint *)autoPinToAspectRatio:(CGFloat)ratio
+{
     self.translatesAutoresizingMaskIntoConstraints = NO;
```
charlesmchen: OWSAssert(ration > 0.f);
OWSAssert(ration < 1.f);

charlesmchen: We also might want to clamp this value so that it fails gracefully in production, e.g.

```
// Clamp to ensure view has reasonable aspect ratio.
ratio = MAX(0.05f, (MIN(0.95f, ratio));
```

Signal/src/call/CallAudioService.swift
```diff
@@ -418,7 +418,11 @@ struct AudioSource: Hashable {
 
         guard let availableInputs = session.availableInputs else {
             // I'm not sure when this would happen.
-            owsFail("No available inputs or inputs not ready")
+            if #available(iOS 9.0, *) {
```
charlesmchen: Isn't this test backwards?  I feel like the condition will be true on iOS >= 9.

Signal/src/views/RemoteVideoView.m
```diff
@@ -0,0 +1,219 @@
...
+    self = [super initWithFrame:frame];
+    if (!self) {
+        return self;
+    }
+
+    self.backgroundColor = UIColor.blackColor;
+
+    UILabel *label = [UILabel new];
+    label.numberOfLines = 0;
+    label.text
```
charlesmchen:
```
label.lineBreakMode = NSLineBreakByWordWrapping;
```

Signal/src/views/RemoteVideoView.m
```diff
@@ -0,0 +1,219 @@
...
+    if (!SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(9, 0)) {
+        _videoRenderer = [NullVideoRenderer new];
+        [self addSubview:_videoRenderer];
+        [_videoRenderer autoPinEdgesToSuperviewEdges];
+    }
+
+// Currently RTC only supports metal on 64bit machines
+#if defined(RTC_SUPPORTS_METAL)
+    // On 64-bit, iOS9+: uses the MetalKit backed view for improved battery/rendering performance.
+    if (_videoRenderer == nil) {
```
charlesmchen: why not wrap this in an else for the condition above?

charlesmchen: Ignore this comment.

Signal/src/views/RemoteVideoView.m
```diff
@@ -0,0 +1,219 @@
...
+
+        // It is insufficient to check the RTC_SUPPORTS_METAL macro to determine Metal support.
+        // RTCMTLVideoView requires the MTKView class, available only in iOS9+
+        // So check that it exists before proceeding.
+        if ([MTKView class]) {
+            _videoRenderer = [[RTCMTLVideoView alloc] initWithFrame:CGRectZero];
+            [self addSubview:_videoRenderer];
+            [_videoRenderer autoPinEdgesToSuperviewEdges];
+            // HACK: Although RTCMTLVideo view is positioned to the top edge of the screen
+            // It's inner (private) MTKView is below the status bar.
```
charlesmchen: So the `MTKView` has constraints but they're wrong?  Seems like we should remove those constraints or make sure that the constraints we add have very high priority.

Signal/src/views/RemoteVideoView.m
```diff
@@ -0,0 +1,219 @@
...
+        && containingView.bounds.size.height > 0) {
+
+        [constraints addObject:[videoView autoPinToAspectRatio:aspectRatio]];
+        [constraints addObject:[videoView autoSetDimension:ALDimensionWidth
+                                                    toSize:containingView.width
+                                                  relation:NSLayoutRelationGreaterThanOrEqual]];
+        [constraints addObject:[videoView autoSetDimension:ALDimensionHeight
+                                                    toSize:containingView.height
+                                                  relation:NSLayoutRelationGreaterThanOrEqual]];
+        [constraints addObjectsFromArray:[videoView autoCenterInSuperview]];
```
charlesmchen: What's to guarantee that `videoView` will have the smallest possible size that satisfies these constraints?

Signal/src/call/CallAudioService.swift
michaelkirk: I still am not sure why this would happen, so I'd like to be alerted by an assert when it occurs.

Currently it's only *known* to occur on iOS8, where I'd like to ignore the assertionFailure, but I'd still like to know when this assert fails on iOS9+

I'll try to make the comment clearer.

Signal/src/UIView+OWS.m
michaelkirk: Good call! Though we'd want a much large max ratio than .95 right? e.g. maybe something like 95.0?

Signal/src/views/RemoteVideoView.m
michaelkirk: done!

Signal/src/views/RemoteVideoView.m
michaelkirk: Done - added low priority "pin to superview edges" constraint.

Signal/src/UIView+OWS.m
charlesmchen: dur, you're right.  
```
// Clamp to ensure view has reasonable aspect ratio.
static const CGFloat kMinRatio = 0.05f;
ratio = MAX(kMinRatio, (MIN(1 / kMinRatio, ratio));
```

Signal/src/views/RemoteVideoView.m
charlesmchen: Still curious about this.

michaelkirk: I made their priority explicitly "required".

michaelkirk:
PTAL @charlesmchen

charlesmchen:
Please be sure to test on iOS 8, 9, 10 and 11.  I can help if you have a gap in your test devices.

michaelkirk:
> Please be sure to test on iOS 8, 9, 10 and 11. I can help if you have a gap in your test devices.

Added to the test plan. Do you have a 64bit iOS10 device @charlesmchen ?

michaelkirk:
In 098906882259690630d314052bac6c064409bac3

PR #2619, opened 2017-10-06
background push notifications

michaelkirk:
Ensure we register for remote notifications when launched from the background. 

The primary motivation for this change is not always reproducible, but I have a couple of compellign logs that indicate the app *is* launched, but doesn't receive the push notification delegate callback.

Documentation suggests that we register for notifications at launch, which we can't quite do, because we need the DB to be setup to determine if we're registered, but this get us closer.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -271,13 +271,13 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
 - (void)application:(UIApplication *)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken
 {
-    DDLogDebug(@"%@ registered vanilla push token: %@", self.tag, deviceToken);
+    DDLogInfo(@"%@ registered vanilla push token: %@", self.tag, deviceToken);
     [PushManager.sharedManager.pushNotificationFutureSource trySetResult:deviceToken];
 }
 
 - (void)application:(UIApplication *)application didFailToRegisterForRemoteNotificationsWithError:(NSError *)error
 {
-    DDLogDebug(@"%@ failed to register vanilla push token with error: %@", self.tag, error);
+    DDLogError(@"%@ failed to register vanilla push token with error: %@", self.tag, error);
 #ifdef DEBUG
     DDLogWarn(@"%@ We're in debug mode. Faking success for remote registration with a fake push identifier", self.tag);
     [PushManager.sharedManager.pushNotificationFutureSource trySetResult:[[NSMutableData dataWithLength:32] copy]];
@@ -290,7 +290,7 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 - (void)application:(UIApplication *)application
     didRegisterUserNotificationSettings:(UIUserNotificationSettings *)notificationSettings
 {
-    DDLogDebug(@"%@ registered user notification settings", self.tag);
+    DDLogInfo(@"%@ registered user notification settings", self.tag);
     [PushManager.sharedManager.userNotificationFutureSource trySetResult:notificationSettings];
 }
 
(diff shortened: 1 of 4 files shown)


charlesmchen:
> Documentation suggests that we register for notifications at launch, which we can't quite do, because we need the DB to be setup to determine if we're registered, but this get us closer.

Could we store an "is registered" flag in `NSUserDefaults`?

Signal/src/AppDelegate.m
```diff
@@ -271,23 +271,26 @@ - (void)setupEnvironment
...
 {
-    DDLogDebug(@"%@ Successfully registered for remote notifications with token: %@", self.tag, deviceToken);
+    DDLogDebug(@"%@ registered vanilla push token: %@", self.tag, deviceToken);
     [PushManager.sharedManager.pushNotificationFutureSource trySetResult:deviceToken];
 }
 
 - (void)application:(UIApplication *)application didFailToRegisterForRemoteNotificationsWithError:(NSError *)error
 {
-    OWSProdError([OWSAnalyticsEvents appDelegateErrorFailedToRegisterForRemoteNotifications]);
+    DDLogDebug(@"%@ failed to register vanilla push token with error: %@", self.tag, error);
```
charlesmchen: Should this be warn?

Signal/src/AppDelegate.m
```diff
@@ -819,6 +817,16 @@ - (void)databaseViewRegistrationComplete
...
     if ([TSAccountManager isRegistered]) {
         DDLogInfo(@"localNumber: %@", [TSAccountManager localNumber]);
+
+        // Fetch messages as soon as possible after launching. In particular, when
+        // launching from the background, without this, we end up waiting some extra
+        // seconds before receiving an actionable push notification.
+        [[Environment getCurrent].messageFetcherJob runAsync];
+
+        // This should happen at any launch, background or foreground.
+        __unused AnyPromise *promise = [OWSSyncPushTokensJob runWithPushManager:[PushManager sharedManager]
```
charlesmchen: why assign to a local variable?

charlesmchen: I thought it was unsafe to do this before `didRegisterUserNotificationSettings:`.  Maybe we should do this in both places, i.e.:

* Start `OWSSyncPushTokensJob` in `didRegisterUserNotificationSettings:` if database view registration is complete.
* Start `OWSSyncPushTokensJob` in `databaseViewRegistrationComplete:` if `didRegisterUserNotificationSettings` has already been called (we'll need a new flag).

michaelkirk: SyncPushTokensJob already makes sure to call `registerUserNotificationSettings`

michaelkirk: This is more firmly established in the followup PR #2620

michaelkirk:
> Could we store an "is registered" flag in NSUserDefaults?

We could do this, but I'm not sure it would solve any problems for us, since ultimately we'll need to access storage anyway, at least to store the new tokens and to access our server credentials (if not other things).

Signal/src/AppDelegate.m
michaelkirk: It's the simplest way, in tandem with the __unused macro to avoid a warning here. Is there a better way?

Signal/src/AppDelegate.m
michaelkirk: Sure!

Signal/src/AppDelegate.m
```diff
@@ -271,23 +271,26 @@ - (void)setupEnvironment
 
 - (void)application:(UIApplication *)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken
 {
-    DDLogDebug(@"%@ Successfully registered for remote notifications with token: %@", self.tag, deviceToken);
+    DDLogDebug(@"%@ registered vanilla push token: %@", self.tag, deviceToken);
```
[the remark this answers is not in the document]
michaelkirk: Sure!

michaelkirk:
PTAL @charlesmchen

Signal/src/AppDelegate.m
charlesmchen: Or maybe we can have a `tryToSyncPushTokens` method that we can call from both places that only works if both have completed.

michaelkirk:
In ede89a7

PR #2620, opened 2017-10-06
Rework push registration

michaelkirk:
PR is based on  #2619, so if you have blocking concerns with #2619, hold of on reviewing this, otherwise PTAL @charlesmchen.

== Account Registration ==

Not complete until push tokens are uploaded

== Remote Notifications Registration ==

Extracted from PushManager

- wait for notification-settings registration to complete before
  requesting push tokens, otherwise it's possible token requests will
  be ignored.

- Less state required for push notification callbacks, specifically, we
  no longer need to ensure we've created a promise before the
  registration delegate methods get called.

- no more TOCFuture in Signal-iOS (still in SSK for now). It's not in
  cases of inexplicable behavior - one a recently, push notification
  premature free, in redphone, and more popular use, and I've seen two
  futures inexplicably being nil.  Instead, let's consolidate around
  PromiseKit for popularly used, maintained, strongly-typed futures.

- separate logic for registering for vanilla push/voip notifications
  (few dependencies) from responding to UILocalNotifications (lots of
  dependencies). Ultimately I'd like to consolidate the remaining
  UILocalNotifications logic with the existing NotificationsManager

== Misc ==

more debug logging
more uniform logging
remove stale logic around newly registered user

// FREEBIE


SignalServiceKit/src/Account/TSAccountManager.m
```diff
@@ -290,11 +290,6 @@ - (void)verifyAccountWithCode:(NSString *)verificationCode
                                                                                forNumber:phoneNumber
                                                                             signalingKey:signalingKey
                                                                                  authKey:authToken];
-    void (^completedRegistrationBlock)() = ^{
-        [self didRegister];
-        [TSSocketManager requestSocketOpen];
```
michaelkirk: 1. didRegister is now called only after PushTokens have been uploaded.
2. no need to requestSocketOpen since TSSocketManager observes the registrationDidChange notification.

michaelkirk:
Closing this for now.

PR #2624, opened 2017-10-10
Use separate Giphy API key for Signal iOS.

charlesmchen:
PTAL @michaelkirk


michaelkirk:
Should probably get rid of that TODO now.

charlesmchen:
> Should probably get rid of that TODO now.

Done!

PR #2625, opened 2017-10-10
Suppress build warnings for Cocoapods dependencies.

charlesmchen:
This PR is intended as a proposal.  

* Working through build errors is so much easier without the huge quantity of warnings from our pods.
* We're less likely to miss/ignore a warning from our own codebase amid the noise.

PTAL @michaelkirk

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -2,16 +2,19 @@ platform :ios, '8.0'
 source 'https://github.com/CocoaPods/Specs.git'
 
 target 'Signal' do
-    pod 'ATAppUpdater'
+	pod '25519', :inhibit_warnings => true
+    pod 'Mantle', :inhibit_warnings => true	
+	pod 'YapDatabase/SQLCipher', '~> 2.9.3', :inhibit_warnings => true	
+    pod 'ATAppUpdater', :inhibit_warnings => true
     pod 'AxolotlKit',                 git: 'https://github.com/WhisperSystems/SignalProtocolKit.git'
     #pod 'AxolotlKit',                 path: '../SignalProtocolKit'
-    pod 'JSQMessagesViewController',  git: 'https://github.com/WhisperSystems/JSQMessagesViewController.git', branch: 'signal-master'
+    pod 'JSQMessagesViewController',  git: 'https://github.com/WhisperSystems/JSQMessagesViewController.git', branch: 'signal-master', :inhibit_warnings => true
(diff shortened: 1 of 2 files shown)


Podfile
```diff
@@ -2,16 +2,19 @@ platform :ios, '8.0'
 source 'https://github.com/CocoaPods/Specs.git'
 
 target 'Signal' do
-    pod 'ATAppUpdater'
+	pod '25519', :inhibit_warnings => true
```
michaelkirk: nits:
- Weird indent. Can you make sure your editor is not configured to insert tabs?
- elsewhere we're using the `key: 'value'` hash syntax, not the `'key' => 'value'` hash syntax.

not just nits:

- What version of cocoapods are you using?  I'm using the bundler specified version and unable to `pod install`

with error:
```
NoMethodError - undefined method `match' for 25519:Fixnum
/Users/mkirk/.rbenv/versions/2.0.0-p648/lib/ruby/gems/2.0.0/gems/cocoapods-core-1.3.1/lib/cocoapods-core/dependency.rb:348:in `from_string'
```
probably due to this not being merged: https://github.com/WhisperSystems/Curve25519Kit/pull/1/commits/c87d1eb0ad338389add67ae1f7e0b85582986187

If you're still OK going forward with the framework friendly stuff, I can rebase the relevant PRs. The previous motivation for it is no longer immediately pressing, but annoyances like this keep popping up.

Also, we probably shouldn't be inhibiting warnings for this library anyway, which is code that we're maintaining.

charlesmchen: > Weird indent. 

Fixed.

> elsewhere we're using the key: 'value' hash syntax, not the 'key' => 'value' hash syntax.

That doesn't work for me.  Is that syntax only supported in podspecs?  I'm following the examples in the cocoapods docs, IIRC

charlesmchen: > What version of cocoapods are you using? I'm using the bundler specified version and unable to pod install

Fixed, I removed `25519` altogether from this podfile since as you point out we maintain the code and don't want to suppress its warnings.

charlesmchen:
> If you're still OK going forward with the framework friendly stuff, I can rebase the relevant PRs. The previous motivation for it is no longer immediately pressing, but annoyances like this keep popping up.

Sure, let's do that - but can we also merge this for now?  It saves a lot of time and frustration when working through build errors.

charlesmchen:
PTAL @michaelkirk

michaelkirk:
Yup, LGTM

PR #2629, opened 2017-10-11
Avoid deadlock in profile manager.

charlesmchen:
Warning: Largely a whitespace change.

PTAL @michaelkirk


michaelkirk:
In 44051bd7e7dfae7c41242c8de7d982fa01adf134

PR #2631, opened 2017-10-11
push notification fixups

michaelkirk:
Solves an issue with some iOS11 users not being able to register, but quite a bit of code moved to make the push registration process more coherent.

Some inline comments forthcoming...

PTAL @charlesmchen


Signal/src/AppDelegate.m
```diff
@@ -124,7 +124,6 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     [self setupEnvironment];
 
     [UIUtil applySignalAppearence];
-    [[PushManager sharedManager] registerPushKitNotificationFuture];
```
michaelkirk: We no longer require the AppDelegate to set up this state.

Signal/src/Models/AccountManager.swift
```diff
@@ -33,24 +34,33 @@ class AccountManager: NSObject {
...
-                fulfill()
-            }
-        }.then {
-            Logger.debug("\(self.TAG) verification code looks well formed.")
-            return self.registerForTextSecure(verificationCode: verificationCode)
+        guard verificationCode.characters.count > 0 else {
+            let error = OWSErrorWithCodeDescription(.userError,
+                                                    NSLocalizedString("REGISTRATION_ERROR_BLANK_VERIFICATION_CODE",
+                                                                      comment: "alert body during registration"))
+            return Promise(error: error)
```
michaelkirk: This is just a more concise way to return an already failed promise.

Signal/src/Models/AccountManager.swift
```diff
@@ -33,24 +34,33 @@ class AccountManager: NSObject {
...
-            return SyncPushTokensJob.run(pushManager: self.pushManager, accountManager: self, preferences: self.preferences)
+            self.syncPushTokens()
+        }.recover { (error) -> Promise<Void> in
+            switch error {
+            case PushRegistrationError.pushNotSupported(let description):
+                // This can happen with:
+                // - simulators, none of which support receiving push notifications
+                // - on iOS11 devices which have disabled "Allow Notifications" and disabled "Enable Background Refresh" in the system settings.
+                Logger.info("\(self.TAG) Recovered push registration error. Registering for manual message fetcher because push not supported: \(description)")
+                return self.registerForManualMessageFetching()
```
michaelkirk: To receive messages, a device must either:

1. register push tokens (this is the only way we used to work)
2. specify that they'll fetch messages manually (this is how Desktop always works)

Note:  for simulators we no longer register fake push tokens.

Signal/src/Models/AccountManager.swift
```diff
@@ -33,24 +34,33 @@ class AccountManager: NSObject {
...
+            default:
+                throw error
+            }
         }.then {
-            Logger.debug("\(self.TAG) successfully registered for TextSecure")
+            self.completeRegistration()
```
michaelkirk: Now registration is not considered complete until we've registered to receive messages (again, either by uploading push tokens or specifying manual fetching).

note: registering for manual fetching doesn't preclude us from receiving push notifications if we change our settings later.

Signal/src/ViewControllers/HomeViewController.m
```diff
@@ -490,13 +488,10 @@ - (void)viewDidAppear:(BOOL)animated
 {
     [super viewDidAppear:animated];
 
-    if (self.newlyRegisteredUser) {
-        [self markAllUpgradeExperiencesAsSeen];
```
michaelkirk: This was redundant with what happens in the appDelegate regsitrationStateDidChange notification observer.

Signal/src/contact/OWSContactsManager.m
```diff
@@ -27,8 +27,6 @@
 
 @interface OWSContactsManager () <SystemContactsFetcherDelegate>
 
-@property (atomic) id addressBookReference;
-@property (atomic) TOCFuture *futureAddressBook;
```
michaelkirk: These haven't been used for a while.

Signal/src/environment/PushRegistrationManager.swift
```diff
@@ -0,0 +1,287 @@
...
+public enum PushRegistrationError: Error {
+    case assertionError(description: String)
+    case pushNotSupported(description: String)
+    case timeout
+}
+
+/**
+ * Singleton used to integrate with push notification services - registration and routing received remote notifications.
+ */
+@objc class PushRegistrationManager: NSObject, PKPushRegistryDelegate {
```
michaelkirk: I cleaved the push registration logic from the the PushManager. My hope is too eventually merge the remaining PushManager logic into the NotificationManager.

SignalServiceKit/src/Account/TSAccountManager.m
```diff
@@ -290,11 +304,6 @@ - (void)verifyAccountWithCode:(NSString *)verificationCode
                                                                                forNumber:phoneNumber
                                                                             signalingKey:signalingKey
                                                                                  authKey:authToken];
-    void (^completedRegistrationBlock)() = ^{
-        [self didRegister];
-        [TSSocketManager requestSocketOpen];
```
michaelkirk: didRegister has been moved until after we're sure the user is able to receive messages.
requestSocketOpen was redundant with the AppDelegate's registrationState observer.

Signal/src/Models/AccountManager.swift
charlesmchen: If user later gets a push token, do we need to "unregister for manual message fetching"?

Signal/src/Models/SyncPushTokensJob.swift
```diff
@@ -2,68 +2,72 @@
...
-            Logger.warn("\(self.TAG) Sending new tokens to account servers. pushToken: \(pushToken), voipToken: \(voipToken)")
-
+            Logger.warn("\(self.TAG) uploading tokens to account servers. pushToken: \(pushToken), voipToken: \(voipToken)")
             return self.accountManager.updatePushTokens(pushToken:pushToken, voipToken:voipToken).then {
-                return self.recordNewPushTokens(pushToken:pushToken, voipToken:voipToken)
-            }.then {
-                Logger.debug("\(self.TAG) Successfully ran syncPushTokensJob.")
-            }.catch { error in
-                Logger.error("\(self.TAG) Failed to run syncPushTokensJob with error: \(error).")
+                Logger.info("\(self.TAG) successfully updated push tokens on server")
```
charlesmchen: This is a promise within a promise?  Couldn't we flatten this?

Signal/src/environment/PushRegistrationManager.swift
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    private func registerForVoipPushToken() -> Promise<String> {
+        AssertIsOnMainThread()
+        Logger.info("\(self.TAG) in \(#function)")
+
+        guard self.voipTokenPromise == nil else {
+            let promise = self.voipTokenPromise!
+            assert(promise.isPending)
+            return promise.then { $0.hexEncodedString }
```
charlesmchen: I'm not a big fan of `$0`.  More readable to name the variable.

SignalServiceKit/src/Account/TSAccountManager.h
```diff
@@ -65,6 +65,15 @@ extern NSString *const kNSNotificationName_LocalNumberDidChange;
                       success:(void (^)())successBlock
                       failure:(void (^)(NSError *error))failureBlock;
 
+- (void)registerForManualMessageFetchingWithSuccess:(void (^)())successBlock
+                                            failure:(void (^)(NSError *error))failureBlock;
+
+// Called once registration is complete - meaning the following have succeeded:
+// - obtained signal server credentials
+// - uploaded pre-keys
+// - uploaded push tokens
```
charlesmchen: Don't we do this sometimes even if upload of push tokens has failed?

Signal/src/Models/AccountManager.swift
michaelkirk: I asked moxie the same thing and he said "no need to do that."

Signal/src/Models/SyncPushTokensJob.swift
michaelkirk: `updatePushTokens` and `recordPushTokensLocally` need to happen async and sequentially, and they both need to have `(pushToken, voipToken)` in scope.

I believe this is the most legible way to achieve that.

Signal/src/environment/PushRegistrationManager.swift
michaelkirk: Agreed that `$0` *can* be damaging for readability, but for *really* short blocks, (like 1 or 2 lines), I think it can actually help readability when used judiciously.

Similarly, depending on the context, sometimes you want something like:

```
let myFoo = Foo()
let myBar = myFoo.bar
let myQux = myBar.qux
return myQux
```

And sometimes you want something more like:

```
return Foo().bar.qux
```

SignalServiceKit/src/Account/TSAccountManager.h
michaelkirk: Well - we no longer do. 

Previously we'd show an error, but then if you restarted Signal you'd be in a state where you could send messages, but no one could message you.

PR #2632, opened 2017-10-12
Suspend while answering shows "Missed Call"

michaelkirk:
When awakened by a voip push, we get some arbitrary amount of background
time to connect the call (in practice this is ~30s) before the app is
suspended.

Though we were properly terminating the call upon being suspended, we
were not notifying the user that they had missed a call.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -79,7 +79,7 @@ enum CallError: Error {
 }
 
 // Should be roughly synced with Android client for consistency
-private let connectingTimeoutSeconds = 120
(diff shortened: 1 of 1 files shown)


Signal/src/call/CallService.swift
```diff
@@ -1403,6 +1403,14 @@ protocol CallServiceObserver: class {
         }
 
         if let failedCall = failedCall {
+
+            if failedCall.state == .answering {
```
michaelkirk: This is the significant change - everything else is style cleanup.

PR #2637, opened 2017-10-13
Charlesmchen/rework conversation input toolbar

charlesmchen:

PTAL @michaelkirk 

### Hot areas at minimum size.

Note that the inner border reflects the button and the outer border reflects the wrapper, which wis used to expand the hot area of the button.

Note that these borders are here just to make the screenshots legible.

screen shot 2017-10-13 at 10 18 16 am

### Hot areas at expanded size.

screen shot 2017-10-13 at 10 18 09 am

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
@@ -79,13 +79,11 @@ static void *kConversationInputTextViewObservingContext = &kConversationInputTex
     // using iOS auto layout.
     _leftButtonWrapper = [UIView containerView];
     [self.leftButtonWrapper
-        addGestureRecognizer:[[UITapGestureRecognizer alloc] initWithTarget:self
-                                                                     action:@selector(leadingButtonTapped:)]];
+        addGestureRecognizer:[[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(leftButtonTapped:)]];
     [self addSubview:self.leftButtonWrapper];
     _rightButtonWrapper = [UIView containerView];
     [self.rightButtonWrapper
-        addGestureRecognizer:[[UITapGestureRecognizer alloc] initWithTarget:self
-                                                                     action:@selector(trailingButtonTapped:)]];
+        addGestureRecognizer:[[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(rightButtonTapped:)]];
     [self addSubview:self.rightButtonWrapper];
 
     _attachmentButton = [[UIButton alloc] init];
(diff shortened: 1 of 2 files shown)


charlesmchen:

PTAL @michaelkirk

Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
```diff
@@ -73,15 +96,8 @@ - (void)createContents
...
-
-    //    [_attachButton setFrame:CGRectMake(0,
-    //                                       0,
-    //                                       JSQ_TOOLBAR_ICON_WIDTH + JSQ_IMAGE_INSET * 2,
-    //                                       JSQ_TOOLBAR_ICON_HEIGHT + JSQ_IMAGE_INSET * 2)];
-    //    _attachButton.imageEdgeInsets
-    //    = UIEdgeInsetsMake(JSQ_IMAGE_INSET, JSQ_IMAGE_INSET, JSQ_IMAGE_INSET, JSQ_IMAGE_INSET);
     [self.attachmentButton setImage:[UIImage imageNamed:@"btnAttachments--blue"] forState:UIControlStateNormal];
-    [self addSubview:self.attachmentButton];
+    [self.leftButtonWrapper addSubview:self.attachmentButton];
```
charlesmchen: My feeling is that "attachment" should always be on the left and "send/voice memo" should always be on the right. Here's my argument.

* Regardless of RTL/LTR, most people are right-handed.
* We need to support one-handed use (e.g. while holding a subway pole).
* In one-handed mode, you use your thumb.
* "Send" is the key action; it needs to be highly accessible by the thumb.

charlesmchen: WDYT?

erikstrottmann: iMessage keeps “Attachment” on the left and “Voice Memo” on the right for both LTR and RTL languages. In fact, even the “Send” arrow appears on the right for both language directions.

messages-rtl

charlesmchen: Thanks @erikstrottmann

Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
```diff
@@ -1150,6 +609,40 @@ - (BOOL)textView:(UITextView *)textView shouldChangeTextInRange:(NSRange)range r
...
+                            context:kConversationInputTextViewObservingContext];
+}
+
+- (void)removeKVOObservers
+{
+    @try {
+        [self.inputTextView removeObserver:self
+                                forKeyPath:NSStringFromSelector(@selector(contentSize))
+                                   context:kConversationInputTextViewObservingContext];
+    } @catch (NSException *__unused exception) {
```
michaelkirk: What's the likely exception here?

charlesmchen: KVO throws an exception if remove an observer which isn't currently observing.  So either:

* This view is destroyed before the observer is registered.
* Somehow the observer is de-registered twice.

I'm 99% sure this shouldn't happen now - I've greatly simplified the usage of KVO vs. JSQ - but I decided to hold this over at least for a while until we're sure.  I've added a `owsFail()` and a TODO to remove this.

PR #2638, opened 2017-10-13
Add workaround for bug in iOS 11.1 beta around hit area of custom bac…

charlesmchen:
…k buttons.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -1328,27 +1328,29 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
     (OWSDisappearingMessagesConfiguration *)disappearingMessagesConfiguration
 {
     UIBarButtonItem *backItem = [self createOWSBackButton];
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1328,27 +1328,29 @@ - (void)setBarButtonItemsForDisappearingMessagesConfiguration:
...
-                                          toSize:self.unreadCountViewDiameter
-                                        relation:NSLayoutRelationGreaterThanOrEqual];
-
-    [_backButtonUnreadCountView addSubview:_backButtonUnreadCountLabel];
-    [_backButtonUnreadCountLabel autoPinWidthToSuperviewWithMargin:4];
-    [_backButtonUnreadCountLabel autoPinHeightToSuperview];
-
-    // Initialize newly created unread count badge to accurately reflect the current unread count.
-    [self updateBackButtonUnreadCount];
+    if (backItem.customView) {
```
charlesmchen: Largely whitespace change.  Can't safely add constraints without a valid superview.

PR #2639, opened 2017-10-15
Message details shows entire message

michaelkirk:
It feels broken that we truncate long messages.

When viewing a longer message, to avoid the risk of the user not being
aware they can scroll down, we scroll to the last couple lines of the
message, ensuring they can see the meta-data.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MessageMetadataViewController.swift
+++ b/Signal/src/ViewControllers/MessageMetadataViewController.swift
@@ -25,6 +25,7 @@ class MessageMetadataViewController: OWSViewController {
     let databaseConnection: YapDatabaseConnection
 
     let bubbleFactory = OWSMessagesBubbleImageFactory()
+    var bubbleView: UIView?
 
     var message: TSMessage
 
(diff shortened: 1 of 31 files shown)


Signal/src/ViewControllers/MessageMetadataViewController.swift
```diff
@@ -283,8 +293,7 @@ class MessageMetadataViewController: OWSViewController {
                 bodyLabel.textColor = isIncoming ? UIColor.black : UIColor.white
                 bodyLabel.font = UIFont.ows_regularFont(withSize:16)
                 bodyLabel.text = messageBody
-                // Only show the first N lines.
-                bodyLabel.numberOfLines = 10
+                bodyLabel.numberOfLines = 0
```
charlesmchen: Great. Is there a vertical scroll indicator?

Also, let's make sure the test plan includes a "send/receive moby dick" case.

michaelkirk: > Great. Is there a vertical scroll indicator?

Yup - by default

> Also, let's make sure the test plan includes a "send/receive moby dick" case.
 Write

:+1:

PR #2640, opened 2017-10-15
Pull to refresh on homeview fetches messages.

michaelkirk:
This is useful when you're using censorship circumvention and unable to
receive push notifications.

PTAL @charlesmchen


Signal/src/ViewControllers/HomeViewController.m
```diff
@@ -596,6 +603,16 @@ - (CGFloat)tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPa
     return InboxTableViewCell.rowHeight;
 }
 
+- (void)pullToRefreshPerformed:(UIRefreshControl *)refreshControl
+{
+    OWSAssert([NSThread isMainThread]);
+    DDLogInfo(@"%@ begging refreshing.", self.tag);
```
charlesmchen: begging

michaelkirk: Thanks - fixed.

PR #2642, opened 2017-10-15
compose screen uses table index and search includes groups

michaelkirk:
If you are not searching, we now show a table index so you can quickly navigate to all your (e.g.) contacts whose name starts with an "R".

image

If you are searching, we do not show the table index, but we do separate results into sections, and include groups (only when searching).

image

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/NewContactThreadViewController.m
+++ b/Signal/src/ViewControllers/NewContactThreadViewController.m
@@ -22,6 +22,22 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@interface SignalAccount (Collation)
+
+- (NSString *)stringForCollation;
+
+@end
+
+@implementation SignalAccount (Collation)
+
+- (NSString *)stringForCollation
+{
+    OWSContactsManager *contactsManager = [Environment getCurrent].contactsManager;
+    return [contactsManager comparableNameForSignalAccount:self];
+}
+
+@end
+
 @interface NewContactThreadViewController () <UISearchBarDelegate,
     ContactsViewHelperDelegate,
     OWSTableViewControllerDelegate,
@@ -33,6 +49,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) UIView *noSignalContactsView;
 
 @property (nonatomic, readonly) OWSTableViewController *tableViewController;
+
+@property (nonatomic, readonly) UILocalizedIndexedCollation *collation;
 
 @property (nonatomic, readonly) UISearchBar *searchBar;
 @property (nonatomic, readonly) NSLayoutConstraint *hideContactsPermissionReminderViewConstraint;
@@ -59,6 +77,7 @@ NS_ASSUME_NONNULL_BEGIN
 
     _contactsViewHelper = [[ContactsViewHelper alloc] initWithDelegate:self];
     _nonContactAccountSet = [NSMutableSet set];
+    _collation = [UILocalizedIndexedCollation currentCollation];
 
     ReminderView *contactsPermissionReminderView = [[ReminderView alloc]
         initWithText:NSLocalizedString(@"COMPOSE_SCREEN_MISSING_CONTACTS_PERMISSION",
@@ -285,6 +304,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - Table Contents
 
+- (CGFloat)actionCellHeight
+{
+    return ScaleFromIPhone5To7Plus(round((kOWSTable_DefaultCellHeight + [ContactTableViewCell rowHeight]) * 0.5f),
+        [ContactTableViewCell rowHeight]);
+}
+
 - (void)updateTableContents
 {
     OWSTableContents *contents = [OWSTableContents new];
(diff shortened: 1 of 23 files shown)


Signal/src/ViewControllers/NewContactThreadViewController.m
```diff
@@ -295,46 +320,166 @@ - (void)updateTableContents
...
+    if (hasSearchText) {
+        for (OWSTableSection *section in [self contactsSectionsForSearch]) {
+            [contents addSection:section];
+        }
+    } else {
+        for (OWSTableSection *section in [self collatedContactsSections]) {
+            [contents addSection:section];
+        }
+        contents.sectionForSectionIndexTitleBlock = ^NSInteger(NSString *_Nonnull title, NSInteger index) {
+            // We have at least one non-collated section - the "Find By Phone Number / Inbite Friends" section
```
charlesmchen: s/Inbite/Invite/

Signal/src/ViewControllers/NewContactThreadViewController.m
```diff
@@ -295,46 +320,166 @@ - (void)updateTableContents
...
+    [contents addSection:staticSection];
+
+    BOOL hasSearchText = [self.searchBar text].length > 0;
+
+    if (hasSearchText) {
+        for (OWSTableSection *section in [self contactsSectionsForSearch]) {
+            [contents addSection:section];
+        }
+    } else {
+        for (OWSTableSection *section in [self collatedContactsSections]) {
```
charlesmchen: If you insert...

```
const NSUInteger nonCollatedSectionCount = contents.sections.count;
```
...here before adding the collated sections, can we simplify the logic below?

Signal/src/ViewControllers/NewContactThreadViewController.m
```diff
@@ -295,46 +320,166 @@ - (void)updateTableContents
...
+    
+    NSMutableArray<OWSTableSection *> *contactSections = [NSMutableArray new];
+    
+    NSMutableArray<NSMutableArray<SignalAccount *> *> *collatedSignalAccounts = [NSMutableArray new];
+    for (NSUInteger i = 0; i < self.collation.sectionTitles.count; i++) {
+        collatedSignalAccounts[i] = [NSMutableArray new];
+    }
+    for (SignalAccount *signalAccount in self.contactsViewHelper.signalAccounts) {
+        NSInteger section =
+            [self.collation sectionForObject:signalAccount collationStringSelector:@selector(stringForCollation)];
```
charlesmchen: Let's make sure the test plan includes contacts...

* without a first or last name.
* with a display name that starts with whitespace.
* with a display name that starts with a chinese high-code-plane character.
* with a display name that starts with punctuation.
* with a display name that starts with emoji.

Signal/src/ViewControllers/NewContactThreadViewController.m
```diff
@@ -388,76 +541,107 @@ - (void)updateTableContents
...
-            [section addItem:[OWSTableItem disclosureItemWithText:text
-                                                  customRowHeight:kActionCellHeight
-                                                      actionBlock:^{
-                                                          [weakSelf sendTextToPhoneNumber:phoneNumber.toE164];
-                                                      }]];
+    // When searching, we include matching groups
+    OWSTableSection *groupSection = [OWSTableSection new];
+    groupSection.headerTitle = NSLocalizedString(
+        @"COMPOSE_MESSAGE_GROUP_SECTION_TITLE", @"Table section header for group listing when composing a new message");
+    NSArray<TSGroupThread *> *filteredGroupThreads = [self filteredGroupThreads];
```
charlesmchen: This is awesome.

Signal/src/ViewControllers/NewContactThreadViewController.m
```diff
@@ -388,76 +541,107 @@ - (void)updateTableContents
...
+{
+    NSMutableArray<TSGroupThread *> *matchingThreads = [NSMutableArray new];
+    NSString *searchString = [[self.searchBar.text lowercaseString] ows_stripped];
+    [TSGroupThread enumerateCollectionObjectsUsingBlock:^(id obj, BOOL *stop) {
+        if (![obj isKindOfClass:[TSGroupThread class]]) {
+            // group and contact threads are in the same collection.
+            return;
+        }
+        TSGroupThread *groupThread = (TSGroupThread *)obj;
+        if ([[[groupThread.groupModel.groupName lowercaseString] ows_stripped] containsString:searchString]) {
```
charlesmchen: I think it'd be nice if our matching algorithm supported "splits", e.g. searching for "Chen M" matched "Matthew Chen".  See the search logic in `ContactsViewHelper`.  In fact, I'd love to move this logic into `ContactsViewHelper`, because eventually I'm hoping to do search in the home view that matches (group name/contact name/contact phone number/non-phone identifier/message content!).

Signal/src/ViewControllers/OWSTableViewController.m
```diff
@@ -509,19 +509,36 @@ - (nullable UIView *)tableView:(UITableView *)tableView viewForFooterInSection:(
 - (CGFloat)tableView:(UITableView *)tableView heightForHeaderInSection:(NSInteger)sectionIndex
 {
     OWSTableSection *section = [self sectionForIndex:sectionIndex];
```
charlesmchen: _Nullable?

Signal/src/ViewControllers/OWSTableViewController.m
```diff
@@ -509,19 +509,36 @@ - (nullable UIView *)tableView:(UITableView *)tableView viewForFooterInSection:(
...
-    if (section && section.customHeaderHeight) {
+
+    if (!section) {
+        OWSFail(@"Section index out of bounds.");
+        return 0;
+    }
+
+    if (section.customHeaderHeight) {
         return [section.customHeaderHeight floatValue];
```
charlesmchen:
```
OWSAssert([section.customHeaderHeight floatValue] > 0);
```
?

Signal/src/ViewControllers/OWSTableViewController.m
```diff
@@ -509,19 +509,36 @@ - (nullable UIView *)tableView:(UITableView *)tableView viewForFooterInSection:(
...
+        return UITableViewAutomaticDimension;
+    } else {
+        return 0;
     }
-    return UITableViewAutomaticDimension;
 }
 
 - (CGFloat)tableView:(UITableView *)tableView heightForFooterInSection:(NSInteger)sectionIndex
 {
     OWSTableSection *section = [self sectionForIndex:sectionIndex];
```
charlesmchen: _Nullable

Signal/src/ViewControllers/NewContactThreadViewController.m
```diff
@@ -295,46 +320,166 @@ - (void)updateTableContents
...
+        }
+        contents.sectionForSectionIndexTitleBlock = ^NSInteger(NSString *_Nonnull title, NSInteger index) {
+            // We have at least one non-collated section - the "Find By Phone Number / Inbite Friends" section
+            // If we don't account for these, our collation indices will be offset.
+
+            // Count the sections which are not part of our collation.
+            NSInteger extraSections
+                = (NSInteger)contents.sections.count - (NSInteger)self.collation.sectionTitles.count;
+
+            NSInteger sectionIndex = [self.collation sectionForSectionIndexTitleAtIndex:index] + extraSections;
```
charlesmchen: I guess I also wonder if we add a `NSNumber *collatedSectionIndex` property to `OWSTableSection`, would that simplify things?

Signal/src/views/GroupTableViewCell.swift
```diff
@@ -0,0 +1,65 @@
...
+        // Listen to notifications...
+        // TODO avatar, group name change, group membership change, group member name change
+
+        // Layout
+
+        nameLabel.autoPinEdgesToSuperviewEdges(with: UIEdgeInsets.zero, excludingEdge: .bottom)
+        subtitleLabel.autoPinEdgesToSuperviewEdges(with: UIEdgeInsets.zero, excludingEdge: .top)
+        subtitleLabel.autoPinEdge(.top, to: .bottom, of: nameLabel)
+
+        avatarView.autoPinLeadingToSuperview()
```
charlesmchen: Don't we need the swift equivalent of:
```
    self.preservesSuperviewLayoutMargins = YES;
    self.contentView.preservesSuperviewLayoutMargins = YES;
```
in order for the avatar to align properly?

Signal/src/views/GroupTableViewCell.swift
```diff
@@ -0,0 +1,65 @@
...
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    public func configure(thread: TSGroupThread, contactsManager: OWSContactsManager) {
+        self.nameLabel.text = thread.groupModel.groupName
+
+        let groupMemberIds: [String] = thread.groupModel.groupMemberIds
+        let groupMemberNames = groupMemberIds.map { (recipientId: String) in
```
charlesmchen: Should we sort the group member names?

Signal/src/ViewControllers/NewContactThreadViewController.m
michaelkirk: Good idea - Done.

Signal/src/ViewControllers/NewContactThreadViewController.m
michaelkirk: :+1:

Signal/src/ViewControllers/NewContactThreadViewController.m
michaelkirk: fixed- thanks!

Signal/src/ViewControllers/NewContactThreadViewController.m
michaelkirk: I've extracted the group search logic, and it now supports splits. Test probably illustrates this best: https://github.com/WhisperSystems/Signal-iOS/blob/master/Signal/test/util/SearcherTest.swift

I also decided to match against group member names as well as the group name. So, for example, searching for "Alice Bob" will list groups to which both Alice and Bob are members.

Signal/src/ViewControllers/OWSTableViewController.m
michaelkirk: done!

Signal/src/ViewControllers/OWSTableViewController.m
michaelkirk: done, and for the footerHeight.

Signal/src/ViewControllers/OWSTableViewController.m
michaelkirk: done!

Signal/src/views/GroupTableViewCell.swift
michaelkirk: I believe this would be true if we were using the stock purelayout: `autoPinEdge(toSuperviewEdge: .leading)`.

However, this is one of our own custom purelayout methods, which as you can see in the screenshot, respects the superview margins. Maybe we should rename these to something like: `avatarView.autoPinLeadingToSuperviewMargin()`

Signal/src/views/GroupTableViewCell.swift
michaelkirk: Probably... We should also probably use the same sorting in the Group's conversation settings "Group Members" list. I'e added a card.

Signal/src/ViewControllers/NewContactThreadViewController.m
michaelkirk: I'm not totally sure what you mean - would that entail adding a map of `(collatedSectionIndex) -> (OWSTableSection)` in OWSTableView?

I'm going to leave it as is for now, but happy to revisit.

PR #2649, opened 2017-10-17
Add actions to debug UI to inject fake incoming messages.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
+++ b/Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
@@ -11,6 +11,7 @@
 #import <25519/Randomness.h>
 #import <AFNetworking/AFNetworking.h>
 #import <AxolotlKit/PreKeyBundle.h>
+#import <SignalServiceKit/OWSBatchMessageProcessor.h>
 #import <SignalServiceKit/OWSDisappearingConfigurationUpdateInfoMessage.h>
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -1028,6 +1041,54 @@ + (void)createNewGroups:(int)counter recipientId:(NSString *)recipientId
...
+        [dataMessageBuilder setGroup:groupBuilder.build];
+    }
+
+    OWSSignalServiceProtosContentBuilder *payloadBuilder = [OWSSignalServiceProtosContentBuilder new];
+    [payloadBuilder setDataMessage:dataMessageBuilder.build];
+    NSData *plaintextData = [payloadBuilder build].data;
+
+    OWSSignalServiceProtosEnvelopeBuilder *envelopeBuilder = [OWSSignalServiceProtosEnvelopeBuilder new];
+    [envelopeBuilder setType:OWSSignalServiceProtosEnvelopeTypeCiphertext];
+    [envelopeBuilder setSource:@"+12345678901"];
```
michaelkirk: Shouldn't this number correspond to the thread rather than hardcoded?

charlesmchen: Okay I reworked this a bit. Thanks.

PR #2655, opened 2017-10-17
Charlesmchen/input placeholder

charlesmchen:
This PR is ready for review. 

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
@@ -8,6 +8,16 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@interface ConversationInputTextView () <UITextViewDelegate>
+
+@property (nonatomic) UILabel *placeholderView;
+@property (nonatomic) NSArray<NSLayoutConstraint *> *placeholderConstraints;
+@property (nonatomic) BOOL isEditing;
+
+@end
+
+#pragma mark -
+
 @implementation ConversationInputTextView
 
 - (instancetype)init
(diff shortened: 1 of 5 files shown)


charlesmchen:
This PR is ready for review. 

PTAL @michaelkirk

Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
```diff
@@ -44,13 +52,89 @@ - (instancetype)init
 
         self.text = nil;
 
-        //        _placeHolder = nil;
-        //        _placeHolderTextColor = [UIColor lightGrayColor];
+        self.placeholderView = [UILabel new];
```
charlesmchen: There's two ways to add a placeholder to `UITextArea`, using a subview (as I do here) or by using `drawRect:`.  Overriding `drawRect:` has serious perf implementations so I've gone this way.

Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
```diff
@@ -44,13 +52,89 @@ - (instancetype)init
...
+        self.placeholderView.text = NSLocalizedString(@"new_message", @"");
+        self.placeholderView.textColor = [UIColor lightGrayColor];
+        self.placeholderView.textAlignment = NSTextAlignmentLeft;
```
michaelkirk: Shouldn't the alignment be default to be RTL friendly?

Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
michaelkirk: 👍

Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
```diff
@@ -44,13 +52,89 @@ - (instancetype)init
...
+    // beginning position.
+    UITextRange *beginningTextRange =
+        [self textRangeFromPosition:self.beginningOfDocument toPosition:self.beginningOfDocument];
+    CGRect beginningTextRect = [self firstRectForRange:beginningTextRange];
+
+    CGFloat hInset = beginningTextRect.origin.x;
+    CGFloat topInset = beginningTextRect.origin.y;
+
+    self.placeholderConstraints = @[
+        [self.placeholderView autoPinEdgeToSuperviewEdge:ALEdgeLeft withInset:hInset],
```
michaelkirk: Do we want to pin to leading for RTL?

Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
charlesmchen: You're right!

Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
charlesmchen: I don't think that's necessary since placeholderView's text alignment will take care of that for us.

PR #2656, opened 2017-10-17
Charlesmchen/menu controller

charlesmchen:

PTAL @michaelkirk


charlesmchen:

PTAL @michaelkirk

Signal/src/ViewControllers/ConversationView/Cells/OWSSystemMessageCell.m
```diff
@@ -305,11 +305,31 @@ - (void)prepareForReuse
...
+
+    DDLogDebug(@"%@ long pressed system message cell: %@", self.logTag, self.viewItem.interaction.debugDescription);
+
+    [self becomeFirstResponder];
+
+    if ([UIMenuController sharedMenuController].isMenuVisible) {
+        [[UIMenuController sharedMenuController] setMenuVisible:NO animated:NO];
+    }
+
+    UIMenuController *menuController = [UIMenuController sharedMenuController];
```
charlesmchen: `UIMenuController` is horrible. It's super stateful.

Going forward, we should be careful to _always_ update `menuItems` before we show `UIMenuController` "manually" (e.g. in response to a long press gesture)  - even if we're just niling them out.  Bear in mind that `UIMenuController` will also be shown by UIKit components (e.g. long pressing on `UITextField` and `UITextArea`).  

We'll leave our custom `menuItems` on `UIMenuController` when we're done with it, but that should be okay so long as our custom menu items always either:

* Use a default menu item action (e.g. copy, paste, delete, share, etc.) or
* Use a custom action which is guaranteed not to correspond with a method on other classes (e.g. `UITextField` and `UITextArea`).

michaelkirk: Seems like a good convention for dealing with menu madness.

PR #2659, opened 2017-10-18
Move write of incoming messages off main thread.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Network/WebSockets/TSSocketManager.m
+++ b/SignalServiceKit/src/Network/WebSockets/TSSocketManager.m
@@ -381,20 +381,25 @@ NSString *const kNSNotification_SocketManagerStateDidChange = @"kNSNotification_
 
     if ([message.path isEqualToString:@"/api/v1/message"] && [message.verb isEqualToString:@"PUT"]) {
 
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Network/WebSockets/TSSocketManager.m
```diff
@@ -381,20 +381,25 @@ - (void)processWebSocketRequestMessage:(WebSocketRequestMessage *)message {
 
     if ([message.path isEqualToString:@"/api/v1/message"] && [message.verb isEqualToString:@"PUT"]) {
 
-        NSData *decryptedPayload =
-            [Cryptography decryptAppleMessagePayload:message.body withSignalingKey:TSStorageManager.signalingKey];
+        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
charlesmchen: Mostly ws changes.

charlesmchen:
> LGTM, but should this target release/2.18.0

IMO no.  I think we need to live with this for a while.

PR #2661, opened 2017-10-18
Charlesmchen/refine gif search

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
+++ b/Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
@@ -56,7 +56,7 @@ class GifPickerViewController: OWSViewController, UISearchBarDelegate, UICollect
 
         self.searchBar = UISearchBar()
         self.layout = GifPickerLayout()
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
```diff
@@ -365,15 +367,15 @@ class GifPickerViewController: OWSViewController, UISearchBarDelegate, UICollect
                 return
             }
 
-            strongSelf.tryToSearch()
+            strongSelf.tryToSearch(dismissKeyboard: false)
```
michaelkirk: I think we only want to dismiss when the user explicitly taps the "search" button right? 

Currently we'll also dismiss upon retry right?

It seems like the dismiss logic should live with the button press in `searchBarSearchButtonClicked`

```
    public func searchBarSearchButtonClicked(_ searchBar: UISearchBar) {
        tryToSearch()
        self.searchBar.resignFirstResponder()
    }
```

No biggie though.

charlesmchen: Yup, that's better. Done.

PR #2666, opened 2017-10-19
Charlesmchen/auto load more

charlesmchen:
PTAL @michaelkirk


charlesmchen:
As you may have noticed, I've added a DEBUG-only "scroll up" button in conversation view.  It's helpful for testing certain scenarios.

PR #2669, opened 2017-10-19
Clean up conversation view.

charlesmchen:
This PR finalizes a bunch of changes from the JSQ rewrite.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationInputTextView.h
+++ b/Signal/src/ViewControllers/ConversationView/ConversationInputTextView.h
@@ -10,8 +10,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)didPasteAttachment:(SignalAttachment *_Nullable)attachment;
 
-- (void)textViewDidChangeLayout;
-
 - (void)inputTextViewDidBecomeFirstResponder;
 
 @end
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -145,11 +145,14 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
 // * The second (required) step is to update messageMappings.
 // * The third (optional) step is to update the messageMappings range using
 //   updateMessageMappingRangeOptions.
+// * The fourth (optional) step is to update the view items using reloadViewItems.
 // * The steps must be done in strict order.
 // * If we do any of the steps, we must do all of the required steps.
-// * We can't use messageMappings in between the first and second steps; e.g.
-//   we can't do any layout, since that uses numberOfItemsInSection: and
-//   interactionAtIndexPath: which use the messageMappings.
+// * We can't use messageMappings or viewItems after the first step until we've
+//   done the last step; i.e.. we can't do any layout, since that uses the view
+//   items which haven't been updated yet.
+// * If the first and/or second steps changes the set of messages
+//   their ordering and/or their state, we must do the third and fourth steps.
 // * If we do the third step, we must call resetContentAndLayout afterward.
 @property (nonatomic) YapDatabaseConnection *uiDatabaseConnection;
 @property (nonatomic) YapDatabaseViewMappings *messageMappings;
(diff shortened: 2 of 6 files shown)


Signal/src/ViewControllers/ConversationView/ConversationInputTextView.h
```diff
@@ -10,8 +10,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)didPasteAttachment:(SignalAttachment *_Nullable)attachment;
 
-- (void)textViewDidChangeLayout;
```
charlesmchen: We don't need to observe the layout of the input toolbar or text view anymore; it's redundant with and less useful than our new approach of observing layout changes to the collection view.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -606,8 +609,6 @@ - (void)scrollToDefaultPosition
     } else {
         [self scrollToBottomAnimated:NO];
     }
-
-    [self updateLastVisibleTimestamp];
```
charlesmchen: We need a thorough test plan around this, but think we were updating the "last visible" state way too often.  It should be sufficient to update it:

* On view will appear.
* When collection view contents are changed with `performBatchUpdates`.
* When collection view scroll state changes.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2239,7 +2226,6 @@ - (void)ensureScrollDownButton
     OWSAssert([NSThread isMainThread]);
 
     BOOL shouldShowScrollDownButton = NO;
-    NSUInteger numberOfMessages = [self.messageMappings numberOfItemsInSection:0];
```
charlesmchen: We want to treat the "view items" as the single source of truth around the collection view contents and consult the `messageMappings` only when updating the "view items".

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2248,12 +2234,11 @@ - (void)ensureScrollDownButton
     // one page.
     BOOL isScrolledUp = scrollSpaceToBottom > pageHeight * 1.f;
 
-    if (numberOfMessages > 0) {
-        TSInteraction *lastInteraction =
-            [self interactionAtIndexPath:[NSIndexPath indexPathForRow:(NSInteger)numberOfMessages - 1 inSection:0]];
```
charlesmchen: Similarly, we want to stop using `interactionAtIndexPath` in favor of the view items.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3184,42 +3163,11 @@ - (void)attachmentButtonPressed
...
-    ConversationViewItem *_Nullable viewItem = [self viewItemForIndex:(NSUInteger)indexPath.row];
-    return viewItem.interaction;
-}
-
 - (void)updateLastVisibleTimestamp
 {
-    TSInteraction *lastVisibleInteraction = [self lastVisibleInteraction];
-    if (lastVisibleInteraction) {
-        uint64_t lastVisibleTimestamp = lastVisibleInteraction.timestampForSorting;
+    ConversationViewItem *_Nullable lastViewItem = [self.viewItems lastObject];
```
michaelkirk: How do we know that the `self.viewItems.lastObject` is visible?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3994,13 +3922,15 @@ - (void)reloadInteractionForViewItem:(ConversationViewItem *)viewItem
...
-- (nullable ConversationViewItem *)viewItemForIndex:(NSUInteger)index
+- (nullable ConversationViewItem *)viewItemForIndex:(NSInteger)index
 {
-    if (index >= self.viewItems.count) {
+    if (index < 0 || index >= (NSInteger)self.viewItems.count) {
         OWSFail(@"%@ Invalid view item index: %zd", self.tag, index);
         return nil;
     }
-    return self.viewItems[index];
+    ConversationViewItem *_Nullable viewItem = self.viewItems[(NSUInteger)index];
```
michaelkirk: When would this be nil?

charlesmchen: You're right; this is redundant with the `OWSFail()` above. Removed.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen: Good catch, this is wrong. Opening a new PR.

Sorry, I missed your comments last night.

PR #2670, opened 2017-10-19
Revert from numeric to default keyboard after message send.

charlesmchen:
PTAL @michaelkirk


Signal/src/ViewControllers/OWSViewController.h
```diff
@@ -15,6 +15,8 @@ NS_ASSUME_NONNULL_BEGIN
 // Use this method in lieu of autoPinToBottomLayoutGuideOfViewController:
 - (void)autoPinViewToBottomGuideOrKeyboard:(UIView *)view;
 
+- (void)setShouldIgnoreKeyboardChanges:(BOOL)value;
```
michaelkirk: TIOLI: consider making this a public, writable, property

charlesmchen: Done.

PR #2671, opened 2017-10-19
Fix "can't send same GIF twice" issue.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2330,9 +2330,6 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
     OWSAssert([NSThread isMainThread]);
     OWSAssert(message);
 
-    [self.editingDatabaseConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
-        [message saveWithTransaction:transaction];
-    }];
     [self updateLastVisibleTimestamp:message.timestampForSorting];
     self.lastMessageSentDate = [NSDate new];
     [self clearUnreadMessagesIndicator];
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2330,9 +2330,6 @@ - (void)messageWasSent:(TSOutgoingMessage *)message
     OWSAssert([NSThread isMainThread]);
     OWSAssert(message);
 
-    [self.editingDatabaseConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
```
charlesmchen: We want to save new outgoing messages "lazily" when they have been "finalized".  In the case of outgoing attachments, that means when the attachments have been (asynchronously) written to disk and their attachment ids adds to the message.  Otherwise the view may try to present them when they are incomplete.  So this should be the responsibility of `ThreadUtil` which will save messages _after_ they are finalized but _before_ they are enqueued in the outgoing message queue.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -364,10 +364,15 @@ - (void)ensureViewState
                 self.messageCellType = OWSMessageCellType_DownloadingAttachment;
                 self.attachmentPointer = (TSAttachmentPointer *)attachment;
                 return;
+            } else {
+                OWSFail(@"%@ Unknown attachment type", self.tag);
             }
+        } else {
```
michaelkirk: TIOLI consider inverting these into a guard pattern.

charlesmchen: I think it's better to "fall through" in the error cases because:

* There's more than one of them.
* Each one requires the same 4-5 lines before it can exit.

charlesmchen:
@michaelkirk Whoops, this should have been opened against `release/2.18.0`.  I've picked it over.

PR #2673, opened 2017-10-19
GIF picker UX tweaks

michaelkirk:
Issues Addressed:

## Less time to animated loading

GIF picker can be difficult to use on a slow data connection. Primarily this was because we were downloading a full-sized animated asset for each cell. Now we download a small one for the picker view, and only download a full-sized asset once it's selected. This results in a 65-90% reduction in overall bandwidth used, and also a quicker picker navigation experience.

## Unable to select cells until animated asset was downloaded.

Previously, tapping on the still wouldn't select the GIF, and there was no user feedback as to why. Now, once you tap a cell (as long as at least the still has been loaded), we indicate which cell is selected, show an activity indicator, and block further interaction with the collection view until the asset is completely fetched.

## Dismiss keyboard after progressive search

We were dismissing keyboard when the user hit the "search" button, but if the user scrolls around after the auto-search, we should dismiss the keyboard.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/GifPicker/GifPickerCell.swift
+++ b/Signal/src/ViewControllers/GifPicker/GifPickerCell.swift
@@ -110,7 +110,7 @@ class GifPickerCell: UICollectionViewCell {
 
         // Record high quality animated rendition, but to save bandwidth, don't start downloading
         // until it's selected.
-        guard let highQualityAnimatedRendition = imageInfo.pickHighQualityAnimatedRendition() else {
+        guard let highQualityAnimatedRendition = imageInfo.pickSendingRendition() else {
             Logger.warn("\(TAG) could not pick gif rendition: \(imageInfo.giphyId)")
             clearAssetRequests()
             return
@@ -119,7 +119,7 @@ class GifPickerCell: UICollectionViewCell {
 
         // The Giphy API returns a slew of "renditions" for a given image. 
         // It's critical that we carefully "pick" the best rendition to use.
-        guard let animatedRendition = imageInfo.pickAnimatedRendition() else {
+        guard let animatedRendition = imageInfo.pickPreviewRendition() else {
             Logger.warn("\(TAG) could not pick gif rendition: \(imageInfo.giphyId)")
             clearAssetRequests()
             return
--- a/Signal/src/ViewControllers/GifPicker/GifPickerCell.swift
+++ b/Signal/src/ViewControllers/GifPicker/GifPickerCell.swift
@@ -3,11 +3,16 @@
 //
 
 import Foundation
+import PromiseKit
 
 class GifPickerCell: UICollectionViewCell {
     let TAG = "[GifPickerCell]"
 
     // MARK: Properties
+    enum GifPickerCellError: Error {
+        case assertionError(description: String)
+        case fetchFailure
+    }
 
     var imageInfo: GiphyImageInfo? {
         didSet {
@@ -34,6 +39,9 @@ class GifPickerCell: UICollectionViewCell {
     var animatedAssetRequest: GiphyAssetRequest?
     var animatedAsset: GiphyAsset?
     var imageView: YYAnimatedImageView?
+
+    // As another bandwidth saving measure, we only fetch the full sized GIF when the user selects it.
+    private var renditionForSending: GiphyRendition?
 
     // MARK: Initializers
 
@@ -93,6 +101,16 @@ class GifPickerCell: UICollectionViewCell {
             clearAssetRequests()
             return
         }
+
+        // Record high quality animated rendition, but to save bandwidth, don't start downloading
+        // until it's selected.
+        guard let highQualityAnimatedRendition = imageInfo.pickHighQualityAnimatedRendition() else {
+            Logger.warn("\(TAG) could not pick gif rendition: \(imageInfo.giphyId)")
+            clearAssetRequests()
+            return
+        }
+        self.renditionForSending = highQualityAnimatedRendition
+
         // The Giphy API returns a slew of "renditions" for a given image. 
         // It's critical that we carefully "pick" the best rendition to use.
         guard let animatedRendition = imageInfo.pickAnimatedRendition() else {
@@ -188,6 +206,31 @@ class GifPickerCell: UICollectionViewCell {
         }
         imageView.image = image
         self.backgroundColor = nil
+    }
+
+    public func fetchRenditionForSending() -> Promise<GiphyAsset> {
+        guard let renditionForSending = self.renditionForSending else {
+            owsFail("\(TAG) renditionForSending was unexpectedly nil")
+            return Promise(error: GifPickerCellError.assertionError(description: "renditionForSending was unexpectedly nil"))
+        }
+
+        let (promise, fulfill, reject) = Promise<GiphyAsset>.pending()
+
+        // We don't retain a handle on the asset request, since there will only ever
+        // be one selected asset, and we never want to cancel it.
+        _ = GiphyDownloader.sharedInstance.requestAsset(rendition: renditionForSending,
+                                                    priority: .high,
+                                                    success: { _, asset in
+                                                        fulfill(asset)
+        },
+                                                    failure: { _ in
+                                                        // TODO GiphyDownloader API shoudl pass through a useful failing error
+                                                        // so we can pass it through here
+                                                        reject(GifPickerCellError.fetchFailure)
+
+        })
+
+        return promise
     }
 
     private func clearViewState() {
--- a/Signal/src/ViewControllers/GifPicker/GifPickerCell.swift
+++ b/Signal/src/ViewControllers/GifPicker/GifPickerCell.swift
@@ -221,7 +221,7 @@ class GifPickerCell: UICollectionViewCell {
             activityIndicator.startAnimating()
 
             // Render activityIndicator on a white tile to ensure it's visible on
-            // when overalayed on a variety of potential gifs.
+            // when overlayed on a variety of potential gifs.
             activityIndicator.backgroundColor = UIColor.white.withAlphaComponent(0.3)
             activityIndicator.autoSetDimension(.width, toSize: 30)
             activityIndicator.autoSetDimension(.height, toSize: 30)
@@ -249,7 +249,7 @@ class GifPickerCell: UICollectionViewCell {
                                                             fulfill(asset)
         },
                                                         failure: { _ in
-                                                            // TODO GiphyDownloader API shoudl pass through a useful failing error
+                                                            // TODO GiphyDownloader API should pass through a useful failing error
                                                             // so we can pass it through here
                                                             Logger.error("\(self.TAG) request failed")
                                                             reject(GiphyError.fetchFailure)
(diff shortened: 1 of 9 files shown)


Signal/src/ViewControllers/GifPicker/GifPickerCell.swift
```diff
@@ -188,6 +212,50 @@ class GifPickerCell: UICollectionViewCell {
...
+
+        if self.isCellSelected {
+            let activityIndicator = UIActivityIndicatorView(activityIndicatorStyle: .gray)
+            self.activityIndicator = activityIndicator
+            addSubview(activityIndicator)
+            activityIndicator.autoCenterInSuperview()
+            activityIndicator.startAnimating()
+
+            // Render activityIndicator on a white tile to ensure it's visible on
+            // when overalayed on a variety of potential gifs.
```
charlesmchen: overlayed

charlesmchen: CALayer makes it easy to add a drop shadow - I wonder if we should use a subtle one to differentiate the tile from the background?

Signal/src/ViewControllers/GifPicker/GifPickerCell.swift
```diff
@@ -188,6 +212,50 @@ class GifPickerCell: UICollectionViewCell {
...
+        // We don't retain a handle on the asset request, since there will only ever
+        // be one selected asset, and we never want to cancel it.
+        _ = GiphyDownloader
+            .sharedInstance.requestAsset(rendition: renditionForSending,
+                                                        priority: .high,
+                                                        success: { _, asset in
+                                                            fulfill(asset)
+        },
+                                                        failure: { _ in
+                                                            // TODO GiphyDownloader API shoudl pass through a useful failing error
```
charlesmchen: should

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
```diff
@@ -169,7 +161,9 @@ class GifPickerViewController: OWSViewController, UISearchBarDelegate, UICollect
         self.collectionView.dataSource = self
         self.collectionView.backgroundColor = UIColor.white
         self.collectionView.register(GifPickerCell.self, forCellWithReuseIdentifier: kCellReuseIdentifier)
-        self.view.addSubview(self.collectionView)
+        // Inserted below searchbar becase we later occlude the collectionview
```
charlesmchen: because

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
```diff
@@ -31,16 +31,16 @@ class GifPickerViewController: OWSViewController, UISearchBarDelegate, UICollect
...
+    let messageSender: MessageSender
 
     let searchBar: UISearchBar
     let layout: GifPickerLayout
     let collectionView: UICollectionView
     var noResultsView: UILabel?
     var searchErrorView: UILabel?
     var activityIndicator: UIActivityIndicatorView?
-
+    var selectedCell: UICollectionViewCell?
```
charlesmchen: I think this should be `selectedImageInfo` or `selectedSendingRendition`, not `selectedCell`.  `UICollectionView` can reload its contents at any time including during download of the "sending rendition".  The selected `GifPickerCell` might be reused, making the "retry download" logic below unsafe.

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
```diff
@@ -295,36 +300,79 @@ class GifPickerViewController: OWSViewController, UISearchBarDelegate, UICollect
...
+            owsFail("\(TAG) Already selected cell")
             return
         }
-        guard let messageSender = messageSender else {
-            owsFail("\(TAG) Missing messageSender.")
-            return
+        self.selectedCell = cell
+
+        // Fade out all cells except the selected one.
+        let maskingView = OWSBezierPathView()
```
charlesmchen: Is a `OWSBezierPathView` necessary? Could we just use a UIView and set it's background color and opacity?

Signal/src/network/GiphyAPI.swift
```diff
@@ -89,8 +104,9 @@ enum GiphyFormat {
 
     // TODO: We may need to tweak these constants.
     let kMaxDimension = UInt(618)
-    let kMinDimension = UInt(101)
-    let kMaxFileSize = UInt(3 * 1024 * 1024)
+    let kMinDimension = UInt(60)
```
charlesmchen: I think we'll want a min dimension of 101 for the sending rendition.

Signal/src/network/GiphyAPI.swift
```diff
@@ -105,20 +121,33 @@ enum GiphyFormat {
 
     public func pickStillRendition() -> GiphyRendition? {
         // Stills are just temporary placeholders, so use the smallest still possible.
-        return pickRendition(isStill:true, pickingStrategy:.smallerIsBetter, maxFileSize:kMaxFileSize)
+        return pickRendition(isStill:true, pickingStrategy:.smallerIsBetter, maxFileSize:kPreferedPreviewFileSize)
     }
 
     public func pickAnimatedRendition() -> GiphyRendition? {
```
charlesmchen: Maybe: `pickPreviewRendition` ?

Signal/src/network/GiphyAPI.swift
```diff
@@ -105,20 +121,33 @@ enum GiphyFormat {
...
+        }
+        // ...but gradually relax the file restriction...
+        if let rendition = pickRendition(isStill:false, pickingStrategy:.smallerIsBetter, maxFileSize:kPreferedPreviewFileSize * 2) {
+            return rendition
+        }
+        // ...and relax even more until we find an animated rendition.
+        return pickRendition(isStill:false, pickingStrategy:.smallerIsBetter, maxFileSize:kPreferedPreviewFileSize * 3)
+    }
+
+    public func pickHighQualityAnimatedRendition() -> GiphyRendition? {
```
charlesmchen: Maybe `pickSendingRendition`?

Signal/src/network/GiphyAPI.swift
```diff
@@ -105,20 +121,33 @@ enum GiphyFormat {
 
     public func pickStillRendition() -> GiphyRendition? {
```
charlesmchen: If we're going to use small animated previews, I wonder if even need still?

charlesmchen:
I'm a bit conflicted about using smaller GIFs because Giphy renditions vary wildly in quality based on the source content: width, height, length, "noisiness".  For longer, noisier content, the previews could be quite crappy in quality and the picker will be pretty far from WYSIWIG.  But I'm definitely will to try this and see how it goes.  

I did have one or two blocking concerns in the implementation.

charlesmchen:
> only download a full-sized asset once it's selected

I want to make sure we're handling the edge cases here.  Is the full-size download cancellable?  If so, is the picker left in a broken state since we cancel all ongoing asset requests when we select a cell?

Signal/src/ViewControllers/GifPicker/GifPickerCell.swift
michaelkirk: done!

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
michaelkirk: done!

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
michaelkirk: Good point. Currently we were only using it's presence as a boolean anyway, so I just changed it to a boolean `hasSelectedCell`.

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
michaelkirk: If we had a solid rectangular mask I would do that, but here we bunch a hole out of the mask, so that the **selected** view is not masked, but everything else is. 

Let me know if there's an easier way to do this.

Sorry I should have included a screenshot in the PR, here's what it looks like:

screen shot 2017-10-20 at 9 52 51 am

Signal/src/network/GiphyAPI.swift
michaelkirk: Done. I've added a rendition type enum: `.stillPreview, .animatedLowQuality, .animatedHighQuality`

Signal/src/network/GiphyAPI.swift
michaelkirk: better. Done.

Signal/src/network/GiphyAPI.swift
michaelkirk: better. Done.

michaelkirk:
> I want to make sure we're handling the edge cases here. Is the full-size download cancellable? If so, is the picker left in a broken state since we cancel all ongoing asset requests when we select a cell?

This brings up a good point. The download itself isn't expclicitly cancelable, but if you hit "cancel" in the navbar while we're waiting to download, we dismiss the VC, so the VC isn't in a bad state, but the download continues and the GIF is sent when it eventually finishes downloading, which is probably wrong. I'll take a look at that.

Signal/src/network/GiphyAPI.swift
michaelkirk: Here's an example load (no scrolling)
```
2017-10-20 11:32:42:752 Signal[72855:2827693] [GiphyDownloader] Still bytes written:    000.06 MB
2017-10-20 11:32:42:752 Signal[72855:2827693] [GiphyDownloader] Animated bytes written: 001.10 MB
2017-10-20 11:32:42:752 Signal[72855:2827693] [GiphyDownloader] Total bytes written:    001.15 MB
```

S the stills are 20x smaller than even the "small" GIF renditions.

On fast connections, probably the stills are unnecessary, but on a slow connection, I think it makes sense to include them. Picking from a still then cancels any downloads of the animatedPreviews.

michaelkirk:
> I'm a bit conflicted about using smaller GIFs because Giphy renditions vary wildly in quality based on the source content: width, height, length, "noisiness". For longer, noisier content, the previews could be quite crappy in quality and the picker will be pretty far from WYSIWIG. But I'm definitely will to try this and see how it goes.

Yeah - give it a try! There is like some more fine tuning to be done on resolution and size limits.

michaelkirk:
OK - I think I've addressed everything.

PTAL @charlesmchen

Signal/src/ViewControllers/GifPicker/GifPickerCell.swift
michaelkirk: done.

charlesmchen:
If you pick cell A, can you pick cell B while A is downloading? Or are you committed?

charlesmchen:
> Yeah - give it a try! There is like some more fine tuning to be done on resolution and size limits.

FYI another option is to use Giphy's "downsampled" renditions which I'm currently ignoring in the parsing or picking logic.

michaelkirk:
> If you pick cell A, can you pick cell B while A is downloading? Or are you committed?

> FYI another option is to use Giphy's "downsampled" renditions which I'm currently ignoring in the parsing or picking logic.

I actually did try this, but found the choppiness too unpleasant. Maybe with the right set of params it could be useful.

michaelkirk:
> If you pick cell A, can you pick cell B while A is downloading? Or are you committed?

You are committed, except that you can cancel-to-dismiss the view controller altogether, which will cancel your selection.

michaelkirk:
rebased and merged in b6c9a2a6

PR #2676, opened 2017-10-20
Fix proxying of Giphy requests.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
> TIOLI, but since this is kind of touchy can we extract and reuse in both places?

Done.

PR #2680, opened 2017-10-22
Use proper color when messages fail to send.

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
@@ -226,9 +226,18 @@ NS_ASSUME_NONNULL_BEGIN
 
     DDLogError(@"%p loadForDisplay: %@", self, NSStringForOWSMessageCellType(self.cellType));
 
-    BOOL isIncoming = self.isIncoming;
-    JSQMessagesBubbleImage *bubbleImageData
-        = isIncoming ? [self.bubbleFactory incoming] : [self.bubbleFactory outgoing];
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -226,9 +226,18 @@ - (void)loadForDisplay
...
-    BOOL isIncoming = self.isIncoming;
-    JSQMessagesBubbleImage *bubbleImageData
-        = isIncoming ? [self.bubbleFactory incoming] : [self.bubbleFactory outgoing];
+    JSQMessagesBubbleImage *bubbleImageData;
+    if ([self.viewItem.interaction isKindOfClass:[TSOutgoingMessage class]]) {
+        TSOutgoingMessage *outgoingMessage = (TSOutgoingMessage *)self.viewItem.interaction;
+        if (outgoingMessage.messageState == TSOutgoingMessageStateUnsent) {
+            bubbleImageData = self.bubbleFactory.outgoingFailed;
+        } else {
+            bubbleImageData = self.bubbleFactory.outgoing;
```
charlesmchen: bubbleImageData isn't used in a number of cases, like images.  See `replaceBubbleWithView:` for a complete list of cases.  We need to do something similar for these cases too right?

michaelkirk: Historically we haven't done more than print the "failed to send" footer for media. I agree that we should do more to indicate failed attachments - probably something like Signal-Android.

PR #2685, opened 2017-10-23
Fix layout of generic and audio messages.

charlesmchen:
PTAL @michaelkirk 

screen shot 2017-10-23 at 1 26 55 pm


Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
```diff
@@ -116,25 +117,37 @@ - (void)createContents
     [contentView autoPinEdgeToSuperviewEdge:ALEdgeTop withInset:self.vMargin];
     [contentView autoPinEdgeToSuperviewEdge:ALEdgeBottom withInset:self.vMargin];
 
+    OWSBezierPathView *iconCircleView = [OWSBezierPathView new];
+    UIColor *iconColor
+        = (self.isIncoming ? [UIColor colorWithRGBHex:0x9e9e9e] : [self foregroundColorWithOpacity:0.15f]);
+    iconCircleView.configureShapeLayerBlock = ^(CAShapeLayer *_Nonnull layer, CGRect bounds) {
+        CGFloat radius = MIN(bounds.size.width, bounds.size.height) * 0.5f;
+        layer.path = [UIBezierPath bezierPathWithRoundedRect:bounds cornerRadius:radius].CGPath;
```
michaelkirk: TIOLI, but would `[UIBezierPath bezierPathWithOvalInRect:bounds].CGPath` save you a line?

Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
```diff
@@ -116,25 +117,37 @@ - (void)createContents
...
+        layer.fillColor = iconColor.CGColor;
+    };
+    [contentView addSubview:iconCircleView];
+    [iconCircleView autoPinEdgeToSuperviewEdge:ALEdgeLeft withInset:self.iconHMargin];
+    [iconCircleView autoVCenterInSuperview];
+    [iconCircleView autoSetDimension:ALDimensionWidth toSize:self.iconSize];
+    [iconCircleView autoSetDimension:ALDimensionHeight toSize:self.iconSize];
+    [iconCircleView setContentHuggingHigh];
```
michaelkirk: No biggie, but since we're explicitly setting the width and height, does setting contentHugging have any affect?

Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
```diff
@@ -116,25 +117,37 @@ - (void)createContents
...
     imageView.tintColor = self.bubbleBackgroundColor;
-    imageView.backgroundColor
-        = (self.isIncoming ? [UIColor colorWithRGBHex:0x9e9e9e] : [self foregroundColorWithOpacity:0.15f]);
-    imageView.layer.cornerRadius = MIN(imageView.bounds.size.width, imageView.bounds.size.height) * 0.5f;
     [contentView addSubview:imageView];
-    [imageView autoPinEdgeToSuperviewEdge:ALEdgeLeft withInset:self.iconHMargin];
-    [imageView autoVCenterInSuperview];
-    [imageView autoSetDimension:ALDimensionWidth toSize:self.iconSize];
-    [imageView autoSetDimension:ALDimensionHeight toSize:self.iconSize];
+    [imageView autoPinEdge:ALEdgeLeft toEdge:ALEdgeLeft ofView:iconCircleView];
```
michaelkirk: TIOLI: it seems like the idiomatic thing to do here is make imageView a subview of iconCircleView with 

```
[iconCircleView addSubview:imageView];
[imageView autoPinEdgesToSuperviewEdges];
```

Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
charlesmchen: Better thanks.

Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
charlesmchen: Done.

Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
charlesmchen: Unfortunately that doesn't work.  TIL `OWSBezierPathView` won't render subviews.  I'm not sure why... could it be because we're constantly recreates the `OWSBezierPathView`'s layers?

michaelkirk: Ah right... probably because we do:

```
    for (CALayer *layer in self.layer.sublayers) {
        [layer removeFromSuperlayer];
    }
```

Probably not worth fixing for this use.

PR #2686, opened 2017-10-23
Fix layout glitches in JSQ rewrite.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationCollectionView.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationCollectionView.m
@@ -15,6 +15,13 @@ NS_ASSUME_NONNULL_BEGIN
     if (frame.size.width == 0 || frame.size.height == 0) {
         // Ignore iOS Auto Layout's tendency to temporarily zero out the
         // frame of this view during the layout process.
+        //
+        // The conversation view has an invariant that the collection view
+        // should always have a "reasonable" (correct width, non-zero height)
+        // size.  This lets us manipulate scroll state at all times, especially
+        // before the view has been presented for the first time.  This
+        // invariant also saves us from needing all sorts of ugly and incomplete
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
```diff
@@ -311,7 +313,10 @@ - (void)ensureContentConstraints
         [rightButton autoPinEdgeToSuperviewEdge:ALEdgeBottom],
     ];
 
-    [self layoutIfNeeded];
+    // Layout immediately, unless the input toolbar hasn't even been laid out yet.
+    if (self.bounds.size.width > 0 && self.bounds.size.height > 0) {
```
charlesmchen: This condition avoids some misleading "iOS auto layout constraints cannot be satisfied" logging.

Signal/src/ViewControllers/ConversationView/ConversationCollectionView.m
```diff
@@ -12,6 +12,11 @@ @implementation ConversationCollectionView
 
 - (void)setFrame:(CGRect)frame
 {
+    if (frame.size.width == 0 || frame.size.height == 0) {
+        // Ignore iOS Auto Layout's tendency to temporarily zero out the
+        // frame of this view during the layout process.
+        return;
+    }
```
michaelkirk: Is this a perf optimization?

I'm a little worried about skipping the `super` call on such fundamental layout logic - but it'll be exercised enough that any terrible bug should be exposed pretty quickly.

charlesmchen: I've added a comment that explains.  This (I believe perfectly safe) hack spares us from all sorts of worse hacks which used to use - which never worked well anyway.

PR #2688, opened 2017-10-23
Charlesmchen/one system contacts fetch at a time

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/contact/SystemContactsFetcher.swift
+++ b/Signal/src/contact/SystemContactsFetcher.swift
@@ -362,7 +362,8 @@ class SystemContactsFetcher: NSObject {
         hasSetupObservation = true
         self.contactStoreAdapter.startObservingChanges { [weak self] in
             DispatchQueue.main.async {
-                self?.updateContacts(completion: nil)
+                // If contacts have changed, don't de-bounce.
+                self?.updateContacts(completion: nil, alwaysNotify: false, shouldDebounce: false)
             }
         }
     }
@@ -418,13 +419,13 @@ class SystemContactsFetcher: NSObject {
         }
     }
 
-    public func fetchIfAlreadyAuthorized(ignoreDebounce: Bool = false) {
+    public func fetchIfAlreadyAuthorized(alwaysNotify: Bool = false) {
         AssertIsOnMainThread()
         guard authorizationStatus == .authorized else {
             return
         }
 
-        updateContacts(completion: nil, ignoreDebounce:ignoreDebounce)
+        updateContacts(completion: nil, alwaysNotify:alwaysNotify)
     }
 
     private func tryToAcquireContactFetchLock() -> Bool {
@@ -444,7 +445,7 @@ class SystemContactsFetcher: NSObject {
         objc_sync_exit(self)
     }
 
-    private func updateContacts(completion: ((Error?) -> Void)?, ignoreDebounce: Bool = false) {
+    private func updateContacts(completion: ((Error?) -> Void)?, alwaysNotify: Bool = false, shouldDebounce: Bool = true) {
         AssertIsOnMainThread()
 
         systemContactsHaveBeenRequestedAtLeastOnce = true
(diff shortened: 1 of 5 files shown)


Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -434,15 +452,21 @@ class SystemContactsFetcher: NSObject {
 
         DispatchQueue.global().async {
 
-            var fetchedContacts: [Contact]?
+            guard self.tryToAcquireContactFetchLock() else {
```
michaelkirk: Seems like there's a race condition where we could lose important contact edits this way. If I launch the app, then while my contacts are still loading, I edit someone in my address book, I might not get that edit pulled into Signal until *next* contact pull.

Maybe that's an edge case that we can reasonably ignore, but we could also use a serial queue instead.

WDYT @charlesmchen

Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -434,15 +452,21 @@ class SystemContactsFetcher: NSObject {
...
+            var fetchedContacts: [Contact]?
             switch self.contactStoreAdapter.fetchContacts() {
             case .success(let result):
                 fetchedContacts = result
             case .error(let error):
                 completion?(error)
+                self.releaseContactFetchLock()
                 return
             }
+            self.releaseContactFetchLock()
```
michaelkirk: maybe a good use for a swift `defer` block.

Signal/src/contact/SystemContactsFetcher.swift
charlesmchen: Yup, that's the definite tradeoff here.  It's hard because this is a major perf win on iPhone 5C if you have a lot of contacts.

Maybe we can have two types of fetch requests: soft and hard.  Hard are always done, serially.  Soft can be de-duplicated.

Signal/src/contact/SystemContactsFetcher.swift
charlesmchen: Done.

charlesmchen:
PTAL @michaelkirk 

I have:

* Renamed `ignoreDebounce` to `alwaysNotify`.
* Added `shouldDebounce` so that when system contacts change we can ensure that we always do a fetch.
* With the introduction of ``shouldDebounce` we can once again more than one fetch at a time, and thus race conditions but I'm not sure this is worth addressing.
* I wonder if we couldn't simplify by modifying `fetchIfAlreadyAuthorized` to noop if `systemContactsHaveBeenRequestedAtLeastOnce` like `requestOnce`.  We could then move fetches onto a serial queue, remove debounce logic. WDYT?

Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -417,25 +419,55 @@ class SystemContactsFetcher: NSObject {
...
         DispatchQueue.global().async {
 
-            var fetchedContacts: [Contact]?
+            if shouldDebounce {
+                guard self.tryToAcquireContactFetchLock() else {
+                    Logger.info("\(self.TAG) ignoring redundant system contacts fetch.")
+                    return
+                }
+            }
+            defer {
```
michaelkirk: Can you put the defer in the previous `if shouldDebounce` block to put the "teardown" code as logically close to the "setup" code as possible.

michaelkirk:
> I wonder if we couldn't simplify by modifying fetchIfAlreadyAuthorized to noop if systemContactsHaveBeenRequestedAtLeastOnce like requestOnce. We could then move fetches onto a serial queue, remove debounce logic. WDYT?

I'm also open to this idea. Adding a "override" flag with a default value is expedient, but it gets harder to understand what the method does. Especially when there are multiple override flags with default values.

Signal/src/contact/SystemContactsFetcher.swift
charlesmchen: No, because then the `defer` executes when you leave the condition block.

michaelkirk: I am surprised and appalled.

charlesmchen:
PTAL @michaelkirk Reworked this PR yet again, as discussed.

PR #2691, opened 2017-10-24
Fixes for input toolbar.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
@@ -12,7 +12,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic) UILabel *placeholderView;
 @property (nonatomic) NSArray<NSLayoutConstraint *> *placeholderConstraints;
-@property (nonatomic) BOOL isEditing;
 
 @end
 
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
```diff
@@ -12,7 +12,6 @@ @interface ConversationInputTextView () <UITextViewDelegate>
 
 @property (nonatomic) UILabel *placeholderView;
 @property (nonatomic) NSArray<NSLayoutConstraint *> *placeholderConstraints;
-@property (nonatomic) BOOL isEditing;
```
charlesmchen: Separate fix: we want to show the input placeholder whenever the text input is empty, even if user is editing.

Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.h
```diff
@@ -34,7 +34,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @protocol ConversationInputTextViewDelegate;
 
-@interface ConversationInputToolbar : UIToolbar
+@interface ConversationInputToolbar : UIView
```
charlesmchen: It turns out that it's not safe to add subviews to `UIToolbar` in its initializer.  We weren't actually using any `UIToolbar` behavior, so we'll just use a `UIView` instead.

Fixes @michaelkirk 's "can tap on input toolbar on iOS 11" issue.

PR #2693, opened 2017-10-24
Interactive/Cancelable slide left for details

michaelkirk:
This is much more tenable with the new perf improvements.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
@@ -162,6 +162,12 @@ NS_ASSUME_NONNULL_BEGIN
     UILongPressGestureRecognizer *longPress =
         [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(handleLongPressGesture:)];
     [self addGestureRecognizer:longPress];
+
+    PanDirectionGestureRecognizer *panGesture =
+        [[PanDirectionGestureRecognizer alloc] initWithDirection:PanDirectionHorizontal
+                                                          target:self
+                                                          action:@selector(handlePanGesture:)];
+    [self addGestureRecognizer:panGesture];
 }
 
 + (NSString *)cellReuseIdentifier
@@ -1027,6 +1033,13 @@ NS_ASSUME_NONNULL_BEGIN
         CGPoint location = [sender locationInView:self];
         [self showMenuController:location];
     }
+}
+
+- (void)handlePanGesture:(UIPanGestureRecognizer *)panRecognizer
+{
+    OWSAssert(self.delegate);
+
+    [self.delegate didPanWithGestureRecognizer:panRecognizer viewItem:self.viewItem];
 }
 
 #pragma mark - UIMenuController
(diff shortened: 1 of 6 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -984,9 +988,14 @@ - (void)viewWillDisappear:(BOOL)animated
 
     [super viewWillDisappear:animated];
 
-    self.isViewVisible = NO;
-
     [self.inputToolbar viewWillDisappear:animated];
+}
+
+- (void)viewDidDisappear:(BOOL)animated
```
michaelkirk: viewWillDisappear is called even if the transition away is cancelled. We should delay all possible teardown until viewDidDisappear.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -162,6 +162,12 @@ - (void)commontInit
     UILongPressGestureRecognizer *longPress =
         [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(handleLongPressGesture:)];
     [self addGestureRecognizer:longPress];
+
+    PanDirectionGestureRecognizer *panGesture =
+        [[PanDirectionGestureRecognizer alloc] initWithDirection:PanDirectionHorizontal
+                                                          target:self
+                                                          action:@selector(handlePanGesture:)];
+    [self addGestureRecognizer:panGesture];
```
charlesmchen: Should this GR hang on the cells or on the conversation collection view?  

* This won't work for pans that start in offers, errors, calls, unread indicator.
* This won't work for pans that start outside messages (on the left of outgoing, on the right of incoming).

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen: Maybe add a comment in `viewWillDisappear` noting that we shouldn't add logic there?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4015,6 +4018,110 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView
...
 }
 
+#pragma mark - swipe to show message details
+
+- (void)didPanWithGestureRecognizer:(UIPanGestureRecognizer *)gestureRecognizer
+                           viewItem:(ConversationViewItem *)conversationItem
+{
+    self.currentShowMessageDetailsPanGesture = gestureRecognizer;
+
+    const CGFloat leftTranslation = -1 * [gestureRecognizer translationInView:self.view].x;
```
charlesmchen: Do we need to clamp this?  Could users reverse the gesture and pan back past the original arrangement?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen: Oh I see this is to show message metadata.  Please ignore.

charlesmchen:
> This is much more tenable with the new perf improvements.

We gotta test this on iOS 8, 9, 10 and 11.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk: Good point. Done!

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk: Not that I could reproduce, but I've now clamped it between 0-1.0 anyway.

PR #2695, opened 2017-10-24
Don't send with return key.

charlesmchen:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
```diff
@@ -48,7 +48,7 @@ - (instancetype)init
         self.dataDetectorTypes = UIDataDetectorTypeNone;
         self.keyboardAppearance = UIKeyboardAppearanceDefault;
         self.keyboardType = UIKeyboardTypeDefault;
-        self.returnKeyType = UIReturnKeySend;
+        self.returnKeyType = UIReturnKeyDefault;
```
michaelkirk: Consider just deleting this.

charlesmchen: Done.

PR #2697, opened 2017-10-24
RTL fixes for JSQ rewrite.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
@@ -55,7 +55,10 @@ NS_ASSUME_NONNULL_BEGIN
         return;
     }
     maskedSubview.frame = self.bounds;
-    [JSQMessagesMediaViewBubbleImageMasker applyBubbleImageMaskToMediaView:maskedSubview isOutgoing:self.isOutgoing];
+    // The JSQ masks are RTL-safe, so we need to invert the
+    // mask orientation manually.
+    BOOL hasOutgoingMask = self.isOutgoing ^ self.isRTL;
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -55,7 +55,10 @@ - (void)updateMask
         return;
     }
     maskedSubview.frame = self.bounds;
-    [JSQMessagesMediaViewBubbleImageMasker applyBubbleImageMaskToMediaView:maskedSubview isOutgoing:self.isOutgoing];
+    // The JSQ masks are RTL-safe, so we need to invert the
```
michaelkirk: typo? `s/are RTL-safe/are not RTL-safe/`

It is a good day when you get to xor.

Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
```diff
@@ -56,6 +56,7 @@ - (instancetype)init
         self.placeholderView.text = NSLocalizedString(@"new_message", @"");
         self.placeholderView.textColor = [UIColor lightGrayColor];
         self.placeholderView.userInteractionEnabled = NO;
+        self.placeholderView.textAlignment = NSTextAlignmentNatural;
```
michaelkirk: Isn't this default? Why do we need to specify this?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen: Done.

Signal/src/ViewControllers/ConversationView/ConversationInputTextView.m
charlesmchen: Removed.

PR #2698, opened 2017-10-24
Fix "back button count" flash. Only call once.

michaelkirk:
This affects returning from message info or when leaving conversation
settings.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -474,6 +474,7 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
 
     [self createScrollButtons];
     [self createHeaderViews];
+    [self createBackButton];
     [self addNotificationListeners];
 }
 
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -964,26 +965,25 @@ - (void)cancelReadTimer
...
-    if (_callOnOpen) {
-        [self callAction];
-        _callOnOpen = NO;
+        if (_composeOnOpen && !self.inputToolbar.hidden) {
+            [self popKeyBoard];
+            _composeOnOpen = NO;
+        }
+        if (_callOnOpen) {
+            [self callAction];
+            _callOnOpen = NO;
```
charlesmchen: I'm not sure we have to clear these flags since they are only applied if `! viewHasEverAppeared`

michaelkirk: Good point. I'm actually rethinking this... conceivably there's a world where we'd want to call/compose on appear without reloading the view. e.g. making a call from contacts app. I'm going to move the compose/call logic outside of this block.

PR #2699, opened 2017-10-24
Make message bubbles a bit wider.

michaelkirk:
This is closer to the previous behavior, at least on 5 and 6 sized devices.

Do you feel like this is *too* wide on your plus sized device @charlesmchen ?

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
@@ -760,7 +760,7 @@ NS_ASSUME_NONNULL_BEGIN
     OWSAssert(self.viewItem);
     OWSAssert([self.viewItem.interaction isKindOfClass:[TSMessage class]]);
 
-    const int maxMessageWidth = (int)floor(contentWidth * 0.7f);
+    const int maxMessageWidth = (int)floor(contentWidth * 0.8f);
 
     CGSize cellSize = CGSizeZero;
     switch (self.cellType) {


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -757,7 +757,7 @@ - (CGSize)cellSizeForViewWidth:(int)viewWidth contentWidth:(int)contentWidth
     OWSAssert(self.viewItem);
     OWSAssert([self.viewItem.interaction isKindOfClass:[TSMessage class]]);
 
-    const int maxMessageWidth = (int)floor(contentWidth * 0.7f);
+    const int maxMessageWidth = (int)floor(contentWidth * 0.8f);
```
charlesmchen: Incidentally, I've been thinking of moving this calculation of max message width into `OWSMessageCell`.

charlesmchen: Scratch that, I see that I already did. =P

charlesmchen:
> Do you feel like this is too wide on your plus sized device @charlesmchen ?

Text feels fine on 7+ but images (and other media) seem too large.

screen shot 2017-10-25 at 9 30 06 am

This can be fixed by tweaking the values in:
```
            const CGFloat maxMediaWidth = maxMessageWidth;
            const CGFloat maxMediaHeight = maxMessageWidth;
```
to something like:
```
            const CGFloat maxMediaWidth = round(maxMessageWidth * 0.9);
            const CGFloat maxMediaHeight = round(maxMessageWidth * 0.8);
```
I feel like the max height should be shorter than the max width (since in a conversation vertical screen real estate is the premium) - WDYT?

The downside to reducing maxMediaWidth is that the edges of media and text bubbles are no longer aligned.  Maybe we should do this instead:
```
            const CGFloat maxMediaWidth = maxMessageWidth;
            const CGFloat maxMediaHeight = round(maxMessageWidth * 0.75);
```

michaelkirk:
> Text feels fine on 7+ but images (and other media) seem too large.

It **is** quite large, but we seem to already being more conservative than iMessage, so I've left it as originally proposed here. Happy to revisit.

PR #2700, opened 2017-10-25
Show timer in navbar without reload

michaelkirk:
PTAL @charlesmchen 

(based on https://github.com/WhisperSystems/Signal-iOS/pull/2698)


michaelkirk:
In 445045b

PR #2701, opened 2017-10-25
Charlesmchen/jsq rewrite vs tests

charlesmchen:
PTAL @michaelkirk


Signal.xcodeproj/project.pbxproj
```diff
@@ -2010,7 +2009,11 @@
 			isa = PBXResourcesBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				34C6B0AE1FA0E4AA00D35993 /* test-jpg.jpg in Resources */,
```
michaelkirk: Maybe for another day, but we should bundle these in debug builds for sending fake attachments through the debug ui.

Signal/test/Models/MesssagesBubblesSizeCalculatorTest.swift
```diff
@@ -22,86 +17,94 @@ class FakeiPhone6JSQMessagesCollectionViewFlowLayout: JSQMessagesCollectionViewF
...
     }
 
     func testHeightForChineseWithEmojiBug() {
-        let messageData = self.messageDataForForText("一二三四五六七八九十甲乙丙😝戊己庚辛壬圭咖啡牛奶餅乾水果蛋糕")
-        let actual = calculator.messageBubbleSize(for: messageData, at: indexPath, with: layout)
+        let viewItem = self.viewItemForText("一二三四五六七八九十甲乙丙😝戊己庚辛壬圭咖啡牛奶餅乾水果蛋糕")
+        let actual = messageBubbleSize(for: viewItem)
         // erroneously seeing 69 with the emoji fix in place.
-        XCTAssertEqual(85.5, actual.height)
+        XCTAssertEqual(86, actual.height)
```
michaelkirk: Do these tests need to be updated now that https://github.com/WhisperSystems/Signal-iOS/commit/418fef33586718eeba1414f702a2780470a44505 is merged?

Signal/test/ViewControllers/ConversationViewItemTest.m
```diff
@@ -0,0 +1,347 @@
...
+@interface ConversationViewItemTest : XCTestCase
+
+@property (nonatomic, readonly) NSData *fakeAudioData;
+@property (nonatomic, readonly) NSData *fakeImageData;
+
+@end
+
+@implementation ConversationViewItemTest
+
+- (NSData *)fakeAudioData
```
michaelkirk: Now that we're including real assets in the test suite, can these `fake*Data` methods go away?

Signal.xcodeproj/project.pbxproj
charlesmchen: Unfortunately, I don't know of a way to do this without creating a separate target which will be a pain.

Signal/test/Models/MesssagesBubblesSizeCalculatorTest.swift
charlesmchen: Yup, done.

Signal/test/ViewControllers/ConversationViewItemTest.m
charlesmchen: Yup!

PR #2704, opened 2017-10-26
Fix edge cases around oversize test messages.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
@@ -303,15 +303,12 @@ NSString *NSStringForOWSMessageCellType(OWSMessageCellType cellType)
     if (!displayableText) {
         NSString *text = textBlock();
 
-        // Only show up to 2kb of text.
-        const NSUInteger kMaxTextDisplayLength = 2 * 1024;
-        text = [text ows_stripped];
+        // Only show up to N characters of text.
+        const NSUInteger kMaxTextDisplayLength = 1024;
         NSString *_Nullable fullText = [DisplayableText displayableText:text];
         BOOL isTextTruncated = NO;
         if (!fullText) {
             fullText = @"";
-        } else {
-            fullText = fullText;
         }
         NSString *_Nullable displayText = fullText;
         if (displayText.length > kMaxTextDisplayLength) {
@@ -324,8 +321,6 @@ NSString *NSStringForOWSMessageCellType(OWSMessageCellType cellType)
         }
         if (!displayText) {
             displayText = @"";
-        } else {
-            displayText = displayText;
         }
 
         displayableText =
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -303,15 +303,12 @@ - (DisplayableText *)displayableTextForInteractionId:(NSString *)interactionId
     if (!displayableText) {
         NSString *text = textBlock();
 
-        // Only show up to 2kb of text.
-        const NSUInteger kMaxTextDisplayLength = 2 * 1024;
-        text = [text ows_stripped];
+        // Only show up to N characters of text.
+        const NSUInteger kMaxTextDisplayLength = 1024;
```
charlesmchen: I propose we should less by default.  WDYT?

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -336,70 +331,81 @@ - (DisplayableText *)displayableTextForInteractionId:(NSString *)interactionId
...
-    OWSAssert([self.interaction isKindOfClass:[TSMessage class]]);
+    OWSAssert([self.interaction isKindOfClass:[TSOutgoingMessage class]] ||
+        [self.interaction isKindOfClass:[TSIncomingMessage class]]);
 
     if (self.hasViewState) {
         return;
     }
     self.hasViewState = YES;
 
-    TSMessage *interaction = (TSMessage *)self.interaction;
```
charlesmchen: It turns out the `TSMessage.body` can be non-nil (Empty string: `@""`) for incoming attachment messages.  So we need to check for an attachment first.

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -30,6 +30,13 @@ class MessageDetailViewController: OWSViewController {
 
     var mediaMessageView: MediaMessageView?
 
+    // See comments on updateTextLayout.
```
charlesmchen: I'd start with the comments on `updateTextLayout` because this is non-trivial.

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -567,4 +589,93 @@ class MessageDetailViewController: OWSViewController {
...
+    //
+    // Therefore we're doing something unusual.  
+    //
+    // * We use an empty UIView "messageTextProxyView" as a placeholder for the
+    //   the UITextView.  It has the size and position of where the UITextView
+    //   would be normally.
+    // * We use a UITextView inside that proxy that is just large enough to
+    //   render the content onscreen. We then move it around within the proxy
+    //   bounds to render the parts of the proxy which are onscreen.
+    private func updateTextLayout() {
```
charlesmchen: I've tested and this works well on iPhone 5C and can handle very large oversize text messages.  It's pretty smooth visually, too. I was afraid there might be some rendering lag.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
michaelkirk: :+1:

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -336,70 +331,81 @@ - (DisplayableText *)displayableTextForInteractionId:(NSString *)interactionId
...
+- (nullable TSAttachment *)firstAttachmentIfAnyOfMessage:(TSMessage *)message
+{
+    if (message.attachmentIds.count == 0) {
+        return nil;
+    }
+    NSString *_Nullable attachmentId = message.attachmentIds.firstObject;
+    if (attachmentId.length == 0) {
+        return nil;
+    }
+    return [TSAttachment fetchObjectWithUniqueID:attachmentId];
```
michaelkirk: This is kind of a sneaky spot to create a transaction. 

And more generally, since the ViewItem is driving UI code, it seems like we should be doing any reads from a transaction parameter passed in from a longLivedRead db connection which hangs on the view controller (e.g. `uiDatabaseConnection`)

This problem isn't new to this PR, but it just occurred to me.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -336,70 +331,81 @@ - (DisplayableText *)displayableTextForInteractionId:(NSString *)interactionId
...
+}
+
 - (void)ensureViewState
 {
-    OWSAssert([self.interaction isKindOfClass:[TSMessage class]]);
+    OWSAssert([self.interaction isKindOfClass:[TSOutgoingMessage class]] ||
+        [self.interaction isKindOfClass:[TSIncomingMessage class]]);
```
michaelkirk: Is this significant? 😱

Signal/src/ViewControllers/MessageDetailViewController.swift
michaelkirk: This is wild. At what file size does this become relevant?

charlesmchen: The issue shows up quite easily if you paste oversize text.  Neither `UILabel` and `UITextView` are designed to be used inside a `UIScrollView` to render long strings.

charlesmchen: Incidentally, I didn't come up with this solution, I've seen it used before to work around other forms of the `UIScrollview`-inside-a-`UIScrollview` issue.

charlesmchen:
> My main concern is about driving the UI on a dbConnection which is not long lived, but that's not new to this PR. Feel free to merge and make a new card.

Card made.

> Also, since this proxy-view code is somewhat clever, it's worth considering only activating it when we're looking at payloads large enough where the simple solution doesn't work. But we could always do that later if it proves problematic.

I think we should always do it:

* This isn't a rare edge case, it's easy to hit this rendering issue with long text.
* There's no safe way to predict what length of text of will trigger this issue.
* If there are any issues with this rendering technique, it'd be nice to know about them sooner rather than later. Using it always will help flush them out.
* Although it takes a minute to understand, I hope this solution is safe.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
charlesmchen: Not really, I just realized that we might someday accidentally use an `OWSMessageCell` for an info or error message.  This assert will quickly expose that class of bug.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
charlesmchen: Good catch - card made.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
michaelkirk: 👍

PR #2708, opened 2017-10-26
Charlesmchen/streamline relayout

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2930,9 +2930,7 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
             OWSAssert([NSThread isMainThread]);
 
             if (!success) {
-                [self resetContentAndLayout];
-            } else {
-                [self.collectionView.collectionViewLayout invalidateLayout];
+                DDLogInfo(@"%@ performBatchUpdates did not finish", self.tag);
             }
 
             [self updateLastVisibleTimestamp];
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2926,10 +2926,10 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
         }
         [self.collectionView reloadItemsAtIndexPaths:rowsToReload];
     }
-        completion:^(BOOL success) {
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2930,9 +2930,7 @@ - (void)yapDatabaseModified:(NSNotification *)notification
             OWSAssert([NSThread isMainThread]);
 
             if (!success) {
-                [self resetContentAndLayout];
-            } else {
-                [self.collectionView.collectionViewLayout invalidateLayout];
```
charlesmchen: This work wasn't necessary - and was interfering with the animations.

michaelkirk: I thought `!success` implied that some of our cells may not be correct, hence starting from scratch. If not that, what **is** it for?

Though I don't know why we were ever invalidating layout in the *else* (success) branch. That seems unnecessary. Happy to see that go.

charlesmchen: I spent an hour this afternoon trying to answer exactly this question, unsuccessfully.  These "failures to finish" occur regularly in the normal operation of our code. 
 I'm not quite sure but it seems like they happen when "database modified" events happen in quick succession and animations for one haven't completed before the next occurs. 
 Regardless, `performBatchUpdates` seems to handle this eventuality itself and to reload cells and/or invalidate the layout as necessary.  Doing so ourselves doesn't just seem to be redundant - it interferes with `performBatchUpdates`'s animations.  In my testing the update animations are noticeably smoother having removed these lines.

charlesmchen: I omitted an important observation: `UICollectionView.h` labels this parameter `finished`:
```
- (void)performBatchUpdates:(void (^ __nullable)(void))updates completion:(void (^ __nullable)(BOOL finished))completion; // allows multiple insert/delete/reload/move calls to be animated simultaneously. Nestable.
```
Our codebase mislabels it `success`.  My working assumption now is that this parameter functions like the `finished` parameter of the completion blocks in the `[UIView animateWithDuration:...]` variants, e.g. it indicates whether or not the requested animations were interrupted before they could complete.  After all `performBatchUpdates` is all about requesting animations.

michaelkirk: OK, let's see how it goes. LGTM

PR #2710, opened 2017-10-27
Attachments require digest or show explanatory error.

michaelkirk:
When checksum is not present we show a somewhat helpful error.

PTAL @charlesmchen 

Note this is for the 2.18.0 release.

## iPhone5 sized

screen shot 2017-10-26 at 5 32 20 pm

screen shot 2017-10-26 at 5 32 13 pm

## 6-sized

screen shot 2017-10-26 at 5 23 41 pm

screen shot 2017-10-26 at 5 23 26 pm

## Plus Sized

screen shot 2017-10-26 at 5 34 59 pm

screen shot 2017-10-26 at 5 34 55 pm

### DIFF
--- a/Signal/translations/en.lproj/Localizable.strings
+++ b/Signal/translations/en.lproj/Localizable.strings
@@ -547,6 +547,9 @@
 /* Error message when attempting to send message */
 "ERROR_DESCRIPTION_UNREGISTERED_RECIPIENT" = "Contact is not a Signal user.";
 
+/* Error message when unable to receive an attachment because the sending client is too old. */
+"ERROR_MESSAGE_ATTACHMENT_FROM_OLD_CLIENT" = "Failure: Ask sender to update Signal and resend.";
+
 /* No comment provided by engineer. */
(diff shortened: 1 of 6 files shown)


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -547,6 +547,9 @@
 /* Error message when attempting to send message */
 "ERROR_DESCRIPTION_UNREGISTERED_RECIPIENT" = "Contact is not a Signal user.";
 
+/* Error message when unable to receive an attachment because the sending client is too old. */
+"ERROR_MESSAGE_ATTACHMENT_FROM_OLD_CLIENT" = "Failure: Ask sender to update Signal and resend.";
```
charlesmchen: Will this always be two lines when translated?

michaelkirk: I don't know, but the layout looks ok with only one line, and it resizes text so longer text will work too - though it gets pretty small. 

This should be a rarely seen piece of UI, so making it super flexible doesn't warrant much priority IMO.

michaelkirk:
In 81386eae0cbb02618530fe8e3796b499af353a70

PR #2711, opened 2017-10-27
Jumbomoji.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
screen shot 2017-10-27 at 10 25 08 am

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -182,6 +180,23 @@ + (NSString *)cellReuseIdentifier
...
+    if (self.displayableText.jumbomojiCount) {
+        switch (self.displayableText.jumbomojiCount.integerValue) {
+            case 1:
+                return [UIFont ows_regularFontWithSize:35.f];
+            case 2:
+                return [UIFont ows_regularFontWithSize:30.f];
+            case 3:
+            case 4:
+            case 5:
+                return [UIFont ows_regularFontWithSize:25.f];
```
charlesmchen: WDYT about these values? See screenshot.

Signal/src/util/DisplayableText.swift
```diff
@@ -11,13 +11,83 @@ import Foundation
     let fullText: String
     let displayText: String
     let isTextTruncated: Bool
+    let jumbomojiCount: NSNumber?
```
charlesmchen: Is there a better way to do an objective-c compatible optional number in Swift?

Signal/src/util/DisplayableText.swift
```diff
@@ -11,13 +11,83 @@ import Foundation
...
+    // * Contains <= maxEmojiCount emoji
+    //
+    // ...return the number of emoji in the string.  Otherwise return nil.
+    //
+    // On iOS 9 and earler, always returns nil.
+    @objc public class func jumbomojiCount(to string: String) -> NSNumber? {
+        if string == "" {
+            return nil
+        }
+        if string.characters.count > Int(kMaxJumbomojiCount) {
```
charlesmchen: I made a few refinements to reduce perf cost, like this early exit.

Signal/src/util/DisplayableText.swift
```diff
@@ -11,13 +11,83 @@ import Foundation
...
+                                                    return
+                                                }
+                                                guard rangeFont.fontName == ".AppleColorEmojiUI" else {
+                                                    didFail = true
+                                                    stop.pointee = true
+                                                    return
+                                                }
+                                                if rangeFont.fontName == ".AppleColorEmojiUI" {
+                                                    Logger.verbose("Detected Emoji at location: \(range.location), for length: \(range.length)")
+                                                    emojiCount += UInt(range.length)
```
charlesmchen: Critically, this code assumes that one character == one emoji.  I haven't found an exception yet, but I suspect there are.  To be continued.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
michaelkirk: Seems about right!

Signal/src/util/DisplayableText.swift
michaelkirk: > Critically, this code assumes that one character == one emoji. I haven't found an exception yet, but I suspect there are. To be continued.

In swift String.characters represent grapheme-clusters, so this should "just work", which is nice.

> I made a few refinements to reduce perf cost, like this early exit.

The downside is , because grapheme clusters can be any length, this requires parsing the string to know how many characters it composes. Said differently, the perf of string.characters.count is related to the length of string, and potentially a bit expensive if the string is very long.

Something to consider if this proves to be a bottleneck is, instead of getting the whole count, to bail if we can enumerate beyond kMaxJumbomojiCount.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -323,8 +323,12 @@ - (DisplayableText *)displayableTextForInteractionId:(NSString *)interactionId
             displayText = @"";
         }
 
-        displayableText =
-            [[DisplayableText alloc] initWithFullText:fullText displayText:displayText isTextTruncated:isTextTruncated];
+        NSNumber *_Nullable jumbomojiCount = [DisplayableText jumbomojiCountTo:fullText];
```
michaelkirk: Are we using this class method elsewhere? 
Do we ever init a DisplayableText without a jumbomojiCount?

nit: If the answer to both is "no" then I'd argue this belongs in the DisplayableText inititializer.

Signal/src/util/DisplayableText.swift
michaelkirk: Since Objc doesn't have optional primitive types, this seems like the right way to do that.

But in looking how we're using it, does it even benefit from being optional? 

In all the cases we're returning nil, it seems like zero is actually the right semantics. e.g. There are 0 jumbo-emoji in the empty string. There are 0 jumbo-emoji in a really long string. etc.

charlesmchen: Great point.

Signal/src/util/DisplayableText.swift
charlesmchen: Surely `[NSMutableAttributedString fixAttributes:]` is much more expensive than `characters.count`, since it processes the whole string.  That's the work I'm trying to avoid here.  Having said that, `fixAttributes:` takes a range - perhaps we could skip the string length check and pass a shorter range to `fixAttributes:`.  That feels like over-optimizing though at the cost of readability.  

Bear in mind that we're now doing a much better job avoiding this work.

* `ConversationViewItem` does this work lazily.
* `ConversationViewItem` never does this work twice, per instance of `ConversationViewController`, since the result is expensive to compute and very cheap to cache.  

Therefore I really doubt this is a hotspot, so long as we do the string length check.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
charlesmchen: Done.

charlesmchen:
PTAL @michaelkirk I've reworked this PR significantly.

Scripts/emoji_ranges.py
```diff
@@ -0,0 +1,84 @@
...
+        (0x2700,0x27BF,   ),
+        (0xFE00,0xFE0F,   ),
+        (0x1F900,0x1F9FF,  ),
+        (65024,65039,), 
+        (8400, 8447,)
+        ]
+
+# def is_in_big_range(range_start, range_end)
+
+if __name__ == '__main__':
```
charlesmchen: This script converts the emoji ranges from the unicode spec to more usable ranges by combining adjacent ranges.

Signal/src/util/DisplayableText.swift
```diff
@@ -4,20 +4,330 @@
...
+            self.rangeStart = rangeStart
+            self.rangeEnd = rangeEnd
+        }
+    }
+
+    // From:
+    // https://www.unicode.org/Public/emoji/
+    // Current Version:
+    // https://www.unicode.org/Public/emoji/6.0/emoji-data.txt
+    static let kEmojiRanges = [
```
charlesmchen: To do binary search, the ranges must be sorted in ascending order.

charlesmchen:
Wait, I'm going to add a unit test.

charlesmchen:
@michaelkirk unit test added (and bugs caught and fixed)!

PTAL

Signal/src/util/DisplayableText.swift
charlesmchen: The old emoji detection solution didn't properly count multi-character emoji.  I've decided to be standards-compliant and just do binary search on the ranges defined in the latest emoji standard.  Should be more accurate, precise and performant.

Signal/test/util/DisplayableTextFilterTest.swift
```diff
@@ -17,24 +17,84 @@ class DisplayableTextTest: XCTestCase {
...
+        XCTAssertEqual(diacriticalText, DisplayableText.displayableText(diacriticalText))
 
         // filter excessive diacritics
-        XCTAssertEqual("HAVING TROUBLE READING TEXT?", filter.displayableText("H҉̸̧͘͠A͢͞V̛̛I̴̸N͏̕͏G҉̵͜͏͢ ̧̧́T̶̛͘͡R̸̵̨̢̀O̷̡U͡҉B̶̛͢͞L̸̸͘͢͟É̸ ̸̛͘͏R͟È͠͞A̸͝Ḑ̕͘͜I̵͘҉͜͞N̷̡̢͠G̴͘͠ ͟͞T͏̢́͡È̀X̕҉̢̀T̢͠?̕͏̢͘͢") )
+        XCTAssertEqual("HAVING TROUBLE READING TEXT?", DisplayableText.displayableText("H҉̸̧͘͠A͢͞V̛̛I̴̸N͏̕͏G҉̵͜͏͢ ̧̧́T̶̛͘͡R̸̵̨̢̀O̷̡U͡҉B̶̛͢͞L̸̸͘͢͟É̸ ̸̛͘͏R͟È͠͞A̸͝Ḑ̕͘͜I̵͘҉͜͞N̷̡̢͠G̴͘͠ ͟͞T͏̢́͡È̀X̕҉̢̀T̢͠?̕͏̢͘͢") )
+
+        XCTAssertEqual("LGO!", DisplayableText.displayableText("L̷̳͔̲͝Ģ̵̮̯̤̩̙͍̬̟͉̹̘̹͍͈̮̦̰̣͟͝O̶̴̮̻̮̗͘͡!̴̷̟͓͓"))
+    }
+
+    func testGlyphCount() {
```
michaelkirk: 👍🏾

Signal/src/util/DisplayableText.swift
michaelkirk: I like this new approach. Can you cross reference the script which generates these ranges for the future peeps?

Scripts/emoji_ranges.py
```diff
@@ -0,0 +1,84 @@
...
+        (0x1F680,0x1F6FF, ),
+        (0x2600,0x26FF,   ),
+        (0x2700,0x27BF,   ),
+        (0xFE00,0xFE0F,   ),
+        (0x1F900,0x1F9FF,  ),
+        (65024,65039,), 
+        (8400, 8447,)
+        ]
+
+# def is_in_big_range(range_start, range_end)
```
michaelkirk: remove?

Scripts/emoji_ranges.py
```diff
@@ -0,0 +1,84 @@
...
+            if range_start >= last_range_start and range_start <= last_range_end + 1:
+            # if last_range_end + 1 == range_start:
+                new_ranges = new_ranges[:-1]
+                print 'merging', last_range_start, last_range_end, 'and', range_start, range_end
+                new_ranges.append((last_range_start, max(range_end, last_range_end),))
+                continue
+
+        new_ranges.append((range_start, range_end,))
+    
+    print
```
michaelkirk: Is this empty print intentional?

Scripts/emoji_ranges.py
```diff
@@ -0,0 +1,84 @@
...
+    for range_start, range_end in new_ranges:
+        # print '0x%X...0x%X, // %d Emotions' % (range_start, range_end, (1 + range_end - range_start), )
+        print 'EmojiRange(rangeStart:0x%X, rangeEnd:0x%X),' % (range_start, range_end, )
+    print 'new_ranges:', len(new_ranges)
```
michaelkirk: So to use this script do you copy some of this output into the DisplayableText class? Can you leave some usage instructions?

Signal/src/util/DisplayableText.swift
charlesmchen: Done.

Scripts/emoji_ranges.py
charlesmchen: Done.

Scripts/emoji_ranges.py
charlesmchen: Yeah it makes the output more readable.

Scripts/emoji_ranges.py
charlesmchen: Done.

PR #2712, opened 2017-10-27
Charlesmchen/split gifs

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
+++ b/Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
@@ -147,6 +147,9 @@ class GifPickerViewController: OWSViewController, UISearchBarDelegate, UICollect
     private func createViews() {
 
         view.backgroundColor = UIColor.white
+
+        // Block UIKit from adjust insets of collection view which screws up
+        // min/max scroll positions.
         self.automaticallyAdjustsScrollViewInsets = false
 
         // Search
--- a/Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
+++ b/Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
@@ -127,6 +127,11 @@ class GifPickerViewController: OWSViewController, UISearchBarDelegate, UICollect
                                                selector: #selector(didBecomeActive),
                                                name: NSNotification.Name.UIApplicationDidBecomeActive,
                                                object: nil)
+
+        DispatchQueue.main.async {
+            self.searchBar.text = "hi"
+            self.searchBar(self.searchBar, textDidChange: self.searchBar.text!)
+        }
     }
 
     override func viewDidAppear(_ animated: Bool) {
--- a/Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
+++ b/Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
@@ -127,11 +127,6 @@ class GifPickerViewController: OWSViewController, UISearchBarDelegate, UICollect
                                                selector: #selector(didBecomeActive),
                                                name: NSNotification.Name.UIApplicationDidBecomeActive,
                                                object: nil)
-
-        DispatchQueue.main.async {
-            self.searchBar.text = "hi"
-            self.searchBar(self.searchBar, textDidChange: self.searchBar.text!)
-        }
     }
 
     override func viewDidAppear(_ animated: Bool) {
--- a/Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
+++ b/Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
@@ -152,6 +152,7 @@ class GifPickerViewController: OWSViewController, UISearchBarDelegate, UICollect
     private func createViews() {
 
         view.backgroundColor = UIColor.white
+        self.automaticallyAdjustsScrollViewInsets = false
 
         // Search
         searchBar.searchBarStyle = .minimal
--- a/Signal/src/network/GiphyDownloader.swift
+++ b/Signal/src/network/GiphyDownloader.swift
@@ -33,11 +33,15 @@ class GiphyAssetSegment {
             AssertIsOnMainThread()
         }
     }
+
     // This state is accessed off the main thread.
     //
     // * During downloads it will be accessed on the task delegate queue.
     // * After downloads it will be accessed on a worker queue. 
     private var segmentData = Data()
+
+    // This state should only be accessed on the main thread.
+    public weak var task: URLSessionDataTask?
 
     init(index: UInt,
          segmentStart: UInt,
@@ -131,6 +135,7 @@ enum GiphyAssetRequestState: UInt {
             createSegments()
         }
     }
+    public weak var contentLengthTask: URLSessionDataTask?
 
     init(rendition: GiphyRendition,
          priority: GiphyRequestPriority,
@@ -294,6 +299,12 @@ enum GiphyAssetRequestState: UInt {
         AssertIsOnMainThread()
 
         wasCancelled = true
+        contentLengthTask?.cancel()
+        contentLengthTask = nil
+        for segment in segments {
+            segment.task?.cancel()
+            segment.task = nil
+        }
 
         // Don't call the callbacks if the request is cancelled.
         clearCallbacks()
--- a/Signal/src/network/GiphyDownloader.swift
+++ b/Signal/src/network/GiphyDownloader.swift
@@ -23,14 +23,21 @@ class GiphyAssetSegment {
     public let index: UInt
     public let segmentStart: UInt
     public let segmentLength: UInt
+    // The amount of the segment that is overlap.  
+    // The overlap lies in the _first_ n bytes of the segment data.
     public let redundantLength: UInt
 
+    // This state should only be accessed on the main thread.
     public var state: GiphyAssetSegmentState = .waiting {
         didSet {
             AssertIsOnMainThread()
         }
     }
-    private var datas = [Data]()
+    // This state is accessed off the main thread.
+    //
+    // * During downloads it will be accessed on the task delegate queue.
+    // * After downloads it will be accessed on a worker queue. 
+    private var segmentData = NSMutableData()
 
     init(index: UInt,
          segmentStart: UInt,
@@ -43,11 +50,7 @@ class GiphyAssetSegment {
     }
 
     public func totalDataSize() -> UInt {
-        var result: UInt = 0
-        for data in datas {
-            result += UInt(data.count)
-        }
-        return result
+        return UInt(segmentData.length)
     }
 
     public func append(data: Data) {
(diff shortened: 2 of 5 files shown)


Signal/src/network/GiphyDownloader.swift
```diff
@@ -227,93 +434,220 @@ extension URLSessionTask {
...
+
+                var request = URLRequest(url: assetRequest.rendition.url as URL)
+                request.httpMethod = "HEAD"
+
+                let task = downloadSession.dataTask(with:request, completionHandler: { [weak self] _, response, error -> Void in
+                    self?.handleAssetSizeResponse(assetRequest:assetRequest, response:response, error:error)
+                })
+
+                task.resume()
+                return
```
michaelkirk: Do we need to kick the run loop here?

Else won't we potentially leave the queue hanging or less-than-full?

And same question for the end of this method?

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
```diff
@@ -147,6 +147,7 @@ class GifPickerViewController: OWSViewController, UISearchBarDelegate, UICollect
     private func createViews() {
 
         view.backgroundColor = UIColor.white
+        self.automaticallyAdjustsScrollViewInsets = false
```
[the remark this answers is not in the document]
charlesmchen: `UIViewController`'s default behavior is to sometimes manipulate the insets of any scroll view that is the root view or a direct descendent thereof - the intent is to get scrolling content to layout out properly vis-a-vis navigation headers etc.  The problem is that UIKit doesn't always guess right about when you want this behavior or not.  In our case, it was breaking scroll behavior in this view.  Trying scrolling up past the top of the GIF content.

Signal/src/network/GiphyDownloader.swift
```diff
@@ -10,11 +10,79 @@ enum GiphyRequestPriority {
...
+    }
+
+    public func mergeData(assetData: NSMutableData) {
+        // In some cases the last two segments will overlap.
+        // In that case, we only want to append the non-overlapping
+        // tail of the last segment.
+        var bytesToIgnore = Int(redundantLength)
+        for data in datas {
+            if data.count <= bytesToIgnore {
+                bytesToIgnore -= data.count
```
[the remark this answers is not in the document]
charlesmchen: I was being clever and trying to avoid doing redundant work:

* merging segment data
* merging asset data.

To avoid the first merge, I was storing a list of segment data pieces.  The code's a lot simpler if we just do two merges though so I've done that.

Signal/src/network/GiphyDownloader.swift
```diff
@@ -28,15 +96,137 @@ enum GiphyRequestPriority {
...
+
+    private func segmentSize() -> UInt {
+        let fileSize = UInt(contentLength)
+        guard fileSize > 0 else {
+            owsFail("\(TAG) rendition missing filesize")
+            requestDidFail()
+            return 0
+        }
+
+        let k1MB: UInt = 1024 * 1024
```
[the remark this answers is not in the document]
charlesmchen: Great question - I'll ask Moxie.

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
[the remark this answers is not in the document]
michaelkirk: :+1: I've actually seen that once before and forgot. Thanks for addressing it. Can you include a comment?

Signal/src/network/GiphyDownloader.swift
```diff
@@ -136,41 +163,46 @@ enum GiphyAssetRequestState: UInt {
...
+        while nextSegmentStart < contentLength {
             var segmentStart: UInt = nextSegmentStart
             var redundantLength: UInt = 0
             // The last segment may overlap the penultimate segment
             // in order to keep the segment sizes uniform.
-            if segmentStart + segmentLength > fileSize {
-                redundantLength = segmentStart + segmentLength - fileSize
-                segmentStart = fileSize - segmentLength
+            if segmentStart + segmentLength > contentLength {
+                redundantLength = segmentStart + segmentLength - contentLength
```
michaelkirk: nit: The logic currently works out, but it seems counterintuitive that `redundantLength` will be negative for all but the last segments. I'd expect it to be 0 in those cases.

Signal/src/network/GiphyDownloader.swift
```diff
@@ -522,7 +585,7 @@ extension URLSessionTask {
             }
             guard !assetRequest.wasCancelled else {
                 // Discard the cancelled asset request and try again.
-                self.processRequestQueue()
+                self.removeAssetRequestFromQueue(assetRequest: assetRequest)
```
michaelkirk: Don't we still want to `processRequestQueue()` so we can try the next one?

Signal/src/network/GiphyDownloader.swift
```diff
@@ -557,34 +620,46 @@ extension URLSessionTask {
                 var request = URLRequest(url: assetRequest.rendition.url as URL)
                 request.httpMethod = "HEAD"
 
-                let task = downloadSession.dataTask(with:request, completionHandler: { [weak self] _, response, error -> Void in
+                let task = downloadSession.dataTask(with:request, completionHandler: { [weak self] data, response, error -> Void in
+                    if let data = data {
```
michaelkirk: style nit: combine these clauses `if let data = data, data.count > 0 {`

Signal/src/network/GiphyDownloader.swift
```diff
@@ -617,19 +692,31 @@ extension URLSessionTask {
...
             for assetRequest in assetRequestQueue where assetRequest.priority == priority {
                 switch assetRequest.state {
                 case .waiting:
-                    break
+                    // This asset request needs its content length.
+                    return assetRequest
                 case .requestingSize:
                     activeAssetRequestsCount += 1
+                    // Ensure that only N requests are active at a time.
+                    guard activeAssetRequestsCount < kMaxAssetRequestCount else {
```
michaelkirk: Naming suggests this should be `<=`

Signal/src/network/GiphyDownloader.swift
```diff
@@ -639,13 +726,18 @@ extension URLSessionTask {
...
-                    activeAssetRequestsCount += 1
-                    // Ensure that only N requests are active at a time.
-                    guard activeAssetRequestsCount < self.kMaxAssetRequestCount else {
-                        return nil
-                    }
-
+                let downloadingSegmentsCount = assetRequest.downloadingSegmentsCount()
+                activeAssetRequestsCount += downloadingSegmentsCount
+                // Ensure that only N segment requests are active per asset at a time.
+                guard downloadingSegmentsCount < kMaxAssetRequestsPerAssetCount else {
```
michaelkirk: Naming suggests this should be `<=`

Signal/src/network/GiphyDownloader.swift
michaelkirk: It seems like this guard belongs in the `case .waiting` instead of here.

I think we want to do what we can to ensure something of a FIFO priority with our GiphyAssets. And requesting content size is a way to insert network requests for Asset which we haven't "started" yet.

Signal/src/network/GiphyDownloader.swift
```diff
@@ -639,13 +726,18 @@ extension URLSessionTask {
...
+                    continue
+                }
+                // Ensure that only N requests are active at a time.
+                guard activeAssetRequestsCount < kMaxAssetRequestCount else {
```
michaelkirk: Naming suggests this should be `<=`

Signal/src/network/GiphyDownloader.swift
```diff
@@ -46,15 +51,27 @@ class GiphyAssetSegment {
...
+        guard state == .downloading else {
+            owsFail("\(TAG) appending data in invalid state: \(state)")
+            return
+        }
+
         datas.append(data)
     }
 
     public func mergeData(assetData: NSMutableData) {
+        guard state == .complete else {
```
michaelkirk: Not functionally wrong, but why are we using `NSMutableData`?  I'd expect something like:

```
mergeData(assetData: inout Data) {
...
assetData.append(data.subdata(in: start..<end))
...
}

var assetData = Data()
mergeData(assetData: &assetData)
```

Signal/src/network/GiphyDownloader.swift
```diff
@@ -46,15 +51,27 @@ class GiphyAssetSegment {
     }
 
     public func append(data: Data) {
+        guard state == .downloading else {
```
michaelkirk: Why does an AssetSegment have multiple datas? 

I expected just one, represented by it's range (whereas the GiphyAsset has multiple data's via it's multiple AssetSegments.)

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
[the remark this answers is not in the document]
charlesmchen: Done.

Signal/src/network/GiphyDownloader.swift
```diff
@@ -227,93 +434,220 @@ extension URLSessionTask {
...
+        }
+
+        assetRequestQueue = assetRequestQueue.filter { $0 != assetRequest }
+    }
+
+    // Start a request if necessary, complete asset requests if possible.
+    private func processRequestQueue() {
         AssertIsOnMainThread()
 
         DispatchQueue.main.async {
```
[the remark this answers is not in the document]
charlesmchen: I've reworked this a bit since we only need to do this dispatch async `requestAsset()` and added a comment.

Signal/src/network/GiphyDownloader.swift
charlesmchen: Already fixed!

Signal/src/network/GiphyDownloader.swift
charlesmchen: Sorry I got interrupted by an interview.  I've just pushed a commit that unifies this.

Signal/src/network/GiphyDownloader.swift
charlesmchen: Done.

Signal/src/network/GiphyDownloader.swift
charlesmchen: Already reworked.  See the new arrangement.

Signal/src/network/GiphyDownloader.swift
charlesmchen: `removeAssetRequestFromQueue` calls `processRequestQueue`

Signal/src/network/GiphyDownloader.swift
charlesmchen: Nice.

Signal/src/network/GiphyDownloader.swift
charlesmchen: > Naming suggests this should be <=

We're try to find an asset for which we can add a request.  If we're at MAX, we don't want to add another request and can exit early.

charlesmchen: > It seems like this guard belongs in the case .waiting instead of here.

We can abandon our search for a request to start as soon as we hit the limit.  We want this method to be cheap since we call it often.

> I think we want to do what we can to ensure something of a FIFO priority with our GiphyAssets. 

We do honor FIFO, within the limitation that we have an upper bound on the number of requests per asset.  Are you saying we shouldn't do that?  

> And requesting content size is a way to insert network requests for Asset which we haven't "started" yet.

Right under the current arrangement we can have requests for multiple assets at a time.  For example, when the picker opens, it'll immediately do HEAD on the first six assets.  When the first HEAD completes, we'll start a segment download.  etc.

If you think we need a different arrangement let's video chat.

Signal/src/network/GiphyDownloader.swift
charlesmchen: See above, we want to enforce the limit and not exceed it.

Signal/src/network/GiphyDownloader.swift
charlesmchen: Ditto.

charlesmchen:
PTAL @michaelkirk I've just pushed yet another commit.

> In my testing (I'm on a crappy network) I'm seemingly able to get the GifPicker into a state where it stops making progress. 

Fixed, we weren't handling all cancellations correctly.

> See my concerns about missing some recursion calls in processRequestQueue

Fixed.

> the request queue can get pretty backlogged (e.g. I'm seeing hundreds of requests when I quickly scroll to the bottom).

Queue entries (the asset requests) should be cheap and stale ones should be quickly cleared.  

> we don't actually cancel the dataTask until the next time the NSURLSession delegate checks in, which could be a while on a slow network.

I've modified the cancellation logic to cancel the tasks sooner, but I'm not sure if that'll have any impact.

Signal/src/network/GiphyDownloader.swift
```diff
@@ -77,11 +81,11 @@ class GiphyAssetSegment {
         // tail of the segment data.
         let bytesToIgnore = Int(redundantLength)
         if bytesToIgnore > 0 {
-            let range = NSMakeRange(bytesToIgnore, segmentData.length - bytesToIgnore)
-            let subdata = (segmentData as NSData).subdata(with: range)
+            let range = NSMakeRange(bytesToIgnore, segmentData.count - bytesToIgnore)
+            let subdata = segmentData.subdata(in: range.location..<range.location + range.length)
```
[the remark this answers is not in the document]
charlesmchen: In the next PR!

PR #2714, opened 2017-10-31
Mkirk 2.18.0/call audio

michaelkirk:
This is kind of an odds-and-ends PR.

- Fixed an assert that can get triggered if hanging up a call right after starting it
- Adjusted volume:  out-ring is louder, sonar-ping is quieter
- Fixed change in ringer volume when toggling between video and speakerphone

I changed some plumbing around call "holding" though we don't support this feature yet, I've paved part of the way as I touched the surrounding code.

This could go in the 2.18.0 or 2.19.0 release.

PTAL @charlesmchen


Signal/src/call/CallAudioService.swift
```diff
@@ -188,7 +205,10 @@ struct AudioSource: Hashable {
             // SoloAmbient plays through speaker, but respects silent switch
             setAudioSession(category: AVAudioSessionCategorySoloAmbient,
                             mode: AVAudioSessionModeDefault)
-        } else if call.hasLocalVideo {
+        } else if call.state == .connected, call.hasLocalVideo {
```
charlesmchen: so the comment here functions like &&?

charlesmchen: comma!

Signal/src/call/CallAudioService.swift
```diff
@@ -216,14 +236,25 @@ struct AudioSource: Hashable {
...
             }
         } catch {
             owsFail("\(TAG) failed setting audio source with error: \(error) isSpeakerPhoneEnabled: \(call.isSpeakerphoneEnabled)")
         }
+
+        if call.state == .connected, !call.isOnHold {
+            audioSession.isRTCAudioEnabled = true
+        } else {
+            audioSession.isRTCAudioEnabled = false
+        }
```
charlesmchen: Would this be simpler?
```
audioSession.isRTCAudioEnabled = call.state == .connected && !call.isOnHold

File: Signal/src/call/CallService.swift [hunk 3]
```diff
@@ -926,7 +926,7 @@ protocol CallServiceObserver: class {
         call.state = .connected
 
         // We don't risk transmitting any media until the remote client has admitted to being connected.
-        peerConnectionClient.setAudioEnabled(enabled: !call.isMuted)
+        ensureAudioState(call: call, peerConnectinClient: peerConnectionClient)
```
charlesmchen: s/peerConnectinClient/peerConnectionClient/

Signal/src/call/CallAudioService.swift
michaelkirk: Yup!

Signal/src/call/CallAudioService.swift
michaelkirk: Just a bit too dense for my arbitrary tastes.

Signal/src/call/CallService.swift
michaelkirk: fixed!

PR #2715, opened 2017-11-01
Charlesmchen/splitgifs2

charlesmchen:
PTAL @michaelkirk


Signal/src/network/GiphyDownloader.swift
```diff
@@ -587,13 +582,21 @@ extension URLSessionTask {
         }
 
         assetRequestQueue = assetRequestQueue.filter { $0 != assetRequest }
-        processRequestQueue()
+        // Process the queue async to ensure that state in the downloader
+        // classes is consistent before we try to start a new request.
+        processRequestQueueAsync()
```
michaelkirk: This is the significant change, right?

charlesmchen: Yeah

PR #2716, opened 2017-11-01
Don't animate message sends.

charlesmchen:
PTAL @michaelkirk 

I might go a step farther and skip animations related to status changes of new outgoing messages, but that won't be necessary if we adopt a more Android-like conversation view style with footers on every message.

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2820,14 +2820,14 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
         return;
     }
 
-    NSArray *messageRowChanges = nil;
-    NSArray *sectionChanges = nil;
+    NSArray<YapDatabaseViewSectionChange *> *sectionChanges = nil;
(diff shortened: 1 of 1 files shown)


michaelkirk:
> I might go a step farther and skip animations related to status changes of new outgoing messages, but that won't be necessary if we adopt a more Android-like conversation view style with footers on every message.

I thought the entirety of the problem was the lack of a footer for every message. In my testing, the glitchy animation before this branch is indistinguishable from the glitchy animation after this branch.

Can you help me understand what this PR accomplishes?

charlesmchen:
Yes, there are remaining imperfections in the animations but they're getting better.

Yes, having a footer on every message will simplify the transitions and thereby eliminate some difficult animations.

The goal of this PR isn't to improve the animations - it's to match Signal Android's behavior of skipping animations when inserting new outgoing (just sent) messages into conversation view so that message sending feels more snappy and instantaneous.

charlesmchen:
We now explicitly suppress animations.

PTAL @michaelkirk

charlesmchen:
Okay I went a step further and suppressed animations for the status changes occur for just-sent messages.

charlesmchen:
It's really snappy now!

PR #2717, opened 2017-11-01
Let users with external keyboards send messages using modifier-return…

charlesmchen:
… (shift, command, option).

PTAL @michaelkirk


charlesmchen:
> I'd prefer that shift+enter continue to insert a newline on Signal-iOS instead of sending, but otherwise LGTM.

Done.  

It was a tough choice.  IMO shift-enter is an a lot more ergonomic/easier keystroke than option-return or command-return, and this will be a very repetitive keystroke for users who use an external keyboard.  But this is a big step forward.  I'll contact users who use external keyboards and see what they think.

PR #2718, opened 2017-11-01
Log voice memo durations.

charlesmchen:
Voice memo file sizes seem reasonable.  A little more than 1 MB/minute.

PTAL @michaelkirk


charlesmchen:
The obvious low-hanging fruit would be to go mono - but I recommend we not do that.  My understanding is that stereo voice recordings have significant accessibility benefits for the visually-impaired.

michaelkirk:
LGTM

PR #2719, opened 2017-11-01
Charlesmchen/digits aint jumbomoji

charlesmchen:
PTAL @michaelkirk


Signal/src/util/DisplayableText.swift
```diff
@@ -30,7 +30,8 @@ extension UnicodeScalar {
     static let kEmojiRanges = [
         EmojiRange(rangeStart:0x23, rangeEnd:0x23),
         EmojiRange(rangeStart:0x2A, rangeEnd:0x2A),
```
michaelkirk: LGTM

I'm somewhat skeptical of applying jumbo-sizing to these other ASCII chars, but am willing to let them go.

0x23: #
0x2A: *
0xA9: ©
0xAE: ®

charlesmchen: Yeah, I guess I'll disable those too.

PR #2720, opened 2017-11-02
Constrain the max text cell height to the height of the screen.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
@@ -304,7 +304,7 @@ NSString *NSStringForOWSMessageCellType(OWSMessageCellType cellType)
         NSString *text = textBlock();
 
         // Only show up to N characters of text.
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -304,7 +304,7 @@ - (DisplayableText *)displayableTextForInteractionId:(NSString *)interactionId
         NSString *text = textBlock();
 
         // Only show up to N characters of text.
-        const NSUInteger kMaxTextDisplayLength = 1024;
+        const NSUInteger kMaxTextDisplayLength = 512;
```
charlesmchen: Okay, for simplicity I've just dialed down the max character count.  I'm not sure exactly what the right value is.  Let's see if we feel like we're seeing "tap for more" too often.

charlesmchen:
PTAL @michaelkirk

PR #2722, opened 2017-11-02
Fix sizing of text bubbles in message detail view.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessageDetailViewController.swift
+++ b/Signal/src/ViewControllers/MessageDetailViewController.swift
@@ -36,6 +36,7 @@ class MessageDetailViewController: OWSViewController, UIScrollViewDelegate {
     var messageTextTopConstraint: NSLayoutConstraint?
     var messageTextHeightLayoutConstraint: NSLayoutConstraint?
     var messageTextProxyViewHeightConstraint: NSLayoutConstraint?
+    var bubbleViewWidthConstraint: NSLayoutConstraint?
 
     var scrollView: UIScrollView?
     var contentView: UIView?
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -343,25 +352,22 @@ class MessageDetailViewController: OWSViewController, UIScrollViewDelegate {
...
             messageTextProxyView.autoPinHeightToSuperview(withMargin: 10)
             self.messageTextProxyViewHeightConstraint = messageTextProxyView.autoSetDimension(.height, toSize:0)
 
             let row = UIView()
             row.addSubview(bubbleView)
             bubbleView.autoPinHeightToSuperview()
-            bubbleView.autoPinLeadingToSuperview(withMargin: 10)
-            bubbleView.autoPinTrailingToSuperview(withMargin: 10)
+            bubbleView.autoPinEdge(toSuperviewEdge: isIncoming ? .leading : .trailing, withInset: bubbleViewHMargin)
+            self.bubbleViewWidthConstraint = bubbleView.autoSetDimension(.width, toSize:0)
```
michaelkirk: Why are we setting the width to 0?

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -617,52 +623,72 @@ class MessageDetailViewController: OWSViewController, UIScrollViewDelegate {
...
+        guard let bubbleViewWidthConstraint = bubbleViewWidthConstraint else {
+            owsFail("\(TAG) Missing bubbleViewWidthConstraint")
             return
         }
 
         if messageTextView.width() != messageTextProxyView.width() {
             owsFail("\(TAG) messageTextView.width \(messageTextView.width) != messageTextProxyView.width \(messageTextProxyView.width)")
         }
 
+        let bubbleViewHMargin: CGFloat = 10
```
michaelkirk: Can we dedupe these to use the equivalent constants extracted around line 310

charlesmchen: Removed!

Signal/src/ViewControllers/MessageDetailViewController.swift
charlesmchen: This is a placeholder value. We set the actual value in `updateTextLayout()` which does measurement, etc..

PR #2724, opened 2017-11-03
Be more defensive about handling local notifications.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -765,21 +765,21 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 }
 
 - (void)application:(UIApplication *)application didReceiveLocalNotification:(UILocalNotification *)notification {
+    OWSAssert([NSThread isMainThread]);
+
     if (!self.isEnvironmentSetup) {
         OWSFail(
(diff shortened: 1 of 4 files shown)


Signal/src/AppDelegate.m
```diff
@@ -766,12 +766,20 @@ - (void)application:(UIApplication *)application
...
+    if (application.applicationState != UIApplicationStateActive) {
+        OWSFail(@"%@ ignoring %s because app is not yet active: %@.", self.tag, __PRETTY_FUNCTION__, notification);
         return;
     }
+    DDLogInfo(@"%@ %s %@", self.tag, __PRETTY_FUNCTION__, notification);
 
     [AppStoreRating preventPromptAtNextTest];
     [[PushManager sharedManager] application:application didReceiveLocalNotification:notification];
+    // We only want to receive a single local notification per launch.
+    [application cancelAllLocalNotifications];
```
michaelkirk: Do we need to reconcile this with NotificationManager/PushManager which keep some local state related to "current" local notifications?

michaelkirk:
> Just curious - what prompted this change? Is there a way to launch the app from multiple notifications?

Actually I've reproduced this myself. Makes sense! 👍

Signal/src/AppDelegate.m
charlesmchen: Done, although I think that stale state was harmless.

PR #2725, opened 2017-11-03
Charlesmchen/message status bias

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/Utils/MessageRecipientStatusUtils.swift
+++ b/Signal/src/ViewControllers/Utils/MessageRecipientStatusUtils.swift
@@ -124,27 +124,23 @@ class MessageRecipientStatusUtils: NSObject {
             return NSLocalizedString("MESSAGE_STATUS_FAILED", comment:"message footer for failed messages")
         } else if outgoingMessage.messageState == .attemptingOut {
             if outgoingMessage.hasAttachments() {
-                assert(outgoingMessage.messageState == .attemptingOut)
-
                 return NSLocalizedString("MESSAGE_STATUS_UPLOADING",
                                          comment:"message footer while attachment is uploading")
             } else {
-                assert(outgoingMessage.messageState == .attemptingOut)
-
                 return NSLocalizedString("MESSAGE_STATUS_SENDING",
                                          comment:"message status while message is sending.")
             }
         }
 
+        assert(outgoingMessage.messageState == .sentToService)
+
         let recipientReadMap = outgoingMessage.recipientReadMap
         if recipientReadMap.count > 0 {
-            assert(outgoingMessage.messageState == .sentToService)
             return NSLocalizedString("MESSAGE_STATUS_READ", comment:"message footer for read messages")
         }
 
         let recipientDeliveryMap = outgoingMessage.recipientDeliveryMap
         if recipientDeliveryMap.count > 0 {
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/Utils/MessageRecipientStatusUtils.swift
```diff
@@ -94,18 +115,32 @@ class MessageRecipientStatusUtils: NSObject {
         }
     }
 
+    // This method is per-message and "biased towards failure".
+    // See comments above.
     public class func statusMessage(outgoingMessage: TSOutgoingMessage,
                                     referenceView: UIView) -> String {
 
+        if outgoingMessage.messageState == .unsent {
```
michaelkirk: nit: The logical flow of this method seems to mirror`recipientStatus(outgoingMesssage:)`

Can we replace the logic in this method with something like:

```
switch recipientStatus(outgoingMessage: outgoingMessage) {
  case .unsent:
       return NSLocalizedString("MESSAGE_STATUS_FAILED", comment:"message footer for failed messages")
   case .attemptingOut:

...etc...

}

File: Signal/src/ViewControllers/Utils/MessageRecipientStatusUtils.swift [hunk 2]
```diff
@@ -13,13 +13,30 @@ import Foundation
...
 class MessageRecipientStatusUtils: NSObject {
     // MARK: Initializers
 
     @available(*, unavailable, message:"do not instantiate this class.")
     private override init() {
     }
 
+    // This method is per-recipient and "biased towards success".
+    // See comments above.
     public class func recipientStatus(outgoingMessage: TSOutgoingMessage,
```
michaelkirk: tldr; move the comments about bias to `recipientStatus(outgoingMessage:)` and remove the other comments.

NBD: As I understand this concept of "bias", we're talking about how to aggregate the status of the recipient(s) into a single status for the overall message.

When there is only a single recipient, there is no aggregation. It's simply the status of the single recipient. When there are multiple recipients we have some more complicated aggregation logic.  e.g. the message bubble footer text "Delivered" appears as soon as at least one person has been delivered the message.

As such, I found the comment introducing the concept of "bias" in single recipient status method confusing rather than helpful. In the group context, where we *have* aggregation logic, it makes sense to spell it out a bit more.

charlesmchen: That's not quite right.  I've elaborated the comment to be more clear why I speak of bias.

Signal/src/ViewControllers/Utils/MessageRecipientStatusUtils.swift
charlesmchen: Done.

Signal/src/ViewControllers/Utils/MessageRecipientStatusUtils.swift
michaelkirk: Ah, so we also need the bias in order to do the right (usually) thing for per-recipient details on legacy messages.

charlesmchen: I guess it's subjective, but in my opinion we're applying bias in the per-message case too.

PR #2730, opened 2017-11-06
Charlesmchen/first send vs linked devices

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Devices/OWSDevice.h
+++ b/SignalServiceKit/src/Devices/OWSDevice.h
@@ -9,6 +9,19 @@ NS_ASSUME_NONNULL_BEGIN
 
 extern uint32_t const OWSDevicePrimaryDeviceId;
 
+@interface OWSDeviceManager : NSObject
+
+- (instancetype)init NS_UNAVAILABLE;
+
++ (instancetype)sharedManager;
+
+- (BOOL)mayHaveLinkedDevices:(YapDatabaseConnection *)dbConnection;
+- (void)setMayHaveLinkedDevices:(BOOL)value dbConnection:(YapDatabaseConnection *)dbConnection;
+
+@end
+
+#pragma mark -
+
 @interface OWSDevice : TSYapDatabaseObject <MTLJSONSerializing>
 
 @property (nonatomic, readonly) NSInteger deviceId;
@@ -51,11 +64,6 @@ extern uint32_t const OWSDevicePrimaryDeviceId;
  *  YES if any values on self changed, else NO
  */
 - (BOOL)updateAttributesWithDevice:(OWSDevice *)other;
-
-#pragma mark - "May Have Linked Devices" Flag
-
-+ (BOOL)mayHaveLinkedDevices:(YapDatabaseConnection *)dbConnection;
-+ (void)setMayHaveLinkedDevices:(BOOL)value dbConnection:(YapDatabaseConnection *)dbConnection;
 
 @end
 
--- a/SignalServiceKit/src/Devices/OWSDevice.h
+++ b/SignalServiceKit/src/Devices/OWSDevice.h
@@ -12,9 +12,9 @@ extern uint32_t const OWSDevicePrimaryDeviceId;
 @interface OWSDevice : TSYapDatabaseObject <MTLJSONSerializing>
 
 @property (nonatomic, readonly) NSInteger deviceId;
-@property (nullable, readonly) NSString *name;
-@property (readonly) NSDate *createdAt;
-@property (readonly) NSDate *lastSeenAt;
+@property (nonatomic, readonly, nullable) NSString *name;
+@property (nonatomic, readonly) NSDate *createdAt;
+@property (nonatomic, readonly) NSDate *lastSeenAt;
 
 + (instancetype)deviceFromJSONDictionary:(NSDictionary *)deviceAttributes error:(NSError **)error;
 
@@ -51,6 +51,11 @@ extern uint32_t const OWSDevicePrimaryDeviceId;
  *  YES if any values on self changed, else NO
  */
 - (BOOL)updateAttributesWithDevice:(OWSDevice *)other;
+
+#pragma mark - "May Have Linked Devices" Flag
+
++ (BOOL)mayHaveLinkedDevices:(YapDatabaseConnection *)dbConnection;
++ (void)setMayHaveLinkedDevices:(BOOL)value dbConnection:(YapDatabaseConnection *)dbConnection;
 
 @end
 
--- a/SignalServiceKit/src/Devices/OWSDevice.m
+++ b/SignalServiceKit/src/Devices/OWSDevice.m
@@ -16,6 +16,67 @@ uint32_t const OWSDevicePrimaryDeviceId = 1;
 NSString *const kTSStorageManager_OWSDeviceCollection = @"kTSStorageManager_OWSDeviceCollection";
 NSString *const kTSStorageManager_MayHaveLinkedDevices = @"kTSStorageManager_MayHaveLinkedDevices";
 
+@interface OWSDeviceManager ()
+
+@property (atomic, nullable) NSNumber *mayHaveLinkedDevicesCached;
+
+@end
+
+#pragma mark -
+
+@implementation OWSDeviceManager
+
++ (instancetype)sharedManager
+{
+    static OWSDeviceManager *instance = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        instance = [[self alloc] initDefault];
+    });
+    return instance;
+}
+
+- (instancetype)initDefault
+{
+    return [super init];
+}
+
+- (BOOL)mayHaveLinkedDevices:(YapDatabaseConnection *)dbConnection
+{
+    OWSAssert(dbConnection);
+
+    @synchronized(self)
+    {
+        if (!self.mayHaveLinkedDevicesCached) {
+            self.mayHaveLinkedDevicesCached = @([dbConnection boolForKey:kTSStorageManager_MayHaveLinkedDevices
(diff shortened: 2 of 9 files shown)


charlesmchen:
PTAL @michaelkirk
--

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -988,7 +988,16 @@ - (void)sendMessage:(TSOutgoingMessage *)message
             // of new secondary devices when this message send fails.
             DDLogWarn(@"%@ sync message has no device messages but account has secondary devices.", self.tag);
         } else if (hasDeviceMessages) {
-            OWSFail(@"%@ sync message has device messages for unknown secondary devices.", self.tag);
+            for (NSDictionary *deviceMessage in deviceMessages) {
+                NSNumber *_Nullable deviceId = deviceMessage[@"destinationDeviceId"];
+                if (deviceId && [deviceId isKindOfClass:[NSNumber class]] && [deviceId isEqualToNumber:@(1)]) {
```
michaelkirk: >  trying to sending sync message to primary device, presumably because user has just registered.

I don't understand why a device would ever send a message to itself, *just registered* or otherwise. Can you expound?

charlesmchen: Because one device messages is created for each `devices` in `SignalRecipient`.  `SignalRecipient`'s device list defaults to a single entry for the primary device `1`, even if the  `SignalRecipient` represents yourself.  When a newly-registered user sends their first sync message (from their primary), it'll just be a single device message - to their own primary.  They'll get a `409` response code and update the actual device list for themselves (presumably empty) at that time.   

We could special-case the default `devices` list for the `SignalRecipient` that represents ourself, but I was looking for the safest, cheapest change in this critical code path.

michaelkirk: OK I now understand how this fixes an overzealous assert upon registration.

> We could special-case the default devices list for the SignalRecipient that represents ourself, but I was looking for the safest, cheapest change in this critical code path.

FWIW this proposal seems easier to understand. We never want to send a message to our own device.

If I understood our conversation earlier, you said this PR also fixes linking new devices - before this change linked devices wouldn't receive new messages (in master). I don't understand how this fixes that.

And stepping back a bit, if we're already involving OWSDevice, why are we using Recipient.devices at all? Why isn't OWSDevices on it's own enough to determine if we should be sending sync messages?

Though I have some concurrency concerns around the way we're doing that in master right now, see https://github.com/WhisperSystems/Signal-iOS/pull/2537#discussion_r149211898, it seems more straight forward to me to fix that, vs. the current approach in master which uses some combination of OWSDevices and Recipient.devices.

michaelkirk: Another argument for building towards something like `DeviceManager.shared.hasLinkedDevice` is that we can more easily continue to save resources.

e.g. the current solution prevents sending the message itself, but it doesn't prevent (e.g.) uploading the attachment that we're inevitably *not* going to sync

michaelkirk: Hm... looking further I'm having second thoughts about my proposal...  We actually don't know that our local list of OWSDevice is complete, e.g. if we start the linking process, then kill the iOS app, then complete linking on desktop, iOS won't know about that device unless we re-visit the devices screen. We could do more work to make sure we're up to date, but at that point, it's not as simple as I was advocating.

Maybe that's why you're proposing inspecting Recipient.devices, which may not be current, but will be *corrected* by the server as we send messages...

charlesmchen: > FWIW the "special-case" the SignalRecipient.selfRecipient proposal seems easier to understand. We never want to send a message to our own device.

Okay, I've taken that approach - and greatly elaborated the comments here.

charlesmchen: > If I understood our conversation earlier, you said this PR also fixes linking new devices - before this change, linked devices wouldn't receive new messages (in master). I don't understand how this fixes that.

Check out the new comments - is it clear now?

charlesmchen: > Maybe that's why you're proposing inspecting Recipient.devices, which may not be current, but will be corrected by the server as we send messages...

That's exactly right - for the purposes of sending, we want the service to be the single source of truth.  Not least, that means we could someday link devices from - for example - a linked iPad and the primary device would learn of that linked device the next time it tried to send a sync message.

charlesmchen:
PTAL @michaelkirk

charlesmchen:
PTAL @michaelkirk

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -954,11 +954,29 @@ - (void)sendMessage:(TSOutgoingMessage *)message
...
+        // linked device list has just changed.
+        //
+        // The linked device list is reflected in two separate pieces of state:
+        //
+        // * OWSDevice's state is updated when you link or unlink a device.
+        // * SignalRecipient's state is updated by 409 "Mismatched devices"
+        //   responses from the service.
+        //
+        // If _both_ of these pieces of state agree that there are no linked
+        // devices, then can safely skip sending sync message.
```
michaelkirk: I think this depends too much about the state of the OWSDevice collection. That collection (in it's current implementation) is not guaranteed to be in sync with the service. e.g. if the user leaves the screen or their app crashes before completing the linking process on desktop, we may not have knowledge of the secondary device. That's not new with this PR, but what's in master as of https://github.com/WhisperSystems/Signal-iOS/commit/1014f27f45fa12645b3cb7c02ea30f36a4ca50f3 is incorrect.

michaelkirk: I think the approach you mentioned earlier, about setting a dirty flag "mayHaveLinkedDevice" or something when the user initiates the device linking process.

charlesmchen:
I think we've got it now.  I consulted with Moxie and we don't need to honor the service response to sent messages since it should be redundant with the approach here and less complete.

charlesmchen:
PTAL @michaelkirk

SignalServiceKit/src/Devices/OWSDevice.h
```diff
@@ -52,6 +52,11 @@ extern uint32_t const OWSDevicePrimaryDeviceId;
  */
 - (BOOL)updateAttributesWithDevice:(OWSDevice *)other;
 
+#pragma mark - "May Have Linked Devices" Flag
+
++ (BOOL)mayHaveLinkedDevices:(YapDatabaseConnection *)dbConnection;
```
michaelkirk: nit: This isn't broken, but it does not seem like a a model level concern. e.g. belongs in a manager or other global singleton.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -954,14 +954,29 @@ - (void)sendMessage:(TSOutgoingMessage *)message
...
 
+        // 1. Check OWSDevice's state.
+        BOOL mayHaveLinkedDevices = [OWSDevice mayHaveLinkedDevices:self.dbConnection];
```
michaelkirk: There is still a concurrency concern here right? We're reading on a connection other than which we write with in the LinkDeviceController.

suggestion, which you can take or leave: move this method off the model class onto the DeviceService class, which is responsible for synchronizing access to `mayHaveLinkedDevice` (or a new tiny DeviceManager class).

SignalServiceKit/src/Storage/YapDatabaseConnection+OWS.m
```diff
@@ -52,6 +52,11 @@ - (nullable NSString *)stringForKey:(NSString *)key inCollection:(NSString *)col
 
 - (BOOL)boolForKey:(NSString *)key inCollection:(NSString *)collection
 {
+    return [self boolForKey:key inCollection:collection defaultValue:NO];
+}
+
+- (BOOL)boolForKey:(NSString *)key inCollection:(NSString *)collection defaultValue:(BOOL)defaultValue
+{
     NSNumber *_Nullable number = [self objectForKey:key inCollection:collection ofExpectedType:[NSNumber class]];
```
michaelkirk: I think we want something like: `return number ? number.boolValue : defaultValue`

charlesmchen: Thanks!

SignalServiceKit/src/Devices/OWSDevice.h
charlesmchen: Done.

SignalServiceKit/src/Messages/OWSMessageSender.m
charlesmchen: Okay, I've added caching which address this kind of race and will improve perf a tad.

charlesmchen:
PTAL @michaelkirk

SignalServiceKit/src/Devices/OWSDevice.m
```diff
@@ -5,26 +5,90 @@
...
+    }
+}
+
+- (void)setMayHaveLinkedDevices:(BOOL)value dbConnection:(YapDatabaseConnection *)dbConnection
+{
+    OWSAssert(dbConnection);
+
+    @synchronized(self)
+    {
+        self.mayHaveLinkedDevicesCached = @(YES);
```
michaelkirk: shouldn't this be ` self.mayHaveLinkedDevicesCached = @(value);`?

michaelkirk: Other than that, LGTM! Thanks for considering my voluminous feedback.

charlesmchen: Done. Thanks for repeatedly saving me from myself.

PR #2734, opened 2017-11-07
Fix crashes in attachment approval view.

charlesmchen:
PTAL @michaelkirk


Signal/src/ViewControllers/AttachmentApprovalViewController.swift
```diff
@@ -155,7 +155,7 @@ class AttachmentApprovalViewController: OWSViewController {
                                     backgroundColor:color,
                                     width:buttonWidth,
                                     height:buttonHeight,
-                                    target:target,
+                                    target:self,
```
charlesmchen: The crazy thing is that this only appears to crash on iOS 8.  I don't know why, but this works on iOS 11.

charlesmchen:
@michaelkirk Note that we don't have to ship this fix if we don't ship a v2.18.2 - this view (and this bug) are removed in v2.19.

PR #2735, opened 2017-11-07
Charlesmchen/permissions vs inactive

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -1434,8 +1434,8 @@ protocol CallServiceObserver: class {
     public func handleFailedCall(failedCall: SignalCall?, error: CallError) {
         AssertIsOnMainThread()
 
-        if case CallError.assertionError(description: let description) = error {
-            Logger.error(description)
+        if case CallError.assertionError(description:let description) = error {
+            owsFail(description)
         }
 
         if let failedCall = failedCall {
--- a/Signal/src/util/UIViewController+Permissions.m
+++ b/Signal/src/util/UIViewController+Permissions.m
@@ -65,9 +65,7 @@ NS_ASSUME_NONNULL_BEGIN
         callback(YES);
     } else if (status == AVAuthorizationStatusNotDetermined) {
         [AVCaptureDevice requestAccessForMediaType:AVMediaTypeVideo
-                                 completionHandler:^(BOOL granted) {
-                                     callback(granted);
-                                 }];
+                                 completionHandler:callback];
     } else {
         DDLogError(@"Unknown AVAuthorizationStatus: %ld", (long)status);
         callback(NO);
@@ -91,9 +89,7 @@ NS_ASSUME_NONNULL_BEGIN
         return;
     }
 
-    [[AVAudioSession sharedInstance] requestRecordPermission:^(BOOL granted) {
-        callback(granted);
-    }];
+    [[AVAudioSession sharedInstance] requestRecordPermission:callback];
 }
 
 @end


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3047,27 +3050,25 @@ - (void)requestRecordingVoiceMemo
     self.voiceMessageUUID = voiceMessageUUID;
 
     __weak typeof(self) weakSelf = self;
-    [[AVAudioSession sharedInstance] requestRecordPermission:^(BOOL granted) {
```
charlesmchen: Centralize the asking for camera and microphone permissions.

Signal/src/call/CallService.swift
```diff
@@ -1103,33 +1103,34 @@ protocol CallServiceObserver: class {
...
-
-        // We don't need to worry about the user granting or remoting this permission
-        // during a call while the app is in the background, because changing this
-        // permission kills the app.
-        if authStatus != .authorized {
-
-            OWSAlerts.showAlert(withTitle:NSLocalizedString("MISSING_CAMERA_PERMISSION_TITLE", comment: "Alert title when camera is not authorized"),
-                                message:NSLocalizedString("MISSING_CAMERA_PERMISSION_MESSAGE", comment: "Alert body when camera is not authorized"))
-
+        guard let frontmostViewController = UIApplication.shared.frontmostViewController else {
```
charlesmchen: Never ask for camera or microphone permissions if app is not active.  I think I'm going to modify this to only skip if in background.  iOS could temporarily make app inactive to show a system alert, e.g. "iOS update available."

Signal/src/util/UIViewController+Permissions.m
```diff
@@ -0,0 +1,101 @@
...
+    DDLogVerbose(@"[%@] ows_askForCameraPermissions", NSStringFromClass(self.class));
+
+    // Ensure callback is invoked on main thread.
+    void (^callback)(BOOL) = ^(BOOL granted) {
+        DispatchMainThreadSafe(^{
+            callbackParam(granted);
+        });
+    };
+
+    if ([UIApplication sharedApplication].applicationState != UIApplicationStateActive) {
```
charlesmchen: Never ask for camera or microphone permissions if app is not active. I think I'm going to modify this to only skip if in background. iOS could temporarily make app inactive to show a system alert, e.g. "iOS update available."

Signal/src/util/UIViewController+Permissions.m
```diff
@@ -0,0 +1,101 @@
...
+
+@implementation UIViewController (Permissions)
+
+- (void)ows_askForCameraPermissions:(void (^)(BOOL granted))callbackParam
+{
+    DDLogVerbose(@"[%@] ows_askForCameraPermissions", NSStringFromClass(self.class));
+
+    // Ensure callback is invoked on main thread.
+    void (^callback)(BOOL) = ^(BOOL granted) {
+        DispatchMainThreadSafe(^{
```
charlesmchen: Make sure permissions callbacks are always invoked on main thread.

Signal/src/call/CallService.swift
```diff
@@ -1434,7 +1435,7 @@ protocol CallServiceObserver: class {
         AssertIsOnMainThread()
 
         if case CallError.assertionError(description:let description) = error {
-            owsFail(description)
```
michaelkirk: I don't think we should get rid of this. Is there some assertion that you're regularly seeing which is making this inconvenient? Maybe we should revisit that assertion.

Signal/src/util/UIViewController+Permissions.m
```diff
@@ -0,0 +1,101 @@
...
+                                                                  callback(NO);
+                                                              }];
+        [alert addAction:dismissAction];
+
+        [self presentViewController:alert animated:YES completion:nil];
+    } else if (status == AVAuthorizationStatusAuthorized) {
+        callback(YES);
+    } else if (status == AVAuthorizationStatusNotDetermined) {
+        [AVCaptureDevice requestAccessForMediaType:AVMediaTypeVideo
+                                 completionHandler:^(BOOL granted) {
```
michaelkirk: nit: unnecessary wrap: `completionHandler:callback`

Signal/src/util/UIViewController+Permissions.m
```diff
@@ -0,0 +1,101 @@
...
+        });
+    };
+
+    if ([UIApplication sharedApplication].applicationState == UIApplicationStateBackground) {
+        DDLogError(@"Skipping microphone permissions request when app is in background.");
+        callback(NO);
+        return;
+    }
+
+    [[AVAudioSession sharedInstance] requestRecordPermission:^(BOOL granted) {
```
michaelkirk: nit: unnecessary wrap.

charlesmchen: Done.

Signal/src/util/UIViewController+Permissions.m
charlesmchen: Done.

Signal/src/call/CallService.swift
charlesmchen: I hit this assert while hanging up a test-device-to-simulator call.  I assumed it was yet another over-aggressive assert in `CallService`.  I'll restore it, but I'm not sure we should `owsFail` edge cases which will recover gracefully in production.

PR #2736, opened 2017-11-07
Fix contact offer

michaelkirk:
We never assign interaction, but we can get it via the viewItem.

// FREEBIE


Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
```diff
@@ -15,7 +15,7 @@
 
 @interface OWSContactOffersCell ()
 
-@property (nonatomic, nullable) OWSContactOffersInteraction *interaction;
+@property (nonatomic, nullable, readonly) OWSContactOffersInteraction *interaction;
```
charlesmchen: Do we even need this private property any more?  Isn't the accessor below sufficient?

michaelkirk: Ok - removed.

PR #2737, opened 2017-11-07
Ensure we re-upload push tokens after re-registering.

michaelkirk:
This fixes a situation where a user could send (but not receive) messages after using the new (debug only) re-register functionality.

I fixed the issue twice - either one would be sufficient, but seems good to be robust about this particular issue.

PTAL @charlesmchen


michaelkirk:
In 48df161e

PR #2742, opened 2017-11-08
Layout cell content when presenting the cell.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
@@ -30,6 +30,15 @@ NS_ASSUME_NONNULL_BEGIN
     return CGSizeZero;
 }
 
+- (void)setIsCellVisible:(BOOL)isCellVisible
+{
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
```diff
@@ -30,6 +30,15 @@ - (CGSize)cellSizeForViewWidth:(int)viewWidth contentWidth:(int)contentWidth
     return CGSizeZero;
 }
 
+- (void)setIsCellVisible:(BOOL)isCellVisible
+{
+    _isCellVisible = isCellVisible;
+
+    if (isCellVisible) {
+        [self layoutIfNeeded];
```
charlesmchen: I'm not sure if this is a complete solution since we don't have repros for all cases, but it helps in the case I can repro.  I'm surprised it's necessary.

michaelkirk: LGTM

PR #2743, opened 2017-11-08
Fixes for scrolling in conversation view.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
@@ -973,6 +973,9 @@ NS_ASSUME_NONNULL_BEGIN
     if (self.isIncoming) {
         return NO;
     }
+    if (self.cellType == OWSMessageCellType_DownloadingAttachment) {
+        return NO;
+    }
     if (!self.attachmentStream) {
         return NO;
     }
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -973,6 +973,9 @@ - (BOOL)isMediaBeingSent
     if (self.isIncoming) {
         return NO;
     }
+    if (self.cellType == OWSMessageCellType_DownloadingAttachment) {
+        return NO;
```
charlesmchen: Fixes a stray assert in attachment stream accessor below.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3505,13 +3505,6 @@ - (void)navigationTitleLongPressed:(UIGestureRecognizer *)gestureRecognizer
 
 #pragma mark - ConversationInputTextViewDelegate
 
-- (void)inputTextViewDidBecomeFirstResponder
-{
-    OWSAssert([NSThread isMainThread]);
-
-    [self scrollToBottomAnimated:YES];
```
charlesmchen: Don't scroll to bottom when we start editing text messages - scroll to bottom when we send.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3630,7 +3623,7 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView
     [self updateLastVisibleTimestamp];
     [self autoLoadMoreIfNecessary];
 
-    if ([self isScrolledAwayFromBottom]) {
+    if (self.isUserScrolling && [self isScrolledAwayFromBottom]) {
```
charlesmchen: Only auto-dismiss keyboard while user scroll gesture is active.

PR #2744, opened 2017-11-08
Increase content insets of attachment button.

charlesmchen:
PTAL @michaelkirk 

The attachment button should be easier to hit _AND_ visually symmetrical with the voice memo button.

attach


charlesmchen:
PTAL @michaelkirk

PR #2745, opened 2017-11-09
Charlesmchen/background vs contacts permission

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/contact/SystemContactsFetcher.swift
+++ b/Signal/src/contact/SystemContactsFetcher.swift
@@ -373,32 +373,40 @@ class SystemContactsFetcher: NSObject {
      *
      * @param   completion  completion handler is called on main thread.
      */
-    public func requestOnce(completion: ((Error?) -> Void)?) {
+    public func requestOnce(completion completionParam: ((Error?) -> Void)?) {
         AssertIsOnMainThread()
 
+        // Ensure completion is invoked on main thread.
+        let completion = { error in
+            DispatchMainThreadSafe({
+                completionParam?(error)
+            })
+        }
+
         guard !systemContactsHaveBeenRequestedAtLeastOnce else {
-            completion?(nil)
+            completion(nil)
             return
         }
         setupObservationIfNecessary()
 
         switch authorizationStatus {
         case .notDetermined:
+            if UIApplication.shared.applicationState == .background {
+                Logger.error("\(self.TAG) do not request contacts permission when app is in background")
+                completion(nil)
+                return
+            }
             self.contactStoreAdapter.requestAccess { (granted, error) in
                 if let error = error {
                     Logger.error("\(self.TAG) error fetching contacts: \(error)")
-                    DispatchQueue.main.async {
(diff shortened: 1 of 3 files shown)


SignalServiceKit/src/Contacts/SignalRecipient.m
```diff
@@ -23,7 +23,23 @@ - (instancetype)initWithTextSecureIdentifier:(NSString *)textSecureIdentifier
         return self;
     }
 
-    _devices = [NSMutableOrderedSet orderedSetWithObject:[NSNumber numberWithInt:1]];
+    OWSAssert([TSAccountManager localNumber].length > 0);
+    if ([[TSAccountManager localNumber] isEqualToString:textSecureIdentifier]) {
```
michaelkirk: This seems unrelated - can you verify this is supposed to be in this PR?

Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -373,32 +373,40 @@ class SystemContactsFetcher: NSObject {
      *
      * @param   completion  completion handler is called on main thread.
      */
-    public func requestOnce(completion: ((Error?) -> Void)?) {
+    public func requestOnce(completion completionParam: ((Error?) -> Void)?) {
         AssertIsOnMainThread()
 
+        // Ensure completion is invoked on main thread.
+        let completion = { error in
```
charlesmchen: We weren't always ensuring that the completion was invoked on main.

charlesmchen: This also avoids the `?()`

Signal/src/contact/SystemContactsFetcher.swift
```diff
@@ -373,32 +373,40 @@ class SystemContactsFetcher: NSObject {
...
         guard !systemContactsHaveBeenRequestedAtLeastOnce else {
-            completion?(nil)
+            completion(nil)
             return
         }
         setupObservationIfNecessary()
 
         switch authorizationStatus {
         case .notDetermined:
+            if UIApplication.shared.applicationState == .background {
```
charlesmchen: The meat of the matter.

charlesmchen:
PTAL @michaelkirk

Signal/src/contact/SystemContactsFetcher.swift
michaelkirk: > We weren't always ensuring that the completion was invoked on main.
  
Huh, I can't see which case we missed. It looks like they're all invoked on main. 

But I guess with this approach at worst we dispatch an empty block which is NBD.

charlesmchen: The missed cases are in `updateContacts`, but I decided to rework the completion handling in both methods with the same goal: ensure the that completion is always called on main thread and DRY.

charlesmchen:
> LGTM - please make sure this is in the test plan.

Done.

PR #2740, opened 2017-11-08
Mkirk/iphone x

michaelkirk:
Fixes a handful of issues around iPhoneX layout.

There should be no noticeable difference for other devices.

I'll attach some screenshots in-line with their corresponding diff.

PR #2749, opened 2017-11-10
Fix "sliding table content" issue on iOS 11.

charlesmchen:
PTAL @michaelkirk 

https://stackoverflow.com/questions/45573829/weird-uitableview-behaviour-in-ios11-cells-scroll-up-with-navigation-push-anima

### DIFF
--- a/Signal/src/ViewControllers/OWSTableViewController.m
+++ b/Signal/src/ViewControllers/OWSTableViewController.m
@@ -390,6 +390,12 @@ NSString * const kOWSTableCellIdentifier = @"kOWSTableCellIdentifier";
     self.tableView.delegate = self;
     self.tableView.dataSource = self;
     self.tableView.tableFooterView = [[UIView alloc] initWithFrame:CGRectZero];
+    if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(11, 0)) {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wpartial-availability"
+        self.tableView.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentNever;
+#pragma clang diagnostic pop
+    }
     [self.view addSubview:self.tableView];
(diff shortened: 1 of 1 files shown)


michaelkirk:
Between which view transitions are you seeing this behavior and on which device?

charlesmchen:
> Between which view transitions are you seeing this behavior and on which device?

Supposedly affects all iOS 11 devices.  I see it on my iPhone 11 simulators in XCode 9.1.

Sample transition: from "app settings" to "linked devices" views and then go back to "app settings".  You'll see it in both.

michaelkirk:
> Supposedly affects all iOS 11 devices. I see it on my iPhone 11 simulators in XCode 9.1.
> 
> Sample transition: from "app settings" to "linked devices" views and then go back to "app settings". You'll see it in both.

You're right, I can easily reproduce on iPhone7 iOS11.1 simulator, but not on my physical iPhone7 running iOS11.1. On the physical device everything seems fine as-is.

So it seems a.) this is a simulator-only bug, or b.) my device isn't affected for some reason.

Reading the documentation, I assume this feature is intended to "automatically handle" some layout issues that arise from iPhone X, so I'd like to make sure we're not going to make it harder to get our iPhone X layout in shape before merging.

Would you mind verifying that it doesn't break things fixed in https://github.com/WhisperSystems/Signal-iOS/pull/2740?

Or we could wait until #2740 is merged and I can verify.

charlesmchen:
> You're right, I can easily reproduce on iPhone7 iOS11.1 simulator, but not on my physical iPhone7 running iOS11.1. On the physical device everything seems fine as-is.

I repro on my iPhone 7+ running iOS 11.1.  We're having a lot of repro discrepancies between our physical devices.

charlesmchen:
> Or we could wait until #2740 is merged and I can verify.

Very happy to wait.

charlesmchen:
> Can we only apply this to iOS11.0 and iOS11.1 (since iOS11.2 seems to have fixed this bug).

Done.

Issue #2748, opened 2017-11-10
Bug in deleting attachments from self-shadow contacts


gsbabil:
- [x] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description

Signal allows sending messages/attachments to self, if there exists
a contact in the iOS address-book with the same number that was use
to register with Signal.

For example, if the number +12101114455 was used to register with
Signal, and there exists a contact in iOS address-book with the name
"John Doe" with +12101114455, both Signal-iOS and Signal-Desktop
will allow exchanging messages/attachments with John Doe.

Once a message is sent to such a "self-shadow" contact, it'll make
the round-trip through the server, and immediately echo-back the
same message or attachment under the same conversation thread.

This unexpected side-effect is in fact great for testing, debugging,
and even saving small notes-to-self as future records. This could
may be even be considered as a serious feature to be implemented
with other bells-and-whistles in the future.

However, there's a bug with exchanging attachment with such
self-shadow contacts. As mentioned above, both messages and
attachments can be sent/received without any issue. But if a
previously sent/received attachment exchanged with a self-shadow
contact is deleted, the remaining sent/received attachment record in
the local iOS database is corrupted.

### Steps to reproduce

1. Register a Signal account, for example with +12101114455.

2. Create an iOS contact called "John Doe" and add +12101114455 to
   this contact.

3. Send an image to "John Doe". The echo-back of the attachment will
   be received immediately and show up in the conversation-view.

4. Now delete either the sent or the received attachment.

5. Try to open the sent/received attachment after step 4.

6. Go back to recent-conversation-view, and come back to the
   conversation-view with John Doe. The remaining sent/received
   attachment will not show up anymore.

**Actual result:**

Corrupted local database as described in step 5 and 6 above.

**Expected result:**

The expected behavior is to have the attachment that was _not_
deleted to show up in the conversation-view with John Doe.

### Screenshots

![signal](https://user-images.githubusercontent.com/1061108/32665482-1fb163be-c680-11e7-8171-8b81b504b548.jpg)

### Device info

 - **Device**         : iPhone 6
 - **iOS version**    : 11.1.1
 - **Signal version** : 2.18.2.1

### Link to debug log

https://gist.github.com/anonymous/cb1ec56ef5d3c5d74f828b811b0e0e4c

charlesmchen:
This has been fixed and will be released in v2.19.  Thanks @gsbabil

gsbabil:
[...]

PR #2750, opened 2017-11-10
Improve handling of attachment edge cases.

charlesmchen:
* Improve handling of invalid message edge cases.
* Properly handle deletion of self-sent attachments.

PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSMessageSender.m
+++ b/SignalServiceKit/src/Messages/OWSMessageSender.m
@@ -1177,6 +1177,37 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
             [TSContactThread getOrCreateThreadWithContactId:contactId transaction:transaction];
         [cThread saveWithTransaction:transaction];
 
+        // We need to clone any attachments for message sent to self; otherwise deleting
+        // the incoming or outgoing copy of the message will break the other.
+        NSMutableArray<NSString *> *attachmentIds = [NSMutableArray new];
+        for (NSString *attachmentId in outgoingMessage.attachmentIds) {
+            TSAttachmentStream *_Nullable outgoingAttachment =
+                [TSAttachmentStream fetchObjectWithUniqueID:attachmentId];
+            OWSAssert(outgoingAttachment);
+            if (!outgoingAttachment) {
+                DDLogError(@"%@ Couldn't load outgoing attachment for message sent to self.", self.logTag);
+            } else {
+                TSAttachmentStream *incomingAttachment =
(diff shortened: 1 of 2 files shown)


charlesmchen:
Fixes: https://github.com/WhisperSystems/Signal-iOS/issues/2748

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -1155,14 +1155,45 @@ - (void)handleSendToMyself:(TSOutgoingMessage *)outgoingMessage
             [TSContactThread getOrCreateThreadWithContactId:contactId transaction:transaction];
         [cThread saveWithTransaction:transaction];
 
+        // We need to clone any attachments for message sent to self; otherwise deleting
+        // the incoming or outgoing copy of the message will break the other.
+        NSMutableArray<NSString *> *attachmentIds = [NSMutableArray new];
+        for (NSString *attachmentId in outgoingMessage.attachmentIds) {
+            TSAttachmentStream *_Nullable outgoingAttachment =
+                [TSAttachmentStream fetchObjectWithUniqueID:attachmentId];
```
michaelkirk: let's use the existing transaction param rather than opening a new one.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -1155,14 +1155,45 @@ - (void)handleSendToMyself:(TSOutgoingMessage *)outgoingMessage
...
+            OWSAssert(outgoingAttachment);
+            if (!outgoingAttachment) {
+                DDLogError(@"%@ Couldn't load outgoing attachment for message sent to self.", self.logTag);
+            } else {
+                TSAttachmentStream *incomingAttachment =
+                    [[TSAttachmentStream alloc] initWithContentType:outgoingAttachment.contentType
+                                                          byteCount:outgoingAttachment.byteCount
+                                                     sourceFilename:outgoingAttachment.sourceFilename];
+                NSError *error;
+                NSData *_Nullable data = [outgoingAttachment readDataFromFileWithError:&error];
```
michaelkirk: TIOLI: Especially for larger attachments, it would be nicer to copy on the filesystem rather than read and then write back the same data.

SignalServiceKit/src/Messages/OWSMessageSender.m
charlesmchen: Done, good one.

SignalServiceKit/src/Messages/OWSMessageSender.m
charlesmchen: I'm going to keep it simple here since this is a true edge case.

PR #2753, opened 2017-11-14
Charlesmchen/dont resurrect zombies

charlesmchen:
If you delete a message before its sent, the message sender should not re-save the message while trying to update its status.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/Jobs/SessionResetJob.swift
+++ b/Signal/src/Jobs/SessionResetJob.swift
@@ -30,9 +30,9 @@ class SessionResetJob: NSObject {
             self.storageManager.deleteAllSessions(forContact: self.recipientId)
 
             DispatchQueue.main.async {
-                let endSessionMessage = EndSessionMessage(timestamp:NSDate.ows_millisecondTimeStamp(), in: self.thread)
+                let endSessionMessage = EndSessionMessage(timestamp: NSDate.ows_millisecondTimeStamp(), in: self.thread)
 
-                self.messageSender.send(endSessionMessage, success: {
+                self.messageSender.enqueue(endSessionMessage, success: {
                     Logger.info("\(self.TAG) successfully sent EndSession<essage.")
                     let message = TSInfoMessage(timestamp: NSDate.ows_millisecondTimeStamp(),
                                                 in: self.thread,
--- a/SignalServiceKit/src/Contacts/TSThread.h
+++ b/SignalServiceKit/src/Contacts/TSThread.h
@@ -157,18 +157,19 @@ NS_ASSUME_NONNULL_BEGIN
 @property (atomic, readonly, nullable) NSDate *mutedUntilDate;
 
 // This model may be updated from many threads. We don't want to save
-// our local copy (this instance) since it may be out of date.  Instead, we
+// our local copy (this instance) since it may be out of date.  We also
+// want to avoid re-saving a model that has been deleted.  Therefore, we
 // use these "updateWith..." methods to:
 //
 // a) Update a property of this instance.
-// b) Load an up-to-date instance of this model from from the data store.
-// c) Update and save that fresh instance.
-// d) If this instance hasn't yet been saved, save this local instance.
+// b) If a copy of this model exists in the database, load an up-to-date copy,
+//    and update and save that copy.
+// b) If a copy of this model _DOES NOT_ exist in the database, do _NOT_ save
+//    this local instance.
 //
 // After "updateWith...":
 //
-// a) An updated copy of this instance will always have been saved in the
-//    data store.
+// a) An copy of this model in the database will have been updated.
 // b) The local property on this instance will always have been updated.
 // c) Other properties on this instance may be out of date.
 //
--- a/SignalServiceKit/src/Contacts/TSThread.h
+++ b/SignalServiceKit/src/Contacts/TSThread.h
@@ -156,28 +156,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (atomic, readonly) BOOL isMuted;
 @property (atomic, readonly, nullable) NSDate *mutedUntilDate;
 
-// This model may be updated from many threads. We don't want to save
-// our local copy (this instance) since it may be out of date.  We also
-// want to avoid re-saving a model that has been deleted.  Therefore, we
-// use these "updateWith..." methods to:
-//
-// a) Update a property of this instance.
-// b) If a copy of this model exists in the database, load an up-to-date copy,
-//    and update and save that copy.
-// b) If a copy of this model _DOES NOT_ exist in the database, do _NOT_ save
-//    this local instance.
-//
-// After "updateWith...":
-//
-// a) An copy of this model in the database will have been updated.
-// b) The local property on this instance will always have been updated.
-// c) Other properties on this instance may be out of date.
-//
-// All mutable properties of this class have been made read-only to
-// prevent accidentally modifying them directly.
-//
-// This isn't a perfect arrangement, but in practice this will prevent
-// data loss and will resolve all known issues.
+#pragma mark - Update With... Methods
+
 - (void)updateWithMutedUntilDate:(NSDate *)mutedUntilDate;
 
 @end
--- a/SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
+++ b/SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
@@ -109,6 +109,7 @@ NS_ASSUME_NONNULL_BEGIN
         return;
     }
 
+    [outgoingMessage saveWithTransaction:transaction];
     [outgoingMessage updateWithWasSentFromLinkedDeviceWithTransaction:transaction];
     [OWSDisappearingMessagesJob becomeConsistentWithConfigurationForMessage:outgoingMessage
                                                             contactsManager:self.contactsManager];
@@ -138,6 +139,7 @@ NS_ASSUME_NONNULL_BEGIN
                                                                      expiresInSeconds:transcript.expirationDuration
                                                                       expireStartedAt:transcript.expirationStartedAt];
         // Since textMessage is a new message, updateWithWasSentAndDelivered will save it.
+        [textMessage saveWithTransaction:transaction];
         [textMessage updateWithWasSentFromLinkedDeviceWithTransaction:transaction];
     }
 }
--- a/SignalServiceKit/src/Messages/Interactions/TSMessage.m
+++ b/SignalServiceKit/src/Messages/Interactions/TSMessage.m
@@ -16,6 +16,10 @@ static const NSUInteger OWSMessageSchemaVersion = 3;
 
 @interface TSMessage ()
 
+@property (nonatomic, nullable) NSString *body;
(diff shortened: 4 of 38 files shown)


charlesmchen:
@michaelkirk I decided to take a more aggressive approach to this solution. I'm going to add a bunch of comments. We should probably discuss offline.

Signal/src/Jobs/SessionResetJob.swift
```diff
@@ -30,9 +30,9 @@ class SessionResetJob: NSObject {
             self.storageManager.deleteAllSessions(forContact: self.recipientId)
 
             DispatchQueue.main.async {
-                let endSessionMessage = EndSessionMessage(timestamp:NSDate.ows_millisecondTimeStamp(), in: self.thread)
+                let endSessionMessage = EndSessionMessage(timestamp: NSDate.ows_millisecondTimeStamp(), in: self.thread)
 
-                self.messageSender.send(endSessionMessage, success: {
+                self.messageSender.enqueue(endSessionMessage, success: {
                     Logger.info("\(self.TAG) successfully sent EndSession<essage.")
                     let message = TSInfoMessage(timestamp: NSDate.ows_millisecondTimeStamp(),
```
charlesmchen: `OWSMessageSender` confusingly used the same naming pattern `sendMessage:` for two different sets of methods: 

* Methods for enqueuing outgoing messages.
* Methods for trying to deliver messages to the service.

I've renamed these methods in an attempt to make the code more readable.

SignalServiceKit/src/Contacts/TSThread.h
```diff
@@ -157,18 +157,19 @@ NS_ASSUME_NONNULL_BEGIN
 @property (atomic, readonly, nullable) NSDate *mutedUntilDate;
 
 // This model may be updated from many threads. We don't want to save
-// our local copy (this instance) since it may be out of date.  Instead, we
+// our local copy (this instance) since it may be out of date.  We also
+// want to avoid re-saving a model that has been deleted.  Therefore, we
```
charlesmchen: I've made parallel changes in `TSOutgoingMessage` and `TSThread`.  These changes are far more impactful for `TSOutgoingMessage`, but I'll explain once here.

The `updateWith...` approach has been fairly successful in avoiding the class of bugs where stale data is saved to the database, but as initially implemented it had the undesirable risk of re-saving deleted models.  

I'm changing the semantics of these methods to only save if there already is an existing copy of the model in the database.

In a number of places, we were relying on this side-effect of `updateWith...` methods and using them to save new models.  This was never a good idea.  We now are explicit about when to save models to the database. In the case of outgoing messages, we save them:

* When they are enqueued in the message sender.
* When we learn of an outgoing message from a linked device via transcript.

Although an important change, I believe this should be safe since outgoing messages should only be saved at these two points.

SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
```diff
@@ -109,6 +109,7 @@ - (void)runWithAttachmentHandler:(void (^)(TSAttachmentStream *attachmentStream)
         return;
     }
 
+    [outgoingMessage saveWithTransaction:transaction];
```
charlesmchen: We now explicitly saved outgoing messages created in response to a sync transcript.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -280,6 +280,15 @@ - (void)main
 
 - (void)tryWithRemainingRetries:(NSUInteger)remainingRetries
 {
+    // If the message has been deleted, abort send.
+    if (![TSOutgoingMessage fetchObjectWithUniqueID:self.message.uniqueId]) {
```
charlesmchen: If we can detect that a user has deleted an outgoing message before the sender could send it, abort the send.

charlesmchen: This is a more desirable UX, I believe.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -554,9 +568,8 @@ - (void)sendAttachmentData:(DataSource *)dataSource
         if (sourceFilename) {
             message.attachmentFilenameMap[attachmentStream.uniqueId] = sourceFilename;
         }
-        [message save];
```
charlesmchen: No need to save here; `enqueueOutgoingMessage:` will save the message.

SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -184,8 +184,7 @@ - (void)setExpirationForMessage:(TSMessage *)message
 
     // Don't clobber if multiple actions simultaneously triggered expiration.
     if (message.expireStartedAt == 0 || message.expireStartedAt > expirationStartedAt) {
-        message.expireStartedAt = expirationStartedAt;
-        [message saveWithTransaction:transaction];
+        [message updateWithExpireStartedAt:expirationStartedAt transaction:transaction];
```
charlesmchen: I also noticed that we weren't applying the `updateWith:...` pattern here.  Fixed.

charlesmchen:
The repro is:

* Put network link conditioner in `edge` mode.
* Send a text message.
* ASAP delete the message.

Expected: Message stays deleted.
Observed: Message reappears after a few seconds.

The symptom is much worse in large groups, since there are N sent/delivered/read updates.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -429,7 +438,11 @@ - (void)sendMessage:(TSOutgoingMessage *)message
         //
         // So we're using YDB behavior to ensure this invariant, which is a bit
         // unorthodox.
-        [message updateWithMessageState:TSOutgoingMessageStateAttemptingOut];
+        [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
+            // All outgoing messages should be saved at the time they are enqueued.
+            [message saveWithTransaction:transaction];
+            [message updateWithMessageState:TSOutgoingMessageStateAttemptingOut transaction:transaction];
```
michaelkirk: IIRC the "point" of `updateWithMessageState` was to avoid clobbers by applying the changes to the latest version of the saved model.

Since any potential clobbering would already be done by the previous line (though we're pretty sure the model shouldn't have been persisted until this moment**), it seems more to the point to say:

```
+  message.messageState = TSOutgoingMessageStateAttemptingOut
   [message saveWithTransaction:transaction]
-  [message updateWithMessageState:TSOutgoingMessageStateAttemptingOut transaction:transaction];
```
Is there more to it?

**On a related note, what would you think about adding an `isPersisted` attribute to TSYapModel that we could assert against?

It would get set in `initWithCoder` and in `saveWithTransaction`

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -410,9 +419,9 @@ - (NSOperationQueue *)sendingQueueForMessage:(TSOutgoingMessage *)message
     }
 }
 
-- (void)sendMessage:(TSOutgoingMessage *)message
-            success:(void (^)(void))successHandler
-            failure:(void (^)(NSError *error))failureHandler
+- (void)enqueueOutgoingMessage:(TSOutgoingMessage *)message
+                       success:(void (^)(void))successHandler
+                       failure:(void (^)(NSError *error))failureHandler
```
michaelkirk: I approve of these name changes!

TIOLI: we could safely omit the word "outgoing" for brevity, since you don't send "incoming" messages

SignalServiceKit/src/Messages/Interactions/TSMessage.m
```diff
@@ -277,6 +281,35 @@ - (BOOL)shouldUseReceiptDateForSorting
     return YES;
 }
 
+#pragma mark - Update Methods
+
+// This method does the work for the "updateWith..." methods.  Please see
+// the header for a discussion of those methods.
+- (void)applyChangeToSelfAndLatestMessage:(YapDatabaseReadWriteTransaction *)transaction
```
michaelkirk: TIOLI 

I got a little confused between this method and the outgoing message version `applyChangeToSelfAndLatestOutgoingMessage` since the names are so similar. Especially if we continue to tweak this method, we'll have to remember to change that logic in all places. And I imagine we'll apply this pattern in more and more places as time goes on.

WDYT about implementing this in TSYapDatabaseModel, and having the changeBlock take a generic (id) param. We'd lose the compiler type-checking on the block params, but I don't like copying around these none-trivial blocks of code and comments.

A middle ground could be that the implementation lives on TSYapDatabase, and any class where we'd like to take advantage of type checking can wrap the method with a type-specific flavor.

SignalServiceKit/src/Contacts/TSThread.h
```diff
@@ -157,18 +157,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+// b) If a copy of this model exists in the database, load an up-to-date copy,
+//    and update and save that copy.
+// b) If a copy of this model _DOES NOT_ exist in the database, do _NOT_ save
+//    this local instance.
 //
 // After "updateWith...":
 //
-// a) An updated copy of this instance will always have been saved in the
-//    data store.
+// a) An copy of this model in the database will have been updated.
```
michaelkirk: typo: "An copy" -> "Any copy"? 

Apart from the grammar error, I idiomatically use "Any" to connote that a thing might not exist.

charlesmchen: Done.

SignalServiceKit/src/Messages/Interactions/TSMessage.m
charlesmchen: Done.  I also applied `NS_ASSUME_NONNULL` to `TSYapDatabaseObject`.  I suggest you take a look at that - although I don't think its controversial.

SignalServiceKit/src/Messages/OWSMessageSender.m

SignalServiceKit/src/Messages/OWSMessageSender.m
charlesmchen:
```
```
+  message.messageState = TSOutgoingMessageStateAttemptingOut
   [message saveWithTransaction:transaction]
-  [message updateWithMessageState:TSOutgoingMessageStateAttemptingOut transaction:transaction];
```
Unfortunately, this would require making the `messageState` property writable, which we don't want to do.  Making these properties `readonly` is invaluable for ensuring correct usage IMO.

```
**On a related note, what would you think about adding an isPersisted attribute to TSYapModel that we could assert against?
```
I've been thinking about that for quite a while - I think it might be quite useful for other problems.  Let's make a card.

michaelkirk: 👍

SignalServiceKit/src/Messages/Interactions/TSMessage.m
michaelkirk: Nice, thanks for doing both those things.

PR #2755, opened 2017-11-15
Evacuate the view item cache.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
> For my understanding - is this only a perf win, or is this meant to address any of the behavior we talked about this AM?

This isn't a perf win; this is about correctness.  I don't _think_ this should affect us, but on the off chance we deleted and then re-saved an interaction (per the zombie resurrection bug), we wouldn't want to recycle a view item if we received the delete and insert events in the same database modified notification.

PR #2757, opened 2017-11-15
Fix edge cases in conversation view.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2893,7 +2893,6 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
                 case YapDatabaseViewChangeDelete: {
                     DDLogVerbose(@"YapDatabaseViewChangeDelete: %@, %@", rowChange.collectionKey, rowChange.indexPath);
                     [self.collectionView deleteItemsAtIndexPaths:@[ rowChange.indexPath ]];
-                    [rowsThatChangedSize removeObject:@(rowChange.indexPath.row)];
                     YapCollectionKey *collectionKey = rowChange.collectionKey;
                     OWSAssert(collectionKey.key.length > 0);
                     break;
@@ -2921,6 +2920,7 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
                         rowChange.newIndexPath);
                     [self.collectionView deleteItemsAtIndexPaths:@[ rowChange.indexPath ]];
                     [self.collectionView insertItemsAtIndexPaths:@[ rowChange.newIndexPath ]];
+                    [rowsThatChangedSize removeObject:@(rowChange.newIndexPath.row)];
                     break;
                 }
                 case YapDatabaseViewChangeUpdate: {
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2893,7 +2893,6 @@ - (void)yapDatabaseModified:(NSNotification *)notification
                 case YapDatabaseViewChangeDelete: {
                     DDLogVerbose(@"YapDatabaseViewChangeDelete: %@, %@", rowChange.collectionKey, rowChange.indexPath);
                     [self.collectionView deleteItemsAtIndexPaths:@[ rowChange.indexPath ]];
-                    [rowsThatChangedSize removeObject:@(rowChange.indexPath.row)];
```
charlesmchen: The values in `rowsThatChangedSize` are expressed as a "new index path", not an "old index path".  By definition, deletes only affect old rows.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2921,6 +2920,7 @@ - (void)yapDatabaseModified:(NSNotification *)notification
                         rowChange.newIndexPath);
                     [self.collectionView deleteItemsAtIndexPaths:@[ rowChange.indexPath ]];
                     [self.collectionView insertItemsAtIndexPaths:@[ rowChange.newIndexPath ]];
+                    [rowsThatChangedSize removeObject:@(rowChange.newIndexPath.row)];
```
charlesmchen: We don't want to reload a row that we just moved.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2938,7 +2938,9 @@ - (void)yapDatabaseModified:(NSNotification *)notification
         for (NSNumber *row in rowsThatChangedSize) {
             [rowsToReload addObject:[NSIndexPath indexPathForRow:row.integerValue inSection:0]];
         }
-        [self.collectionView reloadItemsAtIndexPaths:rowsToReload];
+        if (rowsToReload.count > 0) {
+            [self.collectionView reloadItemsAtIndexPaths:rowsToReload];
```
charlesmchen: Avoid a redundant call (although it should be safe).

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk: Maybe leave this GH comment in the code. 

This is a perf win, but isn't intended to address correctness, right?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk: So during a delete, we were failing to reload a cell. Specifically - they one found at the deleted cells old index - likely the subsequent interaction. 

This could account for message bubbles following a deleted interaction being mis-sized, no?

charlesmchen: That's about right.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen: This is all about correctness - `performBatchUpdates` will throw an exception if you update a row that you also insert.

PR #2758, opened 2017-11-15
Scroll only as far as necessary

michaelkirk:
On first load, message details for longer messages often get scrolled off the top of the screen.

We should only scroll as far as necessary.
 
**Before**
before-scroll-fix

**After**
after-scroll-fix

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MessageDetailViewController.swift
+++ b/Signal/src/ViewControllers/MessageDetailViewController.swift
@@ -38,7 +38,7 @@ class MessageDetailViewController: OWSViewController, UIScrollViewDelegate {
     var messageTextProxyViewHeightConstraint: NSLayoutConstraint?
     var bubbleViewWidthConstraint: NSLayoutConstraint?
 
-    var scrollView: UIScrollView?
+    var scrollView: UIScrollView!
     var contentView: UIView?
 
     var attachment: TSAttachment?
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -96,11 +96,27 @@ class MessageDetailViewController: OWSViewController, UIScrollViewDelegate {
...
-                self.scrollView!.setContentOffset(offset, animated: false)
+                let bubbleViewBottom = bubbleView.superview!.convert(bubbleView.frame, to: scrollView).maxY
+                let maxOffset =  bubbleViewBottom - showAtLeast
+                let lastPage = contentHeight - scrollViewHeight
+
+                let offset = CGPoint(x: 0, y: min(maxOffset, lastPage))
+
+                scrollView.setContentOffset(offset, animated: false)
+
+                //                scrollView.scrollRectToVisible(bubbleView.superview!.convert(bubbleView.frame, to: scrollView), animated: false)
```
charlesmchen: Remove?

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -96,11 +96,27 @@ class MessageDetailViewController: OWSViewController, UIScrollViewDelegate {
...
+
+                // TODO scroll no more than necessary
```
charlesmchen: Remove?

Signal/src/ViewControllers/MessageDetailViewController.swift
michaelkirk: Yup. thanks!

PR #2759, opened 2017-11-15
Fix calling; be explicit about which messages should be saved.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Devices/OWSReadReceiptsForSenderMessage.m
+++ b/SignalServiceKit/src/Devices/OWSReadReceiptsForSenderMessage.m
@@ -67,13 +67,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - TSYapDatabaseObject overrides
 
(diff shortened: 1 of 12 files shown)


SignalServiceKit/src/Devices/OWSReadReceiptsForSenderMessage.m
```diff
@@ -67,13 +67,9 @@ - (OWSSignalServiceProtosReceiptMessage *)buildReceiptMessage:(NSString *)recipi
...
 
-- (void)saveWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
+- (BOOL)shouldBeSaved
 {
-    // override superclass with no-op.
-    //
-    // There's no need to save this message, since it's not displayed to the user.
-    //
-    // Should we find a need to save this in the future, we need to exclude any non-serializable properties.
+    return NO;
```
charlesmchen: Would could DRY this up by adding `TSOutgoingSyncMessage` and `TSOutgoingCallMessage` classes, but it doesn't seem worth it.

PR #2764, opened 2017-11-16
Update UI to reflect changes to disappearing messages configuration.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
> This is only code cleanup with no change in behavior, right?

That's about right.

PR #2765, opened 2017-11-16
Mkirk/push registration

michaelkirk:
Verified that this also affects earlier versions of iOS.

PTAL @charlesmchen


Signal/src/environment/OWSPreferences.m
```diff
@@ -270,8 +270,8 @@ - (nullable NSString *)getVoipToken
 - (void)unsetRecordedAPNSTokens
 {
     DDLogWarn(@"%@ Forgetting recorded APNS tokens", self.tag);
-    [self setValue:nil forKey:OWSPreferencesKeyLastRecordedPushToken];
-    [self setValue:nil forKey:OWSPreferencesKeyLastRecordedVoipToken];
+    [self setValueForKey:OWSPreferencesKeyLastRecordedPushToken toValue:nil];
```
michaelkirk: This is unrelated to the bug, but a small fix to a debug action that I came upon while testing this.

charlesmchen: Nice.

PR #2768, opened 2017-11-16
Reduce font size of "load more" indicator.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
I'm considering getting rid of this indicator altogether now that we're auto-loading more messages based on scroll state.

michaelkirk:
LGTM

Maybe it should say "Loading more messages..."

charlesmchen:
PTAL @michaelkirk 

I significantly reworked this PR.

PR #2775, opened 2017-11-17
Fix layout of table views vs. nav bar.

charlesmchen:
PTAL @michaelkirk 

### Before (first row hidden by navigation bar)

img_6675

### After

img_6676

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
@@ -257,13 +257,13 @@ NS_ASSUME_NONNULL_BEGIN
         case 0:
             break;
         case 1:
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -257,13 +257,13 @@ - (UIFont *)textMessageFont
         case 0:
             break;
         case 1:
-            return [UIFont ows_regularFontWithSize:basePointSize * 2.f];
+            return [UIFont ows_regularFontWithSize:basePointSize + 18.f];
```
michaelkirk: Was this intentional? It seems unrelated.

charlesmchen: Removed!

PR #2777, opened 2017-11-17
Only show contact nag if we've denied contact access

michaelkirk:
PTAL @charlesmchen


michaelkirk:
Need to a do a little more testing... will reopen shortly.

michaelkirk:
Ok, this is ready now. PTAL @charlesmchen

PR #2782, opened 2017-11-17
Charlesmchen/push tokens about

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/AboutTableViewController.m
+++ b/Signal/src/ViewControllers/AboutTableViewController.m
@@ -3,6 +3,9 @@
 //
 
 #import "AboutTableViewController.h"
+#import "Environment.h"
+#import "OWSPreferences.h"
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/AboutTableViewController.m
```diff
@@ -3,20 +3,36 @@
...
 
     [self updateTableContents];
+
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(pushTokensDidChange:)
+                                                 name:[OWSSyncPushTokensJob PushTokensDidChange]
+                                               object:nil];
+}
+
+- (void)pushTokensDidChange:(NSNotification *)notification {
```
michaelkirk: nit: format

charlesmchen: Done.

PR #2783, opened 2017-11-19
Mkirk/calling edgecases

michaelkirk:
Fixes a few edge cases in calling. See details in individual commit messages.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/CallAudioService.swift
+++ b/Signal/src/call/CallAudioService.swift
@@ -137,9 +137,8 @@ struct AudioSource: Hashable {
     // `pulseDuration` is the small pause between the two vibrations in the pair.
     private let pulseDuration = 0.2
 
-    static private let sharedAudioSession = CallAudioSession()
     var audioSession: CallAudioSession {
-        return type(of: self).sharedAudioSession
+        return CallAudioSession.shared
     }
 
     // MARK: - Initializers
@@ -248,12 +247,6 @@ struct AudioSource: Hashable {
             }
         } catch {
             owsFail("\(TAG) failed setting audio source with error: \(error) isSpeakerPhoneEnabled: \(call.isSpeakerphoneEnabled)")
-        }
-
-        if call.state == .connected, !call.isOnHold {
-            audioSession.isRTCAudioEnabled = true
-        } else {
-            audioSession.isRTCAudioEnabled = false
         }
     }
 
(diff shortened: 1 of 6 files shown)


Signal/src/call/PeerConnectionClient.swift
```diff
@@ -359,7 +359,11 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
                 }
                 Logger.verbose("\(self.TAG) setting local session description: \(sessionDescription)")
                 self.peerConnection.setLocalDescription(sessionDescription.rtcSessionDescription,
-                                                        completionHandler: { _ in
+                                                        completionHandler: { error in
+                                                            guard error == nil else {
+                                                                reject(error!)
```
michaelkirk: I've never personally seen this fail, but it seems wrong that we would ignore these errors.

Signal/src/call/CallAudioService.swift
```diff
@@ -249,12 +248,6 @@ struct AudioSource: Hashable {
         } catch {
             owsFail("\(TAG) failed setting audio source with error: \(error) isSpeakerPhoneEnabled: \(call.isSpeakerphoneEnabled)")
         }
-
-        if call.state == .connected, !call.isOnHold {
-            audioSession.isRTCAudioEnabled = true
```
michaelkirk: This was sometimes** occurring before the CallKit provider had activated our audio session. This would result in the recipient not transmitting or hearing call audio.

**The reproduction for this was for the recipient to toggle speakerphone as soon as the call connected.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -697,6 +718,14 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
             assert(self.dataChannel == nil)
             self.dataChannel = dataChannel
             dataChannel.delegate = self
+
+            let pendingMessages = self.pendingDataChannelMessages
+            self.pendingDataChannelMessages = []
+            DispatchQueue.main.async {
```
charlesmchen: Shouldn't we capture weakself and unpack it to strongself here?

michaelkirk: Yes! Fixed.

PR #2784, opened 2017-11-19
Fix connected call showing as "unanswered" until view reload

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -3941,7 +3941,13 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
         CGFloat viewTopToContentBottom = self.safeContentHeight - self.collectionView.contentOffset.y;
 
         NSUInteger oldCellCount = [self.messageMappings numberOfItemsInGroup:self.thread.uniqueId];
+
         [self resetMappings];
+
+        // ViewItem's modified while we were not observing may be stale.
+        // TODO have a more fine-grained cache expiration based on rows modified.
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3992,6 +3992,8 @@ - (void)collectionViewDidChangeLayout
 
             ConversationViewItem *_Nullable viewItem = self.viewItemMap[interaction.uniqueId];
             if (viewItem) {
+                // It's possible the interaction has been updated.
+                [viewItem replaceInteraction:interaction transaction:transaction];
```
charlesmchen: See the existing usage of `reloadInteractionForViewItem` which should already handle this case.

michaelkirk:
Ok - updated based on our prior conversation. PTAL @charlesmchen

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3968,6 +3968,10 @@ - (void)resetMappings
             [self.messageMappings updateWithTransaction:transaction];
         }];
         [self updateMessageMappingRangeOptions:MessagesRangeSizeMode_Normal];
+
+        // ViewItem cache may be stale
+        // TODO have a more fine-grained cache expiration based on rows modified.
+        [self.viewItemMap removeAllObjects];
```
charlesmchen: I suggest we not do this in `resetMappings` but in `setShouldObserveDBModifications`, because `resetMappings` is also used in `loadMoreMessages` and perf will really suffer in the auto-loading experience if we do it here.

PR #2790, opened 2017-11-20
For most views, only try to dismiss keyboard when scroll drag starts.

charlesmchen:
We only want to do this (in the general case) a scroll pan gesture starts.  Specifically, we don't want to do this while a scroll pan is settling as it often won't reflect user intent.

PTAL @michaelkirk


michaelkirk:
In 7c0d372a

PR #2791, opened 2017-11-21
Fix deadlock when responding to contacts sync messages.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSMessageManager.m
+++ b/SignalServiceKit/src/Messages/OWSMessageManager.m
@@ -606,21 +606,30 @@ NS_ASSUME_NONNULL_BEGIN
         }
     } else if (syncMessage.hasRequest) {
         if (syncMessage.request.type == OWSSignalServiceProtosSyncMessageRequestTypeContacts) {
-            OWSSyncContactsMessage *syncContactsMessage =
-                [[OWSSyncContactsMessage alloc] initWithSignalAccounts:self.contactsManager.signalAccounts
-                                                       identityManager:self.identityManager
-                                                        profileManager:self.profileManager];
-            DataSource *dataSource =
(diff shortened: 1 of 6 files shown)


SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -606,21 +606,30 @@ - (void)handleIncomingEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
         }
     } else if (syncMessage.hasRequest) {
         if (syncMessage.request.type == OWSSignalServiceProtosSyncMessageRequestTypeContacts) {
-            OWSSyncContactsMessage *syncContactsMessage =
```
charlesmchen: Mostly a whitespace change.

michaelkirk: Is there a reason we don't want to do the same treatment for group sync requests?

SignalServiceKit/src/Security/OWSRecipientIdentity.m
```diff
@@ -123,6 +123,35 @@ + (YapDatabaseConnection *)dbConnection
     return sharedDBConnection;
 }
 
+- (void)saveWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
```
michaelkirk: Nice. 👌

SignalServiceKit/src/Messages/OWSMessageManager.m
charlesmchen: `[OWSSyncGroupsMessage buildPlainTextAttachmentDataWithTransaction:]` already takes a transaction and none of its inner workings (e.g. `[OWSGroupsOutputStream writeGroup:]` seem unsafe.

PR #2793, opened 2017-11-21
Update layout of "scroll down button" to reflect RTL.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
> Is there a reason we don't use autoLayout for the positioning of this button, or just legacy?

It certainly seems like we could and should.

PR #2795, opened 2017-11-21
Update conversation view UI to reflect unread state.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -11,6 +11,7 @@
 #import "ConversationHeaderView.h"
 #import "ConversationInputTextView.h"
 #import "ConversationInputToolbar.h"
+#import "ConversationScrollButton.h"
 #import "ConversationViewCell.h"
 #import "ConversationViewItem.h"
 #import "ConversationViewLayout.h"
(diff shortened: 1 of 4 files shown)


charlesmchen:
This fixes multiple issues:

* When an unread message arrives in a conversation without an unread indicator AND the user is scrolled away from the bottom of the screen, we should add in the unread indicator at that time.
* When an unread message arrives in a read conversation and the user is scrolled away from the bottom of the screen, the "scroll down" should turn blue to indicate that there are unread messages. (In the long run I think we should do more here, e.g. badge the scroll down button).
* When we scroll to bottom, we should synchronously update conversation UI.
* Use more auto layout.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2222,51 +2225,41 @@ - (void)clearUnreadMessagesIndicator
...
-- (void)createScrollButtons
+- (void)createConversationScrollButtons
 {
-    self.scrollDownButton = [self createScrollButton:@"\uf103" selector:@selector(scrollDownButtonTapped)];
+    self.scrollDownButton = [[ConversationScrollButton alloc] initWithIconText:@"\uf103"];
+    [self.scrollDownButton addTarget:self
+                              action:@selector(scrollDownButtonTapped)
+                    forControlEvents:UIControlEventTouchUpInside];
+    [self.view addSubview:self.scrollDownButton];
+    [self.scrollDownButton autoSetDimension:ALDimensionWidth toSize:self.scrollDownButton.width];
```
michaelkirk: We're constraining it's width to it's current width? Maybe this is idiomatic, but I've never seen it before and it seems a little weird.

Should it be something like self.scrollDownButton.circleSize.width?

charlesmchen: Done in https://github.com/WhisperSystems/Signal-iOS/commit/2ec1e7e92560b2c5f222ab77d479c3839ed79ed4

PR #2796, opened 2017-11-21
Force conversation view cells to update layout immediately.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
@@ -35,8 +35,51 @@ NS_ASSUME_NONNULL_BEGIN
     _isCellVisible = isCellVisible;
 
     if (isCellVisible) {
-        [self layoutIfNeeded];
+        [self forceLayoutImmediately];
     }
+}
+
+- (void)setFrame:(CGRect)frame
+{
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
```diff
@@ -35,10 +35,53 @@ - (void)setIsCellVisible:(BOOL)isCellVisible
     _isCellVisible = isCellVisible;
 
     if (isCellVisible) {
-        [self layoutIfNeeded];
+        [self forceLayoutImmediately];
```
michaelkirk: This seems kind of expensive, especially since we're doing it on all the cell's subviews recursively. It seems like we normally shouldn't have to do this.

Can you include any notes as to why we're forcing layout as opposed to doing the normal layoutIfNeeded pattern of UIKit? (e.g. does it mean that we're failing to mark something as `needsLayout` elsewhere?) and any hints on reproducing the issue so that we can maybe one day solve it at a lower cost?

charlesmchen: I've reverted this change - see this PR.  https://github.com/WhisperSystems/Signal-iOS/pull/2801

PR #2799, opened 2017-11-22
Use dedicated db connection for recipient identities.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Security/OWSRecipientIdentity.m
+++ b/SignalServiceKit/src/Security/OWSRecipientIdentity.m
@@ -103,12 +103,17 @@ OWSSignalServiceProtosVerifiedState OWSVerificationStateToProtoState(OWSVerifica
     }];
 }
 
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Security/OWSRecipientIdentity.m
```diff
@@ -103,12 +103,17 @@ - (void)updateWithChangeBlock:(void (^)(OWSRecipientIdentity *obj))changeBlock
...
+{
+    return self.dbReadWriteConnection;
+}
+
 /**
  * Override to disable the object cache to better enforce transaction semantics on the store.
  * Note that it's still technically possible to access this collection from a different collection,
  * but that should be considered a bug.
  */
-+ (YapDatabaseConnection *)dbConnection
```
charlesmchen: This is when you wish Obj-C let you indicate that your intent is to override, so that if the superclass method changes names, you get a build error.

PR #2800, opened 2017-11-22
Improve handling of unread indicator edge cases.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2859,6 +2859,7 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
 
     // We need to reload any modified interactions _before_ we call
     // reloadViewItems.
+    BOOL hasDeletions = NO;
     for (YapDatabaseViewRowChange *rowChange in rowChanges) {
         switch (rowChange.type) {
             case YapDatabaseViewChangeUpdate: {
@@ -2877,6 +2878,7 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
                 if (collectionKey.key) {
                     [self.viewItemMap removeObjectForKey:collectionKey.key];
                 }
+                hasDeletions = YES;
                 break;
             }
             default:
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2988,6 +2993,9 @@ - (void)yapDatabaseModified:(NSNotification *)notification
             if (scrollToBottom) {
                 [self scrollToBottomAnimated:shouldAnimateScrollToBottom && shouldAnimateUpdates];
             }
+            if (hasDeletions) {
+                [self cleanUpUnreadIndicatorIfNecessary];
```
charlesmchen: `cleanUpUnreadIndicatorIfNecessary` can be a little expensive, so only call it when we learn of deletions.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3431,6 +3433,25 @@ - (void)updateLastVisibleTimestamp
     self.hasUnreadMessages = numberOfUnreadMessages > 0;
 }
 
+- (void)cleanUpUnreadIndicatorIfNecessary
+{
+    BOOL hasUnreadIndicator = self.dynamicInteractions.unreadIndicatorPosition != nil;
```
charlesmchen: If there is an unread indicator and we can't find any unread messages, use `ensureDynamicInteractions` to clean up the obsolete unread indicator.

Signal/src/util/ThreadUtil.m
```diff
@@ -239,18 +239,40 @@ + (ThreadDynamicInteractions *)ensureDynamicInteractionsForThread:(TSThread *)th
             // Due to disappearing messages or manual deletion,
             // firstUnseenInteractionTimestampParameter may refer to an obsolete
             // interaction in which case we want to discard it.
-            TSInteraction *_Nullable lastInteraction =
-                [[transaction ext:TSMessageDatabaseViewExtensionName] lastObjectInGroup:thread.uniqueId];
```
charlesmchen: This approach was too simplistic - it didn't handle the case where there were info or error messages.

Signal/src/util/ThreadUtil.m
```diff
@@ -239,18 +239,40 @@ + (ThreadDynamicInteractions *)ensureDynamicInteractionsForThread:(TSThread *)th
...
-            if (!lastInteraction
-                || lastInteraction.timestampForSorting
-                    < firstUnseenInteractionTimestampParameter.unsignedLongLongValue) {
+            //
+            // Therefore, we should discard the existing unread indicator
+            // position if there are no "unreadable" messages after it.
+            __block TSInteraction *lastCallOrMessage = nil;
+            [[transaction ext:TSMessageDatabaseViewExtensionName]
```
charlesmchen: This search should be fairly quick - the worst case is if there is an obsolete unread indicator followed by hundreds of error or info messages.  That's extremely unlikely.

PR #2801, opened 2017-11-22
Reload adjacent rows using original indices, not final indices.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
@@ -35,51 +35,8 @@ NS_ASSUME_NONNULL_BEGIN
     _isCellVisible = isCellVisible;
 
     if (isCellVisible) {
-        [self forceLayoutImmediately];
+        [self layoutIfNeeded];
     }
-}
-
-- (void)setFrame:(CGRect)frame
-{
-    BOOL didSizeChange = CGSizeEqualToSize(self.frame.size, frame.size);
-
-    [super setFrame:frame];
-
-    if (didSizeChange) {
-        [self forceLayoutImmediately];
-    }
-}
-
-- (void)setBounds:(CGRect)bounds
-{
-    BOOL didSizeChange = CGSizeEqualToSize(self.bounds.size, bounds.size);
-
-    [super setBounds:bounds];
-
-    if (didSizeChange) {
-        [self forceLayoutImmediately];
-    }
-}
-
-- (void)forceLayoutImmediately
-{
-    NSArray<UIView *> *descendents = [ConversationViewCell collectSubviewsOfViewDepthFirst:self];
-    for (UIView *view in descendents) {
-        [view setNeedsLayout];
-    }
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
```diff
@@ -35,53 +35,10 @@ - (void)setIsCellVisible:(BOOL)isCellVisible
     _isCellVisible = isCellVisible;
 
     if (isCellVisible) {
-        [self forceLayoutImmediately];
+        [self layoutIfNeeded];
```
charlesmchen: Reverting this change; I think I misdiagnosed the issue and it should be fixed by the other changes in this PR.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -168,7 +168,7 @@ @interface ConversationViewController () <AVAudioPlayerDelegate,
 @property (nonatomic, readonly) ConversationViewLayout *layout;
 
 @property (nonatomic) NSArray<ConversationViewItem *> *viewItems;
-@property (nonatomic) NSMutableDictionary<NSString *, ConversationViewItem *> *viewItemMap;
+@property (nonatomic) NSMutableDictionary<NSString *, ConversationViewItem *> *viewItemCache;
```
charlesmchen: Renamed the cache.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2920,7 +2920,7 @@ - (void)yapDatabaseModified:(NSNotification *)notification
                         rowChange.finalIndex);
                     [self.collectionView insertItemsAtIndexPaths:@[ rowChange.newIndexPath ]];
                     // We don't want to reload a row that we just inserted.
-                    [rowsThatChangedSize removeObject:@(rowChange.finalIndex)];
+                    [rowsThatChangedSize removeObject:@(rowChange.originalIndex)];
```
charlesmchen: The key fix: Reload actions should use _original_ indices, not _final_ indices.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2970,59 +2970,31 @@ - (void)yapDatabaseModified:(NSNotification *)notification
 
     DDLogVerbose(@"self.viewItems.count: %zd -> %zd", oldViewItemCount, self.viewItems.count);
 
-    BOOL shouldReloadCollection = [self shouldReloadCollection:rowChanges];
```
charlesmchen: We don't need this hack either now, I believe.  Moves are working properly now that we're reloading with the original indices.

Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
michaelkirk: 🎉

PR #2815, opened 2017-11-28
Mkirk/share spike framework friendly

michaelkirk:
Paving the way for a sharing extension

- Add an embedded framework target to share application code between Signal-iOS and the sharing extension.

- Configured cocoapods to work with both the share extension and the main app. This required switching to use_frameworks, which in turn requires some other PR's to be merged.

- Added a template share extension (doesn't currently function).

- The above caused a bunch of build and run failures. I addressed what I easily could, but some will require bigger changes. In the interest of getting us a foundation we can work on in parallel, I broke some functionality,  and marked these places with "FIXME SHARINGEXTENSION". Highlight any that you think are too egregious and we can address them before merging.

PTAL @charlesmchen


SignalMessaging/Info.plist
```diff
@@ -0,0 +1,24 @@
...
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>6.0</string>
+	<key>CFBundleName</key>
+	<string>$(PRODUCT_NAME)</string>
+	<key>CFBundlePackageType</key>
+	<string>FMWK</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
+	<key>CFBundleVersion</key>
+	<string>$(CURRENT_PROJECT_VERSION)</string>
```
charlesmchen: I know you haven't gotten there yet, but my suggestion is that we keep the version_s_ of the main app and the share extension in lockstep.

SignalMessaging/Info.plist
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>$(DEVELOPMENT_LANGUAGE)</string>
+	<key>CFBundleExecutable</key>
+	<string>$(EXECUTABLE_NAME)</string>
+	<key>CFBundleIdentifier</key>
+	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
```
charlesmchen: FYI the app id I selected for the share extension is: `org.whispersystems.signal.shareextension` and the app group id is: `group.org.whispersystems.signal.group`.  

It would have been more standard and aesthetically pleasing to use `group.org.whispersystems.signal` but that was already taken, I'm guessing by an OWS dev before our time.

SignalServiceKit/src/Account/TSPreKeyManager.m
```diff
@@ -90,7 +90,8 @@ + (dispatch_queue_t)prekeyQueue
 
 + (void)checkPreKeysIfNecessary
 {
-    OWSAssert([UIApplication sharedApplication].applicationState == UIApplicationStateActive);
+    // FIXME SHARINGEXTENSION
+    //    OWSAssert([UIApplication sharedApplication].applicationState == UIApplicationStateActive);
```
charlesmchen: Proposal: We add a trivial global to SSK "application type", enum values: main app, app extension.  And we add a condition in a case like this.  And if we need to go further we add an `ApplicationContext` interface and the main app's AppDelegate or the sharing extension's main class implement that interface and register themselves.

This is where I foresee some kind

SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -364,11 +366,12 @@ - (void)timerDidFire
 {
     OWSAssert([NSThread isMainThread]);
 
-    if ([UIApplication sharedApplication].applicationState != UIApplicationStateActive) {
-        // Don't run when inactive.
-        OWSFail(@"%@ Disappearing messages job timer fired while app inactive.", self.logTag);
-        return;
-    }
+    // FIXME SHARINGEXTENSION
+    //    if ([UIApplication sharedApplication].applicationState != UIApplicationStateActive) {
```
charlesmchen:
```
if (AppContext.currentContent.type == AppTypeMain) {
 if ([UIApplication sharedApplication].applicationState != UIApplicationStateActive) {
```

SignalServiceKit/src/Util/NSDate+OWS.m
```diff
@@ -3,17 +3,15 @@
...
 
 @implementation NSDate (millisecondTimeStamp)
 
 + (uint64_t)ows_millisecondTimeStamp
 {
-    uint64_t milliseconds
-        = (uint64_t)(std::chrono::system_clock::now().time_since_epoch() / std::chrono::milliseconds(1));
-    return milliseconds;
+    NSDate *now = [self new];
+    return (uint64_t)(now.timeIntervalSince1970 * 1000);
```
charlesmchen: Huh - we should use `chrono` where we can; it's much higher precision.  Why can't `chrono` be used in the SAE?

charlesmchen: Please add a `FIXME SHARINGEXTENSION` here if you're not going to resolve this in this PR.

charlesmchen:
I applied the `AppContext` introduced in #2822 to resolve nearly all of the FIXMEs in https://github.com/WhisperSystems/Signal-iOS/pull/2828

PR #2820, opened 2017-11-28
Charlesmchen/app extension migration

charlesmchen:
PTAL @michaelkirk


SignalServiceKit/src/TSConstants.h
```diff
@@ -57,6 +57,8 @@ typedef enum { kSMSVerification, kPhoneNumberVerification } VerificationTranspor
 #define textSecureSetProfileNameAPIFormat @"v1/profile/name/%@"
 #define textSecureProfileAvatarFormAPI @"v1/profile/form/avatar"
 
+#define SignalApplicationGroup @"group.org.whispersystems.signal.group"
```
charlesmchen: We could convert these preprocessor symbols to being proper constants.

SignalServiceKit/src/Util/OWSFileSystem.m
```diff
@@ -28,6 +29,57 @@ + (void)protectFolderAtPath:(NSString *)path
...
+    BOOL success = [fileManager moveItemAtPath:oldFilePath toPath:newFilePath error:&error];
+    if (!success || error) {
+        NSString *errorDescription =
+            [NSString stringWithFormat:@"%@ Could not move file or directory from %@ to %@, error: %@",
+                      self.logTag,
+                      oldFilePath,
+                      newFilePath,
+                      error];
+        OWSFail(@"%@", errorDescription);
+        [NSException raise:exceptionName format:@"%@", errorDescription];
```
charlesmchen: We should discuss if raising an exception is the right thing to do here.

Signal/src/AppDelegate.m
```diff
@@ -102,6 +103,18 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
     DDLogWarn(@"%@ application: didFinishLaunchingWithOptions.", self.logTag);
 
+    // We need to this _after_ we set up logging but _before_ we do
+    // anything else.
+    [self ensureIsReadyForAppExtensions];
+
+#if RELEASE
+    // ensureIsReadyForAppExtensions may have changed the state of the logging
```
michaelkirk: I don't understand how ensureIsReadyForAppExtensions would change the state of the logging preference. Can you expound?

Signal/src/Profiles/OWSProfileManager.m
```diff
@@ -1384,14 +1384,29 @@ - (nullable UIImage *)loadProfileAvatarWithFilename:(NSString *)filename
     return image;
 }
 
++ (NSString *)oldProfileAvatarsDirPath
+{
+    return [[OWSFileSystem appDocumentDirectoryPath] stringByAppendingPathComponent:@"ProfileAvatars"];
+}
+
++ (NSString *)newProfileAvatarsDirPath
```
michaelkirk: nit: what's old will always be old, but nothing is new forever, and calling it "new" outside of the context of the migration is not clarifying.

suggested name:
`newProfileAvatarsDirPath` -> `profileAvatarsDirPath`

And maybe a bit clearer:
`oldProfileAvatarsDirPath` -> `profileAvatarsOldDirPath` or `profileAvatarsLegacyDirPath`

I feel the same about the similar names in TSAttachmentStream and TSStorageManager

Signal/src/environment/VersionMigrations.m
```diff
@@ -8,9 +8,10 @@
 #import "OWSDatabaseMigrationRunner.h"
 #import "PushManager.h"
 #import "SignalKeyingStorage.h"
-#import "TSAccountManager.h"
-#import "TSNetworkManager.h"
 #import <SignalServiceKit/AppVersion.h>
+#import <SignalServiceKit/NSUserDefaults+OWS.h>
```
michaelkirk: rebase conflict heads up, I think I already merged a similar change.

SignalServiceKit/src/Util/NSUserDefaults+OWS.m
```diff
@@ -0,0 +1,38 @@
...
+        id value = dictionary[key];
+        OWSAssert(value);
+        [appUserDefaults setObject:value forKey:key];
+    }
+}
+
++ (void)removeAll
+{
+    NSString *appDomain = NSBundle.mainBundle.bundleIdentifier;
+    [NSUserDefaults.standardUserDefaults removePersistentDomainForName:appDomain];
```
michaelkirk: Do we need to:
`[[NSUserDefaults standardUserDefaults] synchronize];`

SignalServiceKit/src/Util/NSUserDefaults+OWS.m
```diff
@@ -0,0 +1,38 @@
...
++ (void)migrateToSharedUserDefaults
+{
+    NSUserDefaults *appUserDefaults = self.appUserDefaults;
+
+    NSDictionary<NSString *, id> *dictionary = [NSUserDefaults standardUserDefaults].dictionaryRepresentation;
+    for (NSString *key in dictionary) {
+        id value = dictionary[key];
+        OWSAssert(value);
+        [appUserDefaults setObject:value forKey:key];
+    }
```
michaelkirk: Do we want to delete the old preferences at this point?

Signal/src/AppDelegate.m
```diff
@@ -102,6 +103,18 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
     DDLogWarn(@"%@ application: didFinishLaunchingWithOptions.", self.logTag);
 
+    // We need to this _after_ we set up logging but _before_ we do
```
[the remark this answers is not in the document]
charlesmchen: Done

Signal/src/AppDelegate.m
charlesmchen: I've elaborated the comment. It's because of `[NSUserDefaults migrateToSharedUserDefaults]`.

Signal/src/Profiles/OWSProfileManager.m
charlesmchen: I can't do: `newProfileAvatarsDirPath` -> `profileAvatarsDirPath` because that would conflict with the `once` getter which does a bunch of work.  

Instead, I'm replacing `old` and `new` with `legacy` and `sharedData` here and everywhere else we're doing a similar migration.

SignalServiceKit/src/Util/NSUserDefaults+OWS.m
charlesmchen: Great question. I decided to leave them in place for a few reasons.

The main thing is that we've never stored much or anything that important in `NSUserDefaults`.

SignalServiceKit/src/Util/NSUserDefaults+OWS.m
charlesmchen: Good idea - and I've reworked how we clear out these defaults too.

PR #2822, opened 2017-11-29
Add application context class.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -8,6 +8,7 @@
 #import "CodeVerificationViewController.h"
 #import "DebugLogger.h"
 #import "Environment.h"
+#import "MainAppContext.h"
 #import "NotificationsManager.h"
 #import "OWSContactsManager.h"
 #import "OWSContactsSyncing.h"
@@ -25,7 +26,6 @@
 #import "ViewControllerUtils.h"
 #import <AxolotlKit/SessionCipher.h>
 #import <SignalMessaging/SignalMessaging.h>
-#import <SignalServiceKit/AppContext.h>
 #import <SignalServiceKit/NSUserDefaults+OWS.h>
 #import <SignalServiceKit/OWSBatchMessageProcessor.h>
 #import <SignalServiceKit/OWSDisappearingMessagesJob.h>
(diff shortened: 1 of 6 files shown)


Signal/src/AppDelegate.m
```diff
@@ -48,7 +49,7 @@
 static NSString *const kURLSchemeSGNLKey                = @"sgnl";
 static NSString *const kURLHostVerifyPrefix             = @"verify";
 
-@interface AppDelegate ()
+@interface AppDelegate () <AppContext>
```
charlesmchen: It might make more sense to have separate `MainAppContext` and `AppExtensionContext` classes rather than a protocol implemented by the AppDelegate, etc.  Hopefully the contexts are stateless, and this might help avoid bad references with the odd lifecycle around app extensions.

Signal/src/util/MainAppContext.h
```diff
@@ -0,0 +1,9 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import <SignalServiceKit/AppContext.h>
+
+@interface MainAppContext : NSObject <AppContext>
```
michaelkirk: Though currently without affect, in case we add more later, can you include the NS_ASSUME_NONNULL

Signal/src/util/MainAppContext.m
```diff
@@ -0,0 +1,25 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "MainAppContext.h"
+
+@implementation MainAppContext
```
michaelkirk: NS_ASSUME_NONNULL

SignalServiceKit/src/Util/AppContext.h
```diff
@@ -0,0 +1,18 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+typedef void (^BackgroundTaskExpirationHandler)(void);
+
+@protocol AppContext <NSObject>
```
michaelkirk: NS_ASSUME_NONNULL

Signal/src/util/MainAppContext.h
charlesmchen: Done.

Signal/src/util/MainAppContext.m
charlesmchen: Done.

SignalServiceKit/src/Util/AppContext.h
charlesmchen: Done.

PR #2823, opened 2017-11-29
Charlesmchen/cross process db modifications

charlesmchen:
We won't be able to test this until other open PRs are merged.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -292,6 +292,10 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
                                                  name:YapDatabaseModifiedNotification
                                                object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(yapDatabaseModified:)
+                                                 name:YapDatabaseModifiedExternallyNotification
+                                               object:nil];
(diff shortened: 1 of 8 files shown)


SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -302,6 +302,7 @@ + (YapDatabaseDeserializer)logOnFailureDeserializer
 - (void)setupDatabaseWithSafeBlockingMigrations:(void (^_Nonnull)())safeBlockingMigrationsBlock
 {
     // Synchronously register extensions which are essential for views.
+    [TSDatabaseView registerCrossProcessNotifier];
```
charlesmchen: I need to measure this and make sure it is fast.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -291,6 +291,10 @@ - (void)addNotificationListeners
                                              selector:@selector(yapDatabaseModified:)
                                                  name:YapDatabaseModifiedNotification
                                                object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(yapDatabaseModified:)
+                                                 name:YapDatabaseModifiedExternallyNotification
+                                               object:nil];
```
charlesmchen: I need to test this and make sure it works appropriately.

Issue #2821, opened 2017-11-29
Signal 2.19.0 crashes from iOS keyboard interrupt


s1381614:
[X] I have searched open and closed issues for duplicates

### Bug description
Signal 2.19.0 crashes on iOS8 for all text input

### Steps to reproduce
- Load Signal
- Click on any contact 
- Compose message
- iOS keyboard is displayed 
- Signal crashes as soon as any letter is pressed on the keyboard
- Image attachments etc however work so it seems to be the iOS keyboard only

**Actual result:** 
Signal 2.19 crashes when any iOS keyboard button is pressed
**Expected result:** 
Doesnt crash and 

### Device info
**Device**: Iphone 6
**iOS version**: 8.4
**Signal version:** 2.19.0

michaelkirk:
Thanks for reporting this. It should be fixed by #2824. 

We have a really small iOS8 user base. Do you have any interest in being a beta-tester? We could really benefit by having one or two iOS8 beta users to help us catch these issues earlier. And if this crash would have happened to you while you were using the beta, at least you'd have the option of switching back to the app store build.

Keep in mind that in general, we tell people the beta is less stable software, more likely to crash or lose data, but all the better if we can fix those crashes in beta. If you're interested please contact support@whispersystems.org with the subject "Sign me up for the iOS beta"

PR #2824, opened 2017-11-29
Fix iOS8 crash on type

michaelkirk:
We were using an iOS9+ API unguarded.

PTAL @charlesmchen 

Fixes https://github.com/WhisperSystems/Signal-iOS/issues/2821


michaelkirk:
Let's discuss if this should be in a hotfix branch or if we want to cut a new release from master. 

Namely - do we want to include the concurrency changes around SignalService: https://github.com/WhisperSystems/Signal-iOS/commit/42e972ef7e68bd12b19bb3a8b1a06b7283900ad8

michaelkirk:
In a9021059

PR #2825, opened 2017-11-29
Fix the add photo permission crash on iOS 11.

charlesmchen:
https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html

screen shot 2017-11-29 at 12 43 43 pm

PTAL @michaelkirk


michaelkirk:
LGTM

PR #2826, opened 2017-11-29
Fix key commands crash on iOS 8.

charlesmchen:
PTAL @michaelkirk


big-r81:
duplicate ;-) see #2824

PR #2828, opened 2017-11-29
Charlesmchen/address sharing extension tod os

charlesmchen:
### This PR will need to be rebased after the spike PR is merged.

PTAL @michaelkirk


SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -366,12 +365,10 @@ - (void)timerDidFire
 {
     OWSAssert([NSThread isMainThread]);
 
-    // FIXME SHARINGEXTENSION
-    //    if ([UIApplication sharedApplication].applicationState != UIApplicationStateActive) {
-    //        // Don't run when inactive.
-    //        OWSFail(@"%@ Disappearing messages job timer fired while app inactive.", self.logTag);
-    //        return;
-    //    }
+    if (!CurrentAppContext().isMainAppAndActive) {
```
michaelkirk: Did you intend to lose this assert?, or do we want something more like:

```
if (CurrentAppContext().isMainApp) {
    OWSAssert(CurrentAppContext().isMainAppAndActive);
}
```

SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -541,9 +548,9 @@ - (NSData *)databasePassword
         // Try to reset app by deleting database.
         // FIXME SHARINGEXTENSION
         OWSFail(@"disabled while Share extension is WIP");
```
michaelkirk: Just making sure you intended to leave this `OWSFail` in. I think it will prevent new installs in the meanwhile, right?

SignalServiceKit/src/Util/AppContext.h
```diff
@@ -9,10 +9,22 @@ typedef void (^BackgroundTaskExpirationHandler)(void);
 - (BOOL)isMainApp;
 - (BOOL)isMainAppAndActive;
 
+// Should only be called if isMainApp is YES.
```
michaelkirk: Good docs, thanks 👏👏 👏

SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -514,20 +515,26 @@ - (NSData *)databasePassword
...
+
+        if (CurrentAppContext().isMainApp) {
+            UIApplicationState applicationState = CurrentAppContext().mainApplicationState;
+            if (applicationState == UIApplicationStateBackground) {
+                // TODO: Rather than crash here, we should detect the situation earlier
+                // and exit gracefully - (in the app delegate?). See the `
+                // This is a last ditch effort to avoid blowing away the user's database.
+                [self backgroundedAppDatabasePasswordInaccessibleWithErrorDescription:errorDescription];
+            }
+        }
```
michaelkirk: Do we want to:

```
else { 
   throw exception
}
```

I don't think the share extension should ever be setting the db password, should it?

SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
charlesmchen: Fixed.

SignalServiceKit/src/Storage/TSStorageManager.m
charlesmchen: Done, good point.

SignalServiceKit/src/Storage/TSStorageManager.m
charlesmchen: Removed.  Wanted to leave it in until I tested a bit.  

NOTE: I haven't tested this code path with the sharing extension yet, so don't run the sharing extension on your primary yet.

charlesmchen:
I'm merging this branch against `master` and closing this PR since I opened it against a temporary branch.

michaelkirk:
> I'm merging this branch against master and closing this PR since I opened it against a temporary branch.

I don't mind, but FYI as of a couple of months ago you can change the branch the PR targets.

charlesmchen:
> I don't mind, but FYI as of a couple of months ago you can change the branch the PR targets.

Yup but it wouldn't let me in this case.

PR #2835, opened 2017-11-30
Charlesmchen/l10n scripts vs sae

charlesmchen:
* Modify l10n string extraction script to scan `SAE` and `SignalMessaging`. 
* Modify l10n string extraction script to copy strings to `SAE`.
* Modify l10n string download script to copy strings to `SAE`.

NOTE: I'm assuming it'll be easiest to have both the main app and the SAE use l10n files that contain the superset of strings in either.

Question: should we follow the main app's convention of having a src folder?

PTAL @michaelkirk


Signal/translations/bin/auto-genstrings
```diff
@@ -64,6 +65,8 @@ echo "$OLDUTF8" | grep -Eo '^".*"' | \
 
 echo "Copying strings from JSQMessagesViewController"
 echo "
-// Strings Copied in from JSQMessagesViewController" >> $STRINGFILE
-cat $JSQ_STRINGS_PATH  | grep -v "^//" >> $STRINGFILE
+// Strings Copied in from JSQMessagesViewController" >> $MAIN_STRINGFILE
+cat $JSQ_STRINGS_PATH  | grep -v "^//" >> $MAIN_STRINGFILE
 
+echo "Copying strings from to iOS Share Extension"
```
michaelkirk: typo: "from to"

charlesmchen: Reverted.

charlesmchen:
> Can we delete the ShareExtension Localizable.strings altogether and instead reference the existing Signal-iOS Localizable.strings in the "Copy Bundle Resources" phase?

I was kind of torn about this, not least because Localizable.strings is "automagically" managed by XCode in some ways that will work for the main app, but not for the SAE.  But since you're leaning that way, I'll adopt that approach and we can switch back if we run into problems.

PR #2839, opened 2017-12-01
Refine loading view of share extension.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/OWSMessagesBubbleImageFactory.swift
+++ b/Signal/src/OWSMessagesBubbleImageFactory.swift
@@ -5,6 +5,7 @@
 import Foundation
 import JSQMessagesViewController
 import SignalServiceKit
+import SignalMessaging
 
 @objc
 class OWSMessagesBubbleImageFactory: NSObject {
--- a/SignalMessaging/SignalMessaging.h
+++ b/SignalMessaging/SignalMessaging.h
@@ -10,4 +10,5 @@ FOUNDATION_EXPORT double SignalMessagingVersionNumber;
 //! Project version string for SignalMessaging.
 FOUNDATION_EXPORT const unsigned char SignalMessagingVersionString[];
 
-// In this header, you should import all the public headers of your framework using statements like #import <SignalMessaging/PublicHeader.h>
+// The public headers of the framework
+#import <SignalMessaging/OWSLogger.h>
--- a/SignalMessaging/utils/OWSSwiftUtils.swift
+++ b/SignalMessaging/utils/OWSSwiftUtils.swift
@@ -7,7 +7,7 @@ import Foundation
 /**
  * We synchronize access to state in this class using this queue.
  */
-func assertOnQueue(_ queue: DispatchQueue) {
+public func assertOnQueue(_ queue: DispatchQueue) {
     if #available(iOS 10.0, *) {
         dispatchPrecondition(condition: .onQueue(queue))
     } else {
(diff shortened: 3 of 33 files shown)


Signal/src/OWSMessagesBubbleImageFactory.swift
```diff
@@ -5,6 +5,7 @@
 import Foundation
 import JSQMessagesViewController
 import SignalServiceKit
+import SignalMessaging
```
charlesmchen: In the course of doing this work, I moved `OWSSwiftUtils` to `SignalMessaging` so that we can use `owsFail()`, etc. in `SignalMessaging` and the SAE.  Unfortunately this means we need to explicitly import `SignalMessaging` in most of our .swift files.

SignalMessaging/SignalMessaging.h
```diff
@@ -10,4 +10,5 @@ FOUNDATION_EXPORT double SignalMessagingVersionNumber;
 //! Project version string for SignalMessaging.
 FOUNDATION_EXPORT const unsigned char SignalMessagingVersionString[];
 
-// In this header, you should import all the public headers of your framework using statements like #import <SignalMessaging/PublicHeader.h>
+// The public headers of the framework
+#import <SignalMessaging/OWSLogger.h>
```
charlesmchen: Any obj-c files that you want to use in `SignalMessaging` swift code needs to be here, like a bridging header.

SignalMessaging/utils/OWSSwiftUtils.swift
```diff
@@ -7,15 +7,15 @@ import Foundation
...
+public func assertOnQueue(_ queue: DispatchQueue) {
     if #available(iOS 10.0, *) {
         dispatchPrecondition(condition: .onQueue(queue))
     } else {
         // Skipping check on <iOS10, since syntax is different and it's just a development convenience.
     }
 }
 
-func owsFail(_ message: String) {
+public func owsFail(_ message: String) {
```
charlesmchen: These methods need to be public so that we can use them in the main app and the SAE.

SignalShareExtension/SAELoadViewController.swift
```diff
@@ -0,0 +1,85 @@
...
+import UIKit
+import SignalMessaging
+import PureLayout
+
+// All Observer methods will be invoked from the main thread.
+protocol SAELoadViewDelegate: class {
+    func shareExtensionWasCancelled()
+}
+
+class SAELoadViewController: UIViewController {
```
charlesmchen: I pulled out the loading view into a separate class.

SignalShareExtension/ShareViewController.swift
```diff
@@ -108,4 +82,10 @@ class ShareViewController: UIViewController {
 
         super.viewDidAppear(animated)
     }
+
+    // MARK: SAELoadViewDelegate
+
+    public func shareExtensionWasCancelled() {
+        self.extensionContext!.completeRequest(returningItems: [], completionHandler: nil)
```
charlesmchen: You "cancel" the SAE by completing with no items.

SignalShareExtension/SAELoadViewController.swift
```diff
@@ -0,0 +1,85 @@
...
+        label.autoHCenterInSuperview()
+        label.autoPinEdge(.top, to: .bottom, of: activityIndicator, withOffset: 25)
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+
+        self.navigationController?.isNavigationBarHidden = false
+
+        guard let activityIndicator = activityIndicator else {
```
michaelkirk: it's idiomatic to use implicitly unwrapped optionals to make this kind of code more readable.

charlesmchen: Missed this, will do this soon.

PR #2847, opened 2017-12-04
Add profile names to thread picker, search profile names

michaelkirk:
PTAL @charlesmchen


Signal/src/ViewControllers/SelectThreadViewController.m
```diff
@@ -131,11 +132,13 @@ - (void)updateTableContents
     __weak SelectThreadViewController *weakSelf = self;
     ContactsViewHelper *helper = self.contactsViewHelper;
     OWSTableContents *contents = [OWSTableContents new];
-    OWSTableSection *section = [OWSTableSection new];
 
-    // Threads
+    // Existing threads are listed first, ordered by most recently active
+    OWSTableSection *recentChatsSection = [OWSTableSection new];
+    recentChatsSection.headerTitle = NSLocalizedString(
+        @"SELECT_THREAD_TABLE_RECENT_CHATS_TITLE", @"Table section header for recently active conversations");
```
charlesmchen: Just spitballing here: I wonder if we should interleave these search results (visually distinguishing contacts from conversations somehow) and sort them by "match weight" where, for example, matches at the start of a name would be weighted more heavily.

e.g. searching for "Bo", a user would see "Bob James <contact>" before "Trip to Boulder <group>" and a conversation with "Bob James <1:1>" before "Alice Arecibo <contact>".

Signal/src/util/Searcher.swift
```diff
@@ -3,6 +3,91 @@
 //
 
 import Foundation
+import SignalServiceKit
+
+@objc
+class ConversationSearcher: NSObject {
```
charlesmchen: Should each Searcher have its own source file?

SignalMessaging/utils/DebugLogger.m
```diff
@@ -106,7 +106,7 @@ - (void)enableTTYLogging
     // This should be redundant with the logic above.
     [logPathSet addObjectsFromArray:self.fileLogger.logFileManager.unsortedLogFilePaths];
     NSArray<NSString *> *logPaths = logPathSet.allObjects;
-    return [logPaths sortedArrayUsingSelector:@selector(compare:)];
+    return [logPaths sortedArrayUsingSelector:@selector((compare:))];
```
charlesmchen: Huh!  Why was this necessary?

charlesmchen: Is this to suppress / work around the "multiple matching methods with varying types" warning?

Signal/src/util/Searcher.swift
michaelkirk: Sure, I'll move ConversationSearcher into it's own file. I'm going to leave AnySearcher with Searcher since it's just a thing wrapper for objc compatibility.

SignalMessaging/utils/DebugLogger.m
michaelkirk: It was - I can't imagine adding those parentheses actually makes it safer though, probably just snuck past the linter.

Signal/src/ViewControllers/SelectThreadViewController.m
michaelkirk: Happy to discuss this more for a future iteration.

PR #2849, opened 2017-12-05
Charlesmchen/sae vs tests 

charlesmchen:
Some important changes here... I'll call them out in comments.

PTAL @michaelkirk

### DIFF
--- a/Signal.xcodeproj/project.pbxproj
+++ b/Signal.xcodeproj/project.pbxproj
@@ -191,36 +191,25 @@
 		34F308A21ECB469700BB7697 /* OWSBezierPathView.m in Sources */ = {isa = PBXBuildFile; fileRef = 34F308A11ECB469700BB7697 /* OWSBezierPathView.m */; };
 		34FD93701E3BD43A00109093 /* OWSAnyTouchGestureRecognizer.m in Sources */ = {isa = PBXBuildFile; fileRef = 34FD936F1E3BD43A00109093 /* OWSAnyTouchGestureRecognizer.m */; };
 		4505C2BF1E648EA300CEBF41 /* ExperienceUpgrade.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4505C2BE1E648EA300CEBF41 /* ExperienceUpgrade.swift */; };
-		4505C2C01E648EA300CEBF41 /* ExperienceUpgrade.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4505C2BE1E648EA300CEBF41 /* ExperienceUpgrade.swift */; };
 		4509E79A1DD653700025A59F /* WebRTC.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4509E7991DD653700025A59F /* WebRTC.framework */; };
 		450D19131F85236600970622 /* RemoteVideoView.m in Sources */ = {isa = PBXBuildFile; fileRef = 450D19121F85236600970622 /* RemoteVideoView.m */; };
 		450DF2051E0D74AC003D14BE /* Platform.swift in Sources */ = {isa = PBXBuildFile; fileRef = 450DF2041E0D74AC003D14BE /* Platform.swift */; };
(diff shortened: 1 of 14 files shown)


Signal.xcodeproj/project.pbxproj
```diff
@@ -191,36 +191,25 @@
 		34F308A21ECB469700BB7697 /* OWSBezierPathView.m in Sources */ = {isa = PBXBuildFile; fileRef = 34F308A11ECB469700BB7697 /* OWSBezierPathView.m */; };
 		34FD93701E3BD43A00109093 /* OWSAnyTouchGestureRecognizer.m in Sources */ = {isa = PBXBuildFile; fileRef = 34FD936F1E3BD43A00109093 /* OWSAnyTouchGestureRecognizer.m */; };
 		4505C2BF1E648EA300CEBF41 /* ExperienceUpgrade.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4505C2BE1E648EA300CEBF41 /* ExperienceUpgrade.swift */; };
-		4505C2C01E648EA300CEBF41 /* ExperienceUpgrade.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4505C2BE1E648EA300CEBF41 /* ExperienceUpgrade.swift */; };
```
charlesmchen: swift files should NOT be added to both the `Signal` and `SignalTests` targets.  Instead we should `@testable import Signal`.  Adding them to both targets causes two separate copies to be linked.  One side effect is that we'd instantiate two copies of our `shared` singletons!

SignalMessaging/environment/Environment.m
```diff
@@ -43,6 +43,11 @@ + (void)setCurrent:(Environment *)environment
     sharedEnvironment = environment;
 }
 
++ (void)clearCurrentForTests
+{
+    sharedEnvironment = nil;
```
charlesmchen: Alternately we could remove the new assert in the setter.

Signal/test/util/DisplayableTextFilterTest.swift
```diff
@@ -82,7 +83,7 @@ class DisplayableTextTest: XCTestCase {
         XCTAssertTrue("❤️💔💌💕💞💓💗💖💘💝💟💜💛💚💙".containsOnlyEmoji)
         XCTAssertTrue("✋🏿💪🏿👐🏿🙌🏿👏🏿🙏🏿".containsOnlyEmoji)
         XCTAssertTrue("🚾🆒🆓🆕🆖🆗🆙🏧".containsOnlyEmoji)
-        XCTAssertTrue("0️⃣1️⃣2️⃣3️⃣4️⃣5️⃣6️⃣7️⃣8️⃣9️⃣🔟".containsOnlyEmoji)
+        XCTAssertFalse("0️⃣1️⃣2️⃣3️⃣4️⃣5️⃣6️⃣7️⃣8️⃣9️⃣🔟".containsOnlyEmoji)
```
michaelkirk: wha? Did this change?

SignalMessaging/environment/Environment.m
michaelkirk: I prefer it as written for now. 👍

SignalMessaging/environment/Release.m
```diff
@@ -14,11 +14,13 @@ @implementation Release
 
 + (Environment *)releaseEnvironment
 {
+    // Order matters here.
+    TSStorageManager *storageManager = [TSStorageManager sharedManager];
```
michaelkirk: Was something broken before? I can't think of any relevant side effects `[TSStorageManager sharedManager]` would have.

Signal/test/util/DisplayableTextFilterTest.swift
charlesmchen: I think we removed this - due to feedback from you in an old PR - and I forgot to update the tests after taking your suggestion.

SignalMessaging/environment/Release.m
charlesmchen: There was a race wherein (if I recall correctly) `OWSContactsManager`'s initializer kicks of `loadCachedDisplayNames` which async uses `TSStorageManager`, which expects to be initialized on the main thread.  Something like that.

PR #2850, opened 2017-12-05
Charlesmchen/environment3

charlesmchen:
This PR is now ready for review.

PTAL @michaelkirk

PR #2851, opened 2017-12-05
Share UI draft

michaelkirk:
- Shows a searchable conversation picker (via the existing SendExternalFileViewController)
- Resurrects the AttachmentApprovalViewController

This is a super rough cut, but it puts a lot of bones in place (e.g. all the attachment related models and views)

Also, this currently only supports sharing jpegs.

PTAL @charlesmchen


Podfile
```diff
@@ -16,6 +16,10 @@ def shared_pods
   #pod 'GRKOpenSSLFramework', path: '../GRKOpenSSLFramework'
 
   # third party pods
+  pod 'AFNetworking', inhibit_warnings: true
+  pod 'JSQMessagesViewController',  git: 'https://github.com/WhisperSystems/JSQMessagesViewController.git', branch: 'mkirk/share-compatible', :inhibit_warnings => true
```
michaelkirk: Avatar building uses JSQ classes, but JSQ is not framework compatible. I think removing this dependency will be straight forward, but I left it as a TODO to get this PR up.

In the meanwhile, I just hacked JSQ to comment out the corresponding methods.

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
```diff
@@ -372,7 +372,7 @@ class GifPickerViewController: OWSViewController, UISearchBarDelegate, UICollect
                 owsFail("\(strongSelf.TAG) couldn't load asset.")
                 return
             }
-            let attachment = SignalAttachment(dataSource: dataSource, dataUTI: asset.rendition.utiType)
+            let attachment = SignalAttachment.imageAttachment(dataSource: dataSource, dataUTI: asset.rendition.utiType)
```
michaelkirk: The stricter rules around frameworks has exposed that we were using the wrong API here.

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -434,7 +434,7 @@ class MessageDetailViewController: OWSViewController, UIScrollViewDelegate {
 
         let contentType = attachment.contentType
         if let dataUTI = MIMETypeUtil.utiType(forMIMEType: contentType) {
-            let attachment = SignalAttachment(dataSource: dataSource, dataUTI: dataUTI)
+            let attachment = SignalAttachment.attachment(dataSource: dataSource, dataUTI: dataUTI)
```
michaelkirk: The stricter rules around frameworks has exposed that we were using the wrong API here.

SignalMessaging/attachments/AttachmentApprovalViewController.swift
```diff
@@ -0,0 +1,171 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+import MediaPlayer
+
+@objc
+public protocol AttachmentApprovalViewControllerDelegate: class {
```
michaelkirk: The old version of AttachmentApprovalViewController took a completion handler, but I went for the delegate approach instead so we could have more flexibility.

SignalMessaging/attachments/MediaMessageView.swift
```diff
@@ -456,6 +479,14 @@ class MediaMessageView: UIView, OWSAudioAttachmentPlayerDelegate {
         videoPlayer?.stop()
         videoPlayer?.view.removeFromSuperview()
         videoPlayer = nil
-        ViewControllerUtils.setAudioIgnoresHardwareMuteSwitch(false)
+        OWSAudioAttachmentPlayer.setAudioIgnoresHardwareMuteSwitch(false)
+    }
+}
+
+extension UIImage {
+    public convenience init?(named: String, inBundleForClass someClass: AnyClass) {
```
michaelkirk: I've just dropped this here for now, but I imagine there may be other resources that want to use this in the SAE. LMK if you have a good/ergonomic approach for having assets potentially distributed across frameworks.

charlesmchen:
@michaelkirk  Screenshots?

charlesmchen:
> This is a super rough cut, but it puts a lot of bones in place (e.g. all the attachment related models and views)

Is all of the remaining work tracked in cards?

Podfile
charlesmchen: Okay, I would _love_ to remove our dependency on JSQ.  I've got a bunch of work in that direction in the "message bubble" branch.

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
charlesmchen: Should we make that constructor private?

I agree this was wrong in principle, but it looks like this would have no negative side effects since the validation in `imageAttachment` should never apply to GIFs from gif picker, right?

.gitignore
```diff
@@ -26,5 +26,6 @@ DerivedData
 *.hmap
 *.ipa
 *.xcuserstate
+Index/
```
charlesmchen: What's this?

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
charlesmchen: Which stricter rules, incidentally?  Around method visibility?

SignalMessaging/Views/ContactsViewHelper.m
```diff
@@ -284,7 +286,7 @@ - (void)presentMissingContactAccessAlertControllerFromViewController:(UIViewCont
                                                                   @"Button text which opens the settings app")
                                                         style:UIAlertActionStyleDefault
                                                       handler:^(UIAlertAction *_Nonnull action) {
-                                                          [[UIApplication sharedApplication] openSystemSettings];
+                                                          [CurrentAppContext() openSystemSettings];
```
charlesmchen: Should we change the button label / alert options if not in main app?

.gitignore
michaelkirk: IDK, something XCode is creating. I think it started appearing with XCode9.1.

michaelkirk: Even though I was using the default DerivedData path, turns out I had (ages ago) specified a custom path in the "Advanced" section.

SignalMessaging/attachments/AttachmentApprovalViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+
+    // MARK: - Create Views
+
+    private func createViews() {
+        let previewTopMargin: CGFloat = 30
+        let previewHMargin: CGFloat = 20
+
+        self.view.addSubview(mediaMessageView)
+        mediaMessageView.autoPinWidthToSuperview(withMargin:previewHMargin)
+        mediaMessageView.autoPin(toTopLayoutGuideOf: self, withInset:previewTopMargin)
```
charlesmchen: * WhatsApp uses a "white background" contact picker view and a "black background" for the approval/caption/etc. view.  I'm inclined to use a white background for both.  
* WhatsApp's attachment approval view has a pretty small "send" button.  I wonder what our caption/sender UI will look like.  I'm not sure it should like like conversation view's, for example.  e.g. Conversation view's input toolbar "grows" as the text message grows in length.  I think this growth might be much more jarring in the context of attachment approval.

Thoughts?

SignalMessaging/attachments/MediaMessageView.swift
```diff
@@ -406,14 +425,18 @@ class MediaMessageView: UIView, OWSAudioAttachmentPlayerDelegate {
...
-        let convertedRect = fromView.convert(fromView.bounds, to:window)
-        let viewController = FullImageViewController(attachment:attachment, from:convertedRect)
-        viewController.present(from:fromViewController)
+
+        // FIXME SHARINGEXTENSION
+//        let window = UIApplication.shared.keyWindow
+//        let convertedRect = fromView.convert(fromView.bounds, to:window)
+//        let viewController = FullImageViewController(attachment:attachment, from:convertedRect)
+//        viewController.present(from:fromViewController)
+        Logger.error("\(TAG) FIXME. image tapped.")
```
charlesmchen: Hmm...

SignalMessaging/attachments/MediaMessageView.swift
```diff
@@ -456,6 +479,14 @@ class MediaMessageView: UIView, OWSAudioAttachmentPlayerDelegate {
...
         videoPlayer?.view.removeFromSuperview()
         videoPlayer = nil
-        ViewControllerUtils.setAudioIgnoresHardwareMuteSwitch(false)
+        OWSAudioAttachmentPlayer.setAudioIgnoresHardwareMuteSwitch(false)
+    }
+}
+
+extension UIImage {
+    public convenience init?(named: String, inBundleForClass someClass: AnyClass) {
+        //    let image = UIImage(named: imageName, inBundleForClass: type(of:self))
```
charlesmchen: Can we remove this line?

SignalMessaging/attachments/MediaMessageView.swift
charlesmchen: * Move to UIImage+OWS?  
* I think the other resources are certs, fonts (and localizable.strings which we don't need to worry about).

charlesmchen: Cert loading is already working fine - but I haven't looked at why.

SignalMessaging/attachments/OWSAudioAttachmentPlayer.m
```diff
@@ -48,7 +58,8 @@ - (void)dealloc
 {
     [[NSNotificationCenter defaultCenter] removeObserver:self];
 
-    [DeviceSleepManager.sharedInstance removeBlockWithBlockObject:self];
+    // FIXME SHARINGEXTENSION
+    //  [DeviceSleepManager.sharedInstance removeBlockWithBlockObject:self];
```
charlesmchen: Add methods to AppContext?

Signal/src/ViewControllers/GifPicker/GifPickerViewController.swift
michaelkirk: > Should we make that constructor private?

done!

> Which stricter rules, incidentally? Around method visibility?

yup. It was "internal", but that breaks across modules.

SignalMessaging/attachments/MediaMessageView.swift
charlesmchen: AppContext?

SignalMessaging/attachments/SendExternalFileViewController.m
```diff
@@ -176,6 +154,61 @@ - (UIView *)createTitleLabel
...
+#pragma mark - AttachmentApprovalViewControllerDelegate
+
+- (void)didApproveAttachment
+{
+    [ThreadUtil addThreadToProfileWhitelistIfEmptyContactThread:self.thread];
+    [ThreadUtil sendMessageWithAttachment:self.attachment inThread:self.thread messageSender:self.messageSender];
+
+    // FIXME SHARINGEXTENSION
+    // Show loading screen and dismiss entire share extnesion once entirely complete
+    [self.navigationController popViewControllerAnimated:YES];
```
charlesmchen: s/extnesion/extension/

Also, see `ShareViewController` which has a method for dismissing the whole share experience.  I think we need to "complete" the extension with an "item" in order to dismiss entire share UI as opposed to dismissing just our SAE.

SignalMessaging/attachments/OWSAudioAttachmentPlayer.m
michaelkirk: Right, sorry I meant to call this one out to make sure we're on the same page.

For the sharing extension context should this just be a no-op?

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -242,8 +242,9 @@ - (void)setBackgroundTaskIdentifier:(UIBackgroundTaskIdentifier)backgroundTaskId
     AssertIsOnMainThread();
 
     // Should only be sent once per operation
-    OWSAssert(_backgroundTaskIdentifier == UIBackgroundTaskInvalid);
-    OWSAssert(backgroundTaskIdentifier != UIBackgroundTaskInvalid);
+    // FIXME SHARINGEXTENSION
+//    OWSAssert(_backgroundTaskIdentifier == UIBackgroundTaskInvalid);
+//    OWSAssert(backgroundTaskIdentifier != UIBackgroundTaskInvalid);
```
charlesmchen: `OWSAssert(!CurrentAppContext() || task identifiers are valid)`

SignalShareExtension/SAELoadViewController.swift
```diff
@@ -64,6 +65,14 @@ class SAELoadViewController: UIViewController {
         activityIndicator.startAnimating()
     }
 
+    override func viewDidAppear(_ animated: Bool) {
+        // FIXME not until ready, and ideally before view appears to avoid any "loading flicker"
+        DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + 1) {
+            Logger.error("Simulating readiness...")
+            self.delegate?.shareExtensionIsReady()
```
charlesmchen: https://github.com/WhisperSystems/Signal-iOS/pull/2850 outlines my proposal here.  Basically `ShareViewController` like the main app will pop the loading view in its "ensure root view" method.

SignalShareExtension/ShareViewController.swift
```diff
@@ -182,4 +194,70 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate {
...
+    // MARK: Helpers
+
+    private func presentConversationPicker() {
+        let conversationPicker = SendExternalFileViewController()
+        buildAttachment().then { attachment -> Void in
+            conversationPicker.attachment = attachment
+            self.pushViewController(conversationPicker, animated: false)
+            Logger.info("presented conversation picker with attachment")
+        }.catch { error in
+            owsFail("\(self.logTag) building attachment failed with error: \(error)")
```
charlesmchen: Show an alert?

SignalMessaging/attachments/MediaMessageView.swift
michaelkirk: Done. Actually, since it seems like you're OK with this approach, I've moved this into `UIImage+OWS.h/m`

SignalShareExtension/ShareViewController.swift
```diff
@@ -182,4 +194,70 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate {
...
+                                            let unexpectedTypeError = ShareViewControllerError.assertionError(description: "unexpected item type: \(String(describing: previewImage))")
+                                            reject(unexpectedTypeError)
+                                            return
+                                        }
+
+                                        fulfill(image)
+        })
+
+        return promise.then { (image: UIImage) in
+            return SignalAttachment.imageAttachment(image: image, dataUTI: kUTTypeJPEG as String, filename: "from-share-extension-2345")
```
charlesmchen: Do these item providers have a filename?  Can we add a TODO to honor it?

michaelkirk:
> @michaelkirk Screenshots?

Sure, but keep in mind this is re-using existing UI, largely as an exercise to get the bulk of the requisite classes and assets moved over.  Because we're both moving a lot of things around, my goal was to do what I could to merge this quickly, and then iterate less disruptively.

It's not intended to reflect what I think the final UI should look like. 

screen shot 2017-12-06 at 10 57 44 am

screen shot 2017-12-06 at 11 37 15 am

SignalMessaging/attachments/OWSAudioAttachmentPlayer.m
charlesmchen: Yup

Podfile
michaelkirk: Made a card.

SignalMessaging/attachments/MediaMessageView.swift
michaelkirk: Based on our discussion, I've removed this method altogether since we're duplicating image assets to both bundles.

SignalMessaging/Views/ContactsViewHelper.m
michaelkirk: Added a card - we're calling openSystemSettings in potentially lots of places and should probably handle it uniformly.

SignalMessaging/attachments/OWSAudioAttachmentPlayer.m
michaelkirk: done!

SignalMessaging/attachments/SendExternalFileViewController.m
michaelkirk: This sounds good, but I'm going to leave it for the next PR.

SignalServiceKit/src/Messages/OWSMessageSender.m
michaelkirk: I guess it is that simple, huh.

I'm assuming you meant:
```
OWSAssert(!CurrentAppContext().isMainApp || _backgroundTaskIdentifier == UIBackgroundTaskInvalid);
```

SignalShareExtension/SAELoadViewController.swift
michaelkirk: Done!

SignalShareExtension/ShareViewController.swift
```diff
@@ -92,6 +95,15 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate {
...
 
         // TODO:
+        self.isNavigationBarHidden = false
+
+        // FIXME we should avoid this interstitial presentation and skip right to the content when possible
+        if (isReady) {
+            self.presentConversationPicker()
+        } else {
+            let loadViewController = SAELoadViewController(delegate:self)
+            self.pushViewController(loadViewController, animated: false)
```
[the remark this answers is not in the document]
michaelkirk: Are you talking about the animation with which we present thiew view or are you suggesting only that this should happen in the ensureRootViewController?

If it's the latter, I've now addressed it.

If it's the former... well I'm not sure what the corollary to replacing the root window in the app extension is, but it's something we can look into.

SignalShareExtension/ShareViewController.swift
michaelkirk: Done!

SignalShareExtension/ShareViewController.swift
```diff
@@ -182,4 +194,70 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate {
...
+    }
+
+    private func buildAttachment() -> Promise<SignalAttachment> {
+        guard let inputItem: NSExtensionItem = self.extensionContext?.inputItems.first as? NSExtensionItem else {
+            let error = ShareViewControllerError.assertionError(description: "no input item")
+            return Promise(error: error)
+        }
+
+        // TODO Multiple attachments. In that case I'm unclear if we'll
+        // be given multile inputItems or a signle inputItem with multiple attachments.
```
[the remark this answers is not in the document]
michaelkirk: thanks =)

SignalShareExtension/ShareViewController.swift
michaelkirk: I found a better API to use, so now we're including the filename and paved the way for non-image sharing types, though for now I've only broadened to all types of images (e.g. GIF, PNG)

SignalMessaging/attachments/MediaMessageView.swift
michaelkirk: Done!

Signal/src/util/MainAppContext.m
```diff
@@ -44,6 +49,23 @@ - (void)endBackgroundTask:(UIBackgroundTaskIdentifier)backgroundTaskIdentifier
     [UIApplication.sharedApplication endBackgroundTask:backgroundTaskIdentifier];
 }
 
+- (void)ensureSleepBlocking:(BOOL)shouldBeBlocking blockingObjects:(NSArray<id> *)blockingObjects;
```
charlesmchen: Wouldn't it be simpler to keep `DeviceSleepManager` in the main app only, forward `addBlock` and `removeBlock` using the `AppContext`, have the SAE context methods noop, done?  

I'm hoping that the `AppContext`s can remain thin "shims" that mostly do forwarding.  I recently modified the SAE context do some setup and I have mixed feelings about it.

WDYT?

SignalServiceKit/src/Util/AppContext.h
```diff
@@ -10,8 +10,16 @@ typedef void (^BackgroundTaskExpirationHandler)(void);
...
+@property (nonatomic, readonly) BOOL isMainApp;
+@property (nonatomic, readonly) BOOL isMainAppAndActive;
+
+// Whether the user is using a right-to-left language like Arabic
+@property (nonatomic, readonly) BOOL isRTL;
+
+@property (nonatomic, readonly) BOOL isRunningTests;
+
+// Useful for translating coordinates to the "entire screen"
+@property (nonatomic, readonly) UIView *keyWindow;
```
charlesmchen: nullable, right?

charlesmchen: Oh huh, this doesn't actually return a window.  Let's rename it to ... `rootReferenceView`?  

I wonder if this is a good idea.  Ideally, we'll only be using this to convert from the device coordinate system to a specific view's coordinate system.  

Currently, the SAE context is returning `self.rootViewController.view` and the main app context is returning the `keyWindow`.  I'm not sure this will yield the correct behavior in all cases.  I'm worried about:

* When app supports multiple orientations (i.e. landscape), this coordinate system conversion has to take into account rotation, etc.  
* I think there may be other cases where the device coordinate system gets "complicated", e.g. ios multitasking on iPad.

I'm not sure `self.rootViewController.view` will do the right thing in these cases.

I guess for the purposes of this release let's just make sure that this works properly in landscape.

SignalShareExtension/ShareViewController.swift
```diff
@@ -214,6 +215,8 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate {
             //                SignalsNavigationController *navigationController =
             //                    [[SignalsNavigationController alloc] initWithRootViewController:homeView];
             //                self.window.rootViewController = navigationController;
```
charlesmchen: Can we remove these commented-out lines?

SignalShareExtension/ShareViewController.swift
```diff
@@ -315,4 +318,91 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate {
...
         self.extensionContext!.completeRequest(returningItems: [], completionHandler: nil)
     }
+
+    // MARK: Helpers
+
+    private func presentConversationPicker() {
+        let conversationPicker = SendExternalFileViewController()
+        buildAttachment().then { attachment -> Void in
+            conversationPicker.attachment = attachment
+            self.pushViewController(conversationPicker, animated: true)
```
charlesmchen: I think we want to replace the root view controller (and thereby discard the loading view):

```
        self.setViewControllers([conversationPicker], animated: false)
```

SignalShareExtension/ShareViewController.swift
```diff
@@ -315,4 +318,91 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate {
...
+            Logger.info("presented conversation picker with attachment: \(attachment)")
+        }.catch { error in
+            OWSAlerts.showAlert(withTitle: NSLocalizedString("SHARE_EXTENSION_UNABLE_TO_BUILD_ATTACHMENT_ALERT_TITLE", comment: "Shown when trying to share content to a Signal user for the share extension. Followed by failure details."),
```
charlesmchen: Option: we could also replace the loading view:

```
   failureTitle = NSLocalizedString("SHARE_EXTENSION_UNABLE_TO_BUILD_ATTACHMENT_ALERT_TITLE", comment: "Shown when trying to share content to a Signal user for the share extension. Followed by failure details.")
   failureMessage = error.localizedDescription
        showErrorView(title:failureTitle, message:failureMessage)
```

Also it seems like this line should have wrapped?

charlesmchen:
We can discuss `DeviceSleepManager` and the coordinate systems issue offline if that's helpful.

SignalShareExtension/ShareViewController.swift
michaelkirk: done!

SignalShareExtension/ShareViewController.swift
michaelkirk: done!

SignalShareExtension/ShareViewController.swift
michaelkirk: I think an alert will work fine for this error case. 

I've extracted the method to be assigned to a local variable to minimize source width.

SignalServiceKit/src/Util/AppContext.h
michaelkirk: I've renamed per your suggestion.

Zoom functionality seems to work fine in landscape, portrait and when switching between the two.

PR #2852, opened 2017-12-06
Add SAE error views.

charlesmchen:
PTAL @michaelkirk 

screen shot 2017-12-06 at 1 50 52 pm

### DIFF
--- a/SignalShareExtension/SAEFailedViewController.swift
+++ b/SignalShareExtension/SAEFailedViewController.swift
@@ -1,0 +1,98 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+import UIKit
+import SignalMessaging
+import PureLayout
+
+// All Observer methods will be invoked from the main thread.
+protocol SAEFailedViewDelegate: class {
+    func shareExtensionWasCancelled()
+}
+
+class SAEFailedViewController: UIViewController {
+
+    weak var delegate: SAEFailedViewDelegate?
+
+    let failureTitle: String
+    let failureMessage: String
+
+    // MARK: Initializers and Factory Methods
+
+    init(delegate: SAEFailedViewDelegate, title: String, message: String) {
+        self.delegate = delegate
+        self.failureTitle = title
+        self.failureMessage = message
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        self.failureTitle = ""
+        self.failureMessage = ""
+        super.init(coder: aDecoder)
+    }
+
+    override func loadView() {
+        super.loadView()
+
+        self.navigationItem.leftBarButtonItem = UIBarButtonItem(barButtonSystemItem: .cancel,
+                                                                target: self,
+                                                                action: #selector(cancelPressed))
+        self.navigationItem.title = "Signal"
+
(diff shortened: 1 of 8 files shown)


SignalShareExtension/ShareViewController.swift
```diff
@@ -310,7 +334,7 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate {
         Logger.flush()
     }
 
-    // MARK: SAELoadViewDelegate
+    // MARK: SAELoadViewDelegate, SAEFailedViewDelegate
```
charlesmchen: I propose that we create a generic `SAEViewDelegate` that  can be used by all subviews to cancel or complete.  It might be nice to add a base class `SAEViewController` for all SAE view controllers that DRYs up common logic.

SignalShareExtension/SAEFailedViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+    // MARK: Initializers and Factory Methods
+
+    init(delegate: SAEFailedViewDelegate, title: String, message: String) {
+        self.delegate = delegate
+        self.failureTitle = title
+        self.failureMessage = message
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    required init?(coder aDecoder: NSCoder) {
```
michaelkirk: nit: fatalError("not implemented")

SignalShareExtension/ShareViewController.swift
```diff
@@ -51,7 +51,11 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate {
 
         isReadyForAppExtensions = OWSPreferences.isReadyForAppExtensions()
         if !isReadyForAppExtensions {
-            // TODO: Show the "You need to launch main app" view.
+            if (OWSPreferences.isRegistered()) {
```
michaelkirk: TIOLI:

I might be wrong, but I'm guessing this state is being duplicated from TSStorageManagaer so that we can show "Not Registered" immediately, rather than briefly showing "Loading..." before showing "Not Registered". Is that right?

I think in practice, if the user is not registered, the db should set up *really* quickly, so we wouldn't be waiting for long before showing "Not Registered". But maybe this produces an undesirable flicker while the "Loading..." is shown and removed. 

What I'm getting at, is that I prefer my hacks at the UI layer rather than the persistence layer. Maybe there's something we can do to address the flicker issue without duplicating state.

**Pedantry aside**, I only have one specific concern: Will a user who restores signal from a backup and launches the share extension before re-registering be in a confused state due to their out-of-sync NSUserDefaults?

The current consequences of which wouldn't be severe, but it would be easy to use this preference elsewhere in a more dangerous place.

SignalShareExtension/SAEFailedViewController.swift
charlesmchen: Done.

SignalShareExtension/ShareViewController.swift
charlesmchen: > I might be wrong, but I'm guessing this state is being duplicated from TSStorageManagaer so that we can show "Not Registered" immediately, rather than briefly showing "Loading..." before showing "Not Registered". Is that right?

Nope.  Sorry, I should have added comments here.  Done.

PR #2853, opened 2017-12-06
Charlesmchen/sae vs statics

charlesmchen:
This PR gets us to the point where we can launch the SAE twice without issue.  There's likely to be more work to be done here after the MVP of the actual UI is merged.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -25,6 +25,7 @@
 #import "VersionMigrations.h"
 #import "ViewControllerUtils.h"
 #import <AxolotlKit/SessionCipher.h>
+#import <SignalMessaging/AppSetup.h>
 #import <SignalMessaging/OWSMath.h>
 #import <SignalMessaging/OWSProfileManager.h>
 #import <SignalMessaging/SignalMessaging.h>
@@ -138,7 +139,12 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     // This block will be cleared in databaseViewRegistrationComplete.
     [DeviceSleepManager.sharedInstance addBlockWithBlockObject:self];
 
-    [self setupEnvironment];
+    [AppSetup setupEnvironment:^{
+        return SignalApp.sharedApp.callMessageHandler;
+    }
+        notificationsProtocolBlock:^{
+            return SignalApp.sharedApp.notificationsManager;
+        }];
 
     [UIUtil applySignalAppearence];
 
@@ -166,6 +172,7 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
     [self prepareScreenProtection];
 
+    // Ensure OWSContactsSyncing is instantiated.
     [OWSContactsSyncing sharedManager];
 
     [[NSNotificationCenter defaultCenter] addObserver:self
@@ -292,27 +299,6 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     }
 
     return viewController;
-}
-
-- (void)setupEnvironment
-{
-    [Environment setCurrent:[Release releaseEnvironment]];
-
-    // Encryption/Decryption mutates session state and must be synchronized on a serial queue.
-    [SessionCipher setSessionCipherDispatchQueue:[OWSDispatch sessionStoreQueue]];
-
-    TextSecureKitEnv *sharedEnv =
-        [[TextSecureKitEnv alloc] initWithCallMessageHandler:SignalApp.sharedApp.callMessageHandler
-                                             contactsManager:[Environment current].contactsManager
-                                               messageSender:[Environment current].messageSender
-                                        notificationsManager:SignalApp.sharedApp.notificationsManager
-                                              profileManager:OWSProfileManager.sharedManager];
-    [TextSecureKitEnv setSharedEnv:sharedEnv];
-
-    [[TSStorageManager sharedManager] setupDatabaseWithSafeBlockingMigrations:^{
-        [VersionMigrations runSafeBlockingMigrations];
-    }];
-    [[Environment current].contactsManager startObserving];
 }
 
 - (void)application:(UIApplication *)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken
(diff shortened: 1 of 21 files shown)


charlesmchen:

PTAL @michaelkirk

Signal/src/AppDelegate.m
```diff
@@ -56,7 +56,6 @@
 @interface AppDelegate ()
 
 @property (nonatomic) UIWindow *screenProtectionWindow;
-@property (nonatomic) OWSContactsSyncing *contactsSyncing;
```
charlesmchen: Convert OWSContactsSyncing to singleton to ensure it's only created once.

Signal/test/TestUtil.h
```diff
@@ -10,7 +10,5 @@
 #define test(expressionExpectedToBeTrue) XCTAssert(expressionExpectedToBeTrue, @"")
 #define testThrows(expressionExpectedToThrow) XCTAssertThrows(expressionExpectedToThrow, @"")
 #define testDoesNotThrow(expressionExpectedToNotThrow) expressionExpectedToNotThrow
-#define testEnv [Release unitTestEnvironment:@[]]
-#define testEnvWith(options) [Release unitTestEnvironment:(@[options])]
```
charlesmchen: Dispensed with the old Release environments.

SignalShareExtension/utils/ShareAppExtensionContext.m
```diff
@@ -29,8 +37,6 @@ - (instancetype)initWithRootViewController:(UIViewController *)rootViewControlle
 
     _rootViewController = rootViewController;
 
-    OWSSingletonAssert();
```
charlesmchen: `OWSSingletonAssert` is turning out to be really useful.  I'd like to do something similar for Swift.  I have an idea...

SignalMessaging/environment/Release.m
```diff
@@ -14,56 +14,25 @@ @implementation Release
 
 + (Environment *)releaseEnvironment
 {
```
michaelkirk: 🔥 🔥 🔥 🔥

SignalShareExtension/ShareViewController.swift
```diff
@@ -77,10 +76,7 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate, S
 
         // We don't need to use "screen protection" in the SAE.
 
-        contactsSyncing = OWSContactsSyncing(contactsManager:Environment.current().contactsManager,
-                                             identityManager:OWSIdentityManager.shared(),
-                                             messageSender:Environment.current().messageSender,
-                                             profileManager:OWSProfileManager.shared())
+        OWSContactsSyncing.sharedManager()
```
michaelkirk: nit: This reads like a no-op. But we're actually calling this getter here for it's side affects.  Can you add a comment, to lessen the chances that it get's inadvertently shuffled around.

SignalShareExtension/utils/ShareAppExtensionContext.h
```diff
@@ -6,12 +6,15 @@
...
 NS_ASSUME_NONNULL_BEGIN
 
+// This is _NOT_ a singleton and will be instantiated each time that the SAE is used.
 @interface ShareAppExtensionContext : NSObject <AppContext>
 
 - (instancetype)init NS_UNAVAILABLE;
 
 - (instancetype)initWithRootViewController:(UIViewController *)rootViewController;
 
+- (void)setupEnvironment;
```
michaelkirk: /me puts on my architect astronaut suit  




"environment setup" entails a lot of things...

Much of the logic which sets up our environment is duplicated between AppDelegate and ShareViewController, with some divergence (currently documented as comments in the ShareViewController where it diverges).

That all makes pretty good sense, and is likely accurate given that we just copy and pasted and hacked and slashed. But I'm concerned that this is not going to be a good long term approach as this logic continues to change, we will be required to make sure to make changes in both files, and keep comments referencing intentionally missing code up to date in order for it to continue to make as much sense as it makes now.

**tldr; I think app setup should be extracted to a place where the shared logic can truly be shared and the divergences are implemented via inheritance or protocol conformance. (e.g. an AppSetup class)**

As I said earlier, I think AppContext should serve only as an extension-compatible adapter-pattern around UIApplication, and we should try to resist it becoming a junk drawer for helper methods needing to know about lots of other classes.

But I'm less concerned with precisely where shared "environment setup" logic lives, and mostly just want it *to be shared*. e.g. if you like this approach of putting more logic into the AppContext, maybe we could use swift protocol extensions to share code, or we could have AppContext be a base class instead of a protocol. IMO that would be preferable to having this touchy code duplicated.

image

SignalMessaging/environment/Environment.m
```diff
@@ -37,7 +38,7 @@ + (Environment *)current
 
 + (void)setCurrent:(Environment *)environment
 {
-    OWSAssert(!sharedEnvironment);
```
michaelkirk: Can you leave a comment similar to the one in AppContext.m about why we (somewhat surprisingly) expect to set the environment multiple times in the context of SAE?

Signal/src/AppDelegate.m
michaelkirk: 👍

SignalMessaging/environment/Environment.m
charlesmchen: Done.

SignalShareExtension/ShareViewController.swift

SignalShareExtension/utils/ShareAppExtensionContext.h
charlesmchen: I agree that we want to:

* DRY
* Avoid god objects / "junk drawers".

Note that this logic isn't quite the same between the two apps.  Our notifications manager and call message handler are quite different in the SAE - they fail when used.  

I'm trying to avoid dragging the main app's notifications manager and call message handler into `SignalMessaging`: one requires WebRTC etc. and the other uses non-SAE-safe API.  So if we used an `AppSetup` class, we have to inject factory blocks for these - that feels like over-engineering.

This relates to a similar question I've been thinking about: can we do better by replacing `Environment` and `TextSecureKitEnv` (and perhaps our "shared" singletons) with a global "singleton registry".  

* We could easily mock/modify singletons in our tests.
* We would be more explicit about dependencies and creation/kickoff order.
* It might make it possible to let us explicitly teardown singletons - which might be quite valuable in the SAE.  We'd need to specify teardown order/procedure.

Anyhow, I've taken your suggestion and created an `AppSetup` class but it feels over-engineered to me.  I'd like to see if we can come up with a better approach.

PR #2855, opened 2017-12-07
Charlesmchen/swift singletons

charlesmchen:
Asserts that we don't instantiate Swift singletons twice. A Swift analogue of `OWSSingletonAssert()`.  Newly relevant with the "statics persist" issue in SAE.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeViewController.m
@@ -110,6 +110,7 @@ typedef NS_ENUM(NSInteger, CellState) { kArchiveState, kInboxState };
     _blockingManager = [OWSBlockingManager sharedManager];
     _blockedPhoneNumberSet = [NSSet setWithArray:[_blockingManager blockedPhoneNumbers]];
 
+    // Ensure ExperienceUpgradeFinder has been initialized.
     ExperienceUpgradeFinder.sharedManager;
 
     [[NSNotificationCenter defaultCenter] addObserver:self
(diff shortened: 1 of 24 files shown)


Signal/src/ViewControllers/HomeViewController.m
```diff
@@ -111,7 +110,7 @@ - (void)commonInit
     _blockingManager = [OWSBlockingManager sharedManager];
     _blockedPhoneNumberSet = [NSSet setWithArray:[_blockingManager blockedPhoneNumbers]];
 
-    _experienceUpgradeFinder = [ExperienceUpgradeFinder new];
+    ExperienceUpgradeFinder.sharedManager;
```
michaelkirk: This reads like a no-op. Why are we calling the sharedManager here and not doing anything with it? Can you either delete this line or leave a comment about why this line is here.

TIOLI: Rather than sprinkling references to singletons all over the code base, hiding it behind either a getter or an assignable property makes it easier for future dependency stubbing.

e.g. 

either: 

```
_experienceUpgradeFinder =  ExperienceUpgradeFinder.sharedManager
```

or 

```
- (ExperienceUpgradeFinder *)experienceUpgradeFinder
{
    return ExperienceUpgradeFinder.sharedManager;
}
```

charlesmchen: What do you think of my proposal for a singleton registry?  It's the best way I know of to safely and easily manage singletons...

As for why this is here, for now I'm just trying to make sure that singletons are still constructed in the same order they used to be.  In the long run, I'd like to do something quite different.

PR #2856, opened 2017-12-07
Convert logTag to property.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Util/NSObject+OWS.h
+++ b/SignalServiceKit/src/Util/NSObject+OWS.h
@@ -8,9 +8,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - Logging
 
-+ (NSString *)logTag;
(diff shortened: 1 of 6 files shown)


SignalServiceKit/src/Util/NSObject+OWS.h
```diff
@@ -8,9 +8,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - Logging
 
-+ (NSString *)logTag;
+@property (nonatomic) NSString *logTag;
 
-- (NSString *)logTag;
++ (NSString *)logTag;
```
michaelkirk: does this mean class methods in swift will still need to do `self.logTag()`

charlesmchen: Good q - this PR fixes the instance usages.

PR #2859, opened 2017-12-07
Modal conversation picker, hide loading screen when possible

michaelkirk:
The first usable screen is the conversation picker, that's the first
thing we want to show the user, and the modal presentation feels like
the right way to introduce this new context.

Long load times should be the exception, not the normal flow, so we
delay it's presentation in hopes that it will generally never be seen.

PTAL @charlesmchen

### DIFF
--- a/SignalShareExtension/ShareViewController.swift
+++ b/SignalShareExtension/ShareViewController.swift
@@ -23,7 +23,6 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate, S
         // We can't show the conversation picker until the DB is set up.
         // Normally this will only take a moment, so rather than flickering and then hiding the loading screen
         // We start as invisible, and only fade it in if it's going to take a while
-        self.view.isHidden = true
         self.view.alpha = 0
         UIView.animate(withDuration: 0.1, delay: 0.5, options: [.curveEaseInOut], animations: {
             self.view.alpha = 1
@@ -288,7 +287,7 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate, S
         // ensure view is visible.
         self.view.layer.removeAllAnimations()
         UIView.animate(withDuration: 0.1, delay: 0, options: [.curveEaseInOut], animations: {
-            self.view.isHidden = false
+
             self.view.alpha = 1
         }, completion: nil)
 
--- a/SignalShareExtension/ShareViewController.swift
+++ b/SignalShareExtension/ShareViewController.swift
@@ -20,6 +20,12 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate, S
 
         Logger.debug("\(self.logTag) \(#function)")
 
+        // We can't show the conversation picker until the DB is set up.
+        // Normally this will only take a moment, so rather than flickering and then hiding the loading screen
+        // We start with as invisible, and only fade it in if it's going to take a while
+        self.view.isHidden = true
+        self.view.alpha = 0
+
         // This should be the first thing we do.
         let appContext = ShareAppExtensionContext(rootViewController:self)
         SetCurrentAppContext(appContext)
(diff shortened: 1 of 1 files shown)


SignalShareExtension/ShareViewController.swift
```diff
@@ -322,6 +327,14 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate, S
...
 
         super.viewDidAppear(animated)
+
+        // We can't show the conversation picker until the DB is set up.
+        // Normally this will only take a moment, so rather than flickering and then hiding the loading screen
+        // We start with as invisible, and only fade it in if it's going to take a while
+        self.view.isHidden = false
+        UIView.animate(withDuration: 0.1, delay: 0.5, options: [.curveEaseInOut], animations: {
+            self.view.alpha = 1
+        }, completion: nil)
```
charlesmchen: I think you've got a race here, because `presentConversationPicker()`  could be called before or after this animation.  There's a bunch of possible solutions, but I suggest that you:

* Move this animation to loadView() to ensure the animation is always started first.
* `self.view.layer.removeAllAnimations()` everywhere we do `self.present(...)` to ensure that we cancel the animation if necessary when presenting our picker _or_ an error view.

michaelkirk: What is the harm of self.view controller becoming "visible" behind the conversationPicker?

michaelkirk: Ah, I guess it could be weird to see a blue screen fade in behind the conversation picker as the picker is being presented?

I've addressed that with another commit. PTAL @charlesmchen

SignalShareExtension/ShareViewController.swift
```diff
@@ -20,6 +20,15 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate, S
     override open func loadView() {
         super.loadView()
 
+        // We can't show the conversation picker until the DB is set up.
+        // Normally this will only take a moment, so rather than flickering and then hiding the loading screen
+        // We start as invisible, and only fade it in if it's going to take a while
+        self.view.isHidden = true
```
charlesmchen: @michaelkirk This is kind of weird but some of my comments (from an hour? ago) on this PR got lost.  I'm 99% sure I made them the right way.  Anyway, I'll try to recreate them now.

charlesmchen: Why use `isHidden` to hide?  Isn't that redundant with setting the alpha to zero?  Also, won't it interfere with the alpha animation below?

SignalShareExtension/ShareViewController.swift
```diff
@@ -295,9 +304,15 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate, S
     }
 
     private func showErrorView(title: String, message: String) {
+        // ensure view is visible.
+        self.view.layer.removeAllAnimations()
+        UIView.animate(withDuration: 0.1, delay: 0, options: [.curveEaseInOut], animations: {
```
charlesmchen: Why fade in the error view?  Why not just "snap" it in?

michaelkirk: Totally subjective - but to me snapping feels jarring. Is this a blocking concern?

SignalShareExtension/ShareViewController.swift
michaelkirk: You're right, it's not necessary. I've remove it. PTAL @charlesmchen

SignalShareExtension/ShareViewController.swift
charlesmchen: Okay. We snap in the recipient picker, right?  I don't understand why we'd snap in one and animate in the other.  Not a blocking concern though.

michaelkirk: No we animate it in from the bottom, as a modal.

PR #2861, opened 2017-12-07
Mkirk/dismiss share view

michaelkirk:
This PR is now ready to review @charlesmchen


Signal/Signal-Info.plist
```diff
@@ -15,12 +15,6 @@
 			<string>Signal</string>
```
charlesmchen: Shouldn't we remove the whole `CFBundleDocumentTypes` section?

PR #2863, opened 2017-12-07
Fix order of operations in SAE setup.

charlesmchen:
PTAL @michaelkirk


SignalShareExtension/ShareViewController.swift
```diff
@@ -71,6 +65,12 @@ public class ShareViewController: UINavigationController, SAELoadViewDelegate, S
             return
         }
 
+        AppSetup.setupEnvironment({
```
charlesmchen: The fix: We shouldn't set up our environment until _after_ we've consulted `isReadyForAppExtensions`.

michaelkirk: Good point! Can you leave a comment to that affect?

PR #2864, opened 2017-12-07
Don't notify user about identity changes in SAE.

charlesmchen:
Is this right?

PTAL @michaelkirk


michaelkirk-signal:
This is no longer relevant - we only poll for profile data in the main app and we notify via a UIAlert when identity changes are learned via share extension message sending.

PR #2866, opened 2017-12-08
Move some work in ContactsUpdater off the main thread.

charlesmchen:
PTAL @michaelkirk


SignalServiceKit/src/Contacts/ContactsUpdater.m
```diff
@@ -202,7 +209,8 @@ - (void)contactIntersectionWithSet:(NSSet<NSString *> *)idSet
                       }
                   }];
 
-              success([NSSet setWithArray:attributesForIdentifier.allKeys]);
+                  success([NSSet setWithArray:attributesForIdentifier.allKeys]);
```
michaelkirk: shouldn't we call this success handler on main?

Imperiopolis-Signal:

PR #2869, opened 2017-12-08
Keep header affixed to navbar.

michaelkirk:
Move search bar outside of table content.

Fixes issues like this:
img_0175

PTAL @charlesmchen

### DIFF
--- a/SignalMessaging/contacts/SelectThreadViewController.m
+++ b/SignalMessaging/contacts/SelectThreadViewController.m
@@ -83,18 +83,22 @@ NS_ASSUME_NONNULL_BEGIN
     [searchBar sizeToFit];
 
     UIView *header = [self.selectThreadViewDelegate createHeaderWithSearchBar:searchBar];
+    if (!header) {
+        header = searchBar;
+    }
+    [self.view addSubview:header];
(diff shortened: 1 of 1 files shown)


SignalMessaging/contacts/SelectThreadViewController.m
```diff
@@ -83,18 +83,22 @@ - (void)createViews
...
+    [header setContentHuggingVerticalHigh];
 
     // Table
     _tableViewController = [OWSTableViewController new];
     _tableViewController.delegate = self;
     [self.view addSubview:self.tableViewController.view];
     [_tableViewController.view autoPinWidthToSuperview];
-    [_tableViewController.view autoPinToTopLayoutGuideOfViewController:self withInset:0];
-    if (header) {
-        _tableViewController.tableView.tableHeaderView = header;
```
charlesmchen: We're using `tableHeaderView` in other views too ... should we fix those as well?

michaelkirk: Good question!

I see two other places we're using it.

1. in the compose view, I think it actually looks good as is.
2. in the country code picker, it looks a little weird, but not as egregious as this. I've made a card to touch that up.

PR #2870, opened 2017-12-08
Charlesmchen/profile manager concurrency

charlesmchen:
Even though we've fixed all known deadlocks in profile manager, I feel like the use of `@synchronized` there is an anti-pattern.  We were using it for a few different reasons in profile manger - I've migrated to other approaches, such as: using the "update with..." pattern.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ProfileViewController.m
+++ b/Signal/src/ViewControllers/ProfileViewController.m
@@ -577,6 +577,7 @@ NSString *const kProfileView_LastPresentedDate = @"kProfileView_LastPresentedDat
 
 - (void)avatarDidChange:(UIImage *)image
 {
+    OWSAssert([NSThread isMainThread]);
     OWSAssert(image);
 
     self.avatar = [image resizedImageToFillPixelSize:CGSizeMake(kOWSProfileManager_MaxAvatarDiameter,
--- a/Signal/src/environment/ExperienceUpgrades/ExperienceUpgrade.swift
+++ b/Signal/src/environment/ExperienceUpgrades/ExperienceUpgrade.swift
@@ -54,7 +54,7 @@ class ExperienceUpgrade: TSYapDatabaseObject {
         // these models in a "change log" archive.
         if propertyKey == "title" || propertyKey == "body" || propertyKey == "image" {
             return MTLPropertyStorageNone
-        } else if propertyKey == "uniqueId" || propertyKey == "seenAt" || propertyKey == "hasEverBeenSaved" {
+        } else if propertyKey == "uniqueId" || propertyKey == "seenAt" {
             return super.storageBehaviorForProperty(withKey: propertyKey)
         } else {
             // Being conservative here in case we rename a property.
--- a/Signal/src/environment/ExperienceUpgrades/ExperienceUpgrade.swift
+++ b/Signal/src/environment/ExperienceUpgrades/ExperienceUpgrade.swift
@@ -54,7 +54,7 @@ class ExperienceUpgrade: TSYapDatabaseObject {
         // these models in a "change log" archive.
         if propertyKey == "title" || propertyKey == "body" || propertyKey == "image" {
             return MTLPropertyStorageNone
-        } else if propertyKey == "uniqueId" || propertyKey == "seenAt" {
+        } else if propertyKey == "uniqueId" || propertyKey == "seenAt" || propertyKey == "hasEverBeenSaved" {
             return super.storageBehaviorForProperty(withKey: propertyKey)
         } else {
             // Being conservative here in case we rename a property.
--- a/SignalMessaging/profiles/OWSProfileManager.h
+++ b/SignalMessaging/profiles/OWSProfileManager.h
@@ -6,12 +6,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-extern NSString *const kNSNotificationName_LocalProfileDidChange;
-extern NSString *const kNSNotificationName_OtherUsersProfileWillChange;
-extern NSString *const kNSNotificationName_OtherUsersProfileDidChange;
 extern NSString *const kNSNotificationName_ProfileWhitelistDidChange;
-extern NSString *const kNSNotificationKey_ProfileRecipientId;
-extern NSString *const kNSNotificationKey_ProfileGroupId;
 
 extern const NSUInteger kOWSProfileManager_NameDataLength;
 extern const NSUInteger kOWSProfileManager_MaxAvatarDiameter;
--- a/SignalMessaging/profiles/OWSProfileManager.h
+++ b/SignalMessaging/profiles/OWSProfileManager.h
@@ -32,7 +32,7 @@ extern const NSUInteger kOWSProfileManager_MaxAvatarDiameter;
 - (OWSAES256Key *)localProfileKey;
 // localUserProfileExists is true if there is _ANY_ local profile.
 - (BOOL)localProfileExists;
-// localUserProfileExists is true if there is a local profile with a name or avatar.
+// hasLocalProfile is true if there is a local profile with a name or avatar.
 - (BOOL)hasLocalProfile;
 - (nullable NSString *)localProfileName;
 - (nullable UIImage *)localProfileAvatarImage;
--- a/SignalMessaging/profiles/OWSProfileManager.h
+++ b/SignalMessaging/profiles/OWSProfileManager.h
@@ -30,6 +30,9 @@ extern const NSUInteger kOWSProfileManager_MaxAvatarDiameter;
 
 // These two methods should only be called from the main thread.
 - (OWSAES256Key *)localProfileKey;
+// localUserProfileExists is true if there is _ANY_ local profile.
+- (BOOL)localProfileExists;
+// localUserProfileExists is true if there is a local profile with a name or avatar.
 - (BOOL)hasLocalProfile;
 - (nullable NSString *)localProfileName;
 - (nullable UIImage *)localProfileAvatarImage;
--- a/SignalMessaging/profiles/OWSProfileManager.m
+++ b/SignalMessaging/profiles/OWSProfileManager.m
@@ -5,6 +5,7 @@
 #import "OWSProfileManager.h"
 #import "Environment.h"
 #import "NSString+OWS.h"
+#import "OWSUserProfile.h"
 #import "UIImage+OWS.h"
 #import <AFNetworking/AFNetworking.h>
 #import <SignalMessaging/SignalMessaging-Swift.h>
@@ -32,69 +33,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-// UserProfile properties may be read from any thread, but should
-// only be mutated when synchronized on the profile manager.
-@interface UserProfile : TSYapDatabaseObject
-
-@property (atomic, readonly) NSString *recipientId;
-@property (atomic, nullable) OWSAES256Key *profileKey;
-@property (atomic, nullable) NSString *profileName;
-@property (atomic, nullable) NSString *avatarUrlPath;
-// This filename is relative to OWSProfileManager.profileAvatarsDirPath.
-@property (atomic, nullable) NSString *avatarFileName;
-
-// This should reflect when either:
-//
-// * The last successful update finished.
-// * The current in-flight update began.
-@property (atomic, nullable) NSDate *lastUpdateDate;
-
-- (instancetype)init NS_UNAVAILABLE;
-
-@end
-
-#pragma mark -
-
-@implementation UserProfile
-
-@synthesize profileName = _profileName;
-
-- (instancetype)initWithRecipientId:(NSString *)recipientId
-{
-    self = [super initWithUniqueId:recipientId];
-
-    if (!self) {
-        return self;
-    }
-
-    OWSAssert(recipientId.length > 0);
-    _recipientId = recipientId;
-
-    return self;
-}
-
-- (nullable NSString *)profileName
-{
-    @synchronized(self)
-    {
-        return _profileName;
-    }
-}
-
-- (void)setProfileName:(nullable NSString *)profileName
-{
-    @synchronized(self)
-    {
-        _profileName = [profileName ows_stripped];
-    }
-}
-
-@end
-
-#pragma mark -
-
-NSString *const kLocalProfileUniqueId = @"kLocalProfileUniqueId";
-
 NSString *const kNSNotificationName_LocalProfileDidChange = @"kNSNotificationName_LocalProfileDidChange";
 NSString *const kNSNotificationName_OtherUsersProfileWillChange = @"kNSNotificationName_OtherUsersProfileWillChange";
 NSString *const kNSNotificationName_OtherUsersProfileDidChange = @"kNSNotificationName_OtherUsersProfileDidChange";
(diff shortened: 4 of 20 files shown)


SignalMessaging/profiles/OWSProfileManager.m
```diff
@@ -32,69 +33,6 @@
...
-// * The current in-flight update began.
-@property (atomic, nullable) NSDate *lastUpdateDate;
-
-- (instancetype)init NS_UNAVAILABLE;
-
-@end
-
-#pragma mark -
-
-@implementation UserProfile
```
charlesmchen: Renamed `UserProfile` to `OWSUserProfile` and made it "all atomic + update with..."

SignalMessaging/profiles/OWSProfileManager.m
```diff
@@ -118,19 +56,12 @@ @interface OWSProfileManager ()
 @property (nonatomic, readonly) OWSIdentityManager *identityManager;
 
 // This property can be accessed on any thread, while synchronized on self.
-@property (nonatomic, readonly) UserProfile *localUserProfile;
-// This property can be accessed on any thread, while synchronized on self.
-@property (atomic, nullable) UIImage *localCachedAvatarImage;
```
charlesmchen: Just use `profileAvatarImageCache` to cache all profile avatars, including the local avatar.  Simpler.

SignalMessaging/profiles/OWSProfileManager.m
```diff
@@ -217,46 +146,11 @@ - (OWSIdentityManager *)identityManager
 
 #pragma mark - User Profile Accessor
 
-// This method can be safely called from any thread.
-- (UserProfile *)getOrBuildUserProfileForRecipientId:(NSString *)recipientId
```
charlesmchen: Moved this method to `OWSUserProfile`.

SignalMessaging/profiles/OWSProfileManager.m
```diff
@@ -217,121 +141,35 @@ - (OWSIdentityManager *)identityManager
...
-    if (!instance) {
-        instance = [[UserProfile alloc] initWithRecipientId:recipientId];
-    }
-
-    OWSAssert(instance);
-
-    return instance;
-}
-
-- (void)saveUserProfile:(UserProfile *)userProfile
```
charlesmchen: `saveUserProfile` has been superceded by "update with..." methods.

SignalMessaging/profiles/OWSProfileManager.m
```diff
@@ -217,121 +141,35 @@ - (OWSIdentityManager *)identityManager
...
-    __unused UserProfile *profile = [self localUserProfile];
+    __unused OWSUserProfile *profile = [self localUserProfile];
 }
 
 #pragma mark - Local Profile
 
-- (UserProfile *)localUserProfile
+- (OWSUserProfile *)localUserProfile
 {
     @synchronized(self)
```
charlesmchen: The only place we still sync on `self`.

SignalMessaging/profiles/OWSProfileManager.m
```diff
@@ -718,62 +518,36 @@ - (void)clearProfileWhitelist
...
-            [self.userProfileWhitelistCache removeAllObjects];
-            [self.groupProfileWhitelistCache removeAllObjects];
-
-            [self.dbConnection purgeCollection:kOWSProfileManager_UserWhitelistCollection];
-            [self.dbConnection purgeCollection:kOWSProfileManager_GroupWhitelistCollection];
-            OWSAssert(0 == [self.dbConnection numberOfKeysInCollection:kOWSProfileManager_UserWhitelistCollection]);
-            OWSAssert(0 == [self.dbConnection numberOfKeysInCollection:kOWSProfileManager_GroupWhitelistCollection]);
-        }
-    });
+    [self.dbConnection asyncReadWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
```
charlesmchen: Use one big transaction.   Also, I've started using async writes throughout the profile logic.

SignalMessaging/profiles/OWSProfileManager.m
```diff
@@ -1029,131 +768,123 @@ - (nullable UIImage *)profileAvatarForRecipientId:(NSString *)recipientId
...
+                if (decryptedData) {
+                    BOOL success = [decryptedData writeToFile:filePath atomically:YES];
+                    if (success) {
+                        image = [UIImage imageWithContentsOfFile:filePath];
+                    }
                 }
-                destination:^NSURL *_Nonnull(NSURL *_Nonnull targetPath, NSURLResponse *_Nonnull response) {
-                    return [NSURL fileURLWithPath:tempFilePath];
+
+                @synchronized(self.currentAvatarDownloads)
```
charlesmchen: There's still a couple of properties we need to synchronize access to - but we now do that narrowly and sync on those objects themselves which is safer.

SignalMessaging/profiles/OWSUserProfile.h
```diff
@@ -0,0 +1,90 @@
...
+                                           dbConnection:(YapDatabaseConnection *)dbConnection;
+
+#pragma mark - Update With... Methods
+
+- (void)updateWithProfileName:(nullable NSString *)profileName
+                   profileKey:(OWSAES256Key *)profileKey
+                avatarUrlPath:(nullable NSString *)avatarUrlPath
+               avatarFileName:(nullable NSString *)avatarFileName
+                 dbConnection:(YapDatabaseConnection *)dbConnection
+                   completion:(nullable OWSUserProfileCompletion)completion;
```
charlesmchen: The problem with the "update with..." pattern is that if you want to modify just a subset of the properties, you need to generate a separate method for each permutation of possibly modified properties.

SignalMessaging/profiles/OWSUserProfile.m
```diff
@@ -0,0 +1,297 @@
...
+
+@implementation OWSUserProfile
+
++ (OWSUserProfile *)getOrBuildUserProfileForRecipientId:(NSString *)recipientId
+                                           dbConnection:(YapDatabaseConnection *)dbConnection
+{
+    OWSAssert(recipientId.length > 0);
+
+    __block OWSUserProfile *userProfile;
+    [dbConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
```
charlesmchen: Since these entities are only ever accessed using the profile manager's db connection, we don't need to worry about staleness.

Signal/src/ViewControllers/ProfileViewController.m
```diff
@@ -577,6 +577,7 @@ - (NSString *)avatarActionSheetTitle
 
 - (void)avatarDidChange:(UIImage *)image
 {
+    OWSAssert([NSThread isMainThread]);
```
charlesmchen: I fixed an unrelated concurrency issue in the avatar helper while testing these changes.

Signal/src/environment/ExperienceUpgrades/ExperienceUpgrade.swift
```diff
@@ -54,7 +54,7 @@ class ExperienceUpgrade: TSYapDatabaseObject {
         // these models in a "change log" archive.
         if propertyKey == "title" || propertyKey == "body" || propertyKey == "image" {
             return MTLPropertyStorageNone
-        } else if propertyKey == "uniqueId" || propertyKey == "seenAt" {
+        } else if propertyKey == "uniqueId" || propertyKey == "seenAt" || propertyKey == "hasEverBeenSaved" {
```
charlesmchen: I finally added a `hasEverBeenSaved` property to `TSYapDatabaseObject`.  Let's discuss!

SignalShareExtension/ShareViewController.swift
```diff
@@ -232,10 +232,14 @@ public class ShareViewController: UINavigationController, ShareViewDelegate, SAE
...
         Logger.info("Presenting initial root view controller")
 
-        if TSAccountManager.isRegistered() {
-            presentConversationPicker()
-        } else {
+        if !TSAccountManager.isRegistered() {
             showNotRegisteredView()
+        } else if !OWSProfileManager.shared().localProfileExists() {
+            // This is a rare edge case, but we want to ensure that the user
+            // is has already saved their local profile key in the main app.
```
charlesmchen: Being a little bit more cautious here.

SignalMessaging/profiles/OWSUserProfile.m
```diff
@@ -0,0 +1,507 @@
...
+                postNotificationNameAsync:kNSNotificationName_OtherUsersProfileDidChange
+                                   object:nil
+                                 userInfo:@{
+                                     kNSNotificationKey_ProfileRecipientId : self.recipientId,
+                                 }];
+        }
+    });
+}
+
+- (void)updateWithProfileName:(nullable NSString *)profileName
```
charlesmchen: The `update with...` methods are incredibly repetitious but I don't see a good way to DRY them up.

We could `[MTLModel dictionaryValue]` to generate "before" and "after" snapshots and compare those, but we'd have to assume that we'll never run afoul of the class of "stale YDB model" bugs which the "update with..." pattern was intended to address.  Since we only use a single dedicated connection for `OWSUserProfile`, we should be safe, but it feels a little dangerous.

charlesmchen: Okay, I talked myself into taking the "model diffing" approach.

SignalMessaging/profiles/OWSProfileManager.m
michaelkirk: 👍

SignalMessaging/profiles/OWSProfileManager.m
```diff
@@ -787,87 +566,70 @@ - (void)addUsersToProfileWhitelist:(NSArray<NSString *> *)recipientIds
...
-            return [value boolValue];
-        }
-
-        value =
-            @([self.dbConnection hasObjectForKey:recipientId inCollection:kOWSProfileManager_UserWhitelistCollection]);
-        self.userProfileWhitelistCache[recipientId] = value;
-        return [value boolValue];
-    }
+    __block BOOL result = NO;
+    [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
```
michaelkirk: Getting rid of the caches simplifies things. I guess it was a premature optimization?

SignalMessaging/profiles/OWSProfileManager.m
```diff
@@ -896,19 +658,15 @@ - (BOOL)isGroupIdInProfileWhitelist:(NSData *)groupId
...
-        }
+    NSString *groupIdKey = [groupId hexadecimalString];
 
-        value = @(nil !=
-            [self.dbConnection objectForKey:groupIdKey inCollection:kOWSProfileManager_GroupWhitelistCollection]);
-        self.groupProfileWhitelistCache[groupIdKey] = value;
-        return [value boolValue];
-    }
+    __block BOOL result = NO;
+    [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
```
michaelkirk: I think we want `readWithBlock:` here

SignalMessaging/profiles/OWSUserProfile.m
```diff
@@ -0,0 +1,347 @@
...
+@property (atomic, nullable) NSString *avatarFileName;
+@property (atomic, nullable) NSDate *lastUpdateDate;
+
+@end
+
+#pragma mark -
+
+@implementation OWSUserProfile
+
++ (OWSUserProfile *)getOrBuildUserProfileForRecipientId:(NSString *)recipientId
```
michaelkirk: We should define `+(NSString *)collection` to be the legacy class name so users don't have to re-build their profile key collection from scratch.

SignalMessaging/profiles/OWSProfileManager.m
michaelkirk: I think we want `readWithBlock:` here.

SignalMessaging/profiles/OWSUserProfile.h
michaelkirk: The only thing I could think of would be to just expose a method that takes the change block, but we'd be putting more responsibility on the caller to do the right thing.

I mean, I guess all the painful work of writing it has been done now anyway. =)

SignalShareExtension/ShareViewController.swift
michaelkirk: 👍

SignalMessaging/profiles/OWSUserProfile.h
charlesmchen: > The only thing I could think of would be to just expose a method that takes the change block, but we'd be putting more responsibility on the caller to do the right thing.

That wouldn't even work since the setters are (by design) private to prevent non-updateWith... access

SignalMessaging/profiles/OWSProfileManager.h
```diff
@@ -35,6 +30,9 @@ extern const NSUInteger kOWSProfileManager_MaxAvatarDiameter;
 
 // These two methods should only be called from the main thread.
 - (OWSAES256Key *)localProfileKey;
+// localUserProfileExists is true if there is _ANY_ local profile.
+- (BOOL)localProfileExists;
+// localUserProfileExists is true if there is a local profile with a name or avatar.
```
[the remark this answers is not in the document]
charlesmchen: Good catch.

SignalMessaging/profiles/OWSProfileManager.m
```diff
@@ -118,19 +51,12 @@ @interface OWSProfileManager ()
...
-@property (atomic, nullable) UIImage *localCachedAvatarImage;
+@property (atomic, readonly) OWSUserProfile *localUserProfile;
 
-// These caches are lazy-populated.  The single point of truth is the database.
-//
-// This property can be accessed on any thread, while synchronized on self.
-@property (atomic, readonly) NSMutableDictionary<NSString *, NSNumber *> *userProfileWhitelistCache;
 // This property can be accessed on any thread, while synchronized on self.
-@property (atomic, readonly) NSMutableDictionary<NSString *, NSNumber *> *groupProfileWhitelistCache;
+//@property (atomic, readonly) NSCache<NSString *, UIImage *> *otherUsersProfileAvatarImageCache;
```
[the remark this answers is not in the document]
charlesmchen: Done.

SignalMessaging/profiles/OWSProfileManager.m
```diff
@@ -369,108 +200,97 @@ - (void)updateLocalProfileName:(nullable NSString *)profileName
...
+                OWSUserProfile *userProfile = self.localUserProfile;
+                OWSAssert(userProfile);
+
+                [userProfile updateWithProfileName:profileName
+                                     avatarUrlPath:avatarUrlPath
+                                    avatarFileName:avatarFileName
+                                      dbConnection:self.dbConnection
+                                        completion:^{
+                                            if (avatarFileName) {
+                                                [self updateProfileAvatarCache:avatarImage filename:avatarFileName];
```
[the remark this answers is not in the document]
charlesmchen: Nope, that method synchronizes on `profileAvatarImageCache` so it should be safe.

SignalMessaging/profiles/OWSProfileManager.m
charlesmchen: Done.

SignalMessaging/profiles/OWSProfileManager.m
charlesmchen: Done.

SignalMessaging/profiles/OWSUserProfile.m
charlesmchen: Great point.

SignalMessaging/profiles/OWSUserProfile.m
```diff
@@ -0,0 +1,347 @@
...
+    return TSYapDatabaseObject.dbReadWriteConnection;
+}
+
+- (NSString *)debugDescription
+{
+    return [NSString stringWithFormat:@"%@ %p %@ %@ %@ %@ %@ %f",
+                     self.logTag,
+                     self,
+                     self.recipientId,
+                     self.profileKey.keyData.hexadecimalString,
```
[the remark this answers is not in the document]
charlesmchen: This is only intended to be used in `verbose` (e.g. developer) logs, but sure, let's be cautious.

PR #2874, opened 2017-12-11
approval view revamp

michaelkirk:
Approval UI for Images and Video, including in-line pan and zoom functionality.

Note there is still some work to be done in the approval UI, but in practice this gets pretty far in having a usable demo.

Here's the remaining work that I know of, related to the approval UI:

1. Approval view only works with mp4 video (fixed in an upcoming PR)
2. Add caption field and restyle the send button
3. Audio Files (this might actually work, I just haven't tested yet)
4. Generic Attachments in landscape
5. Sending progress indicator
6. recipient list below caption and send button

This work is already tracked, I just wanted to make clear what this PR **wasn't** intending to address

### image

screen shot 2017-12-11 at 10 23 34 am

screen shot 2017-12-11 at 10 36 31 am

### image zoomed

screen shot 2017-12-11 at 10 23 47 am

screen shot 2017-12-11 at 10 36 39 am

### video

video-unzoomed

### video zoomed

video-zoomed

PTAL @charlesmchen


SignalMessaging/attachments/AttachmentApprovalViewController.swift
```diff
@@ -175,6 +228,90 @@ public class AttachmentApprovalViewController: OWSViewController {
...
+    private var scrollViewCenter: CGPoint {
+        let size = scrollViewVisibleSize
+        return CGPoint(x: (size.width / 2), y: (size.height / 2))
+    }
+
+    // Return scrollview size without the area overlapping with tab and nav bar.
+    private var scrollViewVisibleSize: CGSize {
+        let contentInset = scrollView.contentInset
+        let scrollViewSize = scrollView.bounds.standardized.size
+        let width = scrollViewSize.width - contentInset.left - contentInset.right
```
charlesmchen: This is personal, but I always prefer to express this kind of thing like so:

```
scrollViewSize.width - (contentInset.left + contentInset.right)
```

Operator ordering can vary between languages and people tend to make mistakes around them, so I always try to be explicit and readable using parens.

charlesmchen: TIOLI obvs

michaelkirk: Agreed this is more legible, done!

PR #2876, opened 2017-12-12
Convert non mp4 video

michaelkirk:
- Fixes the video preview/play feature when sharing from the Photos App
- Ensures shared videos are converted to compatible format when host app doesn't do it for us (e.g. iCloud Drive)
- Ensures video from the Document picker (in the main app) are mp4's.

More comments in line.

In the case that the host app hasn't converted the object for us (e.g. .mov from iCloud Drive), we hijack the existing loading screen to show a UIProgressView reflecting the export progress before the contact picker is shown.

progress-portrait

progress-landscape

PTAL @charlesmchen


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2722,50 +2731,33 @@ - (void)sendQualityAdjustedAttachmentForVideo:(NSURL *)movieURL
...
-                      [exportSession exportAsynchronouslyWithCompletionHandler:^{
-                          dispatch_async(dispatch_get_main_queue(), ^{
-                              OWSAssert([NSThread isMainThread]);
-
-                              if (modalActivityIndicator.wasCancelled) {
-                                  return;
+                      DataSource *dataSource = [DataSourcePath dataSourceWithURL:movieURL];
+                      dataSource.sourceFilename = filename;
+                      VideoCompressionResult *compressionResult =
+                          [SignalAttachment compressVideoAsMp4WithDataSource:dataSource
```
michaelkirk: parallel with the jpegCompression logic, the mp4Compression logic is now consolidated into SignalAttachment.

SignalShareExtension/ShareViewController.swift
```diff
@@ -15,21 +15,14 @@ public class ShareViewController: UINavigationController, ShareViewDelegate, SAE
...
 
     override open func loadView() {
         super.loadView()
 
         Logger.debug("\(self.logTag) \(#function)")
 
-        // We can't show the conversation picker until the DB is set up.
-        // Normally this will only take a moment, so rather than flickering and then hiding the loading screen
-        // We start as invisible, and only fade it in if it's going to take a while
-        self.view.alpha = 0
```
michaelkirk: I simplified the "Loading UI" logic. The Loading VC is now presented modally (after a short delay) and all this animation stuff went away.

SignalShareExtension/ShareViewController.swift
```diff
@@ -364,19 +364,21 @@ public class ShareViewController: UINavigationController, ShareViewDelegate, SAE
...
         self.buildAttachment().then { attachment -> Void in
             let conversationPicker = SharingThreadPickerViewController(shareViewDelegate: self)
             let navigationController = UINavigationController(rootViewController: conversationPicker)
             navigationController.isNavigationBarHidden = true
             conversationPicker.attachment = attachment
-            self.present(navigationController, animated: true, completion: nil)
+            if let presentedViewController = self.presentedViewController {
+                Logger.debug("\(self.logTag) dismissing \(presentedViewController) before presenting conversation picker")
+                self.dismiss(animated: true) {
+                    self.present(navigationController, animated: true)
```
michaelkirk: If the loading modal is presented, we dismiss it, and then immediately present the conversation picker.

This isn't a great look. Any suggestions?

SignalShareExtension/ShareViewController.swift
```diff
@@ -448,7 +450,30 @@ public class ShareViewController: UINavigationController, ShareViewDelegate, SAE
         // TODO accept other data types
         // TODO whitelist attachment types
         // TODO coerce when necessary and possible
-        return promise.then { (url: URL) -> SignalAttachment in
+        return promise.then { (itemUrl: URL) -> Promise<SignalAttachment> in
+
+            let url: URL = try {
+                // iOS converts at least some video formats (e.g. com.apple.quicktime-movie) into mp4s as part of the
```
michaelkirk: Please read this, if only to help justify the hours I spent debugging.

SignalShareExtension/SignalShareExtension-Bridging-Header.h
```diff
@@ -6,19 +6,19 @@
...
-#import "OWSLogger.h"
-#import "OWSMath.h"
-#import "OWSPreferences.h"
-#import "Release.h"
 #import "ShareAppExtensionContext.h"
-#import "UIColor+OWS.h"
-#import "UIFont+OWS.h"
-#import "UIView+OWS.h"
-#import "VersionMigrations.h"
+#import <SignalMessaging/DebugLogger.h>
```
michaelkirk: This cleanup is now irrelevant, but I left it in as I prefer using the full module path, even when not required.

SignalMessaging/attachments/SignalAttachment.swift
```diff
@@ -767,10 +781,137 @@ public class SignalAttachment: NSObject {
...
+        return toUrl
+    }
+
+    private class var videoTempPath: URL {
+        let videoDir = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent("video")
+        OWSFileSystem.ensureDirectoryExists(videoDir.path)
+        return videoDir
+    }
+
+    public class func compressVideoAsMp4(dataSource: DataSource, dataUTI: String) -> (Promise<SignalAttachment>, AVAssetExportSession?) {
```
charlesmchen: Hmm. This doesn't feel like it belongs in `SignalAttachment`.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2722,50 +2731,33 @@ - (void)sendQualityAdjustedAttachmentForVideo:(NSURL *)movieURL
...
+                                                                     dataUTI:(NSString *)kUTTypeMPEG4];
+                      [compressionResult.attachmentPromise retainUntilComplete];
+
+                      compressionResult.attachmentPromise.then(^(SignalAttachment *attachment) {
```
charlesmchen: There's two `attachment` and two `dataSource` symbols of the same type in scope, right?  How about `original...`/`compressed...` or `src...`/`dst...` prefixes?

charlesmchen: Oh sorry, misleading diff.

SignalMessaging/attachments/SignalAttachment.swift
```diff
@@ -767,10 +781,137 @@ public class SignalAttachment: NSObject {
...
+
+            guard let dataSource = DataSourcePath.dataSource(with: exportURL) else {
+                owsFail("Failed to build data source for exported video URL")
+                let attachment = SignalAttachment(dataSource : DataSourceValue.emptyDataSource(), dataUTI: dataUTI)
+                attachment.error = .couldNotConvertToMpeg4
+                fulfill(attachment)
+                return
+            }
+
+            dataSource.sourceFilename = mp4Filename
```
charlesmchen: I think you'll want to do `dataSource .setShouldDeleteOnDeallocation()` so that we eagerly clean up this temp file.  It looks like it will reside in the temp dir but video tends to be large so...

SignalMessaging/attachments/SignalAttachment.swift
```diff
@@ -767,10 +781,137 @@ public class SignalAttachment: NSObject {
...
+        guard isInputVideoValidOutputVideo(dataSource: dataSource, dataUTI: dataUTI) else {
+            // found a video which needs to be converted
+            return true
+        }
+
+        // It is a video, but it's not invalid
+        return false
+    }
+
+    private class func isInputVideoValidOutputVideo(dataSource: DataSource?, dataUTI: String) -> Bool {
```
charlesmchen: Why not just `isValidOutputVideo`?

SignalShareExtension/SAELoadViewController.swift
```diff
@@ -39,6 +68,16 @@ class SAELoadViewController: UIViewController {
         self.view.addSubview(activityIndicator)
         activityIndicator.autoCenterInSuperview()
 
+        progressView = UIProgressView(progressViewStyle: .default)
```
charlesmchen: `OWSProgressView` for consistency?  `UIProgressView` is too minimal to use in the contexts where we already use `OWSProgressView`.

charlesmchen: Note that `OWSProgressView` requires you to `autoSetDimension` its size.

SignalShareExtension/ShareViewController.swift
```diff
@@ -75,6 +68,22 @@ public class ShareViewController: UINavigationController, ShareViewDelegate, SAE
             return
         }
 
+        let loadViewController = SAELoadViewController(delegate: self)
+        self.loadViewController = loadViewController
+
+        // Don't display load screen immediately, in hopes that we can avoid it altogether.
+        after(seconds: 2).then { () -> Void in
```
charlesmchen: 2 seconds feels too long for our SAE to apparently do nothing/not appear.  2 seconds is long enough for users to try to launch the SAE again (or another share extension). I'd suggest .4 or .5 seconds.

SignalShareExtension/ShareViewController.swift
charlesmchen: `ShareViewController` is a `UINavigationController`.  Why are we creating a new `UINavigationController`?  I would just do:
```
        self.setViewControllers([conversationPicker], animated: false)
```
or `animated` could be true, but IMO our users would rather skip the animation.  People are in a hurry when they're trying to share.

SignalShareExtension/ShareViewController.swift
```diff
@@ -448,7 +450,31 @@ public class ShareViewController: UINavigationController, ShareViewDelegate, SAE
...
+                // NSItemProvider `loadItem` API.
+                // However, for some reason, AVFoundation operations such as generating a preview image and playing
+                // the url in the AVMoviePlayer fail on these converted formats until unless we first copy the media
+                // into our container. (These operations succeed when sending a non-converted mp4 (e.g. one received,
+                // saved, and resent in Signal)
+                //
+                // I don't understand why this is, and I haven't found any relevant documentation in the NSItemProvider
+                // or AVFoundation docs.
+                //
+                // I *did* verify that the size and sah256 sum of the original url matches that of the copied url.
```
charlesmchen: s/sah256/SHA256/

SignalShareExtension/ShareViewController.swift
```diff
@@ -448,7 +450,31 @@ public class ShareViewController: UINavigationController, ShareViewDelegate, SAE
...
+                // Perhaps the AVFoundation API's require some extra file system permssion we don't have in the
+                // passed through URL.
+                if MIMETypeUtil.isSupportedVideoFile(itemUrl.path) {
+                    return try SignalAttachment.copyToVideoTempDir(url: itemUrl)
```
charlesmchen: Isn't this backwards?  It reads as though you "copy to temp" if the URL _IS_ supported.  I thought the copy was only necessary if the URL wasn't supported, e.g. if the video has just been converted by iOS.

charlesmchen: It seems like we'll want to `setShouldDeleteOnDeallocation` IFF you copied to temp.

SignalShareExtension/ShareViewController.swift
```diff
@@ -459,13 +485,82 @@ public class ShareViewController: UINavigationController, ShareViewDelegate, SAE
...
 
-            let attachment = SignalAttachment.attachment(dataSource: dataSource, dataUTI: specificUTIType, imageQuality:.medium)
+            guard !SignalAttachment.isInvalidVideo(dataSource: dataSource, dataUTI: specificUTIType) else {
+                // This can happen, e.g. when sharing a quicktime-video from iCloud drive.
+
+                let (promise, exportSession) = SignalAttachment.compressVideoAsMp4(dataSource: dataSource, dataUTI: specificUTIType)
+
+                // TODO: How can we move waiting for this export to the end of the share flow rather than having to do it up front?
+                // Ideally we'd be able to start it here, and not block the UI on conversion unless there's still work to be done
+                // when the user hit's "send".
```
charlesmchen: "You can solve every problem with another level of indirection, except for the problem of too many levels of indirection."

* `SignalAttachment` is itself a layer of indirection around a `DataSource`. 
* `DataSource` is a layer of indirection (adapter pattern) around a file or NSData.

I guess we could have a `SignalAttachmentSource`, but IMO the complexity isn't worth it.

SignalShareExtension/ShareViewController.swift
```diff
@@ -459,13 +485,82 @@ public class ShareViewController: UINavigationController, ShareViewDelegate, SAE
...
+                if let exportSession = exportSession {
+                    let progressPoller = ProgressPoller(timeInterval: 0.1, ratioCompleteBlock: { return exportSession.progress })
+                    self.progressPoller = progressPoller
+                    progressPoller.startPolling()
```
charlesmchen: Shouldn't we clean the `progressPoller` variable?

SignalShareExtension/ShareViewController.swift
```diff
@@ -75,6 +68,22 @@ public class ShareViewController: UINavigationController, ShareViewDelegate, SAE
             return
         }
 
+        let loadViewController = SAELoadViewController(delegate: self)
+        self.loadViewController = loadViewController
```
charlesmchen: Don't we want to clear the `loadViewController` property?

SignalShareExtension/SignalShareExtension-Bridging-Header.h
charlesmchen: +1.  IMO we should write a script that auto-fixes these, but I just can't justify the time to do it.  But if we find ourselves fixing them by hand one-by-one, that's worse.

SignalMessaging/attachments/SignalAttachment.swift
michaelkirk: Does compressAsJpeg?

SignalMessaging/attachments/SignalAttachment.swift
michaelkirk: Great point - done!

SignalMessaging/attachments/SignalAttachment.swift
michaelkirk: I was matching the convention that we used for image output, but I can rename them both:

`isInputImageValidOutputImage` -> `isValidOutputImage`
`isInputVideoValidOutputVideo ` -> `isValidOutputVideo`

SignalShareExtension/ShareViewController.swift
michaelkirk: sure, done.

SignalShareExtension/ShareViewController.swift
michaelkirk: We need to do it for basically *any* video. It's not related to the file type - I think it's related to where the file lives on disk. I agree that it's confusing as written. I'll clarify this.

charlesmchen: Okay, just make sure you only use `setShouldDeleteOnDeallocation` for temp files - not originals.

PR #2877, opened 2017-12-12
Fixup approval view for audio and generic attachments

michaelkirk:
Based on #2876, but stands alone.

PTAL @charlesmchen 

audio-portrait

audio-landscape

file-portrait

file-landscape


michaelkirk:
> But white-on-black appearance isn't final, is it? IMO it doesn't match our white/black "bright" color palette.

You're right - it should probably change, but for expediency I made as few alterations as possible to the existing attachment view.

Also, I snuck this commit in: https://github.com/WhisperSystems/Signal-iOS/commit/6fb5990faa0d8662eaeb61392d63c04114213696

PR #2882, opened 2017-12-14
persist signal accounts

michaelkirk:
Fixes seeing an empty contact list when launching compose view too soon.
Fixes the flickering contact avatar seen when first launching Signal.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/OWSConversationSettingsViewController.m
+++ b/Signal/src/ViewControllers/OWSConversationSettingsViewController.m
@@ -171,7 +171,7 @@ NS_ASSUME_NONNULL_BEGIN
     OWSAssert([self.thread isKindOfClass:[TSContactThread class]]);
     TSContactThread *contactThread = (TSContactThread *)self.thread;
     NSString *recipientId = contactThread.contactIdentifier;
-    return [self.contactsManager signalAccountForRecipientId:recipientId] != nil;
+    return [self.contactsManager hasSignalAccountForRecipientId:recipientId];
 }
 
 #pragma mark - ContactEditingDelegate
--- a/Signal/src/ViewControllers/OWSConversationSettingsViewController.m
+++ b/Signal/src/ViewControllers/OWSConversationSettingsViewController.m
@@ -171,7 +171,7 @@ NS_ASSUME_NONNULL_BEGIN
     OWSAssert([self.thread isKindOfClass:[TSContactThread class]]);
     TSContactThread *contactThread = (TSContactThread *)self.thread;
     NSString *recipientId = contactThread.contactIdentifier;
-    return [self.contactsManager.lastKnownContactRecipientIds containsObject:recipientId];
+    return [self.contactsManager signalAccountForRecipientId:recipientId] != nil;
 }
 
 #pragma mark - ContactEditingDelegate
--- a/Signal/src/contact/OWSContactsManager.m
+++ b/Signal/src/contact/OWSContactsManager.m
@@ -23,7 +23,6 @@ NSString *const kTSStorageManager_AccountDisplayNames = @"kTSStorageManager_Acco
 NSString *const kTSStorageManager_AccountFirstNames = @"kTSStorageManager_AccountFirstNames";
 NSString *const kTSStorageManager_AccountLastNames = @"kTSStorageManager_AccountLastNames";
 NSString *const kTSStorageManager_OWSContactsManager = @"kTSStorageManager_OWSContactsManager";
-NSString *const kTSStorageManager_lastKnownContactRecipientIds = @"lastKnownContactRecipientIds";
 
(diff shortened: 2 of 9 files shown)


Signal/src/ViewControllers/OWSConversationSettingsViewController.m
```diff
@@ -171,7 +171,7 @@ - (BOOL)hasExistingContact
     OWSAssert([self.thread isKindOfClass:[TSContactThread class]]);
     TSContactThread *contactThread = (TSContactThread *)self.thread;
     NSString *recipientId = contactThread.contactIdentifier;
-    return [self.contactsManager.lastKnownContactRecipientIds containsObject:recipientId];
+    return [self.contactsManager signalAccountForRecipientId:recipientId] != nil;
```
charlesmchen: We do this a few places.  I'd make a `hasSignalAccountForRecipientId` method personally to yield more readable code, but I know many people prefer to avoid the redundancy. TIOLI

Signal/src/contact/OWSContactsManager.m
```diff
@@ -240,170 +232,76 @@ - (void)updateSignalAccounts
...
-            }];
+        // Update cached SignalAccounts on disk
+        [self.dbWriteConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
+            NSArray<NSString *> *allKeys = [transaction allKeysInCollection:[SignalAccount collection]];
+            NSMutableSet<NSString *> *orphanedKeys = [NSMutableSet setWithArray:allKeys];
+            
+            for (SignalAccount *signalAccount in signalAccounts) {
+                // TODO only save the ones that changed
+                [orphanedKeys removeObject:signalAccount.uniqueId];
+                [signalAccount saveWithTransaction:transaction];
```
charlesmchen: I thought you were going to persist the collection as a whole. Why the change of heart?

Signal/src/contact/OWSContactsManager.m
```diff
@@ -240,170 +232,76 @@ - (void)updateSignalAccounts
...
-    });
+    return [OWSProfileManager sharedManager];
 }
 
 - (NSString *_Nullable)cachedDisplayNameForRecipientId:(NSString *)recipientId
 {
     OWSAssert(recipientId.length > 0);
 
-    return self.cachedAccountNameMap[recipientId];
+    SignalAccount *signalAccount = [self signalAccountForRecipientId:recipientId];
```
charlesmchen: `SignalAccount *_Nullable signalAccount` here and below...

SignalServiceKit/src/Contacts/SignalAccount.m
```diff
@@ -46,6 +47,23 @@ - (nullable SignalRecipient *)signalRecipientWithTransaction:(YapDatabaseReadTra
...
+{
+    return _recipientId;
+}
+
+- (NSString *)displayName
+{
+    NSString *baseName = (self.contact.fullName.length > 0 ? self.contact.fullName : self.recipientId);
+
+    OWSAssert(self.hasMultipleAccountContact == (self.multipleAccountLabelText != nil));
+    NSString *displayName = (self.multipleAccountLabelText
```
charlesmchen: This isn't redundant with logic in the `OWSContactsManager` is it?

SignalServiceKit/src/Contacts/SignalAccount.m
```diff
@@ -46,6 +47,23 @@ - (nullable SignalRecipient *)signalRecipientWithTransaction:(YapDatabaseReadTra
     return [SignalRecipient recipientWithTextSecureIdentifier:self.recipientId withTransaction:transaction];
 }
 
+- (nullable NSString *)uniqueId
+{
+    return _recipientId;
```
charlesmchen: I think we might want to check for empty recipientIds wherever we create new instances of `SignalRecipient`.

Signal/src/contact/OWSContactsManager.m
michaelkirk: I started that way, but I realized there are still API's to look up the SignalRecipient by ID. If the cache wasn't been loaded when that happens we can produce the correct result by fetching them from disk directly.

SignalServiceKit/src/Contacts/SignalAccount.m
michaelkirk: This logic was moved from OWSContactsManager, but it lives only here now.

SignalServiceKit/src/Contacts/SignalAccount.m
michaelkirk: done

Signal/src/contact/OWSContactsManager.m
michaelkirk: done

PR #2886, opened 2017-12-15
Charlesmchen/attachment downloads vs background

charlesmchen:
PTAL @michaelkirk 

* Use a background task during attachment downloads.
* Use a background task during message processing.
* Create `OWSBackgroundTask` to DRY up and improve handling of background tasks.  

We often weren't handling edge cases correctly, like:

* Ensuring that completion handlers were called exactly once.
* Handling the "background task could not be completed" case.
* Ensuring that completion handlers were thread-safe.

### DIFF
--- a/SignalServiceKit/src/Util/OWSBackgroundTask.m
+++ b/SignalServiceKit/src/Util/OWSBackgroundTask.m
@@ -1,0 +1,85 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "OWSBackgroundTask.h"
+#import "AppContext.h"
+
+@interface OWSBackgroundTask ()
+
+@property (nonatomic) UIBackgroundTaskIdentifier backgroundTaskId;
+@property (nonatomic, readonly) NSString *label;
+
+@end
+
+#pragma mark -
+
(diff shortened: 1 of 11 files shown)


SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -0,0 +1,157 @@
...
+
+    if (backgroundTaskId == UIBackgroundTaskInvalid) {
+        OWSAssert(!completionBlock);
+        return;
+    }
+
+    // endBackgroundTask must be called on the main thread.
+    DispatchMainThreadSafe(^{
+        DDLogVerbose(@"%@ %@ background task completed.", logTag, label);
+        [CurrentAppContext() endBackgroundTask:backgroundTaskId];
```
michaelkirk: Should this come *after* the completion block so that we can help ensure the completion block runs before we're suspended?

(in which case you'd probably want to ensure that the completion block  is not redundantly called in the expirationHandler)

charlesmchen: I've juggled things around here a bit.  I've added some logic to copy/clear the `completionBlock` to ensure that it's called exactly once.

PR #2887, opened 2017-12-15
replace cache for migration

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/contact/OWSContactsManager.h
+++ b/Signal/src/contact/OWSContactsManager.h
@@ -55,7 +55,7 @@ extern NSString *const OWSContactsManagerSignalAccountsDidChangeNotification;
 
 // This variant will fetch system contacts if contact access has already been granted,
 // but not prompt for contact access. Also, it will always notify delegates, even if
-// contacts haven't changed, and will clear out any stale cached SignalAccount's
+// contacts haven't changed, and will clear out any stale cached SignalAccounts
 - (void)userRequestedSystemContactsRefreshWithCompletion:(void (^)(NSError *_Nullable error))completionHandler;
 
 #pragma mark - Util
--- a/Signal/src/contact/OWSContactsManager.h
+++ b/Signal/src/contact/OWSContactsManager.h
@@ -54,9 +54,9 @@ extern NSString *const OWSContactsManagerSignalAccountsDidChangeNotification;
 - (void)fetchSystemContactsOnceIfAlreadyAuthorized;
 
 // This variant will fetch system contacts if contact access has already been granted,
-// but not prompt for contact access. Also, it will always fire a notification.
-- (void)fetchSystemContactsIfAlreadyAuthorizedAndAlwaysNotifyWithCompletion:
-    (void (^)(NSError *_Nullable error))completionHandler;
+// but not prompt for contact access. Also, it will always notify delegates, even if
+// contacts haven't changed, and will clear out any stale cached SignalAccount's
+- (void)userRequestedSystemContactsRefreshWithCompletion:(void (^)(NSError *_Nullable error))completionHandler;
 
 #pragma mark - Util
 
--- a/Signal/src/contact/OWSContactsManager.h
+++ b/Signal/src/contact/OWSContactsManager.h
@@ -52,9 +52,11 @@ extern NSString *const OWSContactsManagerSignalAccountsDidChangeNotification;
 // Ensure's the app has the latest contacts, but won't prompt the user for contact
 // access if they haven't granted it.
 - (void)fetchSystemContactsOnceIfAlreadyAuthorized;
+
 // This variant will fetch system contacts if contact access has already been granted,
 // but not prompt for contact access. Also, it will always fire a notification.
-- (void)fetchSystemContactsIfAlreadyAuthorizedAndAlwaysNotify;
+- (void)fetchSystemContactsIfAlreadyAuthorizedAndAlwaysNotifyWithCompletion:
+    (void (^)(NSError *_Nullable error))completionHandler;
 
 #pragma mark - Util
 
--- a/Signal/src/contact/OWSContactsManager.m
+++ b/Signal/src/contact/OWSContactsManager.m
@@ -63,14 +63,17 @@ NSString *const OWSContactsManagerSignalAccountsDidChangeNotification
 - (void)loadSignalAccountsFromCache
 {
     __block NSMutableArray<SignalAccount *> *signalAccounts;
-    [self.dbReadConnection readWithBlock:^(YapDatabaseReadTransaction * _Nonnull transaction) {
-        signalAccounts = [[NSMutableArray alloc] initWithCapacity:[SignalAccount numberOfKeysInCollectionWithTransaction:transaction]];
-        
+    [self.dbReadConnection readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
+        NSUInteger signalAccountCount = [SignalAccount numberOfKeysInCollectionWithTransaction:transaction];
+        DDLogInfo(@"%@ loading %lu signal accounts from cache.", self.logTag, (unsigned long)signalAccountCount);
+
+        signalAccounts = [[NSMutableArray alloc] initWithCapacity:signalAccountCount];
+
         [SignalAccount enumerateCollectionObjectsWithTransaction:transaction usingBlock:^(SignalAccount *signalAccount, BOOL * _Nonnull stop) {
             [signalAccounts addObject:signalAccount];
         }];
     }];
-    
+
     [self updateSignalAccounts:signalAccounts];
 }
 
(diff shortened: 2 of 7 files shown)


Signal/src/contact/OWSContactsManager.m
```diff
@@ -193,17 +209,16 @@ - (void)updateWithContacts:(NSArray<Contact *> *)contacts
 
             [self.avatarCache removeAllImages];
 
-            [self intersectContacts];
-
-            [self buildSignalAccounts];
+            [self intersectContactsWithCompletion:^(NSError *_Nullable error) {
```
michaelkirk: Previously we were calling buildSignalAccounts immediately, and then again after contacts intersected. 

Previously that made sense, since we had no local store of SignalAccounts. Now that we persist them, there's no reason to rebuild until we have a fresh intersection.

SignalServiceKit/src/Contacts/Contact.m
```diff
@@ -133,6 +134,23 @@ - (instancetype)initWithSystemContact:(CNContact *)contact
     return self;
 }
 
+- (instancetype)initWithCoder:(NSCoder *)coder
+{
+    self = [super initWithCoder:coder];
+    if (_imageData) {
+        _image = [UIImage imageWithData:_imageData];
```
michaelkirk: This is a way for `isEqual:` to work when determining if two Contact's are equivalent.

Signal/src/contact/OWSContactsManager.m
```diff
@@ -264,7 +339,11 @@ - (void)buildSignalAccounts
 - (void)updateSignalAccounts:(NSArray<SignalAccount *> *)signalAccounts
 {
     AssertIsOnMainThread();
-    
+    if ([signalAccounts isEqual:self.signalAccounts]) {
```
michaelkirk: This is significant - we only notify of the change when the value of the accounts actually changes. Because a lot of things observe this notification, it's expensive to fire it all the time.

This is particularly helps with the "settling time" immediately after launch now that we set once from cache, and then set again a few seconds later when contact's fetch and then intersect, which usually won't result in a a change.

SignalServiceKit/src/Contacts/Contact.m
```diff
@@ -127,12 +129,26 @@ - (instancetype)initWithSystemContact:(CNContact *)contact
...
     if (contact.thumbnailImageData) {
-        _image = [UIImage imageWithData:contact.thumbnailImageData];
+        _imageData = contact.thumbnailImageData;
     }
 
     return self;
 }
 
+- (nullable UIImage *)image
+{
```
michaelkirk: For equality checks, it's easier to compare the underlying image data than an instantiated UIImage, so I've shuffled things around to make the Data the persisted property, and the UIImage generated lazily from it.

Signal/src/contact/OWSContactsManager.m
```diff
@@ -218,43 +234,102 @@ - (void)buildSignalAccounts
...
+                DDLogVerbose(@"%@ Saving SignalAccount: %@", self.logTag, signalAccount);
                 [signalAccount saveWithTransaction:transaction];
             }
-            
-            if (orphanedKeys.count > 0) {
-                DDLogInfo(@"%@ Removing %lu orphaned SignalAccounts", self.logTag, (unsigned long)orphanedKeys.count);
-                [transaction removeObjectsForKeys:orphanedKeys.allObjects inCollection:[SignalAccount collection]];
+
+            if (clearStaleCache) {
+                DDLogInfo(@"%@ Removing %lu old SignalAccounts.", self.logTag, (unsigned long)oldSignalAccounts.count);
```
michaelkirk: SignalAccounts are only deleted when the user pulls-to-refresh

Signal/src/contact/OWSContactsManager.h
```diff
@@ -52,9 +52,11 @@ extern NSString *const OWSContactsManagerSignalAccountsDidChangeNotification;
 // Ensure's the app has the latest contacts, but won't prompt the user for contact
 // access if they haven't granted it.
 - (void)fetchSystemContactsOnceIfAlreadyAuthorized;
+
 // This variant will fetch system contacts if contact access has already been granted,
-// but not prompt for contact access. Also, it will always fire a notification.
-- (void)fetchSystemContactsIfAlreadyAuthorizedAndAlwaysNotify;
+// but not prompt for contact access. Also, it will always notify delegates, even if
+// contacts haven't changed, and will clear out any stale cached SignalAccount's
```
charlesmchen: Is `always notify delegates` a behavior change or comment change?

charlesmchen: s/SignalAccount's/SignalAccounts/

Signal/src/contact/OWSContactsManager.m
```diff
@@ -63,17 +63,32 @@ - (id)init {
 - (void)loadSignalAccountsFromCache
 {
     __block NSMutableArray<SignalAccount *> *signalAccounts;
-    [self.dbReadConnection readWithBlock:^(YapDatabaseReadTransaction * _Nonnull transaction) {
-        signalAccounts = [[NSMutableArray alloc] initWithCapacity:[SignalAccount numberOfKeysInCollectionWithTransaction:transaction]];
-        
+    [self.dbReadConnection readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
+        NSUInteger signalAccountCount = [SignalAccount numberOfKeysInCollectionWithTransaction:transaction];
+        DDLogInfo(@"%@ loading %lu signal accounts from cache.", self.logTag, (unsigned long)signalAccountCount);
```
charlesmchen: I always use `%zd` for `NSInteger` and `NSUinteger` to avoid a cast. Huh I wonder if I should use `%zu` for `NSUInteger`?

Signal/src/contact/OWSContactsManager.m
```diff
@@ -193,17 +210,16 @@ - (void)updateWithContacts:(NSArray<Contact *> *)contacts
...
         });
     });
 }
 
-- (void)buildSignalAccounts
+- (void)buildSignalAccountsAndClearStaleCache:(BOOL)clearStaleCache;
 {
-    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        NSMutableDictionary<NSString *, SignalAccount *> *signalAccountMap = [NSMutableDictionary new];
+    dispatch_async(self.serialQueue, ^{
```
michaelkirk: Doing this concurrently could give us strange results since it uses multiple transactions (there's good reason for using multiple transactions, most of them are reads, and only the writes intend to be narrowly scoped)

Signal/src/contact/OWSContactsManager.m
```diff
@@ -116,27 +131,30 @@ - (BOOL)supportsContactEditing
 
 - (void)systemContactsFetcher:(SystemContactsFetcher *)systemsContactsFetcher
               updatedContacts:(NSArray<Contact *> *)contacts
+                userRequested:(BOOL)userRequested
 {
-    [self updateWithContacts:contacts];
+    [self updateWithContacts:contacts clearStaleCache:userRequested];
```
charlesmchen: `shouldClearStaleCache`?

Signal/src/contact/OWSContactsManager.m
```diff
@@ -116,27 +131,30 @@ - (BOOL)supportsContactEditing
 
 - (void)systemContactsFetcher:(SystemContactsFetcher *)systemsContactsFetcher
               updatedContacts:(NSArray<Contact *> *)contacts
+                userRequested:(BOOL)userRequested
```
charlesmchen: `isUserRequested`? `isUserInitiated`?

Signal/src/contact/OWSContactsManager.m
charlesmchen: Great.

Signal/src/contact/OWSContactsManager.m
```diff
@@ -218,43 +234,102 @@ - (void)buildSignalAccounts
...
                 [signalAccounts addObject:signalAccount];
             }
         }
 
+        NSMutableDictionary<NSString *, SignalAccount *> *oldSignalAccounts = [NSMutableDictionary new];
+        [self.dbReadConnection readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
+            [SignalAccount
+                enumerateCollectionObjectsWithTransaction:transaction
+                                               usingBlock:^(id _Nonnull object, BOOL *_Nonnull stop) {
+                                                   if (![object isKindOfClass:[SignalAccount class]]) {
```
charlesmchen: I tend to just assert this ... how often do we have the wrong model type in a collection in the wild?

Signal/src/contact/OWSContactsManager.h
michaelkirk: > Is always notify delegates a behavior change or comment change?

 Just a comment change to clarify that this `notification` happens via a delegate callback, not a NSNotification.

> s/SignalAccount's/SignalAccounts/

done!

Signal/src/contact/OWSContactsManager.m
```diff
@@ -620,6 +699,16 @@ - (UIImage * _Nullable)imageForPhoneIdentifier:(NSString * _Nullable)identifier
     return image;
 }
 
+- (NSComparisonResult (^)(SignalAccount *left, SignalAccount *right))signalAccountComparator
+{
+    return ^NSComparisonResult(SignalAccount *left, SignalAccount *right) {
+        NSString *leftName = [self comparableNameForSignalAccount:left];
+        NSString *rightName = [self comparableNameForSignalAccount:right];
```
charlesmchen: To be a stable sort, we should sort by uniqueId or recipientId if the display names are the same.

SignalServiceKit/src/Contacts/Contact.m
charlesmchen: Comparing image-sized `NSData` instances will be expensive; we should make sure we never do it on the main thread.

michaelkirk: They are thumbnails, so they don't get too large (a couple dozen kb). Admittedly it's not free, but doing this check, even when on the main thread, is far cheaper than the consequence of firing the unnecessary notification.

Signal/src/contact/OWSContactsManager.m
michaelkirk: cowboy committed in 76481a86a5961521aac19fdcd176a3d45c47f70e

PR #2889, opened 2017-12-19
Charlesmchen/session database2

charlesmchen:

PTAL @michaelkirk


Signal/src/AppDelegate.m
```diff
@@ -7,27 +7,27 @@
...
 #import "RegistrationViewController.h"
-#import "Release.h"
 #import "Signal-Swift.h"
 #import "SignalApp.h"
 #import "SignalsNavigationController.h"
-#import "VersionMigrations.h"
 #import "ViewControllerUtils.h"
 #import <AxolotlKit/SessionCipher.h>
 #import <SignalMessaging/AppSetup.h>
+#import <SignalMessaging/Environment.h>
```
charlesmchen: Slowly trying to fix our imports as I touch files.

Signal/src/AppDelegate.m
```diff
@@ -135,7 +135,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     // Prevent the device from sleeping during database view async registration
     // (e.g. long database upgrades).
     //
-    // This block will be cleared in databaseViewRegistrationComplete.
+    // This block will be cleared in storageIsReady.
     [DeviceSleepManager.sharedInstance addBlockWithBlockObject:self];
```
charlesmchen: `OWSStorage.isStorageReady` is YES when _ALL_ storage classes have completed both their sync _AND_ async view registrations.

Signal/src/AppDelegate.m
```diff
@@ -219,8 +219,6 @@ - (void)ensureIsReadyForAppExtensions
     [TSStorageManager migrateToSharedData];
     [OWSProfileManager migrateToSharedData];
     [TSAttachmentStream migrateToSharedData];
-
-    [OWSPreferences setIsReadyForAppExtensions:YES];
```
charlesmchen: I'm moving this flag to until later in the main app launch sequence, since we'll soon be adding additional steps, e.g. "decompose primary storage."

SignalServiceKit/src/Messages/OWSFailedAttachmentDownloadsJob.m
```diff
@@ -121,21 +121,21 @@ - (YapDatabaseSecondaryIndex *)indexDatabaseExtension
 // Useful for tests, don't use in app startup path because it's slow.
 - (void)blockingRegisterDatabaseExtensions
 {
-    [self.storageManager registerExtension:[self indexDatabaseExtension]
+    [self.storageManager registerExtension:[self.class indexDatabaseExtension]
                                   withName:OWSFailedAttachmentDownloadsJobAttachmentStateIndex];
 }
 
-- (void)asyncRegisterDatabaseExtensions
++ (void)asyncRegisterDatabaseExtensionsWithStorageManager:(TSStorageManager *)storageManager
```
charlesmchen: We were needlessly instantiating some of our "jobs" when we registered their views.

SignalServiceKit/src/Storage/OWSStorage+Subclass.h
```diff
@@ -0,0 +1,23 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "OWSStorage.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OWSStorage (Subclass)
```
charlesmchen: The iOS frameworks use `+Subclass` headers for methods that only subclasses should have access to.  I've followed that pattern here.

SignalServiceKit/src/Storage/OWSStorage.h
```diff
@@ -6,24 +6,34 @@
...
 
 - (instancetype)init NS_UNAVAILABLE;
 - (instancetype)initStorage NS_DESIGNATED_INITIALIZER;
 
-- (void)setDatabaseInitialized;
+// Returns YES if _ALL_ storage classes have completed both their
+// sync _AND_ async view registrations.
++ (BOOL)isStorageReady;
+
+/**
```
charlesmchen: `OWSStorage` will have a bunch of class methods for:

* ...coordinating and observing storage setup/readiness.
* ...resetting storage.

SignalServiceKit/src/Storage/OWSStorage.m
```diff
@@ -79,7 +84,7 @@ - (void)readWriteWithBlock:(void (^)(YapDatabaseReadWriteTransaction *transactio
 {
     id<OWSDatabaseConnectionDelegate> delegate = self.delegate;
     OWSAssert(delegate);
-    OWSAssert(delegate.isDatabaseInitialized);
+    OWSAssert(delegate.areSyncRegistrationsComplete);
```
charlesmchen: Note that sometimes we gate on whether the _sync_ registrations are complete and sometimes we gate on whether _all_ registrations are complete.

charlesmchen:

PTAL @michaelkirk

SignalServiceKit/src/Storage/OWSStorage.m
```diff
@@ -243,37 +248,107 @@ - (instancetype)initStorage
...
+            //
+            // The best we can try to do is to discard the current database
+            // and behave like a clean install.
+            OWSProdCritical([OWSAnalyticsEvents storageErrorCouldNotLoadDatabase]);
 
-        // Try to reset app by deleting database.
-        // Disabled resetting storage until we have better data on why this happens.
-        // [self resetAllStorage];
+            // Try to reset app by deleting all databases.
+            [OWSStorage deleteDatabaseFiles];
```
michaelkirk: Just making sure this was intentionally enabled since it was previously commented out, and why the switch from `resetAllStorage` -> `deleteDatabaseFiles`

Seems like we'll have a lot of orphaned data, e.g. attachments, profiles

SignalServiceKit/src/Storage/TSStorageManager.h
```diff
@@ -18,22 +12,15 @@ NS_ASSUME_NONNULL_BEGIN
...
-
-// TODO: Deprecate.
+- (YapDatabaseConnection *)dbReadConnection;
+- (YapDatabaseConnection *)dbReadWriteConnection;
 + (YapDatabaseConnection *)dbReadConnection;
 + (YapDatabaseConnection *)dbReadWriteConnection;
 
 + (void)migrateToSharedData;
 
++ (NSString *)databaseFilePath;
```
michaelkirk: No biggie, but I don't see why this needs to be public. Can we get away with it only in +Subclass.h ?

SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -43,7 +51,29 @@ + (instancetype)sharedManager {
...
+
+        OWSSingletonAssert();
+    }
+
+    return self;
+}
+
+- (void)resetStorage
+{
+    _dbReadConnection = nil;
```
michaelkirk: It's an edge case, but is it safe to set these to nil? They aren't marked as nullable nor lazily initialized.

SignalServiceKit/src/Messages/OWSFailedMessagesJob.m
```diff
@@ -6,7 +6,7 @@
 #import "TSMessage.h"
 #import "TSOutgoingMessage.h"
 #import "TSStorageManager.h"
-#import <YapDatabase/YapDatabaseConnection.h>
+#import <YapDatabase/YapDatabase.h>
```
michaelkirk: ditto

SignalMessaging/Views/ThreadViewHelper.m
```diff
@@ -7,7 +7,7 @@
 #import <SignalServiceKit/TSDatabaseView.h>
 #import <SignalServiceKit/TSStorageManager.h>
 #import <SignalServiceKit/TSThread.h>
-#import <YapDatabase/YapDatabaseConnection.h>
```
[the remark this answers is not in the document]
charlesmchen: Done.  Also, `YapDatabase.h` isn't an umbrella header in the traditional self.  It's still necessary to do other imports once you use features like secondary indices, etc.

SignalServiceKit/src/Storage/OWSStorage.m
charlesmchen: Re-commented it with a TODO to revisit other cleanup.

SignalServiceKit/src/Storage/TSStorageManager.h
charlesmchen: Let's revisit in the next PR where I continue to rework this class.

SignalServiceKit/src/Storage/TSStorageManager.m
charlesmchen: We've always done it this way.  I consider this a work in progress towards closing (some) databases in the background.

PR #2893, opened 2017-12-20
attachment captioning

michaelkirk:
Adds attachment captioning, currently only via the share extension.

In broad strokes this PR:

1. At the data level, we now "understand" that a message can have both an attachment and a text caption. No more need to split incoming captioned attachments into two TSInteractons.

2. UI changes to send captions in the AttachmentApproval view.

3. UI changes to view captioned attachments in the ConversationView and the MessageDetail view.

More comments in line...

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
@@ -159,9 +159,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, nullable) OWSExpirationTimerView *expirationTimerView;
 @property (nonatomic, nullable) NSArray<NSLayoutConstraint *> *payloadConstraints;
 @property (nonatomic, nullable) NSArray<NSLayoutConstraint *> *dateHeaderConstraints;
-@property (nonatomic, nullable) NSArray<NSLayoutConstraint *> *contentConstraints;
+@property (nonatomic, nullable) NSMutableArray<NSLayoutConstraint *> *contentConstraints;
 @property (nonatomic, nullable) NSArray<NSLayoutConstraint *> *footerConstraints;
 @property (nonatomic) BOOL isPresentingMenuController;
+//@property (nonatomic) NSLayoutConstraint *textViewWidthConstraint;
+//@property (nonatomic) NSLayoutConstraint *textViewHeightConstraint;
 
 @end
 
@@ -180,6 +182,8 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)commontInit
 {
     OWSAssert(!self.textView);
+
+    _contentConstraints = [NSMutableArray new];
 
     self.layoutMargins = UIEdgeInsetsZero;
     self.contentView.layoutMargins = UIEdgeInsetsZero;
@@ -226,7 +230,11 @@ NS_ASSUME_NONNULL_BEGIN
     self.textView.textContainerInset = UIEdgeInsetsZero;
     self.textView.contentInset = UIEdgeInsetsZero;
     self.textView.scrollEnabled = NO;
+    //    self.textViewWidthConstraint = [self.textView autoSetDimension:ALDimensionWidth toSize:0];
+    //    self.textViewHeightConstraint = [self.textView autoSetDimension:ALDimensionHeight toSize:0];
+
     [self.myBubbleImageView addSubview:self.textView];
+
     OWSAssert(self.textView.superview);
 
     self.footerLabel = [UILabel new];
@@ -249,12 +257,12 @@ NS_ASSUME_NONNULL_BEGIN
 
     [self.myBubbleImageView autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:self.mediaMaskingView];
     [self.myBubbleImageView autoPinEdgeToSuperviewEdge:ALEdgeBottom];
-    
+
     // want sized to fit...
     //    [self.textMaskingView autoPinEdgeToSuperviewEdge:ALEdgeLeading];
     //    [self.textMaskingView autoPinEdgeToSuperviewEdge:ALEdgeTrailing];
     //    [self.footerView autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:self.textMaskingView];
-//    [self.footerView autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:self.myBubbleImageView];
+    //    [self.footerView autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:self.myBubbleImageView];
 
     [self.footerView autoPinEdgeToSuperviewEdge:ALEdgeBottom];
     [self.footerView autoPinWidthToSuperview];
@@ -740,24 +748,29 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)loadCaptionForAttachmentView
 {
-    [self loadForTextDisplay];
-
-    NSMutableArray *accumulatedConstraints = [self.contentConstraints mutableCopy];
-    [accumulatedConstraints addObjectsFromArray:@[
-        [self.myBubbleImageView autoPinEdgeToSuperviewEdge:(self.isIncoming ? ALEdgeLeading : ALEdgeTrailing)],
-        [self.myBubbleImageView autoPinEdgeToSuperviewEdge:(self.isIncoming ? ALEdgeTrailing : ALEdgeLeading) withInset:0 relation:NSLayoutRelationGreaterThanOrEqual],
-        [self.textView autoPinLeadingToSuperviewWithMargin:self.textLeadingMargin],
-        [self.textView autoPinTrailingToSuperviewWithMargin:self.textTrailingMargin],
-        [self.textView autoPinEdgeToSuperviewEdge:ALEdgeTop withInset:self.textVMargin],
-        [self.textView autoPinEdgeToSuperviewEdge:ALEdgeBottom withInset:self.textVMargin],
-    ]];
-
-    self.contentConstraints = [accumulatedConstraints copy];
+    [self loadForStandaloneTextDisplay];
+    return;
+    //    [self loadForTextDisplay];
+    //
+    //    NSMutableArray *accumulatedConstraints = [self.contentConstraints mutableCopy];
+    //    [accumulatedConstraints addObjectsFromArray:@[
+    //        [self.myBubbleImageView autoPinEdgeToSuperviewEdge:(self.isIncoming ? ALEdgeLeading : ALEdgeTrailing)],
+    ////        [self.myBubbleImageView autoPinEdgeToSuperviewEdge:(self.isIncoming ? ALEdgeTrailing :
+    /// ALEdgeLeading)withInset:0 /
+    /// relation:NSLayoutRelationGreaterThanOrEqual],
+    //        [self.textView autoPinLeadingToSuperviewWithMargin:self.textLeadingMargin],
+    //        [self.textView autoPinTrailingToSuperviewWithMargin:self.textTrailingMargin],
+    //        [self.textView autoPinEdgeToSuperviewEdge:ALEdgeTop withInset:self.textVMargin],
+    //        [self.textView autoPinEdgeToSuperviewEdge:ALEdgeBottom withInset:self.textVMargin],
+    //    ]];
+    //
+    //    self.contentConstraints = [accumulatedConstraints copy];
 }
 
 - (void)loadForStandaloneTextDisplay
 {
     [self loadForTextDisplay];
+
     if (self.displayableText.isTextTruncated) {
         self.tapForMoreLabel = [UILabel new];
         self.tapForMoreLabel.text = NSLocalizedString(@"CONVERSATION_VIEW_OVERSIZE_TEXT_TAP_FOR_MORE",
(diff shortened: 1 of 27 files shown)


SignalMessaging/attachments/AttachmentApprovalViewController.swift
```diff
@@ -170,26 +164,18 @@ public class AttachmentApprovalViewController: OWSViewController {
...
-        sendButton.tintColor = UIColor.white
+        let captioningToolbar = CaptioningToolbar()
+        captioningToolbar.captioningToolbarDelegate = self
+        self.bottomToolbar = captioningToolbar
+    }
 
-        let flexibleSpace = UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil)
-        bottomToolbar.items = [flexibleSpace, sendButton]
+    override public var inputAccessoryView: UIView? {
+        self.bottomToolbar.layoutIfNeeded()
```
michaelkirk: Rather than observing keyboard - change notifications, we can use the inputAccessoryView/becomeFirstResponder combo. 

I'm hopeful that we can do something similar with the input toolbar in the ConversationView to get a nicer dismiss animations when leaving the view and also to apply the interactive dismiss when scrolling - but that's for another day.

SignalMessaging/attachments/AttachmentApprovalViewController.swift
```diff
@@ -321,3 +356,181 @@ private class GradientView: UIView {
...
+    }
+
+    func didTapSend() {
+        self.captioningToolbarDelegate?.captioningToolbarDidTapSend(self, captionText: self.textView.text)
+    }
+
+    // MARK: - UIView Overrides
+
+    // We do progammatic layout, explicitly computing and setting frames since autoLayout does
+    // not seem to work with inputAccessory views, even when forcing a layout.
```
michaelkirk: This comment took me a while to arrive at.

SignalMessaging/categories/UIColor+OWS.h
```diff
@@ -8,21 +8,23 @@ NS_ASSUME_NONNULL_BEGIN
...
-+ (UIColor *)ows_yellowColor;
-+ (UIColor *)ows_reminderYellowColor;
-+ (UIColor *)ows_reminderDarkYellowColor;
-+ (UIColor *)ows_greenColor;
-+ (UIColor *)ows_redColor;
-+ (UIColor *)ows_blackColor;
-+ (UIColor *)ows_errorMessageBorderColor;
-+ (UIColor *)ows_infoMessageBorderColor;
-+ (UIColor *)ows_inputToolbarBackgroundColor;
+@property (class, readonly, nonatomic) UIColor *ows_systemPrimaryButtonColor;
```
michaelkirk: class properties just read a little more idiomatically for things like this in Swift.

SignalMessaging/utils/ThreadUtil.m
```diff
@@ -118,8 +118,10 @@ + (TSOutgoingMessage *)sendMessageWithAttachment:(SignalAttachment *)attachment
     TSOutgoingMessage *message =
         [[TSOutgoingMessage alloc] initWithTimestamp:[NSDate ows_millisecondTimeStamp]
                                             inThread:thread
+                                         messageBody:attachment.captionText
```
michaelkirk: data model changes were basically this easy.

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -327,17 +323,21 @@ class MessageDetailViewController: OWSViewController, UIScrollViewDelegate {
     private func contentRows() -> [UIView] {
         var rows = [UIView]()
 
+        if hasMediaAttachment {
```
michaelkirk: screen shot 2017-12-20 at 12 12 52 am

We now show the attachment+caption in the detail view. Design on this is TBD. I'm wondering if we can actually leverage the OWSMessageCell in this view to get the exact same layout in the detail screen as we do in the conversation view (though maybe special case oversized text messages), but it's not trivial. WDYT?

SignalMessaging/attachments/AttachmentApprovalViewController.swift
```diff
@@ -12,7 +12,7 @@ public protocol AttachmentApprovalViewControllerDelegate: class {
 }
 
 @objc
-public class AttachmentApprovalViewController: OWSViewController {
+public class AttachmentApprovalViewController: OWSViewController, CaptioningToolbarDelegate {
```
michaelkirk: screen shot 2017-12-20 at 12 11 14 am

screen shot 2017-12-20 at 12 11 44 am

charlesmchen:
Some initial questions before I start my PR.

* Will orphan data cleaner do the right thing?
* How will hybrid messages show up in home view and in notifications?
* I feel like there's a million places that we've baked in the assumption that a message has only text or attachment.  How do we verify that we've found and fixed them all?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -726,16 +784,21 @@ - (void)loadForTextDisplay
...
-        self.tapForMoreLabel.textColor = [textColor colorWithAlphaComponent:0.85];
+        self.tapForMoreLabel.textColor = [self.textColor colorWithAlphaComponent:0.85];
         self.tapForMoreLabel.textAlignment = [self.tapForMoreLabel textAlignmentUnnatural];
-        [self.bubbleImageView addSubview:self.tapForMoreLabel];
+        [self.textBubbleImageView addSubview:self.tapForMoreLabel];
 
-        self.contentConstraints = @[
+        [self.contentConstraints addObjectsFromArray:@[
+            [self.textBubbleImageView autoSetDimension:ALDimensionWidth toSize:textBubbleSize.width],
+            [self.textBubbleImageView autoSetDimension:ALDimensionHeight toSize:textBubbleSize.height],
```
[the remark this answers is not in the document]
charlesmchen: ...but that's what WA appears to do.  

wa hv 2
wa hv 1

verti test
horiz test

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -745,14 +808,17 @@ - (void)loadForTextDisplay
...
             [self.tapForMoreLabel autoPinEdgeToSuperviewEdge:ALEdgeBottom withInset:self.textVMargin],
             [self.tapForMoreLabel autoSetDimension:ALDimensionHeight toSize:self.tapForMoreHeight],
-        ];
+        ]];
     } else {
-        self.contentConstraints = @[
+        [self.contentConstraints addObjectsFromArray:@[
+            [self.textBubbleImageView autoSetDimension:ALDimensionWidth toSize:textBubbleSize.width],
+            [self.textBubbleImageView autoSetDimension:ALDimensionHeight toSize:textBubbleSize.height],
+            [self.textBubbleImageView autoPinEdgeToSuperviewEdge:(self.isIncoming ? ALEdgeLeading : ALEdgeTrailing)],
```
charlesmchen: TBD but having looked at WA a bunch I think we should be mroe like them: both media + caption inside a single bubble, media content rounded but not quite touching the bubble edges.  All message bubbles "full width" except for short text-only messages. Media always "full width" and with clamped aspect ratio.

michaelkirk:
> Will orphan data cleaner do the right thing?

Luckily the data-model level changes were pretty small. As I read the OrphanDataCleaner class, it cares about all attachments - but doesn't care whether the corresponding interaction has a message body or not. TSAttachments will continue to look the same as will the TSInteractions which reference them.

The change here is that we no longer make an extra non-attachment text message to simulate a caption. The presence or absence of these non-attachment messages shouldn't affect the ODC.

> How will hybrid messages show up in home view and in notifications?

I'd eventually like to incorporate some sort of media preview in the notification/inbox view. 

But for now, I've left this unchanged. When you receiving an attachment with a caption (which we've supported forever), a single notification appears like "📷: Attachment" (no caption appears in the notification).

We could change this easily enough, what do you think of something like:

(when there is a caption)
- "📷: This is my photo caption"

(when there's no caption)
- "📷: Attachment"

> I feel like there's a million places that we've baked in the assumption that a message has only text or attachment. How do we verify that we've found and fixed them all?

Agreed! Any ideas other than testing like mad?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
michaelkirk: I'm leaning towards that as well, but I didn't want to make a big design change until we'd discussed it, so this PR tries to largely matches the existing design.

The lions share of this work is compatible with that approach - a MessageCell can now contain both an attachment container and a text container.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -466,51 +471,80 @@ - (nullable TSAttachmentPointer *)attachmentPointer
...
         [[UIMenuItem alloc] initWithTitle:NSLocalizedString(@"EDIT_ITEM_SHARE_ACTION",
                                               @"Short name for edit menu item to share contents of media message.")
-                                   action:self.shareActionSelector],
+                                   action:self.shareTextActionSelector],
+        [[UIMenuItem alloc] initWithTitle:NSLocalizedString(@"EDIT_ITEM_MESSAGE_METADATA_ACTION",
+                                              @"Short name for edit menu item to show message metadata.")
+                                   action:self.metadataActionSelector],
+        // FIXME: when deleting a caption, users will be surprised that it also deletes the attachment.
+        // We either need to implement a way to remove the caption separate from the attachment
+        // or make a design change which clarifies that the whole message is getting deleted.
```
charlesmchen: Is there a card?

charlesmchen: Another argument for WA-style UI which visually groups this content.

It's a shame that we can't make a long-press on hybrid messages to show "copy image" "copy caption", etc.

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -378,7 +378,12 @@ + (void)sendAttachment:(NSString *)filePath
     NSString *utiType = [MIMETypeUtil utiTypeForFileExtension:filename.pathExtension];
     DataSource *_Nullable dataSource = [DataSourcePath dataSourceWithFilePath:filePath];
     [dataSource setSourceFilename:filename];
-    SignalAttachment *attachment = [SignalAttachment attachmentWithDataSource:dataSource dataUTI:utiType];
+    SignalAttachment *attachment =
+        [SignalAttachment attachmentWithDataSource:dataSource dataUTI:utiType imageQuality:TSImageQualityOriginal];
+    if (arc4random_uniform(100) > 50) {
```
charlesmchen: Maybe create a `randomBoolean()` function?

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -148,11 +148,9 @@ class MessageDetailViewController: OWSViewController, UIScrollViewDelegate {
         contentView.autoPinEdge(toSuperviewEdge: .bottom)
         scrollView.layoutMargins = UIEdgeInsets(top: 0, left: 0, bottom: 0, right: 0)
 
-        let hasAttachment = message.attachmentIds.count > 0
```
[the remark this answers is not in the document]
charlesmchen: Okay.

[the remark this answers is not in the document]
charlesmchen: Part of me is tempted to let users share all attachments, including text.  e.g. share to notes. share to an email.

Signal/src/ViewControllers/MessageDetailViewController.swift
charlesmchen: Looks nice like that - I'd vote that we _not_ use the `OWSMessageCell` layout.  In particular, it'd be nice for media to _not_ be cropped in this view.

SignalMessaging/attachments/AttachmentApprovalViewController.swift
charlesmchen: Not bad.  I'd argue for bigger margins (between textarea/button + edges of screen) and spacing (between textarea + button).  And I'd vote that the send button's height should match height of empty/single-line textarea height.

SignalMessaging/attachments/AttachmentApprovalViewController.swift
charlesmchen: Let's discuss offline.

charlesmchen:
> Agreed! Any ideas other than testing like mad?

I suspect that many of these take the form:
```
if (message.attachmentIds.count > 0) {
   // Only process attachment
} else {
   // Only process text.
}
```
but there could be some of the form:
```
// Or a nil test.
if (message.body.length > 0) {
   // Only process attachment
} else {
   // Only process text.
}
```
So part of me thinks the right way to do this would be to rename both `body` and `attachmentIds` and fix/audit all of the breakage, then rename them back.

Signal/src/ViewControllers/MessageDetailViewController.swift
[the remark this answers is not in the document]
michaelkirk: My thinking here is that users should not be able to tell that oversized-text attachments exist. They just type a really long message (or copy/paste) and send it off oversized-text attachments are a client mechanism to achieve that.

I agree that users should be able to share plain text attachments, e.g. if they import someFile.txt from the document picker.

SignalMessaging/attachments/AttachmentApprovalViewController.swift
michaelkirk: You're right, I've increased the margins and matched the button height.

screen shot 2017-12-20 at 4 59 37 pm

michaelkirk:
In 37ee9f0e7f867517bd4a8ae0b20422f8bee3cb01

PR #2898, opened 2017-12-23
sharing vs sn changes

michaelkirk:
PTAL @charlesmchen


Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
```diff
@@ -81,56 +77,6 @@ - (nullable OWSTableSection *)sectionForThread:(nullable TSThread *)threadParame
...
-                                                      handler:^(UIAlertAction *_Nonnull action) {
-                                                          NSData *identityKey =
-                                                              [identityManger identityKeyForRecipientId:recipientId];
-                                                          [[OWSIdentityManager sharedManager]
-                                                               setVerificationState:OWSVerificationStateVerified
-                                                                        identityKey:identityKey
-                                                                        recipientId:recipientId
-                                                              isUserInitiatedChange:NO];
-                                                      }]];
-    [alertController addAction:[UIAlertAction actionWithTitle:@"No Longer Verified"
```
michaelkirk: This skips some important machinery - if you want someone to be marked as no-longer-verified, first verify them in the fingerprint controller, then use the debug UI to toggle a SN change.

SignalMessaging/attachments/AttachmentApprovalViewController.swift
```diff
@@ -231,36 +231,53 @@ public class AttachmentApprovalViewController: OWSViewController, CaptioningTool
         // Toolbar flickers in and out if there are errors
         // and remains visible momentarily after share extension is dismissed.
         // It's easiest to just hide it at this point since we're done with it.
-        self.bottomToolbar.isUserInteractionEnabled = false
-        self.bottomToolbar.isHidden = true
+        shouldAllowAttachmentViewResizing = false
```
michaelkirk: Presenting/dismissing alerts can cause the view controller to become/resign first responder, which has implication for when the keyboard is presented which in turn affects the zoom status of the attachment view.

So we make sure to not fire any more of those animations after the send button has been hit.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1540,6 +1540,9 @@
 /* action sheet item */
 "SHARE_ACTION_TWEET" = "Twitter";
 
+/* alert body when sharing filefailed because of untrusted/changed identity keys */
+"SHARE_EXTENSION_FAILED_SENDING_BECAUSE_UNTRUSTED_IDENTITY_KEY" = "Your safety number with %@ has recently changed. You may wish to verify it in the main app before resending.";
```
charlesmchen: TIOLI - I tend to use a `_FORMAT` suffix for l10n strings that should be formatted.

SignalMessaging/attachments/AttachmentApprovalViewController.swift
```diff
@@ -231,36 +231,53 @@ public class AttachmentApprovalViewController: OWSViewController, CaptioningTool
...
+    // so we sometimes allows resizing the attachment.
+    private var shouldAllowAttachmentViewResizing: Bool = true
 
-                    let originalHeight = self.scrollView.bounds.size.height
-
-                    // Position the new scaled item to be centered with respect
-                    // to it's new size.
-                    let heightDelta = originalHeight * (1 - kScaleFactor)
-                    let translate = CGAffineTransform(translationX: 0, y: -heightDelta / 2)
+    enum AttachmetViewScale {
```
charlesmchen: s/AttachmetViewScale/AttachmentViewScale/

SignalMessaging/attachments/AttachmentApprovalViewController.swift
```diff
@@ -231,36 +231,53 @@ public class AttachmentApprovalViewController: OWSViewController, CaptioningTool
...
+        case fullsize, compact
+    }
```
charlesmchen: IMO all class properties (e.g. `shouldAllowAttachmentViewResizing `) and inner types (e.g. `AttachmetViewScale`) belong at the top of the class, before any methods.

SignalMessaging/attachments/SharingThreadPickerViewController.m
```diff
@@ -233,44 +216,183 @@ - (void)tryToSendAttachment:(SignalAttachment *)attachment fromViewController:(U
...
-    UIAlertAction *retryAction =
-        [UIAlertAction actionWithTitle:[CommonStrings retryButton]
-                                 style:UIAlertActionStyleDefault
-                               handler:^(UIAlertAction *action) {
-                                   [self tryToSendAttachment:attachment fromViewController:fromViewController];
-                               }];
-    [failureAlert addAction:retryAction];
-
-    [fromViewController presentViewController:failureAlert animated:YES completion:nil];
+    if (error.code == OWSErrorCodeUntrustedIdentity) {
```
charlesmchen: Don't you need to check the error domain too?

Signal/translations/en.lproj/Localizable.strings
michaelkirk: Done!

SignalMessaging/attachments/AttachmentApprovalViewController.swift
michaelkirk: Thanks.

SignalMessaging/attachments/AttachmentApprovalViewController.swift
michaelkirk: Done.

SignalMessaging/attachments/SharingThreadPickerViewController.m
michaelkirk: Done!

PR #2899, opened 2017-12-23
Fix crash when viewing non-attachment message details.

michaelkirk:
// FREEBIE


michaelkirk:
PTAL @charlesmchen  

(cowboying this one)

PR #2904, opened 2018-01-02
Charlesmchen/session database5

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Network/API/TSNetworkManager.m
+++ b/SignalServiceKit/src/Network/API/TSNetworkManager.m
@@ -3,9 +3,12 @@
 //
 
 #import "TSNetworkManager.h"
+#import "AppContext.h"
 #import "NSURLSessionDataTask+StatusCode.h"
 #import "OWSSignalService.h"
 #import "TSAccountManager.h"
+#import "TSRecipientPrekeyRequest.h"
+#import "TSSubmitMessageRequest.h"
 #import "TSVerifyCodeRequest.h"
 #import <AFNetworking/AFNetworking.h>
 
@@ -56,6 +59,13 @@ typedef void (^failureBlock)(NSURLSessionDataTask *task, NSError *error);
             failure:(void (^)(NSURLSessionDataTask *task, NSError *error))failureBlock
 {
     DDLogInfo(@"%@ Making request: %@", self.logTag, request);
+    if (!CurrentAppContext().isMainApp) {
+        if (![request isKindOfClass:[TSRecipientPrekeyRequest class]]
+            && ![request isKindOfClass:[TSSubmitMessageRequest class]]) {
+            // The SAE should only make requests directly related to message sending.
+            OWSFail(@"%@ Making request: %@", self.logTag, request);
+        }
+    }
 
     void (^failure)(NSURLSessionDataTask *task, NSError *error) =
         [TSNetworkManager errorPrettifyingForFailureBlock:failureBlock];
--- a/SignalServiceKit/src/Storage/AxolotlStore/OWSSessionStorage+SessionStore.m
+++ b/SignalServiceKit/src/Storage/AxolotlStore/OWSSessionStorage+SessionStore.m
@@ -8,8 +8,8 @@
 #import <YapDatabase/YapDatabase.h>
 
 NSString *const OWSSessionStorageSessionStoreCollection = @"TSStorageManagerSessionStoreCollection";
-NSString *const OWSSessionStore_Collection = @"OWSSessionStore_Collection";
-NSString *const OWSSessionStore_Key_HasMigratedToSessionStorage = @"OWSSessionStore_Key_HasMigratedToSessionStorage";
+NSString *const OWSSessionStore_Collection = @"OWSSessionStorage_Collection";
+NSString *const OWSSessionStore_Key_HasMigratedToSessionStorage = @"OWSSessionStorage_Key_HasMigratedToSessionStorage";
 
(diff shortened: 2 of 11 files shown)


SignalServiceKit/src/Network/API/TSNetworkManager.m
```diff
@@ -56,6 +59,13 @@ - (void)makeRequest:(TSRequest *)request
             failure:(void (^)(NSURLSessionDataTask *task, NSError *error))failureBlock
 {
     DDLogInfo(@"%@ Making request: %@", self.logTag, request);
+    if (!CurrentAppContext().isMainApp) {
+        if (![request isKindOfClass:[TSRecipientPrekeyRequest class]]
+            && ![request isKindOfClass:[TSSubmitMessageRequest class]]) {
+            // The SAE should only make requests directly related to message sending.
+            OWSFail(@"%@ Making request: %@", self.logTag, request);
```
charlesmchen: I'm hoping to add a lot of asserts to help identify unexpected behavior in the SAE.  The more broad ones (like this) don't `return` or otherwise abort.

SignalServiceKit/src/Storage/AxolotlStore/OWSSessionStorage+SessionStore.m
```diff
@@ -8,9 +8,8 @@
 #import <YapDatabase/YapDatabase.h>
 
 NSString *const OWSSessionStorageSessionStoreCollection = @"TSStorageManagerSessionStoreCollection";
-NSString *const OWSSessionStorage_Collection = @"OWSSessionStorage_Collection";
-NSString *const OWSSessionStorage_Key_HasMigratedToSessionStorage
-    = @"OWSSessionStorage_Key_HasMigratedToSessionStorage";
+NSString *const OWSSessionStore_Collection = @"OWSSessionStore_Collection";
+NSString *const OWSSessionStore_Key_HasMigratedToSessionStorage = @"OWSSessionStore_Key_HasMigratedToSessionStorage";
```
charlesmchen: This doesn't look right - need to push a commit.

SignalServiceKit/src/Storage/OWSSessionStorage.m
```diff
@@ -177,6 +178,41 @@ + (YapDatabaseConnection *)dbConnection
...
+                                         }
+
+                                         collectionContents[key] = value;
+                                     }];
+    }];
+
+    // 2. Write to new storage.
+    [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
+        [collectionContents enumerateKeysAndObjectsUsingBlock:^(NSString *_Nonnull key, id value, BOOL *_Nonnull stop){
+        }];
```
charlesmchen: There's a missing line here - the bug is fixed in the next PR.

SignalServiceKit/src/Storage/AxolotlStore/OWSSessionStorage+SessionStore.m
charlesmchen: Pushed.

SignalServiceKit/src/Storage/OWSStorage.m
```diff
@@ -271,6 +273,13 @@ + (void)setupWithSafeBlockingMigrations:(void (^_Nonnull)(void))safeBlockingMigr
     // app to crash on launch.
     safeBlockingMigrationsBlock();
 
+    // We need to do this _before_ anyone accesses the session or identity store state.
+    //
+    // None of this state is used by a view, so we can safely do this before the
```
michaelkirk: >  // None of this state is used by a view, so we can safely do this before the

We're talking about **YapDB** views here (as opposed to UIView's), right? 

Regardless of that fact that **we** know that no YapDB views use these collections, the views themselves don't know what happened in the write transaction (they only know *that* a write transaction occurred), so creating a write transaction before a view is registered will cause that view to be re-built from scratch, as soon as it does register, right?

SignalServiceKit/src/Storage/OWSSessionStorage.m
charlesmchen: Right.  There's a bunch more work though that I around improving this method and I moved it elsewhere, etc. hence why I don't want to just fix it here.

SignalServiceKit/src/Storage/OWSSessionStorage.m
```diff
@@ -177,6 +178,41 @@ + (YapDatabaseConnection *)dbConnection
...
+{
+    OWSAssert(collection.length > 0);
+    OWSAssert(storage);
+
+    DDLogInfo(@"%@: migrating %@", self.logTag, collection);
+
+    NSMutableDictionary<NSString *, id> *collectionContents = [NSMutableDictionary new];
+
+    // 1. Read from old storage.
+    [storage.newDatabaseConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
```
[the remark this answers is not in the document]
charlesmchen: The idea was to block on any pending writes in case there were migrations etc. afoot.

SignalServiceKit/src/Storage/OWSStorage.m
charlesmchen: Huh, you might be right.  I guess we should use _read_ transactions when migrating/copying collections.

charlesmchen:
PTAL @michaelkirk

michaelkirk:
LGTM

Issue #2879, opened 2017-12-12
Swipe right to go back sporadically breaks


TempSpas:
- [x] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description

Signal provides an easy way to navigate back from a conversation to the inbox by swiping right from the left edge of the screen. However, on occasion, this functionality will break entirely, and no navigation will occur.

### Steps to reproduce
- Attempt to swipe right from the left edge
- Observe that there is no screen transition, no navigation
- Observe that the contact name at the top will become overlapped with the back arrow as if the screen transition were occurring, but nothing happens.

**Actual result:** Swipe should navigate back to inbox.
**Expected result:** It doesn't.

### Screenshots

![signal bug](https://user-images.githubusercontent.com/3801275/33911302-ed33b0d8-df5f-11e7-8805-22e5029e6807.png)

The above is a screenshot of the overlap that occurs. Again, this is the only thing on the screen that changes at all when the swipe is initiated. No other transition occurs. I have not found a way to reliably reproduce this. Force quitting the app and restarting it does not fix the issue.

### Device info

**Device**: iPhone 6
**iOS version**: 11.2
**Signal version:** 2.19.2.0

TempSpas:
I've figured out part of what's happening. The bug occurs **when you attempt to swipe anywhere along the left edge that someone else's message is.** The swipe-back works correctly if you swipe along the edge anywhere across from one of your own messages. In other words:

![signal-ios-app](https://user-images.githubusercontent.com/3801275/33912969-e977a48a-df65-11e7-8f46-d66777425a93.png)

The red areas along the side are where the bug occurs and the swipe is broken. The green areas are where it works as intended.

mmiller42:
Also occurs for me on iOS 11 on iPhone 8.

michaelkirk:
This will be fixed in the 2.20 release, stay tuned!

PR #2905, opened 2018-01-02
Only activate "Show detail" pan gesture when swiping left

michaelkirk:
Fixes https://github.com/WhisperSystems/Signal-iOS/issues/2879

PTAL @charlesmchen

### DIFF
--- a/Signal/src/views/DirectionalPanGestureRecognizer.swift
+++ b/Signal/src/views/DirectionalPanGestureRecognizer.swift
@@ -6,7 +6,7 @@ import UIKit.UIGestureRecognizerSubclass
 
 @objc
 enum PanDirection: Int {
-    case forward, backward, up, down, any
+    case left, right, up, down, any
 }
 
 @objc
@@ -33,13 +33,13 @@ class PanDirectionGestureRecognizer: UIPanGestureRecognizer {
             let deltaX = previousLocation.x - location.x
 
             switch direction {
-            case .down where deltaY < 0:
-                return
             case .up where deltaY > 0:
                 return
-            case .forward where deltaX < 0:
+            case .down where deltaY < 0:
                 return
-            case .backward where deltaX > 0:
+            case .left where deltaX > 0:
+                return
+            case .right where deltaX < 0:
                 return
             default:
                 break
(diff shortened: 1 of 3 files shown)


Signal/src/views/DirectionalPanGestureRecognizer.swift
```diff
@@ -17,19 +16,50 @@ class PanDirectionGestureRecognizer: UIPanGestureRecognizer {
...
+            let deltaX = previousLocation.x - location.x
+
+            switch direction {
+            case .down where deltaY < 0:
+                return
+            case .up where deltaY > 0:
+                return
+            case .forward where deltaX < 0:
+                return
+            case .backward where deltaX > 0:
```
charlesmchen: What about RTL?

Signal/src/views/DirectionalPanGestureRecognizer.swift
```diff
@@ -17,19 +16,50 @@ class PanDirectionGestureRecognizer: UIPanGestureRecognizer {
...
         if state == .began {
             let vel = velocity(in: view)
             switch direction {
-            case .horizontal where fabs(vel.y) > fabs(vel.x):
-                state = .cancelled
-            case .vertical where fabs(vel.x) > fabs(vel.y):
-                state = .cancelled
+            case .forward, .backward:
+                if fabs(vel.y) > fabs(vel.x) {
+                    state = .cancelled
```
charlesmchen: `UIGestureRecognizer`s can't always be cancelled; IMO safer to have an `abortGesture()` method that not only _tries_ to enter cancelled state but also toggles the enabled/disabled state, thereby forcing the gesture to abort.

michaelkirk: When can't they be cancelled?

Signal/src/views/DirectionalPanGestureRecognizer.swift
michaelkirk: Good call, thanks. Updated!

PTAL @charlesmchen

Signal/src/views/DirectionalPanGestureRecognizer.swift
charlesmchen: It's a particularity of `UIGestureRecognizer` that cancelling doesn't always work.  I can't remember the exact repro, but it's not terribly obscure.

PR #2908, opened 2018-01-03
Fix storage reset.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Storage/OWSStorage.m
+++ b/SignalServiceKit/src/Storage/OWSStorage.m
@@ -1,5 +1,5 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 #import "OWSStorage.h"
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Storage/OWSStorage.m
```diff
@@ -450,9 +461,7 @@ - (void)resetStorage
 
 + (void)resetAllStorage
 {
-    for (OWSStorage *storage in self.allStorages) {
-        [storage resetStorage];
-    }
+    [[NSNotificationCenter defaultCenter] postNotificationName:OWSResetStorageNotification object:nil];
```
michaelkirk: Can you help me understand this change - what was wrong with the old approach?

charlesmchen: We reset storage in a number of scenarios including failure to load key from keychain.  This happens while initializing `TSStorageManager`. Therefore we can't iterate over allStorages 
 - by definition those storages might not be initialized or might be in the process of initializing.  

Correct approach - fire a notification so that any pre-existing storage instances reset.

michaelkirk:
LGTM

PR #2909, opened 2018-01-04
Charlesmchen/session database6 

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -301,7 +301,7 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(yapDatabaseModifiedExternally:)
                                                  name:YapDatabaseModifiedExternallyNotification
-                                               object:TSStorageManager.sharedManager.dbNotificationObject];
+                                               object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(applicationWillEnterForeground:)
                                                  name:OWSApplicationWillEnterForegroundNotification
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -1,5 +1,5 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 #import "ConversationViewController.h"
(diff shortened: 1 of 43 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -297,11 +297,11 @@ - (void)addNotificationListeners
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(yapDatabaseModified:)
                                                  name:YapDatabaseModifiedNotification
-                                               object:nil];
+                                               object:TSStorageManager.sharedManager.dbNotificationObject];
```
charlesmchen: Filter by database so that the database copying process doesn't spam the views with irrelevant modifications.

SignalMessaging/environment/AppSetup.m
```diff
@@ -40,6 +41,10 @@ + (void)setupEnvironment:(CallMessageHandlerBlock)callMessageHandlerBlock
                                                   profileManager:OWSProfileManager.sharedManager];
         [TextSecureKitEnv setSharedEnv:sharedEnv];
 
+        // Register renamed classes.
+        [NSKeyedUnarchiver setClass:[OWSUserProfile class] forClassName:[OWSUserProfile collection]];
+        [NSKeyedUnarchiver setClass:[OWSDatabaseMigration class] forClassName:[OWSDatabaseMigration collection]];
```
charlesmchen: We need to register these before the database copy.

SignalMessaging/utils/OWSPreferences.m
```diff
@@ -24,7 +25,7 @@
 NSString *const OWSPreferencesKeyCallsHideIPAddress = @"CallsHideIPAddress";
 NSString *const OWSPreferencesKeyHasDeclinedNoContactsView = @"hasDeclinedNoContactsView";
 NSString *const OWSPreferencesKeyIOSUpgradeNagVersion = @"iOSUpgradeNagVersion";
-NSString *const OWSPreferencesKey_IsReadyForAppExtensions = @"isReadyForAppExtensions_3";
+NSString *const OWSPreferencesKey_IsReadyForAppExtensions = @"isReadyForAppExtensions_5";
```
charlesmchen: FYI I've worked and reworked the SAE migration process so many times that there may be conflicts with dev devices that did previous migrations.

SignalServiceKit/src/Storage/OWSDatabaseConnection.m
```diff
@@ -47,32 +47,13 @@ - (void)readWithBlock:(void (^)(YapDatabaseReadTransaction *transaction))block
...
-    id<OWSDatabaseConnectionDelegate> delegate = self.delegate;
-    OWSAssert(delegate);
-    OWSAssert(delegate.areSyncRegistrationsComplete);
-
-    [delegate readTransactionWillBegin];
-    [super asyncReadWithBlock:block
-              completionBlock:^{
-                  [delegate readTransactionDidComplete];
-              }];
+    [self asyncReadWithBlock:block completionQueue:NULL completionBlock:nil];
```
charlesmchen: The base class implementation just defers to the all-params flavor of this method. Do that explicitly here to DRY up the logic in this class.

SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -201,6 +298,204 @@ + (YapDatabaseConnection *)dbReadWriteConnection
     return TSStorageManager.sharedManager.dbReadWriteConnection;
 }
 
+#pragma mark - Primary Copy
+
+
+// To avoid the 0xdead10cc crashes (iOS app can't retain file lock
```
charlesmchen: I'm not going to finish this "essay" until we finalize our solution.

Issue #2760, opened 2017-11-16
Broken iOS Layout


PlainSite:
- [X] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description
See screenshot.

### Steps to reproduce
- Talk on the phone. End conversation.

**Actual result:** Broken layout with input bar in the middle of the screen.
**Expected result:** Correct layout.

### Screenshots
![img_1717](https://user-images.githubusercontent.com/8204476/32876157-64de6738-ca50-11e7-8555-10e915b35741.PNG)

### Device info

**Device**: iPod Touch
**iOS version**: 11.1.1
**Signal version:** 2.18.2.1

charlesmchen:
Thanks @PlainSite .  This is fixed in the next release, v2.19.

axfelix:
Are you behind on pushing releases to the app store? Just curious as I've just installed the app for the first time and it seems to be running behind this repository for at least a couple numbered builds...

michaelkirk:
This fix has been released for a while now. Please re-open if you see it on anything 2.19.5 or newer.

Issue #1126, opened 2016-03-24
fullscreen video player's presenting animation should "zoom in" on message


michaelkirk:
Currently zoom is from top left corner, which is a little disorienting.

We should center the zoom animation on the message you're zooming into for fullscreen video, images, and GIF.

e.g. here you can see mid animation, I just tapped the video, and it's growing from the top left corner.

![image](https://cloud.githubusercontent.com/assets/217057/14033497/0cc3af42-f1d8-11e5-8c3a-c42ca15abbf5.png)

Instead, the affect should be more like we're expanding the thumbnail to fill the screen (see built in messages app for example).

michaelkirk:
Fixed in #2911

PR #2911, opened 2018-01-08
Manage keyboard via inputAccessoryView and revamp full-screen media view

### LINKED ISSUE #2812: Text entry field is not positioned correctly on iPhone X for emoji keyboard

### Bug description
On iPhone X, the text entry field is not positioned at the top of the keyboard when using the emoji keyboard.

### Steps to reproduce
- Launch Signal on iPhone X
- Begin a conversation
- Switch to the emoji keyboard

**Actual result:** The text entry field is not visible until you switch back to a textual keyboard.
**Expected result:** The text entry field should be repositioned above the keyboard.

### Screenshots
img_0207

### Device info

**Device**: iPhone X
**iOS version**: 11.1.2 (15B202)
**Signal version:** 2.19.0.22

michaelkirk:
Using an inputAccessoryView works around a bug in iOS11.2 wherein switching to the emoji keyboard doesn't always trigger keyboard frame change notifications.

It has some other niceties too:
- keyboard can be dismissed interactively in scrollview
- smooths out some animations when exiting the view, e.g. swipe left for details with keyboard up

One unexpected hiccup was the presentation of the old move player view. Since it's in the same hierarchy, the input toolbar wanted to display above it. I started down a path to sort of hack around this problem, but the hack was enough work that I decided to try to solve it more thoroughly by moving the video playback into the FullImageViewController (and using the modern AVPlayer rather than the legacy MPMoviePlayer. I'm hopeful that this will help pave the way for a PageView controller style thread media gallery in the future.

I ended up renaming FullImageViewController -> MediaDetailViewController, and since there were so many changes, GH isn't rendering a useful diff. I put the rename in a separate commit so you can omit that from the comparison if you'd like to see a slightly more useful diff.

fixes #2812 
fixes https://github.com/WhisperSystems/Signal-iOS/issues/1126
fixes https://github.com/WhisperSystems/Signal-iOS/issues/2760 (if it wasn't already)
closes #2028 

**full screen image view**
fs-image

**tap once to hide bars**
fs-image-dark

**pinch to zoom image. Delete bar button shows confirmation**
fs-image-delete

**a paused movie**
fs-image-movie

PTAL @charlesmchen


charlesmchen:
@michaelkirk You opened this PR against master.  Shouldn't it be against `hotfix/2.19.3`?

Signal/src/UIView+OWS.m
```diff
@@ -299,6 +299,32 @@ - (NSLayoutConstraint *)autoPinTrailingToSuperviewWithMargin:(CGFloat)margin
     }
 }
 
+- (NSLayoutConstraint *)autoPinBottomToSuperviewWithMargin:(CGFloat)margin
+{
+    if (@available(iOS 9.0, *)) {
+        NSLayoutConstraint *constraint =
+        [self.bottomAnchor constraintEqualToAnchor:self.superview.layoutMarginsGuide.bottomAnchor
+                                          constant:-margin];
```
charlesmchen: What's the difference in behavior here?

Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
```diff
@@ -68,18 +70,34 @@ - (void)dealloc
     [self removeKVOObservers];
 }
 
+- (CGSize)intrinsicContentSize
+{
+    CGSize newSize = CGSizeMake(self.bounds.size.width, self.toolbarHeight + ConversationInputToolbarBorderViewHeight);
+    return newSize;
+}
+
+- (void)setToolbarHeight:(CGFloat)toolbarHeight
```
charlesmchen: Why bother with this method?

Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
```diff
@@ -223,12 +241,16 @@ - (void)ensureContentConstraints
     const CGFloat kMinTextViewHeight = ceil(self.inputTextView.font.lineHeight
         + self.inputTextView.textContainerInset.top + self.inputTextView.textContainerInset.bottom
         + self.inputTextView.contentInset.top + self.inputTextView.contentInset.bottom);
-    const CGFloat kMaxTextViewHeight = 100.f;
+    // Exactly 4 lines of text with default sizing.
+    const CGFloat kMaxTextViewHeight = 98.f;
```
charlesmchen: So this value doesn't reflect dynamic text - it seems strange to hardcode this value rather than derive it from something like `"UIFont with default text size".lineHeight * 4.f`.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1388,11 +1401,13 @@ - (void)toggleDefaultKeyboard
         return;
     }
 
+    // FIXME inputAccessoryView - if using numeric keyboard, switch back to alpha after
+    // sending.
```
charlesmchen: ?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3658,6 +3639,100 @@ - (void)tryToSendAttachmentIfApproved:(SignalAttachment *_Nullable)attachment
...
+    if (!keyboardEndFrameValue) {
+        OWSFail(@"%@ Missing keyboard end frame", self.logTag);
+        return;
+    }
+    CGRect keyboardEndFrame = [keyboardEndFrameValue CGRectValue];
+
+    //    DDLogVerbose(@"%@ keyboard change. Old Frame: %@, New Frame: %@",
+    //        self.logTag,
+    //        NSStringFromCGRect(keyboardBeginFrame),
+    //        NSStringFromCGRect(keyboardEndFrame));
```
charlesmchen: Remove?

Signal/src/ViewControllers/ConversationView/ConversationViewLayout.m
```diff
@@ -68,6 +68,11 @@ - (void)prepareLayout
         [self clearState];
         return;
     }
+    
+    if (self.collectionView.bounds.size.width <= 0.f || self.collectionView.bounds.size.height <= 0.f) {
+        [self.collectionView layoutIfNeeded];
```
charlesmchen: This doesn't seem right.  There's logic in `ConversationViewController` to ensure that collection view has a valid width from the get-go.  Why should this layout care about the collection view's height?

Signal/src/ViewControllers/MediaDetailViewController.m
```diff
@@ -0,0 +1,902 @@
...
+    self.areToolbarsHidden = !self.areToolbarsHidden;
+}
+
+- (void)didDoubleTapImage:(id)sender
+{
+    DDLogVerbose(@"%@ did tap image.", self.logTag);
+    if (self.scrollView.zoomScale == self.scrollView.minimumZoomScale) {
+        [self.scrollView setZoomScale:self.scrollView.minimumZoomScale * 2 animated:YES];
+    } else {
+        [self.scrollView setZoomScale:self.scrollView.minimumZoomScale animated:YES];
```
charlesmchen: Shouldn't double-tap zoom in on the gesture location?

SignalServiceKit/src/Storage/OWSOrphanedDataCleaner.m
```diff
@@ -32,7 +32,7 @@ + (void)auditAsync
 + (void)auditAndCleanupAsync:(void (^_Nullable)())completion
 {
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        [OWSOrphanedDataCleaner auditAndCleanup:YES completion:completion];
+//        [OWSOrphanedDataCleaner auditAndCleanup:YES completion:completion];
```
charlesmchen: ?

Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
michaelkirk: This is as I intended it. I don't think we want "exactly 4 lines", in the case of larger fonts - it would cover too much of the interface. 

I think we want "about 100 points worth", and made the slight optimization for the majority of our users who use default font size.

Signal/src/UIView+OWS.m
michaelkirk: Similar to the other `autoPin*ToSuperViewMargin` methods you'd added previously, this uses the layoutMargins Guide anchors rather than the edge.

The key is that iPhoneX adjusts the layoutMarginsGuide to reflect the areas that are not obscured by the home button.

Specifically, this method is used by the InputToolbar to make sure it appears above the home button on iPhoneX, and on other devices it has no real effect, since the layoutMarginsGuide.bottomAnchor is 0.

SignalServiceKit/src/Storage/OWSOrphanedDataCleaner.m
michaelkirk: Removed, thanks!

Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
michaelkirk: Whoops - it used to do more, but it's no loner necessary. Removed, thanks!

Signal/src/ViewControllers/ConversationView/ConversationViewLayout.m
michaelkirk: You're right. I can no longer reproduce the scenario that was leading to the subsequent assert failing. I've removed this code.

Signal/src/ViewControllers/MediaDetailViewController.m
michaelkirk: Yeah, great point. done!

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk: Forgot to re-enable this behavior. Done!

PR #2912, opened 2018-01-08
Collinstuart/update build documentation

CollinStuart:
### Description

Clarifies BUILDING documentation so that users can get up and running quickly. Explains what capabilities need to be enabled in order to run the app.


charlesmchen:
Merged in 6532a323, thank you @CollinStuart

PR #2914, opened 2018-01-09
Charlesmchen/backup

charlesmchen:
Implements full (as opposed to "partial", e.g. messages & threads-only) backup for ___internal debugging purposes only___.

I've done some light testing and it works so far.

PTAL @michaelkirk

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -31,6 +31,7 @@ end
 target 'Signal' do
   shared_pods
   pod 'ATAppUpdater', :inhibit_warnings => true
+  pod 'SSZipArchive', :inhibit_warnings => true
 
   target 'SignalTests' do
     inherit! :search_paths
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -130,12 +130,12 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
     SetRandFunctionSeed();
 
+    // XXX - careful when moving this. It must happen before we initialize TSStorageManager.
+    [self verifyDBKeysAvailableBeforeBackgroundLaunch];
+
     // If a backup restore is in progress, try to complete it.
     // Otherwise, cleanup backup state.
     [OWSBackup applicationDidFinishLaunching];
-
-    // XXX - careful when moving this. It must happen before we initialize TSStorageManager.
-    [self verifyDBKeysAvailableBeforeBackgroundLaunch];
 
     // Prevent the device from sleeping during database view async registration
     // (e.g. long database upgrades).
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -1,5 +1,5 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 #import "AppDelegate.h"
@@ -9,6 +9,7 @@
 #import "DebugLogger.h"
 #import "MainAppContext.h"
 #import "NotificationsManager.h"
+#import "OWSBackup.h"
 #import "OWSNavigationController.h"
 #import "Pastelog.h"
 #import "PushManager.h"
@@ -128,6 +129,10 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     [self startupLogging];
 
     SetRandFunctionSeed();
+
+    // If a backup restore is in progress, try to complete it.
+    // Otherwise, cleanup backup state.
+    [OWSBackup applicationDidFinishLaunching];
 
     // XXX - careful when moving this. It must happen before we initialize TSStorageManager.
     [self verifyDBKeysAvailableBeforeBackgroundLaunch];
--- a/Signal/src/ViewControllers/DebugUI/DebugUIMisc.m
+++ b/Signal/src/ViewControllers/DebugUI/DebugUIMisc.m
@@ -4,6 +4,7 @@
 
 #import "DebugUIMisc.h"
 #import "OWSBackup.h"
+#import "OWSBackupViewController.h"
 #import "OWSCountryMetadata.h"
 #import "OWSTableViewController.h"
 #import "RegistrationViewController.h"
@@ -94,10 +95,22 @@ NS_ASSUME_NONNULL_BEGIN
                                              [DebugUIMisc sendUnencryptedDatabase:thread];
                                          }]];
     }
-    [items addObject:[OWSTableItem itemWithTitle:@"Export Backup"
-                                     actionBlock:^{
-                                         [OWSBackup exportBackup];
-                                     }]];
+    [items addObject:[OWSTableItem
+                         subPageItemWithText:@"Export Backup w/ Password"
+                                 actionBlock:^(UIViewController *viewController) {
+                                     OWSBackupViewController *backupViewController = [OWSBackupViewController new];
+                                     [backupViewController exportBackup:thread skipPassword:NO];
+                                     [viewController.navigationController pushViewController:backupViewController
+                                                                                    animated:YES];
+                                 }]];
+    [items addObject:[OWSTableItem
+                         subPageItemWithText:@"Export Backup w/o Password"
+                                 actionBlock:^(UIViewController *viewController) {
+                                     OWSBackupViewController *backupViewController = [OWSBackupViewController new];
+                                     [backupViewController exportBackup:thread skipPassword:YES];
+                                     [viewController.navigationController pushViewController:backupViewController
+                                                                                    animated:YES];
+                                 }]];
 
     return [OWSTableSection sectionWithTitle:self.name items:items];
 }
--- a/Signal/src/ViewControllers/DebugUI/DebugUIMisc.m
+++ b/Signal/src/ViewControllers/DebugUI/DebugUIMisc.m
@@ -3,6 +3,7 @@
 //
 
 #import "DebugUIMisc.h"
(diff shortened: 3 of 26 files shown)


Podfile
```diff
@@ -31,6 +31,7 @@ end
 target 'Signal' do
   shared_pods
   pod 'ATAppUpdater', :inhibit_warnings => true
+  pod 'SSZipArchive', :inhibit_warnings => true
```
charlesmchen: There's a bunch of zip-for-iOS pods, but this one seems active, well-designed and it supports AES.

charlesmchen:
* For now, the only way to backup is through the Debug UI.  
* Backup files can be "shared", but it's more convenient to do them in the context of a thread and "send" them in that thread.
* Encrypted and unencrypted backups can be exported, but so far only unencrypted backups can be imported.
* For now the only way to restore/import a backup is to tap it in a thread.

Signal/src/ViewControllers/OWSBackupExportViewController.m
```diff
@@ -0,0 +1,311 @@
...
+#import <SignalMessaging/Environment.h>
+#import <SignalMessaging/SignalMessaging-Swift.h>
+#import <SignalMessaging/UIColor+OWS.h>
+#import <SignalMessaging/UIFont+OWS.h>
+#import <SignalMessaging/UIView+OWS.h>
+#import <SignalServiceKit/MIMETypeUtil.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OWSBackupExportViewController () <OWSBackupDelegate>
```
charlesmchen: The backup export & import VCs share a lot in common.  We could extract a common base class, or not.

charlesmchen:
simulator screen shot - iphone x - 2018-01-09 at 10 28 04


simulator screen shot - iphone x - 2018-01-09 at 10 27 58

simulator screen shot - iphone x - 2018-01-09 at 10 28 07

simulator screen shot - iphone x - 2018-01-09 at 10 28 10

simulator screen shot - iphone x - 2018-01-09 at 10 29 03

Signal/src/util/OWSBackup.m
```diff
@@ -0,0 +1,908 @@
...
+    OWSAssert(srcDirName.length > 0);
+    OWSAssert(dstDirPath.length > 0);
+    OWSAssert(self.backupDirPath.length > 0);
+
+    NSString *srcDirPath = [self.backupDirPath stringByAppendingPathComponent:srcDirName];
+
+    DDLogInfo(@"%@ restoreDirectoryContents: %@ -> %@", self.logTag, srcDirPath, dstDirPath);
+
+    if (![[NSFileManager defaultManager] fileExistsAtPath:srcDirPath]) {
+        DDLogInfo(@"%@ Skipping restore directory: %@.", self.logTag, srcDirPath);
```
michaelkirk: This seems like it should OWSFail -  or is there a normal reason that our backup would be missing a directory?

Signal/src/util/OWSBackup.m
```diff
@@ -0,0 +1,908 @@
...
+}
+
+#pragma mark - Utils
+
+- (NSString *)relativePathforPath:(NSString *)filePath basePath:(NSString *)basePath
+{
+    OWSAssert(filePath.stringByStandardizingPath.length > 0);
+    OWSAssert([filePath.stringByStandardizingPath hasPrefix:basePath]);
+
+    NSString *relativePath = [filePath.stringByStandardizingPath substringFromIndex:basePath.length];
```
michaelkirk: Should we `basePath.stringByStandardizingPath.length` ?

SignalServiceKit/src/Util/OWSFileSystem.h
```diff
@@ -8,7 +8,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype)init NS_UNAVAILABLE;
 
-+ (void)protectFolderAtPath:(NSString *)path;
+// TODO: We shouldn't ignore the return value of thie method.
++ (BOOL)protectFolderAtPath:(NSString *)path;
```
michaelkirk: Want to add `__attribute__((warn_unused_result))`?

charlesmchen:
> after restoring, won't I have a bunch of potentially outdated crypto sessions?
> Should we archive all sessions upon restore?

I was going to leave that "as an exercise for the user" to manually reset any outdated sessions in the interest of keeping backup narrow.

Signal/src/AppDelegate.m
```diff
@@ -129,6 +130,10 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
     SetRandFunctionSeed();
 
+    // If a backup restore is in progress, try to complete it.
+    // Otherwise, cleanup backup state.
+    [OWSBackup applicationDidFinishLaunching];
```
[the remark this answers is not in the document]
charlesmchen: Fair enough.

Signal/src/ViewControllers/DebugUI/DebugUIMisc.m
```diff
@@ -93,6 +95,24 @@ - (nullable OWSTableSection *)sectionForThread:(nullable TSThread *)thread
                                              [DebugUIMisc sendUnencryptedDatabase:thread];
                                          }]];
     }
+    [items addObject:[OWSTableItem
```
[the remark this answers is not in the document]
charlesmchen: Done.

Signal/src/ViewControllers/OWSBackupExportViewController.m
```diff
@@ -0,0 +1,311 @@
...
+    [button autoSetDimension:ALDimensionWidth toSize:140];
+    [button autoSetDimension:ALDimensionHeight toSize:kButtonHeight];
+    return button;
+}
+
+- (void)copyPassword
+{
+    OWSAssert(self.backup.backupPassword.length > 0);
+
+    [UIPasteboard.generalPasteboard setString:self.backup.backupPassword];
```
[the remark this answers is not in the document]
charlesmchen: I'm going to defer these questions until we take on user-facing backup, especially since encrypted backups are being disabled.

Signal/src/util/OWSBackup.m
```diff
@@ -0,0 +1,908 @@
...
+        return NO;
+    }
+    OWSAES256Key *encryptionKey = [OWSAES256Key keyWithData:encryptionKeyData];
+    if (!encryptionKey) {
+        return NO;
+    }
+    self.encryptionKey = encryptionKey;
+
+    NSError *error = nil;
+    BOOL success = [[NSFileManager defaultManager] removeItemAtPath:encryptionKeyFilePath error:&error];
```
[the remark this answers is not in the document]
charlesmchen: There are hard problems here that we'll need to confront if we do user-facing backup.  For debugging purposes, the approach here should be sufficient.

Restoring a backup involves multiple steps and there's no way to do them atomically.  Therefore there's always going to be a risk that an incomplete restore leaves you with an invalid install.  Worst case is database + password mismatch.  

In terms of interrupted restores, we can err on the side of making a best effort to resume or backing off as soon as any serious errors are detected.  I've chosen to do the latter.

Circling back to your original question, your comment is in the "restore preparation" logic, not the more fraught "restore completion" logic.  If "backup preparation" is interrupted, no restore takes place and no harm is done.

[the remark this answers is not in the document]
charlesmchen: If/when we do user-facing backup, if we only support restores post-registration, the stakes aren't that high if a restore is interrupted and the user is left with an invalid install since by definition they can just re-register and re-restore.

Signal/src/util/OWSBackup.m
```diff
@@ -0,0 +1,908 @@
...
+        [self joinRelativePath:queuedBackupRelativePath basePath:documentDirectoryPath];
+    if (![[NSFileManager defaultManager] fileExistsAtPath:queuedBackupPath]) {
+        OWSFail(@"%@ Missing import backup directory: %@.", self.logTag, queuedBackupPath);
+        return;
+    }
+    self.backupDirPath = queuedBackupPath;
+    self.backupState = OWSBackupState_InProgress;
+    DDLogInfo(@"%@ queuedBackupPath: %@", self.logTag, queuedBackupPath);
+
+    [SAMKeychain setAccessibilityType:kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly];
```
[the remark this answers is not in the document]
charlesmchen: > we could wrap SAMKeyChain with OWSKeyChain

Sure, I'll make a card.

Signal/src/util/OWSBackup.m
charlesmchen: Right, I'll add a comment:

```
Not all backups will have both a "app documents" and "shared data container" folder.  The latter should always be present for "modern" installs, but my intention is to be permissive here about what we accept so that we could easily apply this branch to historic (pre-shared data container) versions of the app and restore from them.
```

Signal/src/util/OWSBackup.m
charlesmchen: Done.

SignalServiceKit/src/Util/OWSFileSystem.h
charlesmchen: Done, and I've added a card to do an audit.

charlesmchen:
The remaining work here is:

* Rename all pre-existing files in "app documents" and "shared data container".
* Possibly clear pre-existing keys in `NSUserDefaults`.

charlesmchen:
PTAL @michaelkirk

michaelkirk:
LGTM

PR #2916, opened 2018-01-10
 Retain changes from session database branch.

charlesmchen:
Salvaging some work from the session database branch.

PTAL @michaelkirk


SignalMessaging/environment/AppSetup.m
```diff
@@ -40,6 +41,10 @@ + (void)setupEnvironment:(CallMessageHandlerBlock)callMessageHandlerBlock
                                                   profileManager:OWSProfileManager.sharedManager];
         [TextSecureKitEnv setSharedEnv:sharedEnv];
 
+        // Register renamed classes.
+        [NSKeyedUnarchiver setClass:[OWSUserProfile class] forClassName:[OWSUserProfile collection]];
+        [NSKeyedUnarchiver setClass:[OWSDatabaseMigration class] forClassName:[OWSDatabaseMigration collection]];
```
charlesmchen: Not strictly necessary, but I think it'll be convenient to ensure `NSKeyedUnarchiver` knows how to load any entity in YDB at all times.

SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -1,5 +1,5 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 #import "OWSBackgroundTask.h"
```
charlesmchen: We could refactor `OWSBackgroundTask` to only create a single background task from the POV of iOS.

Signal/src/views/MarqueeLabel.swift
```diff
@@ -1,10 +1,5 @@
-//  Grabbed from: https://github.com/cbpowell/MarqueeLabel-Swift/blob/cd331f3cfc3f9d7114ffa5aa4f243f1d5eda9d0d/Classes/MarqueeLabel.swift
-//  License: MIT License
 //
-//  MarqueeLabel.swift
-//
-//  Created by Charles Powell on 8/6/14.
-//  Copyright (c) 2015 Charles Powell. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
```
michaelkirk: We should retain the existing copyright content here.

SignalMessaging/environment/AppSetup.m
michaelkirk: 👍 

I think this makes sense - but to make sure I understand, whenever a class stored in Yap is renamed we should register a legacy handler here?

I guess the exception being collections that have more than one object type stored in them, which off the top of my head is only the thread collection (where TSContactThread and TSGroupThread cohabit), which I don't imagine we'll be renaming any time soon.

SignalServiceKit/src/Network/API/TSNetworkManager.m
```diff
@@ -56,6 +60,17 @@ - (void)makeRequest:(TSRequest *)request
             failure:(void (^)(NSURLSessionDataTask *task, NSError *error))failureBlock
 {
     DDLogInfo(@"%@ Making request: %@", self.logTag, request);
+    if (!CurrentAppContext().isMainApp) {
+        // TODO: Discuss which of these requests to suppress.
+        if (![request isKindOfClass:[TSRecipientPrekeyRequest class]]
+            && ![request isKindOfClass:[TSSubmitMessageRequest class]]
+            && ![request isKindOfClass:[OWSGetProfileRequest class]]
```
michaelkirk: per recent changes, I don't think we should be making profile requests either.

SignalMessaging/environment/AppSetup.m
charlesmchen: > whenever a class stored in Yap is renamed we should register a legacy handler here?

That's right.  We don't have to be strict about it, but it provides some safety and convenience.

> I guess the exception being collections that have more than one object type stored in them,

I could be wrong, but I think there's a distinction between the collection name and the class name.  I'm just using the `collection` property of these classes as a convenience, since the collection name and class name coincided before these classes were renamed.  

Let's consider a worst case.  Suppose we stored class A and class B in a shared collection C and we renamed A -> A2 and B -> B2, we would then:

* Make sure `A2.collection` and `B2.collection` both returned C.
* `[NSKeyedUnarchiver setClass:[A2 class] forClassName:@"A"];
* `[NSKeyedUnarchiver setClass:[B2 class] forClassName:@"B"];

Right?

Signal/src/views/MarqueeLabel.swift
charlesmchen: I'm going to revert everything here except the notification changes. clang-format went wild.

SignalServiceKit/src/Network/API/TSNetworkManager.m
charlesmchen: Removed.

SignalMessaging/environment/AppSetup.m
michaelkirk: That sounds right to me.

PR #2917, opened 2018-01-10
Skip redundant sync messages.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/OWSLinkDeviceViewController.m
+++ b/Signal/src/ViewControllers/OWSLinkDeviceViewController.m
@@ -1,5 +1,5 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 #import "OWSLinkDeviceViewController.h"
@@ -147,7 +147,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)provisionWithParser:(OWSDeviceProvisioningURLParser *)parser
 {
     // Optimistically set this flag.
-    [OWSDeviceManager.sharedManager setMayHaveLinkedDevices:YES dbConnection:self.dbConnection];
+    [OWSDeviceManager.sharedManager setMayHaveLinkedDevices];
 
     ECKeyPair *_Nullable identityKeyPair = [[OWSIdentityManager sharedManager] identityKeyPair];
     OWSAssert(identityKeyPair);
--- a/SignalServiceKit/src/Devices/OWSDevice.m
+++ b/SignalServiceKit/src/Devices/OWSDevice.m
@@ -19,6 +19,7 @@ NSString *const kTSStorageManager_MayHaveLinkedDevices = @"kTSStorageManager_May
 
 @interface OWSDeviceManager ()
 
+@property (atomic, nullable) NSNumber *mayHaveLinkedDevicesCached;
 @property (atomic) NSDate *lastReceivedSyncMessage;
 
 @end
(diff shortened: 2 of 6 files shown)


SignalServiceKit/src/Devices/OWSDevice.m
```diff
@@ -52,7 +52,7 @@ - (BOOL)mayHaveLinkedDevices:(YapDatabaseConnection *)dbConnection
         if (!self.mayHaveLinkedDevicesCached) {
             self.mayHaveLinkedDevicesCached = @([dbConnection boolForKey:kTSStorageManager_MayHaveLinkedDevices
                                                             inCollection:kTSStorageManager_OWSDeviceCollection
-                                                            defaultValue:YES]);
+                                                            defaultValue:NO]);
```
charlesmchen: We want to send a sync message in one of two cases:

* We know that we have linked devices by inspecting `SignalRecipient.devices` for our own instance of `SignalRecipient`.
* There might be a new linked device that we've just linked.  This flag is intended to reflect the latter.

SignalServiceKit/src/Devices/OWSDevice.m
```diff
@@ -19,7 +19,6 @@
 
 @interface OWSDeviceManager ()
 
-@property (atomic, nullable) NSNumber *mayHaveLinkedDevicesCached;
```
charlesmchen: I'm increasingly skeptical of the value of caching values from the db.

SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -278,6 +278,7 @@ - (void)handleEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
             [self handleIncomingEnvelope:envelope withSyncMessage:content.syncMessage transaction:transaction];
 
             [[OWSDeviceManager sharedManager] setHasReceivedSyncMessage];
+            [OWSDeviceManager.sharedManager setMayHaveLinkedDevices:YES transaction:transaction];
```
charlesmchen: In the interest of caution, I'm also setting this flag whenever we receive a sync message.

Outcome:

Phone will _erroneously_ not send sync message IFF all of the following are true:

* Phone previously (with an older version) added a linked device but never sent a sync message (this should be extremely rare).
* The linked device never sends a sync message to the phone.

That should be extremely rare - and we should recover as soon as the user does anything significant on linked device.

charlesmchen:
I've reworked this significantly. Please take another look.

SignalServiceKit/src/Devices/OWSDevice.m
```diff
@@ -47,32 +46,27 @@ - (BOOL)mayHaveLinkedDevices:(YapDatabaseConnection *)dbConnection
 {
     OWSAssert(dbConnection);
```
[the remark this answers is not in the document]
charlesmchen: I've reworked this significantly. Please take another look.

Signal/src/ViewControllers/OWSLinkDeviceViewController.m
```diff
@@ -147,7 +147,7 @@ - (void)controller:(OWSQRCodeScanningViewController *)controller didDetectQRCode
 - (void)provisionWithParser:(OWSDeviceProvisioningURLParser *)parser
 {
     // Optimistically set this flag.
-    [OWSDeviceManager.sharedManager setMayHaveLinkedDevices:YES dbConnection:self.dbConnection];
+    [OWSDeviceManager.sharedManager setMayHaveLinkedDevices];
```
charlesmchen: We only ever set this flag.

PR #2919, opened 2018-01-10
Fix media presentation

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MediaDetailViewController.m
+++ b/Signal/src/ViewControllers/MediaDetailViewController.m
@@ -49,6 +49,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic) UIScrollView *scrollView;
 @property (nonatomic) UIView *mediaView;
+@property (nonatomic) UIView *presentationView;
 
 @property (nonatomic) UIButton *shareButton;
 
@@ -68,7 +69,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, nullable) UIBarButtonItem *videoPlayBarButton;
 @property (nonatomic, nullable) UIBarButtonItem *videoPauseBarButton;
 
-@property (nonatomic, nullable) NSArray<NSLayoutConstraint *> *mediaViewConstraints;
+@property (nonatomic, nullable) NSArray<NSLayoutConstraint *> *presentationViewConstraints;
 @property (nonatomic, nullable) NSLayoutConstraint *mediaViewBottomConstraint;
 @property (nonatomic, nullable) NSLayoutConstraint *mediaViewLeadingConstraint;
 @property (nonatomic, nullable) NSLayoutConstraint *mediaViewTopConstraint;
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/MediaDetailViewController.m
```diff
@@ -44,6 +44,9 @@ - (BOOL)canPerformAction:(SEL)action withSender:(nullable id)sender
 
 #pragma mark -
 
+// This approximates the curve of our message bubbles, which makes the animation feel a little smoother.
+const CGFloat MediaDetailViewControllerMediaCornerRadius = 17;
```
charlesmchen: Can you please pull out a constant in `OWSMessageCell` and use it here?
```
        maskedSubview.layer.cornerRadius = 17;
```

Signal/src/ViewControllers/MediaDetailViewController.m
```diff
@@ -725,19 +732,21 @@ - (void)presentFromViewController:(UIViewController *)viewController
...
+                                   [self applyFinalMediaViewConstraints];
 
                                    // 2. Animate imageView from it's initial position, which should match where it was
                                    // in the presenting view to it's final position, front and center in this view. This
-                                   // animation intentionally overlaps the previous
-                                   [UIView animateWithDuration:0.2
-                                       delay:0.08
+                                   // animation duration intentionally overlaps the previous
+                                   [UIView animateWithDuration:2.2
+                                       delay:2.08
```
charlesmchen: Was this just slowed down for debugging?  2 seconds delay/duration seems very long?

charlesmchen:
It'd be nice to have screenshots or an animated GIF, but NBD.

Signal/src/ViewControllers/MediaDetailViewController.m
michaelkirk: Yikes, good catch. Thanks.

Signal/src/ViewControllers/MediaDetailViewController.m
michaelkirk: Yes, done.

michaelkirk:
@charlesmchen I've significantly reworked this to address a couple issues - primarily in 
https://github.com/WhisperSystems/Signal-iOS/pull/2919/commits/3d99f1fe4c443a357782206946c7a2579aa7f17b PTAL

Signal/src/ViewControllers/MediaDetailViewController.m
```diff
@@ -277,7 +280,15 @@ - (void)createContents
     self.mediaView.layer.minificationFilter = kCAFilterTrilinear;
     self.mediaView.layer.magnificationFilter = kCAFilterTrilinear;
 
-    [self applyInitialImageViewConstraints];
+    // The presentationView is only used during present/dismiss animations.
+    // It's a static image of the media content.
+    UIImageView *presentationView = [[UIImageView alloc] initWithImage:self.image];
```
charlesmchen: But isn't `self.image` only true for images?  What about videos?

charlesmchen: gotcha, `image` is a method not a property.

PR #2920, opened 2018-01-11
Retain changes from session database branch.

charlesmchen:
I decided to retain more changes from the abandoned "database workaround" branch, since they'll be useful if we do user-facing backup.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -295,11 +295,11 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(yapDatabaseModified:)
                                                  name:YapDatabaseModifiedNotification
-                                               object:nil];
+                                               object:TSStorageManager.sharedManager.dbNotificationObject];
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(yapDatabaseModifiedExternally:)
                                                  name:YapDatabaseModifiedExternallyNotification
-                                               object:nil];
+                                               object:TSStorageManager.sharedManager.dbNotificationObject];
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(applicationWillEnterForeground:)
                                                  name:OWSApplicationWillEnterForegroundNotification
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -299,7 +299,7 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(yapDatabaseModifiedExternally:)
                                                  name:YapDatabaseModifiedExternallyNotification
-                                               object:TSStorageManager.sharedManager.dbNotificationObject];
+                                               object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self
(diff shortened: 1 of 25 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -295,7 +295,7 @@ - (void)addNotificationListeners
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(yapDatabaseModified:)
                                                  name:YapDatabaseModifiedNotification
-                                               object:nil];
+                                               object:TSStorageManager.sharedManager.dbNotificationObject];
```
charlesmchen: If we do user-facing backup, we'll probably export (a subset of) the messages, attachments and groups to another instance of `OWSStorage`.  We'd want to ignore any modification events that result.

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.h
```diff
@@ -1,19 +1,19 @@
...
+@class OWSStorage;
 
 // This class is used to write incoming (decrypted, unprocessed)
 // messages to a durable queue and then process them in batches,
 // in the order in which they were received.
 @interface OWSBatchMessageProcessor : NSObject
 
 + (instancetype)sharedInstance;
-+ (void)syncRegisterDatabaseExtension:(TSStorageManager *)storageManager;
++ (void)syncRegisterDatabaseExtension:(OWSStorage *)storage;
```
charlesmchen: We should be able to register our views on any "storage", e.g. both our primary and an "export" storage.

SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -98,29 +125,18 @@ - (void)runAsyncRegistrationsWithCompletion:(void (^_Nonnull)(void))completion
...
-    [OWSDisappearingMessagesFinder asyncRegisterDatabaseExtensions:self];
-    [OWSFailedMessagesJob asyncRegisterDatabaseExtensionsWithStorageManager:self];
-    [OWSFailedAttachmentDownloadsJob asyncRegisterDatabaseExtensionsWithStorageManager:self];
+    runAsyncRegistrationsForPrimaryStorage(self);
 
     // Block until all async registrations are complete.
-    [self.newDatabaseConnection asyncReadWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
-        OWSAssert(!self.areAsyncRegistrationsComplete);
+    YapDatabaseConnection *dbConnection = self.newDatabaseConnection;
+    [dbConnection flushTransactionsWithCompletionQueue:dispatch_get_main_queue()
```
charlesmchen: A small refinement: `flushTransactionsWithCompletionQueue` instead of a NOOP transaction.

SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -22,8 +23,41 @@
 NSString *const TSStorageManagerExceptionName_CouldNotCreateDatabaseDirectory
     = @"TSStorageManagerExceptionName_CouldNotCreateDatabaseDirectory";
 
-#pragma mark -
+void runSyncRegistrationsForPrimaryStorage(OWSStorage *storage)
```
michaelkirk: nit: the term `primary` feels kind of oblique given there is no longer "secondary" storage.

SignalServiceKit/src/Storage/TSStorageManager.m
michaelkirk: Cool, TIL.

SignalServiceKit/src/Storage/TSStorageManager.m
charlesmchen: Removed.

PR #2923, opened 2018-01-12
Add debug UI actions around clearing and snapshotting session state.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
+++ b/Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
@@ -80,9 +80,9 @@ NS_ASSUME_NONNULL_BEGIN
                         actionBlock:^{
                             [DebugUISessionState clearSessionAndIdentityStore];
                         }],
-        [OWSTableItem itemWithTitle:@"Archive Session and Identity Store"
+        [OWSTableItem itemWithTitle:@"Snapshot Session and Identity Store"
                         actionBlock:^{
-                            [DebugUISessionState archiveSessionAndIdentityStore];
+                            [DebugUISessionState snapshotSessionAndIdentityStore];
                         }],
         [OWSTableItem itemWithTitle:@"Restore Session and Identity Store"
                         actionBlock:^{
@@ -105,12 +105,12 @@ NS_ASSUME_NONNULL_BEGIN
     });
 }
 
-+ (void)archiveSessionAndIdentityStore
++ (void)snapshotSessionAndIdentityStore
 {
     dispatch_async([OWSDispatch sessionStoreQueue], ^{
-        [[TSStorageManager sharedManager] archiveSessionStore];
+        [[TSStorageManager sharedManager] snapshotSessionStore];
         dispatch_async(dispatch_get_main_queue(), ^{
-            [[OWSIdentityManager sharedManager] archiveIdentityState];
+            [[OWSIdentityManager sharedManager] snapshotIdentityState];
         });
     });
 }
--- a/Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
+++ b/Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
@@ -1,5 +1,5 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 #import "DebugUISessionState.h"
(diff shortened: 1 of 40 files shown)


Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
```diff
@@ -20,63 +20,112 @@ - (NSString *)name
 
 - (nullable OWSTableSection *)sectionForThread:(nullable TSThread *)threadParameter
```
charlesmchen: Modified `DebugUISessionState` to have per-thread and no-thread modes, like the other Debug UI pages.

Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
```diff
@@ -20,63 +20,112 @@ - (NSString *)name
...
-                                           [OWSSessionResetJob runWithContactThread:thread
-                                                                      messageSender:[Environment current].messageSender
-                                                                     storageManager:[TSStorageManager sharedManager]];
-                                       }]
-                   ]];
+    return [OWSTableSection sectionWithTitle:self.name items:items];
 }
 
+#if DEBUG
++ (void)clearSessionAndIdentityStore
```
charlesmchen: Blows away everything except our own identity key.

Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
```diff
@@ -20,63 +20,112 @@ - (NSString *)name
...
+{
+    dispatch_async([OWSDispatch sessionStoreQueue], ^{
+        [[TSStorageManager sharedManager] resetSessionStore];
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [[OWSIdentityManager sharedManager] clearIdentityState];
+        });
+    });
+}
+
++ (void)archiveSessionAndIdentityStore
```
charlesmchen: Takes a snapshot of session and identity state.

Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
```diff
@@ -20,63 +20,112 @@ - (NSString *)name
...
+{
+    dispatch_async([OWSDispatch sessionStoreQueue], ^{
+        [[TSStorageManager sharedManager] archiveSessionStore];
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [[OWSIdentityManager sharedManager] archiveIdentityState];
+        });
+    });
+}
+
++ (void)restoreSessionAndIdentityStore
```
charlesmchen: Restores the last snapshot (if any) of session and identity state.

Signal/translations/ar.lproj/Localizable.strings
```diff
@@ -665,7 +665,7 @@
 "FINGERPRINT_SCAN_VERIFY_BUTTON" = "وضع علامة \"تم التحقق\"";
 
 /* No comment provided by engineer. */
-"FINGERPRINT_SHRED_KEYMATERIAL_BUTTON" = "إعادة تعيين هذه الجلسة.";
+"FINGERPRINT_SHRED_KEYMATERIAL_BUTTON" = "إعادة تعيين هذه الجلسة";
```
charlesmchen: "Reset Session" button label has a trailing period, which is wrong.  But maybe we should just leave it, rather than force re-translation in all of these languages.

SignalServiceKit/src/Account/TSAccountManager.m
```diff
@@ -93,7 +93,9 @@ - (void)resetForRegistration
             [transaction removeAllObjectsInCollection:TSAccountManager_UserAccountCollection];
         }];
     }
-    [[TSStorageManager sharedManager] resetSessionStore];
+    dispatch_async([OWSDispatch sessionStoreQueue], ^{
+        [[TSStorageManager sharedManager] resetSessionStore];
```
charlesmchen: PTAL @michaelkirk 

A small but important change.

charlesmchen: Actually, maybe this should dispatch sync.

michaelkirk: Good catch! Can you add an AssertIsOnSessionStoreQueue() to the underlying method?

I only use dispatch sync if it's difficult to add a completion handler, but it seems like adding a completion handler should be straight forward here.

Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
michaelkirk: Just a naming nit, but "archive" is kind of a reserved word when talking about sessions. I assumed this method did something like the similarly named `archiveAllSessions` method

Can we instead use the term "snapshot" unless we're talking about archiving in the sense of `[session archive]`;

SignalServiceKit/src/Messages/OWSIdentityManager.m
```diff
@@ -745,6 +747,58 @@ - (void)saveChangeMessagesForRecipientId:(NSString *)recipientId
...
+            [transaction removeObjectForKey:key inCollection:TSStorageManagerIdentityKeyStoreCollection];
+        }
+        [transaction removeAllObjectsInCollection:TSStorageManagerTrustedKeysCollection];
+    }];
+}
+
+- (NSString *)identityKeySnapshotFilePath
+{
+    NSString *dirPath = [OWSFileSystem appDocumentDirectoryPath];
+    return [dirPath stringByAppendingPathComponent:@".identity-key-snapshot"];
```
michaelkirk: Do we need to ensure these aren't backed up?

Signal/src/ViewControllers/DebugUI/DebugUISessionState.m
charlesmchen: Good point, done.

SignalServiceKit/src/Account/TSAccountManager.m
charlesmchen: > Good catch! Can you add an AssertIsOnSessionStoreQueue() to the underlying method?

Done.

> Actually, maybe this should dispatch sync.

I've changed my mind - I don't think there's any point in blocking on this.

SignalServiceKit/src/Messages/OWSIdentityManager.m
charlesmchen: OWSBackup's "full backups" backs up files and directories in the "app documents" and "shared data container", but ignores any top-level files or directories in those locations whose names start with a period ".".  

I've added comments to that effect everywhere I'm taking advantage of that.

Note that iOS itself also creates at least one file/directory in these containers which:

* can't/shouldn't be backed up (or otherwise accessed by apps).
* have a period "." prefix to their name.

So I've just extended that convention.

michaelkirk: To be clear, I'm wondering if we need to do something like `protectFileOrFolderAtPath` on these files

PR #2924, opened 2018-01-12
Handle unregistration in SAE.

charlesmchen:
SAE should stop work immediately if user unregisters in main app.

PTAL @michaelkirk

### DIFF
--- a/SignalShareExtension/ShareViewController.swift
+++ b/SignalShareExtension/ShareViewController.swift
@@ -15,7 +15,7 @@ public class ShareViewController: UINavigationController, ShareViewDelegate, SAE
     enum ShareViewControllerError: Error {
         case assertionError(description: String)
         case unsupportedMedia
-        case NotRegistered(description: String)
+        case notRegistered()
     }
 
     private var hasInitialRootViewController = false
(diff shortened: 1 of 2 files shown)


SignalShareExtension/ShareViewController.swift
```diff
@@ -12,6 +12,12 @@ import PromiseKit
 @objc
 public class ShareViewController: UINavigationController, ShareViewDelegate, SAEFailedViewDelegate {
 
+    enum ShareViewControllerError: Error {
+        case assertionError(description: String)
+        case unsupportedMedia
+        case NotRegistered(description: String)
```
michaelkirk: style nit, enum should be lowercased. 

Does this need a description? Seems to speak for itself.

charlesmchen:
> What does the user see in this scenario?

The SAE should throw an uncaught exception which should cause it to close.  Not perfect, but this is a true edge case and I think that behavior should be acceptable.

michaelkirk:
> The SAE should throw an uncaught exception which should cause it to close. Not perfect, but this is a true edge case and I think that behavior should be acceptable.

I agree this is pretty edge case, just wasn't clear what would happen.

Presumably after crashing, the next time it's launched they'd see the "Not ready /please launch the main app" dialog?


LGTM

SignalShareExtension/ShareViewController.swift

charlesmchen:
> ...after crashing, the next time it's launched they'd see the "Not ready /please launch the main app" dialog?

Yup.

PR #2932, opened 2018-01-16
Remove spurious warning.

charlesmchen:
Having looked at the callers of this method, I don't think we need this warning.  I didn't realize we had an `OWSFail()` within the method itself.

PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Storage/OWSStorage.m
+++ b/SignalServiceKit/src/Storage/OWSStorage.m
@@ -263,6 +263,7 @@ typedef NSData *_Nullable (^CreateDatabaseMetadataBlock)(void);
             //
             // The best we can try to do is to discard the current database
             // and behave like a clean install.
+            OWSFail(@"%@ Could not load database", self.logTag);
             OWSProdCritical([OWSAnalyticsEvents storageErrorCouldNotLoadDatabase]);
 
             // Try to reset app by deleting all databases.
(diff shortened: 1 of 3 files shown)


michaelkirk:
>  I didn't realize we had an OWSFail() within the method itself.

Within which method? It seems like we call this in several places, some of which would silently ignore if this returned NO.

charlesmchen:
> Within which method? 

Within `protectFileOrFolderAtPath`.

> It seems like we call this in several places, some of which would silently ignore if this returned NO.

I've audited those call sites and I don't see a single place where we would want to do something different if this failed in production.  Happy to discuss on a call.

michaelkirk:
>> Within which method?
>
>Within protectFileOrFolderAtPath.

That's what I assumed - but I don't see an assert here:

```
+ (BOOL)protectFileOrFolderAtPath:(NSString *)path
{
    if (![NSFileManager.defaultManager fileExistsAtPath:path]) {
        return NO;
    }

    NSError *error;
    NSDictionary *fileProtection = @{ NSFileProtectionKey : NSFileProtectionCompleteUntilFirstUserAuthentication };
    [[NSFileManager defaultManager] setAttributes:fileProtection ofItemAtPath:path error:&error];

    NSDictionary *resourcesAttrs = @{ NSURLIsExcludedFromBackupKey : @YES };

    NSURL *ressourceURL = [NSURL fileURLWithPath:path];
    BOOL success = [ressourceURL setResourceValues:resourcesAttrs error:&error];

    if (error || !success) {
        OWSProdCritical([OWSAnalyticsEvents storageErrorFileProtection]);
        return NO;
    }
    return YES;
}
```

Do you want to add one?

> I've audited those call sites and I don't see a single place where we would want to do something different if this failed in production. Happy to discuss on a call.

Ok, then LGTM

charlesmchen:
> That's what I assumed - but I don't see an assert here:

Oh, I thought `OWSProdCritical()` included an `OWSFail()`.  Perhaps it should.  For now, I've added an `OWSFail()` everywhere that we use `OWSProdCritical()`.

PR #2935, opened 2018-01-16
RI 2.19.4.4 hotfix

michaelkirk:
This wasn't quite trivial since 2.19.4.4 had substantial changes *and* so much has moved around in master. Sorry =(

Most of the changes were straight forward, but there were two significant changes.

1. In the hotfix I had the MediaMessageView gain a reference to a ConversationViewItem so that it could plumb that down through to the MediaDetailView. (The MediaDetailView uses the ConversationViewItem as an easy way to perform editing actions like copy/delete/save)

However, in master MediaMessageView is now in SignalMessage framework, whereas ConversationViewItem remains in the Signal app. The ConversationViewItem has a lot of other dependencies not in the SignalMessaging framework, so I'd rather not move it and all it's dependencies into the SignalMessaging framework. Instead a new delegate pattern is responsible for wiring the ConversationViewItem to the MediaDetailView directly. 

This change should not affect the user experience in any way. This change was made during the merge conflict resolution, so it's somewhat lost in the large diff. I'll try to highlight it inline with some comments.

2. The other significant change does affect user experience. As part of the hotfix, we remove the video-playing faculties of the MediaMessageView, which presented a legacy MPMoviePlayer. Instead, we now play the video in line in the MediaDetailViewController. 

So I've made a similar change in the AttachmentApprovalViewController. It now plays videos inline, without presenting an MPMoviePlayerViewController (unless you're on iOS8). 

This second change is in a separate commit, as it wasn't strictly a merge conflict: 94d58b8


Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -12,7 +12,7 @@ enum MessageMetadataViewMode: UInt {
     case focusOnMetadata
 }
 
-class MessageDetailViewController: OWSViewController, UIScrollViewDelegate {
+class MessageDetailViewController: OWSViewController, UIScrollViewDelegate, MediaDetailPresenter {
```
michaelkirk: MediaDetailPresenter is the new protocol added during merge resolution.

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -442,7 +442,8 @@ class MessageDetailViewController: OWSViewController, UIScrollViewDelegate {
         let contentType = attachment.contentType
         if let dataUTI = MIMETypeUtil.utiType(forMIMEType: contentType) {
             let attachment = SignalAttachment.attachment(dataSource: dataSource, dataUTI: dataUTI, imageQuality: .original)
-            let mediaMessageView = MediaMessageView(attachment: attachment, mode: .small)
+            let mediaMessageView = MediaMessageView(attachment: attachment, mode: .small, mediaDetailPresenter: self)
```
michaelkirk: In the hotfix, we passed the `attachmentStream:viewItem` instead of `mediaDetailPresenter:self`

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -751,4 +752,18 @@ class MessageDetailViewController: OWSViewController, UIScrollViewDelegate {
 
         updateTextLayout()
     }
+
+    // MARK: MediaDetailPresenter
+
+    public func presentDetails(mediaMessageView: MediaMessageView, fromView: UIView) {
```
michaelkirk: This method was added during merge resolution. This is how we avoid passing the attachmentStream/viewItem to the mediaMessageView.

SignalMessaging/attachments/MediaMessageView.swift
```diff
@@ -463,14 +472,8 @@ public class MediaMessageView: UIView, OWSAudioAttachmentPlayerDelegate {
...
         }
-        guard let fromViewController = CurrentAppContext().frontmostViewController() else {
-            return
-        }
 
-        let window = CurrentAppContext().rootReferenceView
-        let convertedRect = fromView.convert(fromView.bounds, to:window)
-        let viewController = FullImageViewController(attachment:attachment, from:convertedRect)
-        viewController.present(from:fromViewController)
+        showMediaDetailViewController(fromView: fromView)
```
michaelkirk: We no longer present the full screen view controller (no called the MediaDetailViewController) directly, instead delegating to a protocol.

Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
```diff
@@ -69,18 +71,25 @@ - (void)dealloc
     [self removeKVOObservers];
 }
 
+- (CGSize)intrinsicContentSize
+{
+    CGSize newSize = CGSizeMake(self.bounds.size.width, self.toolbarHeight + ConversationInputToolbarBorderViewHeight);
+    return newSize;
```
charlesmchen: It might be helpful to add a comment about why this is necessary.

michaelkirk: Will do.

PR #2938, opened 2018-01-17
fullscreen approval/captioning view

michaelkirk:
Reuse the approval view form the share extension as a captioning screen in the main app.

To me there is one glaring issue with this approach as it currently stands affecting picking videos:

video-confirmation-low-res

After picking a video, the UIImagePicker shows it's own a playback/confirmation screen before notifying the delegate of the picked media. (it only does this with video/GIFs, not static images). 

After confirming the video selection in that ImagePicker screen, our delegate is notified, and we show our own approval view (with captioning). The first screen is redundant. It would be nice to just skip the UIImagePicker confirmation screen, and use the caption/approval view as the single confirmation. But it doesn't seem possible, short of creating our own custom image picker experience, which could have some other upsides, but would significantly increase the scope of this work.

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.h
+++ b/Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.h
@@ -22,10 +22,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)voiceMemoGestureDidChange:(CGFloat)cancelAlpha;
 
-#pragma mark - Attachment Approval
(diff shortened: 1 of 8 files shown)


michaelkirk:
This one is now ready for review PTAL @charlesmchen

charlesmchen:
Thanks for that GIF.  Looking at it reminds me how much I want to improve the appearance of `ModalActivityIndicatorViewController`.

Issue #2895, opened 2017-12-21
Unable to select/attach photo (iOS/iPhone 6s)


srippon:
- [x] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description
Unable to attach a photo to message.

### Steps to reproduce
- Open conversation
- Select paper clip and select 'Photo Library'
- Select and album (e.g. 'Camera Roll') and select a photo

**Actual result:** When an image is selected nothing happens.
**Expected result:** When an image is selected it returns to the conversation and inserts the selected photo into the conversation.

### Device info
**Device**: iPhone 6s (MKQM2X/A)
**iOS version**: 11.2.1 (15C153)
**Signal version**: 2.19.3.3

michaelkirk:
Please include a debug log.

giucal:
Same here. iPhone 6s, too. Same versions of Signal and iOS.

But:

I noticed that just *selecting* GIFs and videos works, but you can’t attach them. In the next screen, you can hit Cancel and go back, unless you hit Choose: then the screen freezes.

Here’s a log of me
- trying to attach an image;
- trying to attach a GIF (screen “froze”, quit Signal);
- trying to attach a video (idem).

https://gist.github.com/8c9ee1181902866aedeccc2998043827

giucal:
Well, in my case, this turned out to be just a matter of permissions... :man_facepalming:

@srippon, be sure that Signal is granted access to the photo library (check in **Settings > Privacy**).

michaelkirk:
Oh excellent detail - thanks @giucal. I'm sure we can do something better now that we have a reproduction.

srippon:
@giucal thanks it was an issue with my privacy settings. :man_facepalming: 

@michaelkirk, to improve the usability, would it be possible for Signal to warn people that it doesn't have the appropriate permission to add an image from the Photo Library and give them some helpful info on how to update their permissions?  Thanks.

michaelkirk:
Yup, that sounds like the right fix.

If someone gets to it sooner, we'd happily merge a PR. For UX guidelines, see how we handle sending a voice note when you have disabled mic permissions.

michaelkirk:
This is fixed in #2939

PR #2939, opened 2018-01-17
ensure media permissions

michaelkirk:
Same pattern as we use for camera/mic access. 

Fixes https://github.com/WhisperSystems/Signal-iOS/issues/2895


michaelkirk:
This is rebased and ready for review. 

PTAL @charlesmchen

michaelkirk:
in e0ea3921

PR #2942, opened 2018-01-17
block vs share

michaelkirk:
We were showing a "blocked" label, but not actually preventing the user from trying to send to a blocked conversation. What would happen is that after captioning the attachment when you actually tried to send, you'd get a dialog like "Can't send to Michael because he is blocked."

Now when you tap a blocked conversation, you are met with the familiar "Unblock?" actionsheet before you can proceed.

unblock


charlesmchen:
@michaelkirk Is this PR ready for review?

michaelkirk:
Rebased. PTAL @charlesmchen

PR #2955, opened 2018-01-25
Charlesmchen/sae vs file types2 

charlesmchen:
PTAL @michaelkirk


SignalShareExtension/ShareViewController.swift
```diff
@@ -608,6 +608,19 @@ public class ShareViewController: UINavigationController, ShareViewDelegate, SAE
                 } else {
                     fulfill((itemUrl: url, utiType: srcUtiType))
                 }
+            } else if let image = value as? UIImage {
+                if let data = UIImagePNGRepresentation(image) {
+                    let tempFilePath = OWSFileSystem.temporaryFilePath(withFileExtension:"png")
+                    do {
+                        let url = NSURL.fileURL(withPath:tempFilePath)
+                        try data.write(to: url, options: .atomicWrite)
```
michaelkirk: nit: Why does this need to be an atomic write? We're using a random one-time-use file path, not replacing something.

PR #2959, opened 2018-01-26
Elaborate Debug & Internal UI.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/AppSettingsViewController.m
+++ b/Signal/src/ViewControllers/AppSettingsViewController.m
@@ -108,10 +108,16 @@
 - (void)updateTableContents
 {
     OWSTableContents *contents = [OWSTableContents new];
-    OWSTableSection *section = [OWSTableSection new];
 
     __weak AppSettingsViewController *weakSelf = self;
 
+#ifdef INTERNAL
+    OWSTableSection *internalSection = [OWSTableSection new];
+    [section addItem:[OWSTableItem softCenterLabelItemWithText:@"Internal Build"]];
+    [contents addSection:internalSection];
+#endif
+
+    OWSTableSection *section = [OWSTableSection new];
     [section addItem:[OWSTableItem itemWithCustomCellBlock:^{
         return [weakSelf profileHeaderCell];
     }
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2834,23 +2834,26 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
     // We need to reload any modified interactions _before_ we call
     // reloadViewItems.
     BOOL hasDeletions = NO;
+    BOOL hasMalformedRowChange = NO;
     for (YapDatabaseViewRowChange *rowChange in rowChanges) {
         switch (rowChange.type) {
             case YapDatabaseViewChangeUpdate: {
                 YapCollectionKey *collectionKey = rowChange.collectionKey;
-                OWSAssert(collectionKey.key.length > 0);
                 if (collectionKey.key) {
                     ConversationViewItem *viewItem = self.viewItemCache[collectionKey.key];
                     [self reloadInteractionForViewItem:viewItem];
+                } else {
+                    hasMalformedRowChange = YES;
                 }
                 break;
             }
             case YapDatabaseViewChangeDelete: {
                 // Discard cached view items after deletes.
                 YapCollectionKey *collectionKey = rowChange.collectionKey;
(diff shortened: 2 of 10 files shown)


Signal/src/ViewControllers/AppSettingsViewController.m
```diff
@@ -108,10 +108,16 @@ - (void)dealloc
 - (void)updateTableContents
 {
     OWSTableContents *contents = [OWSTableContents new];
-    OWSTableSection *section = [OWSTableSection new];
 
     __weak AppSettingsViewController *weakSelf = self;
 
+#ifdef INTERNAL
+    OWSTableSection *internalSection = [OWSTableSection new];
+    [section addItem:[OWSTableItem softCenterLabelItemWithText:@"Internal Build"]];
```
charlesmchen: If you're ever confused about whether or not you have an internal build installed, you can now check the app settings view.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2834,30 +2834,46 @@ - (void)yapDatabaseModified:(NSNotification *)notification
...
                 break;
         }
+        if (hasMalformedRowChange) {
+            break;
+        }
+    }
+
+    if (hasMalformedRowChange) {
+        // These errors seems to be very rare; they can only be reproduced
+        // using the more extreme actions in the debug UI.
```
charlesmchen: If you insert 100k messages into a conversation, YapDatabase's "group change" starts yielding bad data.  I'm not sure if this'll make us any more robust in production, but it can't hurt.

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -124,6 +124,14 @@ - (nullable OWSTableSection *)sectionForThread:(nullable TSThread *)thread
                         actionBlock:^{
                             [DebugUIMessages sendFakeMessages:10 * 1000 thread:thread];
                         }],
+        [OWSTableItem itemWithTitle:@"Create 100k fake messages"
+                        actionBlock:^{
+                            [DebugUIMessages sendFakeMessages:100 * 1000 thread:thread];
+                        }],
+        [OWSTableItem itemWithTitle:@"Create 100k fake text messages"
+                        actionBlock:^{
+                            [DebugUIMessages sendFakeMessages:100 * 1000 thread:thread isTextOnly:YES];
```
charlesmchen: These are actually fast if you comment out any logging triggered by saving the messages.  I haven't included all of those changes in this PR since we probably want that logging in production.

SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -141,6 +141,10 @@ - (void)runAsyncRegistrationsWithCompletion:(void (^_Nonnull)(void))completion
 
 + (void)protectFiles
 {
+    DDLogInfo(@"%@ Database file size: %@", self.logTag, [OWSFileSystem fileSizeOfPath:self.legacyDatabaseFilePath]);
+    DDLogInfo(@"%@ \t SHM file size: %@", self.logTag, [OWSFileSystem fileSizeOfPath:self.legacyDatabaseFilePath_SHM]);
+    DDLogInfo(@"%@ \t WAL file size: %@", self.logTag, [OWSFileSystem fileSizeOfPath:self.legacyDatabaseFilePath_WAL]);
```
charlesmchen: This can shed some interesting light.

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -994,8 +1038,8 @@ + (void)sendFakeMessages:(NSUInteger)counter
                     [[TSOutgoingMessage alloc] initWithTimestamp:[NSDate ows_millisecondTimeStamp]
                                                         inThread:thread
                                                      messageBody:randomText];
-                DDLogError(@"%@ sendFakeMessages outgoing timestamp: %llu.", self.logTag, message.timestamp);
                 [message saveWithTransaction:transaction];
+                [message updateWithMessageState:TSOutgoingMessageStateUnsent transaction:transaction];
```
michaelkirk: Not super concerned, but curious: why mark them as unsent?

charlesmchen: Because otherwise `OWSFailedMessagesJob` bogs down the app upon next launch.  The point of these tools isn't to stress test `OWSFailedMessagesJob`.

michaelkirk: ah, that makes sense. I would not have guess that this was a perf optimization. Might be worth a code comment.

PR #2965, opened 2018-01-29
Charlesmchen/async db registrations

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.h
+++ b/Signal/src/AppDelegate.h
@@ -1,15 +1,12 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
-#import <UIKit/UIKit.h>
-
 #import "HomeViewController.h"
+#import <UIKit/UIKit.h>
 
 extern NSString *const AppDelegateStoryboardMain;
 
 @interface AppDelegate : UIResponder <UIApplicationDelegate>
-
-@property (atomic) BOOL isEnvironmentSetup;
 
 @end
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -29,6 +29,7 @@
 #import <SignalMessaging/Release.h>
 #import <SignalMessaging/SignalMessaging.h>
 #import <SignalMessaging/VersionMigrations.h>
+#import <SignalServiceKit/AppReadiness.h>
 #import <SignalServiceKit/NSUserDefaults+OWS.h>
 #import <SignalServiceKit/OWSBatchMessageProcessor.h>
 #import <SignalServiceKit/OWSDisappearingMessagesJob.h>
@@ -134,7 +135,6 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
     [self startupLogging];
 
-
     // If a backup restore is in progress, try to complete it.
     // Otherwise, cleanup backup state.
     [OWSBackup applicationDidFinishLaunching];
@@ -196,7 +196,6 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
     return YES;
 }
-
 
 /**
  *  The user must unlock the device once after reboot before the database encryption key can be accessed.
@@ -394,7 +393,6 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
 - (void)applicationDidBecomeActive:(UIApplication *)application {
     DDLogWarn(@"%@ applicationDidBecomeActive.", self.logTag);
-
     if (CurrentAppContext().isRunningTests) {
         return;
     }
@@ -402,6 +400,19 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     [self removeScreenProtection];
 
     [self ensureRootViewController];
+
+    [AppReadiness.sharedManager runNowOrWhenAppIsReady:^{
+        [self handleActivation];
+    }];
+
+    DDLogInfo(@"%@ applicationDidBecomeActive completed.", self.logTag);
+}
+
+- (void)handleActivation
+{
+    OWSAssertIsOnMainThread();
+
+    DDLogWarn(@"%@ handleActivation.", self.logTag);
 
     // Always check prekeys after app launches, and sometimes check on app activation.
     [TSPreKeyManager checkPreKeysIfNecessary];
@@ -472,10 +483,9 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
                                                     preferences:[Environment preferences]];
             }
         });
-        
     }
 
-    DDLogInfo(@"%@ applicationDidBecomeActive completed.", self.logTag);
+    DDLogInfo(@"%@ handleActivation completed.", self.logTag);
 }
 
 - (void)applicationWillResignActive:(UIApplication *)application {
@@ -753,9 +763,12 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
 - (void)storageIsReady
 {
+    OWSAssertIsOnMainThread();
     DDLogInfo(@"%@ storageIsReady", self.logTag);
 
     [OWSPreferences setIsRegistered:[TSAccountManager isRegistered]];
+
+    [AppReadiness.sharedManager setAppIsReady];
 
     if ([TSAccountManager isRegistered]) {
         DDLogInfo(@"localNumber: %@", [TSAccountManager localNumber]);
@@ -833,6 +846,8 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
 - (void)ensureRootViewController
 {
+    OWSAssertIsOnMainThread();
+
     DDLogInfo(@"%@ ensureRootViewController", self.logTag);
 
     if (![OWSStorage isStorageReady] || self.hasInitialRootViewController) {
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -225,15 +225,17 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
         return;
     }
 
-    DDLogInfo(@"%@ Database file size: %@",
-        self.logTag,
-        [OWSFileSystem fileSizeOfPath:TSStorageManager.legacyDatabaseFilePath]);
-    DDLogInfo(@"%@ \t SHM file size: %@",
-        self.logTag,
-        [OWSFileSystem fileSizeOfPath:TSStorageManager.legacyDatabaseFilePath_SHM]);
-    DDLogInfo(@"%@ \t WAL file size: %@",
-        self.logTag,
-        [OWSFileSystem fileSizeOfPath:TSStorageManager.legacyDatabaseFilePath_WAL]);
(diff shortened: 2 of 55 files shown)


Signal/src/AppDelegate.h
```diff
@@ -1,15 +1,12 @@
...
-#import <UIKit/UIKit.h>
-
 #import "HomeViewController.h"
+#import <UIKit/UIKit.h>
 
 extern NSString *const AppDelegateStoryboardMain;
 
 @interface AppDelegate : UIResponder <UIApplicationDelegate>
 
-@property (atomic) BOOL isEnvironmentSetup;
```
charlesmchen: This flag has bee superceded by `AppReadiness`.

Signal/src/AppDelegate.m
```diff
@@ -367,6 +382,14 @@ - (BOOL)application:(UIApplication *)application
     sourceApplication:(NSString *)sourceApplication
            annotation:(id)annotation
 {
+    OWSAssertIsOnMainThread();
+
+    if (!AppReadiness.isAppReady) {
+        DDLogWarn(@"%@ Ignoring openURL: app not ready.", self.logTag);
+        // TODO: Consider using [AppReadiness runNowOrWhenAppIsReady:].
```
charlesmchen: There's a card to follow up on these TODOs.

Signal/test/util/OWSDatabaseConverterTest.m
```diff
@@ -373,6 +373,71 @@ - (void)testDatabaseConversion_WithKeyspec
...
+        databaseSalt = saltData;
+    };
+    __block NSData *_Nullable databaseKeySpec = nil;
+    YapDatabaseKeySpecBlock keySpecBlock = ^(NSData *keySpecData) {
+        OWSAssert(!databaseKeySpec);
+        OWSAssert(keySpecData);
+
+        databaseKeySpec = keySpecData;
+    };
+    NSError *_Nullable error = [YapDatabaseCryptoUtils convertDatabaseIfNecessary:databaseFilePath
```
charlesmchen: We could assert around how long this conversion takes.

SignalMessaging/environment/AppSetup.m
```diff
@@ -45,9 +45,7 @@ + (void)setupEnvironment:(CallMessageHandlerBlock)callMessageHandlerBlock
         [NSKeyedUnarchiver setClass:[OWSUserProfile class] forClassName:[OWSUserProfile collection]];
         [NSKeyedUnarchiver setClass:[OWSDatabaseMigration class] forClassName:[OWSDatabaseMigration collection]];
 
-        [OWSStorage setupWithSafeBlockingMigrations:^{
-            [VersionMigrations runSafeBlockingMigrations];
```
charlesmchen: Everything is now async, including the database migrations.

SignalMessaging/environment/migrations/OWS104CreateRecipientIdentities.m
```diff
@@ -24,42 +24,36 @@ + (NSString *)migrationId
...
-                                               isFirstKnownKey:NO
-                                                     createdAt:[NSDate dateWithTimeIntervalSince1970:0]
-                                             verificationState:OWSVerificationStateDefault]
-                saveWithTransaction:transaction];
-        }];
-
-        [self saveWithTransaction:transaction];
+    NSMutableDictionary<NSString *, NSData *> *identityKeys = [NSMutableDictionary new];
+
+    [transaction
```
charlesmchen: Maybe I should add asserts in `OWSRecipientIdentity` that these entities are only fetched after the app is ready.

SignalMessaging/environment/migrations/OWSDatabaseMigration.m
```diff
@@ -58,6 +58,42 @@ - (void)runUp
...
+// debug builds.
++ (YapDatabaseConnection *)dbReadWriteConnection
+{
+    static dispatch_once_t onceToken;
+    static YapDatabaseConnection *sharedDBConnection;
+    dispatch_once(&onceToken, ^{
+        sharedDBConnection = [TSStorageManager sharedManager].newDatabaseConnection;
+
+        OWSAssert([sharedDBConnection isKindOfClass:[OWSDatabaseConnection class]]);
+        ((OWSDatabaseConnection *)sharedDBConnection).canWriteBeforeStorageReady = YES;
```
charlesmchen: There are now asserts that prevent writes before the storage is ready.  We need to make an exception here though.

SignalServiceKit/src/Messages/OWSFailedAttachmentDownloadsJob.m
```diff
@@ -117,24 +117,19 @@ + (YapDatabaseSecondaryIndex *)indexDatabaseExtension
...
 }
+#endif
 
 + (void)asyncRegisterDatabaseExtensionsWithStorageManager:(OWSStorage *)storage
 {
     [storage asyncRegisterExtension:[self indexDatabaseExtension]
-                           withName:OWSFailedAttachmentDownloadsJobAttachmentStateIndex
-                    completionBlock:^(BOOL ready) {
-                        if (ready) {
-                            DDLogDebug(@"%@ completed registering extension async.", self.logTag);
```
charlesmchen: I've pulled this kind of logging into `OWSStorage`.

SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -29,12 +29,6 @@ void runSyncRegistrationsForStorage(OWSStorage *storage)
 
     // Synchronously register extensions which are essential for views.
     [TSDatabaseView registerCrossProcessNotifier:storage];
```
charlesmchen: We still register the cross process notifier synchronously.

SignalServiceKit/src/Util/AppReadiness.h
```diff
@@ -0,0 +1,28 @@
...
+
+// This method should only be called on the main thread.
++ (void)setAppIsReady;
+
+// If the app is ready, the block is called immediately;
+// otherwise it is called when the app becomes ready.
+//
+// This method should only be called on the main thread.
+// The block will always be called on the main thread.
++ (void)runNowOrWhenAppIsReady:(AppReadyBlock)block;
```
michaelkirk: 👏 👏 👏

SignalServiceKit/src/Messages/OWSFailedAttachmentDownloadsJob.m
michaelkirk: 👍

SignalMessaging/environment/migrations/OWS104CreateRecipientIdentities.m
charlesmchen: > - I don't think the app should be considered "ready" until all the migrations have run.

Great point.  I've formalized this with two non-trivial changes (corresponding to two commits):

* be1fde905 Don't mark app as ready until all version migrations are done.
* a30170b3b Prefer "app is ready" flag to "storage is ready" flag.

SignalMessaging/environment/migrations/OWSDatabaseMigration.m
charlesmchen: I think the concerns you're raising are already addressed (per the two commits I just pushed), but let's go through this in detail...

* We kick off any version/database migrations after the database registrations.  The migrations' r/w transactions will block on those registrations.  So I don't think we need a separate concept of `StorageExtensionsAreRegistered` since its implicit and guaranteed by the ordering here.
* We now don't mark the app as ready until storage is ready AND the migrations are completed (or skipped).
* We only block read/writes on "storage is ready" (i.e. all db registrations are complete but not all migrations).  We could block read/writes on "app is ready" and make an exception for the db migrations, but I'm not sure it's necessary now that we're being more careful to honor the "app is ready" flag.

> At least one migration is run sync, 

`OWS103EnableVideoCalling` looks like a sync migration, but the core work is a network request which is async.  `OWS104CreateRecipientIdentities` was sync but I've converted it to be async.  

Very happy to continue discussing this.

Signal/src/AppDelegate.m
```diff
@@ -221,6 +220,16 @@ - (void)ensureIsReadyForAppExtensions
         return;
     }
 
+    DDLogInfo(@"%@ Database file size: %@",
```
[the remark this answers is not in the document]
charlesmchen: I've added a check for the legacy database's existence.  I think it's valuable to log the database's size _before_ the conversion, if we're about to do one.

Signal/src/AppDelegate.m
```diff
@@ -783,7 +856,6 @@ - (void)storageIsReady
 
     [[OWSProfileManager sharedManager] ensureLocalProfileCached];
```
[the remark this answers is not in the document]
charlesmchen: Interesting idea.  Made a card.

Signal/src/AppDelegate.m
```diff
@@ -833,6 +905,8 @@ - (void)registrationStateDidChange
 
 - (void)ensureRootViewController
 {
+    OWSAssertIsOnMainThread();
+
     DDLogInfo(@"%@ ensureRootViewController", self.logTag);
 
     if (![OWSStorage isStorageReady] || self.hasInitialRootViewController) {
```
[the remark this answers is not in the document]
charlesmchen: Done, and I've also converted many other "storage ready" checks to use "app ready" instead.

Signal/src/network/PushManager.m
```diff
@@ -108,11 +109,15 @@ - (void)application:(UIApplication *)application didReceiveRemoteNotification:(N
 {
     DDLogInfo(@"%@ received remote notification", self.logTag);
 
-    [self.messageFetcherJob run];
+    [AppReadiness runNowOrWhenAppIsReady:^{
```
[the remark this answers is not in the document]
charlesmchen: Hmm, let's discuss this in person.  Feels like a tough call.  It might be just a subjective matter of taste.

Some of the benefits to the current arrangement:

* Gather concerns, e.g. push-related logic here in this push manager.
* DRY, i.e. if we pull this logic into AppDelegate we also have to pull it into ShareViewController.

Going to make a card.

SignalMessaging/contacts/SystemContactsFetcher.swift
```diff
@@ -62,16 +62,18 @@ class ContactsFrameworkContactStoreAdaptee: ContactStoreAdaptee {
 
     @objc
     func didBecomeActive() {
-        let currentSortOrder = CNContactsUserDefaults.shared().sortOrder
+        AppReadiness.runNowOr {
```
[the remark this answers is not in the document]
charlesmchen: Done.

SignalServiceKit/src/Storage/TSDatabaseView.m
```diff
@@ -72,19 +80,13 @@ + (void)registerMessageDatabaseViewWithName:(NSString *)viewName
...
-        [storage asyncRegisterExtension:view
-                               withName:viewName
-                        completionBlock:^(BOOL ready) {
-                            OWSCAssert(ready);
-
-                            DDLogInfo(@"%@ asyncRegisterExtension: %@ -> %d", self.logTag, viewName, ready);
-                        }];
+        [storage asyncRegisterExtension:view withName:viewName];
     } else {
         [storage registerExtension:view withName:viewName];
```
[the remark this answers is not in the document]
charlesmchen: Done.

PR #2966, opened 2018-01-29
Add support for images as documents.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -233,6 +233,7 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
 @property (nonatomic) BOOL shouldObserveDBModifications;
 @property (nonatomic) BOOL viewHasEverAppeared;
 @property (nonatomic) BOOL hasUnreadMessages;
+@property (nonatomic) BOOL isPickingMediaAsDocument;
 
 @end
 
@@ -2300,9 +2301,21 @@ typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
     // UIDocumentPickerModeImport copies to a temp file within our container.
     // It uses more memory than "open" but lets us avoid working with security scoped URLs.
     UIDocumentPickerMode pickerMode = UIDocumentPickerModeImport;
+    // TODO: UIDocumentMenuViewController is deprecated; we should use UIDocumentPickerViewController
+    //       instead.
     UIDocumentMenuViewController *menuController =
         [[UIDocumentMenuViewController alloc] initWithDocumentTypes:documentTypes inMode:pickerMode];
     menuController.delegate = self;
+
+    UIImage *takeMediaImage = [UIImage imageNamed:@"actionsheet_camera_black"];
+    OWSAssert(takeMediaImage);
+    [menuController addOptionWithTitle:NSLocalizedString(
+                                           @"MEDIA_FROM_LIBRARY_BUTTON", @"media picker option to choose from library")
+                                 image:takeMediaImage
+                                 order:UIDocumentMenuOrderFirst
+                               handler:^{
+                                   [self chooseFromLibraryAsDocument];
+                               }];
 
     [self dismissKeyBoard];
     [self presentViewController:menuController animated:YES completion:nil];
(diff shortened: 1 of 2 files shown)


SignalMessaging/attachments/SignalAttachment.swift
```diff
@@ -682,6 +682,12 @@ public class SignalAttachment: NSObject {
     private class func compressImageAsJPEG(image: UIImage, attachment: SignalAttachment, filename: String?, imageQuality: TSImageQuality) -> SignalAttachment {
         assert(attachment.error == nil)
 
+        if imageQuality == .original &&
+            attachment.dataLength < kMaxFileSizeGeneric {
```
charlesmchen: This will affect images picked through both "attach > document > photo library" and "attach > document > browse."

Note that I used the "generic" limit for "original/document images", not the "image" limit.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2601,6 +2630,10 @@ - (void)imagePickerController:(UIImagePickerController *)picker
             return failedToPickAttachment(nil);
         }
 
+        // Images chosen from the "attach document" UI should be sent as originals;
+        // images chosen from the "attach media" UI should be resized to "medium" size;
+        TSImageQuality imageQuality = (self.isPickingMediaAsDocument ? TSImageQualityOriginal : TSImageQualityMedium);
```
charlesmchen: I've made this change for images - I wonder if we should do something similar for videos?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2300,10 +2301,22 @@ - (void)showAttachmentDocumentPickerMenu
...
 
+    UIImage *takeMediaImage = [UIImage imageNamed:@"actionsheet_camera_black"];
+    OWSAssert(takeMediaImage);
+    [menuController addOptionWithTitle:NSLocalizedString(
+                                           @"MEDIA_FROM_LIBRARY_BUTTON", @"media picker option to choose from library")
+                                 image:takeMediaImage
+                                 order:UIDocumentMenuOrderFirst
+                               handler:^{
+                                   [self chooseFromLibraryAsDocument];
+                               }];
```
charlesmchen: I didn't add separate "options" for Dropbox, iCloud as a) this appears non-trivial b) they are available via the "browse..." option.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2472,10 +2485,26 @@ - (void)takePictureOrVideo
...
+}
+
+- (void)chooseFromLibraryAsMedia
+{
+    OWSAssertIsOnMainThread();
+
+    [self chooseFromLibrary:NO];
+}
+
+- (void)chooseFromLibrary:(BOOL)shouldTreatAsDocument
```
michaelkirk: style nit: I almost always prefer to follow apple's convention and include the parameter name in the method selector, otherwise it's not clear at the call site what the boolean means.

e.g.
```
chooseFromLibraryWithShouldTreatAsDocument:YES
```
or
```
chooseFromLibraryAsDocument:YES
```

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2472,10 +2485,26 @@ - (void)takePictureOrVideo
...
 {
     OWSAssertIsOnMainThread();
 
+    self.isPickingMediaAsDocument = shouldTreatAsDocument;
```
michaelkirk: Just checking - can we use the method param rather than storing and accessing this property? 

It would be nice to get rid of unnecessary state tracking, and it seems like it's only used in this method.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen: Unfortunately, due to the delegate pattern we need to hang this state somewhere.  I considered hanging it somewhere else, but couldn't find a better place.

michaelkirk: Ah you're right of course, I misread the diff.

PR #2972, opened 2018-01-31
Charlesmchen/experience upgrades vs i phone x

charlesmchen:
PTAL @michaelkirk 

// Non X (e.g. SE) should be unchanged

simulator screen shot - iphone se - 2018-01-31 at 12 58 41

// X should now be reasonable.

simulator screen shot - iphone x - 2018-01-31 at 12 57 54

### DIFF
--- a/Signal/src/ViewControllers/ExperienceUpgradesPageViewController.swift
+++ b/Signal/src/ViewControllers/ExperienceUpgradesPageViewController.swift
@@ -1,5 +1,5 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 import Foundation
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ExperienceUpgradesPageViewController.swift
```diff
@@ -10,7 +10,7 @@ private class IntroductingReadReceiptsExperienceUpgradeViewController: Experienc
     var buttonAction: ((UIButton) -> Void)?
 
     override func loadView() {
-        self.view = UIView()
+        self.view = UIView.container()
```
charlesmchen: I try to use "containers" for these types of views used to host other views, so that we don't need to worry about layout margins.  In practice it won't make any difference here since top/bottom layout margins are zero by default.

Signal/src/ViewControllers/ExperienceUpgradesPageViewController.swift
```diff
@@ -451,21 +455,34 @@ class ExperienceUpgradesPageViewController: OWSViewController, UIPageViewControl
 
         //// Layout Views
 
+        // Status Bar Background layout
+        //
+        // We use a separate backgrounds for the "status bar" area (which has very
+        // different height on iPhone X) and the "header" (which has consistent
+        // height).
```
charlesmchen: The meat.

PR #2975, opened 2018-02-02
Fix attachment approval layout on iPhoneX

michaelkirk:
convert captioning toolbar to autolayout in the process

Respecting the margins of the root view is key to iPhoneX layout, since they vary be device.

## Recipient Picker

**before:**

master x share recipient list

**after:**

proposed x share recipient list

## Approval View

**before**
master x approval view portrait

before x approval landscape keyboard

**after**

screen shot 2018-02-01 at 4 28 31 pm
screen shot 2018-02-01 at 4 28 41 pm
screen shot 2018-02-01 at 4 28 44 pm
screen shot 2018-02-01 at 4 29 05 pm

PTAL @charlesmchen

### DIFF
--- a/SignalMessaging/attachments/AttachmentApprovalViewController.swift
+++ b/SignalMessaging/attachments/AttachmentApprovalViewController.swift
@@ -282,10 +282,6 @@ public class AttachmentApprovalViewController: OWSViewController, CaptioningTool
         self.approveAttachment(captionText: captionText)
     }
 
-    func captioningToolbar(_ captioningToolbar: CaptioningToolbar, didChangeTextViewHeight newHeight: CGFloat) {
-        Logger.info("Changed height: \(newHeight)")
-    }
-
     // MARK: Video
 
     private func playVideo() {
@@ -546,7 +542,6 @@ private class GradientView: UIView {
 
 protocol CaptioningToolbarDelegate: class {
     func captioningToolbarDidTapSend(_ captioningToolbar: CaptioningToolbar, captionText: String?)
-    func captioningToolbar(_ captioningToolbar: CaptioningToolbar, didChangeTextViewHeight newHeight: CGFloat)
     func captioningToolbarDidBeginEditing(_ captioningToolbar: CaptioningToolbar)
     func captioningToolbarDidEndEditing(_ captioningToolbar: CaptioningToolbar)
 }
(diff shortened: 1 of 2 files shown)


SignalMessaging/attachments/AttachmentApprovalViewController.swift
```diff
@@ -282,10 +282,6 @@ public class AttachmentApprovalViewController: OWSViewController, CaptioningTool
         self.approveAttachment(captionText: captionText)
     }
 
-    func captioningToolbar(_ captioningToolbar: CaptioningToolbar, didChangeTextViewHeight newHeight: CGFloat) {
-        Logger.info("Changed height: \(newHeight)")
```
michaelkirk: we weren't using this for anything interesting.

SignalMessaging/attachments/AttachmentApprovalViewController.swift
```diff
@@ -624,58 +628,41 @@ class CaptioningToolbar: UIView, UITextViewDelegate {
...
-        sendButton.sizeToFit()
-    }
-
-    func didTapSend() {
-        self.captioningToolbarDelegate?.captioningToolbarDidTapSend(self, captionText: self.textView.text)
-    }
-
-    // MARK: - UIView Overrides
-
-    // We do progammatic layout, explicitly computing and setting frames since autoLayout does
```
michaelkirk: Turns our you don't need to do programmatic layout if you properly set up auto resizing. I think the autolayout code is a lot easier to reason about.

SignalMessaging/contacts/SelectThreadViewController.m
```diff
@@ -116,8 +116,7 @@ - (void)createViews
     [self.view addSubview:self.tableViewController.view];
     [_tableViewController.view autoPinWidthToSuperview];
     [_tableViewController.view autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:header];
-
-    [self autoPinViewToBottomGuideOrKeyboard:self.tableViewController.view];
+    [_tableViewController.view autoPinEdgeToSuperviewEdge:ALEdgeBottom];
```
michaelkirk: I attempted to follow the behavior of the native iOS contacts app WRT table contents flowing below the iPhoneX button.

SignalMessaging/attachments/AttachmentApprovalViewController.swift
charlesmchen: Awesome.

PR #2976, opened 2018-02-02
Fix receiving calls

michaelkirk:
PTAL @charlesmchen 

Fixes https://github.com/signalapp/Signal-iOS/issues/2958


michaelkirk:
I audited the other places we're using OWSBackgroundTask, and it looks like we're doing the right thing everywhere else.

PR #2977, opened 2018-02-03
Fix CallKit speakerphone issue and media playback after a failed call.

michaelkirk:
PTAL @charlesmchen 

Fixes https://github.com/signalapp/Signal-iOS/issues/2813


michaelkirk:
Squashed into c646f763

PR #2978, opened 2018-02-03
Fixes voiceover accessability after voice memo

michaelkirk:
PTAL @charlesmchen


michaelkirk:
I'm going to make a couple more changes.

PR #2981, opened 2018-02-06
Resume background audio and voiceover after voice note or playing media

michaelkirk:
Based on #2977 so merge that first.

PTAL @charlesmchen


Signal/src/ViewControllers/CallViewController.swift
```diff
@@ -961,6 +962,29 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver {
...
+        // Which sources are available depends on the state of your Session.
+        // When the audio session is not yet in PlayAndRecord none are available
+        // Then if we're in speakerphone, bluetooth isn't available.
+        // So we accrue all possible audio sources in a set, and that list lives as longs as the CallViewController
+        // The downside of this is that if you e.g. unpair your bluetooth mid call, it will still appear as an option
+        // until your next call.
+        // FIXME: There's got to be a better way, but this is where I landed after a bit of work, and seems to work
+        // pretty well in practice.
+        let availableInputs = callAudioService.availableInputs
+        self.allAudioSources.formUnion(availableInputs)
```
charlesmchen: Seems like we'll need to test this on iOS 8, 9, 10, 11.  Please make sure to specify that in the test plan.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3279,20 +3276,21 @@ - (void)endRecordingVoiceMemo
...
 {
     OWSAssertIsOnMainThread();
 
     DDLogDebug(@"cancelRecordingVoiceMemo");
 
-    [self resetRecordingVoiceMemo];
-}
-
-- (void)resetRecordingVoiceMemo
-{
```
charlesmchen: Huh, isn't this `OWSAssertIsOnMainThread` twice in a row?

Signal/src/ViewControllers/CallViewController.swift
michaelkirk: Sure thing.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk: Removed the redundant assert - thanks!

michaelkirk:
Squashed into c646f763

PR #2985, opened 2018-02-09
Rename forked pods to Signal app.

charlesmchen:
PTAL @michaelkirk


SignalServiceKit/src/Util/OWSAsserts.h
```diff
@@ -119,6 +119,9 @@ NS_ASSUME_NONNULL_BEGIN
 #endif
 
 // This macro is intended for use in Objective-C.
+//
+// Note that we use the "C" flavor of the assert to avoid
+// capturing a strong reference to self.
```
charlesmchen: Tiny unrelated change.

michaelkirk:
This is a dupe of the more comprehensive #2984 by @jlund - can we build off of that instead?

PR #2991, opened 2018-02-12
Charlesmchen/sae TODOs

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -554,12 +554,6 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
                 // and continue cleaning in the background.
                 [[OWSDisappearingMessagesJob sharedJob] startIfNecessary];
 
-                // TODO remove this once we're sure our app boot process is coherent.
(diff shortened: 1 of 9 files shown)


Signal/src/AppDelegate.m
```diff
@@ -478,12 +478,6 @@ - (void)handleActivation
                 // and continue cleaning in the background.
                 [[OWSDisappearingMessagesJob sharedJob] startIfNecessary];
 
-                // TODO remove this once we're sure our app boot process is coherent.
-                // Currently this happens *before* db registration is complete when
-                // launching the app directly, but *after* db registration is complete when
-                // the app is launched in the background, e.g. from a voip notification.
-                [[OWSProfileManager sharedManager] ensureLocalProfileCached];
```
charlesmchen: This seems like a good example of the kind of logic we want to do _after_ storage is ready but _before_ app is ready as part of a new cache-warming phase of app readiness that should probably live in AppSetup as previously discussed.

michaelkirk: Agreed.

PR #2993, opened 2018-02-12
Let users submit debug logs if app launch fails.

charlesmchen:
PTAL @michaelkirk 

### Step 1

simulator screen shot - iphone x - 2018-02-12 at 11 39 46

### Step 2

simulator screen shot - iphone x - 2018-02-12 at 11 39 56

### Step 3

simulator screen shot - iphone x - 2018-02-12 at 11 40 36

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -61,6 +61,7 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 @property (nonatomic) UIWindow *screenProtectionWindow;
 @property (nonatomic) BOOL hasInitialRootViewController;
 @property (nonatomic) BOOL areVersionMigrationsComplete;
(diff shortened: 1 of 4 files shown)


Signal/src/AppDelegate.m
```diff
@@ -368,6 +419,11 @@ - (void)application:(UIApplication *)application didRegisterForRemoteNotificatio
 {
     OWSAssertIsOnMainThread();
 
+    if (self.didAppLaunchFail) {
+        OWSFail(@"%@ %s", self.logTag, __PRETTY_FUNCTION__);
```
michaelkirk: Can we include a substring like "app launch failed" in all of these log/failure strings?

charlesmchen: Done.

PR #2994, opened 2018-02-12
Defer handling app delegate hooks until app is ready.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -476,7 +476,9 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
     if (!AppReadiness.isAppReady) {
         DDLogWarn(@"%@ Ignoring openURL: app not ready.", self.logTag);
-        // TODO: Consider using [AppReadiness runNowOrWhenAppIsReady:].
+        // We don't need to use [AppReadiness runNowOrWhenAppIsReady:];
+        // the only URLs we handle in Signal iOS at the moment are used
+        // for resuming the verification step of the registration flow.
         return NO;
     }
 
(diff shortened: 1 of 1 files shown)


Signal/src/AppDelegate.m
```diff
@@ -564,33 +566,31 @@ - (void)application:(UIApplication *)application
...
-        UIViewController *fromViewController = [[UIApplication sharedApplication] frontmostViewController];
-        [fromViewController presentViewController:controller
-                                         animated:YES
-                                       completion:^{
-                                           completionHandler(NO);
-                                       }];
-    }
+    }];
+
+    completionHandler(YES);
```
michaelkirk: It seems like now we could call the completionHandler twice - always here, and sometimes again after presenting the alertController. Why this change in control flow?

Signal/src/AppDelegate.m
```diff
@@ -630,38 +624,42 @@ - (BOOL)application:(UIApplication *)application
...
-        }
 
-        OutboundCallInitiator *outboundCallInitiator = SignalApp.sharedApp.outboundCallInitiator;
-        OWSAssert(outboundCallInitiator);
-        return [outboundCallInitiator initiateCallWithHandle:phoneNumber];
+            OutboundCallInitiator *outboundCallInitiator = SignalApp.sharedApp.outboundCallInitiator;
+            OWSAssert(outboundCallInitiator);
+            [outboundCallInitiator initiateCallWithHandle:phoneNumber];
+        }];
+        return YES;
```
michaelkirk: I'm unfamiliar with the ramifications of always returning YES here.

Signal/src/AppDelegate.m
charlesmchen: Good catch, this belongs inside the block above.

Signal/src/AppDelegate.m
charlesmchen: I've added a comment.

PR #2996, opened 2018-02-13
Fix notification % escaping, debug UI

michaelkirk:
Also, consolidated the notifications debug UI (and fixed it) to make testing
this a bit easier.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/DebugUI/DebugUINotifications.swift
+++ b/Signal/src/ViewControllers/DebugUI/DebugUINotifications.swift
@@ -36,11 +36,15 @@ class DebugUINotifications: DebugUIPage {
         }
 
         var sectionItems = [
-            OWSTableItem(title:"Last Incoming Message") {
-                Logger.info("\(self.logTag) scheduling notification for incoming message.")
-                self.delayedNotificationDispatch {
-                    Logger.info("\(self.logTag) dispatching")
-                    TSStorageManager.shared().newDatabaseConnection().read({ (transaction) in
+            OWSTableItem(title:"Last Incoming Message") { [weak self] in
+                guard let strongSelf = self else {
+                    return
+                }
+
+                Logger.info("\(strongSelf.logTag) scheduling notification for incoming message.")
+                strongSelf.delayedNotificationDispatch {
+                    Logger.info("\(strongSelf.logTag) dispatching")
+                    TSStorageManager.shared().newDatabaseConnection().read { (transaction) in
                         guard let viewTransaction = transaction.ext(TSMessageDatabaseViewExtensionName) as? YapDatabaseViewTransaction  else {
                             owsFail("unable to build view transaction")
                             return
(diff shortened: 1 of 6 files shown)


Signal/src/ViewControllers/DebugUI/DebugUINotifications.swift
```diff
@@ -0,0 +1,117 @@
...
+
+    override func section(thread aThread: TSThread?) -> OWSTableSection? {
+        guard let thread = aThread else {
+            owsFail("\(logTag) Notifications must specify thread.")
+            return nil
+        }
+
+        var sectionItems = [
+            OWSTableItem(title:"Last Incoming Message") {
+                Logger.info("\(self.logTag) scheduling notification for incoming message.")
```
charlesmchen: These `self` references are going to create retain cycles that will leak this UI.  I don't care about the debug UI, but we don't want to use this idiom in production code.  I wonder if there's a convenient way to do something safe in production code?

SignalMessaging/utils/DisplayableText.swift
```diff
@@ -215,6 +215,17 @@ extension String {
         return text.ows_stripped()
     }
 
+    @objc
+    public class func filterNotificationText(_ text: String?) -> String? {
+        guard let text = self.filterText(text) else {
+            return nil
+        }
+
+        // Notifications strip anything that looks lik a printf formatting character,
```
charlesmchen: s/lik/like/

charlesmchen: s/printf formatting character/string formatting specifier/

Signal/src/environment/NotificationsManager.m
```diff
@@ -312,9 +312,12 @@ - (void)notifyUserForIncomingMessage:(TSIncomingMessage *)message
                             [NSString stringWithFormat:NSLocalizedString(@"APN_MESSAGE_IN_GROUP_DETAILED", nil),
                                       senderName,
                                       threadName,
-                                      messageDescription];
+                                      [DisplayableText filterNotificationText:messageDescription]];
```
charlesmchen: I don't understand why this is desirable.  The message description isn't the format string.  So formatting specifiers in it shouldn't be escaped.  If they are (as you have it here), won't they be rendered incorrectly?  E.g. won't "I'm with you 100%" be rendered as "I'm with you 100%%"?

Signal/src/environment/NotificationsManager.m
```diff
@@ -312,9 +312,12 @@ - (void)notifyUserForIncomingMessage:(TSIncomingMessage *)message
...
+
                     } else {
-                        notification.alertBody = [NSString stringWithFormat:@"%@: %@", senderName, messageDescription];
+                        notification.alertBody = [NSString stringWithFormat:@"%@: %@",
+                                                           senderName,
+                                                           [DisplayableText filterNotificationText:messageDescription]];
```
charlesmchen: Ditto.

Signal/src/environment/NotificationsManager.m
michaelkirk: Nope, that's the point of this change.

from https://developer.apple.com/documentation/uikit/uilocalnotification/1616646-alertbody

> Printf style escape characters are stripped from the string prior to display; to include a percent symbol (%) in the message, use two percent symbols (%%).

Signal/src/ViewControllers/DebugUI/DebugUINotifications.swift
michaelkirk: Thanks, fixed!

Signal/src/environment/NotificationsManager.m
michaelkirk: I've included some more details, and a link to the docs in the comment in DisplayableText

SignalMessaging/utils/DisplayableText.swift
michaelkirk: typo fixed - thanks. 

To be consistent with the iOS docs, I'm going to continue to refer to the affected chars as "printf formatting" chars.

michaelkirk:
PTAL @charlesmchen

Signal/src/environment/NotificationsManager.m
charlesmchen: Oh. Can you please add a comment here (and maybe below too) with that link & quote?

michaelkirk: Done!

PR #2997, opened 2018-02-13
crash on search

michaelkirk:
Small crash fix, and some cleanup around it.

PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Messages/TSGroupModel.m
+++ b/SignalServiceKit/src/Messages/TSGroupModel.m
@@ -23,6 +23,22 @@ NS_ASSUME_NONNULL_BEGIN
     return self;
 }
 
+- (nullable instancetype)initWithCoder:(NSCoder *)coder
+{
+    self = [super initWithCoder:coder];
+    if (!self) {
+        return self;
+    }
+
+    // Occasionally seeing this as nil in legacy data,
+    // which causes crashes.
+    if (_groupMemberIds == nil) {
+        _groupMemberIds = [NSArray new];
+    }
+
+    return self;
+}
+
 - (BOOL)isEqual:(id)other {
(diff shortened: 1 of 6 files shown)


SignalServiceKit/src/Messages/TSGroupModel.m
```diff
@@ -21,6 +23,22 @@ - (instancetype)initWithTitle:(NSString *)title
...
+- (nullable instancetype)initWithCoder:(NSCoder *)coder
+{
+    self = [super initWithCoder:coder];
+    if (!self) {
+        return self;
+    }
+
+    // Occasionally seeing this as nil in legacy data,
+    // which causes crashes.
+    if (_groupMemberIds == nil) {
```
michaelkirk: The proximate cause of the crash is that, while searching we enumerate all member ids of all groups, and assume groupMemberIds is non-nil.

This change let's us maintain that invariant rather than making the property nullable and adding nil-checking in multiple places.

SignalServiceKit/src/Messages/TSGroupModel.m
```diff
@@ -1,16 +1,18 @@
...
 
 #if TARGET_OS_IOS
-- (instancetype)initWithTitle:(NSString *)title
+- (instancetype)initWithTitle:(nullable NSString *)title
                     memberIds:(NSArray<NSString *> *)memberIds
-                        image:(UIImage *)image
+                        image:(nullable UIImage *)image
                       groupId:(NSData *)groupId
 {
     _groupName              = title;
```
charlesmchen: Maybe `OWSAssert(memberIds);`?

michaelkirk:
Thanks - in b999cd9e6

PR #2998, opened 2018-02-13
Fix dynamic type check which was too restrictive

michaelkirk:
PTAL @charlesmchen


SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -1328,6 +1328,8 @@ - (NSDictionary *)encryptedMessageWithPlaintext:(NSData *)plainText
                 }
                 dispatch_semaphore_signal(sema);
             }];
+        // FIXME: Currently this happens within a readwrite transaction - meaning our read-write transaction blocks
+        // on a network request.
```
michaelkirk: I made a card for this.

michaelkirk:
In b358a75e3

PR #3000, opened 2018-02-13
Aggressively nag iOS 8 users to upgrade iOS.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/utils/OWSPreferences.h
+++ b/SignalMessaging/utils/OWSPreferences.h
@@ -1,5 +1,5 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 NS_ASSUME_NONNULL_BEGIN
@@ -58,8 +58,8 @@ extern NSString *const OWSPreferencesKeyEnableDebugLog;
 - (BOOL)hasDeclinedNoContactsView;
 - (void)setHasDeclinedNoContactsView:(BOOL)value;
 
-- (void)setIOSUpgradeNagVersion:(NSString *)value;
-- (nullable NSString *)iOSUpgradeNagVersion;
+- (void)setIOSUpgradeNagDate:(NSDate *)value;
+- (nullable NSDate *)iOSUpgradeNagDate;
 
 #pragma mark - Calling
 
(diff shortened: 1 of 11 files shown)


charlesmchen:
PTAL @michaelkirk

SignalMessaging/views/OWSAlerts.swift
```diff
@@ -65,4 +65,32 @@ import Foundation
...
+    @objc
+    public class func showIOSUpgradeNagIfNecessary() {
+        // Only show the nag to iOS 8 users.
+        if #available(iOS 9.0, *) {
+            return
+        }
+
+        // Don't show the nag to users who have just launched
+        // the app for the first time.
+        guard let _ = AppVersion.instance().lastAppVersion else {
```
michaelkirk: style nit: 

Rather than using both implicit-unwrap and ignored-return, I think it would be clearer to just compare to nil

```
guard AppVersion.instance().lastAppVersion != nil else {
```

SignalMessaging/utils/OWSPreferences.h
```diff
@@ -58,8 +58,8 @@ extern NSString *const OWSPreferencesKeyEnableDebugLog;
 - (BOOL)hasDeclinedNoContactsView;
 - (void)setHasDeclinedNoContactsView:(BOOL)value;
 
-- (void)setIOSUpgradeNagVersion:(NSString *)value;
-- (nullable NSString *)iOSUpgradeNagVersion;
```
michaelkirk: I think we want to retain the existing nag behavior for people on iOS9+, right?  We don't want people running super-old versions of Signal regardless of their platform.

Though we may consider disabling the "old nag" for the few people running iOS8, since they may not be able to update to the "latest" version.

charlesmchen: I think you're thinking of the _signal update_ nag implemented in `AppUpdateNag`.  That remains in place.  I'm just removing the state we used to use to throttle the _iOS update_ nag.

michaelkirk: You're correct - sorry for the confusion.

LGTM!

SignalMessaging/views/OWSAlerts.swift
charlesmchen: Done.

PR #3001, opened 2018-02-13
Charlesmchen/message sender deadlocks

charlesmchen:
The deadlocks occur because the network manager completions were getting enqueued on the main thread and deadlocking if the main thread was blocked trying to open a r/w transaction.

PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSMessageSender.m
+++ b/SignalServiceKit/src/Messages/OWSMessageSender.m
@@ -1310,7 +1310,7 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
         TSRequest *request =
             [[TSRecipientPrekeyRequest alloc] initWithRecipient:identifier deviceId:[deviceNumber stringValue]];
         [self.networkManager makeRequest:request
-            shouldCompleteOnMainQueue:NO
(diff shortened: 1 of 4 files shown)


SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -1303,8 +1303,14 @@ - (NSDictionary *)encryptedMessageWithPlaintext:(NSData *)plainText
...
-        [self.networkManager makeRequest:[[TSRecipientPrekeyRequest alloc] initWithRecipient:identifier
-                                                                                    deviceId:[deviceNumber stringValue]]
+        // It's not ideal that we're using a semaphore inside a read/write transaction.
+        // To avoid deadlock, we need to ensure that our success/failure completions
+        // are called _off_ the main thread.  Otherwise we'll deadlock if the main
+        // thread is blocked on opening a transaction.
+        TSRequest *request =
+            [[TSRecipientPrekeyRequest alloc] initWithRecipient:identifier deviceId:[deviceNumber stringValue]];
+        [self.networkManager makeRequest:request
+            shouldCompleteOnMainQueue:NO
```
michaelkirk: It feels a bit more idiomatic to pass the completion queue rather than a boolean, but this works for now.

PR #3002, opened 2018-02-13
Constant time compare

### LINKED ISSUE #29: Is anything preventing the compiler from optimizing isEqualToData_TimingSafe into a short-circuiting loop?
I wrote the method [`isEqualToData_TimingSafe`](https://github.com/WhisperSystems/Signal-iOS/blob/master/Signal/src/crypto/CryptoTools.m#L66) to use non-short-circuiting logic, but the language semantics might allow the compiler to optimize it into a loop that does short-circuit, (The compiler would have to reason across the message passing, but it's possible.)

Can this method be delegated to a crypto library?

CollinStuart:
- fix case when second part of the && conditional is skipped when data is not equal
- isEqual variable marked volatile to prevent case when it doesn't equal 0, the loop can break early since it can never be 0 again
- tested with Fastest O3 and Whole Module optimization (App Store Release)

// FREEBIE

### Contributor checklist

 * iPhone 5C, iOS 10.3.3
 * iPhone 6S, iOS 11.1

- - - - - - - - - -

### Description
This pull request reduces the possibility for compiler optimizations/loop short-circuiting. Testing was done with the Fastest O3 release configuration. 

FWIW there is also a const-time comparison function in the Curve25519Kit library under [compare.c](https://github.com/signalapp/Signal-Pods/blob/a8334500da7c8180341043d84b0b41145e6619e1/Curve25519Kit/Sources/ed25519/additions/compare.c) that also works on unsigned char *.

fixes #29

### DIFF
--- a/SignalServiceKit/src/Util/NSData+OWSConstantTimeCompare.m
+++ b/SignalServiceKit/src/Util/NSData+OWSConstantTimeCompare.m
@@ -10,7 +10,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (BOOL)ows_constantTimeIsEqualToData:(NSData *)other
 {
-    BOOL isEqual = YES;
+    volatile UInt8 isEqual = 0;
 
     if (self.length != other.length) {
         return NO;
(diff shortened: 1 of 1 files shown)


michaelkirk:
In d5e61dac9, thanks!

PR #3006, opened 2018-02-14
Charlesmchen/batch processing glitch

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
+++ b/SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
@@ -269,6 +269,10 @@ NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSMessageContentJo
                                              selector:@selector(appIsReady)
                                                  name:AppIsReadyNotification
                                                object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(yapDatabaseModified:)
+                                                 name:YapDatabaseModifiedNotification
+                                               object:TSStorageManager.sharedManager.dbNotificationObject];
 
     return self;
 }
@@ -280,6 +284,13 @@ NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSMessageContentJo
 
 - (void)appIsReady
 {
+    [self drainQueue];
+}
+
+- (void)yapDatabaseModified:(NSNotification *)notification
+{
+    OWSAssertIsOnMainThread();
+
     [self drainQueue];
 }
 
@@ -316,6 +327,13 @@ NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSMessageContentJo
     dispatch_async(self.serialQueue, ^{
         if (!AppReadiness.isAppReady) {
             // We don't want to process incoming messages until storage is ready.
+            static dispatch_once_t onceToken;
+            dispatch_once(&onceToken, ^{
+                [AppReadiness runNowOrWhenAppIsReady:^{
+                    [self drainQueue];
+                }];
+            });
+
             return;
         }
 
@@ -471,7 +489,8 @@ NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSMessageContentJo
 
     // We need to persist the decrypted envelope data ASAP to prevent data loss.
     [self.processingQueue enqueueEnvelopeData:envelopeData plaintextData:plaintextData transaction:transaction];
-    [self.processingQueue drainQueue];
+    // The new envelope won't be visible to the finder until this transaction commits,
+    // so don't bother calling drainQueue here.
 }
 
 @end
--- a/SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
+++ b/SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
@@ -265,33 +265,12 @@ NSString *const OWSMessageContentJobFinderExtensionGroup = @"OWSMessageContentJo
     _finder = finder;
     _isDrainingQueue = NO;
 
-    [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(appIsReady)
-                                                 name:AppIsReadyNotification
-                                               object:nil];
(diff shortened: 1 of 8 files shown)


SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
```diff
@@ -283,6 +287,13 @@ - (void)appIsReady
     [self drainQueue];
 }
 
+- (void)yapDatabaseModified:(NSNotification *)notification
+{
+    OWSAssertIsOnMainThread();
+
+    [self drainQueue];
```
charlesmchen: The core fix - we used to kick off queue processing in `enqueueEnvelopeData`, but this won't work any more (see the comment below).  So now we observe changes to the database.

SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -139,6 +139,13 @@ - (void)yapDatabaseModified:(NSNotification *)notification
 {
     if (AppReadiness.isAppReady) {
         [OWSMessageUtils.sharedManager updateApplicationBadgeCount];
+    } else {
+        static dispatch_once_t onceToken;
```
charlesmchen: This doesn't fix any known issues, but I realized that anywhere we abort because `isAppReady` isn't true yet, we should enqueue a block that will try again when the app is ready.  Slightly laborious, but that will make this code easier to reason about.  I use a once token to prevent enqueuing the block more than once.

SignalServiceKit/src/Util/AppReadiness.m
```diff
@@ -94,10 +94,12 @@ - (void)runAppReadyBlocks
     OWSAssertIsOnMainThread();
     OWSAssert(self.isAppReady);
 
-    for (AppReadyBlock block in self.appReadyBlocks) {
+    // Make a local copy, then clear this state.
+    NSArray<AppReadyBlock> *appReadyBlocks = self.appReadyBlocks;
+    self.appReadyBlocks = nil;
```
charlesmchen: A third fix.  Again, this does not address a known issue, but it's safer.

This makes us robust to enqueuing a new "app ready block" while executing the existing blocks.

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
```diff
@@ -316,6 +327,13 @@ - (void)drainQueue
     dispatch_async(self.serialQueue, ^{
         if (!AppReadiness.isAppReady) {
             // We don't want to process incoming messages until storage is ready.
+            static dispatch_once_t onceToken;
+            dispatch_once(&onceToken, ^{
```
michaelkirk: I'll just leave this comment only once, but I feel the same about all the places in this PR where this code has been copy pasted.

Is this required for fixing the bug discussed in the PR, or is this just being extra careful?

Seeing this code duplicated in so many classes makes me feel like we have the wrong model. 

Rather than having all these classes know about "app readiness" and react to "app readiness", it seems like an "app readiness coordinator" should be telling these classes when to act. 

IMO AppDelegate is a good candidate for this "app readiness state coordinator" role (and is mostly already doing it):

e.g. 
- The AppDelegate doesn't start jobs until App is ready
- The AppDelegate doesn't act upon notifications until the app is ready
- etc.

The way I'm thinking about it - this block of code should never execute if we've initiated our dependencies coherently. In other words - if we are trying to `drainQueue` before the app is ready, that seems like a bug in whoever told the queue to start draining.

What I'd ultimately like to see is something like`OWSAssert(AppReadiness.isAppReady)`, and make any requisite changes to make that a reality, but in the interest of avoiding any big changes, would you be willing to add an `OWSFail` in all of these blocks as written?

Or are you think about this differently?

SignalServiceKit/src/Util/AppReadiness.m
```diff
@@ -94,10 +94,12 @@ - (void)runAppReadyBlocks
     OWSAssertIsOnMainThread();
     OWSAssert(self.isAppReady);
 
-    for (AppReadyBlock block in self.appReadyBlocks) {
+    // Make a local copy, then clear this state.
```
michaelkirk: Is this pertinent to the fix? It's all main thread only right?

A comment as to "why" might be more helpful than "what"

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
```diff
@@ -471,7 +489,8 @@ - (void)enqueueEnvelopeData:(NSData *)envelopeData
 
     // We need to persist the decrypted envelope data ASAP to prevent data loss.
     [self.processingQueue enqueueEnvelopeData:envelopeData plaintextData:plaintextData transaction:transaction];
```
michaelkirk: Ah - good find! 

I think that your fix works, but dbModified happens all the time, even though this isn't a particularly expensive method, I think we should try to avoid falling back on observing dbModified when there's another straight forward way. How about instead using `transaction`'s completion queue method:

```
- (void)addCompletionQueue:(dispatch_queue_t)completionQueue
           completionBlock:(dispatch_block_t)completionBlock
```

SignalServiceKit/src/Util/AppReadiness.m
charlesmchen: Having thought it over, I think I'm going to revert this change.

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
charlesmchen: As discussed offline, I hope we can move "app readiness coordination" into AppSetup in v2.21.

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
charlesmchen: Nice, done.

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
charlesmchen: I've made a number of changes in this direction in this PR.  Let's pursue in v2.21.

PR #3009, opened 2018-02-15
Fix "lose messages received while in background"

michaelkirk:
A migrated DB file has the NSFileProtectionClassComplete, meaning it's never accessible while the device is locked. We want it to be accessible after first unlock.

Applying a data protection class to a directory only affects newly created files, not existing files in that directory nor files moved to that directory.

Presumably this has been an issue for a while in the 2.20 branch, but you'd only notice when receiving voip notifications.

PTAL: @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Util/OWSFileSystem.m
+++ b/SignalServiceKit/src/Util/OWSFileSystem.m
@@ -9,8 +9,34 @@ NS_ASSUME_NONNULL_BEGIN
 
 @implementation OWSFileSystem
 
++ (BOOL)protectRecursiveContentsAtPath:(NSString *)path
+{
+    BOOL isDirectory;
+    if (![NSFileManager.defaultManager fileExistsAtPath:path isDirectory:&isDirectory]) {
+        return NO;
+    }
+
+    if (!isDirectory) {
+        [self protectFileOrFolderAtPath:path];
+    }
+    NSString *dirPath = path;
+
+    BOOL success = YES;
+    NSDirectoryEnumerator *directoryEnumerator = [[NSFileManager defaultManager] enumeratorAtPath:dirPath];
+
+    for (NSString *relativePath in directoryEnumerator) {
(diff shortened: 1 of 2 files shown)


SignalServiceKit/src/Util/OWSFileSystem.m
```diff
@@ -9,8 +9,34 @@
...
++ (BOOL)protectRecursiveContentsAtPath:(NSString *)path
+{
+    BOOL isDirectory;
+    if (![NSFileManager.defaultManager fileExistsAtPath:path isDirectory:&isDirectory]) {
+        return NO;
+    }
+
+    if (!isDirectory) {
+        [self protectFileOrFolderAtPath:path];
+    }
```
charlesmchen: Shouldn't we early `return YES;` here?

charlesmchen: Oh, I think you meant to `return  [self protectFileOrFolderAtPath:path];`.

SignalServiceKit/src/Util/OWSFileSystem.m
```diff
@@ -32,6 +58,33 @@ + (BOOL)protectFileOrFolderAtPath:(NSString *)path
...
 
++ (void)logAttributesOfItemAtPath:(NSString *)path
+{
+    NSDictionary<NSFileAttributeKey, id> *_Nullable attributes = [self attributesOfItemAtPath:path];
+    DDLogDebug(@"%@ path: %@ has attributes: %@", self.logTag, path, attributes);
+}
+
++ (nullable NSDictionary<NSFileAttributeKey, id> *)attributesOfItemAtPath:(NSString *)path
+{
+    return [[NSFileManager defaultManager] attributesOfItemAtPath:path error:nil];
```
charlesmchen: Should we use an `NSError` and verify it?

michaelkirk: Done!

PR #3014, opened 2018-02-15
Improve handling of attachments with captions.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/Interactions/TSMessage.m
+++ b/SignalServiceKit/src/Messages/Interactions/TSMessage.m
@@ -3,6 +3,7 @@
 //
 
 #import "TSMessage.h"
+#import "AppContext.h"
 #import "NSDate+OWS.h"
 #import "NSString+SSK.h"
 #import "TSAttachment.h"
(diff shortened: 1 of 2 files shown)


SignalServiceKit/src/Messages/Interactions/TSMessage.m
```diff
@@ -245,14 +252,18 @@ - (NSString *)previewTextWithTransaction:(YapDatabaseReadTransaction *)transacti
             return NSLocalizedString(@"UNKNOWN_ATTACHMENT_LABEL", @"In Inbox view, last message label for thread with corrupted attachment.");
         }
     } else {
-        return self.body;
+        // TODO: We should do better here.
```
michaelkirk: Outside of a buggy sending client, does this happen?

In other words, should this be an OWSFail or at least a DDLogError

SignalServiceKit/src/Messages/Interactions/TSMessage.m
```diff
@@ -236,7 +240,10 @@ - (NSString *)debugDescription
 
 - (NSString *)previewTextWithTransaction:(YapDatabaseReadTransaction *)transaction
 {
-    if ([self hasAttachments]) {
+    if (self.body.length > 0) {
```
michaelkirk: I'm not sure this is better, since now we exclude the fact that it's an attachment from the notification. 

What about including "\(attachmentDescription): \(caption)" when both are present, something like:

"📽 Attachment: Cha Cha Cha"

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -1153,8 +1153,9 @@ - (void)handleSendToMyself:(TSOutgoingMessage *)outgoingMessage
     [self handleMessageSentLocally:outgoingMessage];
 
     if (!(outgoingMessage.body || outgoingMessage.hasAttachments)) {
-        DDLogDebug(
-            @"%@ Refusing to make incoming copy of non-standard message sent to self:%@", self.logTag, outgoingMessage);
+        DDLogDebug(@"%@ Refusing to make incoming copy of non-standard message sent to self: %@",
```
michaelkirk: This looks like just a formatting change, but I don't remember when or why this would happen. If you know could you leave a comment?

SignalServiceKit/src/Messages/Interactions/TSMessage.m
charlesmchen: Done.

SignalServiceKit/src/Messages/OWSMessageSender.m

SignalServiceKit/src/Messages/Interactions/TSMessage.m
charlesmchen: Reworked this.

charlesmchen:
PTAL @michaelkirk

SignalServiceKit/src/Messages/Interactions/TSMessage.m
```diff
@@ -226,33 +227,64 @@ - (BOOL)hasAttachments
...
         NSString *attachmentId = self.attachmentIds[0];
         return [NSString stringWithFormat:@"Media Message with attachmentId:%@", attachmentId];
     } else {
         return [NSString stringWithFormat:@"%@ with body: %@", [self class], self.body];
     }
 }
 
+// TODO: This method contains view-specific logic and probably belongs in NotificationsManager, not in SSK.
 - (NSString *)previewTextWithTransaction:(YapDatabaseReadTransaction *)transaction
 {
```
michaelkirk: These changes LGTM, but do you want to duplicate them into the `description` method too?

charlesmchen: Done.

PR #3015, opened 2018-02-15
Surface error messages in SAE as alerts.

charlesmchen:
PTAL @michaelkirk 

This is needed to surface "safety number changes" encountered during message sending to the user.


charlesmchen:
PTAL @michaelkirk

PR #3017, opened 2018-02-15
Charlesmchen/sae filenames

charlesmchen:
The "coerce to NSData" hack won't work; it prevents us from preserving the file names of shared files.  It's probably a tad safer, but I don't know of a specific share type that fixes at the moment.  I think it's far better to preserve filenames and fix broken share types (if any) when we find them.  They should be rare, if they even exist.

PTAL @michaelkirk


SignalShareExtension/ShareViewController.swift
```diff
@@ -713,7 +710,7 @@ public class ShareViewController: UIViewController, ShareViewDelegate, SAEFailed
                 } else {
                     return itemUrl
                 }
-            }()
+                }()
```
michaelkirk: weird indent

SignalShareExtension/ShareViewController.swift
```diff
@@ -91,7 +91,7 @@ public class ShareViewController: UIViewController, ShareViewDelegate, SAEFailed
 
             Logger.debug("\(strongSelf.logTag) setup is slow - showing loading screen")
             strongSelf.showPrimaryViewController(loadViewController)
-        }.retainUntilComplete()
+            }.retainUntilComplete()
```
michaelkirk: weird indent

SignalShareExtension/ShareViewController.swift
```diff
@@ -494,19 +494,19 @@ public class ShareViewController: UIViewController, ShareViewDelegate, SAEFailed
             strongSelf.loadViewController = nil
             strongSelf.showPrimaryViewController(conversationPicker)
             Logger.info("showing picker with attachment: \(attachment)")
-        }.catch {[weak self]  error in
-            AssertIsOnMainThread()
-            guard let strongSelf = self else { return }
+            }.catch {[weak self]  error in
```
michaelkirk: weird indent

charlesmchen:
Does this also fix the ability to share urls from safari/chrome?

@michaelkirk Yes

SignalShareExtension/ShareViewController.swift
charlesmchen: Fixed.

SignalShareExtension/ShareViewController.swift
charlesmchen: Fixed.

SignalShareExtension/ShareViewController.swift

PR #3018, opened 2018-02-15
Charlesmchen/sae filenames 

charlesmchen:
PTAL @michaelkirk

This reverts most of https://github.com/signalapp/Signal-iOS/pull/3017

I realized that we can get the best of both worlds by only using the "coerce item load to data" if none of the item provider's UTI types conform to URL.  Most shares will have at least one UTI type that does, and if it does we should be able to load the share as a URL instead of as data which is desirable.

### DIFF
--- a/SignalShareExtension/ShareViewController.swift
+++ b/SignalShareExtension/ShareViewController.swift
@@ -90,7 +90,7 @@ public class ShareViewController: UIViewController, ShareViewDelegate, SAEFailed
 
             Logger.debug("\(strongSelf.logTag) setup is slow - showing loading screen")
             strongSelf.showPrimaryViewController(loadViewController)
-            }.retainUntilComplete()
+        }.retainUntilComplete()
 
         // We shouldn't set up our environment until after we've consulted isReadyForAppExtensions.
         AppSetup.setupEnvironment({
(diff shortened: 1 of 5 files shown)


SignalShareExtension/ShareViewController.swift
```diff
@@ -690,12 +695,28 @@ public class ShareViewController: UIViewController, ShareViewDelegate, SAEFailed
...
+            // as a URL:
+            //
+            // * So that we can preserve the source filename which we infer from the URL.
+            // * So that we can avoid loading the item into memory if possible.
+
+            itemProvider.loadItem(forTypeIdentifier: srcUtiType, options: nil, completionHandler: loadCompletion)
+        } else {
+            // Otherwise, fall back to using the "load as data" coercion hack since this is
+            // safer.
+            itemProvider.loadData(forTypeIdentifier: srcUtiType, options: nil, completionHandler: loadCompletion)
```
michaelkirk: Do you have a test case/file for this? I'm unclear when we execute this code.

charlesmchen:
@michaelkirk I just found an edge case where "coercing to load as data" doesn't work, so I'm going to close this PR.

PR #3021, opened 2018-02-15
Improve handling of the not migrated case.

charlesmchen:
PTAL @michaelkirk


SignalShareExtension/ShareViewController.swift
```diff
@@ -66,14 +66,7 @@ public class ShareViewController: UIViewController, ShareViewDelegate, SAEFailed
         // abort.
         isReadyForAppExtensions = OWSPreferences.isReadyForAppExtensions()
         guard isReadyForAppExtensions else {
-            // If we don't have TSSStorageManager, we can't consult TSAccountManager
-            // for isRegistered, so we use OWSPreferences which is usually-accurate
-            // copy of that state.
-            if OWSPreferences.isRegistered() {
```
charlesmchen: We can't use `OWSPreferences.isRegistered` since it introduced too recently to be of any use.

PR #3024, opened 2018-02-16
Fix issues around profile updates.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/profiles/OWSUserProfile.m
+++ b/SignalMessaging/profiles/OWSUserProfile.m
@@ -36,6 +36,8 @@ NSString *const kLocalProfileUniqueId = @"kLocalProfileUniqueId";
 
 @implementation OWSUserProfile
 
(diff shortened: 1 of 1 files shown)


SignalMessaging/profiles/OWSUserProfile.m
```diff
@@ -122,13 +153,6 @@ - (void)applyChanges:(void (^)(id))changeBlock
                     @"%@ Ignoring redundant update in %s: %@", self.logTag, functionName, self.debugDescription);
                 didChange = NO;
             } else {
-                NSString *_Nullable oldAvatarUrlPath = beforeSnapshot[@"avatarUrlPath"];
```
charlesmchen: We don't want to enforce this here - it's "too late" in the update process and prevents us from updating the avatar URL and filename in a single update, which we do for _local_ profile changes.  Better for this logic to reside in the avatar URL setter and use a consistent ordering for property updates in the updateWith:... methods.

PR #3027, opened 2018-02-16
Only update file protection for db files.

michaelkirk:
There are potentially many files in Attachments/Profiles, and since movement happens during launch, it must be fast.

PTAL @charlesmchen


michaelkirk:
Superseded by #3031

PR #3031, opened 2018-02-18
remove slow file protection updates from launch path

michaelkirk:
To avoid blocking launch, file protection is now updated async for most
moved files. Out of paranoia, the database files are also update
redundantly on a sync code path.

It's still critical that we update permissions recursively for two
reasons:

1. Updating a containing directory's FileProtection does not affect
   existing files in that directory. Only future files made in that directory.

2. Because we've changed the containers default file protection level
   (from unspecified to NSFileProtectionComplete), some existing files
   will have their file protection updated upon launching Signal 2.20.
   It's not clear to me which files this affects, and I haven't found
   any relevant documentation, but from observation, it seems to affect
   any top-level files in the container. Regardless, we're now doing the
   right thing: after launching 2.20, ensure all file permissions are
   what we expect.

Also removed no-op file protection on legacy db files. They've already
been moved by the time this method runs in AppSetup.

And I added a cheesy-but-helpful file browser to the debug ui.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/DebugUI/DebugUIFileBrowser.swift
+++ b/Signal/src/ViewControllers/DebugUI/DebugUIFileBrowser.swift
@@ -1,0 +1,380 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+class DebugUIFileBrowser: OWSTableViewController {
+
+    // MARK: Dependencies
+    var fileManager: FileManager {
+        return FileManager.default
+    }
+
+//    let logTag = "[DebugUIFileBrowser]"
+
+    // MARK: Overrides
+    let fileURL: URL
+
+    init(fileURL: URL) {
+        self.fileURL = fileURL
+
+        super.init(nibName: nil, bundle: nil)
+
+        self.contents = buildContents()
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        let titleLabel = UILabel()
+        titleLabel.text = "\(fileURL)"
+        titleLabel.sizeToFit()
+        titleLabel.textColor = UIColor.white
+        titleLabel.lineBreakMode = .byTruncatingHead
+        self.navigationItem.titleView = titleLabel
+    }
+
+    fileprivate func updateContents() {
+        self.contents = buildContents()
+        self.tableView.reloadData()
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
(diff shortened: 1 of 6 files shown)


SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -155,12 +155,6 @@ + (void)protectFiles
     DDLogInfo(@"%@ \t SHM file size: %@", self.logTag, [OWSFileSystem fileSizeOfPath:self.sharedDataDatabaseFilePath_SHM]);
     DDLogInfo(@"%@ \t WAL file size: %@", self.logTag, [OWSFileSystem fileSizeOfPath:self.sharedDataDatabaseFilePath_WAL]);
 
-    // The old database location was in the Document directory,
```
michaelkirk: We've already moved the legacy DB at the point this gets called in AppSetup

Signal/src/ViewControllers/DebugUI/DebugUIFileBrowser.swift
```diff
@@ -0,0 +1,380 @@
...
+//
+
+class DebugUIFileBrowser: OWSTableViewController {
+
+    // MARK: Dependencies
+    var fileManager: FileManager {
+        return FileManager.default
+    }
+
+//    let logTag = "[DebugUIFileBrowser]"
```
charlesmchen: ?

Signal/src/ViewControllers/DebugUI/DebugUIFileBrowser.swift
```diff
@@ -0,0 +1,380 @@
...
+
+    fileprivate func updateContents() {
+        self.contents = buildContents()
+        self.tableView.reloadData()
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
+        // In case files were added / removed in child view controller
+        updateContents()
+        super.viewWillAppear(animated)
```
charlesmchen: I usually like to do the super first in these UIKit lifecycle methods unless there's a specific reason to do otherwise.  We always want to call super and by placing it first its easier to nice if its missing.

Signal/src/ViewControllers/DebugUI/DebugUIFileBrowser.swift
michaelkirk: Fixed!

Signal/src/ViewControllers/DebugUI/DebugUIFileBrowser.swift
michaelkirk: Fixed!

michaelkirk:
In fbbf432a4886c4a0c63172472fa886ae107d8910

PR #3037, opened 2018-02-20
Improve the robustness of the migration logic.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -230,34 +230,63 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
 - (BOOL)ensureIsReadyForAppExtensions
 {
+    // Given how sensitive this migration is, we verbosely
+    // log the contents of all involved paths before and after.
+    //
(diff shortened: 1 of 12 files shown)


Signal/src/AppDelegate.m
```diff
@@ -230,35 +230,64 @@ - (void)verifyDBKeysAvailableBeforeBackgroundLaunch
 
 - (BOOL)ensureIsReadyForAppExtensions
 {
+    // Given how sensitive this migration is, we verbosely
+    // log the contents of all involved paths before and after.
+    //
+    // TODO: Remove this logging once we have high confidence
+    // in our migration logic.
+    NSArray<NSString *> *paths = @[
+        TSStorageManager.legacyDatabaseFilePath,
```
michaelkirk: Fine for now, but we have to be sure that none of these class methods ever instantiate an instance of TSStorageManager, like some of our class methods do.

charlesmchen: +1

PR #3040, opened 2018-02-21
Improve the robustness of the migration logic.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Storage/TSStorageManager.m
+++ b/SignalServiceKit/src/Storage/TSStorageManager.m
@@ -18,8 +18,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-NSString *const TSStorageManagerExceptionName_CouldNotMoveDatabaseFile
-    = @"TSStorageManagerExceptionName_CouldNotMoveDatabaseFile";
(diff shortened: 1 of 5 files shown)


SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -255,21 +253,35 @@ + (nullable NSError *)migrateToSharedData
...
 
     NSError *_Nullable error = nil;
-    error = [OWSFileSystem moveAppFilePath:self.legacyDatabaseFilePath
-                        sharedDataFilePath:self.sharedDataDatabaseFilePath
-                             exceptionName:TSStorageManagerExceptionName_CouldNotMoveDatabaseFile];
+    if ([fileManager fileExistsAtPath:self.legacyDatabaseFilePath] &&
+        [fileManager fileExistsAtPath:self.sharedDataDatabaseFilePath]) {
+        // In the case that we have a "database conflict" (i.e. database files
+        // in the src and dst locations), ensure database integrity by renaming
+        // all of the dst database files.
```
charlesmchen: This is the key fix - we wouldn't want to end up with a SHM or WAL file from an "abandoned" database.  So if we detect a database conflict (as discussed today), immediately move aside all existing database files in the shared data container.

PR #3042, opened 2018-02-21
Charlesmchen/custom notification sounds

charlesmchen:
First steps toward this.  The copy is pretty rough.

* Lets users select a global "new message" notification sound.
* Affects "new message" sound in-app.
* Affects "new message" sound in local notification.
* Remove the ker-clunk asset.
* Doesn't support per-thread "new message" sounds yet.
* Doesn't affect ringtones.
* Doesn't affect the misc. sounds like connecting.

PTAL @michaelkirk 

simulator screen shot - iphone x - 2018-02-20 at 22 12 49
simulator screen shot - iphone x - 2018-02-20 at 22 12 42


kaktusztea:
Great!
In the future it would be pretty a much more discrete sound when the conversation is actually open.

PR #3043, opened 2018-02-21
Send ICE updates immediately after sending CallOffer for faster call connection

michaelkirk:
Sending ICE updates immediately will allow calls
to connect without having to wait for an additional serialized round
trip between the caller and call recipient.

For legacy reasons, the call sender used to wait until after receiving
the call answer before sending the ICE updates. The primary motivation
was that if the receiving user hadn't accepted a new identity change,
rather than just seeing one "Tap to Accept New Safety Number" messages
for a call, they'd see one for the call offer and then a dozen more as
ICE updates trickled in.

We changed that behavior long ago, and effected clients have been mostly phased out.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/PeerConnectionClient.swift
+++ b/Signal/src/call/PeerConnectionClient.swift
@@ -1,5 +1,5 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 import Foundation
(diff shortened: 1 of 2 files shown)


Signal/src/call/PeerConnectionClient.swift
```diff
@@ -657,7 +657,7 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
                 Logger.debug("\(self.TAG) \(#function) Ignoring obsolete event in terminated client")
                 return
             }
-            Logger.debug("\(self.TAG) didChange IceConnectionState:\(newState.debugDescription)")
+            Logger.info("\(self.TAG) didChange IceConnectionState:\(newState.debugDescription)")
```
michaelkirk: Turning up logging a bit since we're releasing several call related changes.

PR #3047, opened 2018-02-21
Charlesmchen/conversation view vs modified external

charlesmchen:
PTAL @michaelkirk 

Although I think we'll want to do this, it's too late to merge such a big change in v2.20.  

I haven't thoroughly tested this work yet.

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -229,7 +229,7 @@ typedef enum : NSUInteger {
 @property (nonatomic) BOOL hasUnreadMessages;
 @property (nonatomic) BOOL isPickingMediaAsDocument;
 @property (nonatomic, nullable) NSNumber *previousLastTimestamp;
-@property (nonatomic, nullable) NSNumber *previousLastUnreadTimestamp;
+@property (nonatomic, nullable) NSNumber *viewHorizonTimestamp;
 
 @end
 
@@ -4156,12 +4156,11 @@ typedef enum : NSUInteger {
             CGPoint newContentOffset = CGPointMake(0, MAX(0, newContentHeight - viewTopToContentBottom));
             [self.collectionView setContentOffset:newContentOffset animated:NO];
         }
-        
+
         // When we resume observing database changes, we want to scroll to show the user
         // any new items inserted while we were not observing.  We therefore find the
-        // first item which was previously unread and center it in the view.  We don't
-        // want to make any assumptions, so we select the item to center on based on
-        // "previous unread timestamp" rather than a specific interaction.
+        // first item at or after the "view horizon".  See the comments below which explain
+        // the "view horizon".
         ConversationViewItem *_Nullable lastViewItem = self.viewItems.lastObject;
         BOOL hasAddedNewItems = (lastViewItem &&
             previousLastTimestamp &&
@@ -4169,18 +4168,22 @@ typedef enum : NSUInteger {
 
         DDLogInfo(@"%@ hasAddedNewItems: %d", self.logTag, hasAddedNewItems);
         if (hasAddedNewItems) {
-            NSIndexPath *_Nullable indexPathToShow = [self firstIndexPathAtPreviousLastUnreadTimestamp];
+            NSIndexPath *_Nullable indexPathToShow = [self firstIndexPathAtViewHorizonTimestamp];
             if (indexPathToShow) {
                 [self.collectionView scrollToItemAtIndexPath:indexPathToShow
                                             atScrollPosition:UICollectionViewScrollPositionCenteredVertically
                                                     animated:YES];
             }
         }
-        self.previousLastUnreadTimestamp = nil;
+        self.viewHorizonTimestamp = nil;
         DDLogVerbose(@"%@ resumed observation of database modifications.", self.logTag);
     } else {
         DDLogVerbose(@"%@ pausing observation of database modifications.", self.logTag);
-        // When stopping observation, try to record the timestamp of the last item.
+        // When stopping observation, try to record the timestamp of the "view horizon".
+        // The "view horizon" is where we'll want to focus the users when we resume
+        // observation if any changes have happened while we weren't observing.
+        // Ideally, we'll focus on those changes.  But we can't skip over unread
+        // interactions, so we prioritize those, if any.
         //
         // We'll use this later to update the view to reflect any changes made while
         // we were not observing the database.  See extendRangeToIncludeUnobservedItems
@@ -4197,26 +4200,30 @@ typedef enum : NSUInteger {
             [[TSDatabaseView unseenDatabaseViewExtension:transaction] firstObjectInGroup:self.thread.uniqueId];
         }];
         if (firstUnseenInteraction) {
-            self.previousLastUnreadTimestamp = @(firstUnseenInteraction.timestamp);
+            // If there are any unread interactions, focus on the first one.
+            self.viewHorizonTimestamp = @(firstUnseenInteraction.timestamp);
+        } else if (lastViewItem) {
+            // Otherwise, focus _just after_ the last interaction.
+            self.viewHorizonTimestamp = @(lastViewItem.interaction.timestamp + 1);
         } else {
-            self.previousLastUnreadTimestamp = nil;
+            self.viewHorizonTimestamp = nil;
         }
         DDLogVerbose(@"%@ paused observation of database modifications.", self.logTag);
     }
 }
 
-- (nullable NSIndexPath *)firstIndexPathAtPreviousLastUnreadTimestamp
+- (nullable NSIndexPath *)firstIndexPathAtViewHorizonTimestamp
 {
     OWSAssert(self.shouldObserveDBModifications);
 
-    if (!self.previousLastUnreadTimestamp) {
+    if (!self.viewHorizonTimestamp) {
         return nil;
     }
     if (self.viewItems.count < 1) {
         return nil;
     }
-    uint64_t previousLastUnreadTimestamp = self.previousLastUnreadTimestamp.unsignedLongLongValue;
-    // Binary search for the first view item whose timestamp >= the "previous last unread" timestamp.
+    uint64_t viewHorizonTimestamp = self.viewHorizonTimestamp.unsignedLongLongValue;
+    // Binary search for the first view item whose timestamp >= the "view horizon" timestamp.
     // We want to move "left" rightward, discarding interactions before this cutoff.
     // We want to move "right" leftward, discarding all-but-the-first interaction after this cutoff.
     // In the end, if we converge on an item _after_ this cutoff, it's the one we want.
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -115,13 +115,6 @@
     kMediaTypeVideo,
 } kMediaTypes;
 
-typedef NS_ENUM(NSInteger, MessagesRangeSizeMode) {
-    // This mode should only be used when initially configuring the range,
-    // since we want the range to monotonically grow after that.
-    MessagesRangeSizeMode_Truncate,
-    MessagesRangeSizeMode_Normal
```
charlesmchen: This got simplified out by the `lastRangeLength` changes.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -195,7 +188,7 @@ @interface ConversationViewController () <AttachmentApprovalViewControllerDelega
 @property (nonatomic, readonly) UILabel *backButtonUnreadCountLabel;
 @property (nonatomic, readonly) NSUInteger backButtonUnreadCount;
 
-@property (nonatomic) NSUInteger page;
+@property (nonatomic) NSUInteger lastRangeLength;
```
charlesmchen: It's much simpler and more expressive to represent the size of the current "desired" range using the unit of items, not pages of items.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -235,6 +228,8 @@ @interface ConversationViewController () <AttachmentApprovalViewControllerDelega
 @property (nonatomic) BOOL viewHasEverAppeared;
 @property (nonatomic) BOOL hasUnreadMessages;
 @property (nonatomic) BOOL isPickingMediaAsDocument;
+@property (nonatomic, nullable) NSNumber *previousLastTimestamp;
+@property (nonatomic, nullable) NSNumber *previousLastUnreadTimestamp;
```
charlesmchen: We now do a much better job updating the view when we resume observing database modifications.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1564,20 +1559,19 @@ - (void)setDisappearingMessagesConfiguration:
...
 {
-    // The "old" range length may have been increased by insertions of new messages
-    // at the bottom of the window.
-    NSUInteger oldLength = [self.messageMappings numberOfItemsInGroup:self.thread.uniqueId];
-
-    NSUInteger targetLength = oldLength;
-    if (mode == MessagesRangeSizeMode_Truncate) {
+    NSUInteger rangeLength = 0;
+    
+    if (self.lastRangeLength == 0) {
```
charlesmchen: We were using the "mode" to special case the first time we calculate the range size.  Now we can just use `lastRangeLength`.  We can now execute this logic more than once (for empty conversations) but that's fine.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1586,23 +1580,21 @@ - (void)updateMessageMappingRangeOptions:(MessagesRangeSizeMode)mode
...
-    NSUInteger rangeLength;
-    while (YES) {
-        rangeLength = kYapDatabasePageSize * (self.page + 1);
-        if (rangeLength >= targetLength) {
-            break;
-        }
-        self.page = self.page + 1;
-    }
+    // Always try to load at least a single page of messages.
+    rangeLength = MAX(rangeLength, kYapDatabasePageSize);
```
charlesmchen: I feel like this method is now simpler and more readable.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2797,10 +2789,16 @@ - (void)yapDatabaseModifiedExternally:(NSNotification *)notification
 
     DDLogVerbose(@"%@ %s", self.logTag, __PRETTY_FUNCTION__);
 
-    // External database modifications can't be converted into incremental updates,
-    // so rebuild everything.  This is expensive and usually isn't necessary, but
-    // there's no alternative.
-    [self resetMappings];
+    if (self.shouldObserveDBModifications) {
```
charlesmchen: We only need to reset our view state if we observe `yapDatabaseModifiedExternally` while we are observing db changes.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4123,9 +4124,135 @@ - (void)setShouldObserveDBModifications:(BOOL)shouldObserveDBModifications
             CGPoint newContentOffset = CGPointMake(0, MAX(0, newContentHeight - viewTopToContentBottom));
             [self.collectionView setContentOffset:newContentOffset animated:NO];
         }
```
charlesmchen: We now do two separate things to update our view when we resume observing database modifications.

* We expand the mapping range to include any items added while we were not observing.
* We try to update scroll to reflect new items added while we were not observing.

SignalMessaging/Views/ThreadViewHelper.m
```diff
@@ -129,16 +129,18 @@ - (void)yapDatabaseModifiedExternally:(NSNotification *)notification
...
-    // so rebuild everything.  This is expensive and usually isn't necessary, but
-    // there's no alternative.
-    [self.uiDatabaseConnection beginLongLivedReadTransaction];
-    [self.uiDatabaseConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
-        [self.threadMappings updateWithTransaction:transaction];
-    }];
-
-    [self updateThreads];
-    [self.delegate threadListDidChange];
+    if (self.shouldObserveDBModifications) {
```
charlesmchen: Small, safe optimization: we don't need to do this if we're not observing db modifications since we'll do it when we resume.

SignalServiceKit/src/Storage/OWSStorage.m
```diff
@@ -76,7 +76,10 @@ - (void)readWriteWithBlock:(void (^)(YapDatabaseReadWriteTransaction *transactio
     OWSAssert(delegate);
     OWSAssert(delegate.areAllRegistrationsComplete || self.canWriteBeforeStorageReady);
 
-    OWSBackgroundTask *backgroundTask = [OWSBackgroundTask backgroundTaskWithLabelStr:__PRETTY_FUNCTION__];
+    OWSBackgroundTask *_Nullable backgroundTask = nil;
+    if (CurrentAppContext().isMainApp) {
```
charlesmchen: Background tasks have no effect in SAE and will only clutter the logs with useless warnings.

Signal/src/ViewControllers/HomeViewController.m
```diff
@@ -931,10 +931,12 @@ - (void)yapDatabaseModifiedExternally:(NSNotification *)notification
 
     DDLogVerbose(@"%@ %s", self.logTag, __PRETTY_FUNCTION__);
 
-    // External database modifications can't be converted into incremental updates,
-    // so rebuild everything.  This is expensive and usually isn't necessary, but
-    // there's no alternative.
-    [self resetMappings];
+    if (self.shouldObserveDBModifications) {
```
charlesmchen: Small, safe optimization: we don't need to do this if we're not observing db modifications since we'll do it when we resume.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4156,17 +4291,26 @@ - (void)resetMappings
...
     [self updateBackButtonUnreadCount];
     [self updateNavigationBarSubtitleLabel];
+
+    // There appears to be a bug in YapDatabase that sometimes delays modifications
+    // made in another process (e.g. the SAE) from showing up in other processes.
+    // There's a simple workaround: a trivial write to the database flushes changes
+    // made from other processes.
+    [self.editingDatabaseConnection setObject:[NSUUID UUID].UUIDString
+                                       forKey:@"conversation_view_noop_mod"
+                                 inCollection:@"temp"];
```
charlesmchen: This turned out to be key change.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk: 👏 👏 👏 I agree.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk: It would be good to make this async to avoid blocking the UI when other writes are happening.

But would it still solve our problem? I *think* it would because ultimately we're relying on the a new dbModified notification to fix up the state of the uiDatabase.

Can you also make a card to follow up with upstream about this bug?

TIOLI

Signal/src/ViewControllers/HomeViewController.m
michaelkirk: Makes sense -can you leave a similar comment in the code?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4142,9 +4144,149 @@ - (void)setShouldObserveDBModifications:(BOOL)shouldObserveDBModifications
...
+        BOOL hasAddedNewItems = (lastViewItem &&
+            previousLastTimestamp &&
+                                 lastViewItem.interaction.timestamp > previousLastTimestamp.unsignedLongLongValue);
+
+        DDLogInfo(@"%@ hasAddedNewItems: %d", self.logTag, hasAddedNewItems);
+        if (hasAddedNewItems) {
+            NSIndexPath *_Nullable indexPathToShow = [self firstIndexPathAtViewHorizonTimestamp];
+            if (indexPathToShow) {
+                [self.collectionView scrollToItemAtIndexPath:indexPathToShow
+                                            atScrollPosition:UICollectionViewScrollPositionCenteredVertically
```
michaelkirk: For the sake of longer messages, do we want UICollectionViewScrollPositionTop?

TIOLI

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4142,9 +4144,149 @@ - (void)setShouldObserveDBModifications:(BOOL)shouldObserveDBModifications
...
+                                                    animated:YES];
```
michaelkirk: ntit: This happens before the view is visible, in viewWillAppear, so don't we want `animated:NO`?

Signal/src/ViewControllers/HomeViewController.m
charlesmchen: Done.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen: Right, it's a tough call, because my goal is to show _both_ the last item before the "view horizon" and the first item after the "view horizon".  We can't do "top on first item after" or "bottom on last item before" or we won't see the other. Unfortunately, this gets tricky if either is huge.  The largest cells are oversize text, which should be rare.  Other cells are considerably smaller than a screenful.  So I think this is reasonable.

Left a comment.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen: Reasonable.  I was thinking it'd be better to emphasize that we scrolled to the user, but ... it's not clear what's best.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m

PR #3048, opened 2018-02-22
Charlesmchen/custom notification sounds2

charlesmchen:
PTAL @michaelkirk 

simulator screen shot - iphone x - 2018-02-21 at 23 01 01
simulator screen shot - iphone x - 2018-02-21 at 23 00 57
simulator screen shot - iphone x - 2018-02-21 at 23 00 54
simulator screen shot - iphone x - 2018-02-21 at 23 00 44
simulator screen shot - iphone x - 2018-02-21 at 23 00 31


Signal/src/Signal-Bridging-Header.h
```diff
@@ -10,9 +10,9 @@
 #import "ConversationViewItem.h"
 #import "DateUtil.h"
```
charlesmchen: I haven't tested yet, but I'm assuming the sound used for the fallback push notifications is specified in the payload built by the service.  So I've overwritten `NewMessage.aifc` with the iOS default notification sound "note".

Signal/src/ViewControllers/NotificationSettingsViewController.m
```diff
@@ -31,21 +31,34 @@ - (void)updateTableContents
...
+    // Sounds section.
+    
     OWSTableSection *soundsSection = [OWSTableSection new];
-    soundsSection.headerTitle = NSLocalizedString(
-        @"NOTIFICATIONS_SECTION_SOUNDS", @"Label for settings UI that allows user to change the notification sound.");
+    soundsSection.headerTitle = NSLocalizedString(@"SETTINGS_SECTION_SOUNDS",
+                                                  @"Label for the sounds section of settings views.");
+    [soundsSection
+     addItem:[OWSTableItem disclosureItemWithText:
+              NSLocalizedString(@"SETTINGS_ITEM_NOTIFICATION_SOUND",
```
charlesmchen: The copy is all rough.  The details of how exactly this UI is presented in the app and conversation settings should be discussed too.  I deviated from the WA design.  For one thing, I think our privacy measures prevent us from leveraging the custom ringtones on the device contacts, since the calls aren't associated with a system contact.  Additionally, they have some extra settings that we haven't committed to implementing.

Signal/src/ViewControllers/NotificationSoundsViewController.h
```diff
@@ -1,9 +0,0 @@
-//
-//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
-//
-
-#import "OWSTableViewController.h"
-
-@interface NotificationSoundsViewController : OWSTableViewController
```
charlesmchen: Renamed this and generalized it to handle notification and ringtone selection, globally and per-thread.

Signal/src/ViewControllers/OWSSoundSettingsViewController.m
```diff
@@ -0,0 +1,165 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OWSSoundSettingsViewController ()
+
+@property (nonatomic) BOOL isDirty;
+
+@property (nonatomic) OWSSound currentSound;
+
+@property (nonatomic) AVAudioPlayer *audioPlayer;
```
charlesmchen: We need to retain a reference to the `AVAudioPlayer` so that we can ensure only one "demo" is playing at once, and that we stop demoing sounds if we leave this view.

Signal/src/ViewControllers/OWSSoundSettingsViewController.m
```diff
@@ -0,0 +1,165 @@
...
+{
+    [self updateTableContents];
+}
+
+- (void)updateNavigationItems
+{
+    self.navigationItem.leftBarButtonItem =
+        [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemCancel
+                                                      target:self
+                                                      action:@selector(cancelWasPressed:)];
```
charlesmchen: Should we use "cancel" or "stop"?  We tend to use "stop" throughout our app.

Signal/src/ViewControllers/OWSSoundSettingsViewController.m
```diff
@@ -0,0 +1,165 @@
...
+
+    self.currentSound = sound;
+    self.isDirty = YES;
+    [self updateTableContents];
+    [self updateNavigationItems];
+}
+
+- (void)cancelWasPressed:(id)sender
+{
+    // TODO: Add "discard changes?" alert.
```
charlesmchen: I didn't bother with a warning when abandoning unsaved changes, but I did include a dirty flag towards this.

Signal/src/call/CallAudioService.swift
```diff
@@ -102,42 +102,6 @@ protocol CallAudioServiceDelegate: class {
         }
     }
 
-    class Sound: NSObject {
-
-        static let incomingRing = Sound(filePath: "r", fileExtension: "caf", loop: true)
-        static let outgoingRing = Sound(filePath: "outring", fileExtension: "mp3", loop: true)
-        static let dialing = Sound(filePath: "sonarping", fileExtension: "mp3", loop: true)
-        static let busy = Sound(filePath: "busy", fileExtension: "mp3", loop: false)
-        static let failure = Sound(filePath: "failure", fileExtension: "mp3", loop: false)
```
charlesmchen: Superceded by the more general `OWSSounds`/`OWSSound`.  Maybe I should rename `OWSSounds` to `OWSSoundManager`.

Signal/src/call/Speakerbox/CallKitCallUIAdaptee.swift
```diff
@@ -124,6 +124,12 @@ final class CallKitCallUIAdaptee: NSObject, CallUIAdaptee, CXProviderDelegate {
 
         disableUnsupportedFeatures(callUpdate: update)
 
+        // Update the provider configuration to reflect the caller's ringtone.
+        let sound = OWSSounds.ringtoneSound(for: call.thread)
+        let providerConfiguration = provider.configuration
+        providerConfiguration.ringtoneSound = OWSSounds.filename(for: sound)
+        provider.configuration =  providerConfiguration
```
charlesmchen: We sent `ringtoneSound` in the context of the call, to honor custom ringtones.

Signal/src/environment/NotificationsManager.m
```diff
@@ -100,8 +101,8 @@ - (void)presentMissedCall:(SignalCall *)call callerName:(NSString *)callerName
     };
 
     if ([self shouldPlaySoundForNotification]) {
-        NotificationSound notificationSound = [Environment preferences].globalNotificationSound;
-        notification.soundName = [NotificationSounds filenameForNotificationSound:notificationSound];
+        OWSSound sound = [OWSSounds notificationSoundForThread:thread];
+        notification.soundName = [OWSSounds filenameForSound:sound];
```
charlesmchen: We could DRY this up a little.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1575,6 +1572,12 @@
 /* content of tweet when inviting via twitter */
 "SETTINGS_INVITE_TWITTER_TEXT" = "You can reach me using Signal by @WhisperSystems, get it now.";
 
+/* Label for settings view that allows user to change the notification sound. */
+"SETTINGS_ITEM_NOTIFICATION_SOUND" = "Notification";
+
+/* Label for settings view that allows user to change the ringtone sound. */
+"SETTINGS_ITEM_RINGTONE_SOUND" = "Ringtone";
```
charlesmchen: To be discussed.

SignalMessaging/environment/NotificationSounds.h
```diff
@@ -1,37 +0,0 @@
-//
-//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
-//
-
-typedef NS_ENUM(NSUInteger, NotificationSound) {
-    NotificationSound_Aurora = 0,
-    NotificationSound_Bamboo,
```
charlesmchen: Renamed to `OWSSounds` and generalized to handle notification sounds, ringtones and "call sounds".

SignalMessaging/environment/OWSSounds.m
```diff
@@ -0,0 +1,463 @@
...
+    if (!self) {
+        return self;
+    }
+
+    OWSAssert(storageManager);
+
+    _dbConnection = storageManager.newDatabaseConnection;
+    // TODO: Is it safe to load all of these sounds into memory?
+    //       Probably better to do LRU cache.
+    self.systemSoundIDMap = [NSMutableDictionary new];
```
charlesmchen: Note this TODO.

SignalMessaging/environment/OWSSounds.m
```diff
@@ -0,0 +1,463 @@
...
+            self.systemSoundIDMap[@(sound)] = systemSoundID;
+        }
+        return (SystemSoundID)systemSoundID.unsignedIntegerValue;
+    }
+}
+
+- (void)playSound:(OWSSound)sound
+{
+    SystemSoundID systemSoundID = [self systemSoundIDForSound:sound];
+    AudioServicesPlayAlertSound(systemSoundID);
```
charlesmchen: We might want to use AVAudioPlayer instead...

SignalMessaging/environment/OWSSounds.m
```diff
@@ -0,0 +1,463 @@
...
+
+            // Calls
+        case OWSSound_CallConnecting:
+            return @"ringback_tone_cept.caf";
+        case OWSSound_CallOutboundRinging:
+            return @"ringback_tone_ansi.caf";
+        case OWSSound_CallBusy:
+            return @"busy_tone_ansi.caf";
+        case OWSSound_CallFailure:
+            return @"end_call_tone_cept.caf";
```
charlesmchen: Some of these replacement sounds for the "call sounds" were more obvious than others.  To be discussed.

SignalMessaging/utils/OWSPreferences.h
```diff
@@ -57,9 +55,6 @@ extern NSString *const OWSPreferencesKeyEnableDebugLog;
 - (void)setIOSUpgradeNagDate:(NSDate *)value;
 - (nullable NSDate *)iOSUpgradeNagDate;
 
-- (NotificationSound)globalNotificationSound;
-- (void)setGlobalNotificationSound:(NotificationSound)notificationSound;
```
charlesmchen: Storage of these values is a private implementation detail of `OWSSounds`.

Signal/src/Signal-Bridging-Header.h
michaelkirk: Per external conversation, I think this is an improvement but it may be possible to respect the users "default message tone" choice here (but probably will never be possible to support their per-user choice).

Signal/src/ViewControllers/OWSSoundSettingsViewController.m
michaelkirk: IMO I agree with "Cancel", though I don't feel strongly.

Signal/src/ViewControllers/OWSSoundSettingsViewController.m
```diff
@@ -0,0 +1,165 @@
...
+    [self.audioPlayer stop];
```
michaelkirk: Is this sufficient for background audio to resume when we're done? I'd like to have a mostly drop-in replacement for AVPlayer which does this for us, and OWSAudioAttachmentPlayer is basically that, but it has some additional features that we don't need, like a delegate/timer intended to be used to update playback state UI, but on the whole it's not really "attachment" specific.

So paths forward:

- continue to use AVAudioPlayer here, but add the "beginAudioActivity"/"endAudioActivity" instrumentation.
- use the OWSAttachmentPlayer here which already has that logic (but maybe some undesirable extras)
- extract a new class OWSAudioPlayer from OWSAttachmentPlayer, which does the beginAudioActivtiy/endAudioActivity stuff

SignalMessaging/environment/OWSSounds.h
```diff
@@ -0,0 +1,97 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef NS_ENUM(NSUInteger, OWSSound) {
```
michaelkirk: Is there a reason to write all this new code in objc vs swift?

SignalMessaging/environment/OWSSounds.m
michaelkirk: Do you know how this plays with background audio? e.g. I know if we're using AVAudioPlayer we'll have to explicitly end the audio session afterwards. Not sure about AudioServices*

SignalMessaging/environment/OWSSounds.m
```diff
@@ -0,0 +1,463 @@
...
+{
+    return (sound == OWSSound_CallConnecting || sound == OWSSound_CallOutboundRinging ||
+        [self.allRingtoneSounds containsObject:@(sound)]);
+}
+
++ (nullable AVAudioPlayer *)audioPlayerForSound:(OWSSound)sound
+{
+    NSURL *soundURL = [OWSSounds soundURLForSound:sound];
+    NSError *error;
+    AVAudioPlayer *_Nullable player = [[AVAudioPlayer alloc] initWithContentsOfURL:soundURL error:&error];
```
michaelkirk: Same concerns as my earlier comment starting with: 

> Is this sufficient for background audio to resume when we're done?

Within the context of call audio, which this code may have been extracted from, we were playing fast and loose, because all call audio was book-ended by the CallAdaptee's starting/ending an audio activity. 

But now that we're using it outside of the context of a call, we'll need to properly terminate the audio session when done playing. Again - either by inlining startActivity/stopActivity calls somewhere in this class, or preferably, by using some kind of reusable OWSAudioPlayer class which accounts for this.

Signal/src/ViewControllers/OWSSoundSettingsViewController.m
michaelkirk: 👍

Signal/src/Signal-Bridging-Header.h
charlesmchen: Yup. There's a checklist item in the card.

SignalMessaging/environment/OWSSounds.m
charlesmchen: I've made a item in the checklist for the card.

Signal/src/ViewControllers/OWSSoundSettingsViewController.m
charlesmchen: I like the third option.  I'll pursue in the "fourth" PR around this work, since this get touched in the next PR.

charlesmchen:
>  If it's going to be easier to address this in a subsequent PR, then LGTM

Going to do this in next PR, thanks.

SignalMessaging/environment/OWSSounds.h
charlesmchen: I wasn't 100% sure this wouldn't need to move to SSK.  It doesn't look like it needs to.

SignalMessaging/environment/OWSSounds.m
charlesmchen: +1 will do.

PR #3063, opened 2018-02-26
Share background tasks.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/environment/Release.m
+++ b/SignalMessaging/environment/Release.m
@@ -7,7 +7,6 @@
 #import "NotificationsManager.h"
 #import "OWSContactsManager.h"
 #import <SignalServiceKit/ContactsUpdater.h>
-#import <SignalServiceKit/OWSBackgroundTask.h>
 #import <SignalServiceKit/OWSMessageSender.h>
 #import <SignalServiceKit/TSNetworkManager.h>
 
@@ -19,7 +18,6 @@
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
         // Order matters here.
-        [[OWSBackgroundTaskManager sharedManager] observeNotifications];
         TSStorageManager *storageManager = [TSStorageManager sharedManager];
         TSNetworkManager *networkManager = [TSNetworkManager sharedManager];
         OWSContactsManager *contactsManager = [OWSContactsManager new];
--- a/SignalMessaging/environment/Release.m
+++ b/SignalMessaging/environment/Release.m
@@ -19,7 +19,7 @@
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
         // Order matters here.
-        [OWSBackgroundTaskManager sharedManager];
+        [[OWSBackgroundTaskManager sharedManager] observeNotifications];
         TSStorageManager *storageManager = [TSStorageManager sharedManager];
         TSNetworkManager *networkManager = [TSNetworkManager sharedManager];
         OWSContactsManager *contactsManager = [OWSContactsManager new];
--- a/SignalMessaging/environment/Release.m
+++ b/SignalMessaging/environment/Release.m
@@ -1,5 +1,5 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 #import "Release.h"
@@ -7,6 +7,7 @@
 #import "NotificationsManager.h"
 #import "OWSContactsManager.h"
 #import <SignalServiceKit/ContactsUpdater.h>
+#import <SignalServiceKit/OWSBackgroundTask.h>
 #import <SignalServiceKit/OWSMessageSender.h>
 #import <SignalServiceKit/TSNetworkManager.h>
 
@@ -18,6 +19,7 @@
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
         // Order matters here.
+        [OWSBackgroundTaskManager sharedManager];
         TSStorageManager *storageManager = [TSStorageManager sharedManager];
         TSNetworkManager *networkManager = [TSNetworkManager sharedManager];
         OWSContactsManager *contactsManager = [OWSContactsManager new];
--- a/SignalServiceKit/src/Util/OWSBackgroundTask.m
+++ b/SignalServiceKit/src/Util/OWSBackgroundTask.m
@@ -4,14 +4,290 @@
 
 #import "OWSBackgroundTask.h"
 #import "AppContext.h"
+#import "NSTimer+OWS.h"
 #import "Threading.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef void (^BackgroundTaskExpirationBlock)(void);
+typedef NSNumber *OWSTaskId;
+
+// This class can be safely accessed and used from any thread.
+@interface OWSBackgroundTaskManager ()
+
+// This property should only be accessed while synchronized on this instance.
+@property (nonatomic) UIBackgroundTaskIdentifier backgroundTaskId;
+
+// This property should only be accessed while synchronized on this instance.
+@property (nonatomic) NSMutableDictionary<OWSTaskId, BackgroundTaskExpirationBlock> *expirationMap;
+
+// This property should only be accessed while synchronized on this instance.
+@property (nonatomic) unsigned long long idCounter;
+
+// Note that this flag is set a little early in "will resign active".
+//
+// This property should only be accessed while synchronized on this instance.
+@property (nonatomic) BOOL isAppActive;
+
+// We use this timer to provide continuity and reduce churn,
+// so that if one OWSBackgroundTask ends right before another
+// begins, we use a single uninterrupted background that
+// spans their lifetimes.
+//
+// This property should only be accessed while synchronized on this instance.
+@property (nonatomic, nullable) NSTimer *continuityTimer;
+
+@end
+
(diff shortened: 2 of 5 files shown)


SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -4,14 +4,290 @@
...
+        // We can't create background tasks in the SAE, but pretend that we succeeded.
+        return YES;
+    }
+
+    @synchronized(self)
+    {
+        // We only want to have a background task if we are:
+        // a) "not active" AND
+        // b1) there is more than one active instance of OWSBackgroundTask.
+        // b2) or there _was_ an active instance recently.
```
charlesmchen: This is a key comment.

SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -4,14 +4,290 @@
...
+    // be called on the main thread, so facilitate that by dispatching async
+    // to main queue here.  That way we can ensure that we don't end the
+    // background task until all of the completion blocks have completed.
+    dispatch_async(dispatch_get_main_queue(), ^{
+        for (BackgroundTaskExpirationBlock expirationBlock in expirationMap) {
+            expirationBlock();
+        }
+        if (backgroundTaskId != UIBackgroundTaskInvalid) {
+            // Apparently we need to "end" even expired background tasks.
+            [CurrentAppContext() endBackgroundTask:backgroundTaskId];
```
charlesmchen: We weren't doing this before.

SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -4,14 +4,290 @@
...
+
+#pragma mark -
+
 @interface OWSBackgroundTask ()
 
 @property (nonatomic, readonly) NSString *label;
 
 // This property should only be accessed while synchronized on this instance.
-@property (nonatomic) UIBackgroundTaskIdentifier backgroundTaskId;
+@property (nonatomic, nullable) OWSTaskId taskId;
```
charlesmchen: There's a new implementation detail in this file - "task ids" which are a token used to identify instances of `OWSBackgroundTask` for the manager.

SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -131,40 +396,38 @@ - (void)startBackgroundTask
             self.completionBlock = nil;
         }
         if (completionBlock) {
-            completionBlock(BackgroundTaskState_CouldNotStart);
+            DispatchMainThreadSafe(^{
+                completionBlock(BackgroundTaskState_CouldNotStart);
```
charlesmchen: We now guarantee that we always call `completionBlock` on the main thread, which we should have been doing all along.  I don't think it was an issue though.

SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -76,36 +352,26 @@ - (void)dealloc
...
-        // will always be called on the main thread, but in practice we've observed
-        // otherwise.  We use DispatchSyncMainThreadSafe() (note the sync) to ensure that
-        // this work is done on the main thread.
-        //
-        // See: https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtaskwithexpiratio)
-        //
-        // Note the usage of OWSCAssert() to avoid capturing a reference to self.
-        OWSCAssert([NSThread isMainThread]);
-
-        DispatchSyncMainThreadSafe(^{
```
charlesmchen: We don't need to use `DispatchSyncMainThreadSafe()`.  See `[OWSBackgroundTaskManager 
 backgroundTaskExpired]`.

SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -4,14 +4,290 @@
...
+    }
+
+    // It'd be nice to do this work synchronously, but it seems unsafe to
+    // depend on all expiration blocks being cheap.
+    //
+    // OWSBackgroundTask's API guarantees that completionBlock will always
+    // be called on the main thread, so facilitate that by dispatching async
+    // to main queue here.  That way we can ensure that we don't end the
+    // background task until all of the completion blocks have completed.
+    dispatch_async(dispatch_get_main_queue(), ^{
```
michaelkirk: Dispatching async from the expiration handler seems like a bad idea

https://developer.apple.com/documentation/uikit/uiapplication/1623051-beginbackgroundtaskwithname?language=objc
> The handler is called synchronously on the main thread, blocking the app’s suspension momentarily while the app is notified.

Presumably the calling iOS code looks something like
```
-(void) taskDidExpire
{
    NSAssert([NSThread isMainThread]);
    callTheirExpirationHandler();
    terminateApp();
}
```

I read that as "whatever is blocking your app's suspension will run before any dispatched code." - so I'd say these blocks should be run in line.

SignalServiceKit/src/Util/OWSBackgroundTask.m
michaelkirk: Good to know!

SignalMessaging/environment/Release.m
```diff
@@ -18,6 +19,7 @@ + (Environment *)releaseEnvironment
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
         // Order matters here.
+        [OWSBackgroundTaskManager sharedManager];
```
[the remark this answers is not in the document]
charlesmchen: > Can we instead prefer to consolidate global state mutation related to "app setup" in the aptly named AppSetup class?

I'll try to take this on in a separate PR later today.

> I'd prefer code like this be made explicit

I went with:

`[[OWSBackgroundTaskManager sharedManager] observeNotifications];`

SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -4,14 +4,290 @@
...
+
+    @synchronized(self)
+    {
+        self.isAppActive = NO;
+
+        [self ensureBackgroundTaskState];
+    }
+}
+
+// Returns nil if adding this task _should have_ started a
```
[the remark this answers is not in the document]
charlesmchen: This method registers a new task with this manager.  We only bother requesting a background task from iOS if the app is inactive (or about to become inactive), so this will often not start a background task.

I've elaborated the comment.

SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -4,14 +4,290 @@
...
+// background task, but the background task couldn't be begun.
+// In that case expirationBlock will not be called.
+- (nullable OWSTaskId)addTask:(BackgroundTaskExpirationBlock)expirationBlock
```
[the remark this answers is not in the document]
charlesmchen: Done.

SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -4,14 +4,290 @@
...
+    OWSAssert(taskId);
+
+    @synchronized(self)
+    {
+        OWSAssert(self.expirationMap[taskId] != nil);
+
+        [self.expirationMap removeObjectForKey:taskId];
+
+        // Keep the background task active (if necessary) for an
+        // extra fraction of a second to provide continuity between
```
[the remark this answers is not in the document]
charlesmchen: Sorry, my comment was confusing.  It now reads:

```
        // This timer will ensure that we keep the background task active (if necessary)
        // for an extra fraction of a second to provide continuity between tasks.
        // This makes it easier and safer to use background tasks, since most code
        // should be able to ensure background tasks by "narrowly" wrapping
        // their core logic with a OWSBackgroundTask and not worrying about "hand off"
        // between OWSBackgroundTasks.
```
So this logic is dealing with that concern, not putting that responsibility on to the shoulders of code that uses `OWSBackgroundTask`.

SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -4,14 +4,290 @@
...
+    if (!CurrentAppContext().isMainApp) {
+        // We can't create background tasks in the SAE, but pretend that we succeeded.
+        return YES;
+    }
+
+    @synchronized(self)
+    {
+        // We only want to have a background task if we are:
+        // a) "not active" AND
+        // b1) there is more than one active instance of OWSBackgroundTask.
```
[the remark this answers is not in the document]
charlesmchen: Good catch.

SignalServiceKit/src/Util/OWSBackgroundTask.m
```diff
@@ -4,14 +4,290 @@
...
+            // otherwise.
+            //
+            // See:
+            // https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtaskwithexpiratio)
+            OWSAssert([NSThread isMainThread]);
+
+            [self backgroundTaskExpired];
+        }];
+
+        // If a background task could not be begun, call the completion block.
```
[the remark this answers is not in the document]
charlesmchen: Rewrote this comment.

SignalServiceKit/src/Util/OWSBackgroundTask.m
charlesmchen: Done.

charlesmchen:
PTAL @michaelkirk

charlesmchen:
PTAL @michaelkirk

michaelkirk:
LGTM!

PR #3064, opened 2018-02-26
fall back push notifications respect the configured notification sound 

michaelkirk:
Based on https://github.com/signalapp/Signal-iOS/pull/3054 so merge that first.


SignalMessaging/environment/OWSSounds.m
```diff
@@ -245,6 +246,36 @@ + (OWSSound)globalNotificationSound
...
+    NSString *dirPath = [[OWSFileSystem appLibraryDirectoryPath] stringByAppendingPathComponent:@"Sounds"];
+    [OWSFileSystem ensureDirectoryExists:dirPath];
+
+    // This name is specified in the payload by the Signal Service when requesting fallback push notifications.
+    NSString *kDefaultNotificationSoundFilename = @"NewMessage.aifc";
+    NSString *defaultSoundPath = [dirPath stringByAppendingPathComponent:kDefaultNotificationSoundFilename];
+
+    DDLogDebug(@"%@ writing new default sound to %@", self.logTag, defaultSoundPath);
+
+    NSURL *_Nullable soundURL = [OWSSounds soundURLForSound:sound quiet:NO];
```
charlesmchen: The notification sounds are in different formats - they're not all .aifc files.  It seems like we should generate valid .aifc versions of all of the sounds if we're going to do this.  I can help - it shouldn't be hard.  i.e. we shouldn't count on iOS correctly playing sound files that have a misleading file extension.

Signal.xcodeproj/project.pbxproj
```diff
@@ -901,6 +902,7 @@
 		459311FB1D75C948008DD4F0 /* OWSDeviceTableViewCell.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = OWSDeviceTableViewCell.m; sourceTree = "<group>"; };
 		4597E94E1D8313C100040CDE /* sq */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = sq; path = translations/sq.lproj/Localizable.strings; sourceTree = "<group>"; };
 		4597E94F1D8313CB00040CDE /* bg */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = bg; path = translations/bg.lproj/Localizable.strings; sourceTree = "<group>"; };
+		45A2F004204473A3002E978A /* NewMessage.aifc */ = {isa = PBXFileReference; lastKnownFileType = file; name = NewMessage.aifc; path = Signal/AudioFiles/NewMessage.aifc; sourceTree = SOURCE_ROOT; };
```
charlesmchen: It's weird you had to add this - I thought I already had this file in the project.  Or is this an additional copy?

michaelkirk: The file was in the repo, but removed from the project. 🤷‍♂️

SignalMessaging/environment/OWSSounds.m
michaelkirk: Oh, good catch!

> we shouldn't count on iOS correctly playing sound files that have a misleading file extension.

Totally agree.

michaelkirk: PTAL @charlesmchen - I converted the message tones to aifc via something like:

```
for file in *.m4r
do
ffmpeg  -i $file $(echo $file | sed 's/.m4r/.aifc')
done
```

And updated the -quite script accordingly.

PR #3065, opened 2018-02-26
Mkirk/migration splash

michaelkirk:
Waiting on https://github.com/signalapp/Signal-iOS/pull/3064 so review that first.

There is some art pending, which should arrive tomorrow, but that should be a drop in change.


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1788,6 +1785,15 @@
 /* Header for upgrade experience */
 "UPGRADE_EXPERIENCE_CALLKIT_TITLE" = "Just Swipe to Answer";
 
+/* Description for notification audio customization */
+"UPGRADE_EXPERIENCE_INTRODUCING_NOTIFICATION_AUDIO_DESCRIPTION" = "You can now specify default and per thread text tones. And calls now respect your system contact's ringtone.";
```
michaelkirk: This copy is provisional - I've tasked @jlund.

charlesmchen:
@michaelkirk this can be rebased now that  #3064 is merged, right?

michaelkirk:
> @michaelkirk this can be rebased now that #3064 is merged, right?

Done. PTAL @charlesmchen

Signal.xcodeproj/project.pbxproj
```diff
@@ -2914,7 +2907,6 @@
 			files = (
 				45F59A0A2029140500E8D2B0 /* OWSVideoPlayer.swift in Sources */,
 				344F249B200FD03300CFB4F4 /* SharingThreadPickerViewController.m in Sources */,
-				34F77040203F55D000C7CBB7 /* OWS107LegacySounds.m in Sources */,
```
charlesmchen: I suspect you're going to want to keep this migration if you want to delay upgrading legacy users to "note" until they've seen the splash, as you have it below.

Signal/src/ViewControllers/NotificationSettingsViewController.m
```diff
@@ -53,6 +53,10 @@ - (void)updateTableContents
                                    OWSSoundSettingsViewController *vc = [OWSSoundSettingsViewController new];
                                    [weakSelf.navigationController pushViewController:vc animated:YES];
                                }]];
+    [soundsSection addItem:[OWSTableItem switchItemWithText:NSLocalizedString(@"NOTIFICATIONS_SECTION_INAPP", nil)
```
charlesmchen: Since you're changing the copy of `NOTIFICATIONS_SECTION_INAPP`, should we add a comment now too?

Signal.xcodeproj/project.pbxproj
michaelkirk: Did you see that setting the default sound now happens when the splash screen appears?

Signal/src/ViewControllers/NotificationSettingsViewController.m
michaelkirk: Good call - done!

michaelkirk:
> LGTM but screenshots but would be super helpful so we can review this in iOS standup.

I'll provide an updated one when we have non-preliminary copy/art, otherwise it's the same upgrade splash we always use.

Signal.xcodeproj/project.pbxproj
charlesmchen: I see this:
```
        OWSSounds.setGlobalNotificationSound(.note)
```
But this won't have any effect if you get rid of `OWS107LegacySounds` since the default will have already changed.  I guess it might affect the fallback PN sound.  We can discuss if this isn't clear...

michaelkirk: You're right, thanks for patiently re-explaining =)

PR #3070, opened 2018-02-27
Fix sharing url when text is also present

michaelkirk:
This is sort of a stop gap intended to be minimal and safe.

PTAL @charlesmchen

### DIFF
--- a/SignalShareExtension/Info.plist
+++ b/SignalShareExtension/Info.plist
@@ -61,11 +61,11 @@
                 $extensionItem.attachments,
                 $attachment,
                 (
-                ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO "public.data"
-                || ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO "public.url"
+                ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;public.data&quot;
+                || ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;public.url&quot;
                 )
-                AND NOT (ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO "public.vcard")
-                ).@count == 1
+                AND NOT (ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;public.vcard&quot;)
+                ).@count &gt;= 1
                 ).@count == 1
             </string>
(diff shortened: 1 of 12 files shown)


SignalShareExtension/Info.plist
```diff
@@ -61,11 +61,11 @@
...
                 (
-                ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO "public.data"
-                || ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO "public.url"
+                ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;public.data&quot;
+                || ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;public.url&quot;
                 )
-                AND NOT (ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO "public.vcard")
-                ).@count == 1
+                AND NOT (ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;public.vcard&quot;)
+                ).@count &gt;= 1
```
michaelkirk: Huh - xcode decided to escape this field.

The only significant change is this one. We still only accept one attachment, but it's ok if that attachment matches more than one UTI type.

SignalShareExtension/ShareViewController.swift
```diff
@@ -574,9 +593,14 @@ public class ShareViewController: UIViewController, ShareViewDelegate, SAEFailed
...
         }
 
-        // TODO Multiple attachments. In that case I'm unclear if we'll
-        // be given multiple inputItems or a single inputItem with multiple attachments.
-        guard let itemProvider: NSItemProvider = inputItem.attachments?.first as? NSItemProvider else {
+        // A single inputItem can have multiple attachments, e.g. sharing from Firefox gives
+        // one url attachment and another text attachment, where the the url would be https://some-news.com/articles/123-cat-stuck-in-tree
+        // and the text attachment would be something like "Breaking news - cat stuck in tree"
+        //
+        // FIXME: For now, we prefer the URL provider and discard the text provider, since it's more useful to share the URL than the caption
```
michaelkirk: With this PR we only use the URL, not the suggested text. We should include the suggested text, but it's going to be tricky to apply that change in a non-disruptive way. 

There's quite a bit of machinery based on the assumption that there's only one item provider.

PR #3075, opened 2018-02-28
Fix group-sync disabling disappearing timer

michaelkirk:
Must add properties *before* build

PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Devices/OWSGroupsOutputStream.m
+++ b/SignalServiceKit/src/Devices/OWSGroupsOutputStream.m
@@ -35,11 +35,6 @@ NS_ASSUME_NONNULL_BEGIN
         [groupBuilder setAvatarBuilder:avatarBuilder];
     }
 
-    NSData *groupData = [[groupBuilder build] data];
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Devices/OWSGroupsOutputStream.m
```diff
@@ -35,11 +35,6 @@ - (void)writeGroup:(TSGroupModel *)group transaction:(YapDatabaseReadTransaction
...
     }
 
-    NSData *groupData = [[groupBuilder build] data];
-    uint32_t groupDataLength = (uint32_t)groupData.length;
-    [self.delegateStream writeRawVarint32:groupDataLength];
-    [self.delegateStream writeRawData:groupData];
-
     TSGroupThread *_Nullable groupThread = [TSGroupThread threadWithGroupId:group.groupId transaction:transaction];
     if (groupThread) {
         OWSDisappearingMessagesConfiguration *_Nullable disappearingMessagesConfiguration =
```
michaelkirk: At this point we'd already serialized the protobuf. Adding a new field at this point is too late.

PR #3078, opened 2018-03-01
Audit for missing reverse integration upon committing to master

michaelkirk:
This is meant as a nag. I'm not sure there'd be a good way to catch-us-in-the-act unless we migrate to some merge tools a la git-flow, and in that case I think we'd still want an audit like this.

PTAL @charlesmchen


Scripts/git_hooks/pre-commit
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import os
```
michaelkirk: This is just a straight copy of my existing script, which I'm pretty sure is untouched from what you provided @charlesmchen.

michaelkirk: Ah! Do you mind if I move it over top of this one? 

For dev setup, it would be nice if these were in one place.

michaelkirk: >  Can we de-dupe somehow?

Done! PTAL @charlesmchen

michaelkirk:
In ac8dbf81f

PR #3081, opened 2018-03-01
Charlesmchen/2fa vs registration

charlesmchen:
PTAL @michaelkirk 

simulator screen shot - iphone 8 plus - 2018-03-01 at 16 16 01
simulator screen shot - iphone 8 plus - 2018-03-01 at 16 15 58

### DIFF
--- a/Signal/src/ViewControllers/CodeVerificationViewController.m
+++ b/Signal/src/ViewControllers/CodeVerificationViewController.m
@@ -1,12 +1,13 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 #import "CodeVerificationViewController.h"
+#import "OWS2FARegistrationViewController.h"
 #import "ProfileViewController.h"
 #import "Signal-Swift.h"
-#import "UIViewController+OWS.h"
 #import <PromiseKit/AnyPromise.h>
+#import <SignalMessaging/UIViewController+OWS.h>
 #import <SignalServiceKit/OWSError.h>
 #import <SignalServiceKit/TSAccountManager.h>
 #import <SignalServiceKit/TSNetworkManager.h>
@@ -65,7 +66,13 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)viewDidLoad {
     [super viewDidLoad];
-    
+
+    self.navigationItem.backBarButtonItem =
+        [[UIBarButtonItem alloc] initWithTitle:NSLocalizedString(@"BACK_BUTTON", @"button text for back button")
+                                         style:UIBarButtonItemStylePlain
+                                        target:self
+                                        action:@selector(backButtonWasPressed)];
+
     [self createViews];
     
     [self initializeKeyboardHandlers];
@@ -259,7 +266,7 @@ NS_ASSUME_NONNULL_BEGIN
     [self startActivityIndicator];
     OWSProdInfo([OWSAnalyticsEvents registrationRegisteringCode]);
     __weak CodeVerificationViewController *weakSelf = self;
-    [self.accountManager registerWithVerificationCode:[self validationCodeFromTextField]]
+    [self.accountManager registerWithVerificationCode:[self validationCodeFromTextField] pin:nil]
         .then(^{
             OWSProdInfo([OWSAnalyticsEvents registrationRegisteringSubmittedCode]);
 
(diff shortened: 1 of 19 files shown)


SignalServiceKit/src/Account/TSAccountManager.m
```diff
@@ -392,6 +394,13 @@ - (void)verifyAccountWithCode:(NSString *)verificationCode
                     failureBlock(userError);
                     break;
                 }
+                case 423: {
+                    DDLogError(@"%@ 2FA PIN required: %ld", self.logTag, error.code);
+                    NSError *error = OWSErrorWithCodeDescription(
+                        OWSErrorCodeRegistrationMissing2FAPIN, @"Registration missing 2FA PIN.");
```
charlesmchen: We don't need to l10n this string since it's never presented to the user.

Signal/src/ViewControllers/CodeVerificationViewController.m
```diff
@@ -65,7 +66,13 @@ - (instancetype)init
 
 - (void)viewDidLoad {
     [super viewDidLoad];
-    
+
+    self.navigationItem.backBarButtonItem =
+        [[UIBarButtonItem alloc] initWithTitle:NSLocalizedString(@"BACK_BUTTON", @"button text for back button")
```
michaelkirk: Why can't we use a normal back button?

SignalMessaging/contacts/ViewControllerUtils.m
```diff
@@ -12,7 +12,9 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-#pragma mark -
+// TODO: ?
```
michaelkirk: What's TODO here?

Signal/src/ViewControllers/OWS2FARegistrationViewController.m
```diff
@@ -0,0 +1,291 @@
...
+                                                                 @"auth' failed.")];
+
+                                      [weakSelf.pinTextfield becomeFirstResponder];
+                                  }];
+                              });
+                          });
+                  }];
+}
+
+- (void)vericationWasCompleted
```
michaelkirk: s/vericationWasCompleted/verificationWasCompleted

Signal/src/ViewControllers/OWS2FARegistrationViewController.m
```diff
@@ -0,0 +1,291 @@
...
+    OWSAssert(self.hasValidPin);
+    OWSAssert(self.verificationCode.length > 0);
+    NSString *pin = self.pinTextfield.text;
+    OWSAssert(pin.length > 0);
+
+    DDLogInfo(@"%@ %s", self.logTag, __PRETTY_FUNCTION__);
+
+    __weak OWS2FARegistrationViewController *weakSelf = self;
+
+    [ModalActivityIndicatorViewController
```
michaelkirk: Do we use this progress UI on any of the other steps in registration flow? This feels like an outlier.

Signal/src/ViewControllers/OWS2FARegistrationViewController.m
```diff
@@ -0,0 +1,291 @@
...
+                          })
+                          .catch(^(NSError *error) {
+                              OWSAssertIsOnMainThread();
+                              OWSProdInfo([OWSAnalyticsEvents registrationRegistrationFailed]);
+                              DDLogError(@"%@ error verifying challenge: %@", weakSelf.logTag, error);
+                              dispatch_async(dispatch_get_main_queue(), ^{
+                                  [modalActivityIndicator dismissWithCompletion:^{
+                                      OWSAssertIsOnMainThread();
+
+                                      [OWSAlerts
```
michaelkirk: What if the error is due to something besides wrong PIN e.g. rate limit? or network is down? 

Since we're not getting much value out of the generic error title anyway, what about something like:

```
[OWSAlerts showAlertWithTitle:NSLocalizedString(@"REGISTER_2FA_REGISTRATION_FAILED", ...)
      message:error.localizedString];
```

SignalMessaging/contacts/ViewControllerUtils.m
charlesmchen: Done.

Signal/src/ViewControllers/CodeVerificationViewController.m
charlesmchen: You're right - we need this in the app settings (where the default back button labels are way too long), but not here.  Removed.

Signal/src/ViewControllers/OWS2FARegistrationViewController.m
charlesmchen: Done.

Signal/src/ViewControllers/OWS2FARegistrationViewController.m
charlesmchen: We don't yet, but I am actually about to add a card suggesting we do that.  IMO we should use the same UI pattern everywhere in our app and `ModalActivityIndicatorViewController` has a bunch of advantages over what we have now in the registration flow.

* Easier for developers to use.
* Is more consistent with how other apps do blocking or modal activity indicators like WA.
* Our registration views use "in-button" activity indicators that can lead to broken layout for buttons with long labels (e.g. issues will only show up in some l10ns).

Signal/src/ViewControllers/OWS2FARegistrationViewController.m
charlesmchen: Good idea, thanks.

PR #3082, opened 2018-03-02
Fix redundant sync sends.

charlesmchen:
Took another whack at this.

PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Devices/OWSDevice.m
+++ b/SignalServiceKit/src/Devices/OWSDevice.m
@@ -19,9 +19,6 @@ NSString *const kTSStorageManager_MayHaveLinkedDevices = @"kTSStorageManager_May
 
 @interface OWSDeviceManager ()
 
-// This property should only be accessed while synchronized on self.
-@property (atomic, nullable) NSNumber *mayHaveLinkedDevicesCached;
-
 @property (atomic) NSDate *lastReceivedSyncMessage;
 
 @end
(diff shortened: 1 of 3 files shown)


SignalServiceKit/src/Devices/OWSDevice.m
```diff
@@ -19,9 +19,6 @@
 
 @interface OWSDeviceManager ()
 
-// This property should only be accessed while synchronized on self.
-@property (atomic, nullable) NSNumber *mayHaveLinkedDevicesCached;
```
charlesmchen: I'm trying to cull all caching like this.

SignalServiceKit/src/Devices/OWSDevice.m
```diff
@@ -49,30 +46,34 @@ - (BOOL)mayHaveLinkedDevices:(YapDatabaseConnection *)dbConnection
...
-    @synchronized(self)
-    {
-        if (!self.mayHaveLinkedDevicesCached) {
-            self.mayHaveLinkedDevicesCached = @([dbConnection boolForKey:kTSStorageManager_MayHaveLinkedDevices
-                                                            inCollection:kTSStorageManager_OWSDeviceCollection
-                                                            defaultValue:YES]);
-        }
+    return [dbConnection boolForKey:kTSStorageManager_MayHaveLinkedDevices
+                       inCollection:kTSStorageManager_OWSDeviceCollection
+                       defaultValue:YES];
```
charlesmchen: We rightfully default to YES to avoid dropping sync sends for legacy clients.

SignalServiceKit/src/Devices/OWSDevice.m
```diff
@@ -49,30 +46,34 @@ - (BOOL)mayHaveLinkedDevices:(YapDatabaseConnection *)dbConnection
...
-        return [self.mayHaveLinkedDevicesCached boolValue];
-    }
+// In order to avoid skipping necessary sync messages, the default value
+// for mayHaveLinkedDevices is YES.  Once we've successfully sent a
+// sync message with no device messages (e.g. the service has confirmed
+// that we have no linked devices), we can set mayHaveLinkedDevices to NO
+// to avoid unnecessary message sends for sync messages until we learn
+// of a linked device (e.g. through the device linking UI or by receiving
+// a sync message, etc.).
+- (void)clearMayHaveLinkedDevicesIfNotSet
```
charlesmchen: But as soon as we know that these sends aren't necessary, we should start skipping them... until we have new reason they might be necessary.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -982,13 +983,28 @@ - (void)sendMessageToService:(TSOutgoingMessage *)message
         OWSAssert(deviceMessages.count > 0);
     }
 
+    if (deviceMessages.count == 0) {
+        DDLogWarn(@"%@ Sending a message with no device messages.", self.logTag);
```
michaelkirk: Is the only time that we expect this when sending a sync message to ourself for the first time? Or are there other occasions?

charlesmchen: I think this might occur in other rare edge cases, e.g. after unlinking all linked devices and after trying and failing to link a device.

charlesmchen: I'll leave a comment to this effect.

PR #3085, opened 2018-03-02
Registration Lock reminder view

michaelkirk:
PTAL @charlesmchen 

screen shot 2018-03-02 at 1 18 38 pm


Signal/src/ViewControllers/OWS2FARegistrationViewController.m
```diff
@@ -60,171 +56,53 @@ - (void)viewDidLoad
...
+    PinEntryView *entryView = [PinEntryView new];
+    self.entryView = entryView;
+    entryView.delegate = self;
+    [self.view addSubview:entryView];
+    
+    entryView.instructionsText = NSLocalizedString(@"REGISTER_2FA_INSTRUCTIONS", @"Instructions to enter the 'two-factor auth pin' in the 2FA registration view.");
+    
+    // Layout
+    
+    [entryView autoPinEdgesToSuperviewMargins];
```
charlesmchen: Is this safe for iPhone X?  I thought we had to use `autoPinTopToSuperviewWithMargin` for top and bottom for this kind of root view.

Signal/src/ViewControllers/OWS2FAReminderViewController.swift
```diff
@@ -0,0 +1,110 @@
...
+        self.pinEntryView = pinEntryView
+        pinEntryView.delegate = self
+        let instructionsText = NSLocalizedString("REMINDER_2FA_BODY", comment: "Body text for when user is peridoically prompted to enter their registration lock PIN")
+        pinEntryView.instructionsText = instructionsText
+
+        view.addSubview(pinEntryView)
+
+        pinEntryView.autoPinWidthToSuperview(withMargin: 20)
+        pinEntryView.autoPin(toTopLayoutGuideOf: self, withInset: 0)
+        pinEntryView.autoPin(toBottomLayoutGuideOf: self, withInset: 0)
```
charlesmchen: 8-)

Signal/src/views/PinEntryView.h
```diff
@@ -0,0 +1,30 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class PinEntryView;
+
+@protocol PinEntryViewDelegate <NSObject>
```
charlesmchen: Thanks for pulling this out.

Signal/src/views/PinEntryView.m
```diff
@@ -0,0 +1,205 @@
...
+    const CGFloat kVSpacing = 30.f;
+
+    UILabel *instructionsLabel = [self createLabelWithText:nil];
+    self.instructionsLabel = instructionsLabel;
+    [instructionsLabel autoPinTopToSuperviewWithMargin:kVSpacing];
+    [instructionsLabel autoPinWidthToSuperview];
+
+    UILabel *createForgotLink = [self createForgotLink];
+    [createForgotLink autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:instructionsLabel withOffset:5];
+    [createForgotLink autoPinWidthToSuperview];
```
charlesmchen: But we don't want a forgot link in the "set PIN settings" views, right?

Signal/src/views/PinEntryView.m
```diff
@@ -0,0 +1,205 @@
...
+    self.pinTextfield.textAlignment = NSTextAlignmentCenter;
+    [self addSubview:self.pinTextfield];
+}
+
+- (UILabel *)createForgotLink
+{
+    UILabel *label = [UILabel new];
+    label.textColor = [UIColor ows_materialBlueColor];
+    NSString *text = NSLocalizedString(
+        @"REGISTER_2FA_FORGOT_PIN", @"Label for 'I forgot my PIN' link in the 2FA registration view.");
```
charlesmchen: I suspect we'll want different "forgot" copy in the registration w/ pin view (where the stakes are high) v. the reminder view (where the stakes are low).  They have different resolutions too: in the former, you wait.  In the latter, we might want to offer to clear the PIN and/or let users select a new PIN if they forgot it.

SignalServiceKit/src/Util/OWS2FAManager.m
```diff
@@ -129,6 +140,113 @@ - (void)disable2FAWithSuccess:(nullable OWS2FASuccess)success failure:(nullable
...
+- (NSDate *)nextReminderDate
+{
+    NSDate *lastSuccessfulReminderDate = self.lastSuccessfulReminderDate ?: [NSDate distantPast];
+    
+    return [lastSuccessfulReminderDate dateByAddingTimeInterval:self.repetitionInterval];
+}
+
+- (NSArray<NSNumber *> *)allRepetitionIntervals
+{
+    // Keep sorted monotonically increasing.
```
charlesmchen: So every time they successfully guess, we "back off" and remind them less.  When they fail to guess or set a new PIN, shouldn't we reset the repetition interval to zero?

Signal/src/views/PinEntryView.m
michaelkirk: I agree that the *behavior* of tapping "I forgot my PIN" should be different depending on context (that's achieved via the delegate method `pinEntryViewForgotPinLinkTapped`), but the actual text of the button which prompts that behavior: "I forgot my PIN" is the same in both places, which seems reasonable to me.

Signal/src/views/PinEntryView.m
michaelkirk: This isn't (currently?) used in the settings view. It's used in the registration and reminder views, both of which want a "I forgot my PIN" button.

SignalServiceKit/src/Util/OWS2FAManager.m
michaelkirk: Per the spec, I think we want to just ratchet them back one repetition interval.

Signal/src/ViewControllers/OWS2FARegistrationViewController.m
michaelkirk: I'd like to address this in the pursuit of: "Make 2FA screen look more like the rest of the registration flow." since the layout here's going to get changed regardless.

michaelkirk:
In 07ee3ea84

PR #3087, opened 2018-03-02
Fix profile avatar downloads.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/profiles/OWSProfileManager.m
+++ b/SignalMessaging/profiles/OWSProfileManager.m
@@ -912,25 +912,13 @@ const NSUInteger kOWSProfileManager_MaxAvatarDiameter = 640;
             return;
         }
 
-        // If we're transitioning from (no avatar -> no avatar) or from (same avatar -> same avatar),
-        // don't bother updating the avatar.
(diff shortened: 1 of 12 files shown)


SignalMessaging/profiles/OWSProfileManager.m
```diff
@@ -942,7 +947,7 @@ - (void)updateProfileForRecipientId:(NSString *)recipientId
                                          completion:nil];
         }
 
-        if (userProfile.avatarUrlPath.length > 0 && userProfile.avatarFileName.length == 0) {
+        if (avatarUrlPath.length > 0 && !canSkipAvatarUpdate) {
```
michaelkirk: Isn't this change (and calculating `canSkipAvatarUpdate`) redundant now that we've fixed the avatarUrlPath setter?

The old mental-model, used only the state of the OWSUserProfile model to determine if it needed to download a new avatar.

In other words, anytime we have a non-empty userProfile.avatarFileName, it means we've previously successfully downloaded the avatar specified by userProfile.avatarUrlPath.

Thus, so long as we correctly invalidate avatarFileName upon setting avatarUrlPath, we should be good, right?

charlesmchen: Good point - we can simplify.  Done, and added a comment.

PR #3088, opened 2018-03-02
Make default audio tone more prominent

michaelkirk:
Small UX improvement

Label Note as (default)
Move it to the second item in the list, just after "None"

audio-tone-default

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/OWSSoundSettingsViewController.m
+++ b/Signal/src/ViewControllers/OWSSoundSettingsViewController.m
@@ -74,13 +74,25 @@ NS_ASSUME_NONNULL_BEGIN
     for (NSNumber *nsValue in allSounds) {
         OWSSound sound = (OWSSound)nsValue.intValue;
         OWSTableItem *item;
+
+        NSString *soundLabelText = ^{
+            NSString *baseName = [OWSSounds displayNameForSound:sound];
+            if (sound == OWSSound_Note) {
+                NSString *noteStringFormat = NSLocalizedString(@"SETTINGS_AUDIO_DEFAULT_TONE_LABEL_FORMAT",
+                    @"Format string for the default 'Note' sound. Embeds the system {{sound name}}.");
+                return [NSString stringWithFormat:noteStringFormat, baseName];
+            } else {
+                return [OWSSounds displayNameForSound:sound];
+            }
+        }();
+
         if (sound == self.currentSound) {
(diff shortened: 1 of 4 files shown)


SignalMessaging/environment/migrations/OWSDatabaseMigrationRunner.m
```diff
@@ -65,10 +65,7 @@ - (void)runMigrations:(NSArray<OWSDatabaseMigration *> *)migrations
 
     NSMutableArray<OWSDatabaseMigration *> *migrationsToRun = [NSMutableArray new];
     for (OWSDatabaseMigration *migration in migrations) {
-        if ([OWSDatabaseMigration fetchObjectWithUniqueID:migration.uniqueId]) {
-            DDLogDebug(@"%@ Skipping previously run migration: %@", self.logTag, migration);
-        } else {
-            DDLogWarn(@"%@ Running migration: %@", self.logTag, migration);
```
michaelkirk: We're redundantly logging which migrations get run a few lines down.

Signal/src/ViewControllers/OWSSoundSettingsViewController.m
```diff
@@ -74,13 +74,25 @@ - (void)updateTableContents
     for (NSNumber *nsValue in allSounds) {
         OWSSound sound = (OWSSound)nsValue.intValue;
         OWSTableItem *item;
+
+        NSString *soundLabelText = ^{
+            NSString *baseName = [OWSSounds displayNameForSound:sound];
+            if (sound == OWSSound_Note) {
+                NSString *noteStringFormat = NSLocalizedString(@"SETTINGS_AUDIO_DEFAULT_TONE_LABEL_FORMAT",
+                    @"Format string for the default 'Note' sound. Embeds the system {{sound name}}.");
+                return [NSString stringWithFormat:noteStringFormat, baseName];
```
charlesmchen: This is fine, but lately, I've been doing ("string a" + [" "] + "string b") using the rtl-safe string concatenation so that we don't have to count on translator to get the RTL ordering correct, since they have less context than we do.

michaelkirk:
In 716db9f4c

Signal/src/ViewControllers/OWSSoundSettingsViewController.m
michaelkirk: Good point.

PR #3089, opened 2018-03-02
Fix migration to work with fallback notifications as well

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/SignalMessaging/environment/OWSSounds.m
+++ b/SignalMessaging/environment/OWSSounds.m
@@ -244,6 +244,27 @@ NSString *const kOWSSoundsStorageGlobalNotificationKey = @"kOWSSoundsStorageGlob
 
 + (void)setGlobalNotificationSound:(OWSSound)sound
 {
+    [self.sharedManager setGlobalNotificationSound:sound];
+}
+
+- (void)setGlobalNotificationSound:(OWSSound)sound
+{
+    [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
+        [self setGlobalNotificationSound:sound transaction:transaction];
+    }];
+}
+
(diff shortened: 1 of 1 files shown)


SignalMessaging/environment/OWSSounds.m
```diff
@@ -274,17 +295,6 @@ + (void)setGlobalNotificationSound:(OWSSound)sound
         return;
     }
```
michaelkirk: There's more to do than just set a DB key - we were doing it in the non-transaction flavor of this method, but not in the transaction flavor.

I've re-organized to follow our usual pattern of doing the substantial work in the transaction-taking flavor, and having the transaction-creating flavor simply call the transaction-taking method.

charlesmchen: So with this fix:

* Legacy users will still hear the classic clunk _before_ they see the sound splash... for fallback notifications?
* Legacy users will hear the note sound _after_ they see the sound splash... for fallback notifications?

Also, some day we will stop showing the sound notifications splash (perhaps because there is another splash).  How will we migrate users then?

michaelkirk: So with this fix:

> Legacy users will still hear the classic clunk before they see the sound splash... for fallback notifications?

Affirmative

> Legacy users will hear the note sound after they see the sound splash... for fallback notifications?

Affirmative

> Also, some day we will stop showing the sound notifications splash (perhaps because there is another splash). How will we migrate users then?

Well, depending how long it goes, this will be a pretty small subset, but at such a time we could either:

1. Do nothing. This would leave this small subset of users who haven't launched the Signal App in a couple of months with the ka-klunk until they manually update sounds in their settings.

2. Migrate them other way. 

I think option #1 is pretty tenable for the small tail, considering we were originally considering doing this for *all* legacy users.

PR #3090, opened 2018-03-02
Charlesmchen/account attributes vs pin

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Util/OWS2FAManager.m
+++ b/SignalServiceKit/src/Util/OWS2FAManager.m
@@ -73,14 +73,23 @@ const NSUInteger kDaySecs = kHourSecs * 24;
 
 - (BOOL)is2FAEnabled
 {
-    return [self.dbConnection boolForKey:kOWS2FAManager_IsEnabledKey
-                            inCollection:kOWS2FAManager_Collection
-                            defaultValue:NO];
+    return [self.dbConnection hasObjectForKey:kOWS2FAManager_PINKey inCollection:kOWS2FAManager_Collection];
 }
 
-- (void)setIs2FAEnabled:(BOOL)value
+- (void)set2FANotEnabled
 {
-    [self.dbConnection setBool:value forKey:kOWS2FAManager_IsEnabledKey inCollection:kOWS2FAManager_Collection];
+    [self.dbConnection removeObjectForKey:kOWS2FAManager_PINKey inCollection:kOWS2FAManager_Collection];
+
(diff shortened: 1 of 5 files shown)


SignalServiceKit/src/Util/OWS2FAManager.m
```diff
@@ -65,16 +65,30 @@ - (instancetype)initWithStorageManager:(TSStorageManager *)storageManager
...
+{
+    [self.dbConnection removeObjectForKey:kOWS2FAManager_PINKey inCollection:kOWS2FAManager_Collection];
+
+    [[NSNotificationCenter defaultCenter] postNotificationNameAsync:NSNotificationName_2FAStateDidChange
+                                                             object:nil
+                                                           userInfo:nil];
 }
 
-- (void)setIs2FAEnabled:(BOOL)value
+- (void)set2FAEnabledWithPin:(NSString *)pin
```
michaelkirk: I had to make similar changes in #3085 - can you use those?

charlesmchen:
@michaelkirk I'm going to wait until #3085 is merged before moving ahead with this.

SignalServiceKit/src/Util/OWS2FAManager.m
charlesmchen: Rebased and reconciled.  I drew from both but crucially we now have a single piece of state - a "registration lock" pin.  If it's set, "registration lock" is enabled.  Less state = less surface area for bugs.

charlesmchen:
227
3,187 761 signalapp/Signal-iOS
 Code  Issues 331  Pull requests 30  Projects 0  Wiki  Insights  Settings

charlesmchen:
PTAL @michaelkirk

SignalServiceKit/src/Util/OWS2FAManager.m
```diff
@@ -71,16 +70,19 @@ - (instancetype)initWithStorageManager:(TSStorageManager *)storageManager
...
+{
+    return [self.dbConnection objectForKey:kOWS2FAManager_PinCode inCollection:kOWS2FAManager_Collection];
+}
+
 - (BOOL)is2FAEnabled
 {
-    return [self.dbConnection boolForKey:kOWS2FAManager_IsEnabledKey
-                            inCollection:kOWS2FAManager_Collection
-                            defaultValue:NO];
+    return self.pinCode != nil;
```
michaelkirk: Thanks for getting rid of this redundant state!

PR #3095, opened 2018-03-04
Fix blue new message dot

### LINKED ISSUE #2901: In conversation new, active state does not maintain 'blue new message' counter's background

### Bug description
When in the conversation view, tapping (tap and hold to see active state better) on a conversation that has new messages causes the blue 'new message counter' to disappear. It should remain visible.

### Steps to reproduce
- View conversation list
- One of the conversations must have received new messages so the blue, new message counter is visible
- Tap (& hold to more easily see this) on the new message. The background will go gray as expected, but the blue 'new message counter' disappears. It should remain blue.

**Actual result:** The background goes gray and the blue 'new message counter' disappears. 
**Expected result:** The background goes gray and the blue 'new message counter' should remain visible.

### Device info
**Device**: iPhone 6
**iOS version**: 11.2.1
**Signal version:** 2.19.3.3

ftlno:
// FREEBIE

### Contributor checklist

 * iDevice A, iPhone 7 Plus, iOS 11.2.6
 * iDevice B, iPhone 8, iOS 11.2.6
 * Simulator , iPhone 7 iOS 11.0
- - - - - - - - - -

### Description
Fixes #2901 

When a row is selected in UITableView, background colors of all UIView subviews gets changed. This class will prevent that from happening so that the blue dot can remain blue while the row is selected. 
Description and solution suggestions found in the following link:
https://stackoverflow.com/questions/6745919/uitableviewcell-subview-disappears-when-cell-is-selected/27717607

I've tested this, see provided before and after screenshot. I've removed my phone number from the images.

before

after

PS: I found an unrelated additional bug while debugging, the Norwegian word for Inbox "Innboks" is longer, so that label "Innboks (1)"-label in the segment control gets truncated. I've created this https://github.com/signalapp/Signal-iOS/issues/3096

### DIFF
--- a/Signal/src/ViewControllers/Utils/NeverClearView.swift
+++ b/Signal/src/ViewControllers/Utils/NeverClearView.swift
@@ -1,5 +1,6 @@
-// Created by Fredrik Lillejordet on 04.03.2018.
-// Copyright © 2018 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
 
 @objc class NeverClearView: UIView {
     override var backgroundColor: UIColor? {
--- a/Signal/src/ViewControllers/Utils/NeverClearView.swift
+++ b/Signal/src/ViewControllers/Utils/NeverClearView.swift
@@ -1,7 +1,7 @@
 // Created by Fredrik Lillejordet on 04.03.2018.
 // Copyright © 2018 Open Whisper Systems. All rights reserved.
 
-@objc class NeverClearUIView: UIView {
+@objc class NeverClearView: UIView {
     override var backgroundColor: UIColor? {
         didSet {
             if backgroundColor != nil && backgroundColor!.cgColor.alpha == 0 {
--- a/Signal/src/ViewControllers/Utils/NeverClearView.swift
+++ b/Signal/src/ViewControllers/Utils/NeverClearView.swift
@@ -1,0 +1,12 @@
+// Created by Fredrik Lillejordet on 04.03.2018.
+// Copyright © 2018 Open Whisper Systems. All rights reserved.
+
+@objc class NeverClearUIView: UIView {
+    override var backgroundColor: UIColor? {
+        didSet {
+            if backgroundColor != nil && backgroundColor!.cgColor.alpha == 0 {
+                backgroundColor = oldValue
+            }
+        }
+    }
+}
--- a/Signal/src/ViewControllers/Utils/NeverClearView.swift
+++ b/Signal/src/ViewControllers/Utils/NeverClearView.swift
@@ -5,7 +5,7 @@
 @objc class NeverClearView: UIView {
     override var backgroundColor: UIColor? {
         didSet {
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/Utils/NeverClearView.swift
```diff
@@ -0,0 +1,12 @@
+// Created by Fredrik Lillejordet on 04.03.2018.
+// Copyright © 2018 Open Whisper Systems. All rights reserved.
+
+@objc class NeverClearView: UIView {
+    override var backgroundColor: UIColor? {
+        didSet {
+            if backgroundColor != nil && backgroundColor!.cgColor.alpha == 0 {
```
michaelkirk-signal: nit: prefer implicit unwrap instead of `!` whenever feasible.

i.e.

```
if let backgroundColor = self.backgroundColor, backgroundColor.cgColor.alpha == 0 {
...
```

Signal/src/ViewControllers/Utils/NeverClearView.swift
```diff
@@ -0,0 +1,12 @@
+// Created by Fredrik Lillejordet on 04.03.2018.
+// Copyright © 2018 Open Whisper Systems. All rights reserved.
```
michaelkirk-signal: Please run the Scripts/hooks/post-commit to get the right headers (or copy from another file)

Signal/src/ViewControllers/Utils/NeverClearView.swift
michaelkirk-signal: Outside of the context of this PR, this code is a bit arcane. Could you leave a comment in the code to explain how selecting a cell affects the background color requiring this workaround to avoid recoloring the view?

ftlno:
Thanks for the thorough feedback! I will improve and commit the addressed issues, rebase and tag Charles when ready.

Signal/src/ViewControllers/Utils/NeverClearView.swift
ftlno: ✅ Done!

ftlno:
PTAL @charlesmchen :)

ftlno:
What about this @charlesmchen-signal? It seems like it needs to be rebased again?

ftlno:
@michaelkirk-signal PTAL :)

Signal/src/ViewControllers/Utils/NeverClearView.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+@objc class NeverClearView: UIView {
+    override var backgroundColor: UIColor? {
+        didSet {
+            if backgroundColor?.cgColor.alpha == 0 {
```
michaelkirk-signal: Outside of the context of this PR, this code is a bit arcane. Could you leave a comment in the code to explain how selecting a cell affects the background color requiring this workaround to avoid recoloring the view?

Otherwise, LGTM.

PTAL @charlesmchen

charlesmchen-signal:
Merged in d220883b5.

PR #3101, opened 2018-03-05
Make 2FA registration screen look like the rest of registration

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/OWS2FARegistrationViewController.m
+++ b/Signal/src/ViewControllers/OWS2FARegistrationViewController.m
@@ -54,6 +54,15 @@ NS_ASSUME_NONNULL_BEGIN
 {
     [super viewDidLoad];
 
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/OWS2FARegistrationViewController.m
```diff
@@ -65,7 +74,6 @@ - (void)viewDidLoad
         @"REGISTER_2FA_INSTRUCTIONS", @"Instructions to enter the 'two-factor auth pin' in the 2FA registration view.");
 
     // Layout
-
     [entryView autoPinEdgesToSuperviewMargins];
```
michaelkirk: Since we're pinning to margins here, we're all good on iPhoneX

screen shot 2018-03-05 at 11 47 32 am

PR #3105, opened 2018-03-05
Try to keep-alive registration lock w/ bg fetch

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -20,6 +20,7 @@
 #import "SignalsNavigationController.h"
 #import "ViewControllerUtils.h"
 #import <AxolotlKit/SessionCipher.h>
+#import <PromiseKit/AnyPromise.h>
 #import <SignalMessaging/AppSetup.h>
 #import <SignalMessaging/Environment.h>
 #import <SignalMessaging/OWSContactsManager.h>
@@ -205,6 +206,10 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(registrationStateDidChange)
                                                  name:RegistrationStateDidChangeNotification
+                                               object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(registrationLockDidChange:)
+                                                 name:NSNotificationName_2FAStateDidChange
                                                object:nil];
 
     DDLogInfo(@"%@ application: didFinishLaunchingWithOptions completed.", self.logTag);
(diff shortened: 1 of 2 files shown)


Signal/src/AppDelegate.m
```diff
@@ -594,6 +599,17 @@ - (void)applicationDidBecomeActive:(UIApplication *)application {
     DDLogInfo(@"%@ applicationDidBecomeActive completed.", self.logTag);
 }
 
+- (void)enableBackgroundRefreshIfNecessary
+{
+    if (OWS2FAManager.sharedManager.is2FAEnabled && [TSAccountManager isRegistered]) {
```
charlesmchen: So this is safe to before "app is ready".

Signal/src/AppDelegate.m
```diff
@@ -988,7 +1006,7 @@ - (void)application:(UIApplication *)application didReceiveLocalNotification:(UI
 - (void)application:(UIApplication *)application
     handleActionWithIdentifier:(NSString *)identifier
           forLocalNotification:(UILocalNotification *)notification
-             completionHandler:(void (^)())completionHandler
+             completionHandler:(void (^)(void))completionHandler
```
charlesmchen: I've seen these compiler warnings... but XCode (9.2?) was turning them into errors when I fixed this before.

Signal/src/AppDelegate.m
```diff
@@ -1039,6 +1057,28 @@ - (void)application:(UIApplication *)application
...
+        __block AnyPromise *job = [[SignalApp sharedApp].messageFetcherJob run].then(^{
+            // HACK: Call completion handler after n seconds.
+            //
+            // We don't currently have a convenient API to know when message fetching is *done* when
+            // working with the websocket.
+            //
+            // We *could* substantially rewrite the TSSocketManager to take advantage of the `empty` message
+            // But once our REST endpoint is fixed to properly de-enqueue fallback notifications, we can easily
+            // use the rest endpoint here rather than the websocket and circumvent making changes to critical code.
+            dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(5 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
```
charlesmchen: So my reading of this comment is that five is simply a very conservative estimate for "log enough to receive and ack but not necessarily process at least 1 message/one empty queue message."

michaelkirk: Correct

Signal/src/AppDelegate.m
michaelkirk: I'll revert this for now to avoid breaking anyones build.

Signal/src/AppDelegate.m
michaelkirk: I took care to only run it in other blocks protected by app-readiness, but being redundant here is a good idea in case it would get moved. Updated!

Signal/src/AppDelegate.m
michaelkirk: Actually, I kept the change after verifying it still builds on 9.2

PR #3106, opened 2018-03-05
In-App notifications don't pause background audio

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -3203,7 +3203,7 @@ typedef enum : NSUInteger {
     NSURL *fileURL = [NSURL fileURLWithPath:filepath];
 
     // Setup audio session
-    BOOL configuredAudio = [OWSAudioSession.shared setRecordCategoryWithAudioActivity:self.voiceNoteAudioActivity];
+    BOOL configuredAudio = [OWSAudioSession.shared startRecordingAudioActivity:self.voiceNoteAudioActivity];
     if (!configuredAudio) {
         OWSFail(@"%@ Couldn't configure audio session", self.logTag);
         [self cancelVoiceMemo];
--- a/Signal/src/ViewControllers/ExperienceUpgradesPageViewController.swift
+++ b/Signal/src/ViewControllers/ExperienceUpgradesPageViewController.swift
@@ -9,13 +9,6 @@ private class IntroducingCustomNotificationAudioExperienceUpgradeViewController:
 
     var buttonAction: ((UIButton) -> Void)?
 
-    override func viewDidAppear(_ animated: Bool) {
-        super.viewDidAppear(animated)
-
-        // Opt users in to the new default sound ("note") after they've seen the splash screen.
-        OWSSounds.setGlobalNotificationSound(.note)
-    }
-
     override func loadView() {
         self.view = UIView.container()
 
--- a/Signal/src/environment/NotificationsManager.m
+++ b/Signal/src/environment/NotificationsManager.m
@@ -25,6 +25,7 @@
 @property (nonatomic, readonly) NotificationType notificationPreviewType;
 
 @property (nonatomic, readonly) NSMutableArray<NSDate *> *notificationHistory;
+@property (nonatomic, nullable) OWSAudioPlayer *audioPlayer;
 
 @end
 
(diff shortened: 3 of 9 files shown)


SignalMessaging/environment/OWSSounds.m
```diff
@@ -224,7 +224,7 @@ - (void)playSound:(OWSSound)sound quiet:(BOOL)quiet shouldRespectSilentSwitch:(B
     [self.audioPlayer stop];
     self.audioPlayer = [OWSSounds audioPlayerForSound:sound quiet:quiet];
     if (shouldRespectSilentSwitch) {
-        [self.audioPlayer playWithCurrentAudioCategory];
+        [self.audioPlayer playWithAmbientAudioCategory];
```
michaelkirk: This is a regression from 2.20. 

Previously we used `AudioServicesPlayAlertSound` which had behavior like Ambient+"ducks others"

At first I took a stab at restoring the usage of AudioServicesPlayerAlertSound, but I think actually the *ducks others* behavior is still pretty distracting and not necessary (e.g. see WA).

michaelkirk:
I'm going to make one more change. hold off on reviewing...

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3203,7 +3203,7 @@ - (void)startRecordingVoiceMemo
     NSURL *fileURL = [NSURL fileURLWithPath:filepath];
 
     // Setup audio session
-    BOOL configuredAudio = [OWSAudioSession.shared setRecordCategoryWithAudioActivity:self.voiceNoteAudioActivity];
+    BOOL configuredAudio = [OWSAudioSession.shared startRecordingAudioActivity:self.voiceNoteAudioActivity];
```
michaelkirk: just a rename

Signal/src/ViewControllers/ExperienceUpgradesPageViewController.swift
```diff
@@ -9,13 +9,6 @@ private class IntroducingCustomNotificationAudioExperienceUpgradeViewController:
 
     var buttonAction: ((UIButton) -> Void)?
 
-    override func viewDidAppear(_ animated: Bool) {
-        super.viewDidAppear(animated)
-
-        // Opt users in to the new default sound ("note") after they've seen the splash screen.
-        OWSSounds.setGlobalNotificationSound(.note)
```
michaelkirk: we don't want to migrate legacy user's audio automatically.

Signal/src/environment/NotificationsManager.m
```diff
@@ -343,8 +345,8 @@ - (void)notifyUserForIncomingMessage:(TSIncomingMessage *)message
         } else {
             if (shouldPlaySound && [Environment.preferences soundInForeground]) {
                 OWSSound sound = [OWSSounds notificationSoundForThread:thread];
-                // We play the "quiet" variation of sounds if possible for notifications in the foreground.
-                [OWSSounds playSound:sound quiet:YES shouldRespectSilentSwitch:YES];
```
michaelkirk: This class is the only place we were using the shouldRespectSilentSwitch behavior, we were never setting it to NO. 

So instead I wrapped it behind a method and moved the OWSSounds.sharedInstance.audioPlayer to this class.

SignalMessaging/environment/OWSSounds.m
```diff
@@ -282,11 +259,18 @@ - (void)setGlobalNotificationSound:(OWSSound)sound transaction:(YapDatabaseReadW
     DDLogDebug(@"%@ writing new default sound to %@", self.logTag, defaultSoundPath);
 
     NSURL *_Nullable soundURL = [OWSSounds soundURLForSound:sound quiet:NO];
-    OWSAssert(soundURL);
+
+    NSData *soundData = ^{
```
michaelkirk: "None" sound was broken for fallback notifications

SignalMessaging/utils/OWSAudioPlayer.m
```diff
@@ -99,11 +99,20 @@ - (void)playWithCurrentAudioCategory
...
 
     [self play];
 }
 
+- (void)playAsForegroundAlert
+{
+    OWSAssertIsOnMainThread();
+    [OWSAudioSession.shared startAmbientAudioActivity:self.audioActivity];
+
+    AudioServicesPlaySystemSound(kSystemSoundID_Vibrate);
```
michaelkirk: Replace the haptic feedback which we'd lost (inadvertently I believe) since 2.20

michaelkirk:
FYI @charlesmchen added a commit https://github.com/signalapp/Signal-iOS/pull/3106/commits/4e64b09ad6fd0c2d1a908a638d81552ed0c3bbc5

PR #3107, opened 2018-03-06
more conservative call logging

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/SignalMessaging/utils/OWSPreferences.m
+++ b/SignalMessaging/utils/OWSPreferences.m
@@ -7,6 +7,7 @@
 #import <SignalServiceKit/NSUserDefaults+OWS.h>
 #import <SignalServiceKit/TSStorageHeaders.h>
 #import <SignalServiceKit/YapDatabaseConnection+OWS.h>
+#import <SignalServiceKit/YapDatabaseTransaction+OWS.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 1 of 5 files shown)


SignalMessaging/utils/OWSPreferences.m
```diff
@@ -224,11 +230,41 @@ - (void)setIsSystemCallLogEnabled:(BOOL)flag
...
+    NSNumber *_Nullable callKitPreference =
+        [self tryGetValueForKey:OWSPreferencesKeyCallKitEnabled transaction:transaction];
+    BOOL wasUsingCallKit = callKitPreference ? [callKitPreference boolValue] : YES;
+
+    NSNumber *_Nullable callKitPrivacyPreference =
+        [self tryGetValueForKey:OWSPreferencesKeyCallKitPrivacyEnabled transaction:transaction];
+    BOOL wasUsingCallKitPrivacy = callKitPrivacyPreference ? callKitPrivacyPreference.boolValue : YES;
+
+    BOOL shouldLogCallsInRecents = ^{
+        if (wasUsingCallKit && !wasUsingCallKitPrivacy) {
```
michaelkirk: Per a conversation with jlund, we should be more conservative about which legacy users we opt in to using the "recents" list. 

Notably, people who left these settings as their default will have to manually opt-in to showing calls in the Recents list.

PR #3109, opened 2018-03-06
Show migration screen at first launch.

michaelkirk:
There used to be a migration that ran when the experience upgrade view controller became visible. In an effort to make sure that didn't happen until the user actually had the app in the foreground, I inadvertently suppressed the presentation of the experience upgrade for users explicitly launching the app (until they backgrounded and re-foregrounded the app).

Now that the migration has been removed, I've reverted the presentation logic of the upgrade experience VC to occur when HomeViewController.viewDidAppear.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeViewController.m
@@ -287,6 +287,22 @@ typedef NS_ENUM(NSInteger, CellState) { kArchiveState, kInboxState };
     [self updateBarButtonItems];
 }
 
+- (void)viewDidAppear:(BOOL)animated
+{
+    [super viewDidAppear:animated];
+
+    // Keep in mind viewDidAppear is called while the app is in the background if the app was
+    // launched by a voip notification. This is fine - it will remain visible to the user
+    // when they eventually launch the app, but we shouldn't make any changes assuming
+    // the user has *seen* the upgrade experience at this point.
+    if (!self.hasShownAnyUnseenUpgradeExperiences) {
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [self displayAnyUnseenUpgradeExperience];
+            self.hasShownAnyUnseenUpgradeExperiences = YES;
+        });
+    }
+}
+
 - (void)updateBarButtonItems
 {
     const CGFloat kBarButtonSize = 44;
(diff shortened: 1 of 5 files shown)


SignalMessaging/environment/OWSSounds.m
```diff
@@ -74,8 +74,8 @@ - (instancetype)initWithStorageManager:(TSStorageManager *)storageManager
         @(OWSSound_Keys),
         @(OWSSound_Popcorn),
         @(OWSSound_Pulse),
-        @(OWSSound_Synth),
         @(OWSSound_ClassicNotification),
+        @(OWSSound_Synth),
```
michaelkirk: I thought the bottom of the list was maybe an easier place to find the "Classic" notification, but now that it's called "Signal Classic" (right next to "Synth") it just looked like we had mis-sorted alphabetically.

Signal/src/ViewControllers/HomeViewController.m
```diff
@@ -287,6 +287,22 @@ - (void)viewDidLoad
     [self updateBarButtonItems];
 }
 
+- (void)viewDidAppear:(BOOL)animated
+{
+    [super viewDidAppear:animated];
+
+    // Keep in mind viewDidAppear is called while the app is in the background if the app was
+    // launched by a voip notification. This is fine - it will remain visible to the user
+    // when they eventually launch the app, but we shouldn't make any changes assuming
```
charlesmchen: // it will remain visible to the user when they eventually launch the app

Is that true?  What if:

* App is launched from voip.
* We show upgrade splash & mark all upgrades as "seen".
* App - still in the background - is killed by iOS.
* User launches app, never sees the upgrade.

SignalMessaging/environment/OWSSounds.m
charlesmchen: Maybe we should rename `OWSSound_ClassicNotification` to  `OWSSound_SignalClassic` to make it more obvious why the sort is in this order.

Signal/src/ViewControllers/HomeViewController.m
charlesmchen: Should we observe "app lifecycle events" and only show & mark as seen the upgrade UI when they are actually visible?

SignalMessaging/environment/OWSSounds.m
michaelkirk: Done

Signal/src/ViewControllers/HomeViewController.m
michaelkirk: I took a different approach, which seemed simpler, PTAL @charlesmchen

PR #3114, opened 2018-03-06
Fix debuglogs.org integration

michaelkirk:
content-type is now not only available, but required.

### DIFF
--- a/Scripts/debug_log_upload.py
+++ b/Scripts/debug_log_upload.py
@@ -28,10 +28,13 @@ def execute_command(command):
         print e.output
         sys.exit(1)
 
-        
+def add_field(curl_command, form_key, form_value):
+    curl_command.append('-F')
+    curl_command.append("%s=%s" % (form_key, form_value))
+
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Precommit cleanup script.')
-    parser.add_argument('--file', help='used for starting a new version.')
+    parser.add_argument('--file', required=True, help='used for starting a new version.')
 
     args = parser.parse_args()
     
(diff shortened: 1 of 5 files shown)


Scripts/debug_log_upload.py
```diff
@@ -41,24 +44,30 @@ def execute_command(command):
...
     for field_name in upload_fields:
-        field_value = upload_fields[field_name]
-        curl_command.append('-F')
-        curl_command.append("'%s=%s'" % (field_name, field_value, ))
+        add_field(curl_command, field_name, upload_fields[field_name])
+
+    add_field(curl_command, "content-type", "application/octet-stream")
+ 
     curl_command.append('-F')
-    curl_command.append("'file=@%s'" % (args.file,))
```
michaelkirk: This actually didn't work with my version of curl.

Scripts/debug_log_upload.py
```diff
@@ -41,24 +44,30 @@ def execute_command(command):
...
     upload_key = upload_key + os.path.splitext(args.file)[1]
-    upload_fields['key'] = upload_key
     
     download_url = 'https://debuglogs.org/' + upload_key
     print 'download_url:', download_url
     
     curl_command = ['curl', '-v', '-i', '-X', 'POST']
+
+    # key must appear before other fields
+    add_field(curl_command, 'key', upload_key)
```
michaelkirk: aws requirement, else you get some 4xx response.

SignalMessaging/utils/OWSPreferences.m
```diff
@@ -255,6 +255,7 @@ - (void)applyCallLoggingSettingsForLegacyUsersWithTransaction:(YapDatabaseReadWr
         }
     }();
 
+    DDLogInfo(@"%@ Migrating setting - System Call Enabled: %d", self.logTag, shouldLogCallsInRecents);
```
michaelkirk: unrelated, but wanted to get this in.

Signal/src/util/Pastelog.m
```diff
@@ -150,6 +151,8 @@ - (void)uploadFileWithUploadUrl:(NSString *)uploadUrl fields:(NSDictionary *)fie
                 NSString *fieldValue = fields[fieldName];
                 [formData appendPartWithFormData:[fieldValue dataUsingEncoding:NSUTF8StringEncoding] name:fieldName];
             }
+            [formData appendPartWithFormData:[OWSMimeTypeApplicationZip dataUsingEncoding:NSUTF8StringEncoding] name:@"content-type"];
```
charlesmchen: This should be `self.mimeType`, not `OWSMimeTypeApplicationZip`.

charlesmchen: or probably `weakSelf.mimeType`.

michaelkirk: Done, thanks!

PR #3119, opened 2018-03-07
Fix redundant profile downloads

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/SignalMessaging/profiles/OWSUserProfile.m
+++ b/SignalMessaging/profiles/OWSUserProfile.m
@@ -106,18 +106,17 @@ NSString *const kLocalProfileUniqueId = @"kLocalProfileUniqueId";
 {
     @synchronized(self)
     {
-        BOOL didChange;
-        if (_avatarUrlPath == nil && avatarUrlPath == nil) {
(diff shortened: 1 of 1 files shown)


SignalMessaging/profiles/OWSUserProfile.m
```diff
@@ -106,18 +106,17 @@ - (void)setAvatarUrlPath:(nullable NSString *)avatarUrlPath
 {
     @synchronized(self)
     {
-        BOOL didChange;
-        if (_avatarUrlPath == nil && avatarUrlPath == nil) {
-            didChange = NO;
-        } else if (_avatarUrlPath != nil || avatarUrlPath != nil) {
```
michaelkirk-signal: I could have also just changed this line to:
```
} else if (_avatarUrlPath == nil || avatarUrlPath == nil) {
```

But since we got it wrong... twice... I rethought this idiom.

charlesmchen: Funny, I applied a fix for this in the backup work...

PR #3133, opened 2018-03-14
Swipe through media attachments

michaelkirk-signal:
First pass at a swipe-through media view for conversations.

Future work could include (there's a card tracking these):

- title label per item with sender name, date/time
- photo rail
- overlay caption when available
- using Mapping so we don't have to keep a copy of all mediaMessages for a thread in memory
- reaping GalleryItemCache

It's intended to be visually identical to the existing full screen media view, but now you can page through multiple.

image

One thing, I intentionally omitted the dbModified machinery under the assumption that people in the gallery don't expect new media to appear as messages trickle in while viewing the gallery. I could be wrong about this, but it simplified the implementation.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -18,7 +18,6 @@
 #import "DateUtil.h"
 #import "DebugUITableViewController.h"
 #import "FingerprintViewController.h"
-#import "MediaDetailViewController.h"
 #import "NSAttributedString+OWS.h"
 #import "NewGroupViewController.h"
 #import "OWSAudioPlayer.h"
@@ -2028,8 +2027,15 @@ typedef enum : NSUInteger {
 
     [self dismissKeyBoard];
 
-    MediaDetailViewController *vc = [[MediaDetailViewController alloc] initWithAttachmentStream:attachmentStream
-                                                                                       viewItem:viewItem];
+    if (![viewItem.interaction isKindOfClass:[TSMessage class]]) {
+        OWSFail(@"Unexpected viewItem.interaction");
+        return;
+    }
+    TSMessage *mediaMessage = (TSMessage *)viewItem.interaction;
+
+    MediaPageViewController *vc =
+        [[MediaPageViewController alloc] initWithThread:self.thread mediaMessage:mediaMessage];
+
     [vc presentFromViewController:self replacingView:imageView];
 }
 
@@ -2042,8 +2048,15 @@ typedef enum : NSUInteger {
     OWSAssert(attachmentStream);
 
     [self dismissKeyBoard];
-    MediaDetailViewController *vc = [[MediaDetailViewController alloc] initWithAttachmentStream:attachmentStream
-                                                                                       viewItem:viewItem];
+
+    if (![viewItem.interaction isKindOfClass:[TSMessage class]]) {
+        OWSFail(@"Unexpected viewItem.interaction");
+        return;
+    }
+    TSMessage *mediaMessage = (TSMessage *)viewItem.interaction;
+
+    MediaPageViewController *vc =
+        [[MediaPageViewController alloc] initWithThread:self.thread mediaMessage:mediaMessage];
     [vc presentFromViewController:self replacingView:imageView];
 }
 
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2028,10 +2028,7 @@ typedef enum : NSUInteger {
 
     [self dismissKeyBoard];
 
-    UIWindow *window = [UIApplication sharedApplication].keyWindow;
-    CGRect convertedRect = [imageView convertRect:imageView.bounds toView:window];
     MediaDetailViewController *vc = [[MediaDetailViewController alloc] initWithAttachmentStream:attachmentStream
-                                                                                       fromRect:convertedRect
                                                                                        viewItem:viewItem];
     [vc presentFromViewController:self replacingView:imageView];
 }
@@ -2045,11 +2042,7 @@ typedef enum : NSUInteger {
     OWSAssert(attachmentStream);
 
     [self dismissKeyBoard];
-    UIWindow *window = [UIApplication sharedApplication].keyWindow;
-    CGRect convertedRect = [imageView convertRect:imageView.bounds toView:window];
-
     MediaDetailViewController *vc = [[MediaDetailViewController alloc] initWithAttachmentStream:attachmentStream
-                                                                                       fromRect:convertedRect
                                                                                        viewItem:viewItem];
     [vc presentFromViewController:self replacingView:imageView];
 }
--- a/Signal/src/ViewControllers/MediaPageViewController.swift
+++ b/Signal/src/ViewControllers/MediaPageViewController.swift
@@ -1,0 +1,688 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import UIKit
+
+class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSource, UIPageViewControllerDelegate, MediaDetailViewControllerDelegate {
+
+    private struct MediaGalleryItem: Equatable {
+        let message: TSMessage
+        let attachmentStream: TSAttachmentStream
+        let viewController: MediaDetailViewController
+
+        var isVideo: Bool {
+            return attachmentStream.isVideo()
+        }
+
+        var image: UIImage {
+            guard let image = attachmentStream.image() else {
+                owsFail("\(logTag) in \(#function) unexpectedly unable to build attachment image")
+                return UIImage()
+            }
+
+            return image
+        }
+    }
+
+    private var cachedItems: [MediaGalleryItem] = []
+    private var initialItem: MediaGalleryItem!
+    private var currentItem: MediaGalleryItem! {
+        return cachedItems.first { $0.viewController == viewControllers?.first }
+    }
+
+    private let includeGallery: Bool
+    private let thread: TSThread
+
+    private let mediaGalleryFinder: OWSMediaGalleryFinder
+    private let uiDatabaseConnection: YapDatabaseConnection
+
+    private var mediaMessages: [TSMessage] = []
+
+    convenience init(thread: TSThread, mediaMessage: TSMessage) {
+        self.init(thread: thread, mediaMessage: mediaMessage, includeGallery: true)
+    }
+
+    init(thread: TSThread, mediaMessage: TSMessage, includeGallery: Bool) {
+        self.thread = thread
+        self.uiDatabaseConnection = OWSPrimaryStorage.shared().newDatabaseConnection()
+        self.mediaGalleryFinder = OWSMediaGalleryFinder()
+        self.includeGallery = includeGallery
+
+        let kSpacingBetweenItems: CGFloat = 20
+
+        super.init(transitionStyle: .scroll,
+                   navigationOrientation: .horizontal,
+                   options: [UIPageViewControllerOptionInterPageSpacingKey: kSpacingBetweenItems])
+
+        self.dataSource = self
+        self.delegate = self
+
+        uiDatabaseConnection.beginLongLivedReadTransaction()
+
+        if includeGallery {
+            uiDatabaseConnection.read { transaction in
+                // TODO don't read all media messages in at once. Use Mapping?
+                self.mediaGalleryFinder.enumerateMediaMessages(with: thread, transaction: transaction) { message in
+                    self.mediaMessages.append(message)
+                }
+            }
+        } else {
+            self.mediaMessages = [mediaMessage]
+        }
+
+        guard let initialItem = self.buildGalleryItem(mediaMessage: mediaMessage, thread: thread) else {
+            owsFail("unexpetedly unable to build initial gallery item")
+            return
+        }
+        self.initialItem = initialItem
+        cachedItems.insert(initialItem, at: 0)
+
+        self.setViewControllers([initialItem.viewController], direction: .forward, animated: false, completion: nil)
+    }
+
+    @available(*, unavailable, message: "Unimplemented")
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    deinit {
+        Logger.debug("\(logTag) deinit")
+    }
+
+    var presentationView: UIImageView!
+    var footerBar: UIToolbar!
+    var videoPlayBarButton: UIBarButtonItem!
+    var videoPauseBarButton: UIBarButtonItem!
+    var pagerScrollView: UIScrollView!
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        // Navigation
+
+        self.navigationItem.leftBarButtonItem = UIBarButtonItem(barButtonSystemItem: .stop, target: self, action: #selector(didPressDismissButton))
+
+        // Even though bars are opaque, we want content to be layed out behind them.
+        // The bars might obscure part of the content, but they can easily be hidden by tapping
+        // The alternative would be that content would shift when the navbars hide.
+        self.extendedLayoutIncludesOpaqueBars = true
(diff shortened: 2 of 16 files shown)


rocknjohn:
Resolves https://github.com/signalapp/Signal-iOS/issues/759

charlesmchen:
> One thing, I intentionally omitted the dbModified machinery under the assumption that people in the gallery don't expect new media to appear as messages trickle in while viewing the gallery. I could be wrong about this, but it simplified the implementation.

Sounds good.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2028,11 +2027,15 @@ - (void)didTapImageViewItem:(ConversationViewItem *)viewItem
 
     [self dismissKeyBoard];
 
-    UIWindow *window = [UIApplication sharedApplication].keyWindow;
-    CGRect convertedRect = [imageView convertRect:imageView.bounds toView:window];
-    MediaDetailViewController *vc = [[MediaDetailViewController alloc] initWithAttachmentStream:attachmentStream
-                                                                                       fromRect:convertedRect
```
charlesmchen: Are we not doing the presentation animations anymore?

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -0,0 +1,706 @@
...
+        let attachmentStream: TSAttachmentStream
+        let viewController: MediaDetailViewController
+
+        var isVideo: Bool {
+            return attachmentStream.isVideo()
+        }
+
+        var image: UIImage {
+            guard let image = attachmentStream.image() else {
+                owsFail("\(logTag) in \(#function) unexpectedly unagble to build attachment image")
```
charlesmchen: s/unagble/

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -0,0 +1,706 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import UIKit
+
+class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSource, UIPageViewControllerDelegate, MediaDetailViewControllerDelegate {
+
+    private struct MediaGalleryItem: Equatable {
```
charlesmchen: can you please log dealloc as we do in `OWSViewController`?  It's pretty easy to introduce leaks in the views.

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -0,0 +1,706 @@
...
+        let message: TSMessage
+        let attachmentStream: TSAttachmentStream
+        let viewController: MediaDetailViewController
+
+        var isVideo: Bool {
+            return attachmentStream.isVideo()
+        }
+
+        var image: UIImage {
```
charlesmchen: I wonder if this should be optional.  Missing attachment files are going to be more common once if/when we ship backup since we'll do "lazy restores" where attachments are filled in later.

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -0,0 +1,706 @@
...
+                   options: [UIPageViewControllerOptionInterPageSpacingKey: kSpacingBetweenItems])
+
+        self.dataSource = self
+        self.delegate = self
+
+        uiDatabaseConnection.beginLongLivedReadTransaction()
+
+        if includeGallery {
+            uiDatabaseConnection.read { transaction in
+                // TODO don't read all media messages in at once. Use Mapping?
```
charlesmchen: I suspect we'll have to use a mapping before shipping this.

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -0,0 +1,706 @@
...
+
+        // Hack to avoid "page" bouncing when not in gallery view.
+        // e.g. when getting to media details via message details screen, there's only
+        // one "Page" so the bounce doesn't make sense.
+        if !self.includeGallery {
+            pagerScrollView.isScrollEnabled = false
+        }
+
+        // FIXME dynamic title with sender/date
+        self.title = "Attachment"
```
charlesmchen: l10n?

charlesmchen: Or is this definitely changing in next PR?

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -0,0 +1,706 @@
...
+        self.shouldHideToolbars = !self.shouldHideToolbars
+    }
+
+    private var shouldHideToolbars: Bool = false {
+        didSet {
+            if (oldValue == shouldHideToolbars) {
+                return
+            }
+
+            // Hiding the status bar affects the positioing of the navbar. We don't want to show that in an animation, it's
```
charlesmchen: s/positioing/

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -0,0 +1,706 @@
...
+    var originRect: CGRect?
+
+    func present(fromViewController: UIViewController, replacingView: UIView) {
+
+        self.replacingView = replacingView
+
+        let convertedRect: CGRect = replacingView.convert(replacingView.bounds, to: UIApplication.shared.keyWindow)
+        self.originRect = convertedRect
+
+        // loadView hasn't necesarily been called yet.
```
charlesmchen: s/necesarily/

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -0,0 +1,706 @@
...
+                self.presentationView.layer.cornerRadius = 0
+                self.presentationView.superview?.layoutIfNeeded()
+
+                // We must lay out once *before* we centerMediaViewConstraints
+                // because it uses the imageView.frame to build the constraints
+                // that will center the imageView, and then once again *after*
+                // to ensure that the centered constraints are applied.
+                    //TODO
+//                [self centerMediaViewConstraints];
+//                [self.mediaView.superview layoutIfNeeded];
```
charlesmchen: ?

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -0,0 +1,706 @@
...
+                    self.view.backgroundColor = UIColor.white
+                },
+                completion: { (_: Bool) in
+                    // HACK: Setting the frame to itself *seems* like it should be a no-op, but
+                    // it ensures the content is drawn at the right frame. In particular I was
+                    // reproducibly seeing some images squished (they were EXIF rotated, maybe
+                    // related). similar to this report:
+                    // https://stackoverflow.com/questions/27961884/swift-uiimageview-stretched-aspect
+                    // TODO
+//                    self.mediaView.frame = self.mediaView.frame
```
charlesmchen: should this be uncommented per the comment above?

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -0,0 +1,706 @@
...
+
+        var count: UInt = 0
+        self.uiDatabaseConnection.read { (transaction: YapDatabaseReadTransaction) in
+            count = self.mediaGalleryFinder.mediaCount(thread: self.thread, transaction: transaction)
+        }
+        return Int(count)
+    }
+
+    @nonobjc
+    public func presentationIndex(for: UIPageViewController) -> Int {
```
charlesmchen: Could we do the `presentationIndex(for page:` instead to avoid the "for" funny business?

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -0,0 +1,706 @@
...
+            owsFail("currentIndex was unexpectedly nil in \(#function)")
+            return nil
+        }
+
+        let index: Int = mediaMessages.index(before: currentIndex)
+        return mediaMessages[safe: index]
+    }
+}
+
+fileprivate extension Collection {
```
charlesmchen: Pull out into public extension?

SignalServiceKit/src/Storage/OWSMediaGalleryFinder.m
```diff
@@ -0,0 +1,142 @@
...
+-  (NSUInteger)mediaCountForThread:(TSThread *)thread transaction:(YapDatabaseReadTransaction *)transaction
+{
+    NSString *group = [self mediaGroupWithThreadId:thread.uniqueId];
+    return [[self galleryExtensionWithTransaction:transaction] numberOfItemsInGroup:group];
+}
+
+- (NSUInteger)mediaIndexForMessage:(TSMessage *)message transaction:(YapDatabaseReadTransaction *)transaction
+{
+    NSString *groupId;
+    NSUInteger index;
```
charlesmchen: What will these values be if the message if no longer exists, e.g. it "disappeared"?

SignalServiceKit/src/Storage/OWSMediaGalleryFinder.m
```diff
@@ -0,0 +1,142 @@
...
+    [storage asyncRegisterExtension:[self mediaGalleryDatabaseExtension]
+                           withName:OWSMediaGalleryFinderExtensionName];
+}
+
++ (YapDatabaseAutoView *)mediaGalleryDatabaseExtension
+{
+    YapDatabaseViewSorting *sorting = [YapDatabaseViewSorting withObjectBlock:^NSComparisonResult(YapDatabaseReadTransaction * _Nonnull transaction, NSString * _Nonnull group, NSString * _Nonnull collection1, NSString * _Nonnull key1, id  _Nonnull object1, NSString * _Nonnull collection2, NSString * _Nonnull key2, id  _Nonnull object2) {
+        
+        if (![object1 isKindOfClass:[TSMessage class]]) {
+            return NSOrderedSame;
```
charlesmchen: `OWSFail()`

SignalServiceKit/src/Storage/OWSMediaGalleryFinder.m
```diff
@@ -0,0 +1,142 @@
...
+        }
+        TSMessage *message1 = (TSMessage *)object1;
+        
+        if (![object2 isKindOfClass:[TSMessage class]]) {
+            return NSOrderedSame;
+        }
+        TSMessage *message2 = (TSMessage *)object2;
+        
+        return [@(message1.timestampForSorting) compare:@(message2.timestampForSorting)];
```
charlesmchen: I wonder if we could pull this out into a NSObject+OWS category: `[NSObject orderingForObject:(NSObject *)left andObject:(NSObject *)right expectedClass:(Class)clazz comparisonBlock:(id(^_NonNull)(id))comparisonBlock]`.

then:
```

+ (YapDatabaseAutoView *)mediaGalleryDatabaseExtension
{
    YapDatabaseViewSorting *sorting = [YapDatabaseViewSorting withObjectBlock:^NSComparisonResult(YapDatabaseReadTransaction * transaction, NSString * group, NSString * collection1, NSString * key1, id  object1, NSString * collection2, NSString * key2, id  object2) {
return [NSObject orderingForObject:message1 andObject:message2 expectedClass:[TSMessage class] comparisonBlock:^(TSMessage message) { return message. timestampForSorting; }];
}
```

charlesmchen: TIOLI obvs

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: Yup, but they are now the responsibility of the MediaPageView controller. Which also does the dismiss animation, or a different dismiss animation if we've swiped to a different image.

michaelkirk-signal: We are, but the animation responsibility has moved to the MediaPageView controller, which is also responsible for the dismiss animation. 

When we dismiss from the same media as we presented (e.g. we never swiped, or swiped and swiped back), we smoothly animate the media back to it's position in the conversation, but if we've swiped to a different media message than we started, we instead animate the media off screen.

This responsibility might shift again since I'm hoping to re-use a similar animation with the tiled media gallery.

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: fixed.

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: good call, done.

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: The point of this struct was to consolidate handling the optional, rather than sprinkling the handling of it throughout everything that touches this field. I'm hoping that's possible, as it makes reasoning elsewhere simpler.

Another approach which doesn't introduce nil would be to use a better placeholder image - e.g. in your example something that communicated "loading" would make sense to the user.

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: Yeah, I wouldn't be totally surprised, though I've done some benchmarking with a few thousand attachments on a fast device, and it's not a problem. I'll definitely do benchmarking on a bigger dataset on a slower device to see what we can get away with.

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: I'm going to get this in a followup PR.

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: fixed

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: Yes, much better, thanks!

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: fixed, thanks for the spellcheck =)

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: Fixed, thanks!

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: Removed - I can no longer reproduce the behavior that this code, which used to live in MessageDetailViewController used to fix.

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: Done.

SignalServiceKit/src/Storage/OWSMediaGalleryFinder.m
michaelkirk-signal: Keep in mind our transaction is from a long-lived connection, so this shouldn't happen, but it's good point.

I've added an assert to verify this doesn't happen.

SignalServiceKit/src/Storage/OWSMediaGalleryFinder.m
michaelkirk-signal: Done

SignalServiceKit/src/Storage/OWSMediaGalleryFinder.m
michaelkirk-signal: Though we could use `expectedClass` to dynamically check, we'd still have to cast in the comparisonBlock to avoid compiler warnings right? (unless we wrote it in Swift)

michaelkirk-signal:
> LGTM but I suspect we'll need to use a mapping - or a long-lived transaction (shared with the finder) so that disappearing message (etc.) don't break the view.

Keep in mind that we *are* using a long-lived transaction. So messages won't disappear while in the media gallery. My thought to use a mapping was if the user has more than 500 or so media messages in a thread it could get sluggish on slow devices.

PR #3143, opened 2018-03-17
Charlesmchen/incremental backup4

charlesmchen:
PTAL @michaelkirk


Signal.xcodeproj/project.pbxproj
```diff
@@ -40,6 +40,7 @@
 		340FC8C7204DE64D007AEB0F /* OWSBackupAPI.swift in Sources */ = {isa = PBXBuildFile; fileRef = 340FC8C6204DE64D007AEB0F /* OWSBackupAPI.swift */; };
 		340FC8CA20517B84007AEB0F /* OWSBackupImportJob.m in Sources */ = {isa = PBXBuildFile; fileRef = 340FC8C820517B84007AEB0F /* OWSBackupImportJob.m */; };
 		340FC8CD20518C77007AEB0F /* OWSBackupJob.m in Sources */ = {isa = PBXBuildFile; fileRef = 340FC8CC20518C76007AEB0F /* OWSBackupJob.m */; };
+		340FC8D0205BF2FA007AEB0F /* OWSBackupIO.m in Sources */ = {isa = PBXBuildFile; fileRef = 340FC8CE205BF2FA007AEB0F /* OWSBackupIO.m */; };
```
charlesmchen: I pulled out a `OWSBackupIO` class that does encryption/decryption and compression/decompression.

Signal/src/ViewControllers/AppSettings/AboutTableViewController.m
```diff
@@ -72,18 +72,28 @@ - (void)updateTableContents
 #ifdef DEBUG
     __block NSUInteger threadCount;
     __block NSUInteger messageCount;
+    __block NSUInteger attachmentCount;
```
charlesmchen: These stats weren't complete.

Signal/src/ViewControllers/DebugUI/DebugUIBackup.m
```diff
@@ -40,6 +40,10 @@ - (nullable OWSTableSection *)sectionForThread:(nullable TSThread *)thread
                                      actionBlock:^{
                                          [DebugUIBackup tryToImportBackup];
                                      }]];
+    [items addObject:[OWSTableItem itemWithTitle:@"Log Database Size Stats"
+                                     actionBlock:^{
+                                         [DebugUIBackup logDatabaseSizeStats];
```
charlesmchen: I intend to try to shrink our database size on disk/resist its future growth.  This helps.

Signal/src/util/OWSBackup.m
```diff
@@ -107,31 +106,6 @@ - (void)setup
     });
 }
 
-- (void)setBackupPrivateKey:(NSData *)value
```
charlesmchen: Until we have a real backup encryption key, I've returned this and stubbed out the value.

Signal/src/util/OWSBackupAPI.swift
```diff
@@ -225,52 +225,55 @@ import CloudKit
     // MARK: - Delete
 
     @objc
-    public class func deleteRecordFromCloud(recordName: String,
+    public class func deleteRecordsFromCloud(recordNames: [String],
```
charlesmchen: Added bulk delete for cloud cleanup because the perf benefits are huge.

charlesmchen: We might want to do something similar for saves, downloads and "check if exists".

Signal/src/util/OWSBackupExportJob.m
```diff
@@ -3,64 +3,226 @@
...
 #import <SignalServiceKit/Threading.h>
-#import <SignalServiceKit/YapDatabaseConnection+OWS.h>
-#import <YapDatabase/YapDatabase.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
-NSString *const kOWSBackup_ExportDatabaseKeySpec = @"kOWSBackup_ExportDatabaseKeySpec";
+@class OWSAttachmentExport;
+
+@interface OWSBackupExportItem : NSObject
```
charlesmchen: I've added "items" to export and import.  It considerably simplified the code and improved readability as the complexity grew.

Signal/src/util/OWSBackupExportJob.m
```diff
@@ -3,64 +3,226 @@
...
+    OWSAssert(backupIO);
+
+    self.exportItems = [NSMutableArray new];
+    self.backupIO = backupIO;
+
+    return self;
+}
+
+- (BOOL)writeObject:(TSYapDatabaseObject *)object
+         entityType:(OWSSignalServiceProtosBackupSnapshotBackupEntityType)entityType
```
charlesmchen: It isn't strictly necessary to capture the entity type (the importer doesn't use this state), but I think it'll be helpful to have around to future-proof this work, help with debugging issue, etc.

Signal/src/util/OWSBackupExportJob.m
```diff
@@ -149,35 +308,32 @@ - (void)start
...
+        }
+        [self saveToCloudWithCompletion:^(NSError *_Nullable saveError) {
+            if (saveError) {
+                [weakSelf failWithError:saveError];
                 return;
             }
-            [self saveToCloud:^(NSError *_Nullable saveError) {
-                if (saveError) {
-                    [weakSelf failWithError:saveError];
+            [self cleanUpCloudWithCompletion:^(NSError *_Nullable cleanUpError) {
```
charlesmchen: Yup, I finally made the "with completion" change.

Signal/src/util/OWSBackupExportJob.m
```diff
@@ -211,310 +369,297 @@ - (void)configureExport:(OWSBackupJobBoolCompletion)completion
...
-            [OWSBackupJob loadDatabaseKeySpecWithKeychainKey:kOWSBackup_ExportDatabaseKeySpec];
-        if (!databaseKeySpec) {
-            OWSProdLogAndFail(@"%@ Could not load database keyspec for export.", self.logTag);
-        }
-        return databaseKeySpec;
+    OWSDBExportStream *exportStream = [[OWSDBExportStream alloc] initWithBackupIO:self.backupIO];
+
+    __block BOOL aborted = NO;
+    typedef BOOL (^EntityFilter)(id object);
+    typedef NSUInteger (^ExportBlock)(YapDatabaseReadTransaction *,
```
charlesmchen: I'm never quite sure if this "private block" pattern is more readable than a class method.

Signal/src/util/OWSBackupExportJob.m
```diff
@@ -211,310 +369,297 @@ - (void)configureExport:(OWSBackupJobBoolCompletion)completion
...
         return NO;
     }
-    // TODO: Should we do a database checkpoint?
 
-    DDLogInfo(@"%@ copiedThreads: %llu", self.logTag, copiedThreads);
-    DDLogInfo(@"%@ copiedMessages: %llu", self.logTag, copiedInteractions);
-    DDLogInfo(@"%@ copiedEntities: %llu", self.logTag, copiedEntities);
-    DDLogInfo(@"%@ copiedAttachments: %llu", self.logTag, copiedAttachments);
-    DDLogInfo(@"%@ copiedMigrations: %llu", self.logTag, copiedMigrations);
+    @autoreleasepool {
```
charlesmchen: I was probably overly liberal with my usage of `@autoreleasepool` but I'm trying to future-proof this work and ensure that the (often expensive) memory usage of the import/export process is cleaned up ASAP.

Signal/src/util/OWSBackupExportJob.m
```diff
@@ -672,25 +853,34 @@ - (void)deleteRecordsFromCloud:(NSMutableArray<NSString *> *)obsoleteRecordNames
...
+    NSMutableArray<NSString *> *batchRecordNames = [NSMutableArray new];
+    while (obsoleteRecordNames.count > 0 && batchRecordNames.count < kMaxBatchSize) {
+        NSString *recordName = obsoleteRecordNames.lastObject;
+        [obsoleteRecordNames removeLastObject];
+        [batchRecordNames addObject:recordName];
+    }
 
     __weak OWSBackupExportJob *weakSelf = self;
-    [OWSBackupAPI deleteRecordFromCloudWithRecordName:recordName
+    [OWSBackupAPI deleteRecordsFromCloudWithRecordNames:batchRecordNames
```
charlesmchen: The thing about CloudKit's "batch" methods is that although they offer tantalizing perf benefits, your code might become significantly more complicated to handle "partial failures."

Signal/src/util/OWSBackupIO.h
```diff
@@ -0,0 +1,52 @@
...
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OWSBackupEncryptedItem : NSObject
+
+@property (nonatomic) NSString *filePath;
+
+@property (nonatomic) NSData *encryptionKey;
```
charlesmchen: Per security review, we're now using a separate backup key for each uploaded file in a given backup.  The manifest is the only upload currently encrypted with the "backup encryption key".

Signal/src/util/OWSBackupIO.h
```diff
@@ -0,0 +1,52 @@
...
+
+@end
+
+#pragma mark -
+
+@interface OWSBackupIO : NSObject
```
charlesmchen: This class tries to gather encryption and compression concerns.

Signal/src/util/OWSBackupIO.m
```diff
@@ -0,0 +1,253 @@
...
+        // This assumes that dst will always be smaller than src.
+        //
+        // We slightly pad the buffer size to account for the worst case.
+        size_t dstBufferLength = srcLength + 64 * 1024;
+        uint8_t *dstBuffer = malloc(sizeof(uint8_t) * dstBufferLength);
+        if (!dstBuffer) {
+            return nil;
+        }
+        // TODO: Should we use COMPRESSION_LZFSE?
+        size_t dstLength
```
charlesmchen: I'm tempted to use COMPRESSION_LZFSE, an Apple-specific compression scheme offered by compressionlib that is supposedly better than the traditional alternatives.

Signal/src/util/OWSBackupImportJob.m
```diff
@@ -182,118 +201,174 @@ - (void)downloadAndProcessManifest:(OWSBackupJobCompletion)completion
...
-    self.databaseRecordMap = [databaseRecordMap mutableCopy];
-    self.attachmentRecordMap = [attachmentRecordMap mutableCopy];
+    self.databaseItems = databaseItems;
+    self.attachmentsItems = attachmentsItems;
 
     return completion(YES);
 }
 
-- (void)downloadFilesFromCloud:(NSMutableArray<NSString *> *)recordNames completion:(OWSBackupJobCompletion)completion
+- (nullable NSArray<OWSBackupImportItem *> *)parseItems:(id)json key:(NSString *)key
```
charlesmchen: One nice thing about using "items" is that all of the (defensive) parsing logic is gathered in a single place.

SignalServiceKit/protobuf/OWSSignalServiceProtos.proto
```diff
@@ -238,3 +238,21 @@ message GroupDetails {
   optional bool   active      = 5 [default = true];
   optional uint32 expireTimer = 6;
 }
+
+message BackupSnapshot
```
charlesmchen: Since we're putting this work on pause, I've temporarily put these classes in the "signal service" proto schema, but ultimately they should be pulled out into an "signal ios" proto schema.

SignalServiceKit/src/Storage/OWSPrimaryStorage.m
```diff
@@ -326,6 +326,16 @@ - (NSString *)databaseFilePath
     return OWSPrimaryStorage.databaseFilePath;
 }
 
+- (NSString *)databaseFilePath_SHM
```
charlesmchen: I've made a bunch of changes to the storage classes to support the "backup storage" which we no longer use.  I'm going to leave them since I believe they generally improve things.

SignalServiceKit/src/Storage/OWSStorage.m
```diff
@@ -184,21 +183,15 @@ - (YapDatabaseConnection *)newConnection
 
 - (YapDatabaseConnection *)registrationConnection
 {
-    @synchronized(self)
-    {
-        if (!self.registrationConnectionCached) {
-            YapDatabaseConnection *connection = [super registrationConnection];
+    YapDatabaseConnection *connection = [super registrationConnection];
```
charlesmchen: Backup exports previously needed to block until the backup storage was completely closed.  I did a bunch of work to support this, including this bit.

basnijholt:
I've seen your recent "incremental backup" PRs, but nowhere I've seen it written what it exactly does? Is this really finally the backup feature we've all been hoping for? 😁

Signal/src/util/OWSBackupExportJob.m
```diff
@@ -211,310 +402,297 @@ - (void)configureExport:(OWSBackupJobBoolCompletion)completion
...
+    {
+        unsigned long long totalFileSize = 0;
+        for (OWSAttachmentExport *attachmentExport in self.unsavedAttachmentExports) {
+            totalFileSize += [OWSFileSystem fileSizeOfPath:attachmentExport.attachmentFilePath].unsignedLongLongValue;
+        }
+        DDLogInfo(@"%@ exporting %@: count: %zd, bytes: %llu.",
+            self.logTag,
+            @"attachment items",
+            self.unsavedAttachmentExports.count,
+            totalFileSize);
```
michaelkirk-signal: Could you either only log the attachment size here, or clarify the log to indicate this is the db+attachment size?

Signal/src/util/OWSBackupExportJob.m
```diff
@@ -523,47 +701,69 @@ - (BOOL)saveNextAttachmentFileToCloud:(OWSBackupJobCompletion)completion
...
     }
-    OWSAssert(self.manifestFilePath);
+
+    OWSBackupExportItem *exportItem = [OWSBackupExportItem new];
+    exportItem.encryptedItem = encryptedItem;
 
     __weak OWSBackupExportJob *weakSelf = self;
 
-    [OWSBackupAPI upsertManifestFileToCloudWithFileUrl:[NSURL fileURLWithPath:self.manifestFilePath]
+    [OWSBackupAPI upsertManifestFileToCloudWithFileUrl:[NSURL fileURLWithPath:encryptedItem.filePath]
```
michaelkirk-signal: just geeking out on API, but if the OWSBackupAPI upsert/upload api took an EncryptedItem instead of a filePath, we could use the compiler to help make sure were doing everything in the right order (encrypting before uploading). TIOLI of course.

Signal/src/util/OWSBackupIO.h
```diff
@@ -0,0 +1,57 @@
...
+
+#pragma mark - Compression
+
+- (nullable NSData *)compressData:(NSData *)srcData;
+
+// I'm using the (new in iOS 9) compressionlib.  One of its weaknesses is that it
+// requires you to pre-allocate output buffers during compression and decompression.
+// During decompression this is particularly tricky since there's no way to safely
+// predict how large the output will be based on the input.  So, we store the
+// uncompressed size for compressed backup items.
```
michaelkirk-signal: Good comment - did you investigate the stream implementation?

Might be a bit more work to get it going, but seems like it would offer lower memory consumption and less code to do uncompressed size accounting.

https://developer.apple.com/documentation/compression/data_compression

Signal/src/util/OWSBackupIO.m
michaelkirk-signal: Speaking as a reformed bzip2/xzip user, I would guess there are some pretty diminishing returns on esoteric compression schemes, and that there may be a future benefit to using a more universal compression approach.

So, I'd vote for using something standard unless the wins are big, but I don't feel very strongly about it.

Signal/src/util/OWSBackupImportJob.m
```diff
@@ -182,118 +201,174 @@ - (void)downloadAndProcessManifest:(OWSBackupJobCompletion)completion
...
+        item.recordName = recordName;
+        item.encryptionKey = encryptionKey;
+        item.relativeFilePath = relativeFilePath;
+        item.uncompressedDataLength = uncompressedDataLength;
+        [items addObject:item];
+    }
+    return items;
+}
+
+- (void)downloadFilesFromCloud:(NSMutableArray<OWSBackupImportItem *> *)items
```
michaelkirk-signal: TIOLI: download**Items**FromCloud?

SignalServiceKit/protobuf/OWSSignalServiceProtos.proto
michaelkirk-signal: No biggie - but if we try to pull in upstream changes on this file, its now going to be a bit more of a pain to merge them - is there a reason to not put them in their final resting place now?

Signal/src/util/OWSBackupExportJob.m
charlesmchen: If you send an emoji to someone as old as me, you're running an 80% chance of misinterpretation.  This is presumably the "water evaporating from wet hair after shower" emoji.

charlesmchen:
@basnijholt we're not shipping backup now, but we're trying to answer some questions about what would be involved down the road if/when we did it.

Signal/src/util/OWSBackupExportJob.m
```diff
@@ -3,64 +3,253 @@
...
+
+@end
+
+#pragma mark -
+
+// Used to serialize database snapshot contents.
+// Writes db entities using protobufs into snapshot fragments.
+// Snapshot fragments are compressed (they compress _very well_,
+// around 20x smaller) then encrypted.  Ordering matters in
+// snapshot contents (entities should we restored in the same
```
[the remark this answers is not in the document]
charlesmchen: giphy 4

Signal/src/util/OWSBackupExportJob.m
```diff
@@ -3,64 +3,253 @@
...
+// order they are serialized), so we are always careful to preserve
+// ordering of entities within a snapshot AND ordering of snapshot
+// fragments within a bakckup.
+//
+// This stream is used to write entities one at a time and takes
+// care of sharding them into fragments, compressing and encrypting
+// those fragments.  Fragment size is fixed to reduce worst case
```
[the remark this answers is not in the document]
charlesmchen: I've called it stream because it has a stream-like interface and because eventually it might be nice to use `NSStream`, but I consider that a nice-to-have.  Using `NSStream` would only offer slight benefits - a single file instead of multiple fragments.

Signal/src/util/OWSBackupExportJob.m
```diff
@@ -3,64 +3,253 @@
...
+    if (!self.backupSnapshotBuilder) {
+        self.backupSnapshotBuilder = [OWSSignalServiceProtosBackupSnapshotBuilder new];
+    }
+
+    OWSSignalServiceProtosBackupSnapshotBackupEntityBuilder *entityBuilder =
+        [OWSSignalServiceProtosBackupSnapshotBackupEntityBuilder new];
+    [entityBuilder setType:entityType];
+    [entityBuilder setEntityData:data];
+
+    [self.backupSnapshotBuilder addEntity:[entityBuilder build]];
```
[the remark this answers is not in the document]
charlesmchen: Yup, that would be ideal, but I've looked at the developer time/code complexity/surface area for bugs, and it doesn't make sense... yet.  There's still a ton of more valuable work towards a production-ready MVP.

Signal/src/util/OWSBackupExportJob.m
charlesmchen: Done.

Signal/src/util/OWSBackupIO.h
charlesmchen: Per above, it seems like it'd be non-trivial and not offer critical benefits.  The place streaming encryption would offer key benefits IMO is when encrypting/decrypting large attachments.

Signal/src/util/OWSBackupImportJob.m
charlesmchen: Yeah, naming was hard because you have one "file/item/record" per attachment.  Worse:

* I chose to use "recordName" for CloudKit record names, to emphasize that we were using these as CloudKit identifiers.
* Items are implementation details of the import and export process ... and export has two different kinds of item.

So, sometimes we deliberately download records _without_ downloading the file (e.g. when checking whether or not a record already existing) and other times we download record specifically to download the file.  I used this name here to emphasize that we were trying to do the latter.

SignalServiceKit/protobuf/OWSSignalServiceProtos.proto
charlesmchen: Done.

Signal/src/util/OWSBackupAPI.swift
```diff
@@ -225,52 +225,55 @@ import CloudKit
...
                                              failure: @escaping (Error) -> Void) {
 
-        let recordID = CKRecordID(recordName: recordName)
-
-        database().delete(withRecordID: recordID) {
-            (_, error) in
+        var recordIDs = [CKRecordID]()
+        for recordName in recordNames {
+            recordIDs.append(CKRecordID(recordName: recordName))
+        }
```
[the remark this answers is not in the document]
charlesmchen: Done.

Signal/src/util/OWSBackupExportJob.m
michaelkirk-signal: 🚜 🤒 🇫🇰

basnijholt:
Thanks. 

> if/when we did it

Does this mean there is no plan to do it soon?

Signal/src/util/OWSBackupIO.m
charlesmchen: 10k messages + 5k attachments.

simulator screen shot - iphone x - 2018-03-20 at 11 11 26

### With LZFSE

450kb database snapshot.

### With ZLIB

775kb database snapshot.

### With LZ4

998kb database snapshot.

### With LZMA

338kb database snapshot.

### Conclusion

LZMA has best performance AND is more widely adopted standard than LZFSE.

charlesmchen: Forgot to include one (irrelevant) test result.

### LZ4_RAW

905kb database snapshot.

PR #3144, opened 2018-03-17
Reduce Giphy page size.

charlesmchen:
PTAL @michaelkirk 

This will reduce the max number of search results in half (from 200 -> 100), but:

* search results with load faster.
* we'll be less likely to stumble across bugs that only affect queries with lots of results (as we did this week).
* I suspect most users only look at the first ~20 results anyway.


michaelkirk-signal:
LGTM

I guess infinite scroll would be an obvious future enhancement.

PR #3146, opened 2018-03-17
Charlesmchen/incremental backup5

charlesmchen:

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/DebugUI/DebugUIBackup.m
+++ b/Signal/src/ViewControllers/DebugUI/DebugUIBackup.m
@@ -44,6 +44,10 @@ NS_ASSUME_NONNULL_BEGIN
                                      actionBlock:^{
                                          [DebugUIBackup logDatabaseSizeStats];
                                      }]];
+    [items addObject:[OWSTableItem itemWithTitle:@"Clear All CloudKit Records"
+                                     actionBlock:^{
+                                         [DebugUIBackup clearAllCloudKitRecords];
+                                     }]];
 
     return [OWSTableSection sectionWithTitle:self.name items:items];
 }
@@ -148,6 +152,13 @@ NS_ASSUME_NONNULL_BEGIN
     if (attachmentCount > 0) {
         DDLogInfo(@"%@ attachment average size: %f", self.logTag, attachmentSizeTotal / (double)attachmentCount);
     }
+}
+
++ (void)clearAllCloudKitRecords
+{
+    DDLogInfo(@"%@ clearAllCloudKitRecords.", self.logTag);
+
+    [OWSBackup.sharedManager clearAllCloudKitRecords];
 }
 
 @end
--- a/Signal/src/util/OWSBackupExportJob.m
+++ b/Signal/src/util/OWSBackupExportJob.m
@@ -296,6 +296,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 // If we are replacing an existing backup, we use some of its contents for continuity.
 @property (nonatomic, nullable) NSDictionary<NSString *, OWSBackupManifestItem *> *lastManifestItemMap;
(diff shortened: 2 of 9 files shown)


Signal/src/ViewControllers/DebugUI/DebugUIBackup.m
```diff
@@ -44,6 +44,10 @@ - (nullable OWSTableSection *)sectionForThread:(nullable TSThread *)thread
                                      actionBlock:^{
                                          [DebugUIBackup logDatabaseSizeStats];
                                      }]];
+    [items addObject:[OWSTableItem itemWithTitle:@"Clear All CloudKit Records"
+                                     actionBlock:^{
+                                         [DebugUIBackup clearAllCloudKitRecords];
+                                     }]];
```
charlesmchen: App doesn't retain any local state about whether or not there is a valid backup.  Every new backup export/import consults CloudKit, the "single point of truth" for current backup state.  So clearing CloudKit clears everything.

Signal/src/util/OWSBackupExportJob.m
```diff
@@ -294,6 +294,10 @@ @interface OWSBackupExportJob ()
 
 @property (nonatomic, nullable) OWSBackupExportItem *manifestItem;
 
+// If we are replacing an existing backup, we use some of its contents for continuity.
+@property (nonatomic, nullable) NSDictionary<NSString *, OWSBackupManifestItem *> *lastManifestItemMap;
+@property (nonatomic, nullable) NSSet<NSString *> *lastRecordNames;
```
charlesmchen: In order to do incremental backup, we need to try to "re-use/recycle" backup fragments from previous backups.  To do that, we need to:

* Defensively verify that recycled fragments still exist.
* Extract and reuse the fragment's metadata from the last backup.

Signal/src/util/OWSBackupImportJob.m
```diff
@@ -20,36 +20,14 @@
 
 #pragma mark -
 
-@interface OWSBackupImportItem : NSObject
```
charlesmchen: This class got promoted to the base "job" class since now both import and export jobs need to download and parse fragments from the last backup.

charlesmchen:

PTAL @michaelkirk

Signal/src/util/OWSBackupJob.h
```diff
@@ -11,10 +11,42 @@ extern NSString *const kOWSBackup_ManifestKey_EncryptionKey;
...
 
-@class OWSBackupJob;
+@interface OWSBackupManifestItem : NSObject
+
+@property (nonatomic) NSString *recordName;
+
+@property (nonatomic) NSData *encryptionKey;
+
+// This property is only set for certain types of manifest item;
+@property (nonatomic, nullable) NSString *relativeFilePath;
```
michaelkirk-signal: Can you expound on the comment as to which types?

Signal/src/util/OWSBackupJob.h
```diff
@@ -11,10 +11,42 @@ extern NSString *const kOWSBackup_ManifestKey_EncryptionKey;
...
+
+// This property is only set if the manifest item is downloaded.
+@property (nonatomic, nullable) NSString *downloadFilePath;
+
+// This property is only set if the manifest item is compressed.
```
michaelkirk-signal: I assumed we were uniformly compressing everything. Which things aren't compressed?

Signal/src/util/OWSBackupJob.h
charlesmchen: Done.

Signal/src/util/OWSBackupJob.h
charlesmchen: We currently only compress database snapshots.  I haven't bothered with:

* Manifest files (which are tiny).
* Attachments - which might benefit from this.  It's on the card.

PR #3149, opened 2018-03-19
media gallery/all media

michaelkirk-signal:
Overview:

There are two "versions" of the Media Gallery - the swipeable "page" view (backed by a UIPageViewController) and the "tile" view (backed by a UICollectionViewController).

This PR introduces the tile view, which is accessible from the new "All Media" button in the page view, or accessible directly from the new "All Media" entry in Conversation Settings.

As part of that work, I've pulled some of the presentation logic from the GalleryPageViewController into a new object, the MediaGalleryViewController.

It's responsible for:

- presenting the initial view controller (tile from settings, or page from conversation)
- switching between the view modes (from swipe to gallery and back, or from gallery to swipe and back, ad infinitum)
- a unified data source for both the tile and page views, which lazily loads to allow for faster presentation time.

Some things I've omitted from this PR:

- more aggressive caching
- zoom animation from (Conversation -> Page -> Tile -> Page. (though there is an animation from Settings -> Tile -> Page -> Tile)
- better titles for page view and tile view  #controller
- jettisoning the MediaGalleryDataSource, currently it just fills and fills.
- showing captions in page view

**media pager/details: All Media**

screen shot 2018-03-19 at 3 16 50 pm

**conversation settings: All Media**

image

**Media Tile View**

screen shot 2018-03-19 at 2 53 39 pm

PTAL @charlesmchen


Signal/src/AppDelegate.m
```diff
@@ -1147,6 +1147,18 @@ - (void)checkIfAppIsReady
     [[OWSMessageReceiver sharedInstance] handleAnyUnprocessedEnvelopesAsync];
     [[OWSBatchMessageProcessor sharedInstance] handleAnyUnprocessedEnvelopesAsync];
 
+    if (!Environment.preferences.hasGeneratedThumbnails) {
+        [OWSPrimaryStorage.sharedManager.newDatabaseConnection
```
michaelkirk-signal: Generating thumbnails depends on the size of the media, but in my measuring is about 10 per second on an old iPhone5. We don't need to block launch of the app for this.

If the user goes straight to their gallery after first launch, it's possible they'll see missing thumbnails.

We could build a more robust approach, but it didn't seem very valuable as it's a fleeting problem that will solve itself for most people in a minute or two.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2806,6 +2807,8 @@ - (YapDatabaseConnection *)uiDatabaseConnection
     NSAssert([NSThread isMainThread], @"Must access uiDatabaseConnection on main thread!");
     if (!_uiDatabaseConnection) {
         _uiDatabaseConnection = [self.primaryStorage newDatabaseConnection];
+        // Increase object cache limit. Default is 250.
+        _uiDatabaseConnection.objectCacheLimit = 500;
```
michaelkirk-signal: From https://github.com/yapstudios/YapDatabase/wiki/Performance-Pro

We've gotten into the habit of creating new databaseConnections all the time. Creating the connection isn't expensive in and of itself now that we're using a keySpec, but when possible we should try to re-use the connection to take advantage of it's object cache.

e.g. in the case of the media gallery, it's very likely that much of the "nearby" media messages/attachments are already in cache.

So two changes in this PR:

1. drive the MediaGallery from the ConversationView's uiDatabaseConnection
2. bump up this objectCacheLimit a bit - we could likely benefit from making it even larger, but this is a start.

Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -0,0 +1,710 @@
...
+        self.message = message
+        self.attachmentStream = attachmentStream
+        self.galleryDate = GalleryDate(message: message)
+    }
+
+    var isVideo: Bool {
+        return attachmentStream.isVideo()
+    }
+
+    var thumbnailImage: UIImage {
```
michaelkirk-signal: essential for memory consumption when we're showing 20-40 media tiles. 

We should also consider using pre-downsampled media for the conversation view as well.

Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -0,0 +1,710 @@
...
+        presentDetailView(fromViewController: fromViewController, initialDetailItem: initialDetailItem, replacingView: replacingView)
+    }
+
+    public func presentDetailView(fromViewController: UIViewController, initialDetailItem: MediaGalleryItem, replacingView: UIView) {
+        // For a speedy load, we only fetch a few items on either side of
+        // the initial message
+        ensureGalleryItemsLoaded(.around, item: initialDetailItem, amount: 10)
+        self.initialDetailItem = initialDetailItem
+
+        let pageViewController = MediaPageViewController(initialItem: initialDetailItem, mediaGalleryDataSource: self, uiDatabaseConnection: self.uiDatabaseConnection, includeGallery: self.includeGallery)
```
michaelkirk-signal: This presentation logic was extracted from the MediaPageViewController

Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -0,0 +1,710 @@
...
+    func ensureGalleryItemsLoaded(_ direction: GalleryDirection, item: MediaGalleryItem, amount: UInt, completion: ((IndexSet, [IndexPath]) -> Void)? = nil ) {
+
+        var galleryItems: [MediaGalleryItem] = self.galleryItems
+        var sections: [GalleryDate: [MediaGalleryItem]] = self.sections
+        var sectionDates: [GalleryDate] = self.sectionDates
+
+        var newGalleryItems: [MediaGalleryItem] = []
+        var newDates: [GalleryDate] = []
+
+        Bench(title: "fetching gallery items") {
```
michaelkirk-signal: I was going to remove these `Bench` calls before opening the PR, but they're kind of useful to see, and not in a hot code-path.

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -290,18 +314,100 @@ - (nullable UIImage *)image
...
+    if (![[NSFileManager defaultManager] fileExistsAtPath:self.mediaURL.path]) {
+        OWSFail(@"%@ while generating thumbnail, source file doesn't exist: %@", self.logTag, self.mediaURL) return;
+    }
+
+    // TODO proper resolution?
+    CGFloat thumbnailSize = 200;
+
+    UIImage *_Nullable result;
+    if (self.isImage || self.isAnimated) {
+        CGImageSourceRef imageSource = CGImageSourceCreateWithURL((__bridge CFURLRef)self.mediaURL, NULL);
```
michaelkirk-signal: FYI it's much faster to create thumbnails from a URL ImageSource vs. converting to/from PNG and using a Data ImageSource.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen: Good point.

charlesmchen: We only need to share/re-use connections when doing something perf-intensive (and usually user-facing), right?

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -25,6 +25,13 @@
 
 NS_ASSUME_NONNULL_BEGIN
```
charlesmchen: Nothing to do with this PR, but this Debug UI page has been so useful that it has grown unwieldy.  I intend to eventually clean it up, especially by DRYing repetitive logic.

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -1108,6 +1123,149 @@ + (void)sendFakeMessages:(NSUInteger)counter
...
+                mp4DataSource:(DataSource *)mp4DataSource
+                  transaction:(YapDatabaseReadWriteTransaction *)transaction
+{
+    DDLogInfo(@"%@ sendFakeMediaMessages: %zd", self.logTag, counter);
+
+    for (NSUInteger i = 0; i < counter; i++) {
+
+        // Random time within last n years. Helpful for filling out a media gallery over time.
+        double yearsMillis = 4.0 * 365.0 * 24.0 * 60.0 * 60.0 * 1000.0;
+        uint64_t millisAgo = (uint64_t)(((double)arc4random() / ((double)0xffffffff)) * yearsMillis);
```
charlesmchen: I think you want `FLT_MAX`, not 0xff...

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -1108,6 +1123,149 @@ + (void)sendFakeMessages:(NSUInteger)counter
...
+               jpegDataSource:(DataSource *)jpegDataSource
+                mp4DataSource:(DataSource *)mp4DataSource
+                  transaction:(YapDatabaseReadWriteTransaction *)transaction
+{
+    DDLogInfo(@"%@ sendFakeMediaMessages: %zd", self.logTag, counter);
+
+    for (NSUInteger i = 0; i < counter; i++) {
+
+        // Random time within last n years. Helpful for filling out a media gallery over time.
+        double yearsMillis = 4.0 * 365.0 * 24.0 * 60.0 * 60.0 * 1000.0;
```
charlesmchen: Instead, why not `yearsMillis = 4 * kYearInMs;`.   You'l need to define `kYearInMs` next to the existing constants, e.g. `kMonthInMs`.

Signal/src/ViewControllers/MediaGalleryViewController.swift
charlesmchen: Sounds like a solid card.

Signal/src/ViewControllers/MediaGalleryViewController.swift
charlesmchen: Interesting.

Signal/src/ViewControllers/MediaTileViewController.swift
```diff
@@ -0,0 +1,560 @@
...
+            return nil
+        }
+
+        return galleryItems[mostRecentDate]?.last
+    }
+
+    var isFetchingMoreData: Bool = false
+
+    let kLoadOlderSectionIdx = 0
+    var kLoadNewerSectionIdx: Int {
```
charlesmchen: In Hungarian notation, `k` is used to indicate constants.  This isn't a constant.  

http://web.mst.edu/~cpp/common/hungarian.html

I've been trying to break myself of the habit of using hungarian notation in our codebase since it feels very un-swift-like and when Obj-C identifiers are auto-converted to Swift identifiers these prefixes tend to interfere.

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -224,6 +229,25 @@ - (nullable NSString *)filePath
...
+        return nil;
+    }
+
+    if (!self.isImage && !self.isVideo && !self.isAnimated) {
+        return nil;
+    }
+
+    NSString *filename = filePath.lastPathComponent.stringByDeletingPathExtension;
+    NSString *containingDir = filePath.stringByDeletingLastPathComponent;
+    NSString *newFilename = [filename stringByAppendingString:@"-thumbnail"];
```
charlesmchen: Hmm.  A while back (Spring/Summer 2017?) we switched from storing all attachments in the same directory to storing each within its own subdirectory.  The motive was to let attachments use their "source" filename when possible without having two attachments with the same "source" filename overwrite each other.  But legacy attachments _were not_ moved into subdirectories.

So, this scheme is safe for "new-style" attachments but not for legacy attachments.  i.e. if there were existing "a.jpg" and "a-thumbnail.jpg" attachments, the thumbnail for the first might overwrite the actual second - or the second might be used as the thumbnail for the first.  

I'm not terribly concerned about this so those attachments are really old, so long as:

* We use a more obscure suffix than "-thumbnail", e.g. "-signal-ios-thumbnail".
* We ensure that in the case of a name collision we never overwrite a real attachment.

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -290,18 +314,100 @@ - (nullable UIImage *)image
...
+
+    if (![[NSFileManager defaultManager] fileExistsAtPath:self.mediaURL.path]) {
+        OWSFail(@"%@ while generating thumbnail, source file doesn't exist: %@", self.logTag, self.mediaURL) return;
+    }
+
+    // TODO proper resolution?
+    CGFloat thumbnailSize = 200;
+
+    UIImage *_Nullable result;
+    if (self.isImage || self.isAnimated) {
```
charlesmchen: For images only, I think we need to use `NSData (Image)` category for safety.

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -290,18 +314,100 @@ - (nullable UIImage *)image
...
+        return;
+    }
+
+    if ([[NSFileManager defaultManager] fileExistsAtPath:thumbnailPath]) {
+        // already exists
+        return;
+    }
+
+    if (![[NSFileManager defaultManager] fileExistsAtPath:self.mediaURL.path]) {
+        OWSFail(@"%@ while generating thumbnail, source file doesn't exist: %@", self.logTag, self.mediaURL) return;
```
charlesmchen: Missing semicolon.

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -330,7 +436,7 @@ + (void)deleteAttachments
 - (CGSize)calculateImageSize
 {
     if ([self isVideo]) {
-        return [self videoThumbnail].size;
+        return [self videoStillImage].size;
```
charlesmchen: I think we'll need to update `removeFileWithTransaction` to also remove the thumbnail.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: I think that's right - basically anything where latency is a concern.

Signal/src/AppDelegate.m
```diff
@@ -1147,6 +1147,18 @@ - (void)checkIfAppIsReady
...
+            asyncReadWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
```
charlesmchen: This is going to be disk and CPU-intensive.  Rather than use `asyncReadWithBlock`, it might be better to explicitly dispatch to a low-priority queue to avoid interfering with the UI.

Signal/src/ViewControllers/MediaGalleryViewController.swift
michaelkirk-signal: Added!

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
michaelkirk-signal: done!

Signal/src/ViewControllers/MediaTileViewController.swift
michaelkirk-signal: Whoops, fixed.

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
michaelkirk-signal: Great point!

> We use a more obscure suffix than "-thumbnail", e.g. "-signal-ios-thumbnail".

Done!

> We ensure that in the case of a name collision we never overwrite a real attachment.

Already handled:

```
if ([[NSFileManager defaultManager] fileExistsAtPath:thumbnailPath]) {
```

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
michaelkirk-signal: Done, thanks!

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
michaelkirk-signal: Done, thanks!

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
michaelkirk-signal: Good catch, done!

PR #3150, opened 2018-03-19
Clear PageView cache when receiving memory warning

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MediaPageViewController.swift
+++ b/Signal/src/ViewControllers/MediaPageViewController.swift
@@ -4,32 +4,30 @@
 
 import UIKit
 
-public struct MediaGalleryPage: Equatable {
-
-    public let viewController: MediaDetailViewController
-    public let galleryItem: MediaGalleryItem
-
-    public var message: TSMessage {
-        return galleryItem.message
+// Objc wrapper for the MediaGalleryItem struct
+@objc
+public class GalleryItemBox: NSObject {
+    public let value: MediaGalleryItem
+    
+    init(_ value: MediaGalleryItem) {
+        self.value = value
     }
-
+    
     public var attachmentStream: TSAttachmentStream {
-        return galleryItem.attachmentStream
+        return value.attachmentStream
     }
+}
 
-    public var isVideo: Bool {
-        return galleryItem.isVideo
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -4,59 +4,55 @@
...
+    fileprivate var galleryItem: MediaGalleryItem {
+        return self.galleryItemBox.value
     }
 }
 
 class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSource, UIPageViewControllerDelegate, MediaDetailViewControllerDelegate {
 
     private weak var mediaGalleryDataSource: MediaGalleryDataSource?
 
-    private var cachedPages: [MediaGalleryPage] = []
```
michaelkirk-signal: Since a MediaGalleryPage has a strong reference to a MediaDetailViewController, this cache was getting large fast. Since we were treating it as a sort of DataSource as well as a cache, it wasn't convenient to implement caching with the existing layout.

Moving everything around left the MediaGalleryPage struct not very useful, so I nixed it.

charlesmchen: Can you please ensure that all of the new VCs related to media views log their deallocation (like `OWSViewController`) and verify that none are leaking?

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -175,8 +170,22 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
         view.addGestureRecognizer(verticalSwipe)
     }
 
+    override func didReceiveMemoryWarning() {
+        Logger.info("\(logTag) in \(#function)")
+        super.didReceiveMemoryWarning()
+
+        self.cachedPages = [:]
```
charlesmchen: What does `[:]` do?

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: Yup, they all include:

```
deinit {
        Logger.debug("\(logTag) deinit")
   }
```

PR #3153, opened 2018-03-20
Avoid occasional audio crash

michaelkirk-signal:
Occasionally when doing things like starting a video and then immediately navigating away, we would fail to terminate our audio session.

My first change was to synchronize on self when messing with audioActivities.

Which is a good thing to do, considering we don't always manipulate the audio session on the main thread. Unfortunately, that was insufficient to avoid the crash.

Though it's not elegant, introducing a small delay before deactivation seems to do the trick.

PTAL @charlesmchen

### DIFF
--- a/SignalMessaging/environment/OWSAudioSession.swift
+++ b/SignalMessaging/environment/OWSAudioSession.swift
@@ -19,7 +19,7 @@ public class AudioActivity: NSObject {
     }
 
     deinit {
-        OWSAudioSession.shared.ensureAudioSessionActivationState()
+        OWSAudioSession.shared.ensureAudioSessionActivationStateAfterDelay()
     }
 }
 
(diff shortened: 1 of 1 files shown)


SignalMessaging/environment/OWSAudioSession.swift
```diff
@@ -38,8 +38,10 @@ public class OWSAudioSession: NSObject {
     public func startAmbientAudioActivity(_ audioActivity: AudioActivity) {
         Logger.debug("\(logTag) in \(#function)")
 
-        startAudioActivity(audioActivity)
+        objc_sync_enter(self)
```
charlesmchen: I wonder if we should introduce a `SyncLock`, e.g. 
```
let _ = SyncLock(self)
```
`SyncLock` would be RIAA - take a lock when initialized and release the lock when deallocated.  With let, the lock should be retained throughout the enclosing block.  This is fine too, TIOLI

charlesmchen:
LGTM but we gotta wonder if this just makes the issue harder to repro - which is an improvement.  Could we instead always ensure we're accessing the audio session on the main thread using something like `DispatchMainThreadSafe()`?

michaelkirk-signal:
>  Could we instead always ensure we're accessing the audio session on the main thread using something like DispatchMainThreadSafe()?

That's definitely a valid sentiment. I'm sort of just kicking the can down the road. 

Some reasons not to:
- we may already be accessing it off the main thread, so this could be a non-trivial change.
- Some of it's methods block and can be a bit slow (not multiple seconds, but 100's of ms IMO), switching to main thread only precludes us from handling that nicely

PR #3157, opened 2018-03-20
hide all media from settings gallery

michaelkirk-signal:
PTAL @charlesmchen


Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -106,9 +104,12 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
 
         // Navigation
 
-        self.navigationItem.leftBarButtonItem = UIBarButtonItem(barButtonSystemItem: .stop, target: self, action: #selector(didPressDismissButton))
+        // Note: using a custom leftBarButtonItem breaks the interactive pop gesture, but we don't want to be able
```
charlesmchen: What a shame this was necessary.

PR #3158, opened 2018-03-20
Don't generate thumbnail until attachment has been saved

michaelkirk-signal:
Otherwise the file hasn't been written to disk yet.

PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
+++ b/SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
@@ -49,7 +49,6 @@ NS_ASSUME_NONNULL_BEGIN
     _creationTimestamp = [NSDate new];
 
     [self ensureFilePath];
-    [self ensureThumbnail];
 
     return self;
 }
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -98,6 +96,12 @@ - (nullable instancetype)initWithCoder:(NSCoder *)coder
     return self;
 }
 
+- (void)saveWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
+{
+    [super saveWithTransaction:transaction];
+    [self ensureThumbnail];
```
charlesmchen: But:

* The "convert all at startup" logic only reads and doesn't save.
* For legacy attachments not yet fixed by "convert all at startup", we want to ensure a thumbnail before displaying them.

There's more work to do here, right?

michaelkirk-signal: initWithCoder still calls ensureThumbnail, just not initWithContentType nor initWithPointer, so it will be called for existing attachments.

charlesmchen: Sorry for the misreading.

PR #3164, opened 2018-03-20
Handle empty media gallery

michaelkirk-signal:
PTAL @charlesmchen


Signal/src/ViewControllers/MediaTileViewController.swift
```diff
@@ -518,19 +549,20 @@ fileprivate protocol MediaGalleryCellDelegate: class {
...
 
     override init(frame: CGRect) {
         super.init(frame: frame)
 
-        // TODO add spinnner, start/stop animating on will/end display
         addSubview(label)
 
-        label.autoCenterInSuperview()
+        label.textAlignment = .center
+        label.numberOfLines = 0
```
charlesmchen: Whenever I allow wrap, I tend to explicitly specify the lineBreakMode, e.g. word.

PR #3167, opened 2018-03-22
Fix Tile -> Detail -> Swipe -> Dismiss -> Detail flow.

michaelkirk-signal:
We weren't doing a good job of cleaning up state upon dismissing the DetailView when presented *from* the tile view.

Basically the repro was:

Go to Conversation Settings -> All Media
Tap on a tile, swipe to a different piece of media, dismiss

That dismiss would be fine.

However, when tapping on a *subsequent* tile. That presentation (and eventual dismissal) would not show the presentationView.

And another similar problem, if you'd dismissed the detail view while the navbar was hidden, the next presentation of the detail view would be missing the navbar.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MediaGalleryViewController.swift
+++ b/Signal/src/ViewControllers/MediaGalleryViewController.swift
@@ -283,6 +283,9 @@ class MediaGalleryViewController: UINavigationController, MediaGalleryDataSource
         self.presentationView.image = initialDetailItem.fullSizedImage
         self.applyInitialMediaViewConstraints()
 
+        // Restore presentationView.alpha in case a previous dismiss left us in a bad state.
+        self.presentationView.alpha = 1
+
         // We want to animate the tapped media from it's position in the previous VC
         // to it's resting place in the center of this view controller.
         //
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -418,11 +423,11 @@ class MediaGalleryViewController: UINavigationController, MediaGalleryDataSource
         }
     }
 
-    public func dismissSelf(animated isAnimated: Bool, completion: (() -> Void)? = nil) {
+    public func dismissDetailView(_ pageViewController: MediaPageViewController, animated isAnimated: Bool, completion: (() -> Void)?) {
         self.view.isUserInteractionEnabled = false
         UIApplication.shared.isStatusBarHidden = false
 
-        guard let detailView = pageViewController?.view else {
+        guard let detailView = pageViewController.view else {
```
charlesmchen: So previously we were accessing a local, optional property named `pageViewController` and now we're accessing a non-optional, method parameter of the same name?  This is fine here, but when one identifier "shadows" another, its easy to write bugs.  I wonder if we could avoid that somehow, but don't have a proposal.

michaelkirk-signal: You're right, I shouldn't have a method param that mirrors the name of a local property.

I think the only time we should have multiple symbols with the same name in a single scope is when unwrapping; e.g.

```
guard let foo = self.foo else {
   // ...
}
```

Since in that case there's no worry of referencing the wrong thing, and at worst you could get momentarily confused about the nullability of the thing, but the compiler ultimately ensures you get it right.

In this case, even though it should be fine, since the method param *happens*  to refer to the same instance as the local property, in general it's bad practice. Fixed!

PR #3170, opened 2018-03-22
Refine 'Screen Lock'.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/translations/en.lproj/Localizable.strings
+++ b/Signal/translations/en.lproj/Localizable.strings
@@ -1538,7 +1538,7 @@
 "SCAN_QR_CODE_VIEW_TITLE" = "Scan QR Code";
 
 /* Indicates a delay of zero seconds, and that 'screen lock activity' will timeout immediately. */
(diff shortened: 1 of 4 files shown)


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1538,7 +1538,7 @@
 "SCAN_QR_CODE_VIEW_TITLE" = "Scan QR Code";
 
 /* Indicates a delay of zero seconds, and that 'screen lock activity' will timeout immediately. */
-"SCREEN_LOCK_ACTIVITY_TIMEOUT_NONE" = "None";
+"SCREEN_LOCK_ACTIVITY_TIMEOUT_NONE" = "Instant";
```
michaelkirk-signal: TIOLI: "Immediately" ?

charlesmchen: I think Josh said in standup that he thought Instant was best.

PR #3172, opened 2018-03-22
delete from slider view

michaelkirk-signal:
PTAL @charlesmchen


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -774,7 +774,7 @@
 "FINISH_GROUP_CREATION_LABEL" = "Finish creating group";
 
 /* Label indicating media gallery is empty */
-"GALLERY_TILES_EMPTY_GALLERY" = "You haven't sent or received any media in this conversation yet.";
+"GALLERY_TILES_EMPTY_GALLERY" = "You don't have any media in this conversation.";
```
charlesmchen: There isn't any media in this conversation?

PR #3175, opened 2018-03-23
Batch Delete

michaelkirk-signal:
PTAL @charlesmchen 

**Initial State**

screen shot 2018-03-23 at 3 18 49 pm

**After Tapping Select**

screen shot 2018-03-23 at 3 18 51 pm

** Item Selection **

screen shot 2018-03-23 at 4 07 51 pm

screen shot 2018-03-23 at 4 08 09 pm

screen shot 2018-03-23 at 4 07 55 pm


Signal/src/ViewControllers/MediaTileViewController.swift
```diff
@@ -88,10 +88,30 @@ public class MediaTileViewController: UICollectionViewController, MediaGalleryCe
 
         collectionView.delegate = self
 
-        // TODO iPhoneX
```
michaelkirk-signal: screen shot 2018-03-23 at 4 11 24 pm

Signal/src/ViewControllers/MediaTileViewController.swift
```diff
@@ -343,11 +417,173 @@ public class MediaTileViewController: UICollectionViewController, MediaGalleryCe
...
+            owsFail("\(logTag) in \(#function) collectionView was unexpectedly nil")
+            return
+        }
+
+        guard let indexPaths = collectionView.indexPathsForSelectedItems else {
+            owsFail("\(logTag) in \(#function) indexPaths was unexpectedly nil")
+            return
+        }
+
+        let items: [MediaGalleryItem] = indexPaths.flatMap { return self.galleryItem(at: $0) }
```
charlesmchen: I thought we found that `flatMap` wasn't valid on XCode 9.2?

michaelkirk-signal: You're thinking of `compactMap` which replaces `flatMap` in XCode 9.3.

rocknjohn:
@michaelkirk-signal Confusing text on the confirmation button. "Messages" will mean "text message" to most users.

May I suggest:
Delete
Delete 2
(User already knows what they are selecting.)

Or:
Delete Item
Delete 2 Items

Personal preference, but _not_ grouping by month would be visually cleaner. I don't think grouping by month adds value in this view. Without the headings, more items would be visible for faster selection too.

michaelkirk-signal:
> Confusing text on the confirmation button. "Messages" will mean "text message" to most users.

@rocknjohn - agreed it's awkward, but tagging on "Messages" was intentional. 

We want to clarify that you're deleting the media *and* any corresponding caption. Signal iOS is especially vulnerable to this since captions look like independent messages (for now - we're working on this). 

I guess we could make it less confusing but even longer with "Delete 2 Media Messages"

rocknjohn:
> We want to clarify that you're deleting the media and any corresponding caption. Signal iOS is especially vulnerable to this since captions look like independent messages (for now - we're working on this).

Ah The image contained in a caption bubble would be more clear, as far as the thread view is concerned. I don't think the word "Message" clarifies this behavior on this end of the UX though, maybe you could consider removing this text later after the improved bubbles exist.

PR #3178, opened 2018-03-24
Fix edge cases around pincode-only unlock.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/util/OWSScreenLockUI.m
+++ b/Signal/src/util/OWSScreenLockUI.m
@@ -19,6 +19,10 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic) BOOL hasUnlockedScreenLock;
 @property (nonatomic) BOOL isShowingScreenLockUI;
 
+@property (nonatomic, nullable) NSTimer *screenLockUITimer;
+@property (nonatomic, nullable) NSDate *lastUnlockAttemptDate;
+@property (nonatomic, nullable) NSDate *lastUnlockSuccessDate;
+
 @end
 
 #pragma mark -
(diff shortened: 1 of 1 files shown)


Signal/src/util/OWSScreenLockUI.m
```diff
@@ -19,6 +19,10 @@ @interface OWSScreenLockUI ()
 @property (nonatomic) BOOL hasUnlockedScreenLock;
 @property (nonatomic) BOOL isShowingScreenLockUI;
 
+@property (nonatomic, nullable) NSTimer *screenLockUITimer;
```
charlesmchen: We use a timer to ensure that there's only one "delayed show unlock UI" in flight at a time.

Signal/src/util/OWSScreenLockUI.m
```diff
@@ -132,6 +136,110 @@ - (void)ensureScreenProtection
...
+        shouldHaveScreenProtection,
+        shouldHaveScreenLock,
+        shouldShowBlockWindow);
+    self.screenBlockingWindow.hidden = !shouldShowBlockWindow;
+
+    [self.screenLockUITimer invalidate];
+    self.screenLockUITimer = nil;
+
+    if (shouldHaveScreenLock) {
+        // In pincode-only mode (e.g. device pincode is set
```
charlesmchen: The key comment.  This solution isn't ideal but I don't see a better alternative.

* This will only affect "pincode-only" users which should be in a minority.
* The solution is that users need to realize that they need to double-tap the home button.

michaelkirk: So currently we can't dismiss the locked app because...

1. The "Authenticate to access Signal" alert is in the foreground
2. Hitting the home button dismisses the alert
3. Upon canceling the alert, we immediately re-present the "Authenticate to access Signal" alert.

By delaying the re-presentation of the "Authenticate to access Signal" alert, you're giving the user a moment to re-tap the home button.

That makes sense.

Another approach, rather than automatically re-present the "Authenticate to access Signal", instead we'd have an unlock icon or something in the middle of the screen.

So:

1. Bring Signal to the foreground, the screen is obscured by the blue signal screen with an unlock button. (though you don't have time to push the unlock button)
2. "Unlock" alert is automatically presented once
3. Dismiss Alert by hitting cancel or home button
4. Reveals the blocking screen with an unlock button / icon

5a. Tapping that button re-represents the "Unlock" alert.

-or- 

5b. Tapping home brings you to the home screen.

WDYT?

I'm fine with the current PR for now, but maybe @myles-signal can weigh in.

charlesmchen: @michaelkirk Interesting. Let's revisit on Monday with @myles-signal.  I'm merging this as it stands for next build.

PR #3189, opened 2018-03-26
Fix edge case in screen lock UI.

charlesmchen-signal:
Fixes another edge case.

PTAL @michaelkirk


Signal/src/util/OWSScreenLockUI.m
```diff
@@ -14,7 +14,7 @@ @interface OWSScreenLockUI ()
 // "will enter background."
 @property (nonatomic) BOOL appIsInactive;
 @property (nonatomic) BOOL appIsInBackground;
-@property (nonatomic, nullable) NSDate *appBecameInactiveDate;
+@property (nonatomic, nullable) NSDate *appEnteredBackgroundDate;
```
charlesmchen-signal: Re-auth the user if app enters background, but _do not_ auth the user if app just becomes inactive.

PR #3192, opened 2018-03-26
Don't scroll to bottom on load, since we scroll to focused when view appears.

michaelkirk-signal:
PTAL @charlesmchen


michaelkirk-signal:
superseded by https://github.com/signalapp/Signal-iOS/pull/3194

PR #3193, opened 2018-03-26
Mkirk/end select after delete

michaelkirk-signal:
Just a little UX polish - end select mode after user hits "Delete" in batch select mode.

PTAL @charlesmchen


michaelkirk-signal:
superseded by https://github.com/signalapp/Signal-iOS/pull/3194

PR #3194, opened 2018-03-26
Mkirk/fix crash after delete

michaelkirk-signal:
Rebased #3192 and #3193 into this PR, but the individual commits are semantic if you want to review piecewise.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MediaPageViewController.swift
+++ b/Signal/src/ViewControllers/MediaPageViewController.swift
@@ -18,7 +18,7 @@ public class GalleryItemBox: NSObject {
     }
 }
 
-fileprivate class Box<A> {
+private class Box<A> {
     var value: A
     init(_ val: A) {
         self.value = val
@@ -31,7 +31,7 @@ fileprivate extension MediaDetailViewController {
     }
 }
 
-class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSource, UIPageViewControllerDelegate, MediaDetailViewControllerDelegate {
+class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSource, UIPageViewControllerDelegate, MediaDetailViewControllerDelegate, MediaGalleryDataSourceDelegate {
 
     private weak var mediaGalleryDataSource: MediaGalleryDataSource?
 
@@ -171,7 +171,7 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
         self.view.addSubview(footerBar)
         footerBar.autoPinWidthToSuperview()
         footerBar.autoPin(toBottomLayoutGuideOf: self, withInset: 0)
-        footerBar.autoSetDimension(.height, toSize:kFooterHeight)
+        footerBar.autoSetDimension(.height, toSize: kFooterHeight)
 
         // Gestures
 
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -317,25 +317,51 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
...
         }
         actionSheet.addAction(OWSAlerts.cancelAction)
         actionSheet.addAction(deleteAction)
 
         self.present(actionSheet, animated: true)
     }
 
+    // MARK: MediaGalleryDataSourceDelegate
+
+    func mediaGalleryDataSource(_ mediaGalleryDataSource: MediaGalleryDataSource, willDelete items: [MediaGalleryItem], initiatedBy: MediaGalleryDataSourceDelegate) {
```
michaelkirk-signal: This is the crash fix - we need to remove pages referring to deleted items. This was already happening if the media was deleted in the page view, but not when deleted from the TileView.

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -317,25 +317,51 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
...
+
+        guard items.contains(currentItem) else {
+            Logger.debug("\(self.logTag) in \(#function) irrelevant item")
+            return
+        }
+
+        // If we setCurrentItem with (animated: true) while this VC is in the background, then
+        // the next/previous cache isn't expired, and we're able to swipe back to the just-deleted vc.
+        // So to get the correct behavior, we should only animate these transitions when this
+        // vc is in the foreground
```
michaelkirk-signal: This confounded me for a good while, and is the reason for the somewhat awkward `initiatedBy` param.

Signal/src/ViewControllers/MediaTileViewController.swift
```diff
@@ -536,7 +539,8 @@ public class MediaTileViewController: UICollectionViewController, MediaGalleryDa
         }()
 
         let deleteAction = UIAlertAction(title: confirmationTitle, style: .destructive) { _ in
-            mediaGalleryDataSource.delete(items: items)
+            mediaGalleryDataSource.delete(items: items, initiatedBy: self)
+            self.endSelectMode()
```
michaelkirk-signal: UX Polish: end select mode after hitting delete. This just feels more expected.

Signal/src/ViewControllers/MediaTileViewController.swift
```diff
@@ -111,9 +111,6 @@ public class MediaTileViewController: UICollectionViewController, MediaGalleryDa
         self.footerBarBottomConstraint = footerBar.autoPinEdge(toSuperviewEdge: .bottom, withInset: -kFooterBarHeight)
 
         updateSelectButton()
-
-        self.view.layoutIfNeeded()
-        scrollToBottom(animated: false)
```
michaelkirk-signal: Perf boost: Scrolling to bottom triggers the "autoFetchMore" logic, which is unnecessary because we already ensure we've fetched the items around our "focusedItem" and scroll to that item when viewDidAppear.

PR #3198, opened 2018-03-27
Always remove metadata

michaelkirk-signal:
PTAL @charlesmchen


michaelkirk-signal:
> I hope the test plan includes some interesting images, such as images with non-default orientation.

I've elaborated the test plan around this.

PR #3199, opened 2018-03-27
Improve handling of unexpected failures in local authentication.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/util/OWSScreenLock.swift
+++ b/Signal/src/util/OWSScreenLock.swift
@@ -11,6 +11,7 @@ import LocalAuthentication
         case success
         case cancel
         case failure(error:String)
+        case unexpectedFailure(error:String)
     }
 
     @objc public let screenLockTimeoutDefault = 15 * kMinuteInterval
@@ -122,6 +123,8 @@ import LocalAuthentication
                                         switch outcome {
                                         case .failure(let error):
                                             completion(self.authenticationError(errorDescription: error))
+                                        case .unexpectedFailure(let error):
+                                            completion(self.authenticationError(errorDescription: error))
                                         case .success:
                                             self.setIsScreenLockEnabled(value: true)
                                             completion(nil)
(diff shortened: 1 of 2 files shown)


Signal/src/util/OWSScreenLock.swift
```diff
@@ -122,6 +123,8 @@ import LocalAuthentication
                                         switch outcome {
                                         case .failure(let error):
                                             completion(self.authenticationError(errorDescription: error))
+                                        case .unexpectedFailure(let error):
+                                            completion(self.authenticationError(errorDescription: error))
```
charlesmchen-signal: In non-critical contexts, we treat these failures like any other.

Signal/src/util/OWSScreenLock.swift
```diff
@@ -171,6 +177,8 @@ import LocalAuthentication
                                         switch outcome {
                                         case .failure(let error):
                                             failure(self.authenticationError(errorDescription: error))
+                                        case .unexpectedFailure(let error):
+                                            unexpectedFailure(self.authenticationError(errorDescription: error))
```
charlesmchen-signal: But when trying to unlock the screen lock, we handle them differently.

Signal/src/util/OWSScreenLockUI.m
```diff
@@ -145,6 +145,9 @@ - (void)ensureScreenProtection
         shouldHaveScreenProtection,
         shouldHaveScreenLock,
         shouldShowBlockWindow);
+    if (self.screenBlockingWindow.hidden != shouldShowBlockWindow) {
```
michaelkirk-signal: Presumably we want to log whenever it changes.

So since we assign the negated value, isn't this condition inverted?

charlesmchen-signal: Fixed.

PR #3200, opened 2018-03-27
iCharlesmchen/screen lock7

charlesmchen-signal:
PTAL @michaelkirk 

img_0092
img_0091
img_0090

### DIFF
--- a/Signal/src/util/OWSScreenLock.swift
+++ b/Signal/src/util/OWSScreenLock.swift
@@ -316,7 +316,8 @@ import LocalAuthentication
         let context = LAContext()
 
         // If user has set any non-zero timeout, recycle biometric auth
-        // in the same period as our normal screen lock timeout.
+        // in the same period as our normal screen lock timeout, up to
+        // max of 10 seconds.
         context.touchIDAuthenticationAllowableReuseDuration = TimeInterval(min(10.0, screenLockTimeout()))
 
         if #available(iOS 11.0, *) {
--- a/Signal/src/util/OWSScreenLock.swift
+++ b/Signal/src/util/OWSScreenLock.swift
@@ -11,7 +11,6 @@ import LocalAuthentication
         case success
         case cancel
         case failure(error:String)
-        case unexpectedFailure(error:String)
     }
 
     @objc public let screenLockTimeoutDefault = 15 * kMinuteInterval
@@ -123,8 +122,6 @@ import LocalAuthentication
                                         switch outcome {
                                         case .failure(let error):
                                             completion(self.authenticationError(errorDescription: error))
-                                        case .unexpectedFailure(let error):
-                                            completion(self.authenticationError(errorDescription: error))
                                         case .success:
                                             self.setIsScreenLockEnabled(value: true)
                                             completion(nil)
@@ -147,8 +144,6 @@ import LocalAuthentication
                                         switch outcome {
                                         case .failure(let error):
                                             completion(self.authenticationError(errorDescription: error))
-                                        case .unexpectedFailure(let error):
-                                            completion(self.authenticationError(errorDescription: error))
                                         case .success:
                                             self.setIsScreenLockEnabled(value: false)
                                             completion(nil)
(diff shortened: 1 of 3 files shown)


Signal/src/util/OWSScreenLock.swift
```diff
@@ -11,7 +11,6 @@ import LocalAuthentication
         case success
         case cancel
         case failure(error:String)
-        case unexpectedFailure(error:String)
```
charlesmchen-signal: Not need anymore.

Signal/src/util/OWSScreenLock.swift
```diff
@@ -323,7 +315,10 @@ import LocalAuthentication
     private func screenLockContext() -> LAContext {
         let context = LAContext()
 
-        context.touchIDAuthenticationAllowableReuseDuration = TimeInterval(screenLockTimeout())
+        // If user has set any non-zero timeout, recycle biometric auth
+        // in the same period as our normal screen lock timeout.
+        context.touchIDAuthenticationAllowableReuseDuration = TimeInterval(min(10.0, screenLockTimeout()))
```
charlesmchen-signal: Note.

Signal/src/util/OWSScreenLockUI.m
```diff
@@ -332,9 +427,23 @@ - (void)registrationStateDidChange
     [self ensureScreenProtection];
 }
 
+- (void)clearAuthUIWhenActive
+{
+    if (self.appIsInactive) {
```
charlesmchen-signal: For continuity, continue to present blocking screen in "screen lock" mode while dismissing the "local auth UI".

leonardoporpora:
is possible not to use fingerprint but a password? i would like to have password as i don't trust giving my fingerprint as i can't change it

Signal/src/util/OWSScreenLock.swift
```diff
@@ -323,7 +315,10 @@ import LocalAuthentication
     private func screenLockContext() -> LAContext {
         let context = LAContext()
 
-        context.touchIDAuthenticationAllowableReuseDuration = TimeInterval(screenLockTimeout())
+        // If user has set any non-zero timeout, recycle biometric auth
+        // in the same period as our normal screen lock timeout.
```
michaelkirk-signal: Comment is misleading since it's min(10, x)

Signal/src/util/OWSScreenLock.swift
michaelkirk-signal: 👍

Signal/src/util/OWSScreenLockUI.m
```diff
@@ -307,13 +312,102 @@ - (void)prepareScreenProtectionWithRootWindow:(UIWindow *)rootWindow
...
+        self.logTag,
+        shouldShowBlockWindow,
+        shouldHaveScreenLock,
+        shouldShowUnlockButton);
+
+    if (shouldHaveScreenLock) {
+        const CGFloat kVMargin = 50.f;
+        [rootView addSubview:imageView];
+        [imageView autoHCenterInSuperview];
+        [imageView autoPinTopToSuperviewWithMargin:kVMargin];
```
michaelkirk-signal: So when the screen transitions from the "screen protection" to "lock screen" the logo is moved up higher to make way for some new UI elements.

Some UI feedback on this from the perspective of my iPhone5 (no touch id):

The logo is too high up, it should be closer to the button:
iphone5

The rest of this is TIOLI:

When initially launching the locked app, the "Unlock" button isn't visible and I don't see an alert (no touch id on this device) so it's inexplicable to the user why the logo is not vertically aligned.

Moving the logo top to the top is currently jarring, and it's something that people with this feature enabled will see all the time. Can we animate it?

Also when canceling authentication, I see the system unlock screen without a button for what feels a bit too long. Ideally the button would already be visible when I hit Cancel.

Maybe a complete solution would be:
- Keep the logo vertically centered
- Show the unlock button below center, and render it initially (don't wait for user to cancel, even though the alert will soon be rendered over top of it, it gives the user a hint as to what's going on). That way, if Auth is canceled, unlock button is already there, no need to wait for it to render in
- If and only if an alert is presented, animate up the logo an appropriate amount.

Signal/src/util/OWSScreenLock.swift
charlesmchen-signal: Done.

charlesmchen-signal:
PTAL @michaelkirk

PR #3202, opened 2018-03-28
Start screen lock countdown if app is inactive for more than N seconds.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/util/OWSScreenLockUI.m
+++ b/Signal/src/util/OWSScreenLockUI.m
@@ -36,6 +36,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, nullable) NSDate *lastUnlockAttemptDate;
 @property (nonatomic, nullable) NSDate *lastUnlockSuccessDate;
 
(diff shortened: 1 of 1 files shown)


Signal/src/util/OWSScreenLockUI.m
```diff
@@ -116,19 +118,35 @@ - (void)setupWithRootWindow:(UIWindow *)rootWindow
...
+    if (didChange) {
+        // If app is inactive for more than N seconds,
+        // treat this as "entering the background" for the purposes
+        // of Screen Lock.
+        if (!appIsInactive) {
+            [self.inactiveTimer invalidate];
+            self.inactiveTimer = nil;
+        } else if (!self.isShowingScreenLockUI) {
+            [self.inactiveTimer invalidate];
+            self.inactiveTimer = [NSTimer weakScheduledTimerWithTimeInterval:45.f
```
michaelkirk-signal: Is this to handle the case when the app is in the app switcher for a long time?

charlesmchen-signal: Yes.

PR #3204, opened 2018-03-29
Crash/UI fix fetching edge case

michaelkirk-signal:
This could result in either not fetching the remaining data or a crash
depending on the exact shape of your data.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MediaGalleryViewController.swift
+++ b/Signal/src/ViewControllers/MediaGalleryViewController.swift
@@ -649,8 +649,13 @@ class MediaGalleryViewController: UINavigationController, MediaGalleryDataSource
                 }
 
                 let unfetchedSet = requestSet.subtracting(self.fetchedIndexSet)
-                guard unfetchedSet.count > (requestSet.count / 2) else {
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -649,8 +649,13 @@ class MediaGalleryViewController: UINavigationController, MediaGalleryDataSource
                 }
 
                 let unfetchedSet = requestSet.subtracting(self.fetchedIndexSet)
-                guard unfetchedSet.count > (requestSet.count / 2) else {
-                    // For perf we only want to fetch a relatively full batch, unless the requestSet is very small.
+
+                // For perf we only want to fetch a substantially full batch...
+                let isSubstantialRequest = unfetchedSet.count > (requestSet.count / 2)
+                // ...but we want to make sure always fulfill even small requests if we're getting just the tail end of a gallery.
+                let isFetchingEdgeOfGallery = (self.fetchedIndexSet.count - unfetchedSet.count) < requestSet.count
```
michaelkirk-signal: `isFetchingEdgeOfGallery` is the only logic change, but I pulled them both out for clarity.

PR #3205, opened 2018-03-29
Respect system alert volume for notifications while in app

michaelkirk-signal:
Fixes https://github.com/signalapp/Signal-iOS/issues/3145

PTAL @charlesmchen

### DIFF
--- a/SignalMessaging/environment/OWSSounds.m
+++ b/SignalMessaging/environment/OWSSounds.m
@@ -4,6 +4,7 @@
 
 #import "OWSSounds.h"
 #import "OWSAudioPlayer.h"
+#import <SignalMessaging/SignalMessaging-Swift.h>
 #import <SignalServiceKit/OWSFileSystem.h>
 #import <SignalServiceKit/OWSPrimaryStorage.h>
 #import <SignalServiceKit/TSThread.h>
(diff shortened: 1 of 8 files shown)


SignalMessaging/environment/OWSSounds.m
```diff
@@ -16,6 +16,7 @@
 @interface OWSSounds ()
 
 @property (nonatomic, readonly) YapDatabaseConnection *dbConnection;
+@property (nonatomic, readonly) NSMutableDictionary<NSString *, NSNumber *> *cachedSoundIDs;
```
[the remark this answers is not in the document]
michaelkirk-signal: Done! PTAL @charlesmchen

SignalMessaging/environment/OWSSounds.m
```diff
@@ -207,6 +212,34 @@ + (nullable NSURL *)soundURLForSound:(OWSSound)sound quiet:(BOOL)quiet
...
+
+    NSURL *soundURL = [self.class soundURLForSound:sound quiet:quiet];
+
+    DDLogVerbose(@"%@ creating system sound for %@", self.logTag, soundURL.lastPathComponent);
+    SystemSoundID newSoundID;
+    OSStatus status = AudioServicesCreateSystemSoundID((__bridge CFURLRef _Nonnull)(soundURL), &newSoundID);
+    OWSAssert(status == 0);
+    OWSAssert(newSoundID);
+
+    [self.cachedSoundIDs setWithKey:cacheKey value:@(newSoundID)];
```
charlesmchen-signal: Much closer, but I think we're going to need to wrap the `SystemSoundID` in a thin class that calls `AudioServicesDisposeSystemSoundID` when it is deallocated.

michaelkirk-signal: Oh thanks, that was dumb. 

As you've recommended, created a private helper class to manage the resource. PTAL @charlesmchen

PR #3206, opened 2018-03-29
Charlesmchen/bubble collapse

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Storage/OWSPrimaryStorage.m
+++ b/SignalServiceKit/src/Storage/OWSPrimaryStorage.m
@@ -132,26 +132,42 @@ void runAsyncRegistrationsForStorage(OWSStorage *storage)
 {
     OWSAssert(completion);
 
+    [((OWSDatabase *)self.database)collectRegistrationConnections];
+
     runAsyncRegistrationsForStorage(self);
 
     DDLogVerbose(@"%@ async registrations enqueued.", self.logTag);
 
     // Block until all async registrations are complete.
     //
-    // NOTE: This has to happen on the "registration connection" for this
+    // NOTE: This has to happen on the "registration connections" for this
     //       database.
-    YapDatabaseConnection *dbConnection = self.registrationConnection;
-    OWSAssert(self.registrationConnection);
-    [dbConnection flushTransactionsWithCompletionQueue:dispatch_get_main_queue()
-                                       completionBlock:^{
-                                           OWSAssert(!self.areAsyncRegistrationsComplete);
+    NSMutableSet<YapDatabaseConnection *> *pendingRegistrationConnectionSet =
+        [[((OWSDatabase *)self.database)clearCollectedRegistrationConnections] mutableCopy];
+    DDLogVerbose(@"%@ flushing registration connections: %zd.", self.logTag, pendingRegistrationConnectionSet.count);
 
-                                           DDLogVerbose(@"%@ async registrations complete.", self.logTag);
+    dispatch_async(dispatch_get_main_queue(), ^{
+        for (YapDatabaseConnection *dbConnection in pendingRegistrationConnectionSet) {
+            [dbConnection
+                flushTransactionsWithCompletionQueue:dispatch_get_main_queue()
+                                     completionBlock:^{
+                                         OWSAssertIsOnMainThread();
+                                         OWSAssert(!self.areAsyncRegistrationsComplete);
 
-                                           self.areAsyncRegistrationsComplete = YES;
+                                         [pendingRegistrationConnectionSet removeObject:dbConnection];
+                                         if (pendingRegistrationConnectionSet.count > 0) {
+                                             DDLogVerbose(@"%@ registration connection flushed.", self.logTag);
+                                             return;
+                                         }
 
-                                           completion();
-                                       }];
+                                         DDLogVerbose(@"%@ async registrations complete.", self.logTag);
+
+                                         self.areAsyncRegistrationsComplete = YES;
+
+                                         completion();
+                                     }];
+        }
+    });
 }
 
 + (void)protectFiles
--- a/SignalServiceKit/src/Storage/OWSStorage.h
+++ b/SignalServiceKit/src/Storage/OWSStorage.h
@@ -2,7 +2,7 @@
 //  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
-#import <YapDatabase/YapDatabaseConnection.h>
+#import <YapDatabase/YapDatabase.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -29,6 +29,26 @@ extern NSString *const StorageIsReadyNotification;
 - (instancetype)init NS_UNAVAILABLE;
 - (instancetype)initWithDatabase:(YapDatabase *)database
                         delegate:(id<OWSDatabaseConnectionDelegate>)delegate NS_DESIGNATED_INITIALIZER;
+
+@end
+
+#pragma mark -
+
+@interface OWSDatabase : YapDatabase
+
+- (instancetype)init NS_UNAVAILABLE;
+
+- (id)initWithPath:(NSString *)inPath
+        serializer:(nullable YapDatabaseSerializer)inSerializer
+      deserializer:(YapDatabaseDeserializer)inDeserializer
+           options:(YapDatabaseOptions *)inOptions
+          delegate:(id<OWSDatabaseConnectionDelegate>)delegate NS_DESIGNATED_INITIALIZER;
+
+// Starts collecting references to the registration connections.
+- (void)collectRegistrationConnections;
+// Stops collecting references to the registration connections and returns
+// all collected connections.
+- (NSSet<YapDatabaseConnection *> *)clearCollectedRegistrationConnections;
 
 @end
 
--- a/SignalServiceKit/src/Storage/OWSStorage.m
+++ b/SignalServiceKit/src/Storage/OWSStorage.m
@@ -130,17 +130,11 @@ typedef NSData *_Nullable (^CreateDatabaseMetadataBlock)(void);
 
 #pragma mark -
 
-@interface OWSDatabase : YapDatabase
+@interface OWSDatabase ()
 
 @property (atomic, weak) id<OWSDatabaseConnectionDelegate> delegate;
 
-- (instancetype)init NS_UNAVAILABLE;
-
-- (id)initWithPath:(NSString *)inPath
-        serializer:(nullable YapDatabaseSerializer)inSerializer
-      deserializer:(YapDatabaseDeserializer)inDeserializer
-           options:(YapDatabaseOptions *)inOptions
-          delegate:(id<OWSDatabaseConnectionDelegate>)delegate NS_DESIGNATED_INITIALIZER;
+@property (nonatomic, readonly, nullable) NSMutableSet<YapDatabaseConnection *> *registrationConnectionSet;
 
 @end
 
@@ -191,7 +185,27 @@ typedef NSData *_Nullable (^CreateDatabaseMetadataBlock)(void);
     ((OWSDatabaseConnection *)connection).canWriteBeforeStorageReady = YES;
 #endif
 
+    [self.registrationConnectionSet addObject:connection];
+
     return connection;
+}
+
+- (void)collectRegistrationConnections
+{
+    OWSAssert(!self.registrationConnectionSet);
+
+    _registrationConnectionSet = [NSMutableSet set];
+}
+
+- (NSSet<YapDatabaseConnection *> *)clearCollectedRegistrationConnections
+{
+    OWSAssert(self.registrationConnectionSet);
+
+    NSSet<YapDatabaseConnection *> *registrationConnectionSetCopy = [self.registrationConnectionSet copy];
+
+    _registrationConnectionSet = nil;
+
+    return registrationConnectionSetCopy;
 }
 
 @end


Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
```diff
@@ -0,0 +1,190 @@
...
+                            isRTL:(BOOL)isRTL
+{
+    UIBezierPath *bezierPath = [UIBezierPath new];
+
+    CGFloat bubbleLeft = 0.f;
+    CGFloat bubbleRight = size.width - kBubbleThornSideInset;
+    CGFloat bubbleTop = 0.f;
+    CGFloat bubbleBottom = size.height - kBubbleThornVInset;
+
+    [bezierPath moveToPoint:CGPointMake(bubbleLeft + kBubbleHRounding, bubbleTop)];
```
charlesmchen-signal: This is the path - I hope the constants make it more readable.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -222,71 +95,54 @@ - (void)commontInit
...
-        [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(handleTextTapGesture:)];
-    [self.textBubbleImageView addGestureRecognizer:textTap];
+    self.contentView.userInteractionEnabled = YES;
 
-    UILongPressGestureRecognizer *mediaLongPress =
-        [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(handleMediaLongPressGesture:)];
-    [self.mediaMaskingView addGestureRecognizer:mediaLongPress];
+    UITapGestureRecognizer *tap =
+        [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(handleTapGesture:)];
+    [self.contentView addGestureRecognizer:tap];
```
charlesmchen-signal: I consolidated these gestures and changed which view they hang on, since after quoted replies we'll soon need to distinguish between taps/long presses in many regions of the message.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -222,71 +95,54 @@ - (void)commontInit
...
+- (OWSMessageTextView *)newTextView
+{
+    OWSMessageTextView *textView = [OWSMessageTextView new];
+    textView.backgroundColor = [UIColor clearColor];
+    textView.opaque = NO;
+    textView.editable = NO;
+    textView.selectable = YES;
+    textView.textContainerInset = UIEdgeInsetsZero;
+    textView.contentInset = UIEdgeInsetsZero;
+    textView.textContainer.lineFragmentPadding = 0;
```
charlesmchen-signal: Here's an important fix that we may want to apply elsewhere: the text view has 3 separate properties that lead to a visual margin between the text and the view's bounds.  We need to zero them all out.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -297,7 +153,7 @@ - (UIFont *)textMessageFont
     OWSAssert(DisplayableText.kMaxJumbomojiCount == 5);
 
     CGFloat basePointSize = [UIFont ows_dynamicTypeBodyFont].pointSize;
-    switch (self.displayableText.jumbomojiCount) {
+    switch (self.displayableBodyText.jumbomojiCount) {
```
charlesmchen-signal: As we move towards quoted replies, I've been trying to distinguish the "body" message/media from "quoted" message/media

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -413,83 +315,187 @@ - (void)loadForDisplay
...
+    } else {
+        // Media-only messages should have no background color; they will fill the bubble's bounds
+        // and we don't want artifacts at the edges.
+        self.bubbleView.bubbleColor = nil;
     }
 
-    self.textBubbleImageView.image = bubbleImageData.messageBubbleImage;
-
     [self updateDateHeader];
     [self updateFooter];
```
charlesmchen-signal: Now that we're going to be vertically stacking up 5-6 views within the bubble, we "create if necessary" each possible using other methods, then arrange them (and space them out) appropriately within this method.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -413,83 +315,187 @@ - (void)loadForDisplay
...
+        // e.g. a very tall portrait image + long caption.  The media
+        // view will have "max width", so the image will be cropped to
+        // roughly a square.
+        // TODO: Myles is considering alternatives.
+        [self.viewConstraints addObjectsFromArray:@[
+            [bodyMediaView autoPinLeadingToSuperviewWithMargin:0],
+            [bodyMediaView autoPinTrailingToSuperviewWithMargin:0],
+        ]];
+        [NSLayoutConstraint
+            autoSetPriority:UILayoutPriorityDefaultLow
```
charlesmchen-signal: We need constraints to control the vertical sizing of media and text views, but we use lower priority so that when a message only contains media it uses the exact bounds of the message view.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -510,83 +516,33 @@ - (nullable id)tryToLoadCellMedia:(nullable id (^)(void))loadCellMediaBlock
...
+        [self showAttachmentErrorView:mediaView];
     }
     return cellMedia;
 }
 
-// We want to lazy-load expensive view contents and eagerly unload if the
-// cell is no longer visible.
-- (void)ensureViewMediaState
+// * If cell is visible, lazy-load (expensive) view contents.
+// * If cell is not visible, eagerly unload view contents.
```
charlesmchen-signal: This method got really simple - we now use `loadCellContentBlock` and `unloadCellContentBlock`, so that we can gather concerns and define how to load media views in the same place that we create the media views.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -762,196 +712,275 @@ - (UIFont *)dateHeaderTimeFont
...
+    textView.dataDetectorTypes = (UIDataDetectorTypeLink | UIDataDetectorTypeAddress | UIDataDetectorTypeCalendarEvent);
+    textView.shouldIgnoreEvents = shouldIgnoreEvents;
+}
 
-    if (self.viewItem.interaction.interactionType == OWSInteractionType_OutgoingMessage) {
-        // Ignore taps on links in outgoing messages that haven't been sent yet, as
-        // this interferes with "tap to retry".
-        TSOutgoingMessage *outgoingMessage = (TSOutgoingMessage *)self.viewItem.interaction;
-        self.textView.shouldIgnoreEvents = outgoingMessage.messageState != TSOutgoingMessageStateSentToService;
+- (BOOL)hasTapForMore
```
charlesmchen-signal: This class is pretty complicated, so I'm trying to use this kind of utility method to make things more readable and to DRY.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -762,196 +712,275 @@ - (UIFont *)dateHeaderTimeFont
...
+    __weak OWSMessageCell *weakSelf = self;
+    self.loadCellContentBlock = ^{
+        OWSMessageCell *strongSelf = weakSelf;
+        if (!strongSelf) {
+            return;
+        }
+        if (stillImageView.image) {
+            return;
+        }
+        // Don't cache large still images.
```
charlesmchen-signal: See this comment. I've made a card with follow-up work.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -1043,13 +1068,11 @@ - (CGSize)mediaBubbleSizeForContentWidth:(int)contentWidth
...
 
             // TODO: Adjust this behavior.
-            // TODO: This behavior is a bit different than the old behavior defined
-            //       in JSQMediaItem+OWS.h.  Let's discuss.
 
             CGFloat contentAspectRatio = self.mediaSize.width / self.mediaSize.height;
             // Clamp the aspect ratio so that very thin/wide content is presented
             // in a reasonable way.
-            const CGFloat minAspectRatio = 0.25f;
+            const CGFloat minAspectRatio = 0.35f;
```
charlesmchen-signal: Need to consult with Myles about this.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageTextView.h
```diff
@@ -0,0 +1,13 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OWSMessageTextView : UITextView
```
charlesmchen-signal: I pulled this out into a separate source file.  It didn't change much.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -970,63 +999,59 @@ - (void)addAttachmentUploadViewIfNecessary:(UIView *)attachmentView
 
     if (self.isOutgoing) {
         if (!self.attachmentStream.isUploaded) {
-            self.attachmentUploadView = [[AttachmentUploadView alloc] initWithAttachment:self.attachmentStream
-                                                                               superview:attachmentView
-                                                                 attachmentStateCallback:attachmentStateCallback];
+            __unused AttachmentUploadView *attachmentUploadView =
+                //            self.attachmentUploadView =
+                // This view will be added to attachmentView which will retain a strong reference to it.
+                [[AttachmentUploadView alloc] initWithAttachment:self.attachmentStream
```
michaelkirk-signal: TIOLI: Delete this comment explaining unconventional behavior and replace the method call with self explanatory code like: `[AttachmentUploadView addToView:attachment:attachmentStateCallback:]`

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -970,63 +999,59 @@ - (void)addAttachmentUploadViewIfNecessary:(UIView *)attachmentView
...
+                                                       superview:attachmentView
+                                         attachmentStateCallback:attachmentStateCallback];
         }
     }
 }
 
-- (void)cropMediaViewToBubbbleShape:(UIView *)view
+- (void)showAttachmentErrorView:(UIView *)mediaView
```
michaelkirk-signal: The `mediaView` isn't the error view right? So a better name could be: `showAttachmentErrorViewWithMediaView:(UIView *)mediaView`

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -970,63 +999,59 @@ - (void)addAttachmentUploadViewIfNecessary:(UIView *)attachmentView
...
-    OWSAssert(!self.customView);
+    OWSAssert(mediaView);
 
     // TODO: We could do a better job of indicating that the media could not be loaded.
-    self.customView = [UIView new];
-    self.customView.backgroundColor = [UIColor colorWithWhite:0.85f alpha:1.f];
-    self.customView.userInteractionEnabled = NO;
-    [self.payloadView addSubview:self.customView];
-    [self.contentConstraints addObjectsFromArray:[self.customView autoPinToSuperviewEdges]];
+    UIView *customView = [UIView new];
```
michaelkirk-signal: s/customView/errorView

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -1093,13 +1116,15 @@ - (CGSize)cellSizeForViewWidth:(int)viewWidth contentWidth:(int)contentWidth
...
+        cellSize.height += self.tapForMoreHeight;
+    }
 
     if (self.shouldHaveFailedSendBadge) {
         cellSize.width += self.failedSendBadgeSize;
     }
 
-    cellSize.width = ceil(cellSize.width);
-    cellSize.height = ceil(cellSize.height);
+    cellSize = CGSizeCeil(cellSize);
```
michaelkirk-signal: Nice, didn't know about this method.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -1168,50 +1208,41 @@ - (void)prepareForReuse
...
-    self.footerConstraints = nil;
+    [NSLayoutConstraint deactivateConstraints:self.viewConstraints];
+    self.viewConstraints = [NSMutableArray new];
 
     self.dateHeaderLabel.text = nil;
     self.dateHeaderLabel.hidden = YES;
-    self.textView.text = nil;
-    self.textView.hidden = YES;
-    self.textView.dataDetectorTypes = UIDataDetectorTypeNone;
+    [self.bodyTextViewCached removeFromSuperview];
```
michaelkirk-signal: What are you trying to indicate with the term `cached` in this variable? There is no `uncached` bodyTextView is there? 

Is it that we're only creating the textView during init, and not nilling/rebuilding every re-use cycle for perf?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -1168,50 +1208,41 @@ - (void)prepareForReuse
...
+    self.bubbleView.hidden = YES;
+    self.bubbleView.bubbleColor = nil;
+    self.bubbleView.bubbleStrokeView = nil;
+
+    for (UIView *subview in self.bubbleView.subviews) {
+        [subview removeFromSuperview];
+    }
+
+    if (self.unloadCellContentBlock) {
+        self.unloadCellContentBlock();
```
michaelkirk-signal: I don't think this is dangerous or anything, but if we're removing all the subviews, this should be redundant right?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -1243,6 +1274,30 @@ - (void)setIsCellVisible:(BOOL)isCellVisible {
...
+        return;
+    }
+
+    if (self.lastBodyMediaView) {
+        // Treat this as a "body media" gesture if:
+        //
+        // * There is a "body media" view.
+        // * The gesture occured within or above the "body media" view.
+        CGPoint location = [sender locationInView:self.lastBodyMediaView];
+        if (location.y <= self.lastBodyMediaView.height) {
```
michaelkirk-signal: Not a big deal, but this strikes me as more error prone than adding separate explicit gestures to the views, and I think the problem will become worse as we add more views, e.g. if we change margins and don't properly account for it here.

SignalMessaging/utils/OWSMessagesBubbleImageFactory.swift
```diff
@@ -58,6 +57,32 @@ public class OWSMessagesBubbleImageFactory: NSObject {
         }
     }
 
+    public static let bubbleColorIncoming = UIColor.jsq_messageBubbleLightGray()!
+
+    public static let bubbleColorOutgoingUnsent = UIColor.gray
```
michaelkirk-signal: We should dedupe and use these constants in the above JSQMessagesBubbleImage builders, or delete those other methods if they're no longer being used.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: Done - there's already a card.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -762,196 +712,275 @@ - (UIFont *)dateHeaderTimeFont
...
-    self.textView.linkTextAttributes = @{
-        NSForegroundColorAttributeName : self.textColor,
+    textView.font = font;
+    textView.linkTextAttributes = @{
+        NSForegroundColorAttributeName : textColor,
         NSUnderlineStyleAttributeName : @(NSUnderlineStyleSingle | NSUnderlinePatternSolid)
     };
-    self.textView.dataDetectorTypes
-        = (UIDataDetectorTypeLink | UIDataDetectorTypeAddress | UIDataDetectorTypeCalendarEvent);
+    textView.dataDetectorTypes = (UIDataDetectorTypeLink | UIDataDetectorTypeAddress | UIDataDetectorTypeCalendarEvent);
```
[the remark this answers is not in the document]

Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
```diff
@@ -0,0 +1,190 @@
...
+    _isTruncated = isTruncated;
+
+    if (didChange || !self.shapeLayer) {
+        [self updateLayers];
+    }
+}
+
+- (void)setFrame:(CGRect)frame
+{
+    BOOL didChange = !CGSizeEqualToSize(self.frame.size, frame.size);
```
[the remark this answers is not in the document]
charlesmchen-signal: Good point.  Added comments here; slightly tweaked the stroke view.

Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
```diff
@@ -0,0 +1,190 @@
...
+        [self updateLayers];
+    }
+
+    // We need to inform the "bubble stroke view" (if any) any time our frame changes.
+    [self.bubbleStrokeView updateLayers];
+}
+
+- (void)setBounds:(CGRect)bounds
+{
+    BOOL didChange = !CGSizeEqualToSize(self.bounds.size, bounds.size);
```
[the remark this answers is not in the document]
charlesmchen-signal: Ditto.

Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
```diff
@@ -0,0 +1,190 @@
...
+        [self updateLayers];
+    }
+
+    // We need to inform the "bubble stroke view" (if any) any time our frame changes.
+    [self.bubbleStrokeView updateLayers];
+}
+
+- (void)setCenter:(CGPoint)center
+{
+    [super setCenter:center];
```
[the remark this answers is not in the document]
charlesmchen-signal: Nope, added a comment explaining why.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -762,196 +712,275 @@ - (UIFont *)dateHeaderTimeFont
...
 
-- (void)addCaptionIfNecessary
-{
-    if (self.hasText) {
-        [self loadForTextDisplay];
-    } else {
-        [self.contentConstraints addObject:[self.textBubbleImageView autoSetDimension:ALDimensionHeight toSize:0]];
-    }
+    __weak OWSMessageCell *weakSelf = self;
+    self.loadCellContentBlock = ^{
```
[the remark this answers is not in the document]
charlesmchen-signal: Yup, it's frustrating. You really want to DRY where you can, but there's no good way to return a tuple in obj-c and you'd still want asserts (as you suggest) in the load/unload blocks which can't be DRY'd.  I've added the asserts you suggested.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: I went one better and pulled out the logic to "add" the attachment upload view from its constructor.  That's vestigial.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: It's custom - see above.  Usually on obj-c projects I end up with a header full of these kinds of custom macros.  They make CG code way more readable.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: You're right - the name is confusing.  Removed `Cached`.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: Yup, it should be - but let's be defensive.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: I guess its subjective, but...

* It's not as simple as "media gesture" v. "text gesture."  We'll also have gestures that originate in the "status indicator", the "failed to send indicator", the "tap for more view", the margins/whitespace/bubble edges of the cell.  
* IMO we want to be "permissive about what we accept" in terms of gestures". e.g. it would be a bug if taps in the whitespace of the bubble bounds didn't register.
* With quoted replies, we'll soon have more (3 more?) subviews in these cells.

Therefore, I think we'll need to explicitly dispatch these gestures using rules like these.

If we try to resolve these complexities by using more gesture recognizers, we'll leave ourselves vulnerable to difficult bugs: 

* GRs attached to views which overlap.
* Regions of the bubble view not covered by any subview with a GR.
* Visible bounds of contents (e.g. body text) not corresponding exactly to the hot area for that content.  Note: We don't want to try to fix that by ensuring that subviews exactly fill bubble bounds without any overlap; that will make a difficult layout problem very unwieldy.

SignalMessaging/utils/OWSMessagesBubbleImageFactory.swift
charlesmchen-signal: Done. Once we've worked on message detail view, I hope to rename and refactor `OWSMessagesBubbleImageFactory`.

PR #3209, opened 2018-03-30
Don't obscure "Unlock" button with keyboard

michaelkirk-signal:
I tried to leave thorough comments in the code, as this is a bit arcane.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/util/OWSScreenLock.swift
+++ b/Signal/src/util/OWSScreenLock.swift
@@ -37,6 +37,9 @@ import LocalAuthentication
     // Passcode-code only authentication process deactivates the app.
     private var ignoreUnlockUntilActive = false
 
+    // We temporarily resign any first responder while the Screen Lock is presented.
+    var firstResponderBeforeLockscreen: UIResponder?
+
     // MARK - Singleton class
 
     @objc(sharedManager)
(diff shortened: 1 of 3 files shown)


Signal/src/util/OWSScreenLock.swift
```diff
@@ -37,6 +37,9 @@ import LocalAuthentication
     // Passcode-code only authentication process deactivates the app.
     private var ignoreUnlockUntilActive = false
 
+    // We temporarily resign any first responder while the Screen Lock is presented.
+    var firstResponderBeforeLockscreen: UIResponder?
```
charlesmchen-signal: Should this be `weak`?

Signal/src/util/UI Categories/UIResponder+OWS.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+// Based on https://stackoverflow.com/questions/1823317/get-the-current-first-responder-without-using-a-private-api/11768282#11768282
+extension UIResponder {
+    private weak static var firstResponder: UIResponder?
```
charlesmchen-signal: The symbol `firstResponder` seems like it could collide with a symbol on UIResponder.  I wonder if we should rename this to `owsFirstResponder` or something?

michaelkirk-signal: Swift won't let us collide without specifying `override`.

Signal/src/util/OWSScreenLock.swift
michaelkirk-signal: Yeah it should, good catch!

PR #3214, opened 2018-04-02
Fix flush of registration connections.

charlesmchen-signal:
PTAL @michaelkirk


SignalServiceKit/src/Storage/OWSPrimaryStorage.m
```diff
@@ -131,26 +131,42 @@ - (void)runAsyncRegistrationsWithCompletion:(void (^_Nonnull)(void))completion
...
     runAsyncRegistrationsForStorage(self);
 
     DDLogVerbose(@"%@ async registrations enqueued.", self.logTag);
 
     // Block until all async registrations are complete.
     //
-    // NOTE: This has to happen on the "registration connection" for this
+    // NOTE: This has to happen on the "registration connections" for this
     //       database.
-    YapDatabaseConnection *dbConnection = self.registrationConnection;
```
charlesmchen-signal: This code incorrectly assumed that there would only be a single registration connection, since YapDatabase supposedly caches and reuses db connections for 5 seconds (and we start registration of all of our async db views within 5 seconds).  Not sure why, but this assumption is incorrect.  

We need to collect and flush _all_ registration connections.

michaelkirk-signal:
LGMT

PR #3215, opened 2018-04-02
Only resume lazy attachment restore in debug.

charlesmchen-signal:
PTAL @michaelkirk


michaelkirk-signal:
Should I make a branch with this commit for 2.23.1 as well?

charlesmchen-signal:
> Should I make a branch with this commit for 2.23.1 as well?

Nope, these changes aren't relevant to v2.23.1.

PR #3216, opened 2018-04-02
Fix assert in bubble views.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleStrokeView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleStrokeView.m
@@ -31,6 +31,8 @@ NS_ASSUME_NONNULL_BEGIN
     self.shapeLayer = [CAShapeLayer new];
     [self.layer addSublayer:self.shapeLayer];
 
+    [self updateLayers];
+
     return self;
 }
 
@@ -79,7 +81,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)updateLayers
 {
-    OWSAssert(self.shapeLayer);
+    if (!self.shapeLayer) {
+        return;
+    }
 
     // Don't fill the shape layer; we just want a stroke around the border.
     self.shapeLayer.fillColor = [UIColor clearColor].CGColor;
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleStrokeView.m
```diff
@@ -79,7 +81,9 @@ - (void)setCenter:(CGPoint)center
 
 - (void)updateLayers
 {
-    OWSAssert(self.shapeLayer);
+    if (!self.shapeLayer) {
```
michaelkirk-signal: As I read it, `shapeLayer` is set in `init` and never unset.

Why do we have nil checks for self.shapeLayer sprinkled throughout this file?

Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
```diff
@@ -132,8 +134,12 @@ - (void)setBubbleColor:(nullable UIColor *)bubbleColor
 
 - (void)updateLayers
 {
-    OWSAssert(self.maskLayer);
-    OWSAssert(self.shapeLayer);
+    if (!self.maskLayer) {
```
michaelkirk-signal: Same question - shapeLayer and maskLayer are set in `init` and never unset.

Why do we have the nil checks here and elsewhere?

michaelkirk-signal:
Ohhhh. I get it.

Because we're doing the Objc thing where we don't set our own properties until *after* we call `[super init]`, and `[super init]` is calling some of our settings which breaks our invariants.

This couldn't happen in swift, but LGTM

Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleStrokeView.m
charlesmchen-signal: Because the frame/bounds/center can be changed in the super constructor before `init` completes.

charlesmchen-signal:
> Ohhhh. I get it.

Yeah, that's it.

PR #3228, opened 2018-04-03
Modify attachment emojis.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/Attachments/TSAttachment.m
+++ b/SignalServiceKit/src/Messages/Attachments/TSAttachment.m
@@ -181,9 +181,9 @@ NSUInteger const TSAttachmentSchemaVersion = 4;
     if ([MIMETypeUtil isImage:contentType]) {
         return @"📷";
     } else if ([MIMETypeUtil isVideo:contentType]) {
-        return @"📽";
+        return @"🎥";
     } else if ([MIMETypeUtil isAudio:contentType]) {
-        return @"📻";
+        return @"🎧";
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Messages/Attachments/TSAttachment.m
```diff
@@ -181,9 +181,9 @@ + (NSString *)emojiForMimeType:(NSString *)contentType
     if ([MIMETypeUtil isImage:contentType]) {
         return @"📷";
     } else if ([MIMETypeUtil isVideo:contentType]) {
-        return @"📽";
+        return @"🎥";
     } else if ([MIMETypeUtil isAudio:contentType]) {
-        return @"📻";
+        return @"🎧";
```
charlesmchen-signal: I think this is more legible as "audio".

SignalServiceKit/src/Messages/Attachments/TSAttachment.m
```diff
@@ -181,9 +181,9 @@ + (NSString *)emojiForMimeType:(NSString *)contentType
     if ([MIMETypeUtil isImage:contentType]) {
         return @"📷";
     } else if ([MIMETypeUtil isVideo:contentType]) {
-        return @"📽";
+        return @"🎥";
```
charlesmchen-signal: IMO this too will be more legible.

michaelkirk-signal:
Provided these look all right on iOS9 and iOS10 as well, LGTM

PR #3230, opened 2018-04-03
Charlesmchen/fixup interactions

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -3568,15 +3568,7 @@ typedef enum : NSUInteger {
 
         groupThread.groupModel = newGroupModel;
         [groupThread saveWithTransaction:transaction];
-        message = [[TSOutgoingMessage alloc] initOutgoingMessageWithTimestamp:[NSDate ows_millisecondTimeStamp]
-                                                                     inThread:groupThread
-                                                                  messageBody:nil
-                                                                attachmentIds:[NSMutableArray new]
-                                                             expiresInSeconds:0
-                                                              expireStartedAt:0
-                                                               isVoiceMessage:NO
-                                                             groupMetaMessage:TSGroupMessageUpdate
-                                                                quotedMessage:nil];
+        message = [TSOutgoingMessage outgoingMessageInThread:groupThread groupMetaMessage:TSGroupMessageUpdate];
         [message updateWithCustomMessage:updateGroupInfo transaction:transaction];
     }];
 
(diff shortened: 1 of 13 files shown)


charlesmchen-signal:
See: https://github.com/signalapp/YapDatabase/commit/764e949142ba1bada99aeedeeaccfb68047a6e79

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3541,15 +3541,7 @@ - (void)updateGroupModelTo:(TSGroupModel *)newGroupModel successCompletion:(void
 
         groupThread.groupModel = newGroupModel;
         [groupThread saveWithTransaction:transaction];
-        message = [[TSOutgoingMessage alloc] initOutgoingMessageWithTimestamp:[NSDate ows_millisecondTimeStamp]
```
charlesmchen-signal: I started working through the process of adding factory methods for interactions.

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -3750,6 +3733,7 @@ + (TSOutgoingMessage *)createFakeOutgoingMessage:(TSThread *)thread
                                     messageState:(TSOutgoingMessageState)messageState
                                      isDelivered:(BOOL)isDelivered
                                           isRead:(BOOL)isRead
+                                  isVoiceMessage:(BOOL)isVoiceMessage
```
michaelkirk-signal: I think you want to pass this through right?

Signal/test/util/OWSDatabaseConverterTest.m
```diff
@@ -150,6 +141,8 @@ - (void)openYapDatabase:(NSString *)databaseFilePath
             [expectation fulfill];
         });
 
+        // YapDatabase can retain references to the registration
+        // connections for up to 5 seconds.
```
michaelkirk-signal: I don't understand this comment. It sounds like it's related to https://github.com/signalapp/Signal-iOS/pull/3214

My understanding is that we're waiting 5 seconds in this test simply because this is an async process that will take "a while", and 5 secs was a safe upper bound.

Ultimately, I'm not super concerned about nitpicking a comment in a test, but I want to make sure I understand whats going on.

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
charlesmchen-signal: Thanks!

Signal/test/util/OWSDatabaseConverterTest.m
charlesmchen-signal: Elaborated.

PR #3231, opened 2018-04-03
Update pods

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -1,18 +1,18 @@
 PODS:
-  - AFNetworking (3.1.0):
-    - AFNetworking/NSURLSession (= 3.1.0)
-    - AFNetworking/Reachability (= 3.1.0)
-    - AFNetworking/Security (= 3.1.0)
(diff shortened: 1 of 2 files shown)


Podfile.lock
```diff
@@ -1,38 +1,38 @@
...
+  - CocoaLumberjack/Extensions (3.4.1):
     - CocoaLumberjack/Default
   - Curve25519Kit (2.1.0)
   - GRKOpenSSLFramework (1.0.2.12)
   - HKDFKit (0.0.4)
   - JSQMessagesViewController (7.3.4):
     - JSQSystemSoundPlayer (~> 2.0.1)
   - JSQSystemSoundPlayer (2.0.1)
-  - libPhoneNumber-iOS (0.9.12)
+  - libPhoneNumber-iOS (0.9.13)
```
charlesmchen-signal: We're going to have to do a solid regression pass.

PR #3232, opened 2018-04-03
Charlesmchen/quoted replies vs oversize text

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSMessageUtils.m
+++ b/SignalServiceKit/src/Messages/OWSMessageUtils.m
@@ -4,6 +4,8 @@
 
 #import "OWSMessageUtils.h"
 #import "AppContext.h"
+#import "MIMETypeUtil.h"
+#import "OWSMessageSender.h"
 #import "OWSPrimaryStorage.h"
 #import "TSAccountManager.h"
 #import "TSAttachment.h"
@@ -136,8 +138,8 @@ NS_ASSUME_NONNULL_BEGIN
     OWSAssert(transaction);
 
     uint64_t timestamp = message.timestamp;
-    NSString *_Nullable body = message.body;
-    BOOL hasText = body.length > 0;
+    NSString *_Nullable quotedText = message.body;
+    BOOL hasText = quotedText.length > 0;
     BOOL hasAttachment = NO;
     NSString *_Nullable sourceFilename = nil;
     NSData *_Nullable thumbnailData = nil;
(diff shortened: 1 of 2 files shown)


SignalServiceKit/src/Messages/OWSMessageUtils.m
```diff
@@ -149,11 +151,46 @@ + (nullable TSQuotedMessage *)quotedMessageForMessage:(TSMessage *)message
         TSAttachment *_Nullable attachment =
             [TSAttachment fetchObjectWithUniqueID:attachmentId transaction:transaction];
         if (attachment) {
-            sourceFilename = attachment.sourceFilename;
-            contentType = attachment.contentType;
-            // Try to generate a thumbnail, if possible.
-            thumbnailData = [self thumbnailDataForAttachment:attachment];
-            hasAttachment = YES;
+            // If the attachment is "oversize text", try to treat it appropriately.
+            if (!hasText && [NSObject isNullableObject:attachment.contentType equalTo:OWSMimeTypeOversizeTextMessage] &&
```
michaelkirk-signal: small readability nit: I like the existence of `[NSObject isNullableObject:equalTo:]` when it would otherwise be replaced by some 5 line nil check pattern, but it's kind of mouth full when it's not really necessary, so I'd rather not start using it all over the place.

In this case, even if contentType **were** nil, it'd do the right thing, but if you don't like the idea of calling the method on nil, you could flip it:

```
[OWSMimeTypeOversizeTextMessage isEqual:attachment.contentType]
```

Or we could avoid it altogether in Swift.

SignalServiceKit/src/Messages/OWSMessageUtils.m
```diff
@@ -149,11 +151,46 @@ + (nullable TSQuotedMessage *)quotedMessageForMessage:(TSMessage *)message
...
+                        [[NSString alloc] initWithData:oversizeTextData encoding:NSUTF8StringEncoding];
+                    // First, truncate to the rough max characters.
+                    NSString *_Nullable truncatedText =
+                        [oversizeText substringToIndex:kOversizeTextMessageSizeThreshold - 1];
+                    // But kOversizeTextMessageSizeThreshold is in _bytes_, not characters,
+                    // so we need to continue to trim the string until it fits.
+                    while (truncatedText && truncatedText.length > 0 &&
+                        [truncatedText dataUsingEncoding:NSUTF8StringEncoding].length
+                            >= kOversizeTextMessageSizeThreshold) {
+                        // A very coarse binomial search by halving is acceptable, since
```
michaelkirk-signal: s/binomial/binary?

SignalServiceKit/src/Messages/OWSMessageUtils.m
```diff
@@ -149,11 +151,46 @@ + (nullable TSQuotedMessage *)quotedMessageForMessage:(TSMessage *)message
...
+                        // kOversizeTextMessageSizeThreshold is much longer than our target
+                        // length of "three short lines of text on any device we might
+                        // display this on.
+                        //
+                        // We don't worry much about the search converging because
```
michaelkirk-signal: comment trails off

SignalServiceKit/src/Messages/OWSMessageUtils.m
```diff
@@ -149,11 +151,46 @@ + (nullable TSQuotedMessage *)quotedMessageForMessage:(TSMessage *)message
...
+                quotedText = @"";
+
+                TSAttachmentStream *attachmentStream = (TSAttachmentStream *)attachment;
+                NSData *_Nullable oversizeTextData = [NSData dataWithContentsOfFile:attachmentStream.filePath];
+                if (oversizeTextData) {
+                    NSString *_Nullable oversizeText =
+                        [[NSString alloc] initWithData:oversizeTextData encoding:NSUTF8StringEncoding];
+                    // First, truncate to the rough max characters.
+                    NSString *_Nullable truncatedText =
+                        [oversizeText substringToIndex:kOversizeTextMessageSizeThreshold - 1];
```
michaelkirk-signal: So the usage of `kOversizeTextMessageSizeThreshold` here is pretty arbitrary, it's just some value which is "much more" than three lines, so that when we display the three lines of text we can be sure there's enough?

SignalServiceKit/src/Messages/OWSMessageUtils.m
```diff
@@ -162,9 +199,11 @@ + (nullable TSQuotedMessage *)quotedMessageForMessage:(TSMessage *)message
         return nil;
     }
 
+    // It's conceivable that the logic above will find neither valid text
+    // or an attachment to quote.
```
michaelkirk-signal: Outside of malformed data, I can't think of why this would happen. Should we assert?

SignalServiceKit/src/Messages/OWSMessageUtils.m
charlesmchen-signal: Done.

SignalServiceKit/src/Messages/OWSMessageUtils.m
charlesmchen-signal: Done.

SignalServiceKit/src/Messages/OWSMessageUtils.m
charlesmchen-signal: Sometimes one falls asleep while coding.

SignalServiceKit/src/Messages/OWSMessageUtils.m
charlesmchen-signal: It's only arbitrary in the sense that `kOversizeTextMessageSizeThreshold` is out limit on how long text should be in protos since they'll be stored in the database.  Similarly, the idea here is to cap the length of the "text snippet" we include in the "quoted reply" proto.

SignalServiceKit/src/Messages/OWSMessageUtils.m
charlesmchen-signal: Done.

PR #3233, opened 2018-04-04
Send quoted reply

michaelkirk-signal:
There are several design TODO's plus some other niceties, but this is at a level of base functionality.

**New reply item in edit menu**:

replyitem

**Quote other person (just text)**:
quoteplaintext

**Quote self (captioned image)**:
quotecaptionediimage

**Quote self (just image)**:
quoteselfimage

PTAL @charlesmchen

### DIFF
--- a/Signal/src/network/PushManager.m
+++ b/Signal/src/network/PushManager.m
@@ -191,9 +191,10 @@ NSString *const Signal_Message_MarkAsRead_Identifier = @"Signal_Message_MarkAsRe
             TSThread *thread = [TSThread fetchObjectWithUniqueID:threadId];
             NSString *replyText = responseInfo[UIUserNotificationActionResponseTypedTextKey];
 
+            // TODO we could do an actual quoted reply here, but most apps don't.
             [ThreadUtil sendMessageWithText:replyText
                 inThread:thread
-                quotedMessage:nil // TODO we could do an actual quoted reply here, but most apps don't.
+                quotedMessage:nil
                 messageSender:self.messageSender
                 success:^{
                     // TODO do we really want to mark them all as read?
--- a/Signal/src/network/PushManager.m
+++ b/Signal/src/network/PushManager.m
@@ -193,6 +193,7 @@ NSString *const Signal_Message_MarkAsRead_Identifier = @"Signal_Message_MarkAsRe
 
             [ThreadUtil sendMessageWithText:replyText
                 inThread:thread
+                quotedMessage:nil // TODO we could do an actual quoted reply here, but most apps don't.
                 messageSender:self.messageSender
                 success:^{
                     // TODO do we really want to mark them all as read?
--- a/Signal/src/views/QuotedReplyPreview.swift
+++ b/Signal/src/views/QuotedReplyPreview.swift
@@ -43,7 +43,7 @@ class QuotedReplyPreview: UIView {
         // used for text and cancel
         let foregroundColor: UIColor = .darkGray
 
-        let  authorLabel: UILabel = UILabel()
+        let authorLabel: UILabel = UILabel()
         authorLabel.textColor = authorColor
         authorLabel.text = Environment.current().contactsManager.displayName(forPhoneIdentifier: quotedMessage.authorId)
         authorLabel.font = .ows_dynamicTypeHeadline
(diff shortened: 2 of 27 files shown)


Signal/src/network/PushManager.m
```diff
@@ -193,6 +193,7 @@ - (void)application:(UIApplication *)application
 
             [ThreadUtil sendMessageWithText:replyText
                 inThread:thread
+                quotedMessage:nil // TODO we could do an actual quoted reply here, but most apps don't.
```
charlesmchen-signal: * Let's check in with Myles.
* Please put param/argument comments on the line before them.

SignalMessaging/utils/ThreadUtil.h
```diff
@@ -44,27 +44,32 @@ NS_ASSUME_NONNULL_BEGIN
...
                                    failure:(void (^)(NSError *error))failureHandler;
 
 + (TSOutgoingMessage *)sendMessageWithText:(NSString *)text
                                   inThread:(TSThread *)thread
+                             quotedMessage:(nullable TSQuotedMessage *)quotedMessage
                              messageSender:(OWSMessageSender *)messageSender;
 
 + (TSOutgoingMessage *)sendMessageWithAttachment:(SignalAttachment *)attachment
                                         inThread:(TSThread *)thread
+                                   quotedMessage:quotedMessage
```
charlesmchen-signal: Aren't you missing a type here?

SignalMessaging/utils/ThreadUtil.m
```diff
@@ -87,11 +90,10 @@ + (TSOutgoingMessage *)sendMessageWithText:(NSString *)text
                                 inThread:thread
                              messageBody:text
                            attachmentIds:[NSMutableArray new]
-                        expiresInSeconds:(configuration.isEnabled ? configuration.durationSeconds : 0)
-                                  expireStartedAt:0
+                        expiresInSeconds:(configuration.isEnabled ? configuration.durationSeconds : 0)expireStartedAt:0
```
charlesmchen-signal: Haha!  `clang-format` gotcha.

SignalServiceKit/src/Messages/Interactions/TSQuotedMessage.m
```diff
@@ -33,6 +33,17 @@ - (instancetype)initWithTimestamp:(uint64_t)timestamp
     return self;
 }
 
+- (nullable UIImage *)thumbnailImage
+{
+    if (self.thumbnailData.length == 0) {
+        return nil;
+    }
+
+    // PERF TODO cache
```
charlesmchen-signal: Is this tracked in Trello?

Signal/src/views/QuotedReplyPreview.swift
```diff
@@ -0,0 +1,120 @@
...
+            if let contentType = quotedMessage.contentType {
+                let emoji = TSAttachmentStream.emoji(forMimeType: contentType)
+                return "\(emoji) \(quotedMessage.body ?? "")"
+            } else {
+                return quotedMessage.body
+            }
+        }()
+
+        let thumbnailView: UIView? = {
+            if let image = quotedMessage.thumbnailImage() {
```
charlesmchen-signal: See the design in abstract.  I suspect we'll want to make this pretty similar - if not exactly similar - to the way quoted messages are shown in the conversation.

* The "quote stripe" should be on the left of just the text, not the quoted author.
* If there's no thumbnail, we should use a circle with a content-appropriate icon.
* If there's no quoted text/caption, we should show some text.

I propose that I pull out my work (not yet merged) to render quoted content in the conversation views into a separate cell (since it's closer to the current design and has some extra measurement functionality we need in the cells) and we try to reuse it here?

michaelkirk-signal: I'll add this to this list of things to go over with myles.

SignalServiceKit/src/Messages/Interactions/TSQuotedMessage.m
michaelkirk-signal: Added a perf card

SignalMessaging/utils/ThreadUtil.m
michaelkirk-signal: How embarrassing. I pulled this into a separate var so this doesn't keep happening.

Signal/src/network/PushManager.m
michaelkirk-signal: updated. I'll add this to the list of things to go over with myles.

SignalMessaging/utils/ThreadUtil.h
michaelkirk-signal: Sure am. I didn't even realize that could compile.

Issue #3224, opened 2018-04-03
Bypass screen lock 


leonardoporpora:
- [x] I have searched open and closed issues for duplicates

----------------------------------------

### Bug description
Bypass screen lock

### Steps to reproduce
- open the app
- Click cancel 
- click home button 
- open app and u get in without passcode

**Actual result:** Diesnt ask passcode
**Expected result:** Should ask passcode

### Device info

**Device**: iPhone 6s
**iOS version**: 11.3
**Signal version:** 2.23.0.14

leonardoporpora:
Debug log: https://debuglogs.org/785f6cd444069b46bae5dc8b5f763db536e9d84216a9d0bd7deb8e9b44eeb757.zip

leonardoporpora:
I made a video if you want I can send you

michaelkirk-signal:
I can't reproduce this. What is your screen lock timeout set to?

Keep in mind, we re-use biometric auth if it just happened. So if you've just unlocked your phone with your thumb a few seconds before, we don't redundantly ask you again when launching Signal.

Can you reproduce this behavior with the following:

- Switch hands so as not to inadvertently auth with TouchID
- Launch Signal
- See screen lock (as expected)
- Cancel alert, hit home button (with non TouchID hand!)
- You're now on your iOS home screen
- Launch Signal
- See screen lock (as expected)
- Signal launches and you see the Signal inbox without the lock screen

leonardoporpora:
I will send you a video for PoC where should I send it?

michaelkirk-signal:
You can send it to support@signal.org, please reference this issue URL.

michaelkirk-signal:
Can you also answer my questions?

> What is your screen lock timeout set to?


Can you reproduce this behavior with the following:

- Switch hands so as not to inadvertently auth with TouchID
- Launch Signal
- See screen lock (as expected)
- Cancel alert, hit home button (with non TouchID hand!)
- You're now on your iOS home screen
- Launch Signal
- See screen lock (as expected)
- Signal launches and you see the Signal inbox without the lock screen

Unless the video shows your hands it's not going to be the helpful.

leonardoporpora:
I sent the video. The screen lock was set to 15 minutes than I tried also with one minute. I don’t use touchID I use passcode as you will be able to see in the video I sent right now

jkic:
[...]

PR #3235, opened 2018-04-04
Fix screen lock edge case.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/util/OWSScreenLockUI.m
+++ b/Signal/src/util/OWSScreenLockUI.m
@@ -33,7 +33,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, nullable) NSDate *appEnteredBackgroundDate;
 @property (nonatomic, nullable) NSDate *appEnteredForegroundDate;
-@property (nonatomic, nullable) NSDate *lastUnlockAttemptDate;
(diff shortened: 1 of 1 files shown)


Signal/src/util/OWSScreenLockUI.m
```diff
@@ -33,7 +33,6 @@ @interface OWSScreenLockUI ()
 
 @property (nonatomic, nullable) NSDate *appEnteredBackgroundDate;
 @property (nonatomic, nullable) NSDate *appEnteredForegroundDate;
-@property (nonatomic, nullable) NSDate *lastUnlockAttemptDate;
```
charlesmchen-signal: This property was obsolete.

Signal/src/util/OWSScreenLockUI.m
```diff
@@ -158,7 +157,23 @@ - (void)setAppIsInBackground:(BOOL)appIsInBackground
 - (void)markAppAsInBackground
 {
     // Record the time when app entered background.
-    self.appEnteredBackgroundDate = [NSDate new];
+    BOOL shouldResetEnteredBackgroundDate = NO;
```
charlesmchen-signal: The fix.

leonardoporpora:
This doesn’t fix #3224 I downloaded 2.23.1 but the issue is not fixed and I am doing the same steps

PR #3243, opened 2018-04-05
Charlesmchen/quoted replies6

charlesmchen-signal:

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -315,7 +315,6 @@ NS_ASSUME_NONNULL_BEGIN
     if (bodyMediaView) {
         OWSAssert(self.loadCellContentBlock);
         OWSAssert(self.unloadCellContentBlock);
-        OWSAssert(!lastSubview);
 
         bodyMediaView.clipsToBounds = YES;
 
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -234,7 +234,7 @@ NS_ASSUME_NONNULL_BEGIN
     CGSize bodyTextContentSize = [self bodyTextSizeForContentWidth:self.contentWidth includeMargins:NO];
 
     self.bubbleView.isOutgoing = self.isOutgoing;
-    self.bubbleView.hideTail = self.viewItem.shouldHideBubbleTail;
+    self.bubbleView.hideTail = self.viewItem.shouldHideBubbleTail && !self.alwaysShowBubbleTail;
 
     if ([self.viewItem.interaction isKindOfClass:[TSMessage class]] && self.hasNonImageBodyContent) {
         TSMessage *message = (TSMessage *)self.viewItem.interaction;
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -898,9 +898,7 @@ NS_ASSUME_NONNULL_BEGIN
     return (int)floor(contentWidth * 0.8f);
 }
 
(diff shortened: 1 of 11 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -920,15 +918,14 @@ - (CGSize)quotedMessageSizeForViewWidth:(int)viewWidth
     return result;
 }
 
-- (CGSize)sizeForViewWidth:(int)viewWidth contentWidth:(int)contentWidth
+- (CGSize)sizeForContentWidth:(int)contentWidth
```
charlesmchen-signal: Currently, the message bubbles look exactly the same in conversation view and in message detail view. For example, they only occupy 80% of the "content width".  Incoming messages hug leading edge, outgoing messages hug trailing edge. 

We could change that so they are more "greedy".

They also skip their "tail" if they would in conversation view.  We could change that too.  I think I will.

Signal/src/ViewControllers/LongTextViewController.swift
```diff
@@ -0,0 +1,100 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+import SignalServiceKit
+import SignalMessaging
+
+@objc
+public class LongTextViewController: OWSViewController {
```
charlesmchen-signal: I added a dedicated view to show "long text".  Pretty simple.  Has a share button.

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -28,15 +28,9 @@ class MessageDetailViewController: OWSViewController, UIScrollViewDelegate, Medi
     var message: TSMessage
     var wasDeleted: Bool = false
 
-    var mediaMessageView: MediaMessageView?
```
charlesmchen-signal: This view got _way_ simpler.

SignalMessaging/utils/DisplayableText.swift
```diff
@@ -217,20 +217,20 @@ extension String {
     @objc
     public class func displayableText(_ rawText: String) -> DisplayableText {
         // Only show up to N characters of text.
-        let kMaxTextDisplayLength = 1024
+        let kMaxTextDisplayLength = 512
```
charlesmchen-signal: I reduced the max length of text we should in conversation view, since we were previously showing more than a screenful of Latin text on even an iPhone X, which has a particularly long screen.  Check out the screenshots.

I've discussed this briefly with Myles, although there's no clear decision on the exact right length.

charlesmchen-signal:

PTAL @michaelkirk 

simulator screen shot - iphone x - 2018-04-06 at 12 38 51
simulator screen shot - iphone x - 2018-04-06 at 12 38 45
simulator screen shot - iphone x - 2018-04-06 at 12 38 21
simulator screen shot - iphone x - 2018-04-06 at 12 38 16
simulator screen shot - iphone x - 2018-04-06 at 12 38 02
simulator screen shot - iphone x - 2018-04-06 at 12 37 56
simulator screen shot - iphone x - 2018-04-06 at 12 37 54
simulator screen shot - iphone x - 2018-04-06 at 12 37 43
simulator screen shot - iphone x - 2018-04-06 at 12 37 40
simulator screen shot - iphone x - 2018-04-06 at 12 37 38

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
charlesmchen-signal: > They also skip their "tail" if they would in conversation view. We could change that too. I think I will.

I just made this change.  It's not in the screenshots.

Signal/src/ViewControllers/LongTextViewController.swift
charlesmchen-signal: Yup, closed that issue.

charlesmchen-signal:
> This screenshot looks wrong on iPhoneX:
https://user-images.githubusercontent.com/36971147/38433337-706795d6-3998-11e8-8cd4-731e34e04e42.png

Card added.

PR #3246, opened 2018-04-06
Fix screen lock edge cases.

charlesmchen-signal:
PTAL @michaelkirk


leonardoporpora:
@charlesmchen-signal could you kindly tell me how have you fixed the problem of both the Screen locker bypass I reported?

PR #3251, opened 2018-04-09
Fix message details layout in iPhone X.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/AppSettings/OWSLinkedDevicesTableViewController.m
+++ b/Signal/src/ViewControllers/AppSettings/OWSLinkedDevicesTableViewController.m
@@ -46,13 +46,7 @@ int const OWSLinkedDevicesTableViewControllerSectionAddDevice = 1;
     self.tableView.rowHeight = UITableViewAutomaticDimension;
     self.tableView.estimatedRowHeight = 70;
 
-    // Fix a bug that only affects iOS 11.0.x and 11.1.x.
-    if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(11, 0) && !SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(11, 2)) {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wpartial-availability"
-        self.tableView.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentNever;
-#pragma clang diagnostic pop
-    }
+    [self.tableView applyScrollViewInsetsFix];
 
     self.dbConnection = [[OWSPrimaryStorage sharedManager] newDatabaseConnection];
     [self.dbConnection beginLongLivedReadTransaction];
--- a/SignalMessaging/categories/UIView+OWS.m
+++ b/SignalMessaging/categories/UIView+OWS.m
@@ -494,6 +494,7 @@ CGFloat ScaleFromIPhone5(CGFloat iPhone5Value)
 - (void)applyScrollViewInsetsFix
 {
     // Fix a bug that only affects iOS 11.0.x and 11.1.x.
+    // The symptom is a fix weird animation that happens when using the interactive pop gesture.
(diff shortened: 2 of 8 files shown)


michaelkirk-signal:
LGTM

Signal/src/ViewControllers/AppSettings/OWSLinkedDevicesTableViewController.m
```diff
@@ -46,13 +46,7 @@ - (void)viewDidLoad
     self.tableView.rowHeight = UITableViewAutomaticDimension;
     self.tableView.estimatedRowHeight = 70;
 
-    // Fix a bug that only affects iOS 11.0.x and 11.1.x.
-    if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(11, 0) && !SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(11, 2)) {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wpartial-availability"
-        self.tableView.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentNever;
-#pragma clang diagnostic pop
-    }
```
charlesmchen-signal: > Just checking - did you check that we have the appropriate content inset behavior?

Are you referring to this fix?

charlesmchen-signal: I've DRY'd it up in a category.

SignalMessaging/categories/UIView+OWS.m
```diff
@@ -487,4 +487,21 @@ - (void)logHierarchyUpwardLaterWithLabel:(NSString *)label
 
 @end
 
+#pragma mark -
+
+@implementation UIScrollView (OWS)
+
+- (void)applyScrollViewInsetsFix
+{
+    // Fix a bug that only affects iOS 11.0.x and 11.1.x.
```
charlesmchen-signal: Am I right in understanding that we should use this for any scroll view that:

* Could be the bottom-most scroll view of a given view.
* Isn't the root view for its view controller?

In my testing, it looks like we also want to apply to this 11.2.  If we don't, we end up with too much content inset.  Let's discuss?

michaelkirk-signal: This fix, as I recall, was not to fix the *amount* of content offset, but to fix the weird animation that happens when using the interactive pop gesture in iOS11.0-11.1.

Now that it's centralized, we can at least elaborate the comment in one place, as the current one's pretty opaque.

Signal/src/ViewControllers/AppSettings/OWSLinkedDevicesTableViewController.m
michaelkirk-signal: It's related to this fix, in that it's about contentInset, but I'm not sure if adding this code to the conversation details is necessary or will fix anything. 

I guess what I'm asking for is:

- have a full page of content 
- scroll all the way down.
- verify the home button isn't overlaying any content.

And to check that this holds for iOS10, iOS11.0-1, and iOS11.2

charlesmchen-signal: > verify the home button isn't overlaying any content.

I think that's now true.  The problem is that now we're adding "double content inset" to account for the home view button, at least in some cases.

SignalMessaging/categories/UIView+OWS.m
charlesmchen-signal: Improved the comment.

Signal/src/ViewControllers/AppSettings/OWSLinkedDevicesTableViewController.m
charlesmchen-signal: Verified that content insets are appropriate on iOS 9 and iOS 11 and iPhone X.

Lots of screenshots here:
[ScrollViewInsetsViPhoneXScreenshots.zip](https://github.com/signalapp/Signal-iOS/files/1908582/ScrollViewInsetsViPhoneXScreenshots.zip)

PR #3252, opened 2018-04-09
Mkirk/attachment handling

michaelkirk-signal:
Extracts thumbnail data into an attachment stream rather than in line data.

The change in functionality is:

If we receive a quoted reply which has a thumbnail, we first try to derive that thumbnail locally if the quoted attachment is available locally. Else, we fetch the thumbnail referenced by the: `DataMessage.Quote.QuotedAttachment.thumbnail` proto

There were two large ancillary changes to support this.

1. Since sending a message may now involve uploading multiple attachments, I converted the UploadingService to the OWSUploadOperation, which inherits from a new OWSOperation base class, which in turn extracted the "retry" logic from the OWSMessageSendOperation.

2. Introduce the QuotedReplyModel, a view model which lives in SignalMessaging. It allows most of our view logic to remain unchanged now that the attachment data lives in a TSAttachmentStream rather than on the TSQuotedMessage.

PTAL @charlesmchen


SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
```diff
@@ -79,24 +80,53 @@ - (void)runWithAttachmentHandler:(void (^)(TSAttachmentStream *attachmentStream)
...
                                                         messageBody:transcript.body
                                                       attachmentIds:[attachmentsProcessor.attachmentIds mutableCopy]
                                                    expiresInSeconds:transcript.expirationDuration
                                                     expireStartedAt:transcript.expirationStartedAt
                                                      isVoiceMessage:NO
                                                    groupMetaMessage:TSGroupMessageNone
-                                                      quotedMessage:nil];
+                                                      quotedMessage:transcript.quotedMessage];
+
+    TSQuotedMessage *_Nullable quotedMessage = transcript.quotedMessage;
```
michaelkirk-signal: We need to retest this once Desktop can actually send quoted replies. I'm going to add a canaray: `OWSAssert(transcript.quotedMessage == nil)`

SignalServiceKit/src/Messages/Attachments/OWSAttachmentsProcessor.m
```diff
@@ -35,37 +35,30 @@
 @interface OWSAttachmentsProcessor ()
 
 @property (nonatomic, readonly) TSNetworkManager *networkManager;
-@property (nonatomic, readonly) OWSPrimaryStorage *primaryStorage;
-@property (nonatomic, readonly) NSArray<TSAttachmentPointer *> *supportedAttachmentPointers;
```
michaelkirk-signal: "supportedAttachmentPointers" is a legacy field from when we didn't support arbitrary attachments.

SignalServiceKit/src/Messages/Attachments/OWSAttachmentsProcessor.m
```diff
@@ -74,52 +67,25 @@ - (instancetype)initWithAttachmentProtos:(NSArray<OWSSignalServiceProtosAttachme
...
-
-        TSAttachmentPointer *pointer = [[TSAttachmentPointer alloc] initWithServerId:attachmentProto.id
-                                                                                 key:attachmentProto.key
-                                                                              digest:digest
-                                                                           byteCount:attachmentProto.size
-                                                                         contentType:attachmentProto.contentType
-                                                                               relay:relay
-                                                                      sourceFilename:attachmentProto.fileName
-                                                                      attachmentType:attachmentType];
+        TSAttachmentPointer *pointer = [TSAttachmentPointer attachmentPointerFromProto:attachmentProto relay:relay];
```
michaelkirk-signal: This new factory method on AttachmentPointers is re-used in a couple of places.

SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
michaelkirk-signal: Done.

SignalServiceKit/src/Messages/Interactions/TSQuotedMessage.h
```diff
@@ -2,11 +2,47 @@
...
-@interface TSQuotedMessage : TSYapDatabaseObject
+@class OWSSignalServiceProtosDataMessage;
+@class OWSSignalServiceProtosEnvelope;
+@class TSAttachment;
+@class TSAttachmentStream;
+@class TSQuotedMessage;
+@class TSThread;
+@class YapDatabaseReadWriteTransaction;
+
+@interface OWSAttachmentInfo: MTLModel
```
michaelkirk-signal: Rather than having an array of attachmentId's and a dictionary of `(attachmentId) -> sourceFileName` I made a small serializable model class to wrap that information.

SignalServiceKit/src/Messages/Interactions/TSQuotedMessage.h
```diff
@@ -2,11 +2,47 @@
...
+
+- (instancetype)initWithAttachmentId:(nullable NSString *)attachmentId
+                         contentType:(NSString *)contentType
+                      sourceFilename:(NSString *)sourceFilename NS_DESIGNATED_INITIALIZER;
+
+- (instancetype)initWithAttachment:(TSAttachment *)attachment;
+
+@end
+
+@interface TSQuotedMessage : MTLModel
```
michaelkirk-signal: Note: TSQuotedMessage is now a MTLModel, not a TSYapDatabaseObject.

Only classes that live in their own collection should inherit from TSYapDatabaseObject.

e.g. we'd never want to call: 

```
[quotedMessage setAttachmentId:foo];
[quotedMessage saveWithTransaction:transaction]
```

Instead we do something like this:

```
[message.quotedMessage setAttachmentId:foo]`
[message saveWithTransaction:transaction];
```

SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -512,11 +515,8 @@ - (void)handleReceivedGroupAvatarUpdateWithEnvelope:(OWSSignalServiceProtosEnvel
     OWSAssert(groupThread);
     OWSAttachmentsProcessor *attachmentsProcessor =
         [[OWSAttachmentsProcessor alloc] initWithAttachmentProtos:@[ dataMessage.group.avatar ]
-                                                        timestamp:envelope.timestamp
```
michaelkirk-signal: these params weren't used.

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -3399,11 +3401,12 @@ + (void)insertAndDeleteNewOutgoingMessages:(NSUInteger)count
...
-            [TSOutgoingMessage outgoingMessageInThread:thread
-                                           messageBody:text
-                                          attachmentId:nil
-                                      expiresInSeconds:(configuration.isEnabled ? configuration.durationSeconds : 0)];
+
+        uint32_t expiresInSeconds = (configuration.isEnabled ? configuration.durationSeconds : 0);
+        TSOutgoingMessage *message = [TSOutgoingMessage outgoingMessageInThread:thread
+                                                                    messageBody:text
+                                                                   attachmentId:nil
+                                                               expiresInSeconds:expiresInSeconds];
```
charlesmchen-signal: Maybe we should modify the `[TSOutgoingMessage outgoingMessageInThread:...` factory method to take a nullable `OWSDisappearingMessagesConfiguration`, to DRY this up.

No need to do this in this PR, just thinking out loud.

SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -1162,7 +1095,36 @@ - (void)finalizeIncomingMessage:(TSIncomingMessage *)incomingMessage
...
+                transaction:transaction
+                success:^(TSAttachmentStream *_Nonnull attachmentStream) {
+                    [self.dbConnection
+                        asyncReadWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
+                            [incomingMessage setQuotedMessageThumbnailAttachmentStream:attachmentStream];
+                            [incomingMessage saveWithTransaction:transaction];
+                        }];
+                }
+                failure:^(NSError *_Nonnull error) {
+                    DDLogWarn(@"%@ failed to fetch thumbnail for message: %tu with error: %@",
```
michaelkirk-signal: We don't currently handle retry. 

Keep in mind, most quotes will be a recent message, which allows the thumbnail to be generated locally rather than fetched over the network, so I think it makes sense to prefer a simple rather than perfect solution here.

With attachments the pattern is for a tap-to-retry. We could do something similar with thumbnail attachments. WDYT @charlesmchen?

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -118,47 +68,36 @@ - (void)setIsFatal:(BOOL)value
  * Used by `OWSMessageSender` to serialize message sending, ensuring that messages are emitted in the order they
  * were sent.
  */
-@interface OWSSendMessageOperation : NSOperation
+@interface OWSSendMessageOperation : OWSOperation
```
michaelkirk-signal: Extracted the retry logic into a base class, which subclasses take advantage of by overriding hooks.

The base class is also used for the new Upload operation.

I feel like this is easier to read and reuse.

SignalMessaging/Models/OWSQuotedReplyModel.m
```diff
@@ -0,0 +1,183 @@
...
+
+    UIImage *_Nullable thumbnailImage;
+    if (attachmentInfo.thumbnailAttachmentStreamId) {
+        TSAttachment *attachment =
+            [TSAttachment fetchObjectWithUniqueID:attachmentInfo.thumbnailAttachmentStreamId transaction:transaction];
+
+        TSAttachmentStream *attachmentStream;
+        if ([attachment isKindOfClass:[TSAttachmentStream class]]) {
+            attachmentStream = (TSAttachmentStream *)attachment;
+            thumbnailImage = attachmentStream.image;
```
charlesmchen-signal: This looks like it's eagerly loading into memory all thumbnails for all quoted messages as soon as the `ConversationViewItem` is created.  I suspect we'll want to do that lazily.  In fact, to support browsing _far_ back into conversation histories, I suspect we'll want to eventually unload these thumbnails if they are _far_ offscreen.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -168,144 +107,98 @@ @implementation OWSSendMessageOperation
...
 {
     self = [super init];
     if (!self) {
         return self;
     }
 
-    _operationState = OWSSendMessageOperationStateNew;
-    self.backgroundTask = [OWSBackgroundTask backgroundTaskWithLabelStr:__PRETTY_FUNCTION__];
-
+    self.remainingRetries = 6;
```
michaelkirk-signal: Note I bumped this up to 6. 

Previously the uploading process used the same retry counter as message sending.

As I recall it, we wanted a lower retry count because upload retries could take a *really* long time.

e.g. the scenario:

1. Alice sends Bob a large attachment
2. Alice sends Bob a short text message

Since sending the attachment (1) blocks sending the message (2), if Alice is on a bad network, it can take a really long time before Alice even tries to fire off (2).

In this case, we want (1) to fail relatively fast if it means that (2) can have a chance to succeed.

So: Uploads still retry 4 times, but message sends now retry a couple more times. Happy to tweak these, but my intuition is that the desired retry behavior for uploading attachments is not the same as that for message sending.

SignalMessaging/Models/OWSQuotedReplyModel.h
```diff
@@ -0,0 +1,48 @@
...
+
+- (instancetype)initWithTimestamp:(uint64_t)timestamp
+                         authorId:(NSString *)authorId
+                             body:(NSString *_Nullable)body
+                 attachmentStream:(nullable TSAttachmentStream *)attachment;
+
+- (instancetype)initWithQuotedMessage:(TSQuotedMessage *)quotedMessage
+                          transaction:(YapDatabaseReadTransaction *)transaction;
+
++ (nullable instancetype)quotedReplyForMessage:(TSMessage *)message
```
charlesmchen-signal: Could you please add a comment on each initializer/factory method documenting how it is intended to be used?

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -168,144 +107,98 @@ @implementation OWSSendMessageOperation
...
         return;
     }
 
-    // Use this flag to ensure a given operation only succeeds or fails once.
-    __block BOOL onceFlag = NO;
-    RetryableFailureHandler retryableFailureHandler = ^(NSError *_Nonnull error) {
-        DDLogInfo(@"%@ Sending failed. Remaining retries: %lu", self.logTag, (unsigned long)remainingRetries);
-
-        OWSAssert(!onceFlag);
-        onceFlag = YES;
```
michaelkirk-signal: This once flag is captured in the synchronized assertion in the super class, which asserts that markForCompletion is only ever called once.

SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
```diff
@@ -79,24 +80,58 @@ - (void)runWithAttachmentHandler:(void (^)(TSAttachmentStream *attachmentStream)
...
                                                     expireStartedAt:transcript.expirationStartedAt
                                                      isVoiceMessage:NO
                                                    groupMetaMessage:TSGroupMessageNone
-                                                      quotedMessage:nil];
+                                                      quotedMessage:transcript.quotedMessage];
+
+    // TODO: When written, desktop didn't yet support sending quotedMessages, so we didn't have a
+    // good way to test the handling of transcripts with a quotedMessage. This assertion can be delete
+    // once we've tested transcripts with quoted messages are processed correctly.
+    OWSAssert(transcript.quotedMessage == nil);
```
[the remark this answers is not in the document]
michaelkirk-signal: Sure, I'll create a card and add it to the backlog. Keep in mind, Desktop is initially not supporting sending - just receiving these messages. AFAIK sending isn't yet imminently scheduled - hence this canary.

SignalServiceKit/src/Messages/Interactions/TSQuotedMessage.h
```diff
@@ -2,11 +2,47 @@
...
+@interface OWSAttachmentInfo: MTLModel
+
+@property (nonatomic, readonly, nullable) NSString *contentType;
+@property (nonatomic, readonly, nullable) NSString *sourceFilename;
+
+// This is only set when sending a new attachment so we have a way
+// to reference the original attachment when generating a thumbnail.
+// We don't want to do this until the message is saved, when the user sends
+// the message so as not to end up with an orphaned file.
+@property (nonatomic, readonly, nullable) NSString *attachmentId;
```
charlesmchen-signal: So when we send an outgoing quoted reply, we have an instance of `OWSQuotedReplyModel` used in the authoring UI that generates a thumbnail, and then when we send the message, it is saved in the database and a new instance of `OWSQuotedReplyModel` (which makes its own thumbnail) is created by the conversation view (i.e. in the `ConversationViewItem`)?

SignalServiceKit/src/Messages/Interactions/TSQuotedMessage.h
```diff
@@ -15,25 +51,49 @@ NS_ASSUME_NONNULL_BEGIN
...
+    receivedQuotedAttachmentInfos:(NSArray<OWSAttachmentInfo *> *)attachmentInfos;
+
+// used when sending quoted messages
 - (instancetype)initWithTimestamp:(uint64_t)timestamp
                          authorId:(NSString *)authorId
                              body:(NSString *_Nullable)body
-                   sourceFilename:(NSString *_Nullable)sourceFilename
-                    thumbnailData:(NSData *_Nullable)thumbnailData
-                      contentType:(NSString *_Nullable)contentType;
+      quotedAttachmentsForSending:(NSArray<TSAttachment *> *)attachments;
```
charlesmchen-signal: When would there be multiple attachments?

SignalServiceKit/src/Messages/Interactions/TSQuotedMessage.m
```diff
@@ -3,46 +3,329 @@
...
+    for (OWSAttachmentInfo *info in self.quotedAttachments) {
+        if (info.thumbnailAttachmentStreamId) {
+            [streamIds addObject:info.thumbnailAttachmentStreamId];
+        }
+    }
+
+    return [streamIds copy];
+}
+
+- (NSArray<TSAttachmentStream *> *)createThumbnailAttachmentsIfNecessaryWithTransaction:
```
charlesmchen-signal: Where is this method called?

SignalMessaging/Models/OWSQuotedReplyModel.m
michaelkirk-signal: That's a good plan. 

Since this behavior is essentially the same as before when we were passing around the raw `(NSData *)thumbnailData`, and likely entails changes in the view layer to request this data be loaded/unloaded I'd like to leave this proposed enhancement for a separate PR.

SignalServiceKit/src/Messages/OWSMessageSender.m
charlesmchen-signal: Should we use a different value here based on the number of pre-send uploads?

SignalMessaging/Models/OWSQuotedReplyModel.h
michaelkirk-signal: done!

SignalServiceKit/src/Storage/OWSOrphanedDataCleaner.m
```diff
@@ -104,6 +105,8 @@ + (void)auditAndCleanup:(BOOL)shouldCleanup completion:(void (^_Nullable)(void))
 
     NSMutableSet<NSString *> *orphanInteractionIds = [NSMutableSet new];
     NSMutableSet<NSString *> *messageAttachmentIds = [NSMutableSet new];
+    NSMutableSet<NSString *> *quotedReplyThumbnailAttachmentIds = [NSMutableSet new];
```
charlesmchen-signal: Maybe we should just delete this class for now.  It'll need a heavy rewrite.

SignalServiceKit/src/Messages/Attachments/TSAttachmentPointer.m
```diff
@@ -51,6 +51,36 @@ - (instancetype)initWithServerId:(UInt64)serverId
...
+
+    TSAttachmentType attachmentType = TSAttachmentTypeDefault;
+    if ([attachmentProto hasFlags]) {
+        UInt32 flags = attachmentProto.flags;
+        if ((flags & (UInt32)OWSSignalServiceProtosAttachmentPointerFlagsVoiceMessage) > 0) {
+            attachmentType = TSAttachmentTypeVoiceMessage;
+        }
+    }
+
+    TSAttachmentPointer *pointer = [[TSAttachmentPointer alloc] initWithServerId:attachmentProto.id
```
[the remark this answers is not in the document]
michaelkirk-signal: I'm inclined to keep the existing behavior for now - this method was extracted as-is from our existing parsing logic which asserts some things about the attachment pointer but always returns a record to be saved.

Presumably malformed attachment pointers would be reflected to the user as some kind of "failure to download" which is, I believe, preferable to just pretending that there was no attachment.

So any change here will require testing the edge cases with malformed data and making sure it's better than the current behavior. I'll add a card and we can prioritize it accordingly.

SignalServiceKit/src/Util/OWSOperation.m
```diff
@@ -0,0 +1,181 @@
...
+        [self failOperationWithError:error];
+        return;
+    }
+
+    self.remainingRetries--;
+
+    // TODO Do we want some kind of exponential backoff?
+    // I'm not sure that there is a one-size-fits all backoff approach
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.1 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
+        [self run];
```
charlesmchen-signal: Huh, I've never seen operations re-used for multiple retries. I thought they were expected start/complete once.  For example, I would expect this to break `NSOperation` dependencies, `NSOperationQueue`, etc.

SignalServiceKit/src/Util/OWSOperation.m
```diff
@@ -0,0 +1,181 @@
...
+#import "OWSBackgroundTask.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+NSString *const OWSOperationKeyIsExecuting = @"isExecuting";
+NSString *const OWSOperationKeyIsFinished = @"isFinished";
+
+@interface OWSOperation ()
+
+@property (nonatomic) OWSOperationState operationState;
```
charlesmchen-signal: Shouldn't this be atomic?

SignalServiceKit/src/Messages/Interactions/TSQuotedMessage.h
michaelkirk-signal: That's right. 

OWSQuotedReplyModel is not persisted, but serves as a "fully loaded" view model on top of the persisted TSQuotedMessage.

SignalServiceKit/src/Messages/Interactions/TSQuotedMessage.h
michaelkirk-signal: I have no idea. 

The protos support multiple attachments and Moxie keeps threatening for this to be a reality, so I tried to accommodate it on the receiving side while adding accessors that just deal with the "first" instance since that will be the normal case.

SignalServiceKit/src/Messages/Interactions/TSQuotedMessage.m
michaelkirk-signal: In OWSMesageSender, before sending. If we persist the thumbnails before that, they risk getting orphaned.

I'll add a comment.

SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
```diff
@@ -79,24 +80,58 @@ - (void)runWithAttachmentHandler:(void (^)(TSAttachmentStream *attachmentStream)
...
+        TSAttachmentPointer *attachmentPointer =
+            [TSAttachmentPointer fetchObjectWithUniqueID:quotedMessage.thumbnailAttachmentPointerId
+                                             transaction:transaction];
+
+        if ([attachmentPointer isKindOfClass:[TSAttachmentPointer class]]) {
+            OWSAttachmentsProcessor *attachmentProcessor =
+                [[OWSAttachmentsProcessor alloc] initWithAttachmentPointer:attachmentPointer
+                                                            networkManager:self.networkManager];
+
+            DDLogDebug(@"%@ downloading thumbnail for transcript: %tu", self.logTag, transcript.timestamp);
```
charlesmchen-signal: Sorry, this PR is long and I may have missed it but: is this the only place where we initiate downloads of quoted reply thumbnails?  I suspect we'll want to initiate them in the views too, e.g.

* Alice receives quoted reply from Bob, starting downloading thumbnail.
* App is killed before download completes.
* Alice launches app, looks at conversation with Bob.  Expects quote-thumbnail download to retry (or _maybe_ tap to retry but I suspect we'll just want to retry automatically).

SignalServiceKit/src/Util/OWSOperation.m
michaelkirk-signal: I think you're right. But we're not re-starting the operation. The operation only starts once, when `main` is called.

Our retry logic all happens within `main`.

SignalServiceKit/src/Util/OWSOperation.m
michaelkirk-signal: updated.

SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
michaelkirk-signal: I posted this comment in MessagesManager after opening the PR so maybe you missed it but I'll repost here:

> We don't currently handle retry.
> 
> Keep in mind, most quotes will be a recent message, which allows the thumbnail to be generated locally rather than fetched over the network, so I think it makes sense to prefer a simple rather than perfect solution here.
>
> With attachments the pattern is for a tap-to-retry. We could do something similar with thumbnail attachments. WDYT @charlesmchen?

SignalServiceKit/src/Messages/Interactions/TSQuotedMessage.h
charlesmchen-signal: +1

SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
charlesmchen-signal: Yup just saw it.  Replied above.

PR #3256, opened 2018-04-09
Enforce quoted reply edge cases.

charlesmchen-signal:
* Don't let users reply to non-sent outgoing messages.
* Don't let users reply to non-downloaded incoming attachments.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
@@ -658,7 +658,11 @@ NSString *NSStringForOWSMessageCellType(OWSMessageCellType cellType)
             TSIncomingMessage *incomingMessage = (TSIncomingMessage *)self.interaction;
             if (incomingMessage.hasAttachments) {
                 NSString *attachmentId = incomingMessage.attachmentIds.firstObject;
-                TSAttachment *_Nullable attachment = [TSAttachment fetchObjectWithUniqueID:attachmentId];
+                __block TSAttachment *_Nullable attachment = nil;
+                [[OWSPrimaryStorage.sharedManager newDatabaseConnection]
+                    readWithBlock:^(YapDatabaseReadTransaction *transaction) {
+                        attachment = [TSAttachment fetchObjectWithUniqueID:attachmentId transaction:transaction];
+                    }];
                 if (![attachment isKindOfClass:[TSAttachmentStream class]]) {
                     // Don't let users reply to attachments which aren't yet downloaded
                     // (or otherwise missing on disk).
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
@@ -647,6 +647,25 @@ NSString *NSStringForOWSMessageCellType(OWSMessageCellType cellType)
     } else if (action == self.metadataActionSelector) {
         return YES;
     } else if (action == self.replyActionSelector) {
+        if ([self.interaction isKindOfClass:[TSOutgoingMessage class]]) {
+            TSOutgoingMessage *outgoingMessage = (TSOutgoingMessage *)self.interaction;
+            if (outgoingMessage.messageState == TSOutgoingMessageStateUnsent
+                || outgoingMessage.messageState == TSOutgoingMessageStateAttemptingOut) {
+                // Don't let users reply to messages which aren't yet delivered to the service.
+                return NO;
+            }
+        } else if ([self.interaction isKindOfClass:[TSIncomingMessage class]]) {
+            TSIncomingMessage *incomingMessage = (TSIncomingMessage *)self.interaction;
+            if (incomingMessage.hasAttachments) {
+                NSString *attachmentId = incomingMessage.attachmentIds.firstObject;
+                TSAttachment *_Nullable attachment = [TSAttachment fetchObjectWithUniqueID:attachmentId];
+                if (![attachment isKindOfClass:[TSAttachmentStream class]]) {
+                    // Don't let users reply to attachments which aren't yet downloaded
+                    // (or otherwise missing on disk).
+                    return NO;
+                }
+            }
+        }
         return YES;
     } else {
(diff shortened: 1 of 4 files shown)


SignalMessaging/Models/OWSQuotedReplyModel.m
```diff
@@ -170,7 +170,8 @@ + (nullable OWSQuotedReplyModel *)quotedReplyForMessage:(TSMessage *)message
 
     if (!hasText && !hasAttachment) {
         OWSFail(@"%@ quoted message has neither text nor attachment", self.logTag);
-        return nil;
+        quotedText = @"";
+        hasText = YES;
```
charlesmchen-signal: In prod, I suspect we should treat this as an "empty" message.  Note that this can happen if, for example, all of the quoted text is filtered.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -647,6 +647,25 @@ - (BOOL)canPerformAction:(SEL)action
...
+            if (outgoingMessage.messageState == TSOutgoingMessageStateUnsent
+                || outgoingMessage.messageState == TSOutgoingMessageStateAttemptingOut) {
+                // Don't let users reply to messages which aren't yet delivered to the service.
+                return NO;
+            }
+        } else if ([self.interaction isKindOfClass:[TSIncomingMessage class]]) {
+            TSIncomingMessage *incomingMessage = (TSIncomingMessage *)self.interaction;
+            if (incomingMessage.hasAttachments) {
+                NSString *attachmentId = incomingMessage.attachmentIds.firstObject;
+                TSAttachment *_Nullable attachment = [TSAttachment fetchObjectWithUniqueID:attachmentId];
```
michaelkirk-signal: This will open a transaction on the default TSYapDatabaseModel.readWriteConnection which could block since we use this connection all over the place.

Can you come up with a solution which uses the uiDatabaseConnection?

SignalMessaging/Models/OWSQuotedReplyModel.m
michaelkirk-signal: 👍

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -347,7 +347,7 @@ - (nullable UIImage *)thumbnailImage
     }
 
     if (![[NSFileManager defaultManager] fileExistsAtPath:thumbnailPath]) {
-        OWSFail(@"%@ missing thumbnail for attachmentId: %@", self.logTag, self.uniqueId);
+        DDLogError(@"%@ missing thumbnail for attachmentId: %@", self.logTag, self.uniqueId);
```
michaelkirk-signal: When does this happen where we'd want to ignore this assertion? Is it during a restore? can you leave a comment?

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
charlesmchen-signal: > This will open a transaction on the default TSYapDatabaseModel.readWriteConnection 

Yup, this was a tough choice.  

 I `uiDatabaseConnection` considered using but:

* I'm not super concerned about race conditions since our authoring/sending code should be robust to edge cases, so we don't need the extra consistency that would come with `uiDatabaseConnection`.
* I don't see how to use `uiDatabaseConnection` without adding a property to the view model.  That's worrisome.  We pass around the the "item" (which is really a view model, imo, I just wanted a shorter name) to multiple other views.  It just seems like we're setting ourselves up for later bugs/surprise if we start baking connections into models or view models.  
* I'm not super concerned about blocking, since although we do use `TSYapDatabaseModel.readWriteConnection` in many places, we shouldn't be using it anywhere that will block for long.

Again, tough choice.  Happy to discuss further.

michaelkirk-signal: > I'm not super concerned about blocking, since although we do use TSYapDatabaseModel.readWriteConnection in many places, we shouldn't be using it anywhere that will block for long.

I think you mean that writes using the `TSYapDatabaseModel.readWriteConnection` should be quick because we, as habit, only use it for quick things like `[someModel save]`. 

Verifying that would be a lot of work - we use that connection all over the place (because it's so convenient), but even if it were true,  we can't control whether a completely unrelated long running write transaction is happening on another connection, which would block the "quick" one from starting. That's why we need to avoid opening even small write transactions that would block the UI.

> I don't see how to use uiDatabaseConnection without adding a property to the view model. That's worrisome. We pass around the the "item" (which is really a view model, imo, I just wanted a shorter name) to multiple other views. It just seems like we're setting ourselves up for later bugs/surprise if we start baking connections into models or view models.

I totally agree with this, and it's for that reason that I don't think using the read write dbConnection on the model is any better, the only difference is the dbConnection owned by the model is hidden a few stack frames down. I'd really like to phase it out because it's too easy to use it on the main thread.

A couple alternatives that spring to mind:

Load the attachment when instantiating the viewItem. The concern there is maybe(?) we'd be swapping one perf problem for another - since maybe(?) there's something to be gained by not fetching the TSAttachment until it's needed. But I *think* we end up fetching every Attachment anyway, by way of `ConversationViewTiem.initWithInteraction->ensureViewState->firstAttachmentIfAnyWithTransaction:` right? What if we maintained a reference to the attachment at that point?

Another alternative, would be to introduce a delegate. Either at the level of the ConversationViewItem or possibly on whatever view(s) are using the ConversationViewItem.

charlesmchen-signal: I pushed a commit, PTAL @michaelkirk 

Basically I use a throwaway database connection.

This method is `canPerformAction`.  It's used when displaying the menu controller to determine which items to include.  We call it rarely and it's not a perf hotspot.  

WDYT?

michaelkirk-signal: LGTM

PR #3259, opened 2018-04-09
Don't use scaledFontForFont.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/categories/UIFont+OWS.m
+++ b/SignalMessaging/categories/UIFont+OWS.m
@@ -122,15 +122,7 @@ NS_ASSUME_NONNULL_BEGIN
     // But this doesn't seem to work in practice on iOS 11 using UIFontWeightMedium.
 
     UIFont *derivedFont = [UIFont systemFontOfSize:self.pointSize weight:UIFontWeightMedium];
-
-    if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(11, 0)) {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wpartial-availability"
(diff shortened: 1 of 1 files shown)


SignalMessaging/categories/UIFont+OWS.m
```diff
@@ -122,15 +122,7 @@ - (UIFont *)ows_medium
     // But this doesn't seem to work in practice on iOS 11 using UIFontWeightMedium.
 
     UIFont *derivedFont = [UIFont systemFontOfSize:self.pointSize weight:UIFontWeightMedium];
-
-    if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(11, 0)) {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wpartial-availability"
-        return [[UIFontMetrics defaultMetrics] scaledFontForFont:derivedFont];
```
charlesmchen-signal: We don't want to use `scaledFontForFont`.

michaelkirk-signal: Your comment above still says "But this doesn't seem to work in practice on iOS 11 using UIFontWeightMedium."

If that's not the case, could you update the comment? If it is still the case seems like theres more work to do?

charlesmchen-signal: Yup, `[UIFontDescriptor fontDescriptorByAddingAttributes:...]` doesn't work so we use `[UIFont systemFontOfSize: weight:]` as it stands here.  

I find the docs around `scaledFontForFont` confusing.  I didn't realize that it would affect the font size in some cases.

PR #3263, opened 2018-04-10
limit caption length

michaelkirk-signal:
PTAL @charlesmchen


SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
```diff
@@ -415,7 +416,21 @@ - (OWSSignalServiceProtosDataMessageBuilder *)dataMessageBuilder
     
     OWSSignalServiceProtosDataMessageBuilder *builder = [OWSSignalServiceProtosDataMessageBuilder new];
     [builder setTimestamp:self.timestamp];
-    [builder setBody:self.body];
+
+
+    if ([self.body lengthOfBytesUsingEncoding:NSUTF8StringEncoding] <= kOversizeTextMessageSizeThreshold) {
```
michaelkirk-signal: Be more defensive about enforcing this limit in production in case we missed another entry point.

charlesmchen-signal:
Screenshots?

michaelkirk-signal:
Merging this as is, but it's likely the design will change. Waiting on direction from myles.

PR #3267, opened 2018-04-11
Fix quoted message overflow.

charlesmchen-signal:
PTAL @michaelkirk


charlesmchen-signal:
> This solves the problem where message bubbles containing a quote from an author with a very long name were being rendered wider than the collection view, right?

Yup

PR #3268, opened 2018-04-11
Fix long press on quoted reply.

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -630,28 +632,26 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)panRecognizer
 
 - (void)showTextMenuController:(CGPoint)fromLocation
```
charlesmchen-signal: DRY'd up nicely.

PR #3270, opened 2018-04-11
Label quoted replies to yourself as such.

charlesmchen-signal:
PTAL @michaelkirk 

signal-2018-04-11-114712
signal-2018-04-11-114705

### DIFF
--- a/Signal/translations/en.lproj/Localizable.strings
+++ b/Signal/translations/en.lproj/Localizable.strings
@@ -1411,6 +1411,9 @@
 /* Indicates the author of a quoted message. Embeds {{the author's name or phone number}}. */
 "QUOTED_REPLY_AUTHOR_INDICATOR_FORMAT" = "Replying to %@";
 
+/* Indicates the author of a quoted message is yourself. */
+"QUOTED_REPLY_AUTHOR_INDICATOR_YOURSELF" = "Yourself";
+
 /* Indicates this message is a quoted reply to an attachment of unknown type. */
 "QUOTED_REPLY_TYPE_ATTACHMENT" = "Attachment";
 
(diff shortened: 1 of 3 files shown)


SignalMessaging/Models/OWSQuotedReplyModel.m
```diff
@@ -154,7 +154,7 @@ + (nullable OWSQuotedReplyModel *)quotedReplyForMessage:(TSMessage *)message
                     // The search will always converge since in the worst case (namely
                     // a single character which in utf-8 is >= 1024 bytes) the loop will
                     // exit when the string is empty.
-                    truncatedText = [truncatedText substringToIndex:oversizeText.length / 2];
+                    truncatedText = [truncatedText substringToIndex:truncatedText.length / 2];
```
charlesmchen-signal: Fixed a crash here.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1411,6 +1411,9 @@
 /* Indicates the author of a quoted message. Embeds {{the author's name or phone number}}. */
 "QUOTED_REPLY_AUTHOR_INDICATOR_FORMAT" = "Replying to %@";
 
+/* Indicates the author of a quoted message is yourself. */
+"QUOTED_REPLY_AUTHOR_INDICATOR_YOURSELF" = "Yourself";
```
michaelkirk-signal: I was thinking instead of "Replying to Alice" we'd say "Replying to Yourself", rather than just "Yourself". WDYT?

If we do go that route, it might still be good to have a separate, albeit redundant-in-english string like "Replying to Yourself" rather than trying to embed "Yourself" into `QUOTED_REPLY_AUTHOR_INDICATOR_FORMAT`. I'm not familiar enough with all our translations to know if that would work.

SignalMessaging/Models/OWSQuotedReplyModel.m
michaelkirk-signal: Good catch!

Signal/translations/en.lproj/Localizable.strings
charlesmchen-signal: Done.

PR #3279, opened 2018-04-11
Add more asserts around thread in SAE.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/CallViewController.swift
+++ b/Signal/src/ViewControllers/CallViewController.swift
@@ -375,7 +375,7 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver, 
     }
 
     func presentAudioSourcePicker() {
-        AssertIsOnMainThread()
+        SwiftAssertIsOnMainThread(#function)
 
         let actionSheetController = UIAlertController(title: nil, message: nil, preferredStyle: .actionSheet)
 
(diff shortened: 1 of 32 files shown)


Signal/src/ViewControllers/CallViewController.swift
```diff
@@ -375,7 +375,7 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver,
     }
 
     func presentAudioSourcePicker() {
-        AssertIsOnMainThread()
```
charlesmchen-signal: We were often confusing `AssertIsOnMainThread()` which is intended for Swift code and `OWS AssertIsOnMainThread()` which is intended for Obj-C, so I've renamed the former to make clear that it is for swift usage only.

SignalShareExtension/ShareViewController.swift
```diff
@@ -136,6 +138,8 @@ public class ShareViewController: UIViewController, ShareViewDelegate, SAEFailed
     }
 
     private func activate() {
+        SwiftAssertIsOnMainThread(#function)
```
charlesmchen-signal: The real purpose of this PR was to add these asserts in the SAE.

Signal/src/ViewControllers/CallViewController.swift
charlesmchen-signal: Well, the asserts compile out entirely in release builds thanks to the preprocessor.  And we use these asserts all over the place, so it's nice to eliminate them.

PR #3281, opened 2018-04-11
Reduce oversize text size.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSMessageSender.m
+++ b/SignalServiceKit/src/Messages/OWSMessageSender.m
@@ -48,7 +48,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -48,7 +48,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-const NSUInteger kOversizeTextMessageSizeThreshold = 16 * 1024;
+const NSUInteger kOversizeTextMessageSizeThreshold = 2 * 1024;
```
charlesmchen-signal: I'm confirming that this is the correct value.

charlesmchen-signal: Greyson has confirmed.

michaelkirk-signal:
LGTM

PR #3284, opened 2018-04-11
Use profile name in quoted messages, fix "multi account" label

michaelkirk-signal:
Turned out to not be *quite* a one line change =/

PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Contacts/SignalAccount.m
+++ b/SignalServiceKit/src/Contacts/SignalAccount.m
@@ -50,16 +50,9 @@ NS_ASSUME_NONNULL_BEGIN
     return _recipientId;
 }
 
-- (NSString *)displayName
+- (nullable NSString *)contactFullName
 {
-    NSString *baseName = (self.contact.fullName.length > 0 ? self.contact.fullName : self.recipientId);
-
-    OWSAssert(self.hasMultipleAccountContact == (self.multipleAccountLabelText != nil));
-    NSString *displayName = (self.multipleAccountLabelText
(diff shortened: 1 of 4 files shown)


SignalServiceKit/src/Contacts/SignalAccount.m
```diff
@@ -50,16 +50,9 @@ - (nullable NSString *)uniqueId
     return _recipientId;
 }
 
-- (NSString *)displayName
```
michaelkirk-signal: The only place we used this was when determining whether to show the profile name, so we should return nil if there is no local contact name.

SignalServiceKit/src/Contacts/SignalAccount.m
```diff
@@ -50,16 +50,9 @@ - (nullable NSString *)uniqueId
...
 {
-    NSString *baseName = (self.contact.fullName.length > 0 ? self.contact.fullName : self.recipientId);
-
-    OWSAssert(self.hasMultipleAccountContact == (self.multipleAccountLabelText != nil));
-    NSString *displayName = (self.multipleAccountLabelText
-            ? [NSString stringWithFormat:@"%@ (%@)", baseName, self.multipleAccountLabelText]
-            : baseName);
-
-    return displayName.filterStringForDisplay;
+    return self.contact.fullName.filterStringForDisplay;
```
charlesmchen-signal: FYI I like to check `length non-zero` of these filtered strings rather than nil.

michaelkirk-signal: ok - updated. PTAL @charlesmchen

PR #3289, opened 2018-04-13
Limit attachment caption length to 2k bytes

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/test/util/StringAdditionsTest.swift
+++ b/Signal/test/util/StringAdditionsTest.swift
@@ -1,0 +1,100 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import XCTest
+
+class StringAdditionsTest: XCTestCase {
+
+    override func setUp() {
+        super.setUp()
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+    }
+
+    override func tearDown() {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+        super.tearDown()
+    }
+
+    func testASCII() {
+        let originalString = "Hello World"
+
+        var truncatedString = originalString.truncated(toByteCount: 8)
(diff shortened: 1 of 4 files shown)


Signal/test/util/StringAdditionsTest.swift
```diff
@@ -0,0 +1,100 @@
...
+        XCTAssertEqual("Oh🇨🇦Canada", truncatedString)
+
+        truncatedString = originalString.truncated(toByteCount: 24)
+        XCTAssertEqual("Oh🇨🇦Canada🇨🇦", truncatedString)
+
+        truncatedString = originalString.truncated(toByteCount: 25)
+        XCTAssertEqual("Oh🇨🇦Canada🇨🇦", truncatedString)
+
+        truncatedString = originalString.truncated(toByteCount: 100)
+        XCTAssertEqual("Oh🇨🇦Canada🇨🇦", truncatedString)
```
charlesmchen-signal: Nice.

SignalMessaging/attachments/AttachmentApprovalViewController.swift
```diff
@@ -634,19 +634,19 @@ class CaptioningToolbar: UIView, UITextViewDelegate {
...
+            Logger.debug("\(self.logTag) in \(#function) long text was truncated")
             self.lengthLimitLabel.isHidden = false
+
             // Accept as much of the input as we can
-            let remainingSpace = maxCharacterCount - textView.text.count
-            if (remainingSpace) > 0 {
-                let acceptableAddition = text.substring(to: text.startIndex.advanced(by: remainingSpace))
-                textView.text = "\(textView.text ?? "")\(acceptableAddition)"
-                updateHeight(textView: textView)
+            let byteBudget = kOversizeTextMessageSizeThreshold + range.length - existingText.utf8.count
```
charlesmchen-signal: I think `range.length` is expressed in characters and the other two values are expressed in bytes.

PR #3290, opened 2018-04-13
Don't animate changes to shape layer properties.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleStrokeView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleStrokeView.m
@@ -87,7 +87,7 @@ NS_ASSUME_NONNULL_BEGIN
 
     // Prevent the shape layer from animating changes.
     [CATransaction begin];
-    [CATransaction setValue:(id)kCFBooleanTrue forKey:kCATransactionDisableActions];
+    [CATransaction setDisableActions:YES];
 
     // Don't fill the shape layer; we just want a stroke around the border.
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleStrokeView.m
```diff
@@ -85,19 +85,21 @@ - (void)updateLayers
         return;
     }
 
+    // Prevent the shape layer from animating changes.
+    [CATransaction begin];
+    [CATransaction setValue:(id)kCFBooleanTrue forKey:kCATransactionDisableActions];
```
michaelkirk-signal: nit here and elsewhere: `[CATransaction setDisableActions:YES]`

Signal/src/views/AudioProgressView.swift
```diff
@@ -73,6 +73,10 @@ import SignalServiceKit
     internal func updateContent() {
         SwiftAssertIsOnMainThread(#function)
 
+        // Prevent the shape layer from animating changes.
+        CATransaction.begin()
+        CATransaction.setValue(kCFBooleanTrue, forKey: kCATransactionDisableActions)
+
```
michaelkirk-signal: Take it or leave it, but since the start and end of the transaction are getting pretty far apart, it might be more coherent to introduce a block based api like:

```
CATransaction.ows_withoutAnimation {

}
```

charlesmchen-signal: I'm going to leave it for now, but it's not a bad idea.  It might make more sense as a method on CALayer.

PR #3296, opened 2018-04-16
Fix animation glitch in blocking window.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/util/OWSScreenLockUI.m
+++ b/Signal/src/util/OWSScreenLockUI.m
@@ -514,16 +514,16 @@ const UIWindowLevel UIWindowLevel_Background = -1.f;
 
     UIView *rootView = self.screenBlockingViewController.view;
 
-    [NSLayoutConstraint deactivateConstraints:self.screenBlockingConstraints];
-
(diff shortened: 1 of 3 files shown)


Signal/src/util/OWSScreenLockUI.m
```diff
@@ -514,17 +514,17 @@ - (void)updateScreenBlockingWindow:(ScreenLockUIState)desiredUIState animated:(B
 
     UIView *rootView = self.screenBlockingViewController.view;
 
-    [NSLayoutConstraint deactivateConstraints:self.screenBlockingConstraints];
```
charlesmchen-signal: Don't deactivate constraints unless we're adding new ones.

SignalMessaging/categories/UIView+OWS.h
```diff
@@ -118,6 +118,8 @@ CGFloat ScaleFromIPhone5(CGFloat iPhone5Value);
 - (void)addRedBorderRecursively;
 
 #ifdef DEBUG
+- (void)logFrame;
+- (void)logFrameWithLabel:(NSString *)label;
```
charlesmchen-signal: Debugging tools; not part of the fix.

PR #3298, opened 2018-04-16
Charlesmchen/home and conversation view design tweaks

charlesmchen-signal:
PTAL @michaelkirk 

simulator screen shot - iphone x - 2018-04-16 at 12 02 44

simulator screen shot - iphone x - 2018-04-16 at 12 02 40

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewCell.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewCell.m
@@ -26,8 +26,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic) UILabel *dateTimeLabel;
 // The unread badge has a larger v-height than the other elements in its
 // row.  We don't want it to distort the v-alignment of the cell's labels
-// so we use a placeholder to reserve the correct width.
-@property (nonatomic) UIView *unreadPlaceholder;
+// so we use a container to reserve the correct width.
+@property (nonatomic) UIView *unreadBadgeContainer;
 @property (nonatomic) UIView *unreadBadge;
 @property (nonatomic) UILabel *unreadLabel;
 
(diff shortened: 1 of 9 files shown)


Signal/src/ViewControllers/HomeView/HomeViewCell.m
```diff
@@ -302,7 +313,7 @@ - (NSAttributedString *)attributedStringForDate:(nullable NSDate *)date
 
 - (UIFont *)dateTimeFont
 {
-    return [UIFont ows_dynamicTypeFootnoteFont];
+    return [UIFont ows_dynamicTypeFootnoteFont].ows_medium;
```
michaelkirk-signal: nit: This might be better renamed to ows_mediumWeight - in context it's not clear what it's doing.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
```diff
@@ -339,7 +350,17 @@ + (CGFloat)scaleValueWithDynamicType:(CGFloat)minValue
 
 + (CGFloat)rowHeight
 {
-    return 72;
+    // Scale the cell height using size of dynamic "body" type as a reference.
+    const CGFloat kReferenceFontSizeMin = 17.f;
+    const CGFloat kReferenceFontSizeMax = 23.f;
+    CGFloat referenceFontSize = UIFont.ows_dynamicTypeBodyFont.pointSize;
+    CGFloat alpha = CGFloatClamp01(CGFloatInverseLerp(referenceFontSize, kReferenceFontSizeMin, kReferenceFontSizeMax));
```
michaelkirk-signal: These helpers make surprisingly readable math. 👏

Signal/src/ViewControllers/HomeView/HomeViewCell.m
```diff
@@ -369,7 +390,7 @@ - (void)prepareForReuse
     self.thread = nil;
     self.contactsManager = nil;
 
-    [self.unreadBadge removeFromSuperview];
```
michaelkirk-signal: ~~Don't we still need to remove the unread badge~~

Never mind, this is correct, but I think calling this the `unreadBadgeContainer` instead of `unreadPlaceholder` better communicates what's going on here.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
charlesmchen-signal: Done.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
charlesmchen-signal: Done.

PR #3300, opened 2018-04-16
Charlesmchen/attachment type assets

charlesmchen-signal:
PTAL @michaelkirk 

img_0050


Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
```diff
@@ -118,62 +118,31 @@ - (void)createContents
...
-        [MIMETypeUtil fileExtensionForMIMEType:self.attachmentStream.contentType];
-    }
-    if (fileExtension.length < 1) {
-        fileExtension = NSLocalizedString(@"GENERIC_ATTACHMENT_DEFAULT_TYPE",
-            @"A default label for attachment whose file extension cannot be determined.");
-    }
-
-    UILabel *fileTypeLabel = [UILabel new];
-    fileTypeLabel.text = fileExtension.uppercaseString;
-    fileTypeLabel.textColor = iconColor;
```
michaelkirk-signal: So we no longer include the file extension in the file icon? (e.g. PDF, etc)

Signal/src/ViewControllers/ConversationView/Cells/OWSQuotedMessageView.m
```diff
@@ -315,6 +327,28 @@ - (nullable NSString *)fileTypeForSnippet
     return nil;
 }
 
+- (BOOL)isAudioAttachment
```
michaelkirk-signal: So we use the mic for audio files *and* voice notes?

Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
charlesmchen-signal: Yes. My understanding is that Myles is already working on a new design for this but its not ready yet.

Signal/src/ViewControllers/ConversationView/Cells/OWSQuotedMessageView.m
charlesmchen-signal: Yes. We can't distinguish them here, right?

PR #3305, opened 2018-04-17
Make screen block view first responder.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/util/OWSScreenLockUI.m
+++ b/Signal/src/util/OWSScreenLockUI.m
@@ -438,6 +438,7 @@ const UIWindowLevel UIWindowLevel_Background = -1.f;
         // Show the blocking window in front of the status bar.
         self.screenBlockingWindow.windowLevel = UIWindowLevelStatusBar + 1;
         self.rootWindow.hidden = YES;
+        [self.screenBlockingViewController becomeFirstResponder];
     } else {
         self.screenBlockingWindow.windowLevel = UIWindowLevel_Background;
         [self.rootWindow makeKeyAndVisible];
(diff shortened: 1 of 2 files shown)


Signal/src/util/OWSScreenLockUI.m
```diff
@@ -32,6 +32,39 @@ typedef NS_ENUM(NSUInteger, ScreenLockUIState) {
 
 const UIWindowLevel UIWindowLevel_Background = -1.f;
 
+// OWSScreenBlockingWindow tries to become first responder
+// when presented to ensure that the input accessory is updated.
+@interface OWSScreenBlockingViewController : UIViewController
```
charlesmchen-signal: I'm not sure whether we should merge this PR.  It may not be necessary, and it doesn't fix a known repro.  

However, we've recently learned (in the case of call view) that a reliable way to update the input accessory is to switch first responders - not just `resignFirstResponder` but to ensure that there is a new first responder.

So I've opened this PR in the spirit of being defensive.

charlesmchen-signal:
@michaelkirk Good questions.  I've made it explicit that we only try to make the blocking screen view first responder when we present the block screen.  

Also, I propose we merge this _after_ the v2.24 release.

Signal/src/util/OWSScreenLockUI.m
```diff
@@ -534,6 +520,7 @@ - (void)updateScreenBlockingWindow:(ScreenLockUIState)desiredUIState animated:(B
         // Show the blocking window in front of the status bar.
         self.screenBlockingWindow.windowLevel = UIWindowLevelStatusBar + 1;
         self.rootWindow.hidden = YES;
+        [self.screenBlockingViewController becomeFirstResponder];
```
michaelkirk-signal: is the screenblockingViewController capable of becoming first responder at this time? e.g. I don't think a not-yet presented view controller can become first responder.

michaelkirk-signal: We could just add an assert that verifies this does what you're intending.

charlesmchen-signal: Done.

PR #3306, opened 2018-04-17
Charlesmchen/fix date comparators

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/util/DateUtil.m
+++ b/Signal/src/util/DateUtil.m
@@ -78,11 +78,8 @@ static NSString *const DATE_FORMAT_WEEKDAY = @"EEEE";
 
 + (BOOL)dateIsOlderThanToday:(NSDate *)date now:(NSDate *)now
 {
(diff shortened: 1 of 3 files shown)


Signal/src/util/DateUtil.m
```diff
@@ -78,11 +78,8 @@ + (BOOL)dateIsOlderThanToday:(NSDate *)date
 
 + (BOOL)dateIsOlderThanToday:(NSDate *)date now:(NSDate *)now
 {
-    NSCalendar *calendar = [NSCalendar currentCalendar];
-
-    NSUInteger dateDayOfEra = [calendar ordinalityOfUnit:NSCalendarUnitDay inUnit:NSCalendarUnitEra forDate:date];
```
charlesmchen-signal: `[NSCalendar  ordinalityOfUnit: inUnit:]` ignores timezone when the ordinality is "era".

PR #3309, opened 2018-04-17
Fix taps in message cells.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -67,10 +67,6 @@ NS_ASSUME_NONNULL_BEGIN
     [self addSubview:self.bubbleView];
     [self.bubbleView autoPinEdgesToSuperviewEdges];
 
-    UITapGestureRecognizer *tap =
-        [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(handleTapGesture:)];
-    [self addGestureRecognizer:tap];
-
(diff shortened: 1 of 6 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -67,10 +67,6 @@ - (void)commontInit
     [self addSubview:self.bubbleView];
     [self.bubbleView autoPinEdgesToSuperviewEdges];
 
-    UITapGestureRecognizer *tap =
```
charlesmchen-signal: This fixes the "can't tap on status message / send failure indicator" bug.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2163,7 +2163,7 @@ - (void)didTapConversationItem:(ConversationViewItem *)viewItem
     }];
 }
 
-- (void)didTapConversationItem:(ConversationViewItem *)viewItem quotedMessage:(OWSQuotedReplyModel *)quotedReply
+- (void)didTapConversationItem:(ConversationViewItem *)viewItem quotedReply:(OWSQuotedReplyModel *)quotedReply
```
charlesmchen-signal: This was causing crashes when tapping on quoted replies.

michaelkirk-signal: Death by rename. Thanks for finding this.

PR #3310, opened 2018-04-17
Tweak call view presentation.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/UIApplication+OWS.swift
+++ b/Signal/src/UIApplication+OWS.swift
@@ -18,7 +18,6 @@ extension UIApplication {
         guard let window = CurrentAppContext().mainWindow else {
             return nil
         }
-        Logger.error("findFrontmostViewController: \(window)")
         guard let viewController = window.rootViewController else {
             owsFail("\(self.logTag) in \(#function) Missing root view controller.")
             return nil
--- a/Signal/src/call/UserInterface/CallUIAdapter.swift
+++ b/Signal/src/call/UserInterface/CallUIAdapter.swift
@@ -39,15 +39,22 @@ extension CallUIAdaptee {
         let callViewController = CallViewController(call: call)
         callViewController.modalTransitionStyle = .crossDissolve
 
-        guard let presentingViewController = UIApplication.shared.frontmostViewControllerIgnoringAlerts else {
+        guard let mainWindow = CurrentAppContext().mainWindow else {
+            owsFail("in \(#function) mainWindow unexpectedly nil")
+            return
+        }
+        guard let presentingViewController = mainWindow.rootViewController else {
             owsFail("in \(#function) view controller unexpectedly nil")
             return
         }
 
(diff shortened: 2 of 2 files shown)


Signal/src/UIApplication+OWS.swift
```diff
@@ -18,7 +18,6 @@ extension UIApplication {
         guard let window = CurrentAppContext().mainWindow else {
             return nil
         }
-        Logger.error("findFrontmostViewController: \(window)")
```
charlesmchen-signal: Stale logging.

Signal/src/call/UserInterface/CallUIAdapter.swift
```diff
@@ -39,15 +39,22 @@ extension CallUIAdaptee {
         let callViewController = CallViewController(call: call)
         callViewController.modalTransitionStyle = .crossDissolve
 
-        guard let presentingViewController = UIApplication.shared.frontmostViewControllerIgnoringAlerts else {
+        guard let mainWindow = CurrentAppContext().mainWindow else {
+            owsFail("in \(#function) mainWindow unexpectedly nil")
+            return
+        }
+        guard let presentingViewController = mainWindow.rootViewController else {
```
charlesmchen-signal: Per our discussion, this should be fine - and make us robust to our presenting VC being dismissed by other logic.

Signal/src/call/UserInterface/CallUIAdapter.swift
```diff
@@ -39,15 +39,22 @@ extension CallUIAdaptee {
...
             owsFail("in \(#function) view controller unexpectedly nil")
             return
         }
 
         if let presentedViewController = presentingViewController.presentedViewController {
-            presentedViewController.dismiss(animated: false)
+            presentedViewController.dismiss(animated: false) {
```
charlesmchen-signal: This is a theory without proof/a repro.  I'm pretty sure that it's unsafe to assume `dismiss` finishes its work synchronously, even when animated is false, hence its completion handler.  So let's use the completion handler instead.

Signal/src/call/UserInterface/CallUIAdapter.swift
michaelkirk-signal: I don't understand what we get from this portion. How does presenting off the rootViewController vs. the frontMostVC make us more robust? We can still be dismissed by any similar logic in the app.

e.g. here how we' "dismiss whatever is presented and present this VC instead", hopefully we're not doing that anywhere else 😬 

Also seems like a minor UX degredation - previously after ending the call, the user would be back where they left off, right? But now we're always returning them to the home view before presenting the call screen.

Does this work if there are multiple VC's presented atop the rootView? seems like we want something like an "unwind" segue.

I'm left feeling like this portion of the changes asks more questions than it answers. I'm OK with working through the questions if you want to, but otherwise I'm inclined to just merge the second half of the PR.

The other changes in this PR LGTM

PR #3311, opened 2018-04-17
Capture and honor intended recipients for outgoing messages.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessageDetailViewController.swift
+++ b/Signal/src/ViewControllers/MessageDetailViewController.swift
@@ -212,7 +212,16 @@ class MessageDetailViewController: OWSViewController, MediaGalleryDataSourceDele
                     groupRows.append(divider)
                 }
 
-                for recipientId in thread.recipientIdentifiers {
+                var messageRecipientIds: [String]
+                if let intendedRecipientIds = outgoingMessage.intendedRecipientIds {
+                    // New messages capture the list of intended recipients.
+                    messageRecipientIds = Array(intendedRecipientIds)
+                } else {
+                    // For legacy messages, just use the list of members of the thread.
+                    messageRecipientIds = thread.recipientIdentifiers
+                }
+
+                for recipientId in messageRecipientIds {
                     let (recipientStatus, shortStatusMessage, _) = MessageRecipientStatusUtils.recipientStatusAndStatusMessage(outgoingMessage: outgoingMessage, recipientId: recipientId, referenceView: self.view)
 
                     guard recipientStatus == recipientStatusGroup else {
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -212,7 +212,16 @@ class MessageDetailViewController: OWSViewController, MediaGalleryDataSourceDele
...
 
-                for recipientId in thread.recipientIdentifiers {
+                var messageRecipientIds: [String]
+                if let intendedRecipientIds = outgoingMessage.intendedRecipientIds {
+                    // New messages capture the list of intended recipients.
+                    messageRecipientIds = Array(intendedRecipientIds)
+                } else {
+                    // For legacy messages, just use the list of members of the thread.
+                    messageRecipientIds = thread.recipientIdentifiers
+                }
```
charlesmchen-signal: First fix: in message details view, only show the "intended recipients" which are captured the first time we try to send the message.

SignalServiceKit/src/Messages/DeviceSyncing/OWSOutgoingSentMessageTranscript.m
```diff
@@ -48,15 +49,21 @@ - (OWSSignalServiceProtosSyncMessageBuilder *)syncMessageBuilder
...
+    [sentBuilder setDestination:self.recipientIdentifierForMessage];
+    [sentBuilder setMessage:[self.message buildDataMessage:self.recipientIdentifierForMessage]];
     [sentBuilder setExpirationStartTimestamp:self.message.timestamp];
 
     [syncMessageBuilder setSentBuilder:sentBuilder];
 
     return syncMessageBuilder;
 }
 
+// Note that this will return nil for group messages.
```
charlesmchen-signal: This isn't a change in behavior, I've just pulled this method from `TSOutgoingMessage` to here, the only place we use it.

I'm going to make a card to investigate this.  It might be correct behavior, but it's strange that.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -450,11 +463,14 @@ - (void)sendMessageToService:(TSOutgoingMessage *)message
...
 
         if ([thread isKindOfClass:[TSGroupThread class]]) {
+            OWSAssert(message.intendedRecipientIds);
+
             TSGroupThread *gThread = (TSGroupThread *)thread;
 
             NSError *error;
-            NSArray<SignalRecipient *> *recipients =
-                [self getRecipients:gThread.groupModel.groupMemberIds error:&error];
+            NSArray<SignalRecipient *> *recipients = [self getRecipients:gThread.groupModel.groupMemberIds
```
charlesmchen-signal: If we re-send a failed message, only send to the intersection of:

* The original intended recipients.
* The current members of the thread.

PR #3312, opened 2018-04-17
Simplify usage of contacts framework.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -44,7 +44,6 @@
 #import "UIViewController+Permissions.h"
 #import "ViewControllerUtils.h"
 #import <AVFoundation/AVFoundation.h>
-#import <AddressBookUI/AddressBookUI.h>
 #import <AssetsLibrary/AssetsLibrary.h>
 #import <ContactsUI/CNContactViewController.h>
 #import <JSQMessagesViewController/JSQMessagesBubbleImage.h>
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -44,7 +44,6 @@
 #import "UIViewController+Permissions.h"
 #import "ViewControllerUtils.h"
 #import <AVFoundation/AVFoundation.h>
-#import <AddressBookUI/AddressBookUI.h>
```
charlesmchen-signal: It'd be nice to remove all dependency on `AddressBookUI`.  This gets us close.

SignalMessaging/contacts/OWSContactsManager.m
```diff
@@ -19,6 +19,7 @@
 #import <SignalServiceKit/PhoneNumber.h>
 #import <SignalServiceKit/SignalAccount.h>
 
+@import AddressBook;
```
charlesmchen-signal: We still use `AddressBook` here to deal with sort order, format order.

SignalServiceKit/src/Contacts/Contact.m
```diff
@@ -65,6 +40,7 @@ - (instancetype)initWithSystemContact:(CNContact *)contact
     _cnContact = contact;
     _firstName = [self trimName:contact.givenName];
     _lastName = [self trimName:contact.familyName];
+    _fullName = [CNContactFormatter stringFromContact:contact style:CNContactFormatterStyleFullName];
```
charlesmchen-signal: We need to test this.  Let's discuss what the complete set of test scenarios should be.

PR #3314, opened 2018-04-18
Charlesmchen/call view optionals

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/CallViewController.swift
+++ b/Signal/src/ViewControllers/CallViewController.swift
@@ -30,7 +30,7 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver, 
     // MARK: Views
 
     var hasConstraints = false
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/CallViewController.swift
```diff
@@ -819,18 +855,18 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver,
     //   before the operation completes.
     func updateAudioSourceButtonIsSelected() {
         guard callUIAdapter.audioService.isSpeakerphoneEnabled else {
-            self.audioSourceButton.isSelected = false
+            self.audioSourceButton?.isSelected = false
```
charlesmchen-signal: This could crash.

charlesmchen-signal:
PTAL @michaelkirk I significantly simplified this PR.

PR #3317, opened 2018-04-19
surface call reconnecting

michaelkirk-signal:
PTAL @charlesmchen 

Video/screenshots sent out of band.

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -547,7 +547,7 @@ protocol CallServiceObserver: class {
                     // If both users are trying to call each other at the same time,
                     // both should see busy.
                     handleRemoteBusy(thread: existingCall.thread, callId: existingCall.signalingId)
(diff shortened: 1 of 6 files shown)


Signal/src/call/CallService.swift
```diff
@@ -327,8 +327,9 @@ protocol CallServiceObserver: class {
 
             // Don't let the outgoing call ring forever. We don't support inbound ringing forever anyway.
             let timeout: Promise<Void> = after(interval: connectingTimeoutSeconds).then { () -> Void in
-                // rejecting a promise by throwing is safely a no-op if the promise has already been fulfilled
-                OWSProdInfo(OWSAnalyticsEvents.callServiceErrorTimeoutWhileConnectingOutgoing(), file: #file, function: #function, line: #line)
```
michaelkirk-signal: As a separate change, I've moved this logging so it only occurs when relevant and updated the comment.

PR #3322, opened 2018-04-20
no reply to disappearing messages

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2317,6 +2317,11 @@ typedef enum : NSUInteger {
         return;
     }
 
+    if (message.isExpiringMessage) {
+        OWSFail(@"%@ unexpected expiring reply message: %@", self.logTag, message);
+        return;
+    }
+
     __block OWSQuotedReplyModel *quotedReply;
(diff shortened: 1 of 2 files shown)


SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -286,6 +286,11 @@ - (void)enqueueMessage:(TSOutgoingMessage *)message
         OWSAssert([message.body lengthOfBytesUsingEncoding:NSUTF8StringEncoding] <= kOversizeTextMessageSizeThreshold);
     }
 
+    if (message.isExpiringMessage && message.quotedMessage != nil) {
+        failureHandler(OWSErrorMakeAssertionError(@"expiring message unexpectedly had a quotedMessage"));
```
charlesmchen-signal: I don't think this part is right.

* Alice talking to Bob.
* Conversation has disappearing messages off.
* Alice says "hi".
* Bob turns disappearing messages on.
* Bob responds to "hi" with "hello!".  

It's okay for "hello" (expiring) to quote "hi" (non-expiring).

michaelkirk-signal: good catch, fixed.

PR #3324, opened 2018-04-22
unjank homescreen

michaelkirk-signal:
Some changes to make scrolling the home screen smoother.

Esentially:

1. Get all data needed for the homeview via explicit db transactions, so we can be sure we're getting it via the uiDBConnection (as opposed to the shared/sometimes-blocked db connections).

2. Move all those db calls onto a new "ThreadModel" view model. It has everything we need to render a HomeViewCell already loaded.

2. a simple cache scheme for these ThreadModel view models

PTAL @charlesmchen

### DIFF
--- a/Signal/src/Models/ThreadModel.swift
+++ b/Signal/src/Models/ThreadModel.swift
@@ -1,0 +1,101 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+@objc
+public class ThreadModel: NSObject {
+    let hasUnreadMessages: Bool
+    let lastMessageDate: Date
+    let isGroupThread: Bool
+    let threadRecord: TSThread
+    let unreadCount: UInt
+    let contactIdentifier: String?
+    let name: String
+    let isMuted: Bool
+    var isContactThread: Bool {
+        return !isGroupThread
+    }
+
+    let lastMessageText: String?
+
+//    func attributedSnippet(blockedPhoneNumberSet: Set<String>) {
+//        let isBlocked: Bool = {
+//            guard let contactIdentifier = self.contactIdentifier else {
+//                return false
+//            }
+//            assert(isContactThread)
+//            return blockedPhoneNumberSet.contains(self.contactIdentifier)
+//        }()
+//            
+//        
+//            
+////            BOOL hasUnreadMessages = thread.hasUnreadMessages;
+//        
+////            NSMutableAttributedString *snippetText = [NSMutableAttributedString new];
+//        var snippetText = NSMutableAttributedString()
+//        if isBlocked {
+//            // If thread is blocked, don't show a snippet or mute status.
+//            let append = NSAttributedString(string: NSLocalizedString("HOME_VIEW_BLOCKED_CONTACT_CONVERSATION",
+//                                                                      comment: "A label for conversations with blocked users."),
+//                attributes: <#T##[String : Any]?#>)
+//            
+////            if (isBlocked) {
+////                // If thread is blocked, don't show a snippet or mute status.
+////                [snippetText
+////                    appendAttributedString:[[NSAttributedString alloc]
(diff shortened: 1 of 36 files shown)


Signal/src/Models/ThreadModel.swift
```diff
@@ -0,0 +1,40 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+@objc
+public class ThreadModel: NSObject {
```
charlesmchen-signal: Shouldn't we call this `ThreadViewModel`?

Signal/src/ViewControllers/ConversationView/Cells/OWSSystemMessageCell.m
```diff
@@ -266,7 +271,11 @@ - (CGSize)cellSizeForViewWidth:(int)viewWidth contentWidth:(int)contentWidth
     result.height += self.topVMargin;
     result.height += self.bottomVMargin;
 
-    [self applyTitleForInteraction:interaction label:self.titleLabel];
+    // FIXME pass in transaction from the uiDBConnection.
```
charlesmchen-signal: Are you going to do this for the same release (v2.25?)?

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -47,7 +47,7 @@ @interface HomeViewController () <UITableViewDelegate, UITableViewDataSource, UI
 @property (nonatomic) UISegmentedControl *segmentedControl;
 @property (nonatomic) id previewingContext;
 @property (nonatomic) NSSet<NSString *> *blockedPhoneNumberSet;
-
+@property (nonatomic, readonly) NSCache<NSString *, ThreadModel *> *threadModelCache;
```
charlesmchen-signal: `ThreadModel` seems pretty cheap to instantiate, but I wonder if we should keep around the complete list of `ThreadModel` instances for the current load window, just like conversation view.

SignalServiceKit/src/Contacts/Threads/TSContactThread.m
```diff
@@ -114,6 +114,7 @@ - (BOOL)hasSafetyNumbers
     return !![[OWSIdentityManager sharedManager] identityKeyForRecipientId:self.contactIdentifier];
 }
 
+// TODO deprecate this? seems weird to access the displayName in the DB model
```
charlesmchen-signal: +1

Signal/src/Models/ThreadModel.swift
michaelkirk-signal: sure, done!

Signal/src/ViewControllers/ConversationView/Cells/OWSSystemMessageCell.m
michaelkirk-signal: If time allows it would be nice, but likely not for this release. To be clear, I don't think this a regression - we were previously creating a transaction here too, it's just that it was buried deep down in `applyTitleForInteraction:label` -> `description`.

`description` opened a transaction from the shared model.dBConnection not the uiDBConnection.

Instead we're now using `previewTextWithTransaction:` instead of `description` for any user facing text.

But yes! I'd like to do a similar audit of the conversation view and apply the same patterns. The existing ConversationViewItem (which maybe should be renamed to `InteractionViewModel` or `ConversationItemViewModel`) gets us part of the way there, but it still relies on self-created (non uiDBConnection transactions).

Though, I think it'll be a bigger endeavor, since the implicit-transaction pattern is used not just in the ConversationViewController, but also in the ConversationLayout classes (and maybe some other places).

Signal/src/ViewControllers/HomeView/HomeViewController.m
michaelkirk-signal: That would definitely be a nice future enhancement! 

Even with all the threadModel's cached, a significant portion of our remaining MainThread time is fetching threads from the Mapping, even though we only use them to get the cache key for the threadModel.

PR #3325, opened 2018-04-23
db view corruption

michaelkirk-signal:
PTAL @charlesmchen 

Rebuilding the messages view seems sufficient to recover the no-longer appearing message.

An iPhone5 with 100k messages takes 10mins to rebuild this view, so though it's faster than the SAE upgrade which rebuilt all views, it's still not cheap. It translates pretty linearly, so I'd expect most users one-time slow load to be a minute or less. 

e.g. an iPhone7 w/ 60k messages took 30s.

screenshot (not "Optimizing Database" is animated - flashing slowly)

screen shot 2018-04-23 at 12 54 22 pm

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -61,6 +61,8 @@ static NSString *const kInitialViewControllerIdentifier = @"UserInitialViewContr
 static NSString *const kURLSchemeSGNLKey                = @"sgnl";
 static NSString *const kURLHostVerifyPrefix             = @"verify";
 
+static NSTimeInterval launchStartedAt;
+
 @interface AppDelegate ()
 
 @property (nonatomic) BOOL hasInitialRootViewController;
@@ -103,6 +105,8 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
 
     // This should be the first thing we do.
     SetCurrentAppContext([MainAppContext new]);
+
+    launchStartedAt = CACurrentMediaTime();
 
     BOOL isLoggingEnabled;
 #ifdef DEBUG
@@ -171,8 +175,8 @@ static NSString *const kURLHostVerifyPrefix             = @"verify";
     UIWindow *mainWindow = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
     self.window = mainWindow;
     CurrentAppContext().mainWindow = mainWindow;
-    // Show the launch screen until the async database view registrations are complete.
-    mainWindow.rootViewController = [self loadingRootViewController];
+    // Show LoadingViewController until the async database view registrations are complete.
+    mainWindow.rootViewController = [LoadingViewController new];
     [mainWindow makeKeyAndVisible];
 
     // performUpdateCheck must be invoked after Environment has been initialized because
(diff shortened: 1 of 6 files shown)


Signal/src/AppDelegate.m
```diff
@@ -411,74 +413,6 @@ - (void)startupLogging
     DDLogInfo(@"iPhone Version: %@", platform);
 }
 
-- (UIViewController *)loadingRootViewController
```
michaelkirk-signal: pulled this into a class

Signal/src/ViewControllers/LoadingViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+    var bottomLabel: UILabel!
+
+    override public func loadView() {
+        self.view = UIView()
+        view.backgroundColor = UIColor.ows_materialBlue
+
+        self.logoView = UIImageView(image: #imageLiteral(resourceName: "logoSignal"))
+        view.addSubview(logoView)
+
+        logoView.autoCenterInSuperview()
```
michaelkirk-signal: Now that we're not using the actual Launch Screen, we can more easily add behavior.

However, it's now possible for the loading screen to look distinguishable from the Launch screen, which would be undesirable.

But it's such a simple view and we see it so often that I'm not very worried about breaking it.

SignalServiceKit/src/Storage/TSDatabaseView.m
```diff
@@ -162,7 +162,7 @@ + (void)asyncRegisterThreadInteractionsDatabaseView:(OWSStorage *)storage
 
     [self registerMessageDatabaseViewWithName:TSMessageDatabaseViewExtensionName
                                  viewGrouping:viewGrouping
-                                      version:@"1"
+                                      version:@"2"
```
michaelkirk-signal: The actual fix is simply incrementing the version.

michaelkirk-signal:
@charlesmchen - though we'll be merging https://github.com/signalapp/Signal-iOS/pull/3325, I think the changes to the loading screen still make sense - especially since there's no "app version" to check whether or not to show the "slow loading" UI when we rebuild a corrupted db view.

Should I pull out my changes to TSDatabaseView and rebase?

charlesmchen-signal:
> Should I pull out my changes to TSDatabaseView and rebase?

@michaelkirk  Please do!

Signal/src/ViewControllers/LoadingViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+        labelStack.setContentHuggingHigh()
+    }
+
+    var isShowingTopLabel = false
+    var isShowingBottomLabel = false
+    override public func viewDidAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+
+        // We only show the "loading" UI if it's a slow launch. Otherwise this ViewController
+        // shoudl be indistinguishable from the launch screen.
```
charlesmchen-signal: "shoudl"

Signal/src/ViewControllers/LoadingViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+        let kTopLabelThreshold: TimeInterval = 5
+        DispatchQueue.main.asyncAfter(deadline: .now() + kTopLabelThreshold) { [weak self] in
```
charlesmchen-signal: * 5 seconds seems long before showing this warning. I'd suggest 2?
* The reason I didn't use a timer before is because the main thread might be blocked on a db read/write or whatever.  It shouldn't be, but this seems like an unsafe assumption.

michaelkirk-signal: > 5 seconds seems long before showing this warning. I'd suggest 2?

I'm not confident that our normal startup is under 2 seconds for most devices. We definitely don't want to risk showing this every time someone launches. In fact, I actually had the opposite concern, that 5 seconds could be too soon for old slow devices.

> The reason I didn't use a timer before is because the main thread might be blocked on a db read/write or whatever. It shouldn't be, but this seems like an unsafe assumption.

I suppose that's true for any code that uses the main thread anywhere. Do you have a proposal?

michaelkirk-signal:
Some changes pulled into: https://github.com/signalapp/Signal-iOS/pull/3367

PR #3328, opened 2018-04-23
Charlesmchen/db migrations vs storage readiness

charlesmchen-signal:
PTAL @michaelkirk


SignalMessaging/environment/migrations/OWSDatabaseMigration.m
```diff
@@ -58,16 +58,6 @@ - (void)runUpWithCompletion:(OWSDatabaseMigrationCompletion)completion
     OWSAssert(completion);
 
     OWSDatabaseConnection *dbConnection = (OWSDatabaseConnection *)self.primaryStorage.newDatabaseConnection;
-    // These migrations won't be run until storage registrations are enqueued,
-    // but this transaction might begin before all registrations are marked as
-    // complete, so disable this checking.
-    //
-    // TODO: Once we move "app readiness" into AppSetup, we should explicitly
-    // not start these migrations until storage is ready.  We can then remove
-    // this statement which disables checking.
```
michaelkirk-signal: Thank you Matthew of yore for writing such a nice comment. =)

SignalShareExtension/ShareViewController.swift
```diff
@@ -90,18 +90,18 @@ public class ShareViewController: UIViewController, ShareViewDelegate, SAEFailed
         // We shouldn't set up our environment until after we've consulted isReadyForAppExtensions.
         AppSetup.setupEnvironment({
```
michaelkirk-signal: nit:  renaming method to `setupEnvironmentWithCallMessageHandlerBlock:` would make this more clear.

SignalShareExtension/ShareViewController.swift
```diff
@@ -90,18 +90,18 @@ public class ShareViewController: UIViewController, ShareViewDelegate, SAEFailed
...
+                                  notificationsProtocolBlock: {
             return NoopNotificationsManager()
-        }
+        },
+                                  migrationCompletion: { [weak self] in
+                                    SwiftAssertIsOnMainThread(#function)
 
-        // performUpdateCheck must be invoked after Environment has been initialized because
-        // upgrade process may depend on Environment.
-        VersionMigrations.performUpdateCheck(completion: { [weak self] in
```
michaelkirk-signal: nice that this was de-duped. 👍

SignalMessaging/environment/AppSetup.m
```diff
@@ -46,7 +51,18 @@ + (void)setupEnvironment:(CallMessageHandlerBlock)callMessageHandlerBlock
...
+                [VersionMigrations performUpdateCheckWithCompletion:^() {
+                    OWSAssertIsOnMainThread();
+
+                    migrationCompletion();
+
+                    backgroundTask = nil;
+                }];
+            });
+        }];
         [[Environment current].contactsManager startObserving];
```
michaelkirk-signal: Old code, but do we really want contactsManager to start observing here? Seems like we wouldn't want it to do anything until **after** the db is up and migrations have completed.

In this case, I think the consequences are benign because it just clears an entry from the avatar cache, but it would be *really* easy to inadvertently add some code which opens a write transaction into `ContactsManager.otherUsersProfileWillChange` without realizing it can be called before the app is ready.

SignalMessaging/environment/AppSetup.m
```diff
@@ -46,7 +51,18 @@ + (void)setupEnvironment:(CallMessageHandlerBlock)callMessageHandlerBlock
         [NSKeyedUnarchiver setClass:[OWSUserProfile class] forClassName:[OWSUserProfile collection]];
         [NSKeyedUnarchiver setClass:[OWSDatabaseMigration class] forClassName:[OWSDatabaseMigration collection]];
 
-        [OWSStorage setupStorage];
+        [OWSStorage setupStorageWithMigrationBlock:^() {
+            dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+                // Don't start database migrations until storage is ready.
+                [VersionMigrations performUpdateCheckWithCompletion:^() {
```
michaelkirk-signal: Hmmm... looks like this used to get called on the main thread. 

Are you backgrounding it because we don't want `runAllOutstandingWithCompletion` to block the main thread? That seems reasonable.

Since we're comparing state in `performUpdateCheckWithCompletion` (e.g. `AppVersion.instance.lastAppVersion` and friends), maybe it makes sense to continue to call `performUpdateCheckWithCompletion` on the main thread, but have `performUpdateCheckWithCompletion` do the background dispatch for the potentially blocking parts: `assumeAllExistingMigrationsRun`/`runAllOutstandingWithCompletion`

In practice, it looks like all the state we're accessing is thread safe anyway, so I don't see a problem with the change as written, just nitting over style. Take it or leave it.

SignalServiceKit/src/Storage/OWSStorage.m
```diff
@@ -332,29 +325,24 @@ - (void)runAsyncRegistrationsWithCompletion:(void (^_Nonnull)(void))completion
     OWS_ABSTRACT_METHOD();
 }
 
-+ (NSArray<OWSStorage *> *)allStorages
++ (void)setupStorageWithMigrationBlock:(OWSStorageMigrationBlock)migrationBlock
```
michaelkirk-signal: naming nit: `registerExtensions` might be more preices than `setupStorage`

SignalShareExtension/ShareViewController.swift
charlesmchen-signal: Good idea, done.

SignalMessaging/environment/AppSetup.m
charlesmchen-signal: Great point, moved to when app becomes ready.  And I'm _not_ going to start this in the SAE.

SignalMessaging/environment/AppSetup.m
charlesmchen-signal: Okay, moved this "dispatch off main thread" to just cover `runAllOutstandingWithCompletion`.  `assumeAllExistingMigrationsRun` is cheap enough not to matter.

SignalServiceKit/src/Storage/OWSStorage.m
charlesmchen-signal: Done.

charlesmchen-signal:
> One more opportunity for simplificiation, de-duplication, it seems like we could get rid of the storageIsReady checks in AppDelegate/ShareViewController.

Since this is harmless, I'm going to leave this in until we move this readiness logic into AppSetup.

PR #3333, opened 2018-04-24
Charlesmchen/input toolbar fix

charlesmchen-signal:
PTAL @michaelkirk 

This fixes a few more scenarios where it's really easy to get the input toolbar to reappear (present view X from conversation view, send app to background, return to app), but I can't help but feel that instead of trying to mop up N leaks we should fix some root cause elsewhere.  

Also, some manifestations of this problem can't be fixed. i.e. `UIAlertController` can't be subclassed.

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -117,6 +117,68 @@ typedef enum : NSUInteger {
 
 #pragma mark -
 
+@interface OWSImagePickerController : UIImagePickerController
+
+@end
+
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -117,6 +117,68 @@
 
 #pragma mark -
 
+@interface OWSImagePickerController : UIImagePickerController
```
charlesmchen-signal: I hesitate to make new source files and increase our build times.. But they're definitely clutter here.

PR #3334, opened 2018-04-25
Toggle rear camera in video call

michaelkirk-signal:
PTAL @charlesmchen 

I'm waiting on the final button asset, but that should be a trivial drop in.

### DIFF
--- a/Signal/src/ViewControllers/CallViewController.swift
+++ b/Signal/src/ViewControllers/CallViewController.swift
@@ -46,7 +46,10 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver, 
 
     // MARK: - Ongoing Call Controls
 
-    var ongoingCallView: UIView!
+    var ongoingCallControls: UIStackView!
+
+    var ongoingAudioCallControls: UIStackView!
+    var ongoingVideoCallControls: UIStackView!
 
     var hangUpButton: UIButton!
     var audioSourceButton: UIButton!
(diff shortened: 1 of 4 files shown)


charlesmchen-signal:
Screenshots?

Signal/src/ViewControllers/CallViewController.swift
```diff
@@ -345,42 +348,60 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver,
 
 //        textMessageButton = createButton(imageName:"message-active-wide",
 //                                                action:#selector(didPressTextMessage))
```
charlesmchen-signal: Remove?

Signal/src/ViewControllers/CallViewController.swift
```diff
@@ -345,42 +348,60 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver,
...
-        audioSourceButton = createButton(imageName: "audio-call-speaker-inactive",
+        audioSourceButton = createButton(image: #imageLiteral(resourceName: "audio-call-speaker-inactive"),
```
charlesmchen-signal: I'm not sure how I feel about these.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -220,6 +224,21 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
         self.videoSender = videoSender
     }
 
+    public func setCameraSource(useBackCamera: Bool) {
+        guard let localVideoSource = self.localVideoSource else {
+            owsFail("\(logTag) in \(#function) localVideoSource was unexpectedly nil")
+            return
+        }
+
+        // certain devices, e.g. 16GB iPod touch don't have a back camera
```
charlesmchen-signal: Should we add one to our test matrix?

Signal/src/ViewControllers/CallViewController.swift
michaelkirk-signal: It renders it in the IDE, which is cute, but not really that helpful. 

The main reason I like them is that they do a better job than logic like this:

```
 let image = UIImage(named: imageName)	
 assert(image != nil)
```

e.g. while developing you can be pretty sure you didn't mistype the name, and while running you know that it exists.

Signal/src/call/PeerConnectionClient.swift
michaelkirk-signal: Personally, I think this accounts for such a small user base that I'm not inclined to dedicate QA resources to verifying it.

michaelkirk-signal:
> Screenshots?

Whoops - I took a screenshot, but never posted it.

after_video_2

PR #3335, opened 2018-04-25
Show database upgrade warning while launching v2.25 or later.

charlesmchen-signal:
PTAL @michaelkirk


charlesmchen-signal:
> LGTM, though maybe obviated by the loading work in #3325

I think so.  This should be for the early access alpha.

PR #3336, opened 2018-04-25
Charlesmchen/dynamic type fixes

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSSystemMessageCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSSystemMessageCell.m
@@ -169,6 +169,9 @@ NS_ASSUME_NONNULL_BEGIN
     OWSAssert(interaction);
     OWSAssert(label);
 
+    // Update cell to reflect changes in dynamic text.
+    self.titleLabel.font = [self titleFont];
+
     // TODO: Should we move the copy generation into this view?
 
     if ([interaction isKindOfClass:[TSErrorMessage class]]) {
--- a/Signal/src/ViewControllers/HomeView/HomeViewCell.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewCell.m
@@ -62,7 +62,7 @@ NS_ASSUME_NONNULL_BEGIN
     [self setTranslatesAutoresizingMaskIntoConstraints:NO];
     self.layoutMargins = UIEdgeInsetsMake(0, self.cellHMargin, 0, self.cellHMargin);
     self.contentView.layoutMargins = UIEdgeInsetsZero;
-    self.contentView.preservesSuperviewLayoutMargins = YES;
+    self.contentView.preservesSuperviewLayoutMargins = NO;
 
     self.backgroundColor = [UIColor whiteColor];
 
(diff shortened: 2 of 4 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSSystemMessageCell.m
```diff
@@ -169,6 +169,9 @@ - (void)applyTitleForInteraction:(TSInteraction *)interaction
     OWSAssert(interaction);
     OWSAssert(label);
 
+    // Update cell to reflect changes in dynamic text.
+    self.titleLabel.font = [self titleFont];
```
charlesmchen-signal: We need to update all cell's fonts in their "configure" method in case dynamic type sizing has changed since this cell was created.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
```diff
@@ -84,8 +84,10 @@ - (void)commontInit
     [self.payloadView autoPinLeadingToTrailingEdgeOfView:self.avatarView offset:self.avatarHSpacing];
     [self.payloadView autoVCenterInSuperview];
     // Ensure that the cell's contents never overflow the cell bounds.
-    [self.payloadView autoPinEdgeToSuperviewMargin:ALEdgeTop relation:NSLayoutRelationGreaterThanOrEqual];
-    [self.payloadView autoPinEdgeToSuperviewMargin:ALEdgeBottom relation:NSLayoutRelationGreaterThanOrEqual];
+    //
+    // NOTE: It's critical that we pin to the superview top and bottom _edge_ and not _margin_.
```
charlesmchen-signal: Important

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -695,15 +695,19 @@ - (UITableViewCell *)cellForArchivedConversationsRow:(UITableView *)tableView
     UIStackView *stackView = [UIStackView new];
     stackView.axis = UILayoutConstraintAxisHorizontal;
     stackView.spacing = 5;
+    // If alignment isn't set, UIStackView uses the height of
+    // disclosureImageView, even if label has a higher desired height.
+    stackView.alignment = UIStackViewAlignmentCenter;
```
charlesmchen-signal: I'm still getting a feel for `UIStackView`, but this was very surprising.  I'm probably just going to get in the habit of always setting alignment to center (unless we need otherwise) to avoid this class of issue.  I don't understand why the default behavior is what it is.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
michaelkirk-signal: I don't understand this change  - Why would we ever want to be closer to the edge than the margin specifies?

This comment doesn't really clarify anything for me - can you use the comment to document *why* not just *what*?

Signal/src/ViewControllers/HomeView/HomeViewController.m
michaelkirk-signal: Yeah, that doesn't sound right. Are you sure that the stackView has enough space to grow vertically?

If we're pinned to the container, and the container isn't large enough, somethings got to give. And based on the hugging/compression configuration we're using that's going to be the label, not the image.

Signal/src/ViewControllers/ConversationView/Cells/OWSSystemMessageCell.m
michaelkirk-signal: 👍 Setting label fonts can be (just a little) expensive, so we might want to set only if not already equal.

PR #3340, opened 2018-04-26
Mkirk/profile pic in conversation

michaelkirk:
PTAL @charlesmchen 

iOS11

ios11-profile-name

iOS 10

ios10

iOS9

ios9


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -489,7 +487,16 @@ - (void)viewDidLoad
 
     [self createConversationScrollButtons];
     [self createHeaderViews];
-    [self createBackButton];
+
+    if (@available(iOS 11, *)) {
```
michaelkirk: This addresses a longstanding glitch in our interactive transitions on iOS11 vis a vis our custom back button.

SignalServiceKit/src/Storage/TSYapDatabaseObject.m
```diff
@@ -223,6 +223,17 @@ - (void)applyChangeToSelfAndLatestCopy:(YapDatabaseReadWriteTransaction *)transa
...
 
+- (void)reload
+{
+    TSYapDatabaseObject *latest = [[self class] fetchObjectWithUniqueID:self.uniqueId];
+    if (!latest) {
+        OWSFail(@"%@ in %s `latest` was unexpectedly nil", self.logTag, __PRETTY_FUNCTION__);
+        return;
+    }
+
+    [self setValuesForKeysWithDictionary:latest.dictionaryValue];
```
charlesmchen-signal: Huh.  Since we often put logic (e.g. migration logic) in `initWithCoder:`, I wonder if reloading this way will be always be safe.

SignalServiceKit/src/Contacts/Threads/TSGroupThread.m
```diff
@@ -12,6 +12,9 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+NSString *const TSGroupThreadAvatarChangedNotification = @"TSGroupThreadAvatarChangedNotification";
+NSString *const TSGroupThread_NotificaitonKey_UniqueId = @"TSGroupThread_NotificaitonKey_UniqueId";
```
charlesmchen-signal: /cation/

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -273,11 +273,17 @@ - (void)viewDidLoad
             self.title = NSLocalizedString(@"HOME_VIEW_TITLE_ARCHIVE", @"Title for the home view's 'archive' mode.");
             break;
     }
+
+    // We don't show any text for the back button, so there's no need to localize it. But because we left align the
+    // conversation title view, we add a little tappable padding after the back button, by having a title of spaces.
+    // Admittedly this is kind of a hack and not super fine grained, but it's simple and results in the interactive pop
+    // gesture animating our title view nicely vs. creating our own back button bar item with custom padding, which does
+    // not properly animate with the "swipe to go back" or "swipe left for info" gestures.
```
charlesmchen-signal: Nice

Signal/src/ViewControllers/ConversationView/ConversationHeaderView.swift
```diff
@@ -0,0 +1,120 @@
...
+
+import Foundation
+
+@objc
+public protocol ConversationHeaderViewDelegate {
+    func didTapConversationHeaderView(_ conversationHeaderView: ConversationHeaderView)
+}
+
+@objc
+public class ConversationHeaderView: UIStackView {
```
charlesmchen-signal: I'm surprised this works.  I thought it was unsafe to rely on iOS auto layout in navigation headers, and I presume UIStackView uses iOS auto layout internally.  

Let's test:

* On iOS 9, 10, 11.
* scenarios like leaving and returning to this view (by pushing convo settings, for example).
* Changes to content (group name length changes).

michaelkirk-signal: > I'm surprised this works. I thought it was unsafe to rely on iOS auto layout in navigation headers, and I presume UIStackView uses iOS auto layout internally.

Here's my understanding:

On older versions of iOS, items added to a navbar are in a weird situation - primarily what is their superview from which we can derive their frame? The inspector shows some privately generated thing that we never have direct access to, but apparently it doesn't play nice with autolayout constraints. 

On iOS11, instead of some weird private class, it's just a stack view.

So before iOS11, it was not possible to use autolayout to set the frame of the item we are adding to the navbar, but in iOS11 it is.

However, it was **always** safe (iOS9-11) to use autolayout to determine the frame of the **subviews**(in this case the two labels and the avatar) of the item added to the navbar.

This is why we manually set the frame size of the ConversationHeaderView on iOS9/10, but on iOS11 use automatic sizing (which the new iOS11 navbar stackview respects).

And since the item we add to the navbar now has a defined frame, its subviews are free to pin to their superview as we would elsewhere.

PR #3341, opened 2018-04-26
Fix call hangup crashes.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/PeerConnectionClient.swift
+++ b/Signal/src/call/PeerConnectionClient.swift
@@ -225,18 +225,22 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
     }
 
     public func setCameraSource(useBackCamera: Bool) {
-        guard let localVideoSource = self.localVideoSource else {
-            owsFail("\(logTag) in \(#function) localVideoSource was unexpectedly nil")
-            return
-        }
+        SwiftAssertIsOnMainThread(#function)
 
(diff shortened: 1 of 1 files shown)


Signal/src/call/PeerConnectionClient.swift
```diff
@@ -225,18 +225,22 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-            return
-        }
+        SwiftAssertIsOnMainThread(#function)
 
-        // certain devices, e.g. 16GB iPod touch don't have a back camera
-        guard localVideoSource.canUseBackCamera else {
-            owsFail("\(logTag) in \(#function) canUseBackCamera was unexpectedly false")
-            return
-        }
+        PeerConnectionClient.signalingQueue.async {
```
charlesmchen-signal: I think we're only supposed to access these properties on the signaling queue.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -525,6 +529,7 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
         audioSender = nil
         audioTrack = nil
         videoSender = nil
+        localVideoSource = nil
```
charlesmchen-signal: This is the fix. We have to eagerly clean up our WebRTC entities.

PR #3343, opened 2018-04-26
Fix issues in outgoing messages status changes.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
+++ b/SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
@@ -10,6 +10,7 @@
 #import "OWSSignalServiceProtos.pb.h"
 #import "ProtoBuf+OWS.h"
 #import "SignalRecipient.h"
+#import "TSAccountManager.h"
 #import "TSAttachmentStream.h"
 #import "TSContactThread.h"
 #import "TSGroupThread.h"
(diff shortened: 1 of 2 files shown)


SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
```diff
@@ -271,7 +272,16 @@ - (instancetype)initOutgoingMessageWithTimestamp:(uint64_t)timestamp
     // New outgoing messages should immediately determine their
     // recipient list from current thread state.
     NSMutableDictionary<NSString *, TSOutgoingMessageRecipientState *> *recipientStateMap = [NSMutableDictionary new];
-    NSArray<NSString *> *recipientIds = [thread recipientIdentifiers];
+    NSArray<NSString *> *recipientIds;
+    if ([self isKindOfClass:[OWSOutgoingSyncMessage class]]) {
+        NSString *_Nullable localNumber = [TSAccountManager localNumber];
+        OWSAssert(localNumber);
+        recipientIds = @[
+            localNumber,
```
charlesmchen-signal: We used to special-case "outgoing sync messages" in the message sender, but hopefully with this change we can do that less.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -456,7 +456,14 @@ - (void)sendMessageToService:(TSOutgoingMessage *)message
     dispatch_async([OWSDispatch sendingQueue], ^{
         TSThread *thread = message.thread;
 
-        if ([thread isKindOfClass:[TSGroupThread class]]) {
+        // TODO: It would be nice to combine the "contact" and "group" send logic here.
```
charlesmchen-signal: I'd really like to collapse how we handle these two kinds of messages.

SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
```diff
@@ -553,6 +563,11 @@ - (void)updateWithDeliveredRecipient:(NSString *)recipientId
     OWSAssert(recipientId.length > 0);
     OWSAssert(transaction);
 
+    // If delivery notification doesn't include timestamp, use "now" as an estimate.
```
michaelkirk-signal: When it's a client generated delivery receipt (which we aren't using, the client should specify), when it's a server generated delivery receipt, we use the timestamp on the envelope.

I can't think of a situation where this would happen and not be a bug, can we fail and log?

michaelkirk-signal: Sorry - I was confusing read receipts with delivery receipts. I believe this is correct as written.

PR #3344, opened 2018-04-26
Don't process multiple local notifications.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -69,6 +69,7 @@ static NSTimeInterval launchStartedAt;
 @property (nonatomic) BOOL hasInitialRootViewController;
 @property (nonatomic) BOOL areVersionMigrationsComplete;
 @property (nonatomic) BOOL didAppLaunchFail;
+@property (nonatomic) BOOL hasReceivedLocalNotification;
 
 @end
 
@@ -525,6 +526,9 @@ static NSTimeInterval launchStartedAt;
         [self handleActivation];
     }];
 
+    // We want to process up to one local notification per activation, so clear the flag.
+    self.hasReceivedLocalNotification = NO;
+
     DDLogInfo(@"%@ applicationDidBecomeActive completed.", self.logTag);
 }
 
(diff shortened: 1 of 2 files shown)


charlesmchen-signal:
PTAL @michaelkirk

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -914,6 +914,9 @@ - (void)presentThread:(TSThread *)thread
                 callOnViewAppearing:callOnViewAppearing];
         self.lastThread = thread;
 
+        // Never push more than one conversation view controller at a time.
```
michaelkirk-signal: Wouldn't this logic be better in `pushTopLevelViewController:`?

In fact, it looks like it already is:
https://github.com/signalapp/Signal-iOS/blob/master/Signal/src/ViewControllers/HomeView/HomeViewController.m#L980

Is it not working? Can we fix that instead of adding more code?

Signal/src/AppDelegate.m
```diff
@@ -592,6 +593,9 @@ - (void)applicationDidBecomeActive:(UIApplication *)application {
         [self handleActivation];
     }];
 
+    // We want to process up to one local notification per activation, so clear the flag.
```
michaelkirk-signal: Can you clarify for me - what problem are you trying to solve here (from the user's perspective) with this code?

Signal/src/ViewControllers/HomeView/HomeViewController.m
charlesmchen-signal: You're right, this part of the change looks redundant. Sorry, I don't have a repro so I can't test this.

Signal/src/AppDelegate.m
charlesmchen-signal: The symptom is: you launch Signal from a local notification for an incoming message.  

Expected: Signal "pops/closes" any presented views and pushes the conversation view for that message's thread - unless that thread is already "frontmost".  
Observed: Signal will do that "pop and present" logic N times - you enter convo view, pop to home view, enter convo view, return to home view, etc.  This issue is very hard to repro although I've seen it myself a couple times.  I decided to add a bunch of logging so that we could debug this without a repro and it paid off.  Riya repro'd and the logging clearly shows the app receiving multiple (identical) notifications (via `didReceiveLocalNotification`) during the activation process (after foreground, before active).  

Also: N can be quite high, like 5-10.  N seems vaguely related to their being multiple signal notifications, but the repro isn't that simple.

PR #3345, opened 2018-04-26
Fix global reference to 'inbox' home view.

charlesmchen-signal:
PTAL @michaelkirk


michaelkirk-signal:
> No biggie, but seems like this would be more coherent in the AppDelegate.

ehh... I guess we can also create a HomeView at the end of the registration flow, so we'd have to add similar code there. Probably not worth changing anything here.

charlesmchen-signal:
Yup, its untidy.

PR #3346, opened 2018-04-26
Charlesmchen/no archived threads

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -70,6 +70,8 @@ @interface HomeViewController () <UITableViewDelegate, UITableViewDataSource, UI
 
 @property (nonatomic) TSThread *lastThread;
 
+@property (nonatomic) BOOL hasArchivedThreadsRow;
```
charlesmchen-signal: Usually updates to our table view are tied to db change notifications. In the case of the "archived threads" row, this is a side effect of db changes but isn't captured in the mappings.

PR #3352, opened 2018-04-27
Rework teardown of PeerConnectionClient.

charlesmchen-signal:
I'm of two minds about this PR.  

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/PeerConnectionClient.swift
+++ b/Signal/src/call/PeerConnectionClient.swift
@@ -66,6 +66,28 @@ protocol PeerConnectionClientDelegate: class {
     func peerConnectionClient(_ peerconnectionClient: PeerConnectionClient, didUpdateRemote videoTrack: RTCVideoTrack?)
 }
 
+class AtomicFlag: NSObject {
+    private var flag = false
+
+    override required init() {
+        self.flag = false
+        super.init()
+    }
+
+    public func isSet() -> Bool {
+        objc_sync_enter(self)
+        let result = flag
+        objc_sync_exit(self)
+        return result
+    }
+
+    public func set() {
+        objc_sync_enter(self)
+        flag = true
+        objc_sync_exit(self)
+    }
+}
+
 /**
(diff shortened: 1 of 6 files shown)


Signal/src/call/PeerConnectionClient.swift
```diff
@@ -246,38 +280,65 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
     public func setLocalVideoEnabled(enabled: Bool) {
         SwiftAssertIsOnMainThread(#function)
 
-        PeerConnectionClient.signalingQueue.async {
-            guard self.peerConnection != nil else {
-                Logger.debug("\(self.TAG) \(#function) Ignoring obsolete event in terminated client")
+        // Define completion early to avoid capturing self within a block
+        // that uses weak self.
+        let completion = { [weak self] in
```
charlesmchen-signal: Change 3: if we enqueue a closure from an enqueued closure, we risk capturing self when the inner closure is defined.  This effectively captures self in the outer closure, even though we're capturing a weak refernce.  We can avoid this by defining our "completions" (as I'm calling these inner closures) outside of the outer closure.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -482,23 +596,37 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
         SwiftAssertIsOnMainThread(#function)
         Logger.debug("\(TAG) in \(#function)")
 
-        PeerConnectionClient.signalingQueue.async {
-            assert(self.peerConnection != nil)
-            self.terminateInternal()
+        // Any other work already enqueued on the signaling queue
+        // (and some work enqueued on the main queue)
+        // should abort once this flag is set.
+        isTerminated.set()
```
charlesmchen-signal: We set the terminated flag before enqueuing our termination logic.

PR #3354, opened 2018-04-27
Fix glitches in outgoing messages.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/DeviceSyncing/OWSOutgoingSentMessageTranscript.m
+++ b/SignalServiceKit/src/Messages/DeviceSyncing/OWSOutgoingSentMessageTranscript.m
@@ -25,6 +25,9 @@ NS_ASSUME_NONNULL_BEGIN
 @interface OWSOutgoingSentMessageTranscript ()
 
 @property (nonatomic, readonly) TSOutgoingMessage *message;
+// sentRecipientId is the recipient of message, for contact thread messages.
+// It is used to identify the thread/conversation to desktop.
+@property (nonatomic, readonly, nullable) NSString *sentRecipientId;
 
 @end
 
@@ -39,6 +42,8 @@ NS_ASSUME_NONNULL_BEGIN
     }
 
     _message = message;
+    // This will be nil for groups.
+    _sentRecipientId = message.thread.contactIdentifier;
 
     return self;
 }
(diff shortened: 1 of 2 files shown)


SignalServiceKit/src/Messages/DeviceSyncing/OWSOutgoingSentMessageTranscript.m
```diff
@@ -50,10 +55,8 @@ - (OWSSignalServiceProtosSyncMessageBuilder *)syncMessageBuilder
     OWSSignalServiceProtosSyncMessageSentBuilder *sentBuilder = [OWSSignalServiceProtosSyncMessageSentBuilder new];
     [sentBuilder setTimestamp:self.message.timestamp];
 
-    // Sync messages have no thread or destination.
-    OWSAssert(!self.message.thread.contactIdentifier);
-    [sentBuilder setDestination:nil];
-    [sentBuilder setMessage:[self.message buildDataMessage:nil]];
+    [sentBuilder setDestination:self.sentRecipientId];
+    [sentBuilder setMessage:[self.message buildDataMessage:self.sentRecipientId]];
```
charlesmchen-signal: This PR fixes two issues.  First "sent transcripts for outgoing messages in 1:1 messages" were broken. We need to include the "sent recipient id" in these transcripts.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -882,7 +882,10 @@ - (void)sendMessageToService:(TSOutgoingMessage *)message
 
             dispatch_async([OWSDispatch sendingQueue], ^{
                 // This emulates the completion logic of an actual successful save (see below).
-                [recipient save];
+                [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
+                    [message updateWithSkippedRecipient:localNumber transaction:transaction];
```
charlesmchen-signal: We need to mark the local recipient as skipped for sync message when we have no linked devices. They are indeed skipped.

SignalServiceKit/src/Messages/DeviceSyncing/OWSOutgoingSentMessageTranscript.m
```diff
@@ -39,6 +42,8 @@ - (instancetype)initWithOutgoingMessage:(TSOutgoingMessage *)message
     }
 
     _message = message;
+    // This will be nil for groups.
+    _sentRecipientId = message.thread.contactIdentifier;
```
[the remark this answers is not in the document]
michaelkirk-signal: I guess this is tricky, because we want it to be nil for group threads... we could pass in a transaction and do some more work, but we've had this logic for a long time, it just got shuffled a lot recently (starting with 9275c67818d4d7f98b82d193c60d8fca9c4208b0). 

Since we're effectively back to what we originally had with this field (though now we're caching it which is probably good), I'm fine with this change as is.

PR #3357, opened 2018-04-30
More outgoing message fixes.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Contacts/ContactsUpdater.m
+++ b/SignalServiceKit/src/Contacts/ContactsUpdater.m
@@ -40,6 +40,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (nullable SignalRecipient *)synchronousLookup:(NSString *)identifier error:(NSError **)error
 {
+    OWSAssert(error);
+
+    DDLogInfo(@"%@ %s %@", self.logTag, __PRETTY_FUNCTION__, identifier);
+
     dispatch_semaphore_t sema = dispatch_semaphore_create(0);
 
     __block SignalRecipient *recipient;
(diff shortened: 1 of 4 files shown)


SignalServiceKit/src/Contacts/ContactsUpdater.m
```diff
@@ -40,6 +40,10 @@ - (instancetype)init
 
 - (nullable SignalRecipient *)synchronousLookup:(NSString *)identifier error:(NSError **)error
```
charlesmchen-signal: I'd like to rework message sending significantly, and one goal would be to remove this method.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -423,7 +425,7 @@ - (void)enqueueAttachment:(DataSource *)dataSource
 {
     OWSAssert(error);
 
-    error = nil;
+    *error = nil;
```
charlesmchen-signal: Omitted `*` lead to NPE.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -186,7 +186,9 @@ - (void)run
 
 - (void)didSucceed
 {
-    OWSAssert(self.message.messageState == TSOutgoingMessageStateSent);
+    if (self.message.messageState != TSOutgoingMessageStateSent) {
+        OWSProdLogAndFail(@"%@ unexpected message status: %@", self.logTag, self.message.statusDescription);
```
charlesmchen-signal: There's a bug here that I haven't fixed yet. This logging should make fixing it a lot easer.

PR #3359, opened 2018-05-01
Charlesmchen/contact share message bubble

charlesmchen-signal:

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSContactShareView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSContactShareView.m
@@ -1,0 +1,178 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+#import "OWSContactShareView.h"
+#import "UIColor+JSQMessages.h"
+#import "UIColor+OWS.h"
+#import "UIFont+OWS.h"
+#import "UIView+OWS.h"
+#import <SignalServiceKit/OWSContactShare.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OWSContactShareView ()
+
+@property (nonatomic) OWSContactShare *contactShare;
+@property (nonatomic) NSString *contactShareName;
+@property (nonatomic) BOOL isIncoming;
+
+@end
+
+#pragma mark -
+
+@implementation OWSContactShareView
+
+- (instancetype)initWithContactShare:(OWSContactShare *)contactShare
+                    contactShareName:(NSString *)contactShareName
+                          isIncoming:(BOOL)isIncoming
+{
+    self = [super init];
+
+    if (self) {
+        _contactShare = contactShare;
+        _contactShareName = contactShareName;
+        _isIncoming = isIncoming;
+    }
+
+    return self;
+}
+
+#pragma mark - JSQMessageMediaData protocol
+
+- (CGFloat)iconHMargin
+{
+    return 12.f;
+}
+
+- (CGFloat)iconHSpacing
+{
+    return 10.f;
+}
+
++ (CGFloat)iconVMargin
+{
+    return 12.f;
+}
+
+- (CGFloat)iconVMargin
+{
+    return [OWSContactShareView iconVMargin];
+}
+
++ (CGFloat)bubbleHeight
+{
+    return self.iconSize + self.iconVMargin * 2;
+}
+
+- (CGFloat)bubbleHeight
+{
+    return [OWSContactShareView bubbleHeight];
+}
+
++ (CGFloat)iconSize
+{
+    return 44.f;
+}
+
+- (CGFloat)iconSize
+{
+    return [OWSContactShareView iconSize];
+}
+
+- (CGFloat)vMargin
+{
+    return 10.f;
+}
+
+- (UIColor *)bubbleBackgroundColor
+{
+    return self.isIncoming ? [UIColor jsq_messageBubbleLightGrayColor] : [UIColor ows_materialBlueColor];
+}
+
+- (void)createContents
+{
+    self.backgroundColor = [UIColor colorWithRGBHex:0xefeff4];
+    self.layoutMargins = UIEdgeInsetsZero;
+
+    // TODO: Verify that this layout works in RTL.
+    const CGFloat kBubbleTailWidth = 6.f;
+
+    UIView *contentView = [UIView containerView];
+    [self addSubview:contentView];
+    [contentView autoPinLeadingToSuperviewMarginWithInset:self.isIncoming ? kBubbleTailWidth : 0.f];
+    [contentView autoPinTrailingToSuperviewMarginWithInset:self.isIncoming ? 0.f : kBubbleTailWidth];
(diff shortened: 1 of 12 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSContactShareView.m
```diff
@@ -0,0 +1,174 @@
...
+{
+    return self.isIncoming ? [UIColor jsq_messageBubbleLightGrayColor] : [UIColor ows_materialBlueColor];
+}
+
+- (void)createContents
+{
+    self.backgroundColor = [UIColor colorWithRGBHex:0xefeff4];
+    self.layoutMargins = UIEdgeInsetsZero;
+
+    // TODO: Verify that this layout works in RTL.
```
charlesmchen-signal: The appearance of this view is very rough.

charlesmchen-signal:

PTAL @michaelkirk

Signal/src/ViewControllers/ConversationView/Cells/OWSContactShareView.m
```diff
@@ -0,0 +1,174 @@
...
+
+    UIStackView *stackView = [UIStackView new];
+    stackView.axis = UILayoutConstraintAxisHorizontal;
+    stackView.spacing = self.iconHSpacing;
+    stackView.alignment = UIStackViewAlignmentCenter;
+    [contentView addSubview:stackView];
+    [stackView autoPinLeadingToSuperviewMarginWithInset:self.iconHMargin];
+    [stackView autoPinTrailingToSuperviewMarginWithInset:self.iconHMargin];
+    [stackView autoVCenterInSuperview];
+    // NOTE: It's critical that we pin to the superview top and bottom _edge_ and not _margin_.
```
michaelkirk-signal: In the future can you try to focus comments more on *why* vs. *what* - the code should explain *what* adequately enough.

I don't understand why this is critical - in this case, the stack views superview is a container view, which has 0 margin, so I can't imagine why there'd even be a difference in pinning to edges vs margin.

If this is a no-op, we should get rid of it, IMO superstitious no-op code is just debt that distracts from the actual important logic.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.h
```diff
@@ -17,6 +17,7 @@ typedef NS_ENUM(NSInteger, OWSMessageCellType) {
     OWSMessageCellType_Video,
     OWSMessageCellType_GenericAttachment,
     OWSMessageCellType_DownloadingAttachment,
+    OWSMessageCellType_ShareContact,
```
michaelkirk-signal: naming nit (TIOLI): `ShareContact` sounds like a verb, e.g. I'd expect interacting with this cell would cause me to share out a contact.

How about `ContactShare` or just `Contact`

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -62,6 +65,7 @@ @interface ConversationViewItem ()
 @property (nonatomic, readonly, nullable) NSString *quotedRecipientId;
 @property (nonatomic, nullable) TSAttachmentStream *attachmentStream;
 @property (nonatomic, nullable) TSAttachmentPointer *attachmentPointer;
+@property (nonatomic, nullable) OWSContact *contactShare;
```
michaelkirk-signal: This is fine for now, but I'm guessing we'll eventually replace this DB record with a view model similar to the way we handled quoted replies, so we can fetch any contact attachment meta-data (TSAttachmentStream) as part of ConversationViewItem hydration.

SignalServiceKit/src/Messages/Interactions/OWSContact.m
```diff
@@ -165,6 +165,28 @@ - (BOOL)ows_isValid
...
+    if (_displayName.length < 1) {
+        CNContact *_Nullable systemContact = [OWSContacts systemContactForContact:self];
+        _displayName = [CNContactFormatter stringFromContact:systemContact style:CNContactFormatterStyleFullName];
+    }
+    if (_displayName.length < 1) {
+        // Fall back to using the organization name.
+        _displayName = self.organizationName;
+    }
+    if (_displayName.length < 1) {
+        DDLogError(@"%@ could not derive a valid display name.", self.logTag);
```
michaelkirk-signal: Should this be a `ProdLogAndFail`?

Signal/src/ViewControllers/ConversationView/Cells/OWSContactShareView.m
charlesmchen-signal: I've rewritten these comments.  The intent is for these overflow safe guards to be orthogonal to changes to the margins.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.h
charlesmchen-signal: Done.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
charlesmchen-signal: Yes, I've been wondering if that's going to be necessary.

SignalServiceKit/src/Messages/Interactions/OWSContact.m
charlesmchen-signal: Done.

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -748,7 +748,9 @@ - (UITableViewCell *)cellForArchivedConversationsRow:(UITableView *)tableView
     // Constrain to cell margins.
     [stackView autoPinEdgeToSuperviewMargin:ALEdgeLeading relation:NSLayoutRelationGreaterThanOrEqual];
     [stackView autoPinEdgeToSuperviewMargin:ALEdgeTrailing relation:NSLayoutRelationGreaterThanOrEqual];
-    // NOTE: It's critical that we pin to the superview top and bottom _edge_ and not _margin_.
+    // Ensure that the cell's contents never overflow the cell bounds.
+    // We pin pin to the superview _edge_ and not _margin_ for the purposes
+    // of overflow, so that changes to the margins do not trip these safe guards.
     [stackView autoPinEdgeToSuperviewEdge:ALEdgeTop withInset:0 relation:NSLayoutRelationGreaterThanOrEqual];
```
michaelkirk-signal: > // We pin pin to the superview _edge_ and not _margin_ for the purposes  
> // of overflow, so that changes to the margins do not trip these safe guards.

nit: typo (pin pin)

If we *did* add margins, the current implementation would allow the view that we're trying to constraint to overflow past the top/bottom margins, which seems undesirable. Pinning to the margins will make the view always at least as big as you have it now, but additionally enforce any desired margin from the edge.

So to me it still seems like pinning to edges here is worse than pinning to margins.

Admittedly this is all hypothetical with a 0 margin as you have here, but this code has now been duplicated into a second place, and presumably you'll continue to use it, so I'm trying to either:

1. convince you that this is wrong and you should stop using it.
- or - 
2. come to understand why this is better and start using this pattern myself.

PR #3360, opened 2018-05-01
Charlesmchen/contact share previews

charlesmchen-signal:

PTAL @michaelkirk


charlesmchen-signal:
> maybe we should move rtlSafeAppend into SSK.

I'm on the cusp of doing so.  I haven't yet because we're currently using 3 separate RTL detection mechanisms, one in `rtlSafeAppend`, and one in each of the `[AppContext isRTL]` implementations.  Moving to SSK would be a good time to consolidate, and I'd just like to take the time to make sure we're doing the right thing.

PR #3361, opened 2018-05-01
Rework contact normalization and validation.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/Interactions/OWSContact.h
+++ b/SignalServiceKit/src/Messages/Interactions/OWSContact.h
@@ -102,6 +102,8 @@ typedef NS_ENUM(NSUInteger, OWSContactAddressType) {
 
 - (instancetype)init NS_UNAVAILABLE;
 
+- (void)normalize;
+
 - (BOOL)ows_isValid;
 
 @end
--- a/SignalServiceKit/src/Messages/Interactions/OWSContact.m
+++ b/SignalServiceKit/src/Messages/Interactions/OWSContact.m
@@ -141,28 +141,49 @@ NS_ASSUME_NONNULL_BEGIN
 
 @implementation OWSContact
 
+- (void)normalize
+{
+    self.phoneNumbers = [self.phoneNumbers
+        filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(OWSContactPhoneNumber *value,
+                                        NSDictionary<NSString *, id> *_Nullable bindings) {
+            return value.ows_isValid;
+        }]];
+    self.emails = [self.emails filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(OWSContactEmail *value,
+                                                               NSDictionary<NSString *, id> *_Nullable bindings) {
+        return value.ows_isValid;
+    }]];
+    self.addresses =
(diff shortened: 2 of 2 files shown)


SignalServiceKit/src/Messages/Interactions/OWSContact.h
```diff
@@ -102,6 +102,8 @@ typedef NS_ENUM(NSUInteger, OWSContactAddressType) {
 
 - (instancetype)init NS_UNAVAILABLE;
 
+- (void)normalize;
```
charlesmchen-signal: I'm thinking we should be explicit about normalizing the data, e.g. do it explicitly during the "share contact" workflow.

SignalServiceKit/src/Messages/Interactions/OWSContact.m
```diff
@@ -141,28 +141,49 @@ @interface OWSContact ()
...
+        }]];
+}
+
 - (BOOL)ows_isValid
 {
     if (self.displayName.ows_stripped.length) {
         return NO;
     }
-    // Contact is valid if at least one entry is valid.
+    BOOL hasValue = NO;
```
charlesmchen-signal: Making this more strict upon reflect.

SignalServiceKit/src/Messages/Interactions/OWSContact.m
```diff
@@ -490,15 +480,11 @@ + (nullable OWSSignalServiceProtosDataMessageContact *)protoForContact:(OWSConta
...
-        if (!phoneNumber.ows_isValid) {
-            OWSProdLogAndFail(@"%@ phone number is invalid.", self.logTag);
-            continue;
-        }
         OWSSignalServiceProtosDataMessageContactPhoneBuilder *phoneBuilder =
             [OWSSignalServiceProtosDataMessageContactPhoneBuilder new];
         phoneBuilder.value = phoneNumber.phoneNumber;
-        if (phoneBuilder.label.ows_stripped.length > 0) {
-            phoneBuilder.label = phoneBuilder.label.ows_stripped;
+        if (phoneNumber.label.ows_stripped.length > 0) {
```
charlesmchen-signal: This code isn't tested or even exercised yet and I'm still working out the glitches.

SignalServiceKit/src/Messages/Interactions/OWSContact.m
michaelkirk-signal: Hmmm.... so if I have a contact entry for Bob which has 3 valid phone numbers and one invalid email, I can't share Bob's card?

I'd expect that I can share bob's card, but only the valid phone numbers are included.

TBH though, I'm not sure what we're getting from this validation logic. If I have something that our parser doesn't think is a valid phonenumber / address / email it seems like we'd still want to share whatever raw text the sender had, rather than just dropping it.

charlesmchen-signal:
> Hmmm.... so if I have a contact entry for Bob which has 3 valid phone numbers and one invalid email, I can't share Bob's card?

That's the intention of the `normalize` method.  Presumably we'd normalize, then validate.

charlesmchen-signal:
> TBH though, I'm not sure what we're getting from this validation logic. If I have something that our parser doesn't think is a valid phonenumber / address / email it seems like we'd still want to share whatever raw text the sender had, rather than just dropping it.

Totally up for discussion. Let's chat tomorrow.

PR #3362, opened 2018-05-01
remove canary assert

michaelkirk-signal:
WORKS GREAT

PTAL @charlesmchen


michaelkirk-signal:
In 6fcbf0529607f1aa4c481f52e12765c83c9df8b3

PR #3364, opened 2018-05-01
Contact picking

michaelkirk-signal:
For now this doesn't actually send a contact message, it just passes the picked contact back to the conversation view controller.

**New Contact Picker for contact shares**

picker-contact-send

**Regression check on old "invite flow"**

picker-invite

PTAL @charlesmchen


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3576,6 +3590,18 @@ - (void)attachmentButtonPressed
...
     [actionSheetController addAction:gifAction];
 
+    UIAlertAction *chooseContactAction = [UIAlertAction
+        actionWithTitle:NSLocalizedString(@"ATTACHMENT_MENU_CONTACT_BUTTON", @"attachment menu option to send contact")
+                  style:UIAlertActionStyleDefault
+                handler:^(UIAlertAction *_Nonnull action) {
+                    [self chooseContactForSending];
+                }];
+    // TODO - proper image
+    UIImage *chooseContactImage = [UIImage imageNamed:@"actionsheet_camera_black"];
```
charlesmchen-signal: I added this to the design qs doc.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4848,6 +4873,37 @@ - (void)collectionView:(UICollectionView *)collectionView
...
+{
+    DDLogDebug(@"%@ in %s with contact: %@", self.logTag, __PRETTY_FUNCTION__, contact);
+
+    // TODO actually build contact message.
+    self.inputToolbar.messageText = contact.fullName;
+}
+
+- (void)contactsPicker:(ContactsPicker *)contactsPicker didSelectMultipleContacts:(NSArray<Contact *> *)contacts
+{
+    DDLogDebug(@"%@ in %s with contacts: %@", self.logTag, __PRETTY_FUNCTION__, contacts);
```
charlesmchen-signal: Shouldn't this be `OWSFail()`?  And shouldn't we configure the contact picker to be "single only" for contact shares?

Signal/src/views/ContactCell.swift
```diff
@@ -39,53 +63,55 @@ class ContactCell: UITableViewCell {
     }
```
charlesmchen-signal: Can you please attach screenshots with dynamic text size on minimum, default and maximum?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: > Shouldn't this be OWSFail()

Yup, added.

> And shouldn't we configure the contact picker to be "single only" for contact shares?

Yup, it is configured via the initializer with the `multiSelection:NO` parameter.

PR #3366, opened 2018-05-02
Charlesmchen/contact view

charlesmchen-signal:
There's a bunch of pending work here, especially around unanswered design questions.

PTAL @michaelkirk 

simulator screen shot - iphone x - 2018-05-02 at 10 09 21
simulator screen shot - iphone x - 2018-05-02 at 10 09 16
simulator screen shot - iphone x - 2018-05-02 at 10 09 12


michaelkirk-signal:
@charlesmchen  - This seems to be based on https://github.com/signalapp/Signal-iOS/pull/3359, so I'll wait until that's merged before reviewing.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+//
+
+import Foundation
+import SignalServiceKit
+import SignalMessaging
+import Reachability
+import ContactsUI
+import MessageUI
+
+class TappableView: UIView {
```
charlesmchen-signal: I might pull this out into a separate source file.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+            addRow(createActionRow(labelText: NSLocalizedString("CONVERSATION_SETTINGS_NEW_CONTACT",
+                                                               comment: "Label for 'new contact' button in conversation settings view."),
+                                   action: #selector(didPressCreateNewContact)))
+
+            addRow(createActionRow(labelText: NSLocalizedString("CONVERSATION_SETTINGS_ADD_TO_EXISTING_CONTACT",
+                                                               comment: "Label for 'new contact' button in conversation settings view."),
+                                   action: #selector(didPressAddToExistingContact)))
+        }
+
+        // TODO: Not designed yet.
```
charlesmchen-signal: There's a bunch of TODOs here that correspond to cards or items in the "open design qs" doc.

Signal/src/ViewControllers/ConversationView/ConversationViewController.h
```diff
@@ -6,15 +6,20 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+typedef NS_ENUM(NSUInteger, ConversationViewAction) {
+    ConversationViewActionNone,
+    ConversationViewActionCompose,
+    ConversationViewActionAudioCall,
+    ConversationViewActionVideoCall,
```
charlesmchen-signal: I finally got around to cleaning up this "action on open" logic for conversation view.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+            strongSelf.tryToDetermineMode()
+        })
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        UIUtil.applySignalAppearence()
+
+        self.becomeFirstResponder()
```
michaelkirk-signal: Why are we becoming first responder here?

If it's because of https://github.com/signalapp/Signal-iOS/pull/3333, then I'd prefer we figure out whats going on rather than continuing to spread around this partial work-around, in case it has unintended consequences.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+
+        return phoneNumbersForContact().filter({ (phoneNumber) -> Bool in
+            return contactsManager.allContactsMap[phoneNumber] != nil
+        })
+    }
+
+    private func phoneNumbersForContact() -> [String] {
+        SwiftAssertIsOnMainThread(#function)
+
+        var result = [String]()
```
michaelkirk-signal: Why is `contact.phoneNumbers` nullable. Seems like it should always be an array, though it may be an empty array.

If we for some reason desire that these fields are sometimes nil, then this might be more to-the-point:

```
private func phoneNumbersForContact() -> [String] {
    return contact.phoneNumbers ?? []
}
```

But if we can make them non-nullable, then this just becomes
```
private func phoneNumbersForContact() -> [String] {
    return contact.phoneNumbers
}
```

At which point, I'd consider inlining `contact.phoneNumbers` and getting rid of this method all together.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+        if let phoneNumbers = contact.phoneNumbers {
+            for phoneNumber in phoneNumbers {
+                result.append(phoneNumber.phoneNumber)
+            }
+        }
+        return result
+    }
+
+    private func tryToDetermineModeRetry() {
```
michaelkirk-signal: This doesn't seem like it's ever called?

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+            subview.removeFromSuperview()
+        }
+
+        // TODO: The design calls for no navigation bar, just a back button.
+        let topView = UIView.container()
+        topView.backgroundColor = UIColor(rgbHex: 0xefeff4)
+        topView.preservesSuperviewLayoutMargins = true
+        rootView.addSubview(topView)
+        topView.autoPinEdge(toSuperviewEdge: .top)
+        topView.autoPinEdge(.left, to: .left, of: self.view, withOffset: 0)
```
michaelkirk-signal: Except for very rare circumstances, prefer .leading/.trailing instead of .left/.right.

I'd say any usage of left/right probably deserves a justifying comment.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+        nameLabel.font = UIFont.ows_dynamicTypeTitle2.ows_bold()
+        nameLabel.textColor = UIColor.black
+        nameLabel.lineBreakMode = .byTruncatingTail
+        nameLabel.textAlignment = .center
+        topView.addSubview(nameLabel)
+        nameLabel.autoPinEdge(.top, to: .bottom, of: avatarView, withOffset: 10)
+        nameLabel.autoPinLeadingToSuperviewMargin(withInset: hMargin)
+        nameLabel.autoPinTrailingToSuperviewMargin(withInset: hMargin)
+
+        var lastView: UIView = nameLabel
```
michaelkirk-signal: I find this `lastView` juggling, wherein it's a different value throughout the course of this layout pass to be a little confusing.

I guess we can't easily just use a stackview because we want custom spacing after each element...

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+        }
+
+        lastView.autoPinEdge(toSuperviewEdge: .bottom, withInset: 15)
+
+        let bottomView = UIView.container()
+        bottomView.backgroundColor = UIColor.white
+        bottomView.layoutMargins = .zero
+        bottomView.preservesSuperviewLayoutMargins = false
+        rootView.addSubview(bottomView)
+        bottomView.autoPinEdge(.top, to: .bottom, of: topView, withOffset: 0)
```
michaelkirk-signal: No biggie, but can't this just be `bottomView.autoPinEdge(.top, to: .bottom, of: topView)`

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+        let addSpacerRow = {
+            guard let prevRow = lastRow else {
+                owsFail("\(self.logTag) missing last row")
+                return
+            }
+            let row = UIView()
+            row.backgroundColor = UIColor(rgbHex: 0xdedee1)
+            bottomView.addSubview(row)
+            row.autoSetDimension(.height, toSize: 1)
+            row.autoPinLeadingToSuperviewMargin(withInset: self.hMargin)
```
michaelkirk-signal: No biggie, but since we're using this `hMargin` all over the place, did you consider defining an actual layout margin with edge insets and pinning directly to them rather than explicitly setting the offset for all these views? (and pinning to the Edge when you don't want the margin)

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+        // TODO: Should we present addresses here too? How?
+
+        lastRow?.autoPinEdge(toSuperviewEdge: .bottom, withInset: 0)
+    }
+
+    private let hMargin = CGFloat(16)
+
+    private func createActionRow(labelText: String, action: Selector) -> UIView {
+        let row = UIView()
+        row.layoutMargins.left = 0
```
michaelkirk-signal: e.g.
```
row.layoutMargins.left = hMargin
row.layoutMargins.right = hMargin
```

then below:
```
label.autoPinToSuperviewMargins()
```

TIOLI of course!

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+    func didPressInvite() {
+        Logger.info("\(self.TAG) \(#function)")
+
+        guard MFMessageComposeViewController.canSendText() else {
+            Logger.info("\(TAG) Device cannot send text")
+            OWSAlerts.showErrorAlert(message: NSLocalizedString("UNSUPPORTED_FEATURE_ERROR", comment: ""))
+            return
+        }
+        let phoneNumbers =
+        phoneNumbersForContact()
```
michaelkirk-signal: format

Signal/src/call/UserInterface/CallUIAdapter.swift
```diff
@@ -73,6 +73,7 @@ extension CallUIAdaptee {
         }
 
         let call = self.startOutgoingCall(handle: recipientId)
+        call.hasLocalVideo = hasLocalVideo
```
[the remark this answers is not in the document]
charlesmchen-signal: Making a card.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1024,15 +1013,22 @@ - (void)viewDidAppear:(BOOL)animated
...
-        [self callAction];
-        _callOnOpen = NO;
+    switch (self.actionOnOpen) {
+        case ConversationViewActionNone:
+            break;
+        case ConversationViewActionCompose:
+            [self popKeyBoard];
+            break;
+        case ConversationViewActionAudioCall:
+            [self audioCallAction];
```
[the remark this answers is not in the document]

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1354,7 +1350,17 @@ - (void)showFingerprintWithRecipientId:(NSString *)recipientId
 
 #pragma mark - Calls
 
-- (void)callAction
+- (void)audioCallAction
+{
+    [self callWithVideo:NO];
```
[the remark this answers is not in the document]
charlesmchen-signal: We still need proper method since they're bound to buttons, etc.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+        }
+        actionBlock()
+    }
+}
+
+// MARK: -
+
+class ContactViewController: OWSViewController, CNContactViewControllerDelegate {
+
+    let TAG = "[ContactView]"
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -0,0 +1,686 @@
...
+    }
+
+    required init(contact: OWSContact) {
+        contactsManager = Environment.current().contactsManager
+        self.contact = contact
+        self.scrollView = UIScrollView()
+
+        super.init(nibName: nil, bundle: nil)
+
+        tryToDetermineMode()
```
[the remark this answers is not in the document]
charlesmchen-signal: Done. This method used to be more complicated.

Signal/src/ViewControllers/ContactViewController.swift
charlesmchen-signal: Removed.

Signal/src/ViewControllers/ContactViewController.swift
charlesmchen-signal: Yep, I've been trying to use `UIStackView` where I can.

Signal/src/ViewControllers/ContactViewController.swift
charlesmchen-signal: Removed.

Signal/src/ViewControllers/ContactViewController.swift
charlesmchen-signal: Yup, I've been following your lead of trying to use layout margins more, but I haven't been strict about always going back and making sure we use them in every possible place.

Signal/src/ViewControllers/ContactViewController.swift
charlesmchen-signal: clang-format is really disappointing me lately.

PR #3367, opened 2018-05-02
loading screen

michaelkirk-signal:
Pulled out of https://github.com/signalapp/Signal-iOS/pull/3325

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/AppSettings/AboutTableViewController.m
+++ b/Signal/src/ViewControllers/AppSettings/AboutTableViewController.m
@@ -79,21 +79,43 @@
         attachmentCount = [transaction numberOfKeysInCollection:[TSAttachment collection]];
     }];
 
+    NSByteCountFormatter *byteCountFormatter = [NSByteCountFormatter new];
+
+    // format counts with thousands separator
+    NSNumberFormatter *numberFormatter = [NSNumberFormatter new];
+    numberFormatter.formatterBehavior = NSNumberFormatterBehavior10_4;
+    numberFormatter.numberStyle = NSNumberFormatterDecimalStyle;
+
     OWSTableSection *debugSection = [OWSTableSection new];
+
     debugSection.headerTitle = @"Debug";
-    [debugSection addItem:[OWSTableItem labelItemWithText:[NSString stringWithFormat:@"Threads: %zd", threadCount]]];
-    [debugSection addItem:[OWSTableItem labelItemWithText:[NSString stringWithFormat:@"Messages: %zd", messageCount]]];
+
+    NSString *formattedThreadCount = [numberFormatter stringFromNumber:@(threadCount)];
     [debugSection
-        addItem:[OWSTableItem labelItemWithText:[NSString stringWithFormat:@"Attachments: %zd", attachmentCount]]];
+        addItem:[OWSTableItem labelItemWithText:[NSString stringWithFormat:@"Threads: %@", formattedThreadCount]]];
+
+    NSString *formattedMessageCount = [numberFormatter stringFromNumber:@(messageCount)];
     [debugSection
-        addItem:[OWSTableItem labelItemWithText:[NSString stringWithFormat:@"Database size: %llu",
-                                                          [OWSPrimaryStorage.sharedManager databaseFileSize]]]];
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/AppSettings/AboutTableViewController.m
```diff
@@ -79,21 +79,43 @@ - (void)updateTableContents
         attachmentCount = [transaction numberOfKeysInCollection:[TSAttachment collection]];
     }];
 
+    NSByteCountFormatter *byteCountFormatter = [NSByteCountFormatter new];
```
charlesmchen-signal: Huh! Didn't know `NSByteCountFormatter` exists.

Signal/src/ViewControllers/LoadingViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+        labelStack.axis = .vertical
+        labelStack.alignment = .center
+        labelStack.spacing = 8
+        view.addSubview(labelStack)
+
+        labelStack.autoPinEdge(.top, to: .bottom, of: logoView, withOffset: 20)
+        labelStack.autoPinLeadingToSuperviewMargin()
+        labelStack.autoPinTrailingToSuperviewMargin()
+        labelStack.setCompressionResistanceHigh()
+        labelStack.setContentHuggingHigh()
```
charlesmchen-signal: We should talk about layout sometime.  I've stopped using the "low" flavors of these methods since they can have unpredictable results.  So I use just "default" and "high".

Signal/src/ViewControllers/LoadingViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+
+    var isShowingTopLabel = false
+    var isShowingBottomLabel = false
+    override public func viewDidAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+
+        // We only show the "loading" UI if it's a slow launch. Otherwise this ViewController
+        // should be indistinguishable from the launch screen.
+        let kTopLabelThreshold: TimeInterval = 5
+        DispatchQueue.main.asyncAfter(deadline: .now() + kTopLabelThreshold) { [weak self] in
```
charlesmchen-signal: Ideally, main thread should not be blocked during a long db migration or when we rebuild database extensions.  But it could. Please add to the test plan some scenario where we rebuild all db extensions and verify that this works correctly.

Signal/src/ViewControllers/LoadingViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    var isShowingTopLabel = false
+    var isShowingBottomLabel = false
+    override public func viewDidAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+
+        // We only show the "loading" UI if it's a slow launch. Otherwise this ViewController
+        // should be indistinguishable from the launch screen.
+        let kTopLabelThreshold: TimeInterval = 5
```
charlesmchen-signal: IMO should be 1-2 seconds, not 5. Users are impatient and give up on the app before 5 seconds.

Signal/src/ViewControllers/LoadingViewController.swift
michaelkirk-signal: Sure thing.

Signal/src/ViewControllers/LoadingViewController.swift
michaelkirk-signal: On my 5c with 100 threads with 100 messages, the average launch time is about 1.8 seconds.

That's way too close to 2 seconds for my comfort - we don't want to risk showing this message to people on every launch.

Signal/src/ViewControllers/LoadingViewController.swift
michaelkirk-signal: In practice I rarely use `low`, but I certainly haven't avoided it, and can imagine certain circumstances where I *would* use it. e.g. if I have ten objects in a stack view, 9 of which I want to stay small and only one which I want to grow, I'd apply low content hugging to the one view, rather than high hugging to the other 9.

I'd be glad to know what these unpredictable results are about.

michaelkirk-signal:
in https://github.com/signalapp/Signal-iOS/commit/fff7b1196114f10c392e8f387e8ce86cba2f265e

PR #3373, opened 2018-05-03
Charlesmchen/edit contact share view

charlesmchen-signal:

PTAL @michaelkirk 

This PR incorporates many of the design changes/clarifications from Myles.

simulator screen shot - iphone x - 2018-05-03 at 14 08 23
simulator screen shot - iphone x - 2018-05-03 at 14 08 18
simulator screen shot - iphone x - 2018-05-03 at 14 08 10

### DIFF
--- a/Signal/src/ViewControllers/ContactViewController.swift
+++ b/Signal/src/ViewControllers/ContactViewController.swift
@@ -690,10 +690,10 @@ class ContactViewController: OWSViewController, CNContactViewControllerDelegate 
         contactViewController.allowsActions = false
         contactViewController.allowsEditing = true
         contactViewController.navigationItem.leftBarButtonItem = UIBarButtonItem(title: CommonStrings.cancelButton, style: .plain, target: self, action: #selector(didFinishEditingContact))
-                contactViewController.navigationItem.leftBarButtonItem = UIBarButtonItem(title: CommonStrings.cancelButton,
-                                                                                         style: .plain,
-                                                                                         target: self,
-                                                                                         action: #selector(didFinishEditingContact))
+        contactViewController.navigationItem.leftBarButtonItem = UIBarButtonItem(title: CommonStrings.cancelButton,
+                                                                                 style: .plain,
+                                                                                 target: self,
+                                                                                 action: #selector(didFinishEditingContact))
 
         self.navigationController?.pushViewController(contactViewController, animated: true)
 
--- a/Signal/src/ViewControllers/ContactViewController.swift
+++ b/Signal/src/ViewControllers/ContactViewController.swift
@@ -14,7 +14,7 @@ class TappableView: UIView {
 
     // MARK: - Initializers
 
-    @available(*, unavailable, message: "use init(call:) constructor instead.")
+    @available(*, unavailable, message: "use other constructor instead.")
     required init?(coder aDecoder: NSCoder) {
         fatalError("Unimplemented")
     }
--- a/Signal/src/ViewControllers/ContactViewController.swift
+++ b/Signal/src/ViewControllers/ContactViewController.swift
@@ -367,48 +367,14 @@ class ContactViewController: OWSViewController, CNContactViewControllerDelegate 
     private func createFieldsView() -> UIView {
         SwiftAssertIsOnMainThread(#function)
 
-        let fieldsView = UIView.container()
-        fieldsView.layoutMargins = .zero
-        fieldsView.preservesSuperviewLayoutMargins = false
-
-        var lastRow: UIView?
-
-        let addSpacerRow = {
-            guard let prevRow = lastRow else {
-                owsFail("\(self.logTag) missing last row")
-                return
-            }
-            let row = UIView()
-            row.backgroundColor = UIColor(rgbHex: 0xdedee1)
-            fieldsView.addSubview(row)
-            row.autoSetDimension(.height, toSize: 1)
-            row.autoPinLeadingToSuperviewMargin(withInset: self.hMargin)
-            row.autoPinTrailingToSuperviewMargin()
-            row.autoPinEdge(.top, to: .bottom, of: prevRow)
-            lastRow = row
-        }
-
-        let addRow: ((UIView) -> Void) = { (row) in
-            if lastRow != nil {
-                addSpacerRow()
-            }
-            fieldsView.addSubview(row)
-            row.autoPinLeadingToSuperviewMargin()
-            row.autoPinTrailingToSuperviewMargin()
-            if let lastRow = lastRow {
-                row.autoPinEdge(.top, to: .bottom, of: lastRow)
-            } else {
-                row.autoPinEdge(toSuperviewEdge: .top)
-            }
-            lastRow = row
-        }
+        var rows = [UIView]()
 
         if viewMode == .nonSystemContact {
-            addRow(createActionRow(labelText: NSLocalizedString("CONVERSATION_SETTINGS_NEW_CONTACT",
+            rows.append(createActionRow(labelText: NSLocalizedString("CONVERSATION_SETTINGS_NEW_CONTACT",
                                                                comment: "Label for 'new contact' button in conversation settings view."),
                                    action: #selector(didPressCreateNewContact)))
 
-            addRow(createActionRow(labelText: NSLocalizedString("CONVERSATION_SETTINGS_ADD_TO_EXISTING_CONTACT",
+            rows.append(createActionRow(labelText: NSLocalizedString("CONVERSATION_SETTINGS_ADD_TO_EXISTING_CONTACT",
                                                                comment: "Label for 'new contact' button in conversation settings view."),
                                    action: #selector(didPressAddToExistingContact)))
         }
@@ -424,7 +390,7 @@ class ContactViewController: OWSViewController, CNContactViewControllerDelegate 
         for phoneNumber in contact.phoneNumbers {
             let formattedPhoneNumber = PhoneNumber.bestEffortLocalizedPhoneNumber(withE164: phoneNumber.phoneNumber)
 
-            addRow(createNameValueRow(name: phoneNumber.localizedLabel(),
+            rows.append(createNameValueRow(name: phoneNumber.localizedLabel(),
                                       value: formattedPhoneNumber,
                                       actionBlock: {
                                         guard let url = NSURL(string: "tel:\(phoneNumber.phoneNumber)") else {
@@ -436,7 +402,7 @@ class ContactViewController: OWSViewController, CNContactViewControllerDelegate 
         }
 
         for email in contact.emails {
-            addRow(createNameValueRow(name: email.localizedLabel(),
+            rows.append(createNameValueRow(name: email.localizedLabel(),
                                       value: email.email,
                                       actionBlock: {
                                         guard let url = NSURL(string: "mailto:\(email.email)") else {
@@ -449,9 +415,7 @@ class ContactViewController: OWSViewController, CNContactViewControllerDelegate 
 
         // TODO: Should we present addresses here too? How?
 
-        lastRow?.autoPinEdge(toSuperviewEdge: .bottom)
-
-        return fieldsView
+        return ContactFieldView(rows: rows, hMargin: hMargin)
     }
 
     private let hMargin = CGFloat(16)
--- a/Signal/src/ViewControllers/ContactsPicker.swift
+++ b/Signal/src/ViewControllers/ContactsPicker.swift
@@ -60,7 +60,8 @@ public class ContactsPicker: OWSViewController, UITableViewDelegate, UITableView
         CNContactFormatter.descriptorForRequiredKeys(for: .fullName),
         CNContactThumbnailImageDataKey as CNKeyDescriptor,
         CNContactPhoneNumbersKey as CNKeyDescriptor,
-        CNContactEmailAddressesKey as CNKeyDescriptor
+        CNContactEmailAddressesKey as CNKeyDescriptor,
+        CNContactPostalAddressesKey as CNKeyDescriptor
     ]
 
     // MARK: - Lifecycle Methods
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -117,6 +117,7 @@ typedef enum : NSUInteger {
 #pragma mark -
 
 @interface ConversationViewController () <AttachmentApprovalViewControllerDelegate,
+    ApproveContactShareViewControllerDelegate,
     AVAudioPlayerDelegate,
     CNContactViewControllerDelegate,
     ContactEditingDelegate,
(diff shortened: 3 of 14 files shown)


Signal/src/ViewControllers/ContactsPicker.swift
```diff
@@ -60,7 +60,8 @@ public class ContactsPicker: OWSViewController, UITableViewDelegate, UITableView
         CNContactFormatter.descriptorForRequiredKeys(for: .fullName),
         CNContactThumbnailImageDataKey as CNKeyDescriptor,
         CNContactPhoneNumbersKey as CNKeyDescriptor,
-        CNContactEmailAddressesKey as CNKeyDescriptor
+        CNContactEmailAddressesKey as CNKeyDescriptor,
+        CNContactPostalAddressesKey as CNKeyDescriptor
```
charlesmchen-signal: We need to fetch addresses too.

SignalMessaging/attachments/ApproveContactShareViewController.swift
```diff
@@ -0,0 +1,602 @@
...
+    let hMargin = CGFloat(0)
+
+    func createContents() {
+        self.layoutMargins.left = 0
+        self.layoutMargins.right = 0
+
+        let checkbox = UIButton(type: .custom)
+        self.checkbox = checkbox
+        // TODO: Use real assets.
+        checkbox.setTitle("☐", for: .normal)
```
charlesmchen-signal: Used emoji as temporary assets.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4915,10 +4937,27 @@ - (void)contactsPicker:(ContactsPicker *)contactsPicker contactFetchDidFail:(NSE
...
 
-    // TODO actually build contact message.
-    self.inputToolbar.messageText = contact.fullName;
+    OWSContact *_Nullable contactShare = [OWSContacts contactForSystemContact:contact.cnContact];
+    if (!contactShare) {
+        DDLogError(@"%@ Could not convert system contact.", self.logTag);
+        return;
+    }
+
+    // TODO: We should probably show this in the same navigation view controller.
```
charlesmchen-signal: There's a card.

SignalMessaging/attachments/ApproveContactShareViewController.swift
```diff
@@ -0,0 +1,556 @@
...
+
+        var rows = [UIView]()
+
+        rows.append(createNameRow())
+
+        for fieldView in fieldViews {
+            rows.append(fieldView)
+        }
+
+        return ContactFieldView(rows: rows, hMargin: hMargin)
```
charlesmchen-signal: I pulled out `ContactFieldView` to DRY up the commonalities between how contact fields are presented 3 different views.

SignalServiceKit/src/Messages/Interactions/OWSContact.m
```diff
@@ -398,6 +407,55 @@ - (NSString *)debugDescription
     return result;
 }
 
+- (OWSContact *)newContactWithNamePrefix:(nullable NSString *)namePrefix
```
charlesmchen-signal: `OWSContact` is intended to be immutable once we've finished building it.  These methods can be used to "create" and "derive" new contacts.

charlesmchen-signal:

PTAL @michaelkirk

SignalMessaging/attachments/ApproveContactShareViewController.swift
```diff
@@ -0,0 +1,556 @@
...
+        var values = [OWSContactEmail]()
+        values += contact.emails
+        values.append(value)
+        contact.emails = values
+    }
+}
+
+// MARK: -
+
+class ContactShareAddress: ContactShareField {
```
michaelkirk-signal: It's not a big deal, but it seems like making ContactShareField generic over the `value` would remove some redundancy in these field classes, so ultimately you could write something like:

```     
class ContactShareAddress: ContactShareField<OWSContactAddress>
```

I can follow up with an example later if you'd like.

SignalMessaging/attachments/ApproveContactShareViewController.swift
michaelkirk-signal: niiiiice.

SignalMessaging/attachments/ApproveContactShareViewController.swift
```diff
@@ -0,0 +1,556 @@
...
+                return strongSelf.previewView(forAddress: address)
+            })
+            fieldViews.append(fieldView)
+        }
+
+        self.fieldViews = fieldViews
+    }
+
+    override public var canBecomeFirstResponder: Bool {
+        return true
```
michaelkirk-signal: Why do we want to become first responder?

SignalMessaging/attachments/ApproveContactShareViewController.swift
```diff
@@ -0,0 +1,556 @@
...
+        super.viewDidDisappear(animated)
+    }
+
+    override public func loadView() {
+        super.loadView()
+
+        self.navigationItem.title = NSLocalizedString("CONTACT_SHARE_APPROVAL_VIEW_TITLE",
+                                                      comment: "Title for the 'Approve contact share' view.")
+
+        self.view.preservesSuperviewLayoutMargins = false
```
michaelkirk-signal: I often see you ignoring the superview layout margins, but I'm not sure why. Don't we generally want to respect the system layout margins? 

i.e. when do we want to pin to the margin, it seems like we'd almost never want to be closer than 16px from the screen edge.

If what you want instead is to pin a view to the edge, and not the margin, then pin to the edge, rather than setting the margin to zero and pinning to the margin.

SignalMessaging/attachments/ApproveContactShareViewController.swift
```diff
@@ -0,0 +1,556 @@
...
+
+        return ContactFieldView(rows: rows, hMargin: hMargin)
+    }
+
+    private let hMargin = CGFloat(16)
+
+    func createNameRow() -> UIView {
+        let nameVMargin = CGFloat(16)
+
+        let row = UIView()
```
michaelkirk-signal: Ah this is similar to what we were talking about yesterday - why this wrapper view?

You can set the margins right on the stack view (but remember to set `stackView.isLayoutMarginsRelativeArrangement = true`)

SignalMessaging/attachments/ApproveContactShareViewController.swift
```diff
@@ -0,0 +1,556 @@
...
+
+        row.isUserInteractionEnabled = true
+        row.addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(didPressEditName)))
+
+        return row
+    }
+
+    func previewView(forPhoneNumber phoneNumber: OWSContactPhoneNumber) -> UIView {
+        let label = UILabel()
+        label.text = PhoneNumber.bestEffortLocalizedPhoneNumber(withE164: phoneNumber.phoneNumber)
```
michaelkirk-signal: ah, that reads a bit better. :+1:

SignalMessaging/attachments/ApproveContactShareViewController.swift
```diff
@@ -0,0 +1,556 @@
...
+        label.font = UIFont.ows_dynamicTypeCaption1
```
michaelkirk-signal: consider pulling out these font attributes to design constants so we don't have to update them in three+ places.

SignalMessaging/attachments/EditContactShareNameViewController.swift
```diff
@@ -0,0 +1,308 @@
...
+        let vMargin = CGFloat(10)
+        self.layoutMargins = UIEdgeInsets(top: vMargin, left: 0, bottom: vMargin, right: 0)
+
+        let stackView = UIStackView()
+        stackView.axis = .horizontal
+        stackView.alignment = .center
+        stackView.layoutMargins = .zero
+        stackView.spacing = 10
+        self.addSubview(stackView)
+        stackView.autoPinTopToSuperviewMargin()
```
michaelkirk-signal: stackView.autoPinEdgesToSuperviewMargins()

SignalMessaging/attachments/EditContactShareNameViewController.swift
```diff
@@ -0,0 +1,308 @@
...
+        guard let delegate = self.delegate else {
+            owsFail("\(logTag) missing delegate.")
+            return
+        }
+
+        delegate.nameFieldDidChange()
+    }
+
+    public func value() -> String {
+        guard let value = valueView.text else {
```
michaelkirk-signal: Generally I'm all about returning non-null values, but we sometimes use a nil `String?` to indicate a value is *missing*, e.g. if someone doesn't have a name prefix.

Do we make other code simpler by returning an empty string vs nil here?

I don't want to confuse empty strings as real values elsewhere - the only thing worse than an unnecessary nil check is an unnecessary `length == 0` check.

Take it or leave it though, maybe we can't avoid those length checks anyway.

SignalMessaging/attachments/EditContactShareNameViewController.swift
```diff
@@ -0,0 +1,308 @@
...
+            namePrefixView ,
+            givenNameView ,
+            middleNameView ,
+            familyNameView ,
+            nameSuffixView
+        ]
+    }
+
+    override public var canBecomeFirstResponder: Bool {
+        return true
```
michaelkirk-signal: again, why/when do we become first responder?

SignalMessaging/attachments/EditContactShareNameViewController.swift
```diff
@@ -0,0 +1,308 @@
...
+                                                     nameSuffix: nameSuffixView.value())
+
+        guard let delegate = self.delegate else {
+            owsFail("\(logTag) missing delegate.")
+            return
+        }
+
+        delegate.editContactShareNameView(self, didEditContactShare: modifiedContactShare)
+
+        self.navigationController?.popViewController(animated: true)
```
michaelkirk-signal: guard/let/fail?

SignalMessaging/attachments/EditContactShareNameViewController.swift
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    func didPressCancel() {
+        Logger.info("\(logTag) \(#function)")
+
+        self.navigationController?.popViewController(animated: true)
```
michaelkirk-signal: guard/let/fail?

SignalServiceKit/src/Messages/Interactions/OWSContact.m
michaelkirk-signal: I like that you're working towards making the model less mutable 👍  

But for now we still mutate fields in the `valueField.applyToContact(contact: contact)` methods right?

SignalMessaging/attachments/ContactFieldView.swift
```diff
@@ -0,0 +1,63 @@
...
+        let addSpacerRow = {
+            guard let prevRow = lastRow else {
+                owsFail("\(self.logTag) missing last row")
+                return
+            }
+            let row = UIView()
+            row.backgroundColor = UIColor(rgbHex: 0xdedee1)
+            self.addSubview(row)
+            row.autoSetDimension(.height, toSize: 1)
+            row.autoPinLeadingToSuperviewMargin(withInset: hMargin)
```
michaelkirk-signal: Something seems a little strange about this inset/margin juggling.

As I understand it:

- You've explicitly set the superview's horizontal margin to 0
- You then pin to that margin (which is now zero, so you could just as well pin to the edge here)
- Then you offset that margin pinning by an inset.

Since you've repeatedly zero'd out horizontal margins on all superviews this works out in the end, but it feels like you're fighting the paradigm. Rather than zeroing out the margin and then pinning to the margin, just pin to the edge. Or if you want a consistent offset, define that as your superviews margin and pin to that.

Apologies if I've covered this same code in another PR, but I keep seeing similar code.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: Yep.

SignalServiceKit/src/Messages/Interactions/OWSContact.m
charlesmchen-signal: Yeah, it's a grey area.  I consider that "building" the derived contact share, so I allowed it.  We could have been more strict with making these models immutable, but it would have involved significant complexity with IMO little real benefit.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -102,10 +102,7 @@ class ContactViewController: OWSViewController, CNContactViewControllerDelegate
 
         UIUtil.applySignalAppearence()
 
-        if let navigationController = self.navigationController {
-            owsFail("\(logTag) missing navigationController")
```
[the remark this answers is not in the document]
charlesmchen-signal: Yeah, I decided to just use ?.

SignalMessaging/attachments/ApproveContactShareViewController.swift
charlesmchen-signal: I've done it, but I think it's marginally worse.

SignalMessaging/attachments/ApproveContactShareViewController.swift
charlesmchen-signal: We still haven't fixed all of the issues around the input toolbar showing up, so it wasn't clear to me if we are going to need this solution too.  Removed for now.

SignalMessaging/attachments/ApproveContactShareViewController.swift
charlesmchen-signal: Sorry, based on previous PR feedback I thought you preferred always pinning to margins and not edges.  I must have misunderstood.  

Note that AFAIK the correct way to pin subviews of a `UIScrollView` to the "full width" of the scroll view is to use the `layoutMarginsGuide`, via `autoPinLeadingToSuperviewMargin` for example.

SignalMessaging/attachments/ApproveContactShareViewController.swift
charlesmchen-signal: This was just a rough MVP to unblock testing.  I've removed this in a subsequent PR.

SignalMessaging/attachments/ApproveContactShareViewController.swift
charlesmchen-signal: They've been de-duplicated in a subsequent PR.

SignalMessaging/attachments/EditContactShareNameViewController.swift
charlesmchen-signal: Done.

SignalMessaging/attachments/EditContactShareNameViewController.swift
charlesmchen-signal: I think this is already fixed in a subsequent PR.

SignalMessaging/attachments/EditContactShareNameViewController.swift
charlesmchen-signal: See above.

SignalMessaging/attachments/EditContactShareNameViewController.swift
charlesmchen-signal: Done.

SignalMessaging/attachments/ContactFieldView.swift
charlesmchen-signal: This has all be reworked in a subsequent PR.

PR #3374, opened 2018-05-03
Fix media page title view on iOS9/10

michaelkirk-signal:
We must specify the frame of the titleView for iOS9/10.

Previously the titleView was a stackview, but we don't want to
"stretch" the labels to fit the size of the stack view, so we wrap the
stack view in a container view which becomes the new titleView.

**iOS 9**
ios9

**iOS 10**
screen shot 2018-05-03 at 2 13 03 pm

**iOS 11**
ios11

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MediaPageViewController.swift
+++ b/Signal/src/ViewControllers/MediaPageViewController.swift
@@ -67,7 +67,7 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
     private let showAllMediaButton: Bool
     private let sliderEnabled: Bool
 
-    private let headerView: UIStackView
+    private let headerView: UIView
 
     init(initialItem: MediaGalleryItem, mediaGalleryDataSource: MediaGalleryDataSource, uiDatabaseConnection: YapDatabaseConnection, options: MediaGalleryOption) {
         assert(uiDatabaseConnection.isInLongLivedReadTransaction())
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -78,20 +78,31 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
 
         let kSpacingBetweenItems: CGFloat = 20
 
-        let headerView = UIStackView()
-        self.headerView = headerView
+        self.headerView = UIView()
+        headerView.layoutMargins = UIEdgeInsets(top: 2, left: 8, bottom: 4, right: 8)
```
charlesmchen-signal: I too find myself wrapping `UIStackView` in a UIView with layout margins often.

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -78,20 +78,31 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
...
+        headerStackView.distribution = .fillProportionally
+        headerStackView.addArrangedSubview(headerNameLabel)
+        headerStackView.addArrangedSubview(headerDateLabel)
+
+        headerStackView.autoPinEdge(toSuperviewMargin: .top, relation: .greaterThanOrEqual)
+        headerStackView.autoPinEdge(toSuperviewMargin: .right, relation: .greaterThanOrEqual)
+        headerStackView.autoPinEdge(toSuperviewMargin: .bottom, relation: .greaterThanOrEqual)
+        headerStackView.autoPinEdge(toSuperviewMargin: .left, relation: .greaterThanOrEqual)
+        headerStackView.setContentHuggingHigh()
+        NSLayoutConstraint.autoSetPriority(UILayoutPriorityDefaultLow) {
```
charlesmchen-signal: Does this have to be low priority?

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -609,5 +621,19 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
...
+            // Do nothing, on iOS11, autolayout grows the stack view as necessary.
+        } else {
+            // Size the titleView to be large enough to fit the widest label,
+            // but no larger. If we go for a "full width" label, our title view
+            // will not be centered (since the left and right bar buttons have different widths)            
+            headerNameLabel.sizeToFit()
+            headerDateLabel.sizeToFit()
+            let maxWidth = max(headerNameLabel.frame.width, headerDateLabel.frame.width)
+
+            let headerFrame: CGRect = CGRect(x: 0, y: 0, width: maxWidth, height: 44)
```
charlesmchen-signal: I wonder if it would be better to hardcode 44 or to use `headerNameLabel.height + headerDateLabel.height`.

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: It just has to be lower priority than the constraints to the superview margin - so anything less than `UILayoutPriorityRequired` would do.

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: Hmm.... I wouldn't normally do something like this, and instead I'd assign the layout margins to the StackView directly, but the hangup here is that because we're dealing with views in the navbar on iOS9/10 we have to set an explicit frame, so the stack view can't "self size".

I can't think of a reason to do it outside of that, but maybe I'm missing something.

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: That could work, but then I'm not sure I could vertically position as flexibly, e.g. note how the container margins on top are 2, but on the bottom are 4.

It's probably doable, but I found the current approach to be flexible.

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: Oh wait, you're right, it doesn't need to be low priority.

PR #3379, opened 2018-05-04
oOnly one system contacts fetch at a time.

charlesmchen-signal:
PTAL @michaelkirk


SignalMessaging/contacts/SystemContactsFetcher.swift
```diff
@@ -269,11 +273,25 @@ public class SystemContactsFetcher: NSObject {
             Logger.error("background task time ran out contacts fetch completed.")
         })
 
+        guard !isFetching else {
```
charlesmchen-signal: I chose not to use a serial queue, since we want to de-bounce.  If multiple fetches are enqueued during a fetch, we only want to kick off _one_ additional fecth when the current fetch completes.

michaelkirk-signal: It's a valid perf concern, but now sometimes our completion handler isn't called at the end of execution.

Immediate thoughts are to either:
1. go with this approach, but retain a list of pending completion handlers
2. or revert most of this and dispatch on a serial queue.

charlesmchen-signal: You're right, using a serial queue.

PTAL @michaelkirk

PR #3380, opened 2018-05-04
Handle completely skipped message sends.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSMessageSender.m
+++ b/SignalServiceKit/src/Messages/OWSMessageSender.m
@@ -501,6 +501,12 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
                 }];
             }
 
+            if (sendingRecipientIds.count < 1) {
+                // All recipients are already sent or can be skipped.
+                successHandler();
+                return;
+            }
+
             NSError *error;
             NSArray<SignalRecipient *> *recipients =
                 [self getRecipientsForRecipientIds:sendingRecipientIds.allObjects error:&error];


SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -501,6 +501,12 @@ - (void)sendMessageToService:(TSOutgoingMessage *)message
                 }];
             }
 
+            if (sendingRecipientIds.count < 1) {
+                // All recipients are already sent or can be skipped.
+                successHandler();
+                return;
```
charlesmchen-signal: For example, if you create a group with no members, there's no need to send a message and this is not an error.

michaelkirk-signal: Hmm... so what's the difference in behavior?

Originally:

- Create a group with no members
- Send
- Shows as failed

And now:

- Create a group with no members
- Send
- Shows as sent (but not delivered)

Is that right?

charlesmchen-signal: It wasn't message sends to empty group that was failing, it was the "group update/creation" message itself that was failing.

charlesmchen-signal: But it's worth noting that this will also affect a number of other edge cases.

* You're sending to a group with 1 member and you learn that that member has left the group between the time a message is enqueued and sent.
* You're sending to a group with 1 member and that member is blocked.

charlesmchen-signal: I'll do my best to write a test plan that covers these and any other edge cases I can think of.

michaelkirk-signal: I'm just trying to understand the new behavior, from the users perspective. 

What behavior does this change (in terms of what the user sees now?) And what did they see before this change?

michaelkirk-signal: I guess you can cover that in the test plan - code changes LGTM

michaelkirk-signal: Thanks for the clarifications!

PR #3385, opened 2018-05-05
contact share: sending/receiving avatars

michaelkirk-signal:
Based on #3373 so merged that first.


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2948,20 +2949,28 @@ - (void)sendMessageAttachment:(SignalAttachment *)attachment
...
     DDLogVerbose(@"%@ Sending contact share.", self.logTag);
 
     BOOL didAddToProfileWhitelist = [ThreadUtil addThreadToProfileWhitelistIfEmptyContactThread:self.thread];
-    TSOutgoingMessage *message = [ThreadUtil sendMessageWithContactShare:contactShare
-                                                                inThread:self.thread
-                                                           messageSender:self.messageSender
-                                                              completion:nil];
 
-    [self messageWasSent:message];
+    [self.editingDatabaseConnection asyncReadWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
```
charlesmchen-signal: If this is going to be an async transaction, the `ensureDynamicInteractions` work below should be moved into the completion block.

SignalMessaging/ViewModels/ContactShareViewModel.swift
```diff
@@ -0,0 +1,132 @@
...
+        if let avatarImage = avatarImage {
+            return avatarImage
+        }
+
+        // TODO: What's the best colorSeed value to use?
+        let avatarBuilder = OWSContactAvatarBuilder(nonSignalName: displayName,
+                                                    colorSeed: displayName,
+                                                    diameter: UInt(diameter),
+                                                    contactsManager: contactsManager)
+        return avatarBuilder.build()
```
charlesmchen-signal: Should we cache this result in `avatarImage`?

SignalServiceKit/src/Contacts/TSThread.h
```diff
@@ -43,16 +43,6 @@ NS_ASSUME_NONNULL_BEGIN
...
 
-#if TARGET_OS_IOS
-
-/**
- *  Returns the image representing the thread. Nil if not available.
- *
- *  @return UIImage of the thread, or nil.
- */
-- (nullable UIImage *)image;
-#endif
```
charlesmchen-signal: Should we also remove `- (nullable UIImage *)image` from `TSThread.m`?

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -703,6 +703,75 @@ - (nullable TSAttachmentStream *)cloneAsThumbnail
...
 }
 
+// MARK: Protobuf serialization
+
++ (nullable OWSSignalServiceProtosAttachmentPointer *)buildProtoForAttachmentId:(nullable NSString *)attachmentId
+                                                                 isVoiceMessage:(BOOL)isVoiceMessage
+{
+    OWSAssert(attachmentId.length > 0);
+
+    // TODO we should hoist this transaction to make sure we're getting a consistent view in the message sending process
```
charlesmchen-signal: What does `hoist` mean here?  Pass in/plumb through?

SignalServiceKit/src/Storage/OWSOrphanedDataCleaner.m
```diff
@@ -129,22 +131,30 @@ + (void)auditAndCleanup:(BOOL)shouldCleanup completion:(void (^_Nullable)(void))
...
+    [orphanAttachmentIds minusSet:contactShareAvatarAttachmentIds];
     NSMutableSet<NSString *> *missingAttachmentIds = [messageAttachmentIds mutableCopy];
     [missingAttachmentIds minusSet:attachmentIds];
 
-    CleanupLogDebug(@"%@ orphan attachmentIds: %zd", self.logTag, orphanAttachmentIds.count);
-    CleanupLogDebug(@"%@ missing attachmentIds: %zd", self.logTag, missingAttachmentIds.count);
-    CleanupLogDebug(@"%@ orphan interactions: %zd", self.logTag, orphanInteractionIds.count);
+    CleanupLogDebug(@"%@ orphan attachmentIds: %tu", self.logTag, orphanAttachmentIds.count);
+    CleanupLogDebug(@"%@ missing attachmentIds: %tu", self.logTag, missingAttachmentIds.count);
+    CleanupLogDebug(@"%@ orphan interactions: %tu", self.logTag, orphanInteractionIds.count);
```
charlesmchen-signal: Is `%tu` right?

> t | Length modifier specifying that a following d, o, u, x, or X conversion specifier applies to a ptrdiff_t.

from https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/Strings/Articles/formatSpecifiers.html

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: Done.

SignalMessaging/ViewModels/ContactShareViewModel.swift
michaelkirk-signal: The avatar builder has built in caching per resolution which ensures our generated initials (text) avatars, which are especially sensitive to compression/resizing artifacts are to-spec, so I'm inclined to leave this as is and rely on the avatar builder cache.

SignalServiceKit/src/Contacts/TSThread.h
michaelkirk-signal: Yes, thanks!

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
michaelkirk-signal: Correct - I'll update the comment.

SignalServiceKit/src/Storage/OWSOrphanedDataCleaner.m
michaelkirk-signal: Yeah, it happens to work on 32 and 64 bit, but it seems like a hack. The problem with the existing zd, is that although there's no warning, it's the wrong type - (i.e. very large unsigned ints are shown as negative.)

I've replaced all these usages with %lu and an explicit cast to `(unsigned long)` which is the party line.

PR #3387, opened 2018-05-07
Consider off/on in addition to duration when becoming consistent

michaelkirk-signal:
Fixes https://github.com/signalapp/Signal-iOS/issues/3365

PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
+++ b/SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
@@ -15,6 +15,7 @@
 #import "OWSPrimaryStorage.h"
 #import "TSIncomingMessage.h"
 #import "TSMessage.h"
+#import "TSThread.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -225,27 +226,25 @@ - (void)becomeConsistentWithConfigurationForMessage:(TSMessage *)message
...
     if (message.expiresInSeconds == 0) {
-        if (disappearingMessagesConfiguration.isEnabled) {
-            changed = YES;
-            DDLogWarn(@"%@ Received remote message which had no expiration set, disabling our expiration to become "
-                      @"consistent.",
-                      self.logTag);
-            disappearingMessagesConfiguration.enabled = NO;
-            [disappearingMessagesConfiguration saveWithTransaction:transaction];
-        }
-    } else if (message.expiresInSeconds != disappearingMessagesConfiguration.durationSeconds) {
```
michaelkirk-signal: Previously we were only considering duration and not "enabled" state when becoming consistent with disappearing messages.

SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -225,27 +226,25 @@ - (void)becomeConsistentWithConfigurationForMessage:(TSMessage *)message
...
+        return;
+    }
+    DDLogInfo(@"%@ becoming consistent message configuration: %@",
+        self.logTag,
+        disappearingMessagesConfiguration.dictionaryValue);
+    [disappearingMessagesConfiguration saveWithTransaction:transaction];
+
+    TSThread *_Nullable thread = [message threadWithTransaction:transaction];
+    if (!thread) {
+        // If there's not yet a message thread, we don't create one.
```
michaelkirk-signal: This isn't a change in behavior, I'm just making existing behavior clearer *and* using the transaction we already have vs the sneaky transaction in `message.thread`.

SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -225,27 +226,25 @@ - (void)becomeConsistentWithConfigurationForMessage:(TSMessage *)message
...
+        disappearingMessagesConfiguration.enabled = NO;
+    } else {
         disappearingMessagesConfiguration.enabled = YES;
         disappearingMessagesConfiguration.durationSeconds = message.expiresInSeconds;
-        [disappearingMessagesConfiguration saveWithTransaction:transaction];
     }
-    
-    if (!changed) {
+
+    if (!disappearingMessagesConfiguration.dictionaryValueDidChange) {
```
charlesmchen-signal: Will `dictionaryValueDidChange` correctly handle the "create" case of `fetchOrCreateDefaultWithThreadId:...`

SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -225,27 +226,25 @@ - (void)becomeConsistentWithConfigurationForMessage:(TSMessage *)message
...
+        return;
+    }
+    DDLogInfo(@"%@ becoming consistent message configuration: %@",
+        self.logTag,
+        disappearingMessagesConfiguration.dictionaryValue);
+    [disappearingMessagesConfiguration saveWithTransaction:transaction];
+
+    TSThread *_Nullable thread = [message threadWithTransaction:transaction];
```
michaelkirk-signal: If it's a thread created within this transaction, the sneaky transaction in message.thread won't see it, so the initial info message gets lost.

PR #3389, opened 2018-05-07
Add contact share actions to conversation view and improve phone numb…

charlesmchen-signal:
…er parsing.

PTAL @michaelkirk


Signal/src/ViewControllers/ContactShareViewHelper.swift
```diff
@@ -0,0 +1,289 @@
...
+import ContactsUI
+import MessageUI
+
+@objc
+public protocol ContactShareViewHelperDelegate: class {
+    func didCreateOrEditContact()
+}
+
+@objc
+public class ContactShareViewHelper: NSObject, CNContactViewControllerDelegate {
```
charlesmchen-signal: I pulled out a helper class that contains to the logic for interacting with contact shares.

Signal/src/ViewControllers/ConversationView/Cells/OWSContactShareView.h
```diff
@@ -5,14 +5,31 @@
 NS_ASSUME_NONNULL_BEGIN
 
 @class ContactShareViewModel;
+@class OWSContact;
+@class OWSContactsManager;
+
+@protocol OWSContactShareViewDelegate <NSObject>
```
charlesmchen-signal: Contact share messages are now interactive in conversation view.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2092,6 +2094,42 @@ - (void)didTapContactShareViewItem:(ConversationViewItem *)conversationItem
...
+- (void)sendMessageToContactShare:(ContactShareViewModel *)contactShare
+{
+    OWSAssertIsOnMainThread();
+    OWSAssert(contactShare);
+
+    self.contactShareViewHelper = [[ContactShareViewHelper alloc] initWithContactShare:contactShare
+                                                                       contactsManager:self.contactsManager
+                                                                    fromViewController:self
+                                                                              delegate:self];
+    [self.contactShareViewHelper sendMessageToContact];
```
charlesmchen-signal: It's unfortunate but we need to retain a strong reference to `ContactShareViewHelper` while we're using it.

SignalMessaging/categories/UIView+OWS.h
```diff
@@ -135,6 +135,14 @@ CGFloat ScaleFromIPhone5(CGFloat iPhone5Value);
 
 @end
 
+#pragma mark -
+
+@interface UIStackView (OWS)
+
+- (void)addBackgroundViewWithBackgroundColor:(UIColor *)backgroundColor;
```
charlesmchen-signal: `UIStackView` is a non-rendering view.

charlesmchen-signal: I might further modify it so that it can have a border.

Signal/src/ViewControllers/ContactShareViewHelper.swift
```diff
@@ -0,0 +1,289 @@
...
+
+    weak var delegate: ContactShareViewHelperDelegate?
+
+    let contactShare: ContactShareViewModel
+    let contactsManager: OWSContactsManager
+    weak var fromViewController: UIViewController?
+
+    public required init(contactShare: ContactShareViewModel, contactsManager: OWSContactsManager, fromViewController: UIViewController, delegate: ContactShareViewHelperDelegate) {
```
charlesmchen-signal: This code largely moved unchanged from contact view.

Signal/src/ViewControllers/ContactShareViewHelper.swift
```diff
@@ -0,0 +1,289 @@
...
+    public func videoCallToContact() {
+        Logger.info("\(logTag) \(#function)")
+
+        presentThreadAndPeform(action: .videoCall)
+    }
+
+    private func presentThreadAndPeform(action: ConversationViewAction) {
+        // TODO: We're taking the first Signal account id. We might
+        // want to let the user select if there's more than one.
+        let phoneNumbers = contactShare.systemContactsWithSignalAccountPhoneNumbers(contactsManager)
```
charlesmchen-signal: The one change is that we're now leveraging the new "phone number convenience accessors" in `OWSContact` and `ContactShareViewModel`.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: It's a tough choice, because we either need "the last helper" (as we have it now) or we'll need the helper to know "the last contact share".  I prefer the former since we're using delegate pattern with system UI.  If we start a new action, we create a new helper which ensures that the old helper will be deallocated.  If we used a single mutable helper, we could conceivably receive a delegate invocation related to a different contact share.

Signal/src/ViewControllers/ContactShareViewHelper.swift
```diff
@@ -0,0 +1,289 @@
...
+            return
+        }
+
+        // TODO: Revisit this.
+        guard let firstPhoneNumber = contactShare.e164PhoneNumbers().first else {
+            owsFail("\(logTag) Missing phone number.")
+            return
+        }
+
+        // TODO: We need to modify OWSAddToContactViewController to take a OWSContact
```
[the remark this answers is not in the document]
charlesmchen-signal: Yes, this comment was written before the contact share view model existed.  Note that this code (and comment) was just moved not written in this PR.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -129,17 +133,19 @@ class ContactViewController: OWSViewController, CNContactViewControllerDelegate
...
     private func systemContactsForContact() -> [String] {
         SwiftAssertIsOnMainThread(#function)
 
-        return contactShare.phoneNumberStrings.filter({ (phoneNumber) -> Bool in
-            return contactsManager.allContactsMap[phoneNumber] != nil
-        })
+        return contactShare.systemContactPhoneNumbers(contactsManager)
+    }
+
+    private func phoneNumbersForContact() -> [String] {
```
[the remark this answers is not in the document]
charlesmchen-signal: Removed.

Signal/src/ViewControllers/ConversationView/Cells/OWSContactShareView.m
```diff
@@ -18,21 +18,30 @@
...
+- (instancetype)initWithContactShare:(ContactShareViewModel *)contactShare
+                          isIncoming:(BOOL)isIncoming
+                            delegate:(id<OWSContactShareViewDelegate>)delegate
 {
     self = [super init];
 
     if (self) {
-        _contactShare = contactShare;
-        _isIncoming = isIncoming;
+        self.delegate = delegate;
```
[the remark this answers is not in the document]
charlesmchen-signal: Done. This seems like a style point.  I converted some of these properties to `readonly`.

Signal/src/ViewControllers/ConversationView/Cells/OWSContactShareView.m
```diff
@@ -65,14 +74,60 @@ - (CGFloat)iconVMargin
...
+
+    return [contactShare systemContactsWithSignalAccountPhoneNumbers:contactsManager].count > 0;
+}
+
++ (BOOL)hasInviteButton:(ContactShareViewModel *)contactShare contactsManager:(OWSContactsManager *)contactsManager
+{
+    OWSAssert(contactShare);
+    OWSAssert(contactsManager);
+
+    return [contactShare systemContactPhoneNumbers:contactsManager].count > 0;
```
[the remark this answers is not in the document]
charlesmchen-signal: I'm skeptical that this is more readable.  I think we should include the parameter name in the first part of the method's name only where there's any ambiguity.  In cases like this I don't see it.  And insanely long method names are less readable.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.h
```diff
@@ -42,6 +44,11 @@ typedef NS_ENUM(NSUInteger, OWSMessageGestureLocation) {
 
 - (void)didTapContactShareViewItem:(ConversationViewItem *)viewItem;
 
+- (void)sendMessageToContactShare:(ContactShareViewModel *)contactShare NS_SWIFT_NAME(sendMessage(toContactShare:));
```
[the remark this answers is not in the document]
charlesmchen-signal: Renamed.

SignalServiceKit/src/Messages/Interactions/OWSContact.h
```diff
@@ -140,10 +141,19 @@ NSString *NSStringForContactAddressType(OWSContactAddressType value);
...
+- (NSArray<NSString *> *)systemContactPhoneNumbers:(id<ContactsManagerProtocol>)contactsManager
+    NS_SWIFT_NAME(systemContactPhoneNumbers(_:));
+- (NSArray<NSString *> *)e164PhoneNumbers NS_SWIFT_NAME(e164PhoneNumbers());
+
 @end
 
 #pragma mark -
 
+// TODO: Move to separate source file, rename to OWSContactConversion.
 @interface OWSContacts : NSObject
```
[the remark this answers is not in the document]
charlesmchen-signal: I don't think we should adopt the `Record` name convention within the codebase.  It's a ROR thing right?  These are model classes.  IMO we should either use Model or no suffix, and I vote no suffix.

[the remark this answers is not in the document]
michaelkirk-signal: > OH also, there's a difference between OWSContact (the model) and OWSContacts

Oh! I didn't realized that OWSContacts was different from OWSContact. I agree with your suggested rename.

> I don't think we should adopt the Record name convention within the codebase. It's a ROR thing right? 

Since this was predicated on my misunderstanding, I'll save this discussion for another day. =)

Signal/src/ViewControllers/ConversationView/Cells/OWSContactShareView.m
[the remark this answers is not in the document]
michaelkirk-signal: Yeah, I guess it's a style thing - but to be clear, it's not just aesthetics.

In general it's risky to call methods on self before self is initialized - since those setters could in turn rely on state that hasn't been initialized (or a future dev could later modify the getters so that they rely on other state). 

That's why Swift strictly enforces this and why I think it's a good convention to enforce upon ourselves.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: I was thinking like this: https://github.com/signalapp/Signal-iOS/pull/3398

PR #3390, opened 2018-05-07
Fix/Simplify outgoing attachment proto building

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
+++ b/SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
@@ -761,7 +761,7 @@ NSString *NSStringForOutgoingMessageRecipientState(OWSOutgoingMessageRecipientSt
             case TSGroupMessageNew: {
                 if (gThread.groupModel.groupImage != nil && self.attachmentIds.count == 1) {
                     attachmentWasGroupAvatar = YES;
-                    [groupBuilder setAvatar:[self buildProtoForAttachmentId:self.attachmentIds[0] filename:nil]];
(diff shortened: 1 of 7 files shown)


SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
```diff
@@ -761,7 +761,7 @@ - (OWSSignalServiceProtosDataMessageBuilder *)dataMessageBuilder
             case TSGroupMessageNew: {
                 if (gThread.groupModel.groupImage != nil && self.attachmentIds.count == 1) {
                     attachmentWasGroupAvatar = YES;
-                    [groupBuilder setAvatar:[self buildProtoForAttachmentId:self.attachmentIds[0] filename:nil]];
```
michaelkirk-signal: this is the fix - 'buildProtoForAttachmentId' got moved to TSAttachmentStream method, but I failed to remove the header definition so this still compiled even though the method didn't exist.

SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
```diff
@@ -781,8 +781,7 @@ - (OWSSignalServiceProtosDataMessageBuilder *)dataMessageBuilder
     if (!attachmentWasGroupAvatar) {
         NSMutableArray *attachments = [NSMutableArray new];
         for (NSString *attachmentId in self.attachmentIds) {
-            NSString *_Nullable sourceFilename = self.attachmentFilenameMap[attachmentId];
```
michaelkirk-signal: This is the change I was least confident with - but it seems like this attachmentFileName map is redundant with what's already living on TSAttachmentStream.

PR #3391, opened 2018-05-07
Fix: Input Toolbar appearing over media gallery, tapping voice note hides input toolbar

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -596,6 +596,28 @@ typedef enum : NSUInteger {
     [self markVisibleMessagesAsRead];
     [self.cellMediaCache removeAllObjects];
     [self cancelReadTimer];
+    [self dismissPresentedViewControllerIfNecessary];
+}
+
+- (void)dismissPresentedViewControllerIfNecessary
+{
+    UIViewController *_Nullable presentedViewController = self.presentedViewController;
+    if (!presentedViewController) {
+        DDLogDebug(@"%@ presentedViewController was nil", self.logTag);
+        return;
+    }
+
+    if ([presentedViewController isKindOfClass:[UIAlertController class]]) {
+        DDLogDebug(@"%@ dismissing presentedViewController: %@", self.logTag, presentedViewController);
+        [self dismissViewControllerAnimated:NO completion:nil];
+        return;
+    }
+
+    if ([presentedViewController isKindOfClass:[UIImagePickerController class]]) {
+        DDLogDebug(@"%@ dismissing presentedViewController: %@", self.logTag, presentedViewController);
+        [self dismissViewControllerAnimated:NO completion:nil];
+        return;
+    }
 }
 
 - (void)viewWillAppear:(BOOL)animated
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -552,15 +552,7 @@ typedef enum : NSUInteger {
 
 - (BOOL)canBecomeFirstResponder
 {
-    // Normally there'd be no mechanism for us to become first responder while presenting
-    // another view controller, however due to our screen lock window juggling, a side effect of
-    // calling `makeKeyAndVisible` is that "last known" first responder is sent "becomeFirstResponder",
-    // regardless of if it is no longer the top most VC.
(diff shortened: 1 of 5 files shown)


Signal/src/util/OWSWindowManager.m
```diff
@@ -67,9 +67,6 @@ @interface OWSWindowManager ()
 
 @property (nonatomic, nullable) UIViewController *callViewController;
 
-@property (nonatomic, nullable) UIResponder *rootWindowResponder;
```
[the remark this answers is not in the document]
charlesmchen-signal: > since becomeFirstResponder is called on window.rootViewController

But we don't want it called on the window's `rootViewController`, we want it called on the "frontmost" vc, right?

Signal/src/util/OWSWindowManager.m
```diff
@@ -473,7 +432,7 @@ - (void)ensureScreenBlockWindowShown
     }
 
     self.screenBlockingWindow.windowLevel = UIWindowLevel_ScreenBlocking();
-    [self.screenBlockingWindow.rootViewController becomeFirstResponder];
+    [self.screenBlockingWindow makeKeyAndVisible];
```
charlesmchen-signal: Hmm, this is a significant change.  I've wondered whether what the exact difference is between `makeKeyAndVisible` and simply manipulating the window's visibility.  

I had something like this in an early iteration of the window manager and I moved away from it because I found an issue.  I'm not against this, but we need a rigorous test plan that includes all permutations of:

* Screen lock enabled or disabled.
* Call View presented or not.
* Input toolbar visible or not.
* Input toolbar selected or not.
* Another view presented over conversation view using `present`.
* Another view presented over conversation view using `navigation controller`.

Note: To my surprise I found that it's necessary to repeat tests around responder state and input toolbar visibility _twice_ because in some cases issue occur only the second time.

SignalMessaging/utils/OWSScreenLock.swift
```diff
@@ -33,9 +33,6 @@ import LocalAuthentication
     private let OWSScreenLock_Key_IsScreenLockEnabled = "OWSScreenLock_Key_IsScreenLockEnabled"
     private let OWSScreenLock_Key_ScreenLockTimeoutSeconds = "OWSScreenLock_Key_ScreenLockTimeoutSeconds"
 
-    // We temporarily resign any first responder while the Screen Lock is presented.
-    weak var firstResponderBeforeLockscreen: UIResponder?
```
charlesmchen-signal: Is this just vestigial?

michaelkirk-signal: Yup.

Signal/src/util/OWSWindowManager.m
```diff
@@ -348,37 +345,11 @@ - (void)ensureRootWindowShown
         DDLogInfo(@"%@ showing root window.", self.logTag);
     }
 
-    BOOL shouldTryToRestoreFirstResponder = self.rootWindow.hidden;
-
+    // By calling makeKeyAndVisible we ensure the rootViewController becomes firt responder.
```
charlesmchen-signal: s/firt/first/

Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -220,7 +220,10 @@ class MediaGalleryViewController: UINavigationController, MediaGalleryDataSource
         fatalError("init(coder:) has not been implemented")
     }
 
-    override var canBecomeFirstResponder: Bool {
+    // HACK: Though we don't have an input accessory view, the VC we are presented above (ConversationVC) does.
```
charlesmchen-signal: s/presented above/presented from/

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -596,6 +596,28 @@ - (void)applicationWillResignActive:(NSNotification *)notification
...
+        return;
+    }
+
+    if ([presentedViewController isKindOfClass:[UIAlertController class]]) {
+        DDLogDebug(@"%@ dismissing presentedViewController: %@", self.logTag, presentedViewController);
+        [self dismissViewControllerAnimated:NO completion:nil];
+        return;
+    }
+
+    if ([presentedViewController isKindOfClass:[UIImagePickerController class]]) {
```
charlesmchen-signal: Can't we collapse these two conditions by using II?

PR #3392, opened 2018-05-08
Avoid spurious "Disabled Disappearing Messages"

michaelkirk-signal:
`dictionaryDidChange` was always true for new records.

I missed a case in https://github.com/signalapp/Signal-iOS/pull/3387 - for conversations which had never before used disappearing messages.

### DIFF
--- a/SignalServiceKit/src/Contacts/OWSDisappearingMessagesConfiguration.m
+++ b/SignalServiceKit/src/Contacts/OWSDisappearingMessagesConfiguration.m
@@ -44,8 +44,8 @@ NS_ASSUME_NONNULL_BEGIN
 
     _enabled = isEnabled;
     _durationSeconds = seconds;
-    _originalDictionaryValue = [NSDictionary new];
     _newRecord = YES;
+    _originalDictionaryValue = self.dictionaryValue;
 
     return self;
 }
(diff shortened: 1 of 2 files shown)


SignalServiceKit/src/Contacts/OWSDisappearingMessagesConfiguration.m
```diff
@@ -44,8 +44,8 @@ - (instancetype)initWithThreadId:(NSString *)threadId enabled:(BOOL)isEnabled du
 
     _enabled = isEnabled;
     _durationSeconds = seconds;
-    _originalDictionaryValue = [NSDictionary new];
     _newRecord = YES;
+    _originalDictionaryValue = self.dictionaryValue;
```
charlesmchen-signal: I'm surprised this fix works, since I would think that using an empty dictionary would cause:

* The dictionary comparison to return "not equal".
* `dictionaryValueDidChange` to be YES.
* The configuration to be saved and applied to the thread.

Can you please explain how this fix works?

michaelkirk-signal: Sorry - I should have been more clear about the problem that this fixes.

The problem was *not* with Alice sending a disappearing message for the first time. It's when Alice sends a regular "non-disappearing" message.

So you are correct when you expect:

- The dictionary comparison to return "not equal".
- dictionaryValueDidChange to be YES.
- The configuration to be saved and applied to the thread.
- Also an info message is printed.

The problem is, the configuration which was just saved is the default configuration of "no" disappearing messages. So we're printing an info message even though nothing changed.

Specifically, to reproduce:

- Alice and Bob reinstall.
- Alice **does not** enable disappearing messages and just says "hi bob"

Bob sees: 

- "Alice has disabled disappearing messages" (even though it was never enabled)
- "hi bob"

Does that clarify the fix?

PR #3397, opened 2018-05-08
Apply contact share assets.

charlesmchen-signal:
PTAL @michaelkirk 

simulator screen shot - iphone x - 2018-05-08 at 15 53 26


Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -417,6 +420,14 @@ class ContactViewController: OWSViewController, CNContactViewControllerDelegate
         circleView.autoPinEdge(toSuperviewEdge: .top)
         circleView.autoHCenterInSuperview()
 
+        guard let image = UIImage(named: imageName) else {
```
michaelkirk-signal: Reminder that you can better capture this pattern with declarative resources - and have the IDE make sure you didn't typo it.

charlesmchen-signal: I'm still thinking this over.

PR #3401, opened 2018-05-09
Show "share contact" flow in single presented navigation controller.

charlesmchen-signal:
PTAL @michaelkirk


michaelkirk-signal:
This is for sending, right?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -5024,11 +5022,7 @@ - (void)contactsPicker:(ContactsPicker *)contactsPicker didSelectContact:(Contac
         [[ApproveContactShareViewController alloc] initWithContactShare:contactShare
                                                         contactsManager:self.contactsManager
                                                                delegate:self];
-
-    UINavigationController *navigationController =
-        [[UINavigationController alloc] initWithRootViewController:approveContactShare];
-    [self dismissKeyBoard];
-    [self presentViewController:navigationController animated:YES completion:nil];
+    [contactsPicker.navigationController pushViewController:approveContactShare animated:YES];
```
michaelkirk-signal: can we assert that contactsPicker has a navigationController?

PR #3405, opened 2018-05-09
Charlesmchen/contact share dynamic type rtl

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/attachments/EditContactShareNameViewController.swift
+++ b/SignalMessaging/attachments/EditContactShareNameViewController.swift
@@ -42,9 +42,11 @@ class ContactNameFieldView: UIView {
         createContents()
     }
 
+    private let hMargin = CGFloat(16)
+
     func createContents() {
         let vMargin = CGFloat(10)
-        self.layoutMargins = UIEdgeInsets(top: vMargin, left: 0, bottom: vMargin, right: 0)
+        self.layoutMargins = UIEdgeInsets(top: vMargin, left: hMargin, bottom: vMargin, right: hMargin)
 
         let stackView = UIStackView()
         stackView.axis = .horizontal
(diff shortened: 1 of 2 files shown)


SignalMessaging/attachments/EditContactShareNameViewController.swift
```diff
@@ -42,9 +42,11 @@ class ContactNameFieldView: UIView {
         createContents()
     }
 
+    private let hMargin = CGFloat(16)
+
     func createContents() {
         let vMargin = CGFloat(10)
-        self.layoutMargins = UIEdgeInsets(top: vMargin, left: 0, bottom: vMargin, right: 0)
+        self.layoutMargins = UIEdgeInsets(top: vMargin, left: hMargin, bottom: vMargin, right: hMargin)
```
michaelkirk-signal: Do we want anything other than the system default horizontal margin?

Can this just be:

```
self.layoutMargins.top = vMargin
self.layoutMargins.bottom = vMargin
```

And an aesthetic nit: if an explicit hMargin is needed for some reason, seems like it might as well live next to the vMargin definition.

charlesmchen-signal: > Do we want anything other than the system default horizontal margin?

I'm trying to match the design - and I think Myle's design looks better than the default margins.

charlesmchen-signal: > if an explicit hMargin is needed for some reason, seems like it might as well live next to the vMargin definition.

Done!

michaelkirk:
But isn’t the default system margin 16? I wasn’t suggesting that you change the design, the code just looked like a noop.

PR #3407, opened 2018-05-10
Add to existing contact

michaelkirk-signal:
## Screenshots

**Contact Picker**

contact_picker

**Add To Existing**
add_to_existing

**Create New**
create_new

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
+++ b/Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
@@ -8,8 +8,12 @@ import ContactsUI
 
 class AddContactShareToExistingContactViewController: ContactsPicker, ContactsPickerDelegate, CNContactViewControllerDelegate {
 
-    // TODO actual protocol?
-    weak var addToExistingContactDelegate: UIViewController?
+    // TODO - there are some hard coded assumptions in this VC that assume we are *pushed* onto a
+    // navigation controller. That seems fine for now, but if we need to be presented as a modal,
+    // or need to notify our presenter about our dismisall or other contact actions, a delegate
+    // would be helpful. It seems like this would require some broad changes to the ContactShareViewHelper,
+    // so I've left it as is for now, since it happens to work.
+    // weak var addToExistingContactDelegate: AddContactShareToExistingContactViewControllerDelegate?
 
     let contactShare: ContactShareViewModel
 
@@ -55,8 +59,7 @@ class AddContactShareToExistingContactViewController: ContactsPicker, ContactsPi
         Logger.debug("\(self.logTag) in \(#function)")
 
         guard let mergedContact: CNContact = self.contactShare.cnContact(mergedWithExistingContact: contact) else {
-            // TODO maybe this should not be optional and return a blank contact so we can still save the (not-actually merged) contact
-            owsFail("\(logTag) in \(#function) navigationController was unexpectedly nil")
+            owsFail("\(logTag) in \(#function) mergedContact was unexpectedly nil")
             return
         }
 
@@ -104,7 +107,16 @@ class AddContactShareToExistingContactViewController: ContactsPicker, ContactsPi
             return
         }
 
-        // We want to pop *this* view and the still presented CNContactViewController in a single animation.
+        // TODO this is weird - ideally we'd do something like
+        //     self.delegate?.didFinishAddingContact
+        // and the delegate, which knows about our presentation context could do the right thing.
+        //
+        // As it is, we happen to always be *pushing* this view controller onto a navcontroller, so the
+        // following works in all current cases.
+        //
+        // If we ever wanted to do something different, like present this in a modal, we'd have to rethink.
+
+        // We want to pop *this* view *and* the still presented CNContactViewController in a single animation.
         // Note this happens for *cancel* and for *done*. Unfortunately, I don't know of a way to detect the difference
         // between the two, since both just call this method.
         guard let myIndex = navigationController.viewControllers.index(of: self) else {
--- a/Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
+++ b/Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
@@ -1,0 +1,121 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+import UIKit
+import ContactsUI
+
+class AddContactShareToExistingContactViewController: ContactsPicker, ContactsPickerDelegate, CNContactViewControllerDelegate {
+
+    // TODO actual protocol?
+    weak var addToExistingContactDelegate: UIViewController?
+
+    let contactShare: ContactShareViewModel
+
+    required init(contactShare: ContactShareViewModel) {
+        self.contactShare = contactShare
+        super.init(allowsMultipleSelection: false, subtitleCellType: .none)
+
+        self.contactsPickerDelegate = self
+    }
+
+    required public init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    @objc required public init(allowsMultipleSelection: Bool, subtitleCellType: SubtitleCellValue) {
+        fatalError("init(allowsMultipleSelection:subtitleCellType:) has not been implemented")
+    }
+
+    // MARK: - ContactsPickerDelegate
+
+    func contactsPicker(_: ContactsPicker, contactFetchDidFail error: NSError) {
+        owsFail("\(logTag) in \(#function) with error: \(error)")
+
+        guard let navigationController = self.navigationController else {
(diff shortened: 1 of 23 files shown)


Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
```diff
@@ -0,0 +1,133 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+import UIKit
+import ContactsUI
+
+class AddContactShareToExistingContactViewController: ContactsPicker, ContactsPickerDelegate, CNContactViewControllerDelegate {
```
michaelkirk-signal: Rather than subclass the ContactsPicker, I would have preferred to compose it, using the contactsPickerDelegate, but it seemed like that was going to require some bigger changes to the contactsShareViewHelper.

Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
```diff
@@ -0,0 +1,133 @@
...
+
+    func contactsPicker(_: ContactsPicker, contactFetchDidFail error: NSError) {
+        owsFail("\(logTag) in \(#function) with error: \(error)")
+
+        guard let navigationController = self.navigationController else {
+            owsFail("\(logTag) in \(#function) navigationController was unexpectedly nil")
+            return
+        }
+
+        navigationController.popViewController(animated: true)
```
michaelkirk-signal: e.g. it's kind of sketchy when a VC dismisses itself.

Signal/src/ViewControllers/ContactsPicker.swift
```diff
@@ -231,13 +219,16 @@ public class ContactsPicker: OWSViewController, UITableViewDelegate, UITableView
...
+            owsFail("\(logTag) in \(#function) cell had unexpected type")
+            return UITableViewCell()
+        }
 
         let dataSource = filteredSections
         let cnContact = dataSource[indexPath.section][indexPath.row]
         let contact = Contact(systemContact: cnContact)
 
-        cell.configure(contact: contact, subtitleType: subtitleCellValue, contactsManager: self.contactsManager)
+        cell.configure(contact: contact, subtitleType: subtitleCellType, showsWhenSelected: self.allowsMultipleSelection, contactsManager: self.contactsManager)
```
michaelkirk-signal: This is the only meaningful change in this file - the rest are just for clarity.

We now tell the ContactCell whether the tableView allowsMultipleSelection so that we only show the ✔️ accessory view when using multiselect.

In the case of single-select contact picker, it felt a little weird to see the checkmark appear just as the view disappears.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4978,11 +4979,13 @@ - (void)collectionView:(UICollectionView *)collectionView
 - (void)contactsPickerDidCancel:(ContactsPicker *)contactsPicker
 {
     DDLogDebug(@"%@ in %s", self.logTag, __PRETTY_FUNCTION__);
+    [self dismissViewControllerAnimated:YES completion:nil];
```
michaelkirk-signal: > // TODO - there are some hard coded assumptions in this VC that assume we are *pushed* onto a
> // navigation controller.

Here we see the inverse of what I was rambling about in the `AddContactShareToExisting` class. 

The ContactsPicker was similarly hardcoding assumptions about it's dismissal. Previously it was only ever used in a modal context, so all these dismisses were hardcoded into the ContactsPicker. Since *here* we were pushing the picker view, the existing dismissal logic didn't work.

On average, I think things work out better when the same view that pushes/presents a VC is responsible for popping/dismissing it - typically as part of a delegate method (even if it means some duplication).

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4992,27 +4995,29 @@ - (void)contactsPicker:(ContactsPicker *)contactsPicker didSelectContact:(Contac
...
+
     OWSContact *_Nullable contactShareRecord = [OWSContacts contactForSystemContact:contact.cnContact];
     if (!contactShareRecord) {
         DDLogError(@"%@ Could not convert system contact.", self.logTag);
         return;
     }
 
     BOOL isProfileAvatar = NO;
-    UIImage *_Nullable avatarImage = contact.image;
+    NSData *_Nullable avatarImageData = contact.imageData;
```
michaelkirk-signal: In order to get contact merging to work, I copped some ideas from your branch - we're now always passing around data to be sure we don't re-encode any images.

Signal/src/ViewControllers/InviteFlow.swift
```diff
@@ -146,7 +151,8 @@ class InviteFlow: NSObject, MFMessageComposeViewControllerDelegate, MFMailCompos
         return UIAlertAction(title: messageTitle, style: .default) { _ in
             Logger.debug("\(self.TAG) Chose message.")
             self.channel = .message
-            let picker = ContactsPicker(delegate: self, multiSelection: true, subtitleCellType: .phoneNumber)
+            let picker = ContactsPicker(allowsMultipleSelection: true, subtitleCellType: .phoneNumber)
+            picker.contactsPickerDelegate = self
```
michaelkirk-signal: Passing a delegate in an initializer can be kind of inflexible/dodgy. 

In this particular case, I wanted the subclass to be it's own delegate, which isn't really possible without implicitly unwrapped optionals (which sometimes crash).

But also, there can be times when the delegate is unintentionally messaged before `self` has been completely initialized.

e.g. see the unfortunate `shouldNotifyDelegateOfUpdatedContacts` dance we do in ContactsViewHelper.

In general I'm trying to default towards always assigning the delegate after the fact.

SignalServiceKit/src/Messages/Interactions/OWSContact.m
```diff
@@ -815,7 +812,8 @@ + (nullable NSData *)vCardDataContact:(OWSContact *)contact
 {
     OWSAssert(contact);
 
-    CNContact *_Nullable systemContact = [self systemContactForContact:contact];
+    // TODO pass in image for vcard
```
michaelkirk-signal: Note: this is still pending.

SignalServiceKit/src/Contacts/Contact.m
```diff
@@ -290,6 +290,63 @@ - (NSUInteger)hash
...
+    if (formattedFullName.length == 0) {
+        mergedCNContact.namePrefix = newCNContact.namePrefix.ows_stripped;
+        mergedCNContact.givenName = newCNContact.givenName.ows_stripped;
+        mergedCNContact.middleName = newCNContact.middleName.ows_stripped;
+        mergedCNContact.familyName = newCNContact.familyName.ows_stripped;
+        mergedCNContact.nameSuffix = newCNContact.nameSuffix.ows_stripped;
+    }
+    
+    // Phone Numbers
+    NSSet<PhoneNumber *> *existingPhoneNumberSet = [NSSet setWithArray:self.parsedPhoneNumbers];
```
michaelkirk-signal: Our system contact adapter already has a lot of tools for dealing with CNContact and their fields. So putting the merge code here was a good fit.

Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
```diff
@@ -0,0 +1,133 @@
...
+import UIKit
+import ContactsUI
+
+class AddContactShareToExistingContactViewController: ContactsPicker, ContactsPickerDelegate, CNContactViewControllerDelegate {
+
+    // TODO - there are some hard coded assumptions in this VC that assume we are *pushed* onto a
+    // navigation controller. That seems fine for now, but if we need to be presented as a modal,
+    // or need to notify our presenter about our dismisall or other contact actions, a delegate
+    // would be helpful. It seems like this would require some broad changes to the ContactShareViewHelper,
+    // so I've left it as is for now, since it happens to work.
```
charlesmchen-signal: I agree - for any non-trivial VC (e.g. VCs with delegates), it'd be nice to let the delegate handle dismissal.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: Agree - for VCs with delegates.

Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
michaelkirk-signal: What I mean is that we should be building view controllers such that adding a delegate is easy. To the point - in this case I couldn't easily do it because the presentation of this VC occurs as a result of a nested series of calls in a view helper class.

The solution isn't totally obvious to me - but I think the problem is: "What's a good way to extract view controller logic so it can be composed without inheritance".

PR #3413, opened 2018-05-10
Refine contact merging.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Contacts/Contact.m
+++ b/SignalServiceKit/src/Contacts/Contact.m
@@ -351,18 +351,18 @@ NS_ASSUME_NONNULL_BEGIN
     }
 
     // Phone Numbers
-    NSMutableSet<PhoneNumber *> *existingPhoneNumberSet = [NSMutableSet setWithArray:self.parsedPhoneNumbers];
-    [existingPhoneNumberSet addObjectsFromArray:self.userTextPhoneNumbers];
+    NSSet<PhoneNumber *> *existingParsedPhoneNumberSet = [NSSet setWithArray:self.parsedPhoneNumbers];
+    NSSet<NSString *> *existingUnparsedPhoneNumberSet = [NSSet setWithArray:self.userTextPhoneNumbers];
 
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Contacts/Contact.m
```diff
@@ -345,16 +345,28 @@ - (CNContact *)buildCNContactMergedWithNewContact:(CNContact *)newCNContact
         mergedCNContact.familyName = newCNContact.familyName.ows_stripped;
         mergedCNContact.nameSuffix = newCNContact.nameSuffix.ows_stripped;
     }
-    
+
+    if (mergedCNContact.organizationName.ows_stripped.length < 1) {
+        mergedCNContact.organizationName = newCNContact.organizationName.ows_stripped;
```
charlesmchen-signal: Handle organization name.

SignalServiceKit/src/Contacts/Contact.m
```diff
@@ -345,16 +345,28 @@ - (CNContact *)buildCNContactMergedWithNewContact:(CNContact *)newCNContact
...
+    }
+
     // Phone Numbers
-    NSSet<PhoneNumber *> *existingPhoneNumberSet = [NSSet setWithArray:self.parsedPhoneNumbers];
-    
+    NSMutableSet<PhoneNumber *> *existingPhoneNumberSet = [NSMutableSet setWithArray:self.parsedPhoneNumbers];
+    [existingPhoneNumberSet addObjectsFromArray:self.userTextPhoneNumbers];
```
charlesmchen-signal: Don't drop unparse-able phone numbers unless they match an existing unparsed phone number.

Phone parsing is hard and I suggest we should be permissive about preserving phone numbers we don't understand. WDYT?

michaelkirk-signal: That seems fine. 

PhoneNumberSet *was* a set of PhoneNumbers, now it looks like it's a set of PhoneNumber's and NSString's, can you remove the incorrect annotation? Or make it an NSObject.

charlesmchen-signal: Whoops, fixed.

PR #3417, opened 2018-05-11
Charlesmchen/build warnings

charlesmchen-signal:
Let's not merge this into v2.25.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -777,10 +777,7 @@ static NSTimeInterval launchStartedAt;
 
             OutboundCallInitiator *outboundCallInitiator = SignalApp.sharedApp.outboundCallInitiator;
             OWSAssert(outboundCallInitiator);
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wunused-result"
             [outboundCallInitiator initiateCallWithHandle:phoneNumber];
-#pragma GCC diagnostic pop
         }];
         return YES;
     } else if ([userActivity.activityType isEqualToString:@"INStartAudioCallIntent"]) {
@@ -824,10 +821,7 @@ static NSTimeInterval launchStartedAt;
 
             OutboundCallInitiator *outboundCallInitiator = SignalApp.sharedApp.outboundCallInitiator;
             OWSAssert(outboundCallInitiator);
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wunused-result"
             [outboundCallInitiator initiateCallWithHandle:phoneNumber];
-#pragma GCC diagnostic pop
         }];
         return YES;
     } else {
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -777,7 +777,10 @@ static NSTimeInterval launchStartedAt;
 
             OutboundCallInitiator *outboundCallInitiator = SignalApp.sharedApp.outboundCallInitiator;
             OWSAssert(outboundCallInitiator);
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-result"
             [outboundCallInitiator initiateCallWithHandle:phoneNumber];
+#pragma GCC diagnostic pop
         }];
         return YES;
     } else if ([userActivity.activityType isEqualToString:@"INStartAudioCallIntent"]) {
@@ -821,7 +824,10 @@ static NSTimeInterval launchStartedAt;
 
             OutboundCallInitiator *outboundCallInitiator = SignalApp.sharedApp.outboundCallInitiator;
             OWSAssert(outboundCallInitiator);
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-result"
             [outboundCallInitiator initiateCallWithHandle:phoneNumber];
+#pragma GCC diagnostic pop
         }];
         return YES;
     } else {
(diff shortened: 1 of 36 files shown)


SignalMessaging/categories/Promise+retainUntilComplete.swift
```diff
@@ -14,6 +14,7 @@ public extension AnyPromise {
         // compiler warning: "Variable 'retainCycle' was written to, but never read"
         var retainCycle: AnyPromise? = self
         self.always {
+            assert(retainCycle != nil)
```
charlesmchen-signal: I think it'll be nice to assert this - and it has the side effect of fixing the "variable assigned but never read" warning.

SignalMessaging/contacts/OWSContactsManager.m
```diff
@@ -19,7 +19,6 @@
 #import <SignalServiceKit/PhoneNumber.h>
 #import <SignalServiceKit/SignalAccount.h>
 
-@import AddressBook;
```
charlesmchen-signal: I think this is our last dependency on AddressBook?

SignalMessaging/contacts/OWSContactsManager.m
```diff
@@ -566,8 +565,12 @@ - (NSAttributedString *)formattedFullNameForRecipientId:(NSString *)recipientId
         NSAttributedString *lastName =
             [[NSAttributedString alloc] initWithString:cachedLastName attributes:lastNameAttributes];
 
+        CNMutableContact *formatContact = [CNMutableContact new];
+        formatContact.givenName = firstName.string;
+        formatContact.familyName = lastName.string;
+        CNContactDisplayNameOrder nameOrder = [CNContactFormatter nameOrderForContact:formatContact];
```
charlesmchen-signal: Docs are opaque about why `[CNContactFormatter nameOrderForContact:]` takes a contact. https://developer.apple.com/documentation/contacts/cncontactformatter/1397731-nameorderforcontact

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -786,6 +786,11 @@ - (void)shareTextAction
         }
         case OWSMessageCellType_Unknown: {
             OWSFail(@"%@ No text to share", self.logTag);
+            break;
+        }
+        case OWSMessageCellType_ContactShare: {
+            OWSFail(@"%@ share contact not implemented.", self.logTag);
```
charlesmchen-signal: I don't think we need these checks in production since `canSaveMedia` and `canPerformAction` should prevent us from ever reaching here.

charlesmchen-signal:
PTAL @michaelkirk

SignalServiceKit/src/Messages/DeviceSyncing/OWSSyncContactsMessage.m
```diff
@@ -30,16 +30,7 @@ - (instancetype)initWithSignalAccounts:(NSArray<SignalAccount *> *)signalAccount
                        identityManager:(OWSIdentityManager *)identityManager
                         profileManager:(id<ProfileManagerProtocol>)profileManager
 {
-    self = [super initOutgoingMessageWithTimestamp:[NSDate ows_millisecondTimeStamp]
```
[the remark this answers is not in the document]
charlesmchen-signal: `OWSOutgoingSyncMessage`'s designated initializer takes no args.

Signal/src/AppDelegate.m
```diff
@@ -772,7 +772,10 @@ - (BOOL)application:(UIApplication *)application
 
             OutboundCallInitiator *outboundCallInitiator = SignalApp.sharedApp.outboundCallInitiator;
             OWSAssert(outboundCallInitiator);
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-result"
             [outboundCallInitiator initiateCallWithHandle:phoneNumber];
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

SignalMessaging/attachments/SignalAttachment.swift
```diff
@@ -38,7 +38,7 @@ extension String {
 }
 
 extension SignalAttachmentError: LocalizedError {
-    public var errorDescription: String {
+    public var errorDescription: String? {
```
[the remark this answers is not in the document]
michaelkirk-signal: ahhhh, makes sense.

SignalMessaging/contacts/OWSContactsManager.m
charlesmchen-signal: > Can you verify this works with the various combinations of display vs. sort in Settings > Contacts?

Yup, this works.

> OWSContact *contact = [self contactForRecipientId:recipientId];

We use this method - `formattedFullNameForRecipientId` - to format recipient ids, for which we don't necessarily have a system contact.

> And in general, when we're touching CNContact*anything, I'd try to put it in our contact adapter. 

Ditto - there isn't necessarily an instance of `Contact`.

charlesmchen-signal: I've modified this to _try_ to use the `CNContact` if we have one, but to fall back to constructing one.

charlesmchen-signal:
PTAL @michaelkirk

charlesmchen-signal:
PTAL @michaelkirk

PR #3418, opened 2018-05-11
Fix crash converting images in SAE.

charlesmchen-signal:
We need to this thoroughly since we don't know exactly why using UIGraphicsBeginImageContext() crashes in share extension after using auth UI, but under some light testing this approach seems to resolve the issue.

PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/attachments/SignalAttachment.swift
+++ b/SignalMessaging/attachments/SignalAttachment.swift
@@ -17,6 +17,7 @@ enum SignalAttachmentError: Error {
     case couldNotConvertToMpeg4
     case couldNotRemoveMetadata
     case invalidFileFormat
+    case couldNotResizeImage
 }
 
 extension String {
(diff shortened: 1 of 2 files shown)


SignalMessaging/attachments/SignalAttachment.swift
```diff
@@ -746,20 +753,61 @@ public class SignalAttachment: NSObject {
...
+    // crashes reliably in the share extension after screen lock's auth UI has been presented.
+    // Resizing using a CGContext seems to work fine.
+    private class func imageScaled(_ uiImage: UIImage, toMaxSize size: CGFloat) -> UIImage? {
+        guard let cgImage = uiImage.cgImage else {
+            owsFail("\(logTag) UIImage missing cgImage.")
+            return nil
+        }
+
+        // It's essential that we work consistently in "CG" coordinates (which are
+        // and pixels and don't reflect orientation), not "UI" coordinates (which
```
michaelkirk-signal: > which are ~~and~~ pixels and don't...

SignalMessaging/attachments/SignalAttachment.swift
```diff
@@ -746,20 +753,61 @@ public class SignalAttachment: NSObject {
...
+        // are in points and do reflect orientation).
+        let srcWidth = CGFloat(cgImage.width)
+        let srcHeight = CGFloat(cgImage.height)
         var scaleFactor: CGFloat
-        let aspectRatio: CGFloat = image.size.height / image.size.width
+        let aspectRatio: CGFloat = srcWidth / srcHeight
```
michaelkirk-signal: So previously AR was height/width but now it's width/height.

Is that because of the CG Coord-> UI Coord  switch?

charlesmchen-signal: Nope, that was a typo.  I'm going to switch to using `AVMakeRectWithAspectRatioInsideRect()` to do this math to avoid this kind of thing.

SignalMessaging/attachments/SignalAttachment.swift
charlesmchen-signal: Done.

SignalMessaging/attachments/SignalAttachment.swift
michaelkirk-signal: Nice, didn't know about that one.

PR #3425, opened 2018-05-14
Fix edge cases in home view.

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -376,6 +376,10 @@ - (UIViewController *)previewingContext:(id<UIViewControllerPreviewing>)previewi
 {
     NSIndexPath *indexPath = [self.tableView indexPathForRowAtPoint:location];
 
+    if ([self isIndexPathForArchivedConversations:indexPath]) {
```
charlesmchen-signal: Fixes a crash when swiping on the "archived threads" row.

PR #3429, opened 2018-05-14
Improve handling of unregistered users.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/translations/en.lproj/Localizable.strings
+++ b/Signal/translations/en.lproj/Localizable.strings
@@ -857,7 +857,7 @@
 /* No comment provided by engineer. */
 "ERROR_MESSAGE_WRONG_TRUSTED_IDENTITY_KEY" = "Safety number changed. Tap to verify.";
 
-/* Format string for 'unregistered user' error. Embeds {{the unregistered user's signal id}}. */
+/* Format string for 'unregistered user' error. Embeds {{the unregistered user's name or signal id}}. */
 "ERROR_UNREGISTERED_USER_FORMAT" = "Unregistered User: %@";
 
 /* action sheet header when re-sending message which failed because of too many attempts */
--- a/Signal/translations/en.lproj/Localizable.strings
+++ b/Signal/translations/en.lproj/Localizable.strings
@@ -857,6 +857,9 @@
 /* No comment provided by engineer. */
 "ERROR_MESSAGE_WRONG_TRUSTED_IDENTITY_KEY" = "Safety number changed. Tap to verify.";
 
+/* Format string for 'unregistered user' error. Embeds {{the unregistered user's signal id}}. */
+"ERROR_UNREGISTERED_USER_FORMAT" = "Unregistered User: %@";
+
 /* action sheet header when re-sending message which failed because of too many attempts */
 "FAILED_SENDING_BECAUSE_RATE_LIMIT" = "Too many failures with this contact. Please try again shortly.";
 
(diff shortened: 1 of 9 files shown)


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -857,6 +857,9 @@
 /* No comment provided by engineer. */
 "ERROR_MESSAGE_WRONG_TRUSTED_IDENTITY_KEY" = "Safety number changed. Tap to verify.";
 
+/* Format string for 'unregistered user' error. Embeds {{the unregistered user's signal id}}. */
+"ERROR_UNREGISTERED_USER_FORMAT" = "Unregistered User: %@";
```
charlesmchen-signal: We now record which user was registered and include it in the user-facing message.

charlesmchen-signal: *unregistered

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -465,6 +469,12 @@ - (void)enqueueAttachment:(DataSource *)dataSource
             SignalRecipient *newRecipient = [self.contactsUpdater synchronousLookup:recipientId error:error];
             if (newRecipient) {
                 [recipients addObject:newRecipient];
+            } else {
+                DDLogWarn(@"%@ No SignalRecipient for recipientId: %@", self.logTag, recipientId);
+                [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
+                    // Mark this recipient as "skipped".
+                    [message updateWithSkippedRecipient:recipientId transaction:transaction];
```
charlesmchen-signal: Mark unregistered recipients as skipped.

michaelkirk-signal: Since the calling method is already responsible for doing async/dispatch and creating transactions, can we pass this transaction in rather than create one here?

It seems like it would be easy to break it, creating a nested transaction if the caller gets refactored, and not notice since this is not happy path.

SignalServiceKit/src/Messages/Interactions/TSInfoMessage.m
```diff
@@ -100,7 +117,14 @@ - (NSString *)previewTextWithTransaction:(YapDatabaseReadTransaction *)transacti
         case TSInfoMessageTypeUnsupportedMessage:
             return NSLocalizedString(@"UNSUPPORTED_ATTACHMENT", nil);
         case TSInfoMessageUserNotRegistered:
-            return NSLocalizedString(@"CONTACT_DETAIL_COMM_TYPE_INSECURE", nil);
+            if (self.unregisteredRecipientId.length > 0) {
+                return [NSString stringWithFormat:NSLocalizedString(@"ERROR_UNREGISTERED_USER_FORMAT",
+                                                      @"Format string for 'unregistered user' error. Embeds {{the "
+                                                      @"unregistered user's signal id}}."),
+                                 self.unregisteredRecipientId];
```
michaelkirk-signal: Why not use the display name?

charlesmchen-signal: Done.

On a side note, I'd like to eventually refactor `previewTextWithTransaction` to take a contact manager param.

SignalServiceKit/src/Messages/OWSMessageSender.m
charlesmchen-signal: Yup. I'm trying to use a narrow, short-lived transaction here since:

* It's an edge case.
* I don't want to bake in the assumption that `[contactsUpdater synchronousLookup]` doesn't use a transaction.

As I mentioned I'd like to refactor this sending logic soon for many reasons, and you're pointing to additional worthy goals for that refactor.  I've updated the card.

PR #3431, opened 2018-05-14
contact flow vs. navbar style issues

michaelkirk-signal:
PTAL @charlesmchen 

Two issues:

1. Make sure we always return the navbar to the correct style. There were certain ways to leave the contact flow yet see navbar labels rendered black.

The issue occurs when mutating NavBar appearance when a navbar is hidden. i.e. if we weren't hiding the navbars in the OWS ContactViewController, this wouldn't occur (obviously for the design we want to hide these navbars, but it explains why we haven't seen this before.)

2. Present the "add a contact" modally. Since the "Add a contact" view needs to use the navbar (for the "done" and "cancel" buttons), animating the navbar back in for this view controller looks janky. 

Tested on iOS 9.4, 10.3, and 11.3

### DIFF
--- a/Signal/src/ViewControllers/ContactShareViewHelper.swift
+++ b/Signal/src/ViewControllers/ContactShareViewHelper.swift
@@ -158,23 +158,18 @@ public class ContactShareViewHelper: NSObject, CNContactViewControllerDelegate {
         contactViewController.delegate = self
         contactViewController.allowsActions = false
         contactViewController.allowsEditing = true
-        contactViewController.navigationItem.leftBarButtonItem = UIBarButtonItem(title: CommonStrings.cancelButton, style: .plain, target: self, action: #selector(didFinishEditingContact))
         contactViewController.navigationItem.leftBarButtonItem = UIBarButtonItem(title: CommonStrings.cancelButton,
                                                                                  style: .plain,
                                                                                  target: self,
                                                                                  action: #selector(didFinishEditingContact))
 
-        guard let navigationController = fromViewController.navigationController else {
-            owsFail("\(logTag) missing navigationController")
-            return
-        }
-
-        navigationController.pushViewController(contactViewController, animated: true)
-
         // HACK otherwise CNContactViewController Navbar is shown as black.
         // RADAR rdar://28433898 http://www.openradar.me/28433898
         // CNContactViewController incompatible with opaque navigation bar
         UIUtil.applyDefaultSystemAppearence()
+
+        let modal = UINavigationController(rootViewController: contactViewController)
+        fromViewController.present(modal, animated: true)
     }
 
     private func presentSelectAddToExistingContactView(contactShare: ContactShareViewModel, fromViewController: UIViewController) {
@@ -208,6 +203,7 @@ public class ContactShareViewHelper: NSObject, CNContactViewControllerDelegate {
             return
         }
 
+        UIUtil.applySignalAppearence()
         delegate.didCreateOrEditContact()
     }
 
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -77,44 +75,38 @@ class ContactViewController: OWSViewController, ContactShareViewHelperDelegate {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
 
-        // Use "dark" style navigation bar in this view.
-        UIUtil.applyDefaultSystemAppearence()
+        // Use "dark" style status bar in this view,
+        // But we don't apply the entire default system UI (UIUtil.applyDefaultSystemAppearence) because
+        // doing so while pushing onto an existing nav controller who's navbars are hidden corrupts the
+        // title label/navbar items elsewhere in the stack.
+        CurrentAppContext().setStatusBarStyle(.default)
```
michaelkirk-signal: To avoid corrupting the hidden navbars label color, we only mutate the status bar style, which is the only side effect  of the `applySystemStyle` we needed here anyway.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -77,44 +75,38 @@ class ContactViewController: OWSViewController, ContactShareViewHelperDelegate {
...
 
         guard let navigationController = self.navigationController else {
             owsFail("\(logTag) in \(#function) navigationController was unexpectedly nil")
             return
         }
         navigationController.isNavigationBarHidden = true
 
-        // self.navigationController is nil in viewDidDisappear when transition via message/call buttons
-        // so we maintain our own reference.
-        self.postDismissNavigationController = navigationController
```
michaelkirk-signal: This was only needed if we're mutating the navigationController in view**Did**Disappear.

Signal/src/ViewControllers/ContactShareViewHelper.swift
```diff
@@ -208,6 +203,7 @@ public class ContactShareViewHelper: NSObject, CNContactViewControllerDelegate {
             return
         }
 
+        UIUtil.applySignalAppearence()
```
charlesmchen-signal: I'm not sure this is right, `ContactViewController`'s header has a dark background.  Therefore I was using"system appearance" aka `applyDefaultSystemAppearence` in `ContactViewController`.  

It looks like if we make this change, `ContactViewController` would have the wrong status bar appearance after editing a contact from it.

charlesmchen-signal: Sorry, disregard this... just saw below.

Signal/src/ViewControllers/ContactViewController.swift
charlesmchen-signal: Huh.  So does that imply that it's not always safe to use `applySignalAppearence` and `applyDefaultSystemAppearence`?

michaelkirk-signal:
> It sounds like you learned some things about nav bar appearance that I'd like to hear about. Let's discuss later.

> Huh. So does that imply that it's not always safe to use applySignalAppearence and applyDefaultSystemAppearence?

> The issue occurs when mutating NavBar appearance when a navbar is hidden. i.e. if we weren't hiding the navbars in the OWS ContactViewController, this wouldn't occur (obviously for the design we want to hide these navbars, but it explains why we haven't seen this before.)

So my best guess is that the navbar labels are updated eagerly when you apply appearance changes, but if the navbar is not visible, they miss their chance to get updated.

PR #3433, opened 2018-05-15
Only mark messages read in conversation view if visible.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -3798,6 +3798,15 @@ typedef enum : NSUInteger {
 
 - (void)markVisibleMessagesAsRead
 {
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3789,6 +3789,15 @@ - (void)updateLastVisibleTimestamp:(uint64_t)timestamp
 
 - (void)markVisibleMessagesAsRead
 {
+    if (self.presentedViewController) {
+        OWSFail(@"%@ Not marking messages as read; another view is presented.", self.logTag);
+        return;
+    }
+    if (self.navigationController.topViewController != self) {
```
charlesmchen-signal: This doesn't fix a bug - the behavior was already correct.  But this is important to get right so I want to future-proof us.

PR #3434, opened 2018-05-15
Rework cleanup of peer connection client delegates.

charlesmchen-signal:
This is not for v2.25.

PTAL @michaelkirk


charlesmchen-signal:
@michaelkirk I'm going to include this in v2.25.  It should be safe, and should help.

PR #3435, opened 2018-05-15
Build & enqueue network requests async.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Network/API/TSNetworkManager.m
+++ b/SignalServiceKit/src/Network/API/TSNetworkManager.m
@@ -68,6 +68,20 @@ typedef void (^failureBlock)(NSURLSessionDataTask *task, NSError *error);
     OWSAssert(successBlock);
     OWSAssert(failureBlock);
 
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        [self makeRequestAsync:request completionQueue:completionQueue success:successBlock failure:failureBlock];
+    });
+}
+
+- (void)makeRequestAsync:(TSRequest *)request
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Network/API/TSNetworkManager.m
```diff
@@ -68,6 +68,20 @@ - (void)makeRequest:(TSRequest *)request
     OWSAssert(successBlock);
     OWSAssert(failureBlock);
 
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        [self makeRequestAsync:request completionQueue:completionQueue success:successBlock failure:failureBlock];
+    });
+}
+
+- (void)makeRequestAsync:(TSRequest *)request
```
michaelkirk-signal: naming nit: Sometimes we have a sync and an async flavor of a method. E.g. run and runAsync.

This one is a little weird because `makeRequest` and `makeRequestAsync` are both mostly async. 

Also, I think you're (reasonably) trying to avoid renaming the public method to keep the diff small.

If anything, what you've named `makeRequestAsync` is *less* async than the `makeRequest` method.

I'm having a hard time coming up with a good name...

`doMakeRequest:` or `makeRequestHelper:` or `_makeRequest:`, while pretty obtuse, would at least not be *against* our current naming conventions.

Feel free to ignore this feedback if you can't come up with anything better.

charlesmchen-signal: Good point. How about simply `makeRequestSync`?

michaelkirk:
Sure, that’s better than anything I came up with.

PR #3439, opened 2018-05-16
Make contact share "bubble/card" reflect share contents.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal.xcodeproj/project.pbxproj
+++ b/Signal.xcodeproj/project.pbxproj
@@ -48,7 +48,7 @@
 		34330A5C1E787A9800DF2FB9 /* dripicons-v2.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 34330A5B1E787A9800DF2FB9 /* dripicons-v2.ttf */; };
 		34330A5E1E787BD800DF2FB9 /* ElegantIcons.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 34330A5D1E787BD800DF2FB9 /* ElegantIcons.ttf */; };
 		34330AA31E79686200DF2FB9 /* OWSProgressView.m in Sources */ = {isa = PBXBuildFile; fileRef = 34330AA21E79686200DF2FB9 /* OWSProgressView.m */; };
-		34382266209A4E400094FEB7 /* ApproveContactShareViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 34382265209A4E3F0094FEB7 /* ApproveContactShareViewController.swift */; };
+		34382266209A4E400094FEB7 /* ContactShareApprovalViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 34382265209A4E3F0094FEB7 /* ContactShareApprovalViewController.swift */; };
 		3438226A209B63500094FEB7 /* EditContactShareNameViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 34382269209B63500094FEB7 /* EditContactShareNameViewController.swift */; };
 		34386A51207D0C01009F5D9C /* HomeViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = 34386A4D207D0C01009F5D9C /* HomeViewController.m */; };
 		34386A52207D0C01009F5D9C /* HomeViewCell.m in Sources */ = {isa = PBXBuildFile; fileRef = 34386A50207D0C01009F5D9C /* HomeViewCell.m */; };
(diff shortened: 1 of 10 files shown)


Signal.xcodeproj/project.pbxproj
```diff
@@ -48,7 +48,7 @@
 		34330A5C1E787A9800DF2FB9 /* dripicons-v2.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 34330A5B1E787A9800DF2FB9 /* dripicons-v2.ttf */; };
 		34330A5E1E787BD800DF2FB9 /* ElegantIcons.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 34330A5D1E787BD800DF2FB9 /* ElegantIcons.ttf */; };
 		34330AA31E79686200DF2FB9 /* OWSProgressView.m in Sources */ = {isa = PBXBuildFile; fileRef = 34330AA21E79686200DF2FB9 /* OWSProgressView.m */; };
-		34382266209A4E400094FEB7 /* ApproveContactShareViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 34382265209A4E3F0094FEB7 /* ApproveContactShareViewController.swift */; };
+		34382266209A4E400094FEB7 /* ContactShareApprovalViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 34382265209A4E3F0094FEB7 /* ContactShareApprovalViewController.swift */; };
```
charlesmchen-signal: I renamed this class to reflect the naming convention we use in `MessageApprovalViewController` and `AttachmentApprovalViewController`.

Signal/src/ViewControllers/ConversationView/Cells/OWSContactShareView.m
```diff
@@ -188,8 +188,9 @@ - (void)createContents
     [contentView autoPinEdgeToSuperviewEdge:ALEdgeTop withInset:self.vMargin];
 
     AvatarImageView *avatarView = [AvatarImageView new];
-    avatarView.image =
-        [self.contactShare getAvatarImageWithDiameter:self.iconSize contactsManager:self.contactsManager];
+    avatarView.image = [self.contactShare getAvatarImageWithDiameter:self.iconSize
+                                                     contactsManager:self.contactsManager
+                                             ignoreContactAndProfile:YES];
```
charlesmchen-signal: This is the bug fix.  Previously, it would often look like you sent an avatar even if you deselected the avatar field in the "contact share approval" view, because `ContactShareViewModel.getAvatarImage()` uses `OWSContactAvatarBuilder` which tries to fill in avatars using profile and system contact avatars from the contacts manager.  

The (design) change is to have the contact share "bubbles/cards" more closely reflect the contact share's contents to avoid this (possibly troubling) confusion.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -237,7 +237,7 @@ class ContactViewController: OWSViewController, ContactShareViewHelperDelegate {
 
         let avatarSize: CGFloat = 100
         let avatarView = AvatarImageView()
-        avatarView.image = contactShare.getAvatarImage(diameter: avatarSize, contactsManager: contactsManager)
+        avatarView.image = contactShare.getAvatarImage(diameter: avatarSize, contactsManager: contactsManager, ignoreContactAndProfile: true)
```
charlesmchen-signal: See below.

Signal.xcodeproj/project.pbxproj
michaelkirk-signal: 👍

SignalMessaging/ViewModels/ContactShareViewModel.swift
```diff
@@ -59,6 +59,7 @@ public class ContactShareViewModel: NSObject {
         let avatarBuilder = OWSContactAvatarBuilder(nonSignalName: displayName,
                                                     colorSeed: colorSeed,
                                                     diameter: UInt(diameter),
+                                                    ignoreContactAndProfile: ignoreContactAndProfile,
                                                     contactsManager: contactsManager)
         return avatarBuilder.build()
```
michaelkirk-signal: So we re-used our existing avatar builder, and have found that it doesn't do *quite* what we need it to do in this circumstance.

Adding a "mode" flag with a default is a quick fix which doesn't alter other code paths, but at the cost of accruing more complexity in the codebase.

What if instead, we revert all the `ignoreContactAndProfile` mode stuff, and instead here we call `avatarBuilder.buildDefaultImage()` directly?

michaelkirk:
Lgtm

PR #3440, opened 2018-05-16
Sort group members.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ThreadSettings/ShowGroupMembersViewController.m
+++ b/Signal/src/ViewControllers/ThreadSettings/ShowGroupMembersViewController.m
@@ -168,7 +168,14 @@ NS_ASSUME_NONNULL_BEGIN
 
     __weak ShowGroupMembersViewController *weakSelf = self;
     ContactsViewHelper *helper = self.contactsViewHelper;
-    for (NSString *recipientId in [recipientIds sortedArrayUsingSelector:@selector(compare:)]) {
+    // Sort the group members using contacts manager.
+    NSArray<NSString *> *sortedRecipientIds =
+        [recipientIds sortedArrayUsingComparator:^NSComparisonResult(NSString *recipientIdA, NSString *recipientIdB) {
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/ThreadSettings/ShowGroupMembersViewController.m
```diff
@@ -168,7 +168,14 @@ - (void)addMembers:(NSArray<NSString *> *)recipientIds
 
     __weak ShowGroupMembersViewController *weakSelf = self;
     ContactsViewHelper *helper = self.contactsViewHelper;
-    for (NSString *recipientId in [recipientIds sortedArrayUsingSelector:@selector(compare:)]) {
+    // Sort the group members using contacts manager.
```
charlesmchen-signal: We were sorting by recipient id which was ... at least a stable ordering.

SignalMessaging/Views/ContactTableViewCell.m
```diff
@@ -78,17 +78,14 @@ - (void)configureProgrammatically
 
     _nameLabel = [UILabel new];
     _nameLabel.lineBreakMode = NSLineBreakByTruncatingTail;
-    _nameLabel.font = [UIFont ows_dynamicTypeBodyFont];
```
charlesmchen-signal: I noticed that we weren't updating these cells to reflect changes to dynamic type.

SignalMessaging/contacts/OWSContactsManager.h
```diff
@@ -70,6 +70,7 @@ extern NSString *const OWSContactsManagerSignalAccountsDidChangeNotification;
  * Used for sorting, respects system contacts name sort order preference.
  */
 - (NSString *)comparableNameForSignalAccount:(SignalAccount *)signalAccount;
+- (NSComparisonResult)compareSignalAccount:(SignalAccount *)left withSignalAccount:(SignalAccount *)right;
```
charlesmchen-signal: It might be better to expose a sortRecipientIds method.

PR #3443, opened 2018-05-17
Use raw text when localization fails

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Contacts/Contact.m
+++ b/SignalServiceKit/src/Contacts/Contact.m
@@ -395,6 +395,27 @@ NS_ASSUME_NONNULL_BEGIN
     return [mergedCNContact copy];
 }
 
++ (nullable NSString *)localizedStringForCNLabel:(nullable NSString *)cnLabel
+{
+    if (cnLabel.length == 0) {
+        return nil;
+    }
+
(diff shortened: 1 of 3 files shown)


SignalServiceKit/src/Contacts/Contact.m
```diff
@@ -395,6 +395,27 @@ - (CNContact *)buildCNContactMergedWithNewContact:(CNContact *)newCNContact
...
+        return nil;
+    }
+
+    NSString *_Nullable localizedLabel = [CNLabeledValue localizedStringForLabel:cnLabel];
+
+    // Docs for localizedStringForLabel say it returns:
+    // > The localized string if a Contacts framework defined label, otherwise just returns the label.
+    // But in practice, at least on iOS11, if the label is not one of CNContacts known labels (like CNLabelHome)
+    // kUnlocalizedStringLabel is returned, rather than the unadultered label.
+    NSString *const kUnlocalizedStringLabel = @"__ABUNLOCALIZEDSTRING";
```
charlesmchen-signal: This is probably fine, just throwing it out there: We could also test that `cnLabel` is in a set of all known CNContact labels and not even try `localizedStringForLabel` if it isn't.

PR #3448, opened 2018-05-17
Improve logging around image resize failed.

charlesmchen-signal:
PTAL @michaelkirk


SignalMessaging/attachments/SignalAttachment.swift
```diff
@@ -704,6 +704,7 @@ public class SignalAttachment: NSObject {
             if image.size.width > maxSize ||
                 image.size.height > maxSize {
                 guard let resizedImage = imageScaled(image, toMaxSize: maxSize) else {
+                    owsFail("\(logTag) resize image failed.")
```
michaelkirk-signal: side note: I think, on average, our assertion failures are log-worthy. I think we should invert the semantics of our objc macros to match our swift ones.

Specifically, rather than OWSFail and OWSProdLogAndFail, we could have OWSFail log by default, and add an OWSFailNoLog() macro for the rare cases where we don't want to log an assertion failure.

PR #3450, opened 2018-05-17
Only clean up messages while active

michaelkirk-signal:
Most of the 10deadCC crashes reported in beta were (6/8) happened while the the DisappearingMessages#runLoop was called - all of them in response to `fallbackTimerDidFire`.

You can see from the stack traces that we're crashing within the readWrite transaction - meaning we already have a background task started, so I didn't bother adding another one here.

My theory is that sometimes this timer is firing just as we're moving from background -> suspended, otherwise, if this timer is fired before we're in the background, the time allocated by the readWrite transaction's background task should be sufficient.

The counter theory is that for some reason this task is just taking much longer than the background task allows (>~30s).

### DIFF
--- a/SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
+++ b/SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
@@ -364,6 +364,10 @@ void AssertIsOnDisappearingMessagesQueue()
     }
 
     dispatch_async(OWSDisappearingMessagesJob.serialQueue, ^{
+        if (!CurrentAppContext().isMainAppAndActive) {
+            DDLogInfo(@"%@ Ignoring fallbacktimer for app which is not main and active.", self.logTag);
+            return;
+        }
         NSUInteger deletedCount = [self runLoop];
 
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -364,6 +364,10 @@ - (void)fallbackTimerDidFire
     }
 
     dispatch_async(OWSDisappearingMessagesJob.serialQueue, ^{
+        if (!CurrentAppContext().isMainAppAndActive) {
```
michaelkirk-signal: I thought this was easier to reason about than invalidating/restarting the timer on app activation/deactivation.

charlesmchen-signal: Maybe we should (crudely) "time" the perf of this work using NSDate or something and do "verbose" (or higher) logging if it takes more than 5 seconds?

charlesmchen-signal:
@michaelkirk didn't this get merged?

michaelkirk-signal:
This was already merged in be5a80e65 and touched up with 2f8b5b8ea

PR #3451, opened 2018-05-17
resize call navbar

michaelkirk-signal:
PTAL @charlesmchen 

Screenshots forthcoming...


michaelkirk-signal:
This isn't quite right on iPhoneX, closing for now.

PR #3453, opened 2018-05-18
Fix "group avatar doesn't update after local change."

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -3835,6 +3835,8 @@ typedef enum : NSUInteger {
         [message updateWithCustomMessage:updateGroupInfo transaction:transaction];
     }];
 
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3835,6 +3835,8 @@ - (void)updateGroupModelTo:(TSGroupModel *)newGroupModel successCompletion:(void
         [message updateWithCustomMessage:updateGroupInfo transaction:transaction];
     }];
 
+    [groupThread fireAvatarChangedNotification];
```
charlesmchen-signal: `AvatarImageView` wasn't updating because avatar updates in conversation settings don't use `updateAvatarWithAttachmentStream`.  This isn't a perfect solution.  We could also fire the notification anytime a group thread is saved, period.

PR #3454, opened 2018-05-18
Charlesmchen/call lifecycle edge cases

charlesmchen-signal:
This fixes at least one reproducible error where ice updates arriving _after_ a call is terminated taint the next call.  

By gather all per-call state in `SignalCallData`,  we can easily reason about it's lifecycle:

* All per-call state is cleaned up when the call is terminated.
* The futures/promises are all created when the call begins.
* The futures/promises are all rejected if necessary when the call is terminated.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -253,47 +253,32 @@ private class SignalCallData: NSObject {
         }
     }
 
-    // TODO: Remove?
     var call: SignalCall? {
         get {
             SwiftAssertIsOnMainThread(#function)
 
-            guard let callData = callData else {
-                return nil
-            }
-            return callData.call
+            return callData?.call
         }
     }
-
     var peerConnectionClient: PeerConnectionClient? {
         get {
             SwiftAssertIsOnMainThread(#function)
 
-            guard let callData = callData else {
-            return nil
-            }
-            return callData.peerConnectionClient
+            return callData?.peerConnectionClient
         }
     }
-
     var localVideoTrack: RTCVideoTrack? {
         get {
             SwiftAssertIsOnMainThread(#function)
 
-            guard let callData = callData else {
-                return nil
-            }
-            return callData.localVideoTrack
+            return callData?.localVideoTrack
         }
     }
-    weak var remoteVideoTrack: RTCVideoTrack? {
+    var remoteVideoTrack: RTCVideoTrack? {
         get {
             SwiftAssertIsOnMainThread(#function)
 
-            guard let callData = callData else {
-                return nil
-            }
-            return callData.remoteVideoTrack
+            return callData?.remoteVideoTrack
         }
     }
     var isRemoteVideoEnabled: Bool {
@@ -832,7 +817,6 @@ private class SignalCallData: NSObject {
 
             let iceUpdateMessage = OWSCallIceUpdateMessage(callId: call.signalingId, sdp: iceCandidate.sdp, sdpMLineIndex: iceCandidate.sdpMLineIndex, sdpMid: iceCandidate.sdpMid)
 
-            Logger.info("\(self.logTag) in \(#function) sending ICE Candidate.")
             Logger.info("\(self.logTag) in \(#function) sending ICE Candidate \(call.identifiersForLogs).")
             let callMessage = OWSOutgoingCallMessage(thread: call.thread, iceUpdateMessage: iceUpdateMessage)
             let sendPromise = self.messageSender.sendPromise(message: callMessage)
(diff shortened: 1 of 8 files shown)


Signal/src/call/CallService.swift
```diff
@@ -1511,44 +1548,16 @@ protocol CallServiceObserver: class {
 
         Logger.debug("\(self.logTag) in \(#function)")
 
-        self.localVideoTrack = nil
-        self.remoteVideoTrack = nil
-        self.isRemoteVideoEnabled = false
+        // Capture a reference to the current call data,
```
michaelkirk-signal: TIOLI: I don't think this comment adds any value and just runs the risk of going out of date. The code is nice and clear as is.

Signal/src/call/CallService.swift
```diff
@@ -98,6 +98,108 @@ protocol CallServiceObserver: class {
...
+            SwiftAssertIsOnMainThread(#function)
+
+            Logger.debug("\(self.logTag) .peerConnectionClient setter: \(oldValue != nil) -> \(peerConnectionClient != nil) \(String(describing: peerConnectionClient))")
+        }
+    }
+
+    required init(call: SignalCall) {
+        self.call = call
+
+        let (callConnectedPromise, fulfillCallConnectedPromise, rejectCallConnectedPromise) = Promise<Void>.pending()
```
michaelkirk-signal: I really like having all these per call callbacks centralized like this. Good work!

Signal/src/call/CallService.swift
```diff
@@ -122,85 +224,86 @@ protocol CallServiceObserver: class {
...
-            fireDidUpdateVideoTracks()
+            guard let callData = callData else {
+                return nil
+            }
+            return callData.localVideoTrack
         }
     }
-    var isRemoteVideoEnabled = false {
-        didSet {
+    weak var remoteVideoTrack: RTCVideoTrack? {
```
michaelkirk-signal: I think this `weak` was copied over - now that videoTrack lives on the callData (which I think is a good call!) we don't need it.

Signal/src/call/CallService.swift
```diff
@@ -360,22 +460,17 @@ protocol CallServiceObserver: class {
...
+        guard let callData = self.callData else {
             self.handleFailedCall(failedCall: call, error: .obsoleteCall(description:"obsolete call in \(#function)"))
             return
         }
 
-        if self.fulfillReadyToSendIceUpdatesPromise == nil {
-            createReadyToSendIceUpdatesPromise()
-        }
-
-        guard let fulfillReadyToSendIceUpdatesPromise = self.fulfillReadyToSendIceUpdatesPromise else {
```
michaelkirk-signal: Cool that we can just get rid of this. 👍

Signal/src/call/CallService.swift
```diff
@@ -706,15 +807,16 @@ protocol CallServiceObserver: class {
...
             OWSProdError(OWSAnalyticsEvents.callServiceCallMissing(), file: #file, function: #function, line: #line)
             self.handleFailedCurrentCall(error: CallError.assertionError(description: "ignoring local ice candidate, since there is no current call."))
             return
         }
+        let call = callData.call
 
         // Wait until we've sent the CallOffer before sending any ice updates for the call to ensure
         // intuitive message ordering for other clients.
-        waitUntilReadyToSendIceUpdates().then { () -> Void in
+        callData.readyToSendIceUpdatesPromise.then { () -> Void in
```
michaelkirk-signal: much better!

Signal/src/call/CallService.swift
```diff
@@ -731,6 +833,7 @@ protocol CallServiceObserver: class {
             let iceUpdateMessage = OWSCallIceUpdateMessage(callId: call.signalingId, sdp: iceCandidate.sdp, sdpMLineIndex: iceCandidate.sdpMLineIndex, sdpMid: iceCandidate.sdpMid)
 
             Logger.info("\(self.logTag) in \(#function) sending ICE Candidate.")
```
michaelkirk-signal: can probably delete this first logger.

Signal/src/call/CallService.swift
```diff
@@ -915,6 +1018,12 @@ protocol CallServiceObserver: class {
         Logger.info("\(self.logTag) in \(#function)")
         SwiftAssertIsOnMainThread(#function)
 
+        guard let callData = self.callData else {
```
michaelkirk-signal: This is a little unfortunate - if the caller used to have a call, presumably it could instead pass us the CallData instead?

i.e. Can we remove this code and change the parameter to **be** the CallData?

Signal/src/call/CallService.swift
```diff
@@ -122,85 +224,86 @@ protocol CallServiceObserver: class {
...
+    var call: SignalCall? {
+        get {
+            SwiftAssertIsOnMainThread(#function)
 
-    // Used by waituntilReadyToSendIceUpdates to make sure CallOffer was 
-    // sent before sending any ICE updates.
-    private var fulfillReadyToSendIceUpdatesPromise: (() -> Void)?
-    private var rejectReadyToSendIceUpdatesPromise: ((Error) -> Void)?
-    private var readyToSendIceUpdatesPromise: Promise<Void>?
+            guard let callData = callData else {
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

Signal/src/call/CallService.swift
charlesmchen-signal: Removed.

Signal/src/call/CallService.swift
charlesmchen-signal: Removed.

Signal/src/call/CallService.swift
charlesmchen-signal: Done.

PR #3457, opened 2018-05-21
Charlesmchen/weak peer connection client

charlesmchen-signal:

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -749,8 +749,7 @@ private class SignalCallData: NSObject {
         Logger.verbose("\(logTag) \(#function) callId: \(callId)")
 
         guard let callData = self.callData else {
-            OWSProdError(OWSAnalyticsEvents.callServiceCallMissing(), file: #file, function: #function, line: #line)
-            self.handleFailedCurrentCall(error: .obsoleteCall(description: "ignoring remote ice update, since there is no current call."))
+            Logger.info("\(logTag) ignoring remote ice update, since there is no current call.")
             return
         }
 
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -449,9 +449,8 @@ private class SignalCallData: NSObject {
             self.handleFailedCall(failedCall: call, error: .obsoleteCall(description:"obsolete call in \(#function)"))
             return
         }
-
         guard callData.call == call else {
-            self.handleFailedCall(failedCall: call, error: .obsoleteCall(description:"obsolete call in \(#function)"))
+            Logger.warn("\(self.logTag) ignoring \(#function) for call other than current call")
             return
         }
 
@@ -746,6 +745,7 @@ private class SignalCallData: NSObject {
      * Remote client (could be caller or callee) sent us a connectivity update
      */
     public func handleRemoteAddedIceCandidate(thread: TSContactThread, callId: UInt64, sdp: String, lineIndex: Int32, mid: String) {
+        SwiftAssertIsOnMainThread(#function)
         Logger.verbose("\(logTag) \(#function) callId: \(callId)")
 
         guard let callData = self.callData else {
@@ -1001,7 +1001,7 @@ private class SignalCallData: NSObject {
         Logger.info("\(self.logTag) in \(#function)")
         SwiftAssertIsOnMainThread(#function)
 
-        guard let peerConnectionClient = self.peerConnectionClient else {
+        guard let peerConnectionClient = callData.peerConnectionClient else {
             OWSProdError(OWSAnalyticsEvents.callServicePeerConnectionMissing(), file: #file, function: #function, line: #line)
             handleFailedCall(failedCall: call, error: CallError.assertionError(description: "\(self.logTag) peerConnectionClient unexpectedly nil in \(#function)"))
             return
--- a/Signal/src/call/PeerConnectionClient.swift
+++ b/Signal/src/call/PeerConnectionClient.swift
@@ -412,7 +412,7 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
             }
 
             peerConnection.offer(for: strongSelf.defaultOfferConstraints, completionHandler: { (sdp: RTCSessionDescription?, error: Error?) in
-                PeerConnectionClient.signalingQueue.async { [weak self] in
+                PeerConnectionClient.signalingQueue.async {
                     completion(sdp, error)
                 }
             })
@@ -549,7 +549,7 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
             Logger.debug("\(strongSelf.logTag) negotiating answer session.")
 
             peerConnection.answer(for: constraints, completionHandler: { (sdp: RTCSessionDescription?, error: Error?) in
-                PeerConnectionClient.signalingQueue.async { [weak self] in
+                PeerConnectionClient.signalingQueue.async {
                     completion(sdp, error)
                 }
             })
--- a/Signal/src/call/PeerConnectionClient.swift
+++ b/Signal/src/call/PeerConnectionClient.swift
@@ -174,7 +174,9 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     private func createSignalingDataChannel() {
         SwiftAssertIsOnMainThread(#function)
-
+        defer {
+            ensureTeardown(#function)
+        }
         guard let peerConnection = peerConnection else {
             owsFail("\(logTag) in \(#function) peerConnection was unexpectedly nil")
             return
@@ -195,6 +197,9 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     fileprivate func createVideoSender() {
         SwiftAssertIsOnMainThread(#function)
+        defer {
+            ensureTeardown(#function)
+        }
 
         Logger.debug("\(logTag) in \(#function)")
         assert(self.videoSender == nil, "\(#function) should only be called once.")
@@ -240,9 +245,7 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
             guard let strongSelf = self else { return }
             Logger.info("\(strongSelf.logTag) \(#function) starting."); Logger.flush()
             defer {
-                DispatchQueue.main.async {
-                    Logger.info("\(strongSelf.logTag) \(#function) complete."); Logger.flush()
-                }
+                strongSelf.ensureTeardown(#function)
             }
             guard let localVideoSource = strongSelf.localVideoSource else {
                 owsFail("\(strongSelf.logTag) in \(#function) localVideoSource was unexpectedly nil")
@@ -261,9 +264,15 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     public func setLocalVideoEnabled(enabled: Bool) {
         SwiftAssertIsOnMainThread(#function)
+        defer {
+            ensureTeardown(#function)
+        }
 
         let completion = { [weak self] in
             guard let strongSelf = self else { return }
+            defer {
+                strongSelf.ensureTeardown(#function)
+            }
             guard let localVideoTrack = strongSelf.localVideoTrack else { return }
             guard let strongDelegate = strongSelf.delegate else { return }
             strongDelegate.peerConnectionClient(strongSelf, didUpdateLocal: enabled ? localVideoTrack : nil)
@@ -273,9 +282,7 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
             guard let strongSelf = self else { return }
             Logger.info("\(strongSelf.logTag) \(#function) starting."); Logger.flush()
             defer {
-                DispatchQueue.main.async {
-                    Logger.info("\(strongSelf.logTag) \(#function) complete."); Logger.flush()
-                }
+                strongSelf.ensureTeardown(#function)
             }
             guard strongSelf.peerConnection != nil else {
                 Logger.debug("\(strongSelf.logTag) \(#function) Ignoring obsolete event in terminated client")
@@ -309,6 +316,9 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     fileprivate func createAudioSender() {
         SwiftAssertIsOnMainThread(#function)
+        defer {
+            ensureTeardown(#function)
+        }
 
         Logger.debug("\(logTag) in \(#function)")
         assert(self.audioSender == nil, "\(#function) should only be called once.")
@@ -336,21 +346,16 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     public func setAudioEnabled(enabled: Bool) {
         SwiftAssertIsOnMainThread(#function)
-
-        Logger.info("\(self.logTag) setAudioEnabled 1.")
-        Logger.flush()
+        defer {
+            ensureTeardown(#function)
+        }
 
         PeerConnectionClient.signalingQueue.async { [weak self] in
             guard let strongSelf = self else { return }
             Logger.info("\(strongSelf.logTag) \(#function) starting."); Logger.flush()
             defer {
-                DispatchQueue.main.async {
-                    Logger.info("\(strongSelf.logTag) \(#function) complete."); Logger.flush()
-                }
+                strongSelf.ensureTeardown(#function)
             }
-
-            Logger.info("\(strongSelf.logTag) setAudioEnabled 2.")
-            Logger.flush()
 
             guard strongSelf.peerConnection != nil else {
                 Logger.debug("\(strongSelf.logTag) \(#function) Ignoring obsolete event in terminated client")
@@ -363,9 +368,6 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
             }
 
             audioTrack.isEnabled = enabled
-
-            Logger.info("\(strongSelf.logTag) setAudioEnabled 3.")
-            Logger.flush()
         }
     }
 
@@ -381,14 +383,15 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     public func createOffer() -> Promise<HardenedRTCSessionDescription> {
         SwiftAssertIsOnMainThread(#function)
+        defer {
+            ensureTeardown(#function)
+        }
 
         let completion: ((((HardenedRTCSessionDescription) -> Void), ((Error) -> Void), RTCSessionDescription?, Error?) -> Void) = { [weak self] (fulfill, reject, sdp, error) in
             guard let strongSelf = self else { return }
             Logger.info("\(strongSelf.logTag) \(#function) starting."); Logger.flush()
             defer {
-                DispatchQueue.main.async {
-                    Logger.info("\(strongSelf.logTag) \(#function) complete."); Logger.flush()
-                }
+                strongSelf.ensureTeardown(#function)
             }
             strongSelf.assertOnSignalingQueue()
             guard strongSelf.peerConnection != nil else {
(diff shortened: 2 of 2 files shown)


Signal/src/call/CallService.swift
```diff
@@ -449,9 +449,8 @@ private class SignalCallData: NSObject {
             self.handleFailedCall(failedCall: call, error: .obsoleteCall(description:"obsolete call in \(#function)"))
             return
         }
-
         guard callData.call == call else {
-            self.handleFailedCall(failedCall: call, error: .obsoleteCall(description:"obsolete call in \(#function)"))
+            Logger.warn("\(self.logTag) ignoring \(#function) for call other than current call")
```
charlesmchen-signal: `handleFailedCall` kills the current call, if any, in an attempt to return to a known good state.  We shouldn't do this in this case which is out of the happy path but can be safely ignored.

Signal/src/call/CallService.swift
```diff
@@ -746,11 +745,11 @@ private class SignalCallData: NSObject {
      * Remote client (could be caller or callee) sent us a connectivity update
      */
     public func handleRemoteAddedIceCandidate(thread: TSContactThread, callId: UInt64, sdp: String, lineIndex: Int32, mid: String) {
+        SwiftAssertIsOnMainThread(#function)
         Logger.verbose("\(logTag) \(#function) callId: \(callId)")
 
         guard let callData = self.callData else {
-            OWSProdError(OWSAnalyticsEvents.callServiceCallMissing(), file: #file, function: #function, line: #line)
-            self.handleFailedCurrentCall(error: .obsoleteCall(description: "ignoring remote ice update, since there is no current call."))
+            Logger.info("\(logTag) ignoring remote ice update, since there is no current call.")
```
charlesmchen-signal: Ditto.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -74,7 +74,31 @@ protocol PeerConnectionClientDelegate: class {
  */
 class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelDelegate {
 
-    let TAG = "[PeerConnectionClient]"
+    private class AtomicHandle<ValueType> : NSObject {
```
charlesmchen-signal: In Swift, weak variables aren't thread safe.  It isn't safe to resolve/acquire/lock a weak reference into a strong reference while the instance might be being deallocated on another thread.  

`AtomicHandle` provides thread-safe access to a strong reference.  `PeerConnectionClient` has an `AtomicHandle` to itself that its many async blocks (which run on more than one thread) can use to safely try to acquire a strong reference to the `PeerConnectionClient`.  In ARC we'd normally, we'd avoid having an instance retain a strong reference to itself to avoid retain cycles, but it's safe in this case: `PeerConnectionClient` is owned (and only used by) a single entity `CallService` and `CallService` always calls `[PeerConnectionClient terminate]` when it is done with a `PeerConnectionClient` instance, so `terminate` is a reliable place where we can break the retain cycle.

charlesmchen-signal: Note: this fixes the symptom of `EXC_BAD_ACCESS` during `PeerConnectionClient` teardown which I'd seen for a long time but misattributed to WebRTC because WebRTC entities like `RTCVideoTrack` cause similar errors if they are not deallocated in a timely when during call teardown.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -121,13 +145,11 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
     // RTCVideoTrack is fragile and prone to throwing exceptions and/or
     // causing deadlock in its destructor.  Therefore we take great care
     // with this property.
-    //
-    // We synchronize access to this property and ensure that we never
-    // set or use a strong reference to the remote video track if 
-    // peerConnection is nil.
```
charlesmchen-signal: We were previously (trying to) synchronize access to `remoteVideoTrack`.  This isn't necessary; we only access `remoteVideoTrack` on the main thread.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -167,13 +193,19 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
     }
 
     deinit {
-        Logger.debug("[PeerConnectionClient] deinit")
+        // TODO: We can demote this log level to debug once we're confident that
+        // this class is always deallocated.
+        Logger.info("[PeerConnectionClient] deinit")
```
charlesmchen-signal: Note: there are still cases where `PeerConnectionClient` isn't always deallocated.  I'll try to address this in a future PR.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -95,7 +119,7 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     // Connection
 
-    private var peerConnection: RTCPeerConnection!
+    private var peerConnection: RTCPeerConnection?
```
charlesmchen-signal: `peerConnection` is cleared after `terminateInternal`, so this needs to be optional.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -227,15 +262,17 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
     public func setCameraSource(useBackCamera: Bool) {
         SwiftAssertIsOnMainThread(#function)
 
-        PeerConnectionClient.signalingQueue.async {
-            guard let localVideoSource = self.localVideoSource else {
-                owsFail("\(self.logTag) in \(#function) localVideoSource was unexpectedly nil")
+        let handle = self.handle
```
charlesmchen-signal: We want to close over/capture the handle, not self.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -227,15 +262,17 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
+        PeerConnectionClient.signalingQueue.async { [weak self] in
```
charlesmchen-signal: I'm still using `[weak self] in` everywhere to ensure that accidental references to `self` cause build errors.  Unfortunately, because we're not using `[weak self]` in these blocks, these cause build warnings.  I think it's worth it - it's easy to accidentally capture `self` while moving code around.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -245,49 +282,55 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     public func setLocalVideoEnabled(enabled: Bool) {
         SwiftAssertIsOnMainThread(#function)
+        let handle = self.handle
+        let completion = { [weak self] in
+            guard let strongSelf = handle.get() else { return }
+            guard let localVideoTrack = strongSelf.localVideoTrack else { return }
+            guard let strongDelegate = strongSelf.delegate else { return }
+            strongDelegate.peerConnectionClient(strongSelf, didUpdateLocal: enabled ? localVideoTrack : nil)
+        }
```
charlesmchen-signal: Previously, we often defined blocks within blocks.  I didn't want to risk accidentally capturing strongSelf from the outer block in an inner block, so I've "unwrapped" these blocks to all be defined separately.  I guess it's worth saying that I didn't want to have `strongSelf1`, `strongSelf2` or whatever either.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     public func createOffer() -> Promise<HardenedRTCSessionDescription> {
         SwiftAssertIsOnMainThread(#function)
+        let handle = self.handle
+        let (promise, fulfill, reject) = Promise<HardenedRTCSessionDescription>.pending()
```
charlesmchen-signal: Since I've "unwrapped" the previously nested blocks, it's a lot easier to use:

```
let (promise, fulfill, reject) = Promise<HardenedRTCSessionDescription>.pending()
```

than:

```
return Promise { fulfill, reject in
```

The former lets us avoid passing around `fulfill` and `reject` which have lengthy type signatures.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
+        let completion: ((RTCSessionDescription?, Error?) -> Void) = { [weak self] (sdp, error) in
+            guard let strongSelf = handle.get() else { return }
+            strongSelf.assertOnSignalingQueue()
+            guard strongSelf.peerConnection != nil else {
```
charlesmchen-signal: I guess it's worth noting that despite all of the whitespace changes that came from unwrapping previously nested blocks, very little has changed in the business logic here.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-        Logger.debug("\(TAG) in \(#function)")
+        Logger.debug("\(logTag) in \(#function)")
 
         // Clear the delegate immediately so that we can guarantee that
         // no delegate methods are called after terminate() returns.
         delegate = nil
 
+        // Clear the handle immediately so that enqueued work is aborted
+        // going forward.
+        handle.clear()
```
charlesmchen-signal: During teardown, we call `terminate` which calls `terminateInternal` on the signaling queue.  We eagerly clear the handle.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -523,12 +605,13 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
         // become nil when it was only a weak property. So we retain it and manually nil the reference here, because
         // we are likely to crash if we retain any peer connection properties when the peerconnection is released
 
-        // See the comments on the remoteVideoTrack property.
-        objc_sync_enter(self)
```
charlesmchen-signal: This is no longer necessary.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -523,12 +605,13 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-
         localVideoTrack?.isEnabled = false
         remoteVideoTrack?.isEnabled = false
 
+        if let dataChannel = self.dataChannel {
+            dataChannel.delegate = nil
```
charlesmchen-signal: This shouldn't matter, but let's be thorough.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -593,177 +678,217 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
+        let completion: (RTCVideoTrack) -> Void = { [weak self] (remoteVideoTrack) in
+            SwiftAssertIsOnMainThread(#function)
+            guard let strongSelf = handle.get() else { return }
+            guard let strongDelegate = strongSelf.delegate else { return }
 
-        // See the comments on the remoteVideoTrack property.
-        //
-        // We only set the remoteVideoTrack property if peerConnection is non-nil.
-        objc_sync_enter(self)
-        if self.peerConnection != nil {
```
charlesmchen-signal: This is one of the only places where there are actual changes to the business logic. 

* We don't need to synchronize access to `remoteVideoTrack` property any more.
* We were _unsafely_ accessing `peerConnection` off the signaling queue here - twice.  This has been fixed.  

Overall, this method got a lot simpler and now looks a lot more like the other methods.

charlesmchen-signal:

PTAL @michaelkirk 

This PR significantly improves the app's ability to return to a good state where it can send and receive calls after call teardown, especially during edge cases.

Signal/src/call/CallService.swift
michaelkirk-signal: 👍

Signal/src/call/PeerConnectionClient.swift
michaelkirk-signal: 👍

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
 
     public func createOffer() -> Promise<HardenedRTCSessionDescription> {
         SwiftAssertIsOnMainThread(#function)
+        let handle = self.handle
+        let (promise, fulfill, reject) = Promise<HardenedRTCSessionDescription>.pending()
+        let completion: ((RTCSessionDescription?, Error?) -> Void) = { [weak self] (sdp, error) in
+            guard let strongSelf = handle.get() else { return }
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-                            let error = OWSErrorMakeUnableToProcessServerResponseError()
-                            reject(error)
-                            return
-                        }
-
-                        fulfill(HardenedRTCSessionDescription(rtcSessionDescription: sessionDescription))
-                    }
-                })
+        PeerConnectionClient.signalingQueue.async { [weak self] in
+            guard let strongSelf = handle.get() else { return }
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-                        return
-                    }
+        PeerConnectionClient.signalingQueue.async { [weak self] in
+            guard let strongSelf = handle.get() else { return }
+            strongSelf.assertOnSignalingQueue()
 
-                    let hardenedSessionDescription = HardenedRTCSessionDescription(rtcSessionDescription: sessionDescription)
+            guard let peerConnection = strongSelf.peerConnection else {
+                Logger.debug("\(strongSelf.logTag) \(#function) Ignoring obsolete event in terminated client")
+                return
```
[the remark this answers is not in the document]
charlesmchen-signal: This isn't a promise?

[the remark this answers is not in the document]
charlesmchen-signal: No, you were right.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
+        }
 
-                    guard let sessionDescription = sdp else {
-                        Logger.error("\(self.TAG) unexpected empty session description, even though no error was reported.")
-                        let error = OWSErrorMakeUnableToProcessServerResponseError()
-                        reject(error)
-                        return
-                    }
+        PeerConnectionClient.signalingQueue.async { [weak self] in
+            guard let strongSelf = handle.get() else { return }
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-
-        return Promise { fulfill, reject in
-            assertOnSignalingQueue()
-
-            guard self.peerConnection != nil else {
-                Logger.debug("\(self.TAG) \(#function) Ignoring obsolete event in terminated client")
+        let handle = self.handle
+        let (promise, fulfill, reject) = Promise<HardenedRTCSessionDescription>.pending()
+        let completion: ((RTCSessionDescription?, Error?) -> Void) = { [weak self] (sdp, error) in
+            guard let strongSelf = handle.get() else { return }
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-                                                            guard error == nil else {
-                                                                reject(error!)
-                                                                return
-                                                            }
-                                                            fulfill()
-                })
+        let handle = self.handle
+        let (promise, fulfill, reject) = Promise<Void>.pending()
+        PeerConnectionClient.signalingQueue.async { [weak self] in
+            guard let strongSelf = handle.get() else { return }
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
 
     public func negotiateSessionDescription(remoteDescription: RTCSessionDescription, constraints: RTCMediaConstraints) -> Promise<HardenedRTCSessionDescription> {
         SwiftAssertIsOnMainThread(#function)
-
+        let handle = self.handle
         return setRemoteSessionDescription(remoteDescription)
-            .then(on: PeerConnectionClient.signalingQueue) {
-                return self.negotiateAnswerSessionDescription(constraints: constraints)
+            .then(on: PeerConnectionClient.signalingQueue) { [weak self] in
+                guard let strongSelf = handle.get() else {
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-            Logger.verbose("\(self.TAG) setting local session description: \(sessionDescription)")
-            self.peerConnection.setLocalDescription(sessionDescription.rtcSessionDescription, completionHandler: resolve)
+        let handle = self.handle
+        return PromiseKit.wrap { [weak self] resolve in
+            guard let strongSelf = handle.get() else { return }
+            strongSelf.assertOnSignalingQueue()
+
+            guard let peerConnection = peerConnection else {
+                Logger.debug("\(strongSelf.logTag) \(#function) Ignoring obsolete event in terminated client")
+                return
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-                                                            guard error == nil else {
-                                                                reject(error!)
-                                                                return
-                                                            }
-                                                            fulfill()
-                })
+        let handle = self.handle
+        let (promise, fulfill, reject) = Promise<Void>.pending()
+        PeerConnectionClient.signalingQueue.async { [weak self] in
+            guard let strongSelf = handle.get() else { return }
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -334,181 +377,220 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
     }
 
     public func setLocalSessionDescriptionInternal(_ sessionDescription: HardenedRTCSessionDescription) -> Promise<Void> {
-        return PromiseKit.wrap { resolve in
-            self.assertOnSignalingQueue()
-            Logger.verbose("\(self.TAG) setting local session description: \(sessionDescription)")
-            self.peerConnection.setLocalDescription(sessionDescription.rtcSessionDescription, completionHandler: resolve)
+        let handle = self.handle
+        return PromiseKit.wrap { [weak self] resolve in
+            guard let strongSelf = handle.get() else { return }
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

Signal/src/call/PeerConnectionClient.swift
charlesmchen-signal: I added a lengthy comment.

charlesmchen-signal:
> Seems like there are some new code paths which aren't ensuring promises get resolved. Please take another look at those.

Done. PTAL @michaelkirk

charlesmchen-signal:
@michaelkirk Oh no need to TAL.  I didn't see your approval.

PR #3458, opened 2018-05-21
Mkirk/resize call navbar

michaelkirk-signal:
screen shot 2018-05-21 at 3 38 28 pm

PTAL @charlesmchen


Signal/src/ViewControllers/DebugUI/DebugUIMisc.m
```diff
@@ -109,7 +109,7 @@ - (nullable OWSTableSection *)sectionForThread:(nullable TSThread *)thread
 
     [items addObject:[OWSTableItem itemWithTitle:@"Show 2FA Reminder"
                                      actionBlock:^() {
-                                         UINavigationController *navController =
+                                         OWSNavigationController *navController =
```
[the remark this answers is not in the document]
charlesmchen-signal: You mean `UINavigationController`, not `UIViewController`, right?

SignalMessaging/attachments/MessageApprovalViewController.swift
```diff
@@ -48,26 +54,13 @@ public class MessageApprovalViewController: OWSViewController, UITextViewDelegat
...
-            items.append(spacer)
-            let sendButton = UIBarButtonItem(title: NSLocalizedString("SEND_BUTTON_TITLE",
-                                                                      comment: "Label for the send button in the conversation view."),
-                                             style: .plain,
-                                             target: self,
-                                             action: #selector(sendPressed))
-            items.append(sendButton)
-        }
+        self.navigationItem.leftBarButtonItem = UIBarButtonItem(barButtonSystemItem: .stop, target: self, action: #selector(cancelPressed))
+        self.navigationItem.rightBarButtonItem = sendButton
```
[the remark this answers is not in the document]
charlesmchen-signal: > Similarly to the AttachmentApprovalVC, the MessageApprovalVC's custom toolbar-that-looks-like-a-navbar is now just a navbar.

Bear in mind that we're using three approval VCs now: attachment, message and contact.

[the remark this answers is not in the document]
charlesmchen-signal: > One small change in behavior is that instead of hiding the "send" button when the text is empty, it remains visible, but we disable it. In general I think it's better to disable controls than to hide them.

Agreed!

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4018,8 +4038,9 @@ - (void)tryToSendAttachmentIfApproved:(SignalAttachment *_Nullable)attachment
         } else if (skipApprovalDialog) {
             [self sendMessageAttachment:attachment];
         } else {
-            AttachmentApprovalViewController *approvalVC = [[AttachmentApprovalViewController alloc] initWithAttachment:attachment delegate:self];
-            [self presentViewController:approvalVC animated:YES completion:nil];
+            OWSNavigationController *modal =
```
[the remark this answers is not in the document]
charlesmchen-signal: Are there any corresponding changes to be made in the share extension, where the approval VCs are pushed to the existing nav controller?

Signal/src/AppDelegate.m
```diff
@@ -1161,4 +1161,17 @@ - (void)ensureRootViewController
     [AppUpdateNag.sharedInstance showAppUpgradeNagIfNecessary];
 }
 
+#pragma mark - status bar touches
+
+- (void)touchesBegan:(NSSet *)touches withEvent:(UIEvent *)event
+{
+    [super touchesBegan:touches withEvent:event];
```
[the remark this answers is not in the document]
charlesmchen-signal: Just throwing this out there: I suppose we could also use _two_ separate windows for the "Return to call" UI: 

* A "visible" one behind the status bar (as we do now).
* An "invisible" (transparent) one in front of the status bar which captures touch events.

Whichever feels more robust to different devices (e.g. iPad)/multiple orientations/future iOS versions... and easier to maintain.

SignalMessaging/ViewControllers/OWSNavigationController.m
```diff
@@ -0,0 +1,146 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface UINavigationController (OWSNavigationController) <UINavigationBarDelegate, NavBarLayoutDelegate>
+
+@end
+
+#pragma mark -
+
+@interface OWSNavigationController () <UIGestureRecognizerDelegate>
```
charlesmchen-signal: Why two separate categories?  Why not just:

```
+@interface OWSNavigationController () <UIGestureRecognizerDelegate, UINavigationBarDelegate, NavBarLayoutDelegate>
```

charlesmchen-signal: Eh, I see why you did this.

SignalMessaging/ViewControllers/OWSNavigationController.m
```diff
@@ -0,0 +1,146 @@
...
+
+@end
+
+#pragma mark -
+
+@implementation OWSNavigationController
+
+- (instancetype)initWithRootViewController:(UIViewController *)rootViewController
+{
+    self = [self initWithNavigationBarClass:[OWSNavigationBar class] toolbarClass:nil];
```
charlesmchen-signal: Don't you want to do the obligatory `if self` dance?

SignalMessaging/ViewControllers/OWSNavigationController.m
```diff
@@ -0,0 +1,146 @@
...
+
+- (void)updateLayoutForNavbar:(OWSNavigationBar *)navbar
+{
+    DDLogDebug(@"%@ in %s", self.logTag, __PRETTY_FUNCTION__);
+
+    if (@available(iOS 11.0, *)) {
+        if (OWSWindowManager.sharedManager.hasCall) {
+            if (UIDevice.currentDevice.isIPhoneX) {
+                // iPhoneX computes status bar height differently.
+                self.additionalSafeAreaInsets = UIEdgeInsetsMake(navbar.navbarWithoutStatusHeight + 20, 0, 0, 0);
```
charlesmchen-signal: Is there any way we can make this safer than baking in this constant?

SignalMessaging/ViewControllers/OWSNavigationController.m
```diff
@@ -0,0 +1,146 @@
...
+        // in iOS11 we have to ensure the navbar frame *in* layoutSubviews.
+        [navbar layoutSubviews];
+    } else {
+        // Pre iOS11 we size the navbar, and position it vertically once.
+        [navbar sizeToFit];
+
+        if (OWSWindowManager.sharedManager.hasCall) {
+            CGRect oldFrame = navbar.frame;
+            CGRect newFrame
+                = CGRectMake(oldFrame.origin.x, navbar.callBannerHeight, oldFrame.size.width, oldFrame.size.height);
```
charlesmchen-signal: TIOLO but in these cases where you're mostly leaving a struct unchanged, I tend to do something like this:
```
CGRect newFrame = navbar.frame;
newFrame.height = navbar.callBannerHeight;
navbar.frame = newFrame;
```

SignalMessaging/categories/UIDevice+featureSupport.swift
```diff
@@ -0,0 +1,30 @@
...
+            // iPhone 5 or 5S or 5C
+            return false
+        case 1334:
+            // iPhone 6/6S/7/8
+            return false
+        case 1920, 2208:
+            // iPhone 6+/6S+/7+/8+//
+            return false
+        case 2436:
+            return true
```
charlesmchen-signal: This will future-proof us better if "iPhone X v2" has same height as iPhone X _and_ other new phones don't.  

`utsname()` would be more conservative. 

Presumably, it'll be easier to update this approach this Summer since the screen sizes of new phones are likely to leak before their machine identifiers do.

If we mis-identify a device, are these the only symptoms:

* the "Return to call" window might have the wrong height.
* nav bars might have the wrong height/layout during a "left" call.

charlesmchen-signal: This if fine, I'm just trying to think it through.

SignalMessaging/utils/OWSWindowManager.m
```diff
@@ -3,29 +3,44 @@
...
+
+const CGFloat OWSWindowManagerCallScreenHeight(void)
+{
+    if ([UIDevice currentDevice].isIPhoneX) {
+        // On an iPhoneX, the system return-to-call banner has been replaced by a much subtler green
+        // circle behind the system clock. Instead, we mimic the old system call banner as on older devices,
+        // but it has to be taller to fit beneath the notch.
+        return 64;
+    } else {
+        return CurrentAppContext().statusBarHeight + 20;
```
charlesmchen-signal: I feel somewhat nervous about all of the places that we are special-casing iPhone X or baking in iOS size constants.  

I wonder if we should leave in a new kind of code marker (like TODO), maybe IOS_DEVICE_CONSTANT or something that we can use to remind ourselves of every constant we should check whenever there's a new major iOS update or new device model.

Signal/src/ViewControllers/DebugUI/DebugUIMisc.m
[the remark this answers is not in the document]
michaelkirk-signal: errp. yes.

Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
```diff
@@ -126,8 +134,21 @@ class AddContactShareToExistingContactViewController: ContactsPicker, ContactsPi
             return
         }
 
+        // HACK otherwise CNContactViewController Navbar is shows window background color.
+        // RADAR rdar://28433898 http://www.openradar.me/28433898
+        // CNContactViewController incompatible with opaque navigation bar
+        navigationController.navigationBar.isTranslucent = false
+        if #available(iOS 10, *) {
```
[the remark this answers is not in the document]
michaelkirk-signal: On iOS9, the CNContactVC has a blue navbar.

Previosly for iOS9 users we showed black text on that blue navbar for the CNContactVC, now we show white text on that blue navbar for the CNContactVC

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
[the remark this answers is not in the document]
michaelkirk-signal: Yup - they are now presented modally.

SignalMessaging/ViewControllers/OWSNavigationController.m
michaelkirk-signal: Better, thanks!

SignalMessaging/ViewControllers/OWSNavigationController.m
michaelkirk-signal: danced!

SignalMessaging/ViewControllers/OWSNavigationController.m
michaelkirk-signal: I added a comment - it wasn't obvious.

SignalMessaging/attachments/MessageApprovalViewController.swift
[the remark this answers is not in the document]
michaelkirk-signal: > Bear in mind that we're using three approval VCs now: attachment, message and contact.

Thankfully, the contact share approval was already using a standard navbar.

SignalMessaging/categories/UIDevice+featureSupport.swift
michaelkirk-signal: > If we mis-identify a device, are these the only symptoms:
> - the "Return to call" window might have the wrong height.
> - nav bars might have the wrong height/layout during a "left" call.

That's right.

SignalMessaging/utils/OWSWindowManager.m
michaelkirk-signal: Unfortunately, I'm not sure that there is a constant that gets us what we want. This kind of layout is so far against the grain of what iOS asks us for, though I tried, I couldn't figure out computable constants for some of them.

I have annotated them with the IOS_DEVICE_CONSTANT comment tag, and also put a note in the sentinel in the UIDevice+featureSupport, so that if we get a new device size and hit that assert we'll have some direction about the things to fix.

charlesmchen-signal:
@michaelkirk I'm going to merge this - please write the test plan.

PR #3462, opened 2018-05-22
Charlesmchen/call service state

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -117,35 +117,67 @@ private class SignalCallData: NSObject {
     let rejectReadyToSendIceUpdatesPromise: ((Error) -> Void)
     let readyToSendIceUpdatesPromise: Promise<Void>
 
-    weak var localVideoTrack: RTCVideoTrack? {
-        didSet {
+    private weak var localVideoTrack_: RTCVideoTrack?
+    var localVideoTrack: RTCVideoTrack? {
+        get {
+        SwiftAssertIsOnMainThread(#function)
+
+        return localVideoTrack_
+        }
+        set {
             SwiftAssertIsOnMainThread(#function)
 
-            Logger.info("\(self.logTag) \(#function)")
+            Logger.info("\(self.logTag)\(#function): \(localVideoTrack_ != nil) -> \(newValue != nil) \(String(describing: newValue))")
+
+            localVideoTrack_ = newValue
         }
     }
 
-    weak var remoteVideoTrack: RTCVideoTrack? {
-        didSet {
+    weak var remoteVideoTrack_: RTCVideoTrack?
+    var remoteVideoTrack: RTCVideoTrack? {
+        get {
             SwiftAssertIsOnMainThread(#function)
 
-            Logger.info("\(self.logTag) \(#function)")
+            return remoteVideoTrack_
+        }
+        set {
+            SwiftAssertIsOnMainThread(#function)
+
+            Logger.info("\(self.logTag)\(#function): \(remoteVideoTrack_ != nil) -> \(newValue != nil) \(String(describing: newValue))")
+
+            remoteVideoTrack_ = newValue
         }
     }
 
-    var isRemoteVideoEnabled = false {
-        didSet {
+    var isRemoteVideoEnabled_ = false
+    var isRemoteVideoEnabled: Bool {
+        get {
             SwiftAssertIsOnMainThread(#function)
 
-            Logger.info("\(self.logTag) \(#function): \(isRemoteVideoEnabled)")
+            return isRemoteVideoEnabled_
+        }
+        set {
+            SwiftAssertIsOnMainThread(#function)
+
+            Logger.info("\(self.logTag)\(#function): \(isRemoteVideoEnabled_) -> \(newValue)")
+
+            isRemoteVideoEnabled_ = newValue
         }
     }
 
+    var peerConnectionClient_: PeerConnectionClient?
     var peerConnectionClient: PeerConnectionClient? {
-        didSet {
+        get {
             SwiftAssertIsOnMainThread(#function)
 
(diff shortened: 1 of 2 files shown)


Signal/src/call/PeerConnectionClient.swift
```diff
@@ -286,7 +286,7 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
         SwiftAssertIsOnMainThread(#function)
 
         let handle = self.handle
-        PeerConnectionClient.signalingQueue.async { [weak self] in
```
charlesmchen-signal: I decided to take these out for now since they cause build warnings.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1014,7 +1014,7 @@
 "INCOMING_DECLINED_CALL" = "You declined a call";
 
 /* No comment provided by engineer. */
-"INCOMING_INCOMPLETE_CALL" = "Incomplete incoming call from";
+"INCOMING_INCOMPLETE_CALL" = "Incomplete incoming call";
```
charlesmchen-signal: This isn't a format string.

Signal/src/call/CallService.swift
```diff
@@ -117,35 +117,67 @@ private class SignalCallData: NSObject {
     let rejectReadyToSendIceUpdatesPromise: ((Error) -> Void)
     let readyToSendIceUpdatesPromise: Promise<Void>
 
-    weak var localVideoTrack: RTCVideoTrack? {
-        didSet {
+    private weak var localVideoTrack_: RTCVideoTrack?
+    var localVideoTrack: RTCVideoTrack? {
```
charlesmchen-signal: I modified `SignalCallData`'s accessors to more strictly enforce that we only access these properties on the main thread.

Signal/src/call/CallService.swift
```diff
@@ -253,34 +289,15 @@ private class SignalCallData: NSObject {
...
+    // This accessor is intended for external use only.
+    // Internally, we should access the call through the call data.
+    public var call: SignalCall? {
         get {
             SwiftAssertIsOnMainThread(#function)
 
             return callData?.call
         }
     }
-    var peerConnectionClient: PeerConnectionClient? {
```
charlesmchen-signal: I got rid of these convenience accessors in an effort to force us to be a little more strict about pulling data from `SignalCallData`.

Signal/src/call/CallService.swift
```diff
@@ -862,22 +876,19 @@ private class SignalCallData: NSObject {
...
+        let call = callData.call
 
         Logger.info("\(self.logTag) in \(#function): \(call.identifiersForLogs).")
 
         switch call.state {
         case .remoteRinging:
             Logger.debug("\(self.logTag) in \(#function) disconnect while ringing... we'll keep ringing")
+        case .localRinging:
+            // This can happen if the incoming caller hangs up while we are ringing.
+            handleFailedCurrentCall(error: CallError.disconnected)
```
charlesmchen-signal: This is a notable change.

Signal/src/call/CallService.swift
```diff
@@ -1495,6 +1534,18 @@ private class SignalCallData: NSObject {
                 // call failed before any call record could be created, make one now.
                 handleMissedCall(failedCall)
             }
+
+            // Ensure there is a call record.
+            let callType = (failedCall.direction == .incoming
```
charlesmchen-signal: We weren't always creating a call record, e.g. if caller hung up while ringing.

charlesmchen-signal:
@michaelkirk These changes reflect an effort to be a little more strict about concurrency and state coherency.  If you don't like them, I'll revert most of them.

michaelkirk-signal:
> LGTM

Oh whoops - I missed the bulk of your changes since GH minimized the CallService file. Taking a look now.

Signal/src/call/CallService.swift
michaelkirk-signal: The case you've described should be covered by our existing handleRemoteHangup logic, what am I missing?

charlesmchen-signal:
> Oh whoops - I missed the bulk of your changes since GH minimized the CallService file. Taking a look now.

PTAL @michaelkirk ?

Signal/src/call/CallService.swift
michaelkirk-signal: @charlesmchen-signal - could you respond to my comment here?

charlesmchen-signal: Makes sense. This change isn't necessary now that you've added `localRinging` to the case above.

Signal/src/call/CallService.swift
charlesmchen-signal: Yup - we've just missed cases before and so I thought it'd be conservative to make sure we always leave a call record, given all of the edge cases in calling.  If you'd rather remove, we can.

charlesmchen-signal:
PTAL @michaelkirk-signal

Signal/src/call/CallService.swift
```diff
@@ -1507,13 +1543,26 @@ private class SignalCallData: NSObject {
                 assert(failedCall.callRecord != nil)
             }
 
+            // Ensure there is a call record.
+            let callType = (failedCall.direction == .incoming
+                ? RPRecentCallTypeIncomingIncomplete
+                : RPRecentCallTypeOutgoingIncomplete)
+            if let callRecord = failedCall.callRecord {
+                callRecord.updateCallType(callType)
```
michaelkirk-signal: Won't this clobber the "missed" call record we just set up on line 1541? 

And in the non-missed case (`default:`, above) we're already asserting that a call record exists, so if this is just a belts and suspenders thing, can we only keep the else case.

Imperiopolis-Signal:

PR #3469, opened 2018-05-23
start migrated disappearing timers

michaelkirk-signal:
PTAL @charlesmchen


SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
```diff
@@ -442,6 +448,13 @@ - (BOOL)shouldStartExpireTimerWithTransaction:(YapDatabaseReadTransaction *)tran
 
     if (!self.isExpiringMessage) {
         return NO;
+    } else if (self.expireStartedAt > 0) {
```
michaelkirk-signal: The above changes to `hasSentToAnyRecipient` should be sufficient for anyone on who went from 2.24->2.25.1

And the following should recover people who migrated to 2.24->2.25.0->2.25.1, since they won't have a `legacyMessageState`.

PR #3477, opened 2018-05-24
Fix call termination reference.

charlesmchen-signal:
PTAL @michaelkirk


charlesmchen-signal:
I'd like to merge this - it's already fixed in #3462 but that isn't merged yet.

PR #3478, opened 2018-05-24
Apply design for call view back button.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/utils/OWSWindowManager.m
+++ b/SignalMessaging/utils/OWSWindowManager.m
@@ -72,7 +72,7 @@ const UIWindowLevel UIWindowLevel_ScreenBlocking(void)
 
 // UIWindowLevel_CallView
 @property (nonatomic) UIWindow *callViewWindow;
-@property (nonatomic) OWSNavigationController *callNavigationController;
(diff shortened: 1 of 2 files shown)


SignalMessaging/utils/OWSWindowManager.m
```diff
@@ -184,8 +184,9 @@ - (UIWindow *)createCallViewWindow:(UIWindow *)rootWindow
     UIViewController *viewController = [OWSWindowRootViewController new];
     viewController.view.backgroundColor = [UIColor ows_materialBlueColor];
 
-    OWSNavigationController *navigationController =
-        [[OWSNavigationController alloc] initWithRootViewController:viewController];
+    // NOTE: Do not use OWSNavigationController for call window.
```
michaelkirk-signal: Why not? 

In general prefer comments that explain why vs. what

charlesmchen-signal: Improved the comment here:

https://github.com/signalapp/Signal-iOS/commit/c71081c8758c94e51859d75e56649f42e2f62990

PR #3482, opened 2018-05-25
Charlesmchen/swift4

charlesmchen-signal:
PTAL @michaelkirk


Signal.xcodeproj/project.pbxproj
```diff
@@ -4431,7 +4439,8 @@
 				PROVISIONING_PROFILE = "";
 				SWIFT_OBJC_BRIDGING_HEADER = "Signal/test/SignalTests-Bridging-Header.h";
 				SWIFT_OPTIMIZATION_LEVEL = "-Owholemodule";
-				SWIFT_VERSION = 3.0;
+				SWIFT_SWIFT3_OBJC_INFERENCE = On;
```
charlesmchen-signal: > SWIFT_SWIFT3_OBJC_INFERENCE 

We might want to turn this off? 

> SWIFT_VERSION

Not sure if we should use 4.0 or 4.1.

Signal/src/call/CallService.swift
```diff
@@ -154,17 +154,23 @@ private class SignalCallData: NSObject {
 
         let (callConnectedPromise, fulfillCallConnectedPromise, rejectCallConnectedPromise) = Promise<Void>.pending()
         self.callConnectedPromise = callConnectedPromise
-        self.fulfillCallConnectedPromise = fulfillCallConnectedPromise
+        self.fulfillCallConnectedPromise = {
```
michaelkirk-signal: #tuplegate =(

Signal/test/call/PeerConnectionClientTest.swift
```diff
@@ -12,35 +12,37 @@ import WebRTC
...
     var dataChannelMessages = [OWSWebRTCProtosData]()
 
-    internal func peerConnectionClientIceConnected(_ peerconnectionClient: PeerConnectionClient) {
+    func peerConnectionClientIceConnected(_ peerconnectionClient: PeerConnectionClient) {
         connectionState = .connected
     }
 
-    internal func peerConnectionClientIceFailed(_ peerconnectionClient: PeerConnectionClient) {
+    func peerConnectionClientIceDisconnected(_ peerconnectionClient: PeerConnectionClient) {
+        connectionState = .disconnected
```
michaelkirk-signal: Thanks for fixing up these tests.

Signal.xcodeproj/project.pbxproj
charlesmchen-signal: > SWIFT_SWIFT3_OBJC_INFERENCE

has been disabled everywhere.

> SWIFT_VERSION

So, if you select 4.1 in the XCode project/target build settings, the actual value that is persisted is 4.0.  So I think we're okay?

Signal/test/call/PeerConnectionClientTest.swift
charlesmchen-signal: I just pushed one more commit to get these tests not just building but passing.

https://github.com/signalapp/Signal-iOS/pull/3482/commits/f7abcc9064eecc911f3fe9d3f10606dd8dcadc35

PR #3483, opened 2018-05-25
Charlesmchen/pcc deinit

charlesmchen-signal:
This PR is ready for review. 

PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/PeerConnectionClient.swift
+++ b/Signal/src/call/PeerConnectionClient.swift
@@ -244,6 +244,9 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
     private var cameraConstraints: RTCMediaConstraints
 
     private let proxy = PeerConnectionProxy()
+    // Note that we're deliberately leaking proxy instances using this
+    // collection to avoid EXC_BAD_ACCESS.  Calls are rare and the proxy
+    // is tiny (a single property), so it's better to leak and be safe.
     private static var expiredProxies = [PeerConnectionProxy]()
 
     init(iceServers: [RTCIceServer], delegate: PeerConnectionClientDelegate, callDirection: CallDirection, useTurnOnly: Bool) {
--- a/Signal/src/call/PeerConnectionClient.swift
+++ b/Signal/src/call/PeerConnectionClient.swift
@@ -108,6 +108,13 @@ class PeerConnectionProxy: NSObject, RTCPeerConnectionDelegate, RTCDataChannelDe
         objc_sync_enter(self)
         let result = value
         objc_sync_exit(self)
+
+        if result == nil {
+            // Every time this method returns nil is a
+            // possible crash avoided.
+            Logger.verbose("\(logTag) cleared get.")
+        }
+
         return result
     }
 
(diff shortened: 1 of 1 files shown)


Signal/src/call/PeerConnectionClient.swift
```diff
@@ -66,62 +66,134 @@ protocol PeerConnectionClientDelegate: class {
...
+// PeerConnectionClient has an PeerConnectionProxy to itself that its many async blocks
+// (which run on more than one thread) can use to safely try to acquire a strong
+// reference to the PeerConnectionClient. In ARC we'd normally, we'd avoid
+// having an instance retain a strong reference to itself to avoid retain
+// cycles, but it's safe in this case: PeerConnectionClient is owned (and only
+// used by) a single entity CallService and CallService always calls
+// [PeerConnectionClient terminate] when it is done with a PeerConnectionClient
+// instance, so terminate is a reliable place where we can break the retain cycle.
+//
+// Note that we use the proxy in two ways:
```
charlesmchen-signal: The bad news is that upgrading to Swift 4.x doesn't fix these crashes.

The good news is that I figured out how to fix the remaining rare cases of this crash: by applying the same fix in more places.  Specifically, by using a delegate proxy that like `AtomicHandle` accesses the PCC in a safe way.  

So, I've renamed `AtomicHandle` to `PeerConnectionProxy` and we now use it two ways.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -171,7 +243,11 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
     private var remoteVideoTrack: RTCVideoTrack?
     private var cameraConstraints: RTCMediaConstraints
 
-    private let handle: AtomicHandle<PeerConnectionClient>
+    private let proxy = PeerConnectionProxy()
+    // Note that we're deliberately leaking proxy instances using this
+    // collection to avoid EXC_BAD_ACCESS.  Calls are rare and the proxy
+    // is tiny (a single property), so it's better to leak and be safe.
+    private static var expiredProxies = [PeerConnectionProxy]()
```
charlesmchen-signal: Note this comment.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -285,9 +359,9 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
     public func setCameraSource(useBackCamera: Bool) {
         SwiftAssertIsOnMainThread(#function)
 
-        let handle = self.handle
-        PeerConnectionClient.signalingQueue.async { [weak self] in
```
charlesmchen-signal: Lastly, I removed `[weak self]` from these blocks.  They cause build warnings.  The intent was to protect us from accidentally creating strong references to `self` if we touch these blocks in the future, but the swift compiler will warn us about implicit usages of `self` anyway.  We just need to be careful to not accidentally use `self.`.

charlesmchen-signal:
This PR is ready for review. 

PTAL @michaelkirk

PR #3487, opened 2018-05-29
Fix breakage from recents changes in share extension.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/attachments/MessageApprovalViewController.swift
+++ b/SignalMessaging/attachments/MessageApprovalViewController.swift
@@ -23,13 +23,7 @@ public class MessageApprovalViewController: OWSViewController, UITextViewDelegat
     let contactsManager: OWSContactsManager
 
     private(set) var textView: UITextView!
-    private var sendButton: UIBarButtonItem = {
-        return UIBarButtonItem(title: NSLocalizedString("SEND_BUTTON_TITLE",
-                                                        comment: "Label for the send button in the conversation view."),
-                               style: .plain,
-                               target: self,
-                               action: #selector(sendPressed))
-    }()
+    private var sendButton: UIBarButtonItem!
 
     // MARK: Initializers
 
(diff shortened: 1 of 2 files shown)


SignalMessaging/attachments/MessageApprovalViewController.swift
```diff
@@ -23,13 +23,7 @@ public class MessageApprovalViewController: OWSViewController, UITextViewDelegat
...
 
     private(set) var textView: UITextView!
-    private var sendButton: UIBarButtonItem = {
-        return UIBarButtonItem(title: NSLocalizedString("SEND_BUTTON_TITLE",
-                                                        comment: "Label for the send button in the conversation view."),
-                               style: .plain,
-                               target: self,
-                               action: #selector(sendPressed))
-    }()
+    private var sendButton: UIBarButtonItem!
```
charlesmchen-signal: TIL that if you access `self` in a block used to initialize a class property, it has an invalid value.  Send button was broken.

@michaelkirk Is there anywhere else we did something similar?

SignalMessaging/attachments/MessageApprovalViewController.swift
```diff
@@ -52,17 +46,29 @@ public class MessageApprovalViewController: OWSViewController, UITextViewDelegat
...
     }
 
     private func updateSendButton() {
         sendButton.isEnabled = textView.text.count > 0
     }
 
+    override public func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        updateSendButton()
```
charlesmchen-signal: This isn't strictly necessary but I think it makes the class easier to reason about.

SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -976,7 +976,7 @@ - (void)sendMessageToService:(TSOutgoingMessage *)message
                                                                      messages:deviceMessages
                                                                         relay:recipient.relay
                                                                     timeStamp:message.timestamp];
-    if (TSSocketManager.sharedManager.canMakeRequests) {
+    if (CurrentAppContext().isMainApp && TSSocketManager.sharedManager.canMakeRequests) {
```
charlesmchen-signal: We don't want to send over the web socket in the share extension.

SignalMessaging/attachments/MessageApprovalViewController.swift
michaelkirk-signal: I'm OK with the change, but I don't understand what was wrong with the existing code.

> initialize a class property

I'm a little confused by terminology here - Swift doesn't have "class" properties the way it has "class" functions, it has instance properties and static properties. I think `static` properties are pretty close to what you might call a `class` property, but this is an instance property not a static property.

If you are saying that you *can't* reference `self` inside a block used to initialize a lazy var, I don't think that's correct (unless you are trying to use the lazy var in the initializer, in which case standard "you can't reference self until initialization is complete" holds).

SignalServiceKit/src/Messages/OWSMessageSender.m
michaelkirk-signal: Though I agree we don't want to send over the web socket in the share extension, I also don't want to sprinkle concerns around that should be centralized.

The existing implementation of `canMakeRequests` relies on socket state. Are you saying that the socket was **open** in the share extension? That seems like a bug.

Or if you did *not* see that the socket was open in the SAE, and this change is just an abundance of caution, then would it be better to centralize the check in `TSSocketManager#canMakeRequests` rather than at every call site?

SignalMessaging/attachments/MessageApprovalViewController.swift
charlesmchen-signal: Sorry, I misspoke - I should have said _instance_ property. 

But the issue here is that the block used to initialize the property is defined in the context of the class, not the instance.  `self` as it was referred to the `MessageApprovalViewController` class, as it would in a class method.  I was surprised too.

SignalServiceKit/src/Messages/OWSMessageSender.m
charlesmchen-signal: Good idea.  See: https://github.com/signalapp/Signal-iOS/pull/3495

PR #3488, opened 2018-05-29
Clear the contacts cache if access to the system contacts is revoked.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/contacts/SystemContactsFetcher.swift
+++ b/SignalMessaging/contacts/SystemContactsFetcher.swift
@@ -109,7 +109,8 @@ class ContactsFrameworkContactStoreAdaptee: ContactStoreAdaptee {
     }
 }
 
-public enum ContactStoreAuthorizationStatus {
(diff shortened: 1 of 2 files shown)


SignalMessaging/contacts/SystemContactsFetcher.swift
```diff
@@ -253,6 +258,7 @@ public class SystemContactsFetcher: NSObject {
         SwiftAssertIsOnMainThread(#function)
         guard authorizationStatus == .authorized else {
             owsFail("should have already requested contact access")
+            self.delegate?.systemContactsFetcher(self, hasAuthorizationStatus: authorizationStatus)
```
michaelkirk-signal: Existing bug, but shouldn't we be calling the completion handler before we return?

charlesmchen-signal: Good point. Fixed.

PR #3493, opened 2018-05-30
Fix more call state edge cases.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -1233,10 +1233,9 @@ private class SignalCallData: NSObject {
         SwiftAssertIsOnMainThread(#function)
 
         guard let call = self.call else {
-            // This should never happen; return to a known good state.
-            owsFail("\(self.logTag) call was unexpectedly nil in \(#function)")
-            OWSProdError(OWSAnalyticsEvents.callServiceCallMissing(), file: #file, function: #function, line: #line)
-            handleFailedCurrentCall(error: CallError.assertionError(description: "\(self.logTag) call unexpectedly nil in \(#function)"))
+            // This can happen if you tap the video button right after the other
+            // user hangs up.
+            Logger.warn("\(self.logTag) ignoring local video change; no call.")
             return
         }
 
--- a/Signal/src/call/CallService.swift
+++ b/Signal/src/call/CallService.swift
@@ -244,10 +244,11 @@ private class SignalCallData: NSObject {
                 if let oldValue = oldValue {
                     DeviceSleepManager.sharedInstance.removeBlock(blockObject: oldValue)
                 }
-                stopAnyCallTimer()
                 if let callData = callData {
                     DeviceSleepManager.sharedInstance.addBlock(blockObject: callData)
                     self.startCallTimer()
+                } else {
+                    stopAnyCallTimer()
                 }
             }
 
@@ -1233,9 +1234,9 @@ private class SignalCallData: NSObject {
         SwiftAssertIsOnMainThread(#function)
 
         guard let call = self.call else {
-            // This can happen if you tap the video button right after the other
-            // user hangs up.
-            Logger.warn("\(self.logTag) ignoring local video change; no call.")
+            // This can happen if you toggle local video right after
+            // the other user ends the call.
+            Logger.debug("\(self.logTag) \(#function) Ignoring event from obsolete call")
             return
         }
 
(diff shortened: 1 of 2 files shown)


Signal/src/call/UserInterface/CallUIAdapter.swift
```diff
@@ -67,8 +67,8 @@ extension CallUIAdaptee {
         SwiftAssertIsOnMainThread(#function)
 
         guard self.callService.call == nil else {
-            Logger.info("unexpectedly found an existing call when trying to start outgoing call: \(recipientId)")
-            //TODO terminate existing call.
+            owsFail("unexpectedly found an existing call when trying to start outgoing call: \(recipientId)")
+            self.callService.terminateCallDueToCriticalError()
```
charlesmchen-signal: Same deal: I haven't figured out a repro yet, but I hit this once today.  I got the phone into a state where it couldn't start outgoing calls because of a lingering call.

charlesmchen-signal: I repro'd again.  You have to terminate a call (locally) then tap call button ASAP.  It's hard to repro, but it's possible.

Signal/src/call/CallService.swift
```diff
@@ -1226,10 +1227,9 @@ private class SignalCallData: NSObject {
         SwiftAssertIsOnMainThread(#function)
 
         guard let call = self.call else {
-            // This should never happen; return to a known good state.
-            owsFail("\(self.logTag) call was unexpectedly nil in \(#function)")
-            OWSProdError(OWSAnalyticsEvents.callServiceCallMissing(), file: #file, function: #function, line: #line)
-            handleFailedCurrentCall(error: CallError.assertionError(description: "\(self.logTag) call unexpectedly nil in \(#function)"))
+            // This can happen if you toggle local video right after
+            // the other user ends the call.
```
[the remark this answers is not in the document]
michaelkirk-signal: Thanks for the notes.

Signal/src/call/UserInterface/CallUIAdapter.swift
michaelkirk-signal: Thinking about this scenario:

1. Alice initiates call to Bob, starts ICE negotiation
2. While Bob/Alice are still setting up the call Signaling (so Bob's phone isn't ringing yet), Bob taps the button to  initiate a call to Charlie.

So before this change:
1. Bob's device would not initiate a call to Charlie, and just seem unresponsive.
2. Then eventually...
2.a. Bob's phone would start to ring with an incoming call from Alice
2.b. or the call would fail to establish and Bob would see a missed call from Alice
3. Now Bob can initiate a call to Charlie.
 
And after this change:
1. the call from Alice would terminate before it has connected
2. Bob would see "Call failed" (for the call from Alice)
3. Bob's call to Charlie would not start.

It's kind of a hairy situation, so I'm not even sure what the ideal behavior for the user is... maybe:

1. When Bob taps to call Charlie
2. Bob's client terminates the call with a "Busy" message (we'd have to ensure all clients handle this: receiving busy *after* a call has started to negotiate ICE might be unexpected).
3. Bob sees a "missed call" notification inserted in his conversation with Alice
4. Bob starts to connect a call with Charlie.

Signal/src/call/CallService.swift
```diff
@@ -238,10 +238,11 @@ private class SignalCallData: NSObject {
                 if let oldValue = oldValue {
                     DeviceSleepManager.sharedInstance.removeBlock(blockObject: oldValue)
                 }
-                stopAnyCallTimer()
                 if let callData = callData {
                     DeviceSleepManager.sharedInstance.addBlock(blockObject: callData)
                     self.startCallTimer()
+                } else {
+                    stopAnyCallTimer()
```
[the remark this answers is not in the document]
michaelkirk-signal: What are you trying to accomplish by moving this code? To me it seems like the outcome is exactly the same. Am I missing something?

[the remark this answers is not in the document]
charlesmchen-signal: It's not dramatic, but I was trying to improve readability and future-proof this code, e.g. make it safe to call `startCallTimer()` without calling `stopAnyCallTimer()` first, etc.  If you don't like it, I'll happily revert.

Signal/src/call/UserInterface/CallUIAdapter.swift
charlesmchen-signal: I agree with your analysis, but after I opened this PR I found that there are multiple repros.  As you point out, there's the "race to initiate call" case, but there's also a "call service ends up in bad state" case.  In the latter, we want to terminate to get bad to a "known good" state.  In the former, we don't.  

Also, obviously the latter case needs more investigation.  I'm going to revert this part of the PR and see if I can repro the latter case.

charlesmchen-signal:
PTAL @michaelkirk I've reverted the "terminate existing call if initiating call conflict" change until I've researched more.

PR #3496, opened 2018-05-30
Avoid spurious animation in call view buttons.

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/CallViewController.swift
```diff
@@ -714,6 +714,8 @@ class CallViewController: OWSViewController, CallObserver, CallServiceObserver,
             ongoingVideoCallControls.removeFromSuperview()
             ongoingCallControls.insertArrangedSubview(ongoingAudioCallControls, at: 0)
         }
+        // Layout immediately to avoid spurious animation.
+        ongoingCallControls.layoutIfNeeded()
```
charlesmchen-signal: The repro is to enable then disable local animations. When disabling, there was an undesired animation.

michaelkirk-signal: LGTM

PR #3498, opened 2018-05-30
Fixed leading/trailing margins for empty home view label.

Gabardone:

- [X ] I have read the [README](https://github.com/WhisperSystems/Signal-iOS/blob/master/README.md) and [CONTRIBUTING](https://github.com/WhisperSystems/Signal-iOS/blob/master/CONTRIBUTING.md) documents
- [X ] I have signed the [Contributor Licence Agreement](https://whispersystems.org/cla/)

### Contributor checklist

 * iPhone 8+, iOS 11.3
 * iPad Pro, iOS 11.3

- - - - - - - - - -

### Description
- The logic to layout the label shown for an empty conversation list is changed so the leading/trailing margins will be respected.
- The label is now centered vertically instead of filling up the superview
- The label can now use more than two lines of text (likely to be an improvement on smaller phones for certain localizations).
- Added a couple comments about some of the localized strings for that label which are mislabeled, to be fixed on a later patch.

- - - - - - - - - -

### Note
- Meant also as a simple, easy fix to help me figure out the process for contributing to Signal.

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -226,7 +226,7 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
 
     [emptyBoxLabel autoPinLeadingToSuperviewMargin];
     [emptyBoxLabel autoPinTrailingToSuperviewMargin];
-    [emptyBoxLabel autoPinVCenterToSuperviewVCenter];
+    [emptyBoxLabel autoAlignAxisToSuperviewAxis:ALAxisHorizontal];
 
     UIRefreshControl *pullToRefreshView = [UIRefreshControl new];
     pullToRefreshView.tintColor = [UIColor grayColor];
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -220,10 +220,13 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
     UILabel *emptyBoxLabel = [UILabel new];
     self.emptyBoxLabel = emptyBoxLabel;
     [self.view addSubview:emptyBoxLabel];
-    [emptyBoxLabel setNumberOfLines:0]; //  Let the label use as many lines as needed. It will very rarely be more than 2.
+
+    //  Let the label use as many lines as needed. It will very rarely be more than 2 but may happen for verbose locs.
+    [emptyBoxLabel setNumberOfLines:0];
+
     [emptyBoxLabel autoPinLeadingToSuperviewMargin];
     [emptyBoxLabel autoPinTrailingToSuperviewMargin];
-    [[[emptyBoxLabel centerYAnchor] constraintEqualToAnchor:[[self view] centerYAnchor]] setActive:YES];
+    [emptyBoxLabel autoPinVCenterToSuperviewVCenter];
 
     UIRefreshControl *pullToRefreshView = [UIRefreshControl new];
     pullToRefreshView.tintColor = [UIColor grayColor];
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -221,9 +221,10 @@ - (void)loadView
     UILabel *emptyBoxLabel = [UILabel new];
     self.emptyBoxLabel = emptyBoxLabel;
     [self.view addSubview:emptyBoxLabel];
-    [emptyBoxLabel autoPinWidthToSuperview];
-    [emptyBoxLabel autoPinToTopLayoutGuideOfViewController:self withInset:0];
-    [emptyBoxLabel autoPinToBottomLayoutGuideOfViewController:self withInset:0];
+    [emptyBoxLabel setNumberOfLines:0]; //  Let the label use as many lines as needed. It will very rarely be more than 2.
```
michaelkirk-signal: style nit: no trailing comments, leave them above on a line of their own

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -221,9 +221,10 @@ - (void)loadView
...
+    [emptyBoxLabel autoPinLeadingToSuperviewMargin];
+    [emptyBoxLabel autoPinTrailingToSuperviewMargin];
+    [[[emptyBoxLabel centerYAnchor] constraintEqualToAnchor:[[self view] centerYAnchor]] setActive:YES];
```
michaelkirk-signal: We use PureLayout (or our own PureLayout inspired methods) for making AL constraints.

So this should be something like:

```
[emptyBoxLabel autoAlignAxis:ALAxisHorizontal toSameAxisOfView:self.view];
```

Or maybe just:
```
[emptyBoxLabel autoVCenterInSuperview];
```

Signal/src/ViewControllers/HomeView/HomeViewController.m
Gabardone: No problem, learning the local coding style as we go.

Signal/src/ViewControllers/HomeView/HomeViewController.m
Gabardone: I got the impression that PureLayout is being slowly phased out from Signal since it doesn't seem to be in active development and the newer Signal code isn't using it, plus the improvements in auto layout API through the years have gotten it to the point where auto layout wrappers aren't really that much of a benefit for most common use cases.

I thought about coming up with a patch to remove usage of the PureLayout library to begin with, but that would be a lot of code churn and require a firm decision of the best way to go forward. I'd advocate for just using the modern UIKit APIs for all but the more complex cases, but the wrappers you're building are ok if a bit redundant for a few too many things for my taste. If neither proved a problem I could go ahead with that, on a separate patch of course. FWIW I've been accused to knowing my auto layouts 👍🏽

I also didn't know if you'd want a utility method for this single use case, but it's an easy enough change to make. I'll get that in place in a second.

michaelkirk-signal: > I got the impression that PureLayout is being slowly phased out from Signal since it doesn't seem to be in active development and the newer Signal code isn't using it

Nope, we're still using PureLayout.

The `autoVCenterInSuperview` method is already defined, can you use the existing definition and remove your newly added method?

Gabardone:
Done as requested (although the PureLayout API for this is outright weird).

Let me know if any further polish is needed.

Gabardone:
img_0003
img_0004

Attached images of the before and after (I did notice this issue with one of the English strings that appears in the archive tab but there's probably been a behavior change there recently where it no longer pops up. It should be easy enough to see with real localized strings too).

michaelkirk:
LGTM 

PTAL @charlesmchen

SignalMessaging/categories/UIDevice+featureSupport.swift
```diff
@@ -11,6 +11,9 @@ public extension UIDevice {
 
     var isIPhoneX: Bool {
         switch UIScreen.main.nativeBounds.height {
+        case 960:
+            //  iPad in iPhone compatibility mode (using old iPhone 4 screen size)
```
charlesmchen-signal: Nice.

charlesmchen-signal:
Merged in https://github.com/signalapp/Signal-iOS/commit/93a4502ee41affed93b5fee83987e418b4aad80a

Thanks @Gabardone

PR #3501, opened 2018-05-31
Use background task while downloading profile avatar.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/profiles/OWSProfileManager.m
+++ b/SignalMessaging/profiles/OWSProfileManager.m
@@ -801,6 +801,8 @@ const NSUInteger kOWSProfileManager_MaxAvatarDiameter = 640;
 {
     OWSAssert(userProfile);
 
+    __block OWSBackgroundTask *backgroundTask = [OWSBackgroundTask backgroundTaskWithLabelStr:__PRETTY_FUNCTION__];
+
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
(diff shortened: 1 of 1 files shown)


michaelkirk-signal:
This doesn't seem dangerous, but why this change? 

It doesn't seem like it would cause 10deadcc's

charlesmchen-signal:
> This doesn't seem dangerous, but why this change?

After we download a profile avatar, we do some db writes, e.g. `updateWithAvatarFileName`.

michaelkirk:
>> This doesn't seem dangerous, but why this change?
>
> After we download a profile avatar, we do some db writes, e.g. updateWithAvatarFileName.

But those writes would start their own background task... so I don't understand how this change would fix anything.

charlesmchen-signal:
> But those writes would start their own background task... so I don't understand how this change would fix anything.

Those writes are delayed by the download process.  I suspect what's happening is that when the download complete while the app is in the background but not yet suspended, we try to open the write transaction and .... 0xdead10cc.  It's not clear under what exact circumstances the app is killed with 0xdead10cc, but it's clearly not just when the app transitions from background -> suspended.  For example, consider all the 0xdead10cc we are seeing when returning from the background.  

We've already had success so far in reducing 0xdead10cc by wrapping not just narrow db writes in background tasks, but entire background processes that involve either db writes + NSUserDefaults access.

michaelkirk-signal:
Ok, I don't understand it, but it doesn't seem dangerous, so I'm fine with trying it.

PR #3502, opened 2018-05-31
Avoid 0xdead10cc crashes in iRate.

charlesmchen-signal:
PTAL @michaelkirk


Libraries/iRate/iRate.m
```diff
@@ -972,9 +977,6 @@ - (void)didDismissAlert:(__unused id)alertView withButtonAtIndex:(NSInteger)butt
 - (void)applicationWillEnterForeground {
     if ([UIApplication sharedApplication].applicationState == UIApplicationStateBackground) {
         [self incrementUseCount];
-        if (self.promptAtLaunch) {
-            [self promptIfAllCriteriaMet];
-        }
```
charlesmchen-signal: It's not safe to try to "prompt at launch" in `applicationWillEnterForeground`, so we do it instead in `applicationDidBecomeActive`.

PR #3503, opened 2018-05-31
Charlesmchen/reported application state

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -1531,8 +1531,8 @@ typedef enum : NSUInteger {
 
 - (void)autoLoadMoreIfNecessary
 {
-    BOOL isAppInBackground = CurrentAppContext().isInBackground;
-    if (self.isUserScrolling || !self.isViewVisible || isAppInBackground) {
+    BOOL isMainAppAndActive = CurrentAppContext().isMainAppAndActive;
+    if (self.isUserScrolling || !self.isViewVisible || !isMainAppAndActive) {
         return;
     }
     if (!self.showLoadMoreHeader) {
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -233,7 +233,6 @@ typedef enum : NSUInteger {
 
 @property (nonatomic) BOOL isViewCompletelyAppeared;
 @property (nonatomic) BOOL isViewVisible;
-@property (nonatomic) BOOL isAppInBackground;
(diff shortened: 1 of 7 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1523,7 +1524,8 @@ - (void)disappearingTimerConfigurationViewWasTapped:(DisappearingTimerConfigurat
 
 - (void)autoLoadMoreIfNecessary
 {
-    if (self.isUserScrolling || !self.isViewVisible || self.isAppInBackground) {
+    BOOL isAppInBackground = CurrentAppContext().isInBackground;
```
michaelkirk-signal: Do we want `!CurrentAppContext.isForegroundAndActive`?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4421,25 +4423,19 @@ - (void)setIsViewVisible:(BOOL)isViewVisible
     [self updateCellsVisible];
 }
 
-- (void)setIsAppInBackground:(BOOL)isAppInBackground
-{
-    _isAppInBackground = isAppInBackground;
-
-    [self updateShouldObserveDBModifications];
```
michaelkirk-signal: Will this present more redrawing to the user since now we're doing it after we become active?

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -484,16 +483,10 @@ - (void)setIsViewVisible:(BOOL)isViewVisible
...
-{
-    _isAppInBackground = isAppInBackground;
-
-    [self updateShouldObserveDBModifications];
-}
-
 - (void)updateShouldObserveDBModifications
 {
-    self.shouldObserveDBModifications = self.isViewVisible && !self.isAppInBackground;
+    BOOL isAppForegroundAndActive = CurrentAppContext().isAppForegroundAndActive;
```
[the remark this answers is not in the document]
charlesmchen-signal: No difference, but the ternary nature of application state can be confusing and I was just trying to chose a name that was completely unambiguous.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: Better, done.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: I don't think so, but that's definitely something we should verify.

PR #3505, opened 2018-05-31
Fix ordering of contacts picker.

charlesmchen-signal:
PTAL @michaelkirk-signal

### DIFF
--- a/Signal/src/ViewControllers/ContactsPicker.swift
+++ b/Signal/src/ViewControllers/ContactsPicker.swift
@@ -196,13 +196,20 @@ public class ContactsPicker: OWSViewController, UITableViewDelegate, UITableView
     }
 
     func collatedContacts(_ contacts: [CNContact]) -> [[CNContact]] {
-        let selector: Selector = #selector(getter: CNContact.nameForCollating)
+        let sortSelector = #selector(getter: CNContact.nameForCollating)
 
+        // 1. Organize the contacts into sections.
         var collated = Array(repeating: [CNContact](), count: collation.sectionTitles.count)
         for contact in contacts {
-            let sectionNumber = collation.section(for: contact, collationStringSelector: selector)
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ContactsPicker.swift
```diff
@@ -196,13 +196,20 @@ public class ContactsPicker: OWSViewController, UITableViewDelegate, UITableView
...
-            let sectionNumber = collation.section(for: contact, collationStringSelector: selector)
+            let sectionNumber = collation.section(for: contact,
+                                                  collationStringSelector: sortSelector)
             collated[sectionNumber].append(contact)
         }
+        // 2. Sort the contents of each section.
+        collated = collated.map({ (sectionContacts) in
+            return collation.sortedArray(from: sectionContacts,
+                                         collationStringSelector: sortSelector)
+        }) as! [[CNContact]]
```
charlesmchen-signal: The key fix here is that we weren't sorting the contents of each section.

Q: is there a way to avoid using `as!` here?

Signal/src/ViewControllers/ContactsPicker.swift
```diff
@@ -365,7 +372,7 @@ fileprivate extension CNContact {
     @objc var nameForCollating: String {
         get {
             if self.familyName.isEmpty && self.givenName.isEmpty {
-                return self.emailAddresses.first?.value as String? ?? ""
+                return (self.emailAddresses.first?.value as String? ?? "").lowercased().ows_stripped()
```
charlesmchen-signal: We want to use case-insensitive sort.

charlesmchen-signal:
> contactFetchRequest.sortOrder = .userDefault

Done, better.

charlesmchen-signal:
PTAL @michaelkirk

PR #3506, opened 2018-05-31
Conversation load perf: Load less messages initially

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -98,15 +98,22 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-// Always load up to 50 messages when user arrives.
-static const int kYapDatabasePageSize = 50;
-// Never show more than 50*500 = 25k messages in conversation view at a time.
-static const int kYapDatabaseMaxPageCount = 500;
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -98,15 +98,22 @@
...
-static const int kYapDatabaseRangeMaxLength = kYapDatabasePageSize * kYapDatabaseMaxPageCount;
+// Always load up to n messages when user arrives.
+//
+// The smaller this number is, the faster the conversation can display.
+// To test, shrink you accessability font as much as possible, then count how many 1-line system info messages (our shortest cells) can
+// fit on screen at a time on an iPhoneX
+//
+// PERF: we could do less messages on shorter (older, slower) devices
+// PERF: we could cache the cell height, since some messages will be much taller.
+static const int kYapDatabasePageSize = 25;
```
charlesmchen-signal: So this value got halfed, and the other values are the same, right?

And this won't have any effect in terms of our making a best effort to show all unread messages when the user enters the conversation, right?

michaelkirk-signal: Yes - I'll include these details in the test plan.

PR #3508, opened 2018-05-31
Conversation load perf: shared ui dbConnection

michaelkirk-signal:
PTAL @charlesmchen 

Based on https://github.com/yapstudios/YapDatabase/wiki/Performance-Pro#sharing_the_ui_databaseconnection

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2075,6 +2075,12 @@ typedef enum : NSUInteger {
 
     [self dismissKeyBoard];
 
+    // In case we were presenting edit menu, we need to become first responder before presenting another VC
+    // else UIKit won't restore first responder status to us when the presented VC is dismissed.
+    if (!self.isFirstResponder) {
+        [self becomeFirstResponder];
+    }
+
     if (![viewItem.interaction isKindOfClass:[TSMessage class]]) {
         OWSFail(@"Unexpected viewItem.interaction");
         return;
@@ -2098,6 +2104,11 @@ typedef enum : NSUInteger {
     OWSAssert(attachmentStream);
 
     [self dismissKeyBoard];
+    // In case we were presenting edit menu, we need to become first responder before presenting another VC
+    // else UIKit won't restore first responder status to us when the presented VC is dismissed.
+    if (!self.isFirstResponder) {
+        [self becomeFirstResponder];
+    }
 
     if (![viewItem.interaction isKindOfClass:[TSMessage class]]) {
         OWSFail(@"Unexpected viewItem.interaction");
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2466,7 +2466,8 @@ typedef enum : NSUInteger {
         [ThreadUtil ensureDynamicInteractionsForThread:self.thread
                                        contactsManager:self.contactsManager
                                        blockingManager:self.blockingManager
-                                          dbConnection:self.editingDatabaseConnection
+                                  uiDatabaseConnection:self.uiDatabaseConnection
+                             editingDatabaseConnection:self.editingDatabaseConnection
                            hideUnreadMessagesIndicator:self.hasClearedUnreadMessagesIndicator
                        firstUnseenInteractionTimestamp:self.dynamicInteractions.firstUnseenInteractionTimestamp
                                           maxRangeSize:maxRangeSize];
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -157,7 +157,7 @@ typedef enum : NSUInteger {
 @property (nullable, nonatomic) UIPanGestureRecognizer *currentShowMessageDetailsPanGesture;
 
 @property (nonatomic) TSThread *thread;
-@property (nonatomic) YapDatabaseConnection *editingDatabaseConnection;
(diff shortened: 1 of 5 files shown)


SignalMessaging/utils/ThreadUtil.m
```diff
@@ -247,7 +249,8 @@ + (ThreadDynamicInteractions *)ensureDynamicInteractionsForThread:(TSThread *)th
 
     ThreadDynamicInteractions *result = [ThreadDynamicInteractions new];
 
-    [dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
+    NSMutableArray *writeBlocks = [NSMutableArray new];
+    [uiDatabaseConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
```
michaelkirk-signal: We now read from the uiDatabaseConnection, and queue any reads to happen in a single transaction on the editing db connection.

SignalServiceKit/src/Storage/OWSPrimaryStorage.m
```diff
@@ -119,15 +124,72 @@ - (instancetype)initStorage
...
 }
 
+- (void)yapDatabaseModifiedExternally:(NSNotification *)notification
+{
+    // Notify observers we're about to update the database connection
+    [[NSNotificationCenter defaultCenter] postNotificationName:OWSUIDatabaseConnectionWillUpdateExternallyNotification object:self.dbNotificationObject];
+    
+    // Move uiDatabaseConnection to the latest commit.
+    // Do so atomically, and fetch all the notifications for each commit we jump.
+    NSArray *notifications = [self.uiDatabaseConnection beginLongLivedReadTransaction];
```
charlesmchen-signal: `beginLongLivedReadTransaction` is expensive.  When we send a message in the SAE, the main app will get 20-30 or more `YapDatabaseModifiedExternallyNotification`.  Previously, we weren't doing any work in response to them because our views were ignoring `YapDatabaseModifiedExternallyNotification` if the view wasn't visible and the app wasn't active.  

I'm not too concerned about the perf cost since the main app will be in the background in this case, but FYI.

SignalMessaging/utils/ThreadUtil.m
```diff
@@ -305,7 +308,9 @@ + (ThreadDynamicInteractions *)ensureDynamicInteractionsForThread:(TSThread *)th
 
         for (TSInteraction *interaction in interactionsToDelete) {
             DDLogDebug(@"Cleaning up interaction: %@", [interaction class]);
-            [interaction removeWithTransaction:transaction];
+            [writeBlocks addObject:^(YapDatabaseReadWriteTransaction *writeTransaction) {
+                [interaction removeWithTransaction:writeTransaction];
+            }];
```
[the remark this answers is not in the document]
michaelkirk-signal: Well there's two goals. 

One is to avoid loading extensions, and the other is to avoid opening write transactions when we can avoid it.

Another thought, since we're doing this on the write connection anyway, what about making this all async?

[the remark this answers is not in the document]
charlesmchen-signal: > Another thought, since we're doing this on the write connection anyway, what about making this all async?

I'm not 100% sure, but I think it's necessary to do this sync otherwise the dynamic interactions will often "move around" right after we enter the conversation.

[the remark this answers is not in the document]
michaelkirk-signal: >  the dynamic interactions will often "move around"

Yeah, good point. 

I've done as you've suggested and moved to a shared editing db connection.

I'm a little bummed that we weren't able to get rid of opening this blocking write transaction every time this method is called, but I don't have a simple solution, so I'll park that aspiration for now.

I kept the shared ui connection, since it has similar benefits, we just don't use it for ensuring dynamic interactions anymore. To be clear, I'm advocating we use shared connections almost everywhere - and work to eliminate existing uiDatabaseConnections and one-off db connections as refactoring allows or performance requirements dictate.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -676,6 +684,11 @@ - (void)viewWillAppear:(BOOL)animated
     }
 
     [self updateLastVisibleTimestamp];
+
+    if (!self.viewHasEverAppeared) {
+        NSTimeInterval appearenceDuration = CACurrentMediaTime() - self.viewControllerCreatedAt;
+        DDLogInfo(@"%@ First viewWillAppear took: %.2fms", self.logTag, appearenceDuration * 1000);
+    }
```
charlesmchen-signal: Did you mean to commit this?  If you want to leave this in, can we make it "verbose"?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -158,6 +159,7 @@ @interface ConversationViewController () <AttachmentApprovalViewControllerDelega
 @property (nonatomic) TSThread *thread;
 @property (nonatomic) YapDatabaseConnection *editingDatabaseConnection;
```
charlesmchen-signal: Don't we want to get rid of the `editingDatabaseConnection` property? For that matter, shouldn't we get rid of `uiDatabaseConnection` too?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: I did mean to keep it. I'll reduce the log level.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: Good catch, I'll make them readonly / computed.

PR #3515, opened 2018-06-01
Fix: input toolbar disappears when tapping media while edit menu is presented

michaelkirk-signal:
Longer term I'd like to replace the UIMenuController with a custom look-a-like that avoids the first responder pattern.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2059,6 +2059,12 @@ typedef enum : NSUInteger {
 
     [self dismissKeyBoard];
 
+    // In case we were presenting edit menu, we need to become first responder before presenting another VC
+    // else UIKit won't restore first responder status to us when the presented VC is dismissed.
+    if (!self.isFirstResponder) {
+        [self becomeFirstResponder];
+    }
+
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2082,6 +2088,11 @@ - (void)didTapVideoViewItem:(ConversationViewItem *)viewItem
     OWSAssert(attachmentStream);
 
     [self dismissKeyBoard];
+    // In case we were presenting edit menu, we need to become first responder before presenting another VC
+    // else UIKit won't restore first responder status to us when the presented VC is dismissed.
+    if (!self.isFirstResponder) {
+        [self becomeFirstResponder];
+    }
```
charlesmchen-signal: This fixes one repro (tapping media after showing menu controller) but not others (e.g. showing an alert (e.g. by tapping "a call item") after showing menu controller.  

I've got a counter-proposal here that covers all of the repros I know of and IMO feels less whack-a-mole:

https://github.com/signalapp/Signal-iOS/pull/3516

michaelkirk-signal: Much better! I'll close this.

PR #3516, opened 2018-06-01
Ensure conversation view is first responder before presenting another view.

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -5114,6 +5114,20 @@ - (void)didCreateOrEditContact
...
+#pragma mark -
+
+- (void)presentViewController:(UIViewController *)viewController
+                     animated:(BOOL)animated
+                   completion:(void (^__nullable)(void))completion
+{
+    // Ensure that we are first responder before presenting other views.
+    // This ensures that the input toolbar will be restored after the
+    // presented view is dismissed.
+    [self becomeFirstResponder];
```
michaelkirk-signal: I think we should only do this if we aren't already first responder, otherwise LGTM!

PR #3518, opened 2018-06-01
Charlesmchen/reduce jsq

charlesmchen-signal:
PTAL @michaelkirk 

We're still using `JSQMessagesAvatarImageFactory`.


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2702,7 +2692,9 @@ - (void)messageWasSent:(TSOutgoingMessage *)message
     self.inputToolbar.quotedReply = nil;
 
     if ([Environment.preferences soundInForeground]) {
-        [JSQSystemSoundPlayer jsq_playMessageSentSound];
+        SystemSoundID soundId = [OWSSounds systemSoundIDForSound:OWSSound_MessageSent quiet:YES];
+        // Vibrate, respect silent switch, respect "Alert" volume, not media volume.
```
michaelkirk-signal: Do we want to vibrate, respect silent switch, and respect alert volume here?

Can you compare with behavior in messages/whatsapp - I'd be surprised.

Signal/translations/ar.lproj/Localizable.strings
```diff
@@ -2257,7 +2257,7 @@
 /* Info message embedding a {{time amount}}, see the *_TIME_AMOUNT strings for context. */
 "YOU_UPDATED_DISAPPEARING_MESSAGES_CONFIGURATION" = "لقد ضبطت وقت إخفاء الرسالة على %@.";
 
-// Strings Copied in from JSQMessagesViewController
+
```
michaelkirk-signal: This change (and the other localization file changes outside of en.lproj) make me thing you edited them manually.

Generally we shouldn't be editing any translation files manually - did you update autogenstrings script appropriately?

I'm not actually worried about the change itself - it just seems like if you're making manual tweaks, they're going to be blown away the next time we run autogenstrings -> sync-translations since these are generated files.

SignalMessaging/environment/OWSSounds.m
```diff
@@ -172,6 +172,8 @@ + (NSString *)displayNameForSound:(OWSSound)sound
             return @"Call Busy";
         case OWSSound_CallFailure:
             return @"Call Failure";
+        case OWSSound_MessageSent:
```
michaelkirk-signal: Should we increase our LRU sound cache size by 1?

Signal/translations/ar.lproj/Localizable.strings
charlesmchen-signal: These were just some whitespace changes left behind by global search-and-replace.  I've reverted them.

SignalMessaging/environment/OWSSounds.m
charlesmchen-signal: Good idea, done.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: You're right, we should be using `AudioServicesPlaySystemSound()` not `AudioServicesPlayAlertSound()` as I had it to match previous behavior in `JSQSystemSoundPlayer jsq_playMessageSentSound`

PR #3520, opened 2018-06-01
https://github.com/signalapp/Signal-iOS/issues/3455: Treat AMR files …

Gabardone:
…as generic attachments (AVFoundation doesn't support playback)

### Contributor checklist

 * iPhone 8+, iOS 11.3.1
 * iPad Pro (compatibility mode), iOS 11.3.1

- - - - - - - - - -

### Description
AVFoundation (specifically AVAudioPlayer) doesn't support playback of .amr files, but it will still load them and report their duration.

Since they won't play, it's better to treat them as generic file attachments until such time as Apple brings back support or a different solution is brought to make them playable if there's enough demand.

Fortunately just removing them from the dictionaries in MIMETypeUtil.m does the trick.


michaelkirk-signal:
LGTM - but holding off on merging for a minute, not sure if we want this in the 2.26 release, or after.

michaelkirk-signal:
In 00c2d47a9 - fyi @charlesmchen I've added a note in the test plan.

PR #3526, opened 2018-06-06
Check contacts access when notified of contacts changes.

charlesmchen-signal:
PTAL @michaelkirk


SignalMessaging/contacts/SystemContactsFetcher.swift
```diff
@@ -266,6 +267,19 @@ public class SystemContactsFetcher: NSObject {
...
 
+    @objc
+    public func refreshAfterContactsChange() {
+        SwiftAssertIsOnMainThread(#function)
+
+        guard authorizationStatus == .authorized else {
+            Logger.info("\(logTag) ignoring contacts change; no access.")
+            self.delegate?.systemContactsFetcher(self, hasAuthorizationStatus: authorizationStatus)
+            return
+        }
```
charlesmchen-signal: By checking here we improve the logging and avoid a spurious `OWSFail()`.

PR #3532, opened 2018-06-10
Fixed settings button margin in home view controller, also fixes land…

Gabardone:
…tches.

### Contributor checklist

 * iPad Pro in iPhone compatibility mode.
 * iPhone 5S simulator
 * iPhone 8+ (landscape glitch doesn't manifest but margin does correct itself).

- - - - - - - - - -

### Description
I've started running Signal in landscape mode to see what needs to get fixed to enable it for good and as an intermediate step to getting an iPad version up. The first thing I noticed is that several of the nav bar icons appear squished when moving to landscape.

Looking at the logic a few things became apparent:
1) The custom UIBarButtonItem seems to have the opposite effect from what the extensive comments around its creation say, as it moves it further in and leaves less space in the nav bar. It's also unclear when the contents of the nav bar are cramped as even in an iPad in compatibility mode (original iPhone screen size) there's more than enough space for "Signal" and two icons.
2) The accessibility labels in the button and its wrapping UIBarButtonItem are the same and redundant. Most of the patch is removing the extra label from all localizations.

I may be missing something, like a UI state where there's a number of extra things in the nav bar or different behaviors in older versions of iOS for nav bar buttons, so let me know if that's the case. I'll add that once seen that the button is farther to the right than usual it cannot be unseen 😝

For reference, this is how the home view looks currently:

screen shot 2018-06-10 at 12 54 46

And here's how it looks with the changes:

patched screen shot 2018-06-10 at 12 51 27

And for further reference (not that we'll be enabling landscape mode for now...), here's how landscape (manually enabled) mode looks in current master:

screen shot 2018-06-10 at 12 54 34

And here's how it looks with the patch:

patched screen shot 2018-06-10 at 12 51 36

Let me know if any further tweaks are needed to this patch or if there's a reason I'm unaware of to keep things as they are.

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -329,40 +329,11 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
     if (self.homeViewMode != HomeViewMode_Inbox) {
         return;
     }
-    const CGFloat kBarButtonSize = 44;
-    // We use UIButtons with [UIBarButtonItem initWithCustomView:...] instead of
-    // UIBarButtonItem in order to ensure that these buttons are spaced tightly.
-    // The contents of the navigation bar are cramped in this view.
-    UIButton *button = [UIButton buttonWithType:UIButtonTypeCustom];
-    UIImage *image = [UIImage imageNamed:@"button_settings_white"];
-    [button setImage:image forState:UIControlStateNormal];
-    UIEdgeInsets imageEdgeInsets = UIEdgeInsetsZero;
-    // We normally would want to use left and right insets that ensure the button
-    // is square and the icon is centered.  However UINavigationBar doesn't offer us
-    // control over the margins and spacing of its content, and the buttons end up
-    // too far apart and too far from the edge of the screen. So we use a smaller
-    // leading inset tighten up the layout.
-    CGFloat hInset = round((kBarButtonSize - image.size.width) * 0.5f);
-    if (self.view.isRTL) {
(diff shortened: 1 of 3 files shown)


Signal/translations/ar.lproj/Localizable.strings
```diff
@@ -1873,9 +1873,6 @@
 /* Label for the block list section of the settings view */
 "SETTINGS_BLOCK_LIST_TITLE" = "تم الحظر";
 
-/* Accessibility hint for the settings button */
```
michaelkirk-signal: I agree with removing this redundant string, and thanks for trying to keep things tidy, but just an FYI for how we manage localizations:

1. Strings/comments are added and removed from the source code as needed using the NSLocalizedString macros.

2. run `Signal/translations/bin/auto-genstrings` to extract translations into en.lproj and remove stale entries from en.lproj. (never add or remove entries directly)

3. Never edit any translations/*.lproj/* files except English. The other localization files are only edited by Signal staff as a result of our release management scripts.

So, just to keep our process orderly, could you remove these changes from the l10n files?

I've added some notes to our [Translations Readme](https://github.com/signalapp/Signal-iOS/blob/master/Signal/translations/TRANSLATIONS.md)

Gabardone: 👍🏽 I'll modify the patch.

Gabardone:
Ok updated, thanks for the pointers about localization process 👍🏽

Also fixed the script, it didn't seem to account for the current position of the JSQMessagesViewController pod.

michaelkirk-signal:
Rebased into 24f30e015, sorry for the delay.

PR #3533, opened 2018-06-11
fts search results controller

michaelkirk-signal:
PTAL @charlesmchen 

This is a rough cut of both the backend and front end for full text search. Some functionality is a little rough, while some is just missing.

While you're reviewing, I'll make sure all outstanding work is cut into separate cards.

### DIFF
--- a/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
+++ b/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
@@ -31,11 +31,14 @@ class ConversationSearchViewController: UITableViewController {
 
         self.view.isHidden = true
 
-        self.tableView.register(ChatSearchResultCell.self, forCellReuseIdentifier: ChatSearchResultCell.reuseIdentifier)
-        self.tableView.register(MessageSearchResultCell.self, forCellReuseIdentifier: MessageSearchResultCell.reuseIdentifier)
+        tableView.rowHeight = UITableViewAutomaticDimension
+        tableView.estimatedRowHeight = 60
+
+        tableView.register(ChatSearchResultCell.self, forCellReuseIdentifier: ChatSearchResultCell.reuseIdentifier)
+        tableView.register(MessageSearchResultCell.self, forCellReuseIdentifier: MessageSearchResultCell.reuseIdentifier)
     }
 
-    // MARK: UITableViewDelegate
+    // MARK: UITableViewDataSource
 
     override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
         guard let searchSection = SearchSection(rawValue: section) else {
@@ -50,99 +53,6 @@ class ConversationSearchViewController: UITableViewController {
             return searchResultSet.contacts.count
         case .messages:
             return searchResultSet.messages.count
-        }
-    }
-
-    class ChatSearchResultCell: UITableViewCell {
-        static let reuseIdentifier = "ChatSearchResultCell"
-
-        let nameLabel: UILabel
-        let snippetLabel: UILabel
-        let avatarView: AvatarImageView
-
-        override init(style: UITableViewCellStyle, reuseIdentifier: String?) {
-            self.nameLabel = UILabel()
-            self.snippetLabel = UILabel()
-            self.avatarView = AvatarImageView()
-            avatarView.autoSetDimensions(to: CGSize(width: 40, height: 40))
-
-            super.init(style: style, reuseIdentifier: reuseIdentifier)
-
-            let textRows = UIStackView(arrangedSubviews: [nameLabel, snippetLabel])
-            textRows.axis = .vertical
-
-            let columns = UIStackView(arrangedSubviews: [avatarView, textRows])
-            columns.axis = .horizontal
-            columns.spacing = 8
-
-            contentView.addSubview(columns)
-            columns.autoPinEdgesToSuperviewMargins()
-        }
-
-        required init?(coder aDecoder: NSCoder) {
-            fatalError("init(coder:) has not been implemented")
-        }
-
-        func configure(searchResult: SearchResult) {
-            self.nameLabel.text = searchResult.thread.name
-            self.snippetLabel.text = searchResult.snippet
-        }
-    }
-
-    class MessageSearchResultCell: UITableViewCell {
-        static let reuseIdentifier = "MessageSearchResultCell"
-
-        let nameLabel: UILabel
-        let snippetLabel: UILabel
-
-        override init(style: UITableViewCellStyle, reuseIdentifier: String?) {
-            self.nameLabel = UILabel()
-            self.snippetLabel = UILabel()
-
-            super.init(style: style, reuseIdentifier: reuseIdentifier)
-
-            let textRows = UIStackView(arrangedSubviews: [nameLabel, snippetLabel])
-            textRows.axis = .vertical
-
-            contentView.addSubview(textRows)
-            textRows.autoPinEdgesToSuperviewMargins()
-        }
-
-        required init?(coder aDecoder: NSCoder) {
-            fatalError("init(coder:) has not been implemented")
-        }
-
-        func configure(searchResult: SearchResult) {
-            self.nameLabel.text = searchResult.thread.name
-
-//            [[NSAttributedString alloc] initWithData:[htmlString dataUsingEncoding:NSUTF8StringEncoding]
-//                options:@{NSDocumentTypeDocumentAttribute: NSHTMLTextDocumentType,
-//                NSCharacterEncodingDocumentAttribute: @(NSUTF8StringEncoding)}
-//                documentAttributes:nil error:nil];
-
-            guard let snippet = searchResult.snippet else {
-                self.snippetLabel.text = nil
-                return
-            }
-
-            guard let encodedString = snippet.data(using: .utf8) else {
-                self.snippetLabel.text = nil
-                return
-            }
-
-//            NSAttributedString(data: <#T##Data#>, options: [, ], documentAttributes: <#T##AutoreleasingUnsafeMutablePointer<NSDictionary?>?#>)
-            // Bold snippet text
-            do {
-
-                // TODO   NSAttributedString.DocumentReadingOptionKey.characterEncoding: .utf8
-                let attributedSnippet = try NSAttributedString(data: encodedString,
-                                                               options: [NSAttributedString.DocumentReadingOptionKey.documentType: NSAttributedString.DocumentType.html],
-                                                               documentAttributes: nil)
-
-                self.snippetLabel.attributedText = attributedSnippet
-            } catch {
-                owsFail("failed to generate snippet: \(error)")
-            }
         }
     }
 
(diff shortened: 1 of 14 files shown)


Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -0,0 +1,244 @@
...
+
+    // MARK: View Lifecyle
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        tableView.rowHeight = UITableViewAutomaticDimension
+        tableView.estimatedRowHeight = 60
+
+        tableView.register(ChatSearchResultCell.self, forCellReuseIdentifier: ChatSearchResultCell.reuseIdentifier)
```
charlesmchen-signal: Why Chat and not `ThreadSearchResultCell` or `ConversationSearchResultCell`?

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -0,0 +1,244 @@
...
+                return UITableViewCell()
+            }
+
+            cell.configure(searchResult: searchResult)
+            return cell
+        }
+    }
+
+    override func numberOfSections(in tableView: UITableView) -> Int {
+        return 3
```
charlesmchen-signal: return `SearchSection.allCases.count`

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    var searcher: ConversationSearcher {
+        return ConversationSearcher.shared
+    }
+
+    enum SearchSection: Int {
+        case conversations = 0
+        case contacts = 1
+        case messages = 2
```
charlesmchen-signal: Should we have an additional "no matches" section we only show if there are no matches?

Signal/test/Models/MesssagesBubblesSizeCalculatorTest.swift
```diff
@@ -48,70 +48,70 @@ class MesssagesBubblesSizeCalculatorTest: XCTestCase {
         let text: String? = ""
         let viewItem = self.viewItemForText(text)
         let actual = messageBubbleSize(for: viewItem)
-        XCTAssertEqual(42, actual.height)
+        XCTAssertEqual(36, actual.height)
```
charlesmchen-signal: IIRC these values reflect a certain version of iOS, a certain device model, and certain dynamic text sizes.

SignalMessaging/utils/ConversationSearcher.swift
```diff
@@ -1,19 +1,77 @@
...
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 import Foundation
 import SignalServiceKit
 
+public class SearchResult {
+    public let thread: ThreadViewModel
+    public let snippet: String?
```
charlesmchen-signal: We need to be careful about assuming we can represent contact matches with `ThreadViewModel`.  Not all contacts have threads.  We want to avoid creating an unsaved `TSThread` instance and/or overwriting `TSThread` instances.

SignalMessaging/utils/ConversationSearcher.swift
```diff
@@ -1,19 +1,77 @@
...
+
```
charlesmchen-signal: Should we add an optional "message id"?

charlesmchen-signal: Or rather an optional TODO to add an optional message id?

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
```diff
@@ -0,0 +1,141 @@
...
+
+        let normalized = FullTextSearchFinder.normalize(text: searchText)
+
+        // We want a forgiving query for phone numbers
+        // TODO a stricter "whole word" query for body text?
+        let prefixQuery = "*\(normalized)*"
+
+        // (snippet: String, collection: String, key: String, object: Any, stop: UnsafeMutablePointer<ObjCBool>)
+        ext.enumerateKeysAndObjects(matching: prefixQuery, with: nil) { (snippet: String, _: String, _: String, object: Any, _: UnsafeMutablePointer<ObjCBool>) in
+            block(object, snippet)
```
charlesmchen-signal: Should we `stop` after some constant max number of matches?

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
```diff
@@ -0,0 +1,141 @@
...
+        guard let ext: YapDatabaseFullTextSearchTransaction = ext(transaction: transaction) else {
+            assertionFailure("ext was unexpectedly nil")
+            return
+        }
+
+        let normalized = FullTextSearchFinder.normalize(text: searchText)
+
+        // We want a forgiving query for phone numbers
+        // TODO a stricter "whole word" query for body text?
+        let prefixQuery = "*\(normalized)*"
```
charlesmchen-signal: Why is this `prefixQuery` and not just `query`?

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    // MARK: View Lifecyle
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        tableView.rowHeight = UITableViewAutomaticDimension
+        tableView.estimatedRowHeight = 60
```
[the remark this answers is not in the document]
michaelkirk-signal: Note this is **estimated**RowHeight - we're sizing the cells with autolayout, so I'm not sure what's TODO here.

SignalMessaging/utils/ConversationSearcher.swift
michaelkirk-signal: Right - I'm assuming we shouldn't represent contact matches with a threadViewModel and that this will change somehow.

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
```diff
@@ -0,0 +1,141 @@
...
+
+    private static let recipientIndexer: SearchIndexer<String> = SearchIndexer { (recipientId: String) in
+        let displayName = contactsManager.displayName(forPhoneIdentifier: recipientId)
+        let searchableContent =  "\(recipientId) \(displayName)"
+
+        return normalize(text: searchableContent)
+    }
+
+    private static let messageIndexer: SearchIndexer<TSMessage> = SearchIndexer { (message: TSMessage) in
+        let searchableContent =  message.body ?? ""
```
charlesmchen-signal: Should we add a TODO to include:

* text from oversize text attachments.
* text from quoted replies.
* etc.?

SignalMessaging/utils/ConversationSearcher.swift
michaelkirk-signal: Sure - if that's what makes sense when adding the Contact results.

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
```diff
@@ -0,0 +1,141 @@
...
+    private static let contactThreadIndexer: SearchIndexer<TSContactThread> = SearchIndexer { (contactThread: TSContactThread) in
+        let recipientId =  contactThread.contactIdentifier()
+        let searchableContent = recipientIndexer.index(recipientId)
+
+        return normalize(text: searchableContent)
+    }
+
+    private static let recipientIndexer: SearchIndexer<String> = SearchIndexer { (recipientId: String) in
+        let displayName = contactsManager.displayName(forPhoneIdentifier: recipientId)
+        let searchableContent =  "\(recipientId) \(displayName)"
```
charlesmchen-signal: I wonder if we should include system contact names, profile names, phone number.

We'd want `+13213214321` and `(321) 321-4321)` to match `+13213214321`, right?

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
```diff
@@ -0,0 +1,141 @@
...
+        }.joined(separator: " ")
+
+        let searchableContent = "\(groupName) \(memberStrings)"
+
+        return normalize(text: searchableContent)
+    }
+
+    private static let contactThreadIndexer: SearchIndexer<TSContactThread> = SearchIndexer { (contactThread: TSContactThread) in
+        let recipientId =  contactThread.contactIdentifier()
+        let searchableContent = recipientIndexer.index(recipientId)
```
charlesmchen-signal: I wonder if we should include system contact names, profile names, phone number.

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
michaelkirk-signal: FTS is doing whole word matches, so without the prefixQuery

- "Al" doesn't match "Alice"
- "123" doesn't match "1-234-567-1234"

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private static let groupThreadIndexer: SearchIndexer<TSGroupThread> = SearchIndexer { (groupThread: TSGroupThread) in
+        let groupName = groupThread.groupModel.groupName ?? ""
+
+        let memberStrings = groupThread.groupModel.groupMemberIds.map { recipientId in
+            recipientIndexer.index(recipientId)
+        }.joined(separator: " ")
+
+        let searchableContent = "\(groupName) \(memberStrings)"
```
charlesmchen-signal: So with this arrangement, could a snippet span a group name and some other content?  Should we use some kind of delimiting token and chop up snippets using those delimiters?

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
michaelkirk-signal: I'm not compelled to index replies by the thing that they're quoting. It wouldn't be difficult from the search index standpoint, we'd just have the indexer include the text.

But showing the user what's matched, and disambiguating the quote from what was quoted seems difficult and dubiously useful.

Searching oversize text attachments seems useful and without the UI pain points, so I'll add a card, but they're not very commonly used, but I'd suggest it's lower priority work.

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
michaelkirk-signal: > We'd want +13213214321 and (321) 321-4321) to match +13213214321, right?

Yes - there's a card for this.

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
michaelkirk-signal: If I have a group called "Book Club" and a member named "Alice"
When I search "Book Alice"

I (personally) would expect "Book Club" to show up, which is the current behavior. It sounds like you're suggesting the opposite.

Can you give me an example in terms of user behavior and the undesirable outcome the existing approach could produce?

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
michaelkirk-signal: We're already searching by their Signal phone number. 

I'll add "index by non-signal phone number" and "index by profile name" in the nice-to-have category.

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
michaelkirk-signal: That seems like a good call, I'll make that change before merging.

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
michaelkirk-signal: There is an existing card for "No results". If adding another section proves to be the best way to do that, that's fine.

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
michaelkirk-signal: Thread is sort of an overloaded term in any programming context. I'll change it to the longer "ConversationSearchResultCell".

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
michaelkirk-signal: I didn't know about `allCases` - looks like it's slated for Swift 4.2.

https://github.com/apple/swift-evolution/blob/master/proposals/0194-derived-collection-of-enum-cases.md

PR #3534, opened 2018-06-11
Open search results.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
+++ b/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
@@ -36,6 +36,36 @@ class ConversationSearchViewController: UITableViewController {
         tableView.register(MessageSearchResultCell.self, forCellReuseIdentifier: MessageSearchResultCell.reuseIdentifier)
     }
 
+    // MARK: UITableViewDelegate
+
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: false)
+
+        guard let searchSection = SearchSection(rawValue: indexPath.section) else {
+            owsFail("\(logTag) unknown section selected.")
+            return
+        }
+
+        var sectionResults: [SearchResult]
+        switch searchSection {
+        case .conversations:
+            sectionResults = searchResultSet.conversations
+        case .contacts:
+            sectionResults = searchResultSet.contacts
+        case .messages:
+            sectionResults = searchResultSet.messages
+        }
+
+        guard indexPath.row < sectionResults.count else {
+            owsFail("\(logTag) unknown row selected.")
+            return
+        }
+
+        let searchResult = sectionResults[indexPath.row]
+        let thread = searchResult.thread
+        SignalApp.shared().presentConversation(for: thread.threadRecord, action: .compose)
+    }
+
     // MARK: UITableViewDataSource
 
     override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -36,6 +36,36 @@ class ConversationSearchViewController: UITableViewController {
...
+            owsFail("\(logTag) unknown section selected.")
+            return
+        }
+
+        var sectionResults: [SearchResult]
+        switch searchSection {
+        case .conversations:
+            sectionResults = searchResultSet.conversations
+        case .contacts:
+            sectionResults = searchResultSet.contacts
```
charlesmchen-signal: Presumably we'll revisit this as we flush out "Search for contacts".

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -296,16 +299,29 @@ - (void)viewDidLoad
...
-    UISearchController *searchController = [[UISearchController alloc] initWithSearchResultsController:searchResultsController];
-    self.searchController = searchController;
-    searchController.searchResultsUpdater = self;
-    self.tableView.tableHeaderView = self.searchController.searchBar;
-    self.definesPresentationContext = YES;
-        
+    [self addChildViewController:searchResultsController];
+    [self.view addSubview:searchResultsController.view];
+    [searchResultsController.view autoPinWidthToSuperview];
+    [searchResultsController.view autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:searchBar];
```
charlesmchen-signal: I declined to do anything fancier for now. It should only be possible to interact with the search bar when the normal home view content is already scrolled to the top of the screen. If necessary we could ensure that.

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -859,11 +875,44 @@ - (BOOL)tableView:(UITableView *)tableView canEditRowAtIndexPath:(NSIndexPath *)
...
+}
 
-- (void)updateSearchResultsForSearchController:(UISearchController *)searchController
+- (void)updateSearchResultsVisibility
 {
-    [self.searchResultsController updateSearchResultsWithSearchText:self.searchController.searchBar.text];
+    OWSAssertIsOnMainThread();
+
+    [self.searchResultsController updateSearchResultsWithSearchText:self.searchBar.text];
+    BOOL isSearching = self.searchBar.text.length > 0;
```
michaelkirk-signal: Should we use ows_stripped,  since it's less obvious when you're in "search" mode now, it might be good to only trigger the search results view when there is visible content in the search bar.

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -36,6 +36,36 @@ class ConversationSearchViewController: UITableViewController {
...
+        case .messages:
+            sectionResults = searchResultSet.messages
```
michaelkirk-signal: Just making sure you noticed there is a separate card to go to the precise message that was matched by the search. This is good stand-in functionality for now though. 👍

charlesmchen-signal: Yup.

Signal/src/ViewControllers/HomeView/HomeViewController.m
charlesmchen-signal: done.

PR #3535, opened 2018-06-11
Include contacts in FTS results

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/SignalMessaging/utils/ConversationSearcher.swift
+++ b/SignalMessaging/utils/ConversationSearcher.swift
@@ -5,7 +5,7 @@
 import Foundation
 import SignalServiceKit
 
-public class SearchResult {
+public class ConversationSearchResult {
     public let thread: ThreadViewModel
     public let snippet: String?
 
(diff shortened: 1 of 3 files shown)


SignalMessaging/utils/ConversationSearcher.swift
```diff
@@ -44,32 +55,42 @@ public class ConversationSearcher: NSObject {
...
-                // TODO show "other contact" results when there is no existing thread
+                let searchResult = ContactSearchResult(signalAccount: signalAccount)
+                contacts.append(searchResult)
             } else {
                 Logger.debug("\(self.logTag) in \(#function) unhandled item: \(match)")
             }
         }
 
-        return SearchResultSet(conversations: conversations, contacts: contacts, messages: messages)
+        // Only show contacts which were not included in an existing conversation.
```
charlesmchen-signal: s/existing conversation/existing 1:1 conversation/

michaelkirk-signal: clarified - thanks.

PR #3539, opened 2018-06-11
Open message search results.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -195,6 +195,8 @@ typedef enum : NSUInteger {
 
 @property (nonatomic) NSUInteger lastRangeLength;
 @property (nonatomic) ConversationViewAction actionOnOpen;
+@property (nonatomic, nullable) NSString *focusMessageIdOnOpen;
+
 @property (nonatomic) BOOL peek;
 
 @property (nonatomic, readonly) OWSContactsManager *contactsManager;
@@ -426,11 +428,16 @@ typedef enum : NSUInteger {
     [self hideInputIfNeeded];
 }
 
-- (void)configureForThread:(TSThread *)thread action:(ConversationViewAction)action
+- (void)configureForThread:(TSThread *)thread
+                    action:(ConversationViewAction)action
+            focusMessageId:(nullable NSString *)focusMessageId
 {
+    OWSAssert(thread);
+
     _thread = thread;
     _isGroupConversation = [self.thread isKindOfClass:[TSGroupThread class]];
     self.actionOnOpen = action;
+    self.focusMessageIdOnOpen = focusMessageId;
     _cellMediaCache = [NSCache new];
     // Cache the cell media for ~24 cells.
     self.cellMediaCache.countLimit = 24;
@@ -698,13 +705,35 @@ typedef enum : NSUInteger {
     return nil;
 }
 
+- (NSIndexPath *_Nullable)indexPathOfMessageOnOpen
+{
+    OWSAssert(self.focusMessageIdOnOpen);
+
+    NSInteger row = 0;
+    for (ConversationViewItem *viewItem in self.viewItems) {
+        if ([viewItem.interaction.uniqueId isEqualToString:self.focusMessageIdOnOpen]) {
+            return [NSIndexPath indexPathForRow:row inSection:0];
+        }
+        row++;
+    }
+    return nil;
+}
+
 - (void)scrollToDefaultPosition
 {
     if (self.isUserScrolling) {
         return;
     }
 
-    NSIndexPath *_Nullable indexPath = [self indexPathOfUnreadMessagesIndicator];
+    NSIndexPath *_Nullable indexPath = nil;
+    if (self.focusMessageIdOnOpen) {
+        indexPath = [self indexPathOfMessageOnOpen];
+    }
+
+    if (!indexPath) {
+        indexPath = [self indexPathOfUnreadMessagesIndicator];
+    }
+
     if (indexPath) {
         if (indexPath.section == 0 && indexPath.row == 0) {
             [self.collectionView setContentOffset:CGPointZero animated:NO];
@@ -1081,8 +1110,9 @@ typedef enum : NSUInteger {
             break;
     }
 
+    // Clear the "on open" state after the view has been presented.
     self.actionOnOpen = ConversationViewActionNone;
-
+    self.focusMessageIdOnOpen = nil;
 
     self.isViewCompletelyAppeared = YES;
     self.viewHasEverAppeared = YES;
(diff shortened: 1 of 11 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -426,11 +428,16 @@ - (void)popped
     [self hideInputIfNeeded];
 }
 
-- (void)configureForThread:(TSThread *)thread action:(ConversationViewAction)action
+- (void)configureForThread:(TSThread *)thread
+                    action:(ConversationViewAction)action
+            focusMessageId:(nullable NSString *)focusMessageId
```
charlesmchen-signal: If there's a name you prefer over `focusMessageId`, I'm open to it.

Signal/src/call/OutboundCallInitiator.swift
```diff
@@ -45,10 +45,7 @@ import SignalMessaging
         isVideo: Bool) -> Bool {
         // Rather than an init-assigned dependency property, we access `callUIAdapter` via Environment
         // because it can change after app launch due to user settings
-        guard let callUIAdapter = SignalApp.shared().callUIAdapter else {
-            owsFail("\(TAG) can't initiate call because callUIAdapter is nil")
-            return false
-        }
+        let callUIAdapter = SignalApp.shared().callUIAdapter
```
charlesmchen-signal: I'm cleaning up nullability in `SignalApp`.

SignalMessaging/utils/ConversationSearcher.swift
```diff
@@ -76,14 +80,14 @@ public class ConversationSearcher: NSObject {
...
-                let searchResult = ConversationSearchResult(thread: threadViewModel, snippet: snippet)
+                let searchResult = ConversationSearchResult(thread: threadViewModel, messageId: message.uniqueId, snippet: snippet)
 
                 messages.append(searchResult)
             } else if let signalAccount = match as? SignalAccount {
                 let searchResult = ContactSearchResult(signalAccount: signalAccount)
                 contacts.append(searchResult)
             } else {
-                Logger.debug("\(self.logTag) in \(#function) unhandled item: \(match)")
+                owsFail("\(self.logTag) in \(#function) unhandled item: \(match)")
```
charlesmchen-signal: Note the fail.

SignalMessaging/utils/ThreadUtil.m
```diff
@@ -615,11 +618,46 @@ + (ThreadDynamicInteractions *)ensureDynamicInteractionsForThread:(TSThread *)th
...
+                                      focusMessageId:(NSString *)focusMessageId
+{
+    OWSAssert(thread);
+    OWSAssert(transaction);
+    OWSAssert(focusMessageId);
+
+    // Enumerate in reverse to count the number of messages after the "focus message".
+    __block NSUInteger count = 0;
+    __block BOOL didMatch = NO;
+    [[transaction ext:TSMessageDatabaseViewExtensionName]
```
charlesmchen-signal: This simply enumerates. We could do binary search, but is it worth it?  

* Searches are rare.
* Search results with recent results will be more useful.
* We only enumerate the collections keys, not objects, which will help considerably.

charlesmchen-signal:

PTAL @michaelkirk

SignalMessaging/utils/ThreadUtil.m
michaelkirk-signal: The `ensureDynamicInteractions` method is already a (the?) major cost of delay when presenting the ConversationVC, so I'd like to try to avoid introducing a linear scan here, even though it's only when searching.

Any reason we can't use the constant time:

```
- (BOOL)getGroup:(NSString * _Nullable * _Nullable)groupPtr
           index:(nullable NSUInteger *)indexPtr
          forKey:(NSString *)key
    inCollection:(nullable NSString *)collection;
```

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -698,13 +705,35 @@ - (NSIndexPath *_Nullable)indexPathOfUnreadMessagesIndicator
     return nil;
 }
 
+- (NSIndexPath *_Nullable)indexPathOfMessageOnOpen
+{
+    OWSAssert(self.focusMessageIdOnOpen);
+
+    NSInteger row = 0;
+    for (ConversationViewItem *viewItem in self.viewItems) {
```
michaelkirk-signal: Are you sure we can't get this in constant time from the mapping?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: Not bad!

Signal/src/call/OutboundCallInitiator.swift
michaelkirk-signal: :+1:

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: We can get constant time by using the position.

charlesmchen-signal:
> - (BOOL)getGroup:(NSString * _Nullable * _Nullable)groupPtr

Done

PTAL @michaelkirk

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -698,13 +705,43 @@ - (NSIndexPath *_Nullable)indexPathOfUnreadMessagesIndicator
     return nil;
 }
 
+- (NSIndexPath *_Nullable)indexPathOfMessageOnOpen
+{
```
michaelkirk-signal: I was expecting something shorter here. We *have* the mapping, view index, and the group, so getting the index path should be a one liner:

- (nullable NSIndexPath *)indexPathForIndex:(NSUInteger)index inGroup:(NSString *)group;

But it would entail storing the viewIndex in addition to (or rather than) the reverse-index. In general I'm not sure introducing the concept of the "reverse index" is helpful. 

Instead we could store the forward index, which simplifies the method which stores it, use it here directly, and only invert it for calculating the range length.

You've already addressed my blocking perf concerns though - so this is TIOLI, I'm just trying to find ways to simplify the remaining code since we'll live with it forever.

charlesmchen-signal: > But it would entail storing the viewIndex in addition to (or rather than) the reverse-index. In general I'm not sure introducing the concept of the "reverse index" is helpful.

We use "reverse indices" because (among other uses) we use this to determine the initial load window size, before a "view index" has any meaning.  Sorry if that wasn't clear; I've elaborated the comments.

PR #3541, opened 2018-06-12
Remove conflicts between call audio and other audio activities.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/environment/OWSAudioSession.swift
+++ b/SignalMessaging/environment/OWSAudioSession.swift
@@ -66,6 +66,12 @@ public class OWSAudioSession: NSObject {
 
         startAudioActivity(audioActivity)
 
+        guard currentActivities.count == 1 else {
+            // We don't want to clobber the audio capabilities configured by (e.g.) media playback or an in-progress call
+            Logger.info("\(logTag) in \(#function) not touching audio session since another currentActivity exists.")
+            return
+        }
+
         do {
             try avAudioSession.setCategory(AVAudioSessionCategoryPlayback)
         } catch {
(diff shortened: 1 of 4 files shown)


SignalMessaging/environment/OWSAudioSession.swift
```diff
@@ -84,6 +90,12 @@ public class OWSAudioSession: NSObject {
 
         startAudioActivity(audioActivity)
 
+        guard currentActivities.count == 1 else {
```
michaelkirk-signal: I'm not sure this is right.

To simplify a bit:

We want to allow "upgrading" audio sessions:

Ambient -> Playback -> PlayAndRecord

But we never want to "downgrade" functionality.

E.g. If we are playing a video while a call comes in, we *want* to go from Playback -> PlayAndRecord

We just want to make sure the reverse doesn't happen - while on a call, when we play a video, we don't want to go from "PlayAndRecord" -> "Playback".

And a disclaimer: I'm not sure that a linear progression is nuanced enough to capture all the various transitions we want to make to our audio session configuration, but I'm pretty sure it would be better than what we have.

michaelkirk-signal: And there's theoretically a similar issue with downgrading from Playback -> Ambient, but off the top of my head, I'm not sure where we use ambient, so it might not be an issue in practice.

The reason the guard exists in ambient right now is that ambient is the "most restrictive" (e.g. respects silent switch, doesn't require recording permission). 

So if we're starting an ambient activity, and any other activity exists, that activity has either already configured the audio session to ambient which is a no-op or either "Playback" or "PlaybackAndRecord", which should suffice for any ambient audio activity.

charlesmchen-signal: Okay, let's address this while stabilizing v2.27.

Imperiopolis-Signal:

PR #3542, opened 2018-06-12
Order the search results.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
+++ b/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
@@ -214,7 +214,7 @@ class ConversationSearchViewController: UITableViewController {
         // TODO: debounce?
 
         self.uiDatabaseConnection.read { transaction in
-            self.searchResultSet = self.searcher.results(searchText: searchText, transaction: transaction)
+            self.searchResultSet = self.searcher.results(searchText: searchText, transaction: transaction, contactsManager: self.contactsManager)
         }
 
         // TODO: more perfomant way to do this?
--- a/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
+++ b/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
@@ -18,6 +18,10 @@ class ConversationSearchViewController: UITableViewController {
         return ConversationSearcher.shared
     }
 
+    private var contactsManager: OWSContactsManager {
+        return Environment.current().contactsManager
+    }
+
     enum SearchSection: Int {
         case noResults
         case conversations
@@ -207,6 +211,7 @@ class ConversationSearchViewController: UITableViewController {
     public func updateSearchResults(searchText: String) {
         guard searchText.stripped.count > 0 else {
             self.searchResultSet = SearchResultSet.empty
+            self.tableView.reloadData()
             return
         }
 
@@ -217,7 +222,7 @@ class ConversationSearchViewController: UITableViewController {
             self.searchResultSet = self.searcher.results(searchText: searchText, transaction: transaction, contactsManager: self.contactsManager)
         }
 
-        // TODO: more perfomant way to do this?
+        // TODO: more performant way to do this?
         self.tableView.reloadData()
     }
 }
--- a/SignalMessaging/utils/ConversationSearcher.swift
+++ b/SignalMessaging/utils/ConversationSearcher.swift
@@ -138,10 +138,10 @@ public class ConversationSearcher: NSObject {
 
         // Order the conversation and message results in reverse chronological order.
         // The contact results are pre-sorted by display name.
-        conversations = conversations.sorted(by: >)
-        messages = messages.sorted(by: >)
+        conversations.sort(by: >)
+        messages.sort(by: >)
         // Order "other" contact results by display name.
-        otherContacts = otherContacts.sorted()
+        otherContacts.sort()
 
         return SearchResultSet(searchText: searchText, conversations: conversations, contacts: otherContacts, messages: messages)
     }
--- a/SignalMessaging/utils/ConversationSearcher.swift
+++ b/SignalMessaging/utils/ConversationSearcher.swift
@@ -5,28 +5,61 @@
 import Foundation
 import SignalServiceKit
 
-public class ConversationSearchResult {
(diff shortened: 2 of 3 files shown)


Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -207,17 +211,18 @@ class ConversationSearchViewController: UITableViewController {
     public func updateSearchResults(searchText: String) {
         guard searchText.stripped.count > 0 else {
             self.searchResultSet = SearchResultSet.empty
+            self.tableView.reloadData()
```
charlesmchen-signal: This fixes an unrelated crash.

SignalMessaging/utils/ConversationSearcher.swift
```diff
@@ -5,28 +5,61 @@
...
+        return lhs.sortKey < rhs.sortKey
+    }
+
+    // MARK: Equatable
+
+    public static func == (lhs: ConversationSearchResult, rhs: ConversationSearchResult) -> Bool {
+        if lhs.thread.threadRecord.uniqueId != rhs.thread.threadRecord.uniqueId {
+            return false
+        }
+        return NSObject.isNullableObject(lhs.messageId as NSObject?, equalTo: rhs.messageId as NSObject?)
```
michaelkirk-signal: The type signature asserts that neither lhs nor rhs are nil, so can this method just become: 

```
return lhs.threadRecord.uniqueId == rhs.thread.threadRecord.uniqueId
```

And in that case, it would be nice to remove the inheritance of the NSObject machinery.

SignalMessaging/utils/ConversationSearcher.swift
```diff
@@ -5,28 +5,61 @@
...
     }
 }
 
-public class ContactSearchResult {
+public class ContactSearchResult: NSObject, Comparable {
```
michaelkirk-signal: Why NSObject here?

SignalMessaging/utils/ConversationSearcher.swift
```diff
@@ -75,30 +111,44 @@ public class ConversationSearcher: NSObject {
...
                 let threadViewModel = ThreadViewModel(thread: thread, transaction: transaction)
-                let searchResult = ConversationSearchResult(thread: threadViewModel, messageId: message.uniqueId, snippet: snippet)
-
+                let sortKey = message.timestamp
+                let searchResult = ConversationSearchResult(thread: threadViewModel, messageId: message.uniqueId, snippet: snippet, sortKey: sortKey)
                 messages.append(searchResult)
             } else if let signalAccount = match as? SignalAccount {
-                let searchResult = ContactSearchResult(signalAccount: signalAccount)
+                let anyFont = UIFont.systemFont(ofSize: 12)
+                let sortKey = contactsManager.attributedStringForConversationTitle(withPhoneIdentifier: signalAccount.recipientId,
```
michaelkirk-signal: How does this work? It looks like we'll sort by display order (e.g. by first name in most english settings). 

Instead we should respect the "sort by first name vs. last name" setting so it's the same order as what we see in the contact picker.

SignalMessaging/utils/ConversationSearcher.swift
```diff
@@ -75,30 +111,44 @@ public class ConversationSearcher: NSObject {
...
             }
         }
 
         // Only show contacts which were not included in an existing 1:1 conversation.
-        let otherContacts: [ContactSearchResult] = contacts.filter { !existingConversationRecipientIds.contains($0.recipientId) }
+        var otherContacts: [ContactSearchResult] = contacts.filter { !existingConversationRecipientIds.contains($0.recipientId) }
+
+        // Order the conversation and message results in reverse chronological order.
+        // The contact results are pre-sorted by display name.
+        conversations = conversations.sorted().reversed()
```
michaelkirk-signal: Instead of enumerating the list twice, let's do `sorted(by: >)`

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
michaelkirk-signal: nice 👍

SignalMessaging/utils/ConversationSearcher.swift
michaelkirk-signal: I misread this - I realize we're checking the thread **and** the message, if any.

In that case:

```
if lhs.thread.threadRecord.uniqueId != rhs.thread.threadRecord.uniqueId {
     return false
}
return lhs.messageId == rhs.messageId
```

or

```
return lhs.thread.threadRecord.uniqueId == rhs.thread.threadRecord.uniqueId &&
lhs.messageId == rhs.messageId
```

(

charlesmchen-signal: Done.

SignalMessaging/utils/ConversationSearcher.swift
charlesmchen-signal: Good point, done.

SignalMessaging/utils/ConversationSearcher.swift
charlesmchen-signal: Done.

charlesmchen-signal:
PTAL @michaelkirk

SignalMessaging/utils/ConversationSearcher.swift
```diff
@@ -75,30 +109,39 @@ public class ConversationSearcher: NSObject {
...
             }
         }
 
         // Only show contacts which were not included in an existing 1:1 conversation.
-        let otherContacts: [ContactSearchResult] = contacts.filter { !existingConversationRecipientIds.contains($0.recipientId) }
+        var otherContacts: [ContactSearchResult] = contacts.filter { !existingConversationRecipientIds.contains($0.recipientId) }
+
+        // Order the conversation and message results in reverse chronological order.
+        // The contact results are pre-sorted by display name.
+        conversations = conversations.sorted(by: >)
```
michaelkirk-signal: nit: since we're just overwriting the vars, let's use the in-place sort.

e.g. 
```
conversations.sort(by: >)
messages.sort(by: >)
otherContacts.sort()
```

charlesmchen-signal: Done.

PR #3543, opened 2018-06-12
Mkirk/fts sin country code

michaelkirk-signal:
PTAL @charlesmchen 

It's kind of a large diff for the actual change in behavior.

The first three commits fix up the tests.

The new functionality is in 
https://github.com/signalapp/Signal-iOS/commit/f57a5dbc77a9e1119f69f589ab975ece3c5c0b49

And the last commit is just a code cleanup.

If you'd prefer, I can make separate PR's. LMK


Signal/test/util/SearcherTest.swift
```diff
@@ -222,18 +210,16 @@ class ConversationSearcherTest: XCTestCase {
         XCTAssertEqual([bookClubThread, snackClubThread, aliceThread], threads)
     }
 
-    // TODO
-    func pending_testSearchContactByNumber() {
-        var resultSet: SearchResultSet = .empty
-
+    func testSearchContactByNumberWithoutCountryCode() {
+        var threads: [ThreadViewModel] = []
```
michaelkirk-signal: This test represents the new feature work in this PR

Signal/test/util/SearcherTest.swift
```diff
@@ -244,16 +230,16 @@ class ConversationSearcherTest: XCTestCase {
         XCTAssertEqual([bookClubThread, snackClubThread, aliceThread], threads)
 
         threads = searchConversations(searchText: "Bob")
-        XCTAssertEqual(2, threads.count)
-        XCTAssertEqual([bookClubThread, bobThread], threads)
+        XCTAssertEqual(1, threads.count)
```
michaelkirk-signal: So this is just a test change - not a feature change. I previously failed to update this test once we started intentionally excluding contact threads which had never been messaged from the "Conversations" search results.

SignalServiceKit/src/Contacts/PhoneNumber.m
```diff
@@ -372,25 +372,23 @@ - (NSNumber *)getCountryCode {
     return self.phoneNumber.countryCode;
 }
 
-- (BOOL)isValid {
-    return [[PhoneNumberUtil sharedThreadLocal].nbPhoneNumberUtil isValidNumber:self.phoneNumber];
-}
-
-- (NSString *)localizedDescriptionForUser {
```
michaelkirk-signal: This diff is confusing because I removed some dead code localizedDescriptionForUser and resolvesInternationallyTo

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
```diff
@@ -47,27 +62,31 @@ public class FullTextSearchFinder: NSObject {
...
+        let normalizedChars = String(String.UnicodeScalarView(nonformattingScalars))
 
-        normalized = String(String.UnicodeScalarView(nonformattingScalars))
+        let digitsOnlyScalars = normalized.unicodeScalars.lazy.filter {
+            CharacterSet.decimalDigits.contains($0)
+        }
+        let normalizedDigits = String(String.UnicodeScalarView(digitsOnlyScalars))
 
-        return normalized
+        if normalizedDigits.count > 0 {
```
charlesmchen-signal: So, a query like: "(321) 321-4321" would match a message body like "321 salesmen sold 321 peanuts to 4321 elephants"?  That sounds fine.

michaelkirk-signal: No, note there are now two different normalization methods, one for the query text the user types into the search bar and one which the yap extension uses when building the search index.

The digit normalization only happens on the query side.

The indexing side makes sure that both the e164 and the national number are stored in the index for any recipient id.

So the inverse is true, a query of "321 salesmen sold 321 peanuts to 4321 elephants" would match a recipient with id "+13213214321"

PR #3546, opened 2018-06-13
Update search results.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
+++ b/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
@@ -13,7 +13,7 @@ class ConversationSearchViewController: UITableViewController {
             SwiftAssertIsOnMainThread(#function)
 
             // Use a slight delay to debounce updates.
-            refreshSearchResults(delay: 0.25)
+            refreshSearchResults()
         }
     }
 
@@ -64,7 +64,7 @@ class ConversationSearchViewController: UITableViewController {
     @objc internal func yapDatabaseModified(notification: NSNotification) {
         SwiftAssertIsOnMainThread(#function)
 
-        refreshSearchResults(delay: 1.0)
+        refreshSearchResults()
     }
 
     // MARK: UITableViewDelegate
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -41,6 +50,17 @@ class ConversationSearchViewController: UITableViewController {
...
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(yapDatabaseModified),
+                                               name: NSNotification.Name.YapDatabaseModified,
+                                               object: OWSPrimaryStorage.shared().dbNotificationObject)
+    }
+
+    @objc internal func yapDatabaseModified(notification: NSNotification) {
+        SwiftAssertIsOnMainThread(#function)
+
+        refreshSearchResults(delay: 1.0)
```
charlesmchen-signal: Notice that we use a longer delay for these events, since they can be frequent.

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -7,10 +7,19 @@ import Foundation
 @objc
 class ConversationSearchViewController: UITableViewController {
 
+    @objc
+    public var searchText = "" {
+        didSet {
+            SwiftAssertIsOnMainThread(#function)
+
+            // Use a slight delay to debounce updates.
+            refreshSearchResults(delay: 0.25)
```
charlesmchen-signal: We use a shorted delay when debouncing updates in response to changes to the search query.

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
michaelkirk-signal: This could interfere with snappiness from the user input. e.g. if a dbModified happens just before the user enters a character they won't see results updated for 1s, instead of 0.25 seconds. 

We could make it more complicated to address this, but I don't see much benefit from observing dbModified notifications in this context.

charlesmchen-signal: Okay, removing.

charlesmchen-signal: Actually, I'm going to make the case for leaving this in - let's discuss at iOS standup.

charlesmchen-signal: Okay, I've re-enable this but both "user changes" and "db modifications" have a much shorter delay and we skip the delay if we're not yet showing any results.

charlesmchen-signal:
> Should we show any feedback? progress indicator, etc.?
> Or does it feel snappy enough without it?

In the "worst case" (slow device, tons of results) searching can take a moment (iPhone 6, thousands of results = 0.3-0.5 seconds).  

I suggest we skip a progress indicator.  Although the results do "stutter" in the worst case, IMO "blinking" in a progress indicator for that short a period will just be jarring.

> I do think we should make sure yap notifications don't delay the users search feedback (possibly by just removing that functionality, but if you want to implement something that handles it separately thats OK too).

Agreed, see above.

charlesmchen-signal:
PTAL @michaelkirk

PR #3547, opened 2018-06-13
Avoid double-scrolling.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -324,7 +324,7 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
     [searchResultsController.view autoPinWidthToSuperview];
     [searchResultsController.view autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:searchBar];
     [searchResultsController.view autoPinEdge:ALEdgeBottom toEdge:ALEdgeBottom ofView:self.tableView];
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -324,7 +324,7 @@ - (void)viewDidLoad
     [searchResultsController.view autoPinWidthToSuperview];
     [searchResultsController.view autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:searchBar];
     [searchResultsController.view autoPinEdge:ALEdgeBottom toEdge:ALEdgeBottom ofView:self.tableView];
-    searchResultsController.view.hidden = self;
+    searchResultsController.view.hidden = YES;
```
charlesmchen-signal: Fixed a typo.

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -913,6 +913,13 @@ - (void)updateSearchResultsVisibility
     [self.searchResultsController updateSearchResultsWithSearchText:searchText];
     BOOL isSearching = searchText.length > 0;
     self.searchResultsController.view.hidden = !isSearching;
+
+    if (isSearching) {
```
charlesmchen-signal: We show the search results table view "on top of" the normal home view table view.  Don't let the user scroll the "background" table view (by swiping on the search field) while the "foreground" table view is visible.

PR #3550, opened 2018-06-13
Charlesmchen/auto size contact cells

charlesmchen-signal:
The only place I haven't fully converted is `ContactPicker`.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/AppSettings/BlockListViewController.m
+++ b/Signal/src/ViewControllers/AppSettings/BlockListViewController.m
@@ -44,7 +44,6 @@ NS_ASSUME_NONNULL_BEGIN
     [_tableViewController.view autoPinToTopLayoutGuideOfViewController:self withInset:0];
     [_tableViewController.view autoPinEdgeToSuperviewEdge:ALEdgeBottom];
     self.tableViewController.tableView.rowHeight = UITableViewAutomaticDimension;
-    //    self.tableView.estimatedRowHeight = 60;
 
     [self updateTableContents];
 }
--- a/Signal/src/ViewControllers/AppSettings/BlockListViewController.m
+++ b/Signal/src/ViewControllers/AppSettings/BlockListViewController.m
@@ -43,6 +43,8 @@ NS_ASSUME_NONNULL_BEGIN
     [_tableViewController.view autoPinWidthToSuperview];
     [_tableViewController.view autoPinToTopLayoutGuideOfViewController:self withInset:0];
     [_tableViewController.view autoPinEdgeToSuperviewEdge:ALEdgeBottom];
+    self.tableViewController.tableView.rowHeight = UITableViewAutomaticDimension;
+    //    self.tableView.estimatedRowHeight = 60;
 
     [self updateTableContents];
 }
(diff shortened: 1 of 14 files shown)


Signal/src/ViewControllers/AppSettings/BlockListViewController.m
```diff
@@ -78,18 +79,22 @@ - (void)updateTableContents
...
-            if (signalAccount) {
-                [cell configureWithSignalAccount:signalAccount contactsManager:helper.contactsManager];
-            } else {
-                [cell configureWithRecipientId:phoneNumber contactsManager:helper.contactsManager];
-            }
-
-            return cell;
-        }
-                                      customRowHeight:[ContactTableViewCell rowHeight]
+        [blocklistSection addItem:[OWSTableItem
```
michaelkirk-signal: Is your clang format feeling OK?

I never really know what to expect, so this could be correct... just wanted to double check.

Signal/src/ViewControllers/AppSettings/OWSLinkedDevicesTableViewController.m
```diff
@@ -44,7 +44,6 @@ - (void)viewDidLoad
 
     self.isExpectingMoreDevices = NO;
     self.tableView.rowHeight = UITableViewAutomaticDimension;
-    self.tableView.estimatedRowHeight = 70;
```
michaelkirk-signal: Why are you removing the estimatedRowHeights here and elsewhere?

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -450,7 +450,7 @@ - (void)ensureThumbnail
     }
 
     if (result == nil) {
-        OWSFail(@"%@ Unable to build thumnail for attachmentId: %@", self.logTag, self.uniqueId);
```
michaelkirk-signal: Why this change? Any clue what got you into this state?

charlesmchen-signal: I've been testing how we handle strange media formats. There are definitely some container/codec combinations where iOS can't extract a thumbnail.  I don't think we'll be able to detect these beforehand (and treat them as a generic attachment).

Signal/src/ViewControllers/AppSettings/OWSLinkedDevicesTableViewController.m
charlesmchen-signal: We could go either way here.

`estimatedRowHeight` improves perf for tables with many cells.  But I'm not sure that we're supplying a good estimate since our cell heights can vary wildly based on dynamic text size.  The docs don't make clear to me if an inaccurate estimate is better or worse for perf than no estimate at all.  WDYT?

Signal/src/ViewControllers/AppSettings/BlockListViewController.m
charlesmchen-signal: I have no idea how to determine that.  It's a mysterious beast.

charlesmchen-signal:
PTAL @michaelkirk

Signal/src/ViewControllers/AppSettings/OWSLinkedDevicesTableViewController.m
michaelkirk-signal: I would leave it as is. It's true we don't know the inner workings of estimatedRowHeight, but based on the documentation, we should be providing a reasonable guess, which we were.

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
michaelkirk-signal: I've hit this assert a few times and it was always a result of some error in our attachment processing logic, so I would have kept it as a good sentinel. 

> I don't think we'll be able to detect these beforehand

We don't have to detect them all - this is just a debug assert. But if you think we'll be hitting this branch under normal usage in such a way that it's more distracting than the sentinel value it provides, I guess we can remove it.

Signal/src/ViewControllers/AppSettings/OWSLinkedDevicesTableViewController.m
charlesmchen-signal: I've already put it back in another PR.  You were right!

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
charlesmchen-signal: Okay, I'll put it back.

PR #3551, opened 2018-06-13
Fix assert around failing sync message sends.

charlesmchen-signal:
PTAL @michaelkirk


michaelkirk-signal:
Ha, I just opened https://github.com/signalapp/Signal-iOS/pull/3555 to fix the same thing. I should review quicker and save myself the trouble!

(LGTM)

PR #3552, opened 2018-06-13
Index oversize text for search.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Storage/FullTextSearchFinder.swift
+++ b/SignalServiceKit/src/Storage/FullTextSearchFinder.swift
@@ -207,10 +207,7 @@ public class FullTextSearchFinder: NSObject {
         guard message.hasAttachments() else {
             return nil
         }
-        guard let dbConnection = dbConnection else {
-            owsFail("Could not load attachment for search indexing.")
-            return nil
-        }
+        let dbConnection = OWSPrimaryStorage.shared().dbReadConnection
         var oversizeText: String?
         dbConnection.read({ (transaction) in
             guard let attachment = message.attachment(with: transaction) else {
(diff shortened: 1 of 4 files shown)


SignalServiceKit/src/Storage/FullTextSearchFinder.swift
```diff
@@ -170,6 +209,12 @@ public class FullTextSearchFinder: NSObject {
     }
 
     private class var dbExtensionConfig: YapDatabaseFullTextSearch {
+        SwiftAssertIsOnMainThread(#function)
+
+        if dbConnection == nil {
+            dbConnection = OWSPrimaryStorage.shared().newDatabaseConnection()
```
charlesmchen-signal: You might not like this; I'm trying to ensure that the db connection gets initialized exactly once, on the main thread, and at a well-defined time.

michaelkirk-signal: This is indeed a strange place for a side effect like this. Why not use the global read connection?

charlesmchen-signal: Removed.  I thought you would prefer this way.

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
```diff
@@ -128,9 +128,48 @@ public class FullTextSearchFinder: NSObject {
...
+                return
+            }
+            guard let attachmentStream = attachment as? TSAttachmentStream else {
+                return
+            }
+            guard attachmentStream.isOversizeText() else {
+                return
+            }
+            Logger.verbose("attachmentStream: \(attachmentStream.contentType)")
+            Logger.flush()
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

charlesmchen-signal:
PTAL @michaelkirk

PR #3553, opened 2018-06-14
Update the README.

jlund-signal:
PTAL @charlesmchen-signal and @michaelkirk-signal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,21 +2,16 @@
 
 Signal is a messaging app for simple private communication with friends.
 
-[![Available on the AppStore](http://cl.ly/WouG/Download_on_the_App_Store_Badge_US-UK_135x40.svg)](https://itunes.apple.com/app/id874139669)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,21 +2,16 @@
 
 Signal is a messaging app for simple private communication with friends.
 
-[![Available on the AppStore](http://cl.ly/WouG/Download_on_the_App_Store_Badge_US-UK_135x40.svg)](https://itunes.apple.com/app/id874139669)
+[![Available on the App Store](http://cl.ly/WouG/Download_on_the_App_Store_Badge_US-UK_135x40.svg)](https://itunes.apple.com/us/app/signal-private-messenger/id874139669?mt=8)
```
charlesmchen-signal: This is a link to the US app store. If we can, we might want to verify that the link works for users who are using app store in other countries.

PR #3558, opened 2018-06-15
Charlesmchen/unify cell and avatar sizes

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/AppSettings/BlockListViewController.m
+++ b/Signal/src/ViewControllers/AppSettings/BlockListViewController.m
@@ -44,6 +44,7 @@ NS_ASSUME_NONNULL_BEGIN
     [_tableViewController.view autoPinToTopLayoutGuideOfViewController:self withInset:0];
     [_tableViewController.view autoPinEdgeToSuperviewEdge:ALEdgeBottom];
     self.tableViewController.tableView.rowHeight = UITableViewAutomaticDimension;
+    self.tableViewController.tableView.estimatedRowHeight = 60;
 
     [self updateTableContents];
 }
--- a/Signal/src/ViewControllers/HomeView/HomeViewCell.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewCell.m
@@ -59,8 +59,6 @@ NS_ASSUME_NONNULL_BEGIN
 {
     OWSAssert(!self.avatarView);
 
-    [self setTranslatesAutoresizingMaskIntoConstraints:NO];
-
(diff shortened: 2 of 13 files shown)


Signal/src/ViewControllers/AppSettings/BlockListViewController.m
```diff
@@ -44,6 +44,7 @@ - (void)loadView
     [_tableViewController.view autoPinToTopLayoutGuideOfViewController:self withInset:0];
     [_tableViewController.view autoPinEdgeToSuperviewEdge:ALEdgeBottom];
     self.tableViewController.tableView.rowHeight = UITableViewAutomaticDimension;
+    self.tableViewController.tableView.estimatedRowHeight = 60;
```
charlesmchen-signal: This value turns out to be required on earlier versions of iOS, e.g. 9.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
```diff
@@ -59,13 +59,6 @@ - (void)commontInit
...
     OWSAssert(!self.avatarView);
 
-    const CGFloat kMinVMargin = 5;
-
-    [self setTranslatesAutoresizingMaskIntoConstraints:NO];
-    self.layoutMargins = UIEdgeInsetsMake(0, self.cellHMargin, 0, self.cellHMargin);
-    self.contentView.layoutMargins = UIEdgeInsetsZero;
-    self.contentView.preservesSuperviewLayoutMargins = YES;
-
     self.backgroundColor = [UIColor whiteColor];
```
charlesmchen-signal: The intention is that as we adopt `UITableViewAutomaticDimension` / auto-sizing of cells throughout our views, to converge upon "cutting with the grain" of how iOS auto layout and `UITableViewCells` are supposed to work.

* Don't `preservesSuperviewLayoutMargins`.
* Don't meddle with cell or `cell.contentView` default layout margins.
* Pin to margins, not edges.

SignalMessaging/Views/ContactTableViewCell.m
```diff
@@ -18,7 +18,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-const NSUInteger kContactTableViewCellAvatarSize = 40;
+const NSUInteger kContactTableViewCellAvatarSize = 48;
```
charlesmchen-signal: Unify the avatar heights. This will unify the cell heights except in cases where dynamic text sizes are very large.

PR #3559, opened 2018-06-15
Shared ui db connection v. auto load more.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -1065,6 +1065,7 @@ typedef enum : NSUInteger {
     [self startReadTimer];
     [self updateNavigationBarSubtitleLabel];
     [self updateBackButtonUnreadCount];
+    [self autoLoadMoreIfNecessary];
 
     switch (self.actionOnOpen) {
         case ConversationViewActionNone:
@@ -1081,7 +1082,6 @@ typedef enum : NSUInteger {
     }
 
     self.actionOnOpen = ConversationViewActionNone;
-
 
     self.isViewCompletelyAppeared = YES;
     self.viewHasEverAppeared = YES;
@@ -1557,6 +1557,10 @@ typedef enum : NSUInteger {
     //
     // Otherwise, tapping on "load more messages" autoscrolls you downward which is completely wrong.
     if (hasEarlierUnseenMessages) {
+        // Ensure view items are updated before trying to scroll to the
+        // unread indicator.
+        [[OWSPrimaryStorage sharedManager] updateUIDatabaseConnectionToLatest];
+
         [self scrollToUnreadIndicatorAnimated];
     }
 }


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1065,6 +1065,7 @@ - (void)viewDidAppear:(BOOL)animated
     [self startReadTimer];
     [self updateNavigationBarSubtitleLabel];
     [self updateBackButtonUnreadCount];
+    [self autoLoadMoreIfNecessary];
```
charlesmchen-signal: This fix is already in v2.27, but we'll want it in v2.26 too.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1557,6 +1557,10 @@ - (void)loadAnotherPageOfMessages
     //
     // Otherwise, tapping on "load more messages" autoscrolls you downward which is completely wrong.
     if (hasEarlierUnseenMessages) {
+        // Ensure view items are updated before trying to scroll to the
+        // unread indicator.
+        [[OWSPrimaryStorage sharedManager] updateUIDatabaseConnectionToLatest];
```
charlesmchen-signal: This should be pretty rare, but I realized that we missed an edge case yesterday.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: For perf we're deliberately only loading a very small window - just enough to fill the screen (though our approach is pretty naive, we could probably load even less if we accounted for device size and message size).

What's wrong with waiting until we scroll to load more?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: If you are going to add a comment, try to have it explain "why", not "what".  The code is already as clear as the comment, and after reading I still don't understand why we want to update the uiDB outside of the normal yapDBModified cycle.

The one scenario I know of when we need to update the uiDB outside of the normal yapDBModified cycle is immediately after the `ensureDynamicInteractions` is called. We need it in that one place because we are performing a synchronous write transaction on the main thread, and want the side effects of that transaction (e.g. a newly created unread indicator) to be available to the rendering process. Waiting for yapDBModified would be too late because it happens async.

As far as I can tell, we're opening no write transaction here, so none of that applies. To be clear, I don't think this will do any significant harm, but I want to make sure we're not superstitiously sprinkling this fix around where it doesn't apply.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: > What's wrong with waiting until we scroll to load more?

There are common scenarios (e.g. more than a load page of unread messages) where we'll load messages up to the unread indicator and then auto-scroll so that the unread-indicator is near the top of the screen.  In these scenarios, the "load more messages" indicator is often visible above the unread indicator.  The rule is: auto-load more messages if we're "near" the "load more messages" indicator so that users don't see it.  That's a courtesy to the user so that the conversation feels infinite and the user doesn't notice loading.  There's another reason too: it avoids a nasty issue wherein it's hard to preserve scroll state if auto-loading occurs while the user is scrolling.

Anyway, performing `autoLoadMoreIfNecessary` here shouldn't do anything most of the time, since most of the time we're not "near enough" to the "load more messages" indicator - and when it does load more, it's for a good reason.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: > We need it in that one place because we are performing a synchronous write transaction on 

A pretty similar scenario is happening here.

> loadNMoreMessages calls resetMappings which calls ensureDynamicInteractions,
> which may move the unread indicator, and for scrollToUnreadIndicatorAnimated
> to work properly, the view items need to be updated to reflect that change.

I've updated the comment to reflect that.

PR #3562, opened 2018-06-18
Charlesmchen/deregistration

charlesmchen-signal:
PTAL @michaelkirk 

* Detects "de-registration" service responses.
* Clears (spurious)) "de-registration" state if we receive a valid response/websocket connection to the service.
* Show "de-registration" alert UI with "re-registration" option in home view.
* Add "re-registration" UI flow from home view.
* Modify "registration" views to resume "re-registration".

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -25,6 +25,7 @@
 #import <SignalServiceKit/OWSBlockingManager.h>
 #import <SignalServiceKit/OWSMessageSender.h>
 #import <SignalServiceKit/OWSMessageUtils.h>
+#import <SignalServiceKit/TSAccountManager.h>
 #import <SignalServiceKit/TSOutgoingMessage.h>
 #import <SignalServiceKit/Threading.h>
 #import <YapDatabase/YapDatabase.h>
@@ -159,6 +160,10 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
                                              selector:@selector(yapDatabaseModifiedExternally:)
                                                  name:YapDatabaseModifiedExternallyNotification
                                                object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(deregistrationStateDidChange:)
+                                                 name:DeregistrationStateDidChangeNotification
+                                               object:nil];
 }
 
 - (void)dealloc
@@ -178,6 +183,13 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
 }
 
 - (void)signalAccountsDidChange:(id)notification
+{
+    OWSAssertIsOnMainThread();
+
+    [self reloadTableViewData];
+}
+
+- (void)deregistrationStateDidChange:(id)notification
 {
     OWSAssertIsOnMainThread();
 
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -211,6 +211,13 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
         [SignalApp.sharedApp setHomeViewController:self];
     }
 
+    UIStackView *reminderStackView = [UIStackView new];
+    reminderStackView.axis = UILayoutConstraintAxisVertical;
+    reminderStackView.spacing = 0;
+    [self.view addSubview:reminderStackView];
+    [reminderStackView autoPinWidthToSuperview];
+    [reminderStackView autoPinToTopLayoutGuideOfViewController:self withInset:0];
+
     __weak HomeViewController *weakSelf = self;
     ReminderView *deregisteredView =
         [ReminderView nagWithText:NSLocalizedString(@"DEREGISTRATION_WARNING",
@@ -222,18 +229,14 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
                             }
                             [RegistrationUtils showReregistrationUIFromViewController:strongSelf];
                         }];
-    [self.view addSubview:deregisteredView];
-    [deregisteredView autoPinWidthToSuperview];
-    [deregisteredView autoPinToTopLayoutGuideOfViewController:self withInset:0];
+    [reminderStackView addArrangedSubview:deregisteredView];
     self.hideDeregisteredViewConstraint = [deregisteredView autoSetDimension:ALDimensionHeight toSize:0];
     self.hideDeregisteredViewConstraint.priority = UILayoutPriorityRequired;
 
     ReminderView *archiveReminderView =
         [ReminderView explanationWithText:NSLocalizedString(@"INBOX_VIEW_ARCHIVE_MODE_REMINDER",
                                               @"Label reminding the user that they are in archive mode.")];
-    [self.view addSubview:archiveReminderView];
-    [archiveReminderView autoPinWidthToSuperview];
-    [archiveReminderView autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:deregisteredView];
+    [reminderStackView addArrangedSubview:archiveReminderView];
     self.hideArchiveReminderViewConstraint = [archiveReminderView autoSetDimension:ALDimensionHeight toSize:0];
     self.hideArchiveReminderViewConstraint.priority = UILayoutPriorityRequired;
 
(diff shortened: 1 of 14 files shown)


Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -1376,6 +1410,80 @@ - (void)updateEmptyBoxText
...
+    DDLogInfo(@"%@ reregisterWithSamePhoneNumber.", self.logTag);
+
+    if (![[TSAccountManager sharedInstance] resetForReregistration]) {
+        OWSFail(@"%@ could not reset for re-registration.", self.logTag);
+        return;
+    }
+
+    [[Environment current].preferences unsetRecordedAPNSTokens];
+
+    [ModalActivityIndicatorViewController
```
charlesmchen-signal: Arguably we could simplify this and just show the `RegistrationViewController` (which will be in re-registration mode) but this UI flow matches Signal Android.

Signal/src/views/ReminderView.swift
```diff
@@ -91,7 +91,7 @@ class ReminderView: UIView {
         // Label
         switch (mode) {
         case .nag:
-            label.font = UIFont.ows_regularFont(withSize: 14)
+            label.font = UIFont.ows_dynamicTypeFootnote
```
charlesmchen-signal: Dynamic type.  Seems to work - will add to test plan.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -659,7 +665,7 @@
 /* table cell label in conversation settings */
 "DISAPPEARING_MESSAGES" = "Disappearing Messages";
 
-/* Info Message when added to {{group name}} which has enabled message expiration after {{time amount}}, see the *_TIME_AMOUNT strings for context. */
+/* Info Message when added to a group which has enabled disappearing messages. Embeds {{time amount}} before messages disappear, see the *_TIME_AMOUNT strings for context. */
```
charlesmchen-signal: Note string comment change.

SignalServiceKit/src/Account/TSAccountManager.h
```diff
@@ -116,6 +114,22 @@ extern NSString *const kNSNotificationName_LocalNumberDidChange;
...
+- (BOOL)isDeregistered;
+- (void)setIsDeregistered:(BOOL)isDeregistered;
+
+#pragma mark - Re-registration
+
+// Re-registration is the process of re-registering _with the same phone number_.
+
+// Returns YES on success.
+- (BOOL)resetForReregistration;
+- (NSString *)reregisterationPhoneNumber;
```
charlesmchen-signal: This isn't the same as `phoneNumberAwaitingVerification`.  'phoneNumberAwaitingVerification' isn't persisted, for one thing.

charlesmchen-signal: Also, during re-registration (as opposed to initial registration) users are "locked" into their old phone number to avoid edge cases.

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -197,12 +211,29 @@ - (void)loadView
...
+                            HomeViewController *strongSelf = weakSelf;
+                            if (!strongSelf) {
+                                return;
+                            }
+                            [RegistrationUtils showReregistrationUIFromViewController:strongSelf];
+                        }];
+    [self.view addSubview:deregisteredView];
+    [deregisteredView autoPinWidthToSuperview];
+    [deregisteredView autoPinToTopLayoutGuideOfViewController:self withInset:0];
+    self.hideDeregisteredViewConstraint = [deregisteredView autoSetDimension:ALDimensionHeight toSize:0];
```
michaelkirk-signal: I understand you're building on the existing pattern, but it's kind of gnarly. Could we instead put all the reminder views into a (potentially empty) reminderViewStackView and remove the `hide*ViewConstraint`s?

It's not a big deal and this should work as is, but I think the alternative would be cleaner. TIOLI

Signal/src/ViewControllers/Registration/CodeVerificationViewController.m
```diff
@@ -112,16 +112,25 @@ - (void)createViews
...
-    backButton.titleLabel.font = [UIFont ows_mediumFontWithSize:14.f];
-    [header addSubview:backButton];
-    [backButton autoPinLeadingToSuperviewMarginWithInset:10.f];
-    [backButton autoAlignAxis:ALAxisHorizontal toSameAxisOfView:titleLabel];
-    [backButton addTarget:self action:@selector(backButtonPressed:) forControlEvents:UIControlEventTouchUpInside];
+    // This view is used in more than one context.
+    //
+    // * Usually, it is pushed atop RegistrationViewController in which
+    //   case we want a "back" button.
+    // * It can also be used to re-register from the app's "de-registration"
```
michaelkirk-signal: >  from the app's "de-registration"

Do you mean "re-registration" ?

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -647,6 +647,12 @@
 /* Title of the alert before redirecting to GitHub Issues. */
 "DEBUG_LOG_GITHUB_ISSUE_ALERT_TITLE" = "GitHub Redirection";
 
+/* Label for button that lets users re-register using the same phone number. */
+"DEREGISTRATION_REREGISTER_WITH_SAME_PHONE_NUMBER" = "Re-register this phone number";
+
+/* Label warning the user that they have been de-registered. */
+"DEREGISTRATION_WARNING" = "Device no longer registered! This is likely because you registered your phone number with Signal on a different device. Tap to re-register.";
```
michaelkirk-signal: PTAL @jlund

Signal/translations/en.lproj/Localizable.strings
michaelkirk-signal: Thanks!

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1309,6 +1315,9 @@
 /* No comment provided by engineer. */
 "NETWORK_STATUS_CONNECTING" = "Connecting";
 
+/* Error indicating that this device is no longer registered. */
+"NETWORK_STATUS_DEREGISTERED" = "No Longer Registered";
```
michaelkirk-signal: In the common case, this happens because the user has registered on their other device. So maybe it would be good to clarify that it's the *device* that's no longer registered, not the user.

e.g. maybe "device deactivated" or  "device no longer active"

/cc @jlund

Signal/src/ViewControllers/Registration/CodeVerificationViewController.m
charlesmchen-signal: Sorry, these changes introduce two concepts:

* De-registration - when the app receives 403 from the service, it becomes "de-registered" and warns the user of that.
* Re-registration - the process of registering again to recover from de-registration.  Unlike registering for the first time, re-registering requires users to use the same phone number.

Signal/translations/en.lproj/Localizable.strings
charlesmchen-signal: This is the exact copy from Android which I believe is our design.

Signal/translations/en.lproj/Localizable.strings
charlesmchen-signal: Great point.

Signal/src/ViewControllers/HomeView/HomeViewController.m
charlesmchen-signal: Done.

charlesmchen-signal: That is, I kept the constraints but used a stack view.  If you'd rather go farther and rework this to not use the constraints, we could do that but I'm not sure it'll be any simpler.

Signal/src/ViewControllers/Registration/CodeVerificationViewController.m
michaelkirk-signal: Right - so if we are in the CodeVerificationViewController, aren't we in the "re-registration" flow?

charlesmchen-signal: Sorry for the lack of clarity.  Users _start_ the re-registration process from two places: app settings and home view's de-registration warning UIs.  When shown from these places, we skip the "registration view" since we already know the phone number and skip to the "code verification" view.  Since there's no "registration view" in these two UI flows, there's no back button on the "code verification" view.

PR #3563, opened 2018-06-19
Charlesmchen/lazy contact

charlesmchen-signal:
PTAL @michaelkirk 

* Modify `Contact` to not hold onto `CNContact`.
* Modify `ContactsManager` to fetch individual `CNContacts` synchronously, using a cache.
* Lazy-load contact avatar data and images. 
* Use `LRUCache` for avatar images.

### DIFF
--- a/Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
+++ b/Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
@@ -60,13 +60,11 @@ class AddContactShareToExistingContactViewController: ContactsPicker, ContactsPi
 
         let contactsManager = Environment.current().contactsManager
         guard let oldCNContact = contactsManager?.cnContact(withId: oldContact.cnContactId) else {
-            // TODO: Alert?
-            Logger.warn("\(logTag) could not load old CNContact.")
+            owsFail("\(logTag) could not load old CNContact.")
             return
         }
         guard let newCNContact = OWSContacts.systemContact(for: self.contactShare.dbRecord, imageData: self.contactShare.avatarImageData) else {
-            // TODO: Alert?
-            Logger.warn("\(logTag) could not load new CNContact.")
+            owsFail("\(logTag) could not load new CNContact.")
             return
         }
         merge(oldCNContact: oldCNContact, newCNContact: newCNContact)
--- a/Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
+++ b/Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
@@ -55,24 +55,21 @@ class AddContactShareToExistingContactViewController: ContactsPicker, ContactsPi
         navigationController.popViewController(animated: true)
     }
 
-    func contactsPicker(_: ContactsPicker, didSelectContact newContact: Contact) {
+    func contactsPicker(_: ContactsPicker, didSelectContact oldContact: Contact) {
         Logger.debug("\(self.logTag) in \(#function)")
 
         let contactsManager = Environment.current().contactsManager
-        contactsManager?.cnContact(withId: self.contactShare.cnContactId,
-                                   success: { [weak self] oldCNContact in
-                                    contactsManager?.cnContact(withId: newContact.cnContactId,
-                                                               success: { newCNContact in
-                                                                guard let strongSelf = weakSelf else {
-                                                                    return
-                                                                }
-                                                                strongSelf.merge(oldCNContact: oldCNContact, newCNContact: newCNContact)
-                                    }, failure: {
-                                        // TODO: Alert?
-                                    })
-        }, failure: {
+        guard let oldCNContact = contactsManager?.cnContact(withId: oldContact.cnContactId) else {
             // TODO: Alert?
-        })
+            Logger.warn("\(logTag) could not load old CNContact.")
+            return
+        }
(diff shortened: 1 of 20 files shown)


Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
```diff
@@ -55,17 +55,29 @@ class AddContactShareToExistingContactViewController: ContactsPicker, ContactsPi
...
     }
 
-    func contactsPicker(_: ContactsPicker, didSelectContact contact: Contact) {
+    func contactsPicker(_: ContactsPicker, didSelectContact oldContact: Contact) {
         Logger.debug("\(self.logTag) in \(#function)")
 
-        guard let mergedContact: CNContact = self.contactShare.cnContact(mergedWithExistingContact: contact) else {
-            owsFail("\(logTag) in \(#function) mergedContact was unexpectedly nil")
+        let contactsManager = Environment.current().contactsManager
+        guard let oldCNContact = contactsManager?.cnContact(withId: oldContact.cnContactId) else {
```
charlesmchen-signal: Everywhere we are using `CNContact` becomes slightly more laborious, but this change dramatically reduces memory usages for users with large numbers of contacts.

SignalMessaging/contacts/OWSContactsManager.m
```diff
@@ -36,6 +36,7 @@ @interface OWSContactsManager () <SystemContactsFetcherDelegate>
 @property (nonatomic, readonly) SystemContactsFetcher *systemContactsFetcher;
 @property (nonatomic, readonly) YapDatabaseConnection *dbReadConnection;
 @property (nonatomic, readonly) YapDatabaseConnection *dbWriteConnection;
+@property (nonatomic, readonly) AnyLRUCache *cnContactCache;
```
charlesmchen-signal: We use `AnyLRUCache` instead of `NSCache` because it is more aggressive about enforcing cache size, which turns out to matter.

michaelkirk-signal: I haven't seen any evidence that NSCache is problematic here. Previously we just were **never** evicting the CNContacts because they were strongly referenced by our list of `OWSContactManager#allContacts: [Contact]`.

So unless you know something I don't, I'd say we should stick with NSCache here unless we have a strong reason not to. NSCache gives us more flexibility to have a larger cache size when possible, while also having a smaller cache size when the system is under pressure.

Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
michaelkirk-signal: We've also recently added some fields in:

```
    public static let allowedContactKeys: [CNKeyDescriptor] = [
        CNContactFormatter.descriptorForRequiredKeys(for: .fullName),
        CNContactThumbnailImageDataKey as CNKeyDescriptor,
        CNContactPhoneNumbersKey as CNKeyDescriptor,
        CNContactEmailAddressesKey as CNKeyDescriptor,
        CNContactPostalAddressesKey as CNKeyDescriptor,
        CNContactViewController.descriptorForRequiredKeys(),
        CNContactVCardSerialization.descriptorForRequiredKeys()
    ]
```

I wonder if any of those could be trimmed out?

SignalMessaging/utils/LRUCache.swift
```diff
@@ -21,39 +21,61 @@ public class AnyLRUCache: NSObject {
...
     private var cacheMap: [KeyType: ValueType] = [:]
     private var cacheOrder: [KeyType] = []
     private let maxSize: Int
 
+    @objc
     public init(maxSize: Int) {
         self.maxSize = maxSize
+
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(didReceiveMemoryWarning),
```
michaelkirk-signal: I think it's cool to make are LRUCache fancier like this, but I still think a lot of work has gone into NSCache, and expect that on average we'll get better performance by using it.

Signal/src/ViewControllers/AddContactShareToExistingContactViewController.swift
charlesmchen-signal: Great question - I don't think so, though.

* `CNContactThumbnailImageDataKey` we recently learned that some `CNContacts` have images and some have thumbnails, some have both and some have neither.  To make a best effort to show an avatar, we need both keys.  I think the thumbnails are always smaller.
* We need the name, phone numbers, emails, addresses.
* `CNContactViewController.descriptorForRequiredKeys` we need this for contact editing.
* `CNContactVCardSerialization.descriptorForRequiredKeys` we need this for contact sharing. We don't do it yet, but my understanding is that this is currently a subset of the keys we already use so it should only serve to future-proof us.

SignalMessaging/contacts/OWSContactsManager.m
charlesmchen-signal: I've done a bunch of measuring.  I thought I had a "smoking gun" scenario where `LRUCache` dramatically out-performs `NSCache`, but I can't remember/find it now.  I'll go with `NSCache`.

charlesmchen-signal:
PTAL @michaelkirk

charlesmchen-signal:
Same thing - no need to re-review unless you want to.  I was looking for LGTM, didn't see it, missed the approval. I've addressed your feedback.

PR #3564, opened 2018-06-19
Charlesmchen/outage detection

charlesmchen-signal:
PTAL @michaelkirk 

* There's probably some follow-up work restyling the "warning" state of reminder view.  I'll do that after #3562 is merged.

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -78,6 +78,7 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
 @property (nonatomic) NSLayoutConstraint *hideDeregisteredViewConstraint;
 @property (nonatomic) NSLayoutConstraint *hideArchiveReminderViewConstraint;
 @property (nonatomic) NSLayoutConstraint *hideMissingContactsPermissionViewConstraint;
+@property (nonatomic) NSLayoutConstraint *outageViewConstraint;
 
 @property (nonatomic) TSThread *lastThread;
 
@@ -166,6 +167,10 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
                                              selector:@selector(deregistrationStateDidChange:)
                                                  name:DeregistrationStateDidChangeNotification
                                                object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(outageStateDidChange:)
+                                                 name:OutageDetection.outageStateDidChange
+                                               object:nil];
 }
 
 - (void)dealloc
@@ -192,6 +197,13 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
 }
 
 - (void)deregistrationStateDidChange:(id)notification
+{
+    OWSAssertIsOnMainThread();
+
+    [self updateReminderViews];
+}
+
+- (void)outageStateDidChange:(id)notification
 {
     OWSAssertIsOnMainThread();
 
@@ -232,6 +244,13 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
     [reminderStackView addArrangedSubview:deregisteredView];
     self.hideDeregisteredViewConstraint = [deregisteredView autoSetDimension:ALDimensionHeight toSize:0];
     self.hideDeregisteredViewConstraint.priority = UILayoutPriorityRequired;
+
+    ReminderView *outageView = [ReminderView
+        nagWithText:NSLocalizedString(@"OUTAGE_WARNING", @"Label warning the user that the Signal service may be down.")
+          tapAction:nil];
+    [reminderStackView addArrangedSubview:outageView];
+    self.outageViewConstraint = [outageView autoSetDimension:ALDimensionHeight toSize:0];
+    self.outageViewConstraint.priority = UILayoutPriorityRequired;
 
     ReminderView *archiveReminderView =
         [ReminderView explanationWithText:NSLocalizedString(@"INBOX_VIEW_ARCHIVE_MODE_REMINDER",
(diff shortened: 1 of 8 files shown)


Signal/src/views/ReminderView.swift
```diff
@@ -126,6 +124,9 @@ class ReminderView: UIView {
     }
 
     @objc func handleTap(gestureRecognizer: UIGestureRecognizer) {
-        tapAction()
+        guard gestureRecognizer.state == .recognized else {
```
michaelkirk-signal: I've always wanted to ask you, since you often insert this incantation - when would this method get called when the gesture state is not recognized?

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1396,6 +1396,9 @@
 /* Info Message when {{other user}} updates message expiration to {{time amount}}, see the *_TIME_AMOUNT strings for context. */
 "OTHER_UPDATED_DISAPPEARING_MESSAGES_CONFIGURATION" = "%@ set disappearing message time to %@.";
 
+/* Label warning the user that the Signal service may be down. */
+"OUTAGE_WARNING" = "Signal is experiencing technical difficulties. We are working hard to restore service as quickly as possible.";
```
michaelkirk-signal: /cc @jlund

SignalServiceKit/src/Network/OutageDetection.swift
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    // We only show the outage warning when we're certain there's an outage.
+    // DNS lookup failures, etc. are not considered an outage.
+    private func checkForOutageSync() -> Bool {
+        let host = CFHostCreateWithName(nil, "uptime.signal.org" as CFString).takeRetainedValue()
+        CFHostStartInfoResolution(host, .addresses, nil)
+        var success: DarwinBoolean = false
+        guard let addresses = CFHostGetAddressing(host, &success)?.takeUnretainedValue() as NSArray? else {
```
michaelkirk-signal: don't we want to takeRetainedValue here?

SignalServiceKit/src/Network/OutageDetection.swift
```diff
@@ -0,0 +1,122 @@
...
+                    return
+                }
+
+                guard let strongSelf = self else {
+                    return
+                }
+
+                strongSelf.checkForOutageAsync()
+            }
+        } else {
```
michaelkirk-signal: Why do we disable the uptime monitor when there's been a network failure? 

Won't there often be network failures during the very service interruptions we want to advertise?

nit: and if we do want this check, a guard would be more legible than this short dangling else.

SignalServiceKit/src/Network/OutageDetection.swift
```diff
@@ -0,0 +1,122 @@
...
+
+        mayHaveOutage = true
+        hasOutage = false
+    }
+
+    @objc
+    public func reportNetworkFailure() {
+        SwiftAssertIsOnMainThread(#function)
+
+        mayHaveOutage = false
```
michaelkirk-signal: It might be more straight forward to call this what it's used for, rather than the state it kind of sort of represents: something like `shouldRunUptimeMonitor`

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -197,12 +209,21 @@ - (void)loadView
         [SignalApp.sharedApp setHomeViewController:self];
     }
 
+    ReminderView *outageView = [ReminderView
```
michaelkirk-signal: A made a similar observation in another PR, but as we're getting more potentially reminder views, it seems like a (potentially empty) UIStackView would be cleaner than juggling constraints.

Signal/src/views/ReminderView.swift
charlesmchen-signal: `UIGestureRecognizers` have a state.  Their `action` is called whenever the state changes and your handler is supposed to respond to that state.  The exception is `UITapGestureRecognizer`.  For some reason (presumably simplicity over consistency), Apple was inconsistent with this GR.  Tap GRs only invoke their action when they recognize.   So this incantation isn't strictly necessary, but I always add it in case they ever decide to make tap GRs consistent with the others, since tap handlers should only do their work when the GR is in the "recognized" state.

Signal/translations/en.lproj/Localizable.strings
charlesmchen-signal: This is the copy from the design doc which I believe Riya already ran by Josh.

SignalServiceKit/src/Network/OutageDetection.swift
charlesmchen-signal: I could be wrong, but here's my take: you only want to use `takeRetainedValue` when the object in question's ownership is transferred to you, e.g. by a `copy` or `create` method or function.  `CFHostGetAddressing` is a `get` function and doesn't transfer ownership, so it doesn't increment the retain count, so it wouldn't be safe for us to decrement the retain count, so we shouldn't `takeRetainedValue`.

SignalServiceKit/src/Network/OutageDetection.swift
charlesmchen-signal: Whoops, I inverted the semantics of these two methods while cleaning up.  I've renamed some of the methods and properties to hopefully make it clearer.  

We _only_ want to check for outages while experiencing failures connecting to the service - and want to stop checking as soon as we successfully connect.

SignalServiceKit/src/Network/OutageDetection.swift
charlesmchen-signal: I've gone with `shouldCheckForOutage`.

Signal/src/ViewControllers/HomeView/HomeViewController.m
charlesmchen-signal: Yup, done.

charlesmchen-signal: Well, I used a stack view - I'm not sure replacing the constraints will simplify things.

charlesmchen-signal:
PTAL @michaelkirk

charlesmchen-signal:
Oh - no need to re-review unless you want to.  I missed the approval.

PR #3565, opened 2018-06-19
Improve logging around deserialization failures.

charlesmchen-signal:
PTAL @michaelkirk


SignalServiceKit/src/Storage/TSRecipient.h
```diff
@@ -1,12 +1,14 @@
...
 //
 
+#import "TSYapDatabaseObject.h"
+
 NS_ASSUME_NONNULL_BEGIN
 
 // Some lingering TSRecipient records in the wild causing crashes.
 // This is a stop gap until a proper cleanup happens.
-@interface TSRecipient : NSObject <NSCoding>
+@interface TSRecipient : TSYapDatabaseObject
```
charlesmchen-signal: @michaelkirk I've expanded the scope of this PR to fix a crash on launch. We now rebuild db extensions that are corrupt on launch.  If a user has a lingering `TSRecipient` instance, this would crash.  It's simpler and safer to make this a `TSYapDatabaseObject` than to use stub implementations which return nil when deserializing. 

FYI this is the `doesNotRecognizeSelector` crash.

SignalServiceKit/src/Util/OWSAnalytics.m
```diff
@@ -399,7 +399,7 @@ - (void)logEvent:(NSString *)eventName
 
 - (BOOL)isSeverityAsync:(OWSAnalyticsSeverity)severity
```
michaelkirk-signal: nit: this method name is not intuitive to me, maybe: `shouldReportAsync:` or invert it's usage to `shouldReportSync:`

charlesmchen-signal: Donee.

PR #3567, opened 2018-06-20
Tidy up type checking in database views.

charlesmchen-signal:
PTAL @michaelkirk


SignalServiceKit/src/Storage/TSRecipient.h
```diff
@@ -1,12 +1,14 @@
...
 //
 
+#import "TSYapDatabaseObject.h"
+
 NS_ASSUME_NONNULL_BEGIN
 
 // Some lingering TSRecipient records in the wild causing crashes.
 // This is a stop gap until a proper cleanup happens.
-@interface TSRecipient : NSObject <NSCoding>
+@interface TSRecipient : TSYapDatabaseObject
```
michaelkirk-signal: Just a reminder that this is going away.

charlesmchen-signal: Yup!

PR #3575, opened 2018-06-21
Improve app settings buttons.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/AppSettings/AppSettingsViewController.m
+++ b/Signal/src/ViewControllers/AppSettings/AppSettingsViewController.m
@@ -227,43 +227,51 @@
                                               }]];
 #endif
 
-    [section addItem:[OWSTableItem itemWithCustomCellBlock:^{
-        UITableViewCell *cell = [UITableViewCell new];
-        cell.preservesSuperviewLayoutMargins = YES;
-        cell.contentView.preservesSuperviewLayoutMargins = YES;
-        cell.selectionStyle = UITableViewCellSelectionStyleNone;
-
-        const CGFloat kButtonHeight = 40.f;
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/AppSettings/AppSettingsViewController.m
```diff
@@ -227,45 +227,53 @@ - (void)updateTableContents
...
-                                             target:self
-                                           selector:@selector(unregisterUser)];
-        }
-        [cell.contentView addSubview:button];
-        [button autoSetDimension:ALDimensionHeight toSize:kButtonHeight];
-        [button autoVCenterInSuperview];
-        [button autoPinLeadingAndTrailingToSuperviewMargin];
-
-        return cell;
+    if (TSAccountManager.sharedInstance.isDeregistered) {
```
charlesmchen-signal: Hopefully this is more readable.  In the "un-registered" state, include two buttons: "re-register" and "delete all data" (which skips the "delete account request" step of the "delete account" button).

Signal/src/ViewControllers/AppSettings/AppSettingsViewController.m
```diff
@@ -227,45 +227,53 @@ - (void)updateTableContents
...
+        [section addItem:[self destructiveButtonItemWithTitle:NSLocalizedString(@"SETTINGS_REREGISTER_BUTTON",
+                                                                  @"Label for re-registration button.")
+                                                     selector:@selector(reregisterUser)
+                                                        color:[UIColor ows_materialBlueColor]]];
```
charlesmchen-signal: We could move this button up under the network status indicator...

PR #3577, opened 2018-06-21
Show legal terms link

michaelkirk-signal:
PTAL @charlesmchen 

We could remove all the ifdef, but let's wait until this feature is actually in the wild in case we have to pull it again.


michaelkirk-signal:
We could remove all the ifdef, but let's wait until this feature is actually in the wild in case we have to pull it again.

PR #3580, opened 2018-06-22
archive banner

michaelkirk-signal:
fixes: 
image

After:
banners

Also fixes our "no contacts" banner which was not being displayed.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -304,7 +304,7 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
 - (void)updateReminderViews
 {
     self.archiveReminderView.hidden = self.homeViewMode != HomeViewMode_Archive;
-    self.missingContactsPermissionView.hidden = !self.shouldShowMissingContactsPermissionView;
+    self.missingContactsPermissionView.hidden = !self.contactsManager.isSystemContactsDenied;
     self.deregisteredView.hidden = !TSAccountManager.sharedInstance.isDeregistered;
     self.outageView.hidden = !OutageDetection.sharedManager.hasOutage;
 }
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -308,25 +303,10 @@ - (void)loadView
 
 - (void)updateReminderViews
 {
-    BOOL shouldHideArchiveReminderView = self.homeViewMode != HomeViewMode_Archive;
```
michaelkirk-signal: Since we're using a stack view, we can show/hide the subviews rather than track a separate constraint. I think this is simpler.

Signal/src/views/ReminderView.swift
```diff
@@ -73,49 +73,36 @@ class ReminderView: UIView {
...
         // Label
         label.font = UIFont.ows_dynamicTypeSubheadline
-        container.addSubview(label)
+        container.addArrangedSubview(label)
         label.textColor = UIColor.black.withAlphaComponent(0.9)
         label.numberOfLines = 0
         label.lineBreakMode = .byWordWrapping
-        label.autoPinLeadingToSuperviewMargin()
-        label.autoPinEdge(toSuperviewEdge: .top)
-        label.autoPinEdge(toSuperviewEdge: .bottom)
```
michaelkirk-signal: the primary fix is here - we weren't pinning the trailing edge of the label when there wasn't an icon.

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -719,15 +699,6 @@ - (void)tableViewSetUp
     self.tableView.tableFooterView = [[UIView alloc] initWithFrame:CGRectZero];
 }
 
-- (BOOL)shouldShowMissingContactsPermissionView
-{
-    if (!self.contactsManager.systemContactsHaveBeenRequestedAtLeastOnce) {
```
michaelkirk-signal: We introduced a guard such that this `systemContactsHaveBeenRequestedAtLeastOnce` was never being set if contacts access was denied. I think it's just simpler to check for the explicit denied state rather than "has been requested and not authorized"

Signal/src/ViewControllers/HomeView/HomeViewController.m
charlesmchen-signal: Definitely simpler. Sorry, I didn't realize `UIStackView` would ignore hidden views.

Signal/src/ViewControllers/HomeView/HomeViewController.m
charlesmchen-signal: +1

PR #3582, opened 2018-06-22
FIX: mute icon corrupted in homeview

michaelkirk-signal:
Don't clobber icon font

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewCell.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewCell.m
@@ -177,18 +177,16 @@ NS_ASSUME_NONNULL_BEGIN
     self.payloadView.spacing = 0.f;
     self.topRowView.spacing = self.topRowHSpacing;
 
+    // We update the fonts every time this cell is configured to ensure that
+    // changes to the dynamic type settings are reflected.
+    self.snippetLabel.font = [self snippetFont];
+
     if (overrideSnippet) {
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/HomeView/HomeViewCell.m
```diff
@@ -177,18 +177,16 @@ - (void)configureWithThread:(ThreadViewModel *)thread
     self.payloadView.spacing = 0.f;
     self.topRowView.spacing = self.topRowHSpacing;
 
+    // We update the fonts every time this cell is configured to ensure that
```
michaelkirk-signal: I'm not sure this is right - it seems like you deliberately moved this below, but I can't tell why. What specified font do we *want* to remove?

charlesmchen-signal: I can't recall the exact repro, but it shouldn't be hard to find - just test with dynamic type set to a fairly large size.

michaelkirk-signal: > I can't recall the exact repro, but it shouldn't be hard to find - just test with dynamic type set to a fairly large size.

In my testing, even very large dynamic text seems fine with this approach. 

To be clear - the part I'm concerned about was this part of the comment which no longer exists:

// Note: we apply this font _after_ we set the attributed text to	
// override any font attributes.

What font attributes would we want to override?

PR #3583, opened 2018-06-22
Charlesmchen/tweak messages1

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.h
+++ b/Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.h
@@ -74,7 +74,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)loadForDisplayWithTransaction:(YapDatabaseReadTransaction *)transaction;
 
-- (CGSize)cellSize;
+- (CGSize)cellSizeWithTransaction:(YapDatabaseReadTransaction *)transaction;
 
 @end
 
--- a/Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.h
+++ b/Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.h
@@ -4,6 +4,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class ConversationLayoutInfo;
 @class ConversationViewCell;
 @class ConversationViewItem;
 @class OWSContactOffersInteraction;
@@ -69,12 +70,11 @@ NS_ASSUME_NONNULL_BEGIN
 // * Users enters another view (e.g. conversation settings view, call screen, etc.).
 @property (nonatomic) BOOL isCellVisible;
 
-// The width of the collection view.
-@property (nonatomic) int contentWidth;
+@property (nonatomic, nullable) ConversationLayoutInfo *layoutInfo;
 
 - (void)loadForDisplayWithTransaction:(YapDatabaseReadTransaction *)transaction;
 
-- (CGSize)cellSizeForViewWidth:(int)viewWidth contentWidth:(int)contentWidth;
+- (CGSize)cellSize;
 
 @end
 
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.h
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.h
@@ -8,10 +8,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 extern const CGFloat kOWSMessageCellCornerRadius;
 
-extern const CGFloat kBubbleVRounding;
-extern const CGFloat kBubbleHRounding;
-extern const CGFloat kBubbleThornSideInset;
-extern const CGFloat kBubbleThornVInset;
 extern const CGFloat kBubbleTextHInset;
 extern const CGFloat kBubbleTextTopInset;
 extern const CGFloat kBubbleTextBottomInset;
@@ -29,10 +25,6 @@ extern const CGFloat kBubbleTextBottomInset;
 #pragma mark -
 
 @interface OWSBubbleView : UIView
-
-@property (nonatomic) BOOL isOutgoing;
-@property (nonatomic) BOOL hideTail;
-@property (nonatomic) BOOL isTruncated;
 
 @property (nonatomic, nullable) UIColor *bubbleColor;
 
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
@@ -9,10 +9,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 const CGFloat kOWSMessageCellCornerRadius = 18;
 
-const CGFloat kBubbleVRounding = kOWSMessageCellCornerRadius;
-const CGFloat kBubbleHRounding = kOWSMessageCellCornerRadius;
-const CGFloat kBubbleThornSideInset = 5.f;
-const CGFloat kBubbleThornVInset = 0;
 const CGFloat kBubbleTextHInset = 10.f;
 const CGFloat kBubbleTextTopInset = 8.f;
 const CGFloat kBubbleTextBottomInset = 6.f;
@@ -46,39 +42,6 @@ const CGFloat kBubbleTextBottomInset = 6.f;
     _partnerViews = [NSMutableArray new];
 
     return self;
-}
-
-- (void)setIsOutgoing:(BOOL)isOutgoing
-{
-    BOOL didChange = _isOutgoing != isOutgoing;
-
-    _isOutgoing = isOutgoing;
-
-    if (didChange) {
-        [self updateLayers];
-    }
-}
-
-- (void)setHideTail:(BOOL)hideTail
-{
-    BOOL didChange = _hideTail != hideTail;
-
-    _hideTail = hideTail;
-
-    if (didChange) {
-        [self updateLayers];
-    }
-}
-
-- (void)setIsTruncated:(BOOL)isTruncated
-{
-    BOOL didChange = _isTruncated != isTruncated;
-
-    _isTruncated = isTruncated;
-
-    if (didChange) {
-        [self updateLayers];
-    }
 }
 
 - (void)setFrame:(CGRect)frame
@@ -165,64 +128,14 @@ const CGFloat kBubbleTextBottomInset = 6.f;
 
 - (UIBezierPath *)maskPath
 {
-    return [self.class maskPathForSize:self.bounds.size
-                            isOutgoing:self.isOutgoing
-                              hideTail:self.hideTail
-                           isTruncated:self.isTruncated
-                                 isRTL:self.isRTL];
+    return [self.class maskPathForSize:self.bounds.size];
 }
 
 + (UIBezierPath *)maskPathForSize:(CGSize)size
-                       isOutgoing:(BOOL)isOutgoing
-                         hideTail:(BOOL)hideTail
-                      isTruncated:(BOOL)isTruncated
-                            isRTL:(BOOL)isRTL
 {
-    UIBezierPath *bezierPath = [UIBezierPath new];
-
-    CGFloat bubbleLeft = 0.f;
-    CGFloat bubbleRight = size.width - kBubbleThornSideInset;
-    CGFloat bubbleTop = 0.f;
-    CGFloat bubbleBottom = size.height - kBubbleThornVInset;
-
-    [bezierPath moveToPoint:CGPointMake(bubbleLeft + kBubbleHRounding, bubbleTop)];
-    [bezierPath addLineToPoint:CGPointMake(bubbleRight - kBubbleHRounding, bubbleTop)];
-    [bezierPath addQuadCurveToPoint:CGPointMake(bubbleRight, bubbleTop + kBubbleVRounding)
-                       controlPoint:CGPointMake(bubbleRight, bubbleTop)];
-    [bezierPath addLineToPoint:CGPointMake(bubbleRight, bubbleBottom - kBubbleVRounding)];
-
-    if (hideTail) {
-        [bezierPath addQuadCurveToPoint:CGPointMake(bubbleRight - kBubbleHRounding, bubbleBottom)
-                           controlPoint:CGPointMake(bubbleRight, bubbleBottom)];
-    } else {
-        // Thorn Tip
-        CGPoint thornTip = CGPointMake(size.width + 1, size.height);
-        CGPoint thornB = CGPointMake(bubbleRight, bubbleBottom - kBubbleVRounding);
-        // Approximate intersection of the thorn and the bubble edge.
-        CGPoint thornPrime
-            = CGPointMake(bubbleRight - kBubbleHRounding * 0.25f, bubbleBottom - kBubbleVRounding * 0.25f);
-        CGPoint thornPrimeA = CGPointMake(thornPrime.x, bubbleBottom - kBubbleVRounding * 0.08f);
-
-        [bezierPath addQuadCurveToPoint:thornTip controlPoint:CGPointMake(thornB.x, bubbleBottom)];
-        [bezierPath addQuadCurveToPoint:thornPrime controlPoint:thornPrimeA];
-        [bezierPath addQuadCurveToPoint:CGPointMake(bubbleRight - kBubbleHRounding, bubbleBottom)
-                           controlPoint:thornPrimeA];
-    }
-
-    [bezierPath addLineToPoint:CGPointMake(bubbleLeft + kBubbleHRounding, bubbleBottom)];
-    [bezierPath addQuadCurveToPoint:CGPointMake(bubbleLeft, bubbleBottom - kBubbleVRounding)
-                       controlPoint:CGPointMake(bubbleLeft, bubbleBottom)];
-    [bezierPath addLineToPoint:CGPointMake(bubbleLeft, bubbleTop + kBubbleVRounding)];
-    [bezierPath addQuadCurveToPoint:CGPointMake(bubbleLeft + kBubbleHRounding, bubbleTop)
-                       controlPoint:CGPointMake(bubbleLeft, bubbleTop)];
-
-    // Horizontal Flip If Necessary
-    BOOL shouldFlip = isOutgoing == isRTL;
-    if (shouldFlip) {
-        CGAffineTransform flipTransform = CGAffineTransformMakeTranslation(size.width, 0.0);
-        flipTransform = CGAffineTransformScale(flipTransform, -1.0, 1.0);
-        [bezierPath applyTransform:flipTransform];
-    }
+    CGRect bounds = CGRectZero;
+    bounds.size = size;
+    UIBezierPath *bezierPath = [UIBezierPath bezierPathWithRoundedRect:bounds cornerRadius:kOWSMessageCellCornerRadius];
     return bezierPath;
 }
 
@@ -255,7 +168,7 @@ const CGFloat kBubbleTextBottomInset = 6.f;
 
 + (CGFloat)minWidth
 {
-    return (kBubbleHRounding * 2 + kBubbleThornSideInset);
+    return (kOWSMessageCellCornerRadius * 2);
 }
 
 @end
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
@@ -74,7 +74,6 @@ NS_ASSUME_NONNULL_BEGIN
     buttonStackView.spacing = self.vSpacing;
     // Ensure all of the buttons have the same width.
     buttonStackView.alignment = UIStackViewAlignmentFill;
-    buttonStackView.layoutMargins = UIEdgeInsetsZero;
 
     self.stackView = [[UIStackView alloc] initWithArrangedSubviews:@[
         self.titleLabel,
@@ -83,7 +82,6 @@ NS_ASSUME_NONNULL_BEGIN
     self.stackView.axis = UILayoutConstraintAxisVertical;
     self.stackView.spacing = self.vSpacing;
     self.stackView.alignment = UIStackViewAlignmentCenter;
-    self.stackView.layoutMargins = UIEdgeInsetsZero;
     [self.contentView addSubview:self.stackView];
 }
 
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
@@ -41,8 +41,6 @@ NS_ASSUME_NONNULL_BEGIN
 {
     OWSAssert(!self.titleLabel);
 
-    self.preservesSuperviewLayoutMargins = NO;
-    self.contentView.preservesSuperviewLayoutMargins = NO;
     self.layoutMargins = UIEdgeInsetsZero;
     self.contentView.layoutMargins = UIEdgeInsetsZero;
 
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
@@ -165,7 +165,7 @@ NS_ASSUME_NONNULL_BEGIN
     return (self.buttonVPadding + CGSizeCeil([self.addToContactsButton sizeThatFits:CGSizeZero]).height);
 }
 
-- (CGSize)cellSize
+- (CGSize)cellSizeWithTransaction:(YapDatabaseReadTransaction *)transaction
 {
     OWSAssert(self.layoutInfo);
     OWSAssert(self.layoutInfo.viewWidth > 0);
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
@@ -18,6 +18,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic) UIButton *addToContactsButton;
 @property (nonatomic) UIButton *addToProfileWhitelistButton;
 @property (nonatomic) UIButton *blockButton;
+@property (nonatomic) NSArray<NSLayoutConstraint *> *layoutConstraints;
+@property (nonatomic) UIStackView *stackView;
 
 @end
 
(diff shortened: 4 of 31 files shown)


Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.h
```diff
@@ -69,12 +70,11 @@ NS_ASSUME_NONNULL_BEGIN
 // * Users enters another view (e.g. conversation settings view, call screen, etc.).
 @property (nonatomic) BOOL isCellVisible;
 
-// The width of the collection view.
-@property (nonatomic) int contentWidth;
+@property (nonatomic, nullable) ConversationLayoutInfo *layoutInfo;
```
charlesmchen-signal: This PR introduces `ConversationLayoutInfo`, which gathers layout-related metadata like sizes, spacing, margins.  I suspect it will get elaborated.

Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.h
```diff
@@ -8,10 +8,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 extern const CGFloat kOWSMessageCellCornerRadius;
 
-extern const CGFloat kBubbleVRounding;
-extern const CGFloat kBubbleHRounding;
-extern const CGFloat kBubbleThornSideInset;
-extern const CGFloat kBubbleThornVInset;
```
charlesmchen-signal: Tweaked the thorns.

Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
```diff
@@ -38,16 +41,17 @@ - (void)commontInit
 {
     OWSAssert(!self.titleLabel);
 
-    //    [self setTranslatesAutoresizingMaskIntoConstraints:NO];
+    self.preservesSuperviewLayoutMargins = NO;
+    self.contentView.preservesSuperviewLayoutMargins = NO;
+    self.layoutMargins = UIEdgeInsetsZero;
+    self.contentView.layoutMargins = UIEdgeInsetsZero;
```
charlesmchen-signal: Converted this cell to use auto layout.

Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
```diff
@@ -85,30 +106,55 @@ + (NSString *)cellReuseIdentifier
...
-{
-    return UIFont.ows_dynamicTypeBodyFont.ows_mediumWeight;
-}
-
-- (UIFont *)buttonFont
-{
-    return UIFont.ows_dynamicTypeBodyFont;
-}
+    CGFloat buttonWidth = 0.f;
+    if (interaction.hasAddToContactsOffer) {
```
charlesmchen-signal: In the next PR, I'm going to just manipulate the button's `hidden` property.

Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
```diff
@@ -85,30 +106,55 @@ + (NSString *)cellReuseIdentifier
...
+    CGFloat buttonHeight = self.buttonHeight;
+    [NSLayoutConstraint deactivateConstraints:self.layoutConstraints];
+    self.layoutConstraints = @[
+        [self.addToContactsButton autoSetDimension:ALDimensionWidth toSize:buttonWidth],
+        [self.addToProfileWhitelistButton autoSetDimension:ALDimensionWidth toSize:buttonWidth],
+        [self.blockButton autoSetDimension:ALDimensionWidth toSize:buttonWidth],
+
+        [self.addToContactsButton autoSetDimension:ALDimensionHeight toSize:buttonHeight],
+        [self.addToProfileWhitelistButton autoSetDimension:ALDimensionHeight toSize:buttonHeight],
+        [self.blockButton autoSetDimension:ALDimensionHeight toSize:buttonHeight],
```
charlesmchen-signal: I might rework this to use a second stack view, which would hopefully let us skip this sizing.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -855,41 +850,46 @@ - (void)showAttachmentErrorViewWithMediaView:(UIView *)mediaView
...
+- (CGSize)bodyTextSize:(BOOL)includeMargins
 {
+    OWSAssert(self.layoutInfo);
+    OWSAssert(self.layoutInfo.maxMessageWidth > 0);
+
     if (!self.hasBodyText) {
         return CGSizeZero;
     }
 
-    BOOL isRTL = self.isRTL;
```
charlesmchen-signal: I'm trying to slowly converge on using `CurrentAppContext.isRTL`.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -926,28 +926,32 @@ - (CGSize)bodyMediaSizeForContentWidth:(int)contentWidth
...
+            result = CGSizeMake(
                 maxMessageWidth, [OWSContactShareView bubbleHeightForContactShare:self.viewItem.contactShare]);
+            break;
     }
-}
 
-- (int)maxMessageWidthForContentWidth:(int)contentWidth
-{
-    return (int)floor(contentWidth * 0.8f);
+    return CGSizeCeil(result);
```
charlesmchen-signal: I'm trying to move towards using `CGSizeCeil()` to ensure point-align our layouts where possible.

Signal/src/ViewControllers/ConversationView/Cells/OWSSystemMessageCell.m
```diff
@@ -22,6 +23,8 @@ @interface OWSSystemMessageCell ()
 
 @property (nonatomic) UIImageView *imageView;
 @property (nonatomic) UILabel *titleLabel;
+@property (nonatomic) UIStackView *stackView;
+@property (nonatomic) NSArray<NSLayoutConstraint *> *layoutConstraints;
```
charlesmchen-signal: Converted this to use auto layout and stack view.

Signal/src/ViewControllers/ConversationView/Cells/OWSUnreadIndicatorCell.m
```diff
@@ -15,12 +16,9 @@ @interface OWSUnreadIndicatorCell ()
...
 @property (nonatomic, nullable) TSUnreadIndicatorInteraction *interaction;
 
-@property (nonatomic) UIView *bannerView;
-@property (nonatomic) UIView *bannerTopHighlightView;
-@property (nonatomic) UIView *bannerBottomHighlightView1;
-@property (nonatomic) UIView *bannerBottomHighlightView2;
 @property (nonatomic) UILabel *titleLabel;
-@property (nonatomic) UILabel *subtitleLabel;
+@property (nonatomic) UIView *strokeView;
+@property (nonatomic) NSArray<NSLayoutConstraint *> *layoutConstraints;
```
charlesmchen-signal: Started to tweak this indicator.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -252,6 +260,14 @@ - (nullable ConversationViewCell *)measurementCell
     return measurementCell;
 }
 
+- (CGFloat)vSpacingWithLastLayoutItem:(id<ConversationViewLayoutItem>)lastLayoutItem
+{
+    OWSAssert(lastLayoutItem);
+
+    // TODO:
+    return 0.f;
```
charlesmchen-signal: I'm going to tweak intra-cell spacing here.

Signal/src/ViewControllers/ConversationView/ConversationLayoutInfo.swift
```diff
@@ -0,0 +1,74 @@
...
+
+    @objc public let contentMarginTop: CGFloat = 10
+    @objc public let contentMarginBottom: CGFloat = 10
+
+    @objc public var gutterLeading: CGFloat = 0
+    @objc public var gutterTrailing: CGFloat = 0
+    // These are the gutters used by "full width" views
+    // like "date headers" and "unread indicator".
+    @objc public var fullWidthGutterLeading: CGFloat = 0
+    @objc public var fullWidthGutterTrailing: CGFloat = 0
```
charlesmchen-signal: Tweaks the gutters, etc.

SignalServiceKit/src/Messages/Interactions/OWSDisappearingConfigurationUpdateInfoMessage.h
```diff
@@ -20,6 +20,8 @@ NS_ASSUME_NONNULL_BEGIN
               createdByRemoteName:(nullable NSString *)remoteName
            createdInExistingGroup:(BOOL)createdInExistingGroup;
 
+- (NSString *)previewText;
```
charlesmchen-signal: In some cases we can skip a transaction.

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -83,6 +87,22 @@ class MessageDetailViewController: OWSViewController, MediaGalleryDataSourceDele
             object: OWSPrimaryStorage.shared().dbNotificationObject)
     }
 
+    override public func viewWillLayoutSubviews() {
+        Logger.debug("\(self.logTag) in \(#function)")
+
+        super.viewWillLayoutSubviews()
+
+        self.conversationLayoutInfo.viewWidth = self.view.width()
```
michaelkirk-signal: Most of the time the view size isn't changing. Would `viewWillTransition:toSize:` be a better candidate?

Signal/src/ViewControllers/ConversationView/ConversationViewLayout.m
```diff
@@ -94,60 +94,42 @@ - (void)prepareLayout
...
-                break;
-            }
-            case ConversationViewLayoutAlignment_FullWidth:
-                itemFrame = CGRectMake(0, y, viewWidth, layoutSize.height);
-                break;
-            case ConversationViewLayoutAlignment_Center:
-                itemFrame = CGRectMake(
-                    hInset + round((viewWidth - layoutSize.width) * 0.5f), y, layoutSize.width, layoutSize.height);
-                break;
+        if (lastLayoutItem) {
```
michaelkirk-signal: nit: `last` is ambiguous. I think you mean `previous`?

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -1980,7 +1983,10 @@ + (DebugUIMessagesAction *)fakeRandomTextAction:(TSThread *)thread
                 OWSAssert(messageToQuote);
                 DDLogVerbose(@"%@ %@", self.logTag, label);
                 [DDLog flushLog];
-                ConversationViewItem *viewItem = [[ConversationViewItem alloc] initWithInteraction:messageToQuote isGroupThread:thread.isGroupThread transaction:transaction];
+                ConversationViewItem *viewItem = [[ConversationViewItem alloc] initWithInteraction:messageToQuote
+                                                                                     isGroupThread:thread.isGroupThread
+                                                                                       transaction:transaction
+                                                                                        layoutInfo:layoutInfo];
```
michaelkirk-signal: I'm somewhat resistant to caching the layoutInfo on the ConversationViewItem. As you can see here, we've perviously used them in multiple contexts, often outside of the ConversationViewController.

Also, it seems kind of awkward in the MessageDetailViewController.

If possible, it seems like the ConversationViewLayout or the ConversationViewController would be a better place to maintain a cache of any layout attributes.

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -589,9 +605,7 @@ class MessageDetailViewController: OWSViewController, MediaGalleryDataSourceDele
             return
         }
 
-        messageBubbleView.contentWidth = contentWidth()
-
-        let messageBubbleSize = messageBubbleView.size(forContentWidth: contentWidth())
+        let messageBubbleSize = messageBubbleView.measureSize()
```
michaelkirk-signal: FWIW the old interface where messageBubbleView doesn't maintain state about it's layout context made more sense to me.  It seems like the VC should be directing the views, not the other way around.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -252,6 +260,14 @@ - (nullable ConversationViewCell *)measurementCell
     return measurementCell;
 }
 
+- (CGFloat)vSpacingWithLastLayoutItem:(id<ConversationViewLayoutItem>)lastLayoutItem
```
michaelkirk-signal: s/Last/Previous ?

Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
```diff
@@ -62,19 +66,48 @@ - (void)commontInit
...
+    UIStackView *buttonStackView = [[UIStackView alloc] initWithArrangedSubviews:@[
+        self.addToContactsButton,
+        self.addToProfileWhitelistButton,
+        self.blockButton,
+    ]];
+    buttonStackView.axis = UILayoutConstraintAxisVertical;
+    buttonStackView.spacing = self.vSpacing;
+    // Ensure all of the buttons have the same width.
+    buttonStackView.alignment = UIStackViewAlignmentFill;
+    buttonStackView.layoutMargins = UIEdgeInsetsZero;
```
michaelkirk-signal: For stack views, unless you're setting `isLayoutMarginsRelativeArrangement` I don't think layoutMargins matter.

Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
```diff
@@ -126,67 +159,36 @@ - (CGFloat)buttonVPadding
...
-    };
-
-    layoutButton(self.addToContactsButton, interaction.hasAddToContactsOffer);
-    layoutButton(self.addToProfileWhitelistButton, interaction.hasAddToProfileWhitelistOffer);
-    layoutButton(self.blockButton, interaction.hasBlockOffer);
+    return (self.buttonVPadding + CGSizeCeil([self.addToContactsButton sizeThatFits:CGSizeZero]).height);
 }
 
-- (CGSize)cellSizeForViewWidth:(int)viewWidth contentWidth:(int)contentWidth
+- (CGSize)cellSize
```
michaelkirk-signal: What is the motivation for removing the parameter and instead keeping layout info statefully on this model?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -926,28 +926,32 @@ - (CGSize)bodyMediaSizeForContentWidth:(int)contentWidth
             }
 
             return CGSizeRound(CGSizeMake(mediaWidth, mediaHeight));
```
michaelkirk-signal: why this break after a return? Did you intend to leave this return?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -51,11 +49,13 @@ - (void)commontInit
     // Ensure only called once.
     OWSAssert(!self.messageBubbleView);
 
-    _viewConstraints = [NSMutableArray new];
-
+    self.preservesSuperviewLayoutMargins = NO;
```
michaelkirk-signal: Isn't this default? Why this change?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -256,11 +234,9 @@ - (void)updateDateHeader
         self.dateHeaderLabel.hidden = NO;
 
         [self.viewConstraints addObjectsFromArray:@[
-            // Date headers should be visually centered within the conversation view,
-            // so they need to extend outside the cell's boundaries.
-            [self.dateHeaderLabel autoSetDimension:ALDimensionWidth toSize:self.contentWidth],
-            (self.isIncoming ? [self.dateHeaderLabel autoPinEdgeToSuperviewEdge:ALEdgeLeading]
-                             : [self.dateHeaderLabel autoPinEdgeToSuperviewEdge:ALEdgeTrailing]),
+            // TODO: Are data headers symmetric or are they asymmetric? gutters are asymmetric?
```
michaelkirk-signal: re: this TODO

My understanding is that dates should be horizontally centered between the screen edges (not the the gutters).

Let's try to nip these little questions in the bud by asking the authority. Figuring out the right thing before opening the PR means less throw-away code for me to review. (and then re-review once you get the answer).

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -432,10 +408,6 @@ - (CGSize)cellSizeForViewWidth:(int)viewWidth contentWidth:(int)contentWidth
         cellSize.height += self.footerHeight;
     }
 
-    if (self.shouldHaveFailedSendBadge) {
```
michaelkirk-signal: Just making sure - this is intentionally going away with no replacement for now?

Signal/src/ViewControllers/ConversationView/ConversationLayoutInfo.swift
```diff
@@ -0,0 +1,74 @@
...
+            SwiftAssertIsOnMainThread(#function)
+
+            updateProperties()
+        }
+    }
+
+    @objc public let contentMarginTop: CGFloat = 10
+    @objc public let contentMarginBottom: CGFloat = 10
+
+    @objc public var gutterLeading: CGFloat = 0
```
michaelkirk-signal: Are we really mutating instances of ConversationLayoutInfo? Or can these all be readonly?

Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
charlesmchen-signal: We want to layout with regard to the cell _edges_, not it's margins.  Our design uses non-standard margins.  This is the default value and redundant with our setting the layout margins to zero as we do here, so I'll remove them - the intent was to make our intention clear (use non-standard layout with respect to edges) and future-proof us against future changes.

SignalServiceKit/src/Messages/Interactions/OWSDisappearingConfigurationUpdateInfoMessage.h
charlesmchen-signal: > It's true that sometimes we can skip using the transaction but, since we do need it sometimes we need to support having a transaction available to configure cells anyway, right?

We never need them for call/info/error messages which are rendered by "system message" cells, so I was trying to simply those cells.  But I've removed this since you don't like it.

Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
```diff
@@ -165,64 +128,14 @@ - (void)updateLayers
 
 - (UIBezierPath *)maskPath
 {
-    return [self.class maskPathForSize:self.bounds.size
-                            isOutgoing:self.isOutgoing
-                              hideTail:self.hideTail
-                           isTruncated:self.isTruncated
```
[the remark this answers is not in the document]
charlesmchen-signal: This was only used in `OWSBubbleView`'s mask generation logic - which didn't even consult that parameter, so I guess it vestigial.

Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
```diff
@@ -38,16 +41,17 @@ - (void)commontInit
 {
     OWSAssert(!self.titleLabel);
 
-    //    [self setTranslatesAutoresizingMaskIntoConstraints:NO];
+    self.preservesSuperviewLayoutMargins = NO;
```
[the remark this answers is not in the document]
charlesmchen-signal: See above - I was just trying to leave a reminder in these cells that - unlike our table view cells - they use non-standard margins.  As noted above, I've removed these.

SignalServiceKit/src/Messages/Interactions/OWSDisappearingConfigurationUpdateInfoMessage.h
charlesmchen-signal: Oh wait, I just remembered another reason why I did it - to avoid plumbing through a transaction in the measurement methods (e.g. `cellSize`) which also needs to do this kind of configuration.

Signal/src/ViewControllers/MessageDetailViewController.swift
charlesmchen-signal: Sure, that should be equivalent.

Signal/src/ViewControllers/ConversationView/ConversationViewLayout.m
charlesmchen-signal: That's better, thanks.

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
charlesmchen-signal: * As an aside, I'm thinking of renaming `ConversationLayoutInfo` to `ConversationAppearance` or `ConversationStyle`, since I think it'll better capture the intention. 
* Conversation cell layout and styling (e.g. colors) is contextual (to the thread, the containing view size, etc.) so we definitely need either this kind of class or possibly a protocol/delegate pattern.
* `ConversationViewController` is a massive file - 5k+ lines.  I'm hoping that we can slowly decompose it into smaller components with well-defined responsibilities.  And I'm trying to resist it growing any larger.  The "conversation style" info seems like an opportunity to gather/bundle some related code.
* `ConversationViewLayout` is a layout for Conversation view's collection view - none of its logic or state is relevant to message details view.  Even if we did move these properties and logic there, we'd need to do just as much work to keep it's "view width" state updated.
* There are other alternatives that would make using this "appearance/style" class a little easier, but they would require KVO of the root view's frame & bounds, and the docs discourage that - and the hacks to do it are awkward and not future-proof.

Want to discuss when you come online?

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
charlesmchen-signal: Done.

Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
charlesmchen-signal: Removed.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
charlesmchen-signal: Fixed, thanks.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: Removed, see above.

Signal/src/ViewControllers/ConversationView/ConversationLayoutInfo.swift
charlesmchen-signal: * Some of the future properties (added in later PRs) are mutable, e.g. to reflect changes to dynamic type size.
* I'm trying to build this logic to make future work to support orientation changes easy, where these views will automatically update to reflect changes to the root view size.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: It's on my list of questions to review with Myles.  I'm trying to batch my questions to him (I already have ~30 from just a day's work).

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: It's on my list of questions to review with Myles.  I'm trying to batch my questions to him (I already have ~30 from just a day's work).

Signal/src/ViewControllers/ConversationView/Cells/OWSContactOffersCell.m
charlesmchen-signal: See above - there's now a ton of "constants" and "semi-constants" the are calculated from each other.  I'm trying to avoid: 

* Passing them all around separately
* Calculating them all for each cell.

We could maintain an instance of "conversation layout info/style/appearance" and pass it into the cell's "configure" and "measurement" methods?  Would that feel better to you?  If it's easy, I'd be happy to do it.

charlesmchen-signal:
PTAL @michaelkirk

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -963,31 +967,28 @@ - (CGSize)quotedMessageSizeForContentWidth:(int)contentWidth includeMargins:(BOO
...
     cellSize.width = MAX(cellSize.width, quotedMessageSize.width);
     cellSize.height += quotedMessageSize.height;
 
-    CGSize mediaContentSize = [self bodyMediaSizeForContentWidth:contentWidth];
+    CGSize mediaContentSize = [self bodyMediaSize];
     cellSize.width = MAX(cellSize.width, mediaContentSize.width);
     cellSize.height += mediaContentSize.height;
 
-    CGSize textContentSize = [self bodyTextSizeForContentWidth:contentWidth includeMargins:YES];
+    CGSize textContentSize = [self bodyTextSize:YES];
```
michaelkirk-signal: nit: Can you label the parameter in the message signature e.g. `bodyTextSizeWithIncludeMargins:YES`?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -411,16 +384,17 @@ - (UIFont *)dateHeaderTimeFont
...
+    OWSAssert(self.layoutInfo.viewWidth > 0);
     OWSAssert(self.viewItem);
     OWSAssert([self.viewItem.interaction isKindOfClass:[TSMessage class]]);
     OWSAssert(self.messageBubbleView);
 
     self.messageBubbleView.viewItem = self.viewItem;
-    self.messageBubbleView.contentWidth = self.contentWidth;
     self.messageBubbleView.cellMediaCache = self.delegate.cellMediaCache;
-    CGSize messageBubbleSize = [self.messageBubbleView sizeForContentWidth:contentWidth];
+    CGSize messageBubbleSize = [self.messageBubbleView measureSize];
```
michaelkirk-signal: I was surprised that `measureSize` didn't ever need a transaction to compute it's sizing, but it looks like everything it uses is already cached on the viewModel. 👍

Signal/src/ViewControllers/ConversationView/Cells/OWSSystemMessageCell.m
```diff
@@ -240,48 +264,31 @@ - (CGFloat)iconSize
...
+    CGSize result = CGSizeMake(self.layoutInfo.viewWidth, 0);
 
-    // FIXME pass in transaction from the uiDBConnection.
-    [[TSYapDatabaseObject dbReadConnection] readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
-        [self applyTitleForInteraction:interaction label:self.titleLabel transaction:transaction];
-    }];
-
-    CGFloat maxTitleWidth = (contentWidth - ([self hMargin] * 2.f + [self hSpacing] + [self iconSize]));
-    CGSize titleSize = [self.titleLabel sizeThatFits:CGSizeMake(maxTitleWidth, CGFLOAT_MAX)];
+    [self applyTitleForInteraction:interaction label:self.titleLabel transaction:transaction];
```
michaelkirk-signal: 🎉

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
charlesmchen-signal: Done.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: Yup, that's where I'd like to get for all of this.

PR #3584, opened 2018-06-22
Mkirk/no unread status for search results

michaelkirk-signal:
Don't show the unread badge or bold text for FTS message results (we still show them for conversation results).

PTAL @charlesmchen 

## before 

before_messages

## after messages

after_messgaes

## after conversations 

after_conversation

### DIFF
--- a/SignalMessaging/Views/ContactTableViewCell.m
+++ b/SignalMessaging/Views/ContactTableViewCell.m
@@ -93,7 +93,6 @@ NS_ASSUME_NONNULL_BEGIN
 
     [self.cellView prepareForReuse];
 
-    self.accessoryView = nil;
     self.accessoryType = UITableViewCellAccessoryNone;
 }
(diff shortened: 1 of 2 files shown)


SignalMessaging/Views/ContactTableViewCell.m
```diff
@@ -93,7 +93,6 @@ - (void)prepareForReuse
 
     [self.cellView prepareForReuse];
 
-    self.accessoryView = nil;
```
charlesmchen-signal: Was this hitting an assert?

michaelkirk-signal: Yes:

```
- (void)setAccessoryView:(nullable UIView *)accessoryView
{
    OWSFail(@"%@ use ows_setAccessoryView instead.", self.logTag);
}
```

PR #3593, opened 2018-06-25
Mkirk/webrtc m67

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/call/PeerConnectionClient.swift
+++ b/Signal/src/call/PeerConnectionClient.swift
@@ -58,12 +58,12 @@ protocol PeerConnectionClientDelegate: class {
     /**
      * Fired whenever the local video track become active or inactive.
      */
-    func peerConnectionClient(_ peerconnectionClient: PeerConnectionClient, didUpdateLocal videoTrack: RTCVideoTrack?)
+    func peerConnectionClient(_ peerconnectionClient: PeerConnectionClient, didUpdateLocalVideoTrack videoTrack: RTCVideoTrack?, captureSession: AVCaptureSession?)
 
     /**
      * Fired whenever the remote video track become active or inactive.
      */
-    func peerConnectionClient(_ peerconnectionClient: PeerConnectionClient, didUpdateRemote videoTrack: RTCVideoTrack?)
+    func peerConnectionClient(_ peerconnectionClient: PeerConnectionClient, didUpdateRemoteVideoTrack videoTrack: RTCVideoTrack?)
 }
 
 // In Swift (at least in Swift v3.3), weak variables aren't thread safe. It
@@ -383,9 +383,16 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
         let proxyCopy = self.proxy
         let completion = {
             guard let strongSelf = proxyCopy.get() else { return }
+
+            // Should these really be guards? Don't we want to pass nil when it's been disabled?
             guard let localVideoTrack = strongSelf.localVideoTrack else { return }
+            guard let videoCaptureSession = strongSelf.videoCaptureSession else { return }
             guard let strongDelegate = strongSelf.delegate else { return }
-            strongDelegate.peerConnectionClient(strongSelf, didUpdateLocal: enabled ? localVideoTrack : nil)
+
+            let videoTrack = enabled ? localVideoTrack : nil
+            let captureSession = enabled ? videoCaptureSession : nil
+
+            strongDelegate.peerConnectionClient(strongSelf, didUpdateLocalVideoTrack: videoTrack, captureSession: captureSession)
         }
 
         PeerConnectionClient.signalingQueue.async {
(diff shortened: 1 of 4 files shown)


Signal/src/call/PeerConnectionClient.swift
```diff
@@ -393,30 +400,42 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
 
             DispatchQueue.main.async(execute: completion)
         }
     }
 
-    // MARK: Audio
+    // MARK: VideoCaptureSettingsDelegate
+
+    // MJK: fixme
+    var videoWidth: Int32 {
```
michaelkirk-signal: Default width/height is TBD, but if you'd like to review the remaining changes, I could let you know which values I settle on.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -331,49 +336,38 @@ class PeerConnectionClient: NSObject, RTCPeerConnectionDelegate, RTCDataChannelD
...
-        let cameraConstraints = RTCMediaConstraints(mandatoryConstraints: nil,
-                                                    optionalConstraints: nil)
-
-        // TODO: Revisit the cameraConstraints.
-        let videoSource = factory.avFoundationVideoSource(with: cameraConstraints)
-        self.localVideoSource = videoSource
-
-        self.videoCaptureSession = videoSource.captureSession
-        videoSource.useBackCamera = false
+        let videoSource = factory.videoSource()
```
michaelkirk-signal: These are the primary upstream API changes, the fallout of which is this PR.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -1092,6 +1110,90 @@ class HardenedRTCSessionDescription {
     }
 }
 
+protocol VideoCaptureSettingsDelegate: class {
+    var videoWidth: Int32 { get }
+    var videoHeight: Int32 { get }
+}
+
+class VideoCaptureController {
```
michaelkirk-signal: This is based on the example class from upstream: https://webrtc.googlesource.com/src/+/c522e758f6a364c23e7232d4f402713601f0b35e%5E%21/

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -1092,6 +1110,90 @@ class HardenedRTCSessionDescription {
...
+            return
+        }
+
+        let fps = self.framesPerSecond(format: format)
+
+        capturer.startCapture(with: device, format: format, fps: fps)
+    }
+
+    public func stopCapture() {
+        self.capturer.stopCapture()
```
charlesmchen-signal: why all of this unnecessary `self.`?  NBD.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -1092,6 +1110,90 @@ class HardenedRTCSessionDescription {
...
+protocol VideoCaptureSettingsDelegate: class {
+    var videoWidth: Int32 { get }
+    var videoHeight: Int32 { get }
+}
+
+class VideoCaptureController {
+
+    let capturer: RTCCameraVideoCapturer
+    weak var settingsDelegate: VideoCaptureSettingsDelegate?
+    var isUsingFrontCamera: Bool = true
```
charlesmchen-signal: I'm slightly concerned about thread safety.  We seem to access `VideoCaptureController` on the main thread and the "signaling queue".  `isUsingFrontCamera` for example is mutated and accessed on both, right?  Should we make the properties of this class private and synchronize access to them?

Signal/src/call/PeerConnectionClient.swift
michaelkirk-signal: I'm personally not a fan of the "implicit self", as I find it makes it harder to distinguish from a local variable. 

I would prefer that we always add the explicit self, even when not required. But I don't feel that strongly about it and am happy to conform if you have a preference for the implicit self.

Signal/src/call/PeerConnectionClient.swift
michaelkirk-signal: Good point - I've serialized read/write to the affected methods.

Signal/src/call/PeerConnectionClient.swift
michaelkirk-signal: After some testing, this actually seems like a reasonable default. I'm checking with Android to make sure we're consistent.

Signal/src/call/PeerConnectionClient.swift
```diff
@@ -1092,6 +1109,121 @@ class HardenedRTCSessionDescription {
...
+    var videoWidth: Int32 { get }
+    var videoHeight: Int32 { get }
+}
+
+class VideoCaptureController {
+
+    let serialQueue = DispatchQueue(label: "org.signal.videoCaptureController")
+    let capturer: RTCCameraVideoCapturer
+    weak var settingsDelegate: VideoCaptureSettingsDelegate?
+    var isUsingFrontCamera: Bool = true
```
charlesmchen-signal: private?

michaelkirk-signal:
merged into release/2.27.0 with 89402d0d2, and followups in #3601

Signal/src/call/PeerConnectionClient.swift
michaelkirk-signal: Good point, done!

PR #3597, opened 2018-06-26
Refine conversation style.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationStyle.swift
+++ b/Signal/src/ViewControllers/ConversationView/ConversationStyle.swift
@@ -48,8 +48,8 @@ public class ConversationStyle: NSObject {
         }
     }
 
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/ConversationStyle.swift
```diff
@@ -48,8 +48,8 @@ public class ConversationStyle: NSObject {
         }
     }
 
-    @objc public let contentMarginTop: CGFloat = 10
-    @objc public let contentMarginBottom: CGFloat = 10
+    @objc public let contentMarginTop: CGFloat = 24
+    @objc public let contentMarginBottom: CGFloat = 24
```
charlesmchen-signal: These changes aren't complete or final, but they reflect conversations with Myles and are closer to where we'll land.

PR #3598, opened 2018-06-26
Charlesmchen/footer view

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/Cells/OWSExpirableMessageView.h
```diff
@@ -1,5 +0,0 @@
-//
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
-//
```
charlesmchen-signal: Vestigial.

Signal/src/ViewControllers/ConversationView/Cells/OWSExpirationTimerView.h
```diff
@@ -1,25 +0,0 @@
-//
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
-//
-
-#import <UIKit/UIKit.h>
-
-NS_ASSUME_NONNULL_BEGIN
-
-extern const CGFloat kExpirationTimerViewSize;
```
charlesmchen-signal: This will be replaced by something like "message status indicator view".

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -22,6 +23,8 @@ @interface OWSMessageBubbleView () <OWSQuotedMessageViewDelegate, OWSContactShar
 
 @property (nonatomic) OWSBubbleView *bubbleView;
 
+@property (nonatomic) UIStackView *stackView;
```
charlesmchen-signal: Use a stack view to layout content of message bubbles.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -1004,14 +1012,28 @@ - (CGSize)measureSize
...
 
     cellSize = CGSizeCeil(cellSize);
 
     return cellSize;
 }
 
+- (BOOL)hasFooter
+{
+    // TODO:
+    return YES;
```
charlesmchen-signal: There's work here around show/hide of footers.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
```diff
@@ -0,0 +1,210 @@
...
+#import "Signal-Swift.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OWSMessageFooterView ()
+
+@property (nonatomic) UILabel *timestampLabel;
+@property (nonatomic) UIView *spacerView;
+@property (nonatomic) UILabel *statusLabel;
+@property (nonatomic) UIView *statusIndicatorView;
```
charlesmchen-signal: This is a placeholder.

Signal/src/ViewControllers/ConversationView/ConversationStyle.swift
```diff
@@ -68,7 +40,9 @@ public class ConversationStyle: NSObject {
     @objc public var maxMessageWidth: CGFloat = 0
     @objc public var maxFooterWidth: CGFloat = 0
 
-    @objc public var textInsets = OWSDirectionalEdgeInsets.zero
+    @objc public var textInsetTop: CGFloat = 0
+    @objc public var textInsetBottom: CGFloat = 0
+    @objc public var textInsetHorizontal: CGFloat = 0
```
charlesmchen-signal: In some contexts, we'll use leading/trailing; in other cases, we'll use left/right.  They're always the same value, so rather than do a bunch of conversions, easiest to use a single value.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -390,56 +358,66 @@ - (void)configureViews
...
-            [tapForMoreLabel autoSetDimension:ALDimensionHeight toSize:self.tapForMoreHeight],
+            [footerView autoPinLeadingToSuperviewMarginWithInset:self.conversationStyle.textInsetHorizontal],
+            [footerView autoPinTrailingToSuperviewMarginWithInset:self.conversationStyle.textInsetHorizontal],
+            [footerView autoPinBottomToSuperviewMarginWithInset:self.conversationStyle.textInsetBottom],
         ]];
-        lastSubview = tapForMoreLabel;
-        bottomMargin = textInsets.bottom;
+        [self.footerView setHasShadows:YES viewItem:self.viewItem];
+    } else {
+        // Display footer over media.
```
michaelkirk-signal: Is this comment in the right spot? It seems like it belongs in the previous branch, or you could just delete it.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -390,56 +358,66 @@ - (void)configureViews
...
+        OWSFail(@"%@ could not display footer.", self.logTag);
     }
 
-    OWSAssert(lastSubview);
-    [self.viewConstraints addObjectsFromArray:@[
-        [lastSubview autoPinEdgeToSuperviewEdge:ALEdgeBottom withInset:bottomMargin],
-    ]];
+    if (textStackView) {
```
michaelkirk-signal: There seems to be some subtlety to this that I'm not following. Why do we want the BubbleView to be less than it's measured size? Wouldn't we want it to be exactly equal?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
```diff
@@ -0,0 +1,185 @@
...
+    self.statusIndicatorView.backgroundColor = [UIColor orangeColor];
+
+    for (UIView *subview in @[
+             self.timestampLabel,
+             self.statusLabel,
+             self.statusIndicatorView,
+         ]) {
+        [subview removeFromSuperview];
+    }
+    if (viewItem.interaction.interactionType == OWSInteractionType_OutgoingMessage) {
```
michaelkirk-signal: Would it be cleaner instead of removing and adding the subviews here in `configureWith...` to add all of them only once in `commonInit` and only touch their isHidden property in this method?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
charlesmchen-signal: Good q. I changed my mind and modified this in a subsequent branch.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
charlesmchen-signal: Done.

PR #3599, opened 2018-06-26
Compact layout / widow reduction.

charlesmchen-signal:
We might not want to do that; we'll need to do a perf analysis of the cost, but we should do that after other tweaking is complete.

PTAL @michaelkirk


charlesmchen-signal:
> For ease of profiling, can you extract this into a method?

Done, good idea.

PR #3601, opened 2018-06-26
CR: make members private where possible

michaelkirk-signal:
add missing weak capture

PTAL @charlesmchen


michaelkirk-signal:
In 9c73dbc88

PR #3603, opened 2018-06-26
Charlesmchen/fixed bubble size

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/ConversationStyle.swift
```diff
@@ -38,9 +38,6 @@ public class ConversationStyle: NSObject {
     @objc public var fullWidthContentWidth: CGFloat = 0
 
     @objc public var maxMessageWidth: CGFloat = 0
-    // TODO: Can we eliminate this after moving timestamps and
-    //       message status inside the message bubbles.
-    @objc public var maxFooterWidth: CGFloat = 0
```
charlesmchen-signal: This property isn't necessary any more.

PR #3605, opened 2018-06-26
Fix corner rounding.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -76,8 +76,6 @@ NS_ASSUME_NONNULL_BEGIN
     self.stackView = [UIStackView new];
     self.stackView.axis = UILayoutConstraintAxisVertical;
     self.stackView.alignment = UIStackViewAlignmentFill;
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -260,6 +258,9 @@ - (void)configureViews
     CGSize bodyMediaContentSize = [self bodyMediaSize];
     CGSize bodyTextContentSize = [self bodyTextSizeWithIncludeMargins:NO];
 
+    [self.bubbleView addSubview:self.stackView];
```
michaelkirk-signal: Why this change - as far as I can tell we never remove the bubbleView.?

charlesmchen-signal: In `prepareForReuse` we remove all subviews from `bubbleView`.

michaelkirk-signal: oh sorry I somehow misread this as `[self addSubview:bubleView]`

This looks fine!

PR #3606, opened 2018-06-26
FIX: obscured searchbar upon returning

michaelkirk-signal:
When navigating back from a conversation, we could be scrolled down a ways.
Since the search bar lives in the inbox table view, we need to make sure the
inbox table view is scrolled all the way up to keep the search bar visible.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -529,18 +529,24 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
 
     self.isViewVisible = YES;
 
-    // When returning to home view, try to ensure that the "last" thread is still
-    // visible.  The threads often change ordering while in conversation view due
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -529,18 +529,24 @@ - (void)viewWillAppear:(BOOL)animated
...
     self.isViewVisible = YES;
 
-    // When returning to home view, try to ensure that the "last" thread is still
-    // visible.  The threads often change ordering while in conversation view due
-    // to incoming & outgoing messages.
-    if (self.lastThread) {
+    BOOL isShowingSearchResults = !self.searchResultsController.view.hidden;
+    if (isShowingSearchResults) {
+        OWSAssert(self.searchBar.text.ows_stripped.length > 0);
+        self.tableView.contentOffset = CGPointZero;
```
michaelkirk-signal: This should not be necessary, but it seems like a good safety.

PR #3612, opened 2018-06-28
Migrate SignalAccount collection.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/contacts/OWSContactsManager.m
+++ b/SignalMessaging/contacts/OWSContactsManager.m
@@ -75,7 +75,7 @@ NSString *const OWSContactsManagerSignalAccountsDidChangeNotification
 - (void)loadSignalAccountsFromCache
 {
     __block NSMutableArray<SignalAccount *> *signalAccounts;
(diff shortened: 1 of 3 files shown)


SignalMessaging/contacts/OWSContactsManager.m
```diff
@@ -384,7 +384,10 @@ - (void)buildSignalAccountsAndClearStaleCache:(BOOL)shouldClearStaleCache
         }
 
         // Update cached SignalAccounts on disk
-        [self.dbWriteConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
+        [self.dbWriteConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
+            // Clear legacy collection.
+            [transaction removeAllObjectsInCollection:[SignalAccount collection_old]];
```
michaelkirk-signal: If this is meant to be run one time only, a migration seems more fitting.

PR #3613, opened 2018-06-28
Charlesmchen/tweak colors

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/ConversationStyle.swift
```diff
@@ -34,7 +34,6 @@ public class ConversationStyle: NSObject {
     @objc public var contentWidth: CGFloat = 0
 
     // viewWidth - (gutterfullWidthGutterLeadingLeading + fullWidthGutterTrailing)
-    // TODO: Is this necessary?
```
charlesmchen-signal: I suggest we move this class to SignalMessaging and pull the contents of `OWSMessagesBubbleImageFactory`/`OWSMessagesBubbleColors` into this class.

SignalMessaging/categories/UIColor+OWS.h
```diff
@@ -32,6 +29,37 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (UIColor *)blendWithColor:(UIColor *)otherColor alpha:(CGFloat)alpha;
 
+#pragma mark - New Colors
```
michaelkirk-signal: nit: Nothing is new for long, maybe "Conversation"

SignalMessaging/categories/UIColor+OWS.m
```diff
@@ -180,6 +162,148 @@ - (UIColor *)blendWithColor:(UIColor *)otherColor alpha:(CGFloat)alpha
                            alpha:CGFloatLerp(a0, a1, alpha)];
 }
 
+#pragma mark - New Colors
+
++ (UIColor *)ows_signalBlueColor
+{
+    return [UIColor colorWithRGBHex:0x2090EA];
```
michaelkirk-signal: glad to have this RGBHex method. 👍

Signal/src/ViewControllers/ConversationView/ConversationStyle.swift
michaelkirk-signal: Sure - is it OK if I include that in my work, or is that blocking something you're doing?

SignalMessaging/categories/UIColor+OWS.h
charlesmchen-signal: Removed.  I think we'll be using these colors beyond conversations.

PR #3616, opened 2018-06-28
Tweak message cells.

charlesmchen-signal:
Tweaked to add two shadows under body media.  If perf needs improvement, this is something we could eliminate.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -23,7 +23,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic) OWSBubbleView *bubbleView;
 
-@property (nonatomic) OWSBubbleShapeView *mediaShadowView;
+// TODO: We may only end up using a single shadow.
+@property (nonatomic) OWSBubbleShapeView *mediaShadowView1;
+
+@property (nonatomic) OWSBubbleShapeView *mediaShadowView2;
 
 @property (nonatomic) OWSBubbleShapeView *mediaClipView;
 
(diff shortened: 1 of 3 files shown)


SignalMessaging/categories/UIView+OWS.h
```diff
@@ -106,6 +106,8 @@ CGFloat ScaleFromIPhone5(CGFloat iPhone5Value);
 // When using a UIView as a "div" to structure layout, we don't want it to have margins.
 - (void)setHLayoutMargins:(CGFloat)value;
 
+- (NSArray<NSLayoutConstraint *> *)autoPinToEdgesOfView:(UIView *)view;
```
michaelkirk-signal: nice

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -23,7 +23,10 @@ @interface OWSMessageBubbleView () <OWSQuotedMessageViewDelegate, OWSContactShar
 
 @property (nonatomic) OWSBubbleView *bubbleView;
 
-@property (nonatomic) OWSBubbleShapeView *mediaShadowView;
+// TODO: We may only end up using a single shadow.
+@property (nonatomic) OWSBubbleShapeView *mediaShadowView1;
```
michaelkirk-signal: Can you leave a comment explaining what the difference is between these shadows? 

It seems like you're trying to create a sort of non-linear effect by overlapping the two shadows?

Would innerShadow and outerShadow be accurate names?

charlesmchen-signal: "Inner shadows" is a term (I think it originated with photoshop?) for shadows _on_ an object (as opposed to outside it) that highlight its edges, sort of like bas-relief.  

Anyway, the design has changed and we're not doing inner shadows any more.  I don't understand why we have two separate shadows - please ask Myles.

PR #3618, opened 2018-06-28
Tweak break spacing.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
@@ -275,9 +275,9 @@ NSString *NSStringForOWSMessageCellType(OWSMessageCellType cellType)
     OWSAssert(previousLayoutItem);
 
     if (self.interaction.interactionType == OWSInteractionType_UnreadIndicator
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -275,9 +275,9 @@ - (CGFloat)vSpacingWithPreviousLayoutItem:(ConversationViewItem *)previousLayout
     OWSAssert(previousLayoutItem);
 
     if (self.interaction.interactionType == OWSInteractionType_UnreadIndicator
-        || previousLayoutItem.interaction.interactionType == OWSInteractionType_UnreadIndicator) {
-        // The unread indicator has its own v-margins.
-        return 0.f;
+        || previousLayoutItem.interaction.interactionType == OWSInteractionType_UnreadIndicator
+        || self.shouldShowDate) {
```
michaelkirk-signal: I've seen a couple changes lately where you pull the margin logic "up" a level, and I think that makes sense. e.g. the views don't have built in padding, allowing us to coordinate all the spacing at the same level. 👍

charlesmchen-signal: Glad you like it.

PR #3619, opened 2018-06-28
Fix audio message layout.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSAudioMessageView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSAudioMessageView.m
@@ -209,11 +209,6 @@ NS_ASSUME_NONNULL_BEGIN
     [self addArrangedSubview:self.audioPlayPauseButton];
     [self.audioPlayPauseButton setContentHuggingHigh];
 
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSAudioMessageView.m
```diff
@@ -247,8 +240,24 @@ - (void)createContents
...
+
+    UIStackView *labelsView = [UIStackView new];
+    labelsView.axis = UILayoutConstraintAxisVertical;
+    labelsView.spacing = [OWSAudioMessageView labelVSpacing];
+    [labelsView addArrangedSubview:topLabel];
+    [labelsView addArrangedSubview:audioProgressView];
     [labelsView addArrangedSubview:bottomLabel];
 
+    // Ensure the "audio progress" and "play button" are v-center-aligned using a container.
+    UIView *labelsContainerView = [UIView containerView];
```
michaelkirk-signal: No biggie, but I don't see why we'd need this containerView to v-align the audioProgressView to the playPauseButton.

PR #3620, opened 2018-06-28
Fix bubble strokes.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -447,6 +447,10 @@ NS_ASSUME_NONNULL_BEGIN
                          }];
 
     [self updateBubbleColor];
+
+    // If we're stroking the bubble edge, ensure the stroke
+    // view is in front of its peers to prevent it from being occluded.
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -447,6 +447,10 @@ - (void)configureViews
                          }];
 
     [self updateBubbleColor];
+
+    // If we're stroking the bubble edge, ensure the stroke
```
michaelkirk-signal: TIOLI, but rather than adding the bubbleStrokeView and then juggling it's position, can instead just add the bubbleStrokeView later? Just trying to keep our concerns colocated.

A quick grep shows we only use the bubbleStrokeView for contactShares - is that right?

charlesmchen-signal: > can instead just add the bubbleStrokeView later?

Ordering matters as we insert subviews into the stack view, so we can't move contact share handling here.  If we move adding the stroke view here, I think we we'd duplicate logic that I'd rather DRY.  

It's admittedly not ideal.  If you have a proposal, I'd be happy to see it.

PR #3621, opened 2018-06-28
Tweak media view gradients.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -422,7 +422,26 @@ NS_ASSUME_NONNULL_BEGIN
         // Do nothing.
     } else if (shouldFooterOverlayMedia) {
         OWSAssert(bodyMediaView);
-        [self.footerView configureWithConversationViewItem:self.viewItem hasShadows:YES];
+
+        CGFloat maxGradientHeight = 48.f;
+        CAGradientLayer *gradientLayer = [CAGradientLayer new];
+        gradientLayer.colors = @[
+            (id)[UIColor colorWithWhite:0.f alpha:0.f].CGColor,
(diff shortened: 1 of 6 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -422,7 +422,26 @@ - (void)configureViews
...
         OWSAssert(bodyMediaView);
-        [self.footerView configureWithConversationViewItem:self.viewItem hasShadows:YES];
+
+        CGFloat maxGradientHeight = 48.f;
+        CAGradientLayer *gradientLayer = [CAGradientLayer new];
+        gradientLayer.colors = @[
+            (id)[UIColor colorWithWhite:0.f alpha:0.f].CGColor,
+            (id)[UIColor colorWithWhite:0.f alpha:0.4f].CGColor,
+        ];
+        OWSLayerView *gradientView =
```
michaelkirk-signal: Did you take a look at using the existing GradientView class?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
```diff
@@ -95,7 +95,7 @@ - (void)configureWithConversationViewItem:(ConversationViewItem *)viewItem hasSh
     }
 
     UIColor *textColor;
-    if (hasShadows) {
+    if (isOverlayingMedia) {
```
michaelkirk-signal: better name 👍

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
charlesmchen-signal: The layout of this gradient is slightly complicated.  IMO it's simpler (and more readable) to express that using a 4-line block than ~5 layout constraints.

PR #3623, opened 2018-06-28
Tweak attachment download view.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -896,6 +896,8 @@ NS_ASSUME_NONNULL_BEGIN
     [wrapper addSubview:downloadView];
     [downloadView autoPinWidthToSuperview];
     [downloadView autoVCenterInSuperview];
+    [downloadView autoPinEdgeToSuperviewMargin:ALEdgeTop relation:NSLayoutRelationGreaterThanOrEqual];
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -862,26 +862,13 @@ - (UIView *)loadViewForDownloadingAttachment
...
+    AttachmentPointerView *downloadView =
         [[AttachmentPointerView alloc] initWithAttachmentPointer:self.attachmentPointer isIncoming:self.isIncoming];
-    [customView addSubview:attachmentPointerView];
-    [attachmentPointerView autoPinWidthToSuperviewWithMargin:20.f];
-    [attachmentPointerView autoVCenterInSuperview];
+
+    UIView *wrapper = [UIView new];
+    [wrapper addSubview:downloadView];
+    [downloadView autoPinWidthToSuperview];
+    [downloadView autoVCenterInSuperview];
```
michaelkirk-signal: do we need to constrain `downloadView` height?

charlesmchen-signal: Done. There's places in our codebase where we skip these kinds of constraints, e.g. when our "measurement" logic guarantees there will be enough space for the centered view.  I'm never sure if these extra constraints are worth it.

PR #3624, opened 2018-06-29
tweak picker

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ColorPickerViewController.swift
+++ b/Signal/src/ViewControllers/ColorPickerViewController.swift
@@ -1,0 +1,145 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+let colorSwatchHeight: CGFloat = 60
+
+class ColorView: UIView {
+    let color: UIColor
+    let swatchView: UIView
+
+    required init(color: UIColor) {
+        self.color = color
+        self.swatchView = UIView()
+
+        super.init(frame: .zero)
+
+        swatchView.backgroundColor = color
+
+        self.swatchView.layer.cornerRadius = colorSwatchHeight / 2
+
+        self.addSubview(swatchView)
+
+        swatchView.autoVCenterInSuperview()
+        swatchView.autoSetDimension(.height, toSize: colorSwatchHeight)
+        swatchView.autoPinEdge(toSuperviewMargin: .top, relation: .greaterThanOrEqual)
+        swatchView.autoPinEdge(toSuperviewMargin: .bottom, relation: .greaterThanOrEqual)
+        swatchView.autoPinLeadingToSuperviewMargin()
+        swatchView.autoPinTrailingToSuperviewMargin()
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+@objc
+protocol ColorPickerDelegate: class {
+    func colorPickerDidCancel(_ colorPicker: ColorPickerViewController)
+    func colorPicker(_ colorPicker: ColorPickerViewController, didPickColorName colorName: String)
+}
+
+@objc
+class ColorPickerViewController: UIViewController, UIPickerViewDelegate, UIPickerViewDataSource {
+
+    private let pickerView: UIPickerView
+    private let thread: TSThread
+    private let colors: [UIColor]
+
+    @objc public weak var delegate: ColorPickerDelegate?
+
+    @objc
+    required init(thread: TSThread) {
+        self.thread = thread
+        self.pickerView = UIPickerView()
+        self.colors = UIColor.ows_conversationColors
+
+        super.init(nibName: nil, bundle: nil)
+
+        self.navigationItem.leftBarButtonItem = UIBarButtonItem(barButtonSystemItem: .cancel, target: self, action: #selector(didTapCancel))
+        self.navigationItem.rightBarButtonItem = UIBarButtonItem(barButtonSystemItem: .save, target: self, action: #selector(didTapSave))
+
+        pickerView.dataSource = self
+        pickerView.delegate = self
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func loadView() {
+        self.view = UIView()
+        view.backgroundColor = .white
+        view.addSubview(pickerView)
+
+        pickerView.autoVCenterInSuperview()
+        pickerView.autoPinLeadingToSuperviewMargin()
+        pickerView.autoPinTrailingToSuperviewMargin()
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        if let colorName = thread.conversationColorName,
+            let currentColor = UIColor.ows_conversationColor(colorName: colorName),
+            let index = colors.index(of: currentColor) {
+            pickerView.selectRow(index, inComponent: 0, animated: false)
+        }
+    }
+
+    // MARK: UIPickerViewDataSource
+
+    public func numberOfComponents(in pickerView: UIPickerView) -> Int {
+        return 1
+    }
+
+    public func pickerView(_ pickerView: UIPickerView, numberOfRowsInComponent component: Int) -> Int {
+        return self.colors.count
+    }
+
+    // MARK: UIPickerViewDelegate
+
+    public func pickerView(_ pickerView: UIPickerView, rowHeightForComponent component: Int) -> CGFloat {
+        let vMargin: CGFloat = 2
+        return colorSwatchHeight + vMargin * 2
+    }
+
+    public func pickerView(_ pickerView: UIPickerView, viewForRow row: Int, forComponent component: Int, reusing view: UIView?) -> UIView {
+        guard let color = colors[safe: row] else {
+            owsFail("\(logTag) in \(#function) color was unexpectedly nil")
+            return ColorView(color: .white)
+        }
+
+        return ColorView(color: color)
+    }
+
+    // MARK: Actions
+
+    var currentColor: UIColor {
+        let index = pickerView.selectedRow(inComponent: 0)
+        guard let color = self.colors[safe: index] else {
+            owsFail("\(self.logTag) in \(#function) index was unexpectedly nil")
+            return UIColor.white
+        }
+
+        return color
+    }
+
+    @objc
+    public func didTapSave() {
+        guard let colorName = UIColor.ows_conversationColorName(color: self.currentColor) else {
+            owsFail("\(self.logTag) in \(#function) colorName was unexpectedly nil")
+            self.delegate?.colorPickerDidCancel(self)
+            return
+        }
+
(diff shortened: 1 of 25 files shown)


SignalMessaging/categories/UIColor+OWS.m
```diff
@@ -98,40 +98,6 @@ + (UIColor *)ows_messageBubbleLightGrayColor
     return [UIColor colorWithHue:240.0f / 360.0f saturation:0.02f brightness:0.92f alpha:1.0f];
 }
 
-+ (UIColor *)backgroundColorForContact:(NSString *)contactIdentifier
```
michaelkirk-signal: pending myles approval, I think it's better to just get rid of these rather than to map legacy values to something similar.

SignalServiceKit/src/Contacts/TSThread.m
```diff
@@ -395,16 +412,58 @@ - (BOOL)isMuted
...
+    NSData *hashData = [Cryptography computeSHA256Digest:contactData truncatedToBytes:hashingLength];
+    [hashData getBytes:&choose length:hashingLength];
+    
+    NSUInteger index = (choose % [self.conversationColorNames count]);
+    return [self.conversationColorNames objectAtIndex:index];
+}
+
++ (NSArray<NSString *> *)conversationColorNames
+{
+    return @[
```
michaelkirk-signal: I felt bad just copying these in, but I wasn't sure what the best way to dedupe was.

We could an assert that does some accounting, we could pull UIColor+OWS into SSK, or we could have a separate class that lives in SSK which UIColor+OWS uses as a helper...

Signal/src/ViewControllers/ColorPickerViewController.swift
```diff
@@ -0,0 +1,143 @@
...
+        pickerView.autoPinLeadingToSuperviewMargin()
+        pickerView.autoPinTrailingToSuperviewMargin()
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        if let colorName = thread.conversationColorName,
+            let currentColor = UIColor.ows_conversationColor(colorName: colorName),
+            let index = colors.index(of: currentColor) {
```
charlesmchen-signal: One has to wonder how UIColor comparison is done and whether this is reliable.  Would it be safer to replace `colors` with `colorNames` and compare color names here?

Signal/src/ViewControllers/ColorPickerViewController.swift
```diff
@@ -0,0 +1,143 @@
...
+        swatchView.backgroundColor = color
+
+        self.swatchView.layer.cornerRadius = colorSwatchHeight / 2
+
+        self.addSubview(swatchView)
+
+        swatchView.autoVCenterInSuperview()
+        swatchView.autoSetDimension(.height, toSize: colorSwatchHeight)
+        swatchView.autoPinLeadingToSuperviewMargin()
+        swatchView.autoPinTrailingToSuperviewMargin()
```
charlesmchen-signal: TIOLO ensure that swatch inside parent.
```
    [swatchView autoPinEdgeToSuperviewMargin:ALEdgeTop relation:NSLayoutRelationGreaterThanOrEqual];
    [swatchView autoPinEdgeToSuperviewMargin:ALEdgeBottom relation:NSLayoutRelationGreaterThanOrEqual];
```

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4326,6 +4326,12 @@ - (void)resendGroupUpdateForErrorMessage:(TSErrorMessage *)message
            }];
 }
 
+- (void)conversationColorWasUpdated
+{
+    [self.conversationStyle updateProperties];
+    [self.headerView updateAvatar];
```
charlesmchen-signal: Don't we need to reload our collection view?

Signal/src/ViewControllers/ThreadSettings/OWSConversationSettingsViewController.m
```diff
@@ -37,10 +37,13 @@
...
 NS_ASSUME_NONNULL_BEGIN
 
-@interface OWSConversationSettingsViewController () <ContactEditingDelegate, ContactsViewHelperDelegate>
+@interface OWSConversationSettingsViewController () <ContactEditingDelegate,
+    ContactsViewHelperDelegate,
+    ColorPickerDelegate>
 
 @property (nonatomic) TSThread *thread;
 @property (nonatomic) YapDatabaseConnection *uiDatabaseConnection;
+@property (nonatomic) YapDatabaseConnection *editingDatabaseConnection;
```
charlesmchen-signal: Why declare this private property - isn't the method below sufficient?

Signal/src/ViewControllers/ThreadSettings/OWSConversationSettingsViewController.m
```diff
@@ -271,6 +279,19 @@ - (void)updateTableContents
                                  [weakSelf showMediaGallery];
                              }]];
 
+
+    [mainSection addItem:[OWSTableItem
+                             itemWithCustomCellBlock:^{
+                                 NSString *colorName = self.thread.conversationColorName;
+                                 UIColor *currentColor = [UIColor ows_conversationColorForColorName:colorName];
+                                 NSString *title = NSLocalizedString(@"CONVERSATION_SETTINGS_CONVERSATION_COLOR",
+                                     @"Indicates that user's profile has been shared with a group.");
```
charlesmchen-signal: l10n comment

Signal/src/ViewControllers/ThreadSettings/OWSConversationSettingsViewController.m
```diff
@@ -1200,6 +1257,36 @@ - (void)otherUsersProfileDidChange:(NSNotification *)notification
...
+    pickerController.delegate = self;
+    OWSNavigationController *modal = [[OWSNavigationController alloc] initWithRootViewController:pickerController];
+
+    [self presentViewController:modal animated:YES completion:nil];
+}
+
+- (void)colorPicker:(ColorPickerViewController *)colorPicker didPickColorName:(NSString *)colorName
+{
+    DDLogDebug(@"%@ in %s picked color: %@", self.logTag, __PRETTY_FUNCTION__, colorName);
+    [self.editingDatabaseConnection asyncReadWriteWithBlock:^(YapDatabaseReadWriteTransaction * _Nonnull transaction) {
```
charlesmchen-signal: Personally, I would make this sync, not async.  We _want_ to block on completion.  If we don't, the user could navigate before the completion is performed. TIOLI

SignalMessaging/categories/UIColor+OWS.h
```diff
@@ -24,9 +24,17 @@ NS_ASSUME_NONNULL_BEGIN
...
 
-+ (UIColor *)backgroundColorForContact:(NSString *)contactIdentifier;
 + (UIColor *)colorWithRGBHex:(unsigned long)value;
 
+#pragma mark - ConversationColor
+
++ (nullable UIColor *)ows_conversationColorForColorName:(NSString *)colorName NS_SWIFT_NAME(ows_conversationColor(colorName:));
++ (nullable NSString *)ows_conversationColorNameForColor:(UIColor *)color
+    NS_SWIFT_NAME(ows_conversationColorName(color:));
+;
```
charlesmchen-signal: extraneous ;

SignalMessaging/categories/UIColor+OWS.m
charlesmchen-signal: +1

SignalMessaging/categories/UIColor+OWS.m
```diff
@@ -304,6 +270,50 @@ + (UIColor *)ows_grey600Color
...
+                     @"grey": self.ows_grey600Color
+                     };
+    });
+    
+    return colorMap;
+}
+
++ (NSArray<NSString *> *)ows_conversationColorNames
+{
+    return self.ows_conversationColorMap.allKeys;
```
charlesmchen-signal: If you take my suggestion above to key the color picker on "color names" instead of "color values", we could sort here to ensure that they are presented with a consistent ordering.

SignalServiceKit/src/Contacts/TSThread.m
```diff
@@ -395,16 +412,58 @@ - (BOOL)isMuted
...
+    NSUInteger count = self.conversationColorNames.count;
+    NSUInteger index = arc4random_uniform((uint32_t)count);
+    return [self.conversationColorNames objectAtIndex:index];
+}
+
++ (NSString *)stableConversationColorNameForString:(NSString *)colorSeed
+{
+    NSData *contactData = [colorSeed dataUsingEncoding:NSUTF8StringEncoding];
+    
+    NSUInteger hashingLength = 8;
```
charlesmchen-signal: `sizeof(unsigned long long)` instead of 8?

Signal/src/ViewControllers/ColorPickerViewController.swift
michaelkirk-signal: Done.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: It's apparently not necessary, but I've added it anyway to be safe.

Signal/src/ViewControllers/ThreadSettings/OWSConversationSettingsViewController.m
michaelkirk-signal: Fixed, thanks!

Signal/src/ViewControllers/ThreadSettings/OWSConversationSettingsViewController.m
michaelkirk-signal: Done.

SignalMessaging/categories/UIColor+OWS.h
michaelkirk-signal: Thanks!

SignalServiceKit/src/Contacts/TSThread.m
michaelkirk-signal: done

Signal/src/ViewControllers/ThreadSettings/OWSConversationSettingsViewController.m
michaelkirk-signal: The method below is sufficient, but, and this is a small subjective style point, properties aren't just for wrapping ivars - I (and Apple) prefer to model idempotent getters as properties. 

e.g. if I wanted to access this connection elsewhere in the code I will probably write: `self.editingDatabaseConnection` not `[self editingDatabaseConnection]`

To me that's a sign that this should be a property. In objc they are of course absolutely equivalent, but Swift actually enforces this distinction (e.g. `self.editingDatabaseConnection()` vs `self.editingDatabaseConnection`).

Signal/src/ViewControllers/ColorPickerViewController.swift
michaelkirk-signal: Done.

PR #3625, opened 2018-06-29
Charlesmchen/tweak cells

charlesmchen-signal:
PTAL @michaelkirk 

Multiple small tweaks.

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
@@ -16,7 +16,7 @@ NS_ASSUME_NONNULL_BEGIN
 // to always keep one around.
 
 @property (nonatomic) OWSMessageBubbleView *messageBubbleView;
-@property (nonatomic) UIView *dateHeaderView;
+@property (nonatomic) UIStackView *dateHeaderView;
 @property (nonatomic) UIView *dateStrokeView;
 @property (nonatomic) UILabel *dateHeaderLabel;
 @property (nonatomic) AvatarImageView *avatarView;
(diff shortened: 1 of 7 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -250,9 +248,11 @@ - (void)updateDateHeader
             [self.dateHeaderView autoPinLeadingToSuperviewMarginWithInset:self.conversationStyle.gutterLeading],
             [self.dateHeaderView autoPinTrailingToSuperviewMarginWithInset:self.conversationStyle.gutterTrailing],
             [self.dateHeaderView autoPinEdgeToSuperviewEdge:ALEdgeTop],
-            [self.dateHeaderView autoSetDimension:ALDimensionHeight toSize:self.dateHeaderHeight],
 
-            [self.messageBubbleView autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:self.dateHeaderView],
+            [self.messageBubbleView autoPinEdge:ALEdgeTop
```
michaelkirk-signal: It seems like there's some subtlety behind this change. This looks equivalent, but less to the point. Why this change?

Signal/src/ViewControllers/ConversationView/Cells/OWSUnreadIndicatorCell.m
```diff
@@ -94,26 +97,14 @@ - (void)loadForDisplayWithTransaction:(YapDatabaseReadTransaction *)transaction
...
-
-        [self.subtitleLabel autoPinEdge:ALEdgeTop
-                                 toEdge:ALEdgeBottom
-                                 ofView:self.titleLabel
-                             withOffset:self.subtitleVSpacing],
-        [self.subtitleLabel autoPinLeadingToSuperviewMarginWithInset:self.conversationStyle.fullWidthGutterLeading],
-        [self.subtitleLabel autoPinTrailingToSuperviewMarginWithInset:self.conversationStyle.fullWidthGutterTrailing],
+        [self.stackView autoPinEdgeToSuperviewEdge:ALEdgeTop],
+        [self.stackView autoPinEdgeToSuperviewEdge:ALEdgeBottom],
+        [self.stackView autoPinLeadingToSuperviewMarginWithInset:self.conversationStyle.fullWidthGutterLeading],
```
michaelkirk-signal: For full width stuff, don't we want to pin to the edge here? not the margin?

Signal/src/ViewControllers/ConversationView/Cells/OWSUnreadIndicatorCell.m
```diff
@@ -135,14 +126,9 @@ - (NSString *)subtitleForInteraction:(TSUnreadIndicatorInteraction *)interaction
...
-- (CGFloat)strokeHeight
-{
-    return 1.f;
-}
-
-- (CGFloat)subtitleVSpacing
+- (CGFloat)strokeThickness
 {
-    return 3.f;
+    return 1.f / UIScreen.mainScreen.scale;
```
michaelkirk-signal: TIOLI: Should we extract a free function like `CGHairlineWidth()`?

charlesmchen-signal: Nice, done.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: Added a comment:

```
            // DO NOT pin to the bottom of dateHeaderView.
            //
            // Being a UIStackView, it doesn't reflect the spacing below the date
            // header contents.  Instead pin using dateHeaderHeight which includes
            // the spacing.
```

Signal/src/ViewControllers/ConversationView/Cells/OWSUnreadIndicatorCell.m
charlesmchen-signal: I was counting on `autoPinLeadingToSuperviewMarginWithInset ` being equivalent `autoPinEdgeToSuperviewEdge:ALEdgeLeading withInset:` since the cell margins are empty.  I preferred the former because it's more concise, but I've switched for clarity.

PR #3626, opened 2018-06-29
Animate sending icon.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
@@ -5,6 +5,7 @@
 #import "OWSMessageFooterView.h"
 #import "DateUtil.h"
 #import "Signal-Swift.h"
+#import <QuartzCore/QuartzCore.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
```diff
@@ -146,6 +148,19 @@ - (void)configureWithConversationViewItem:(ConversationViewItem *)viewItem hasSh
...
+{
+    CABasicAnimation *animation;
+    animation = [CABasicAnimation animationWithKeyPath:@"transform.rotation.z"];
+    animation.toValue = @(M_PI * 2.0);
+    const CGFloat kPeriodSeconds = 1.f;
+    animation.duration = kPeriodSeconds;
+    animation.cumulative = YES;
+    animation.repeatCount = HUGE_VALF;
+
+    [self.statusIndicatorImageView.layer addAnimation:animation forKey:@"animation"];
```
michaelkirk-signal: nit: Do we want a more distinct key name than "animation"?

charlesmchen-signal: I think that only matters if you're using more than one animation on a given layer, since you can then manipulate them by name.

michaelkirk-signal: Exactly - if we add another animation do you think you'll remember to come back and make this one distinct? Or will you again use a generic name and inadvertently clobber this one?

PR #3627, opened 2018-06-29
Charlesmchen/generic attachment file extension

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
@@ -119,6 +119,31 @@ NS_ASSUME_NONNULL_BEGIN
     [self addArrangedSubview:imageView];
     [imageView setContentHuggingHigh];
 
+    NSString *filename = self.attachmentStream.sourceFilename;
+    if (!filename) {
+        filename = [[self.attachmentStream filePath] lastPathComponent];
+    }
+    NSString *fileExtension = filename.pathExtension;
+    if (fileExtension.length < 1) {
+        [MIMETypeUtil fileExtensionForMIMEType:self.attachmentStream.contentType];
+    }
+    if (fileExtension.length < 1) {
+        fileExtension = NSLocalizedString(@"GENERIC_ATTACHMENT_DEFAULT_TYPE",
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
```diff
@@ -118,6 +118,31 @@ - (void)createContents
     [self addArrangedSubview:imageView];
     [imageView setContentHuggingHigh];
 
+    NSString *filename = self.attachmentStream.sourceFilename;
+    if (!filename) {
+        filename = [[self.attachmentStream filePath] lastPathComponent];
+    }
+    NSString *fileExtension = filename.pathExtension;
+    if (fileExtension.length < 1) {
+        [MIMETypeUtil fileExtensionForMIMEType:self.attachmentStream.contentType];
```
michaelkirk-signal: missing `fileExtension = `

Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
```diff
@@ -118,6 +118,31 @@ - (void)createContents
...
+    }
+    if (fileExtension.length < 1) {
+        fileExtension = NSLocalizedString(@"GENERIC_ATTACHMENT_DEFAULT_TYPE",
```
michaelkirk-signal: This text is used to label the icon, right, e.g. the "PDF" that appears atop the file icon?

I'm not clear how this localized string would be used. Can you give an example english localization?

Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
charlesmchen-signal: Fixed.

Signal/src/ViewControllers/ConversationView/Cells/OWSGenericAttachmentView.m
charlesmchen-signal: The contents of this commit were simply restored from when we removed them a few days ago (the design has changed back to including these labels).  I didn't look that closely at this when I restored it.  I'm just going to skip these default label values - I don't think they add anything, but I'll check with Myles.

PR #3628, opened 2018-06-29
Tweak old attachment upload view.

charlesmchen-signal:
PTAL @michaelkirk


charlesmchen-signal:
> Is there a different attachment upload view that we're using instead?

It's in the queue of daily questions for Myles, but from what I can tell, no.

michaelkirk-signal:
It seems like a big change - let's wait until we have the authoritative answer so I don't have to review again if we add it back.

charlesmchen-signal:
Closing in favor of https://github.com/signalapp/Signal-iOS/pull/3680

PR #3630, opened 2018-06-29
Charlesmchen/sender names3

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -4910,7 +4910,7 @@ typedef enum : NSUInteger {
                             || viewItem.shouldShowDate);
                 }
                 if (shouldShowSenderName) {
-                    senderName = [self.contactsManager displayNameForPhoneIdentifier:incomingSenderId];
+                    senderName = [self.contactsManager contactOrProfileNameForPhoneIdentifier:incomingSenderId];
                 }
 
                 // Show the sender avatar for incoming group messages unless
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -1206,10 +1206,10 @@ typedef enum : NSUInteger {
         }
     } else {
         OWSAssert(self.thread.contactIdentifier);
-        name = [self.contactsManager
-            attributedStringForConversationTitleWithPhoneIdentifier:self.thread.contactIdentifier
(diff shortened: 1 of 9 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4923,7 +4923,7 @@ - (void)reloadViewItems
                             || viewItem.shouldShowDate);
                 }
                 if (shouldShowSenderName) {
-                    senderName = [self.contactsManager displayNameForPhoneIdentifier:incomingSenderId];
+                    senderName = [self.contactsManager contactOrProfileNameForPhoneIdentifier:incomingSenderId];
```
michaelkirk-signal: Is this styled appropriately? e.g. we generally want to use the attributed flavor to make sure profile names are distinctly recognizable from something fetched from our own contacts.

charlesmchen-signal: I was wondering the same thing. I'll add a note to discuss with myles.

charlesmchen-signal: Talked with Myles.  Design isn't final, but I've made some changes per him.

SignalMessaging/contacts/OWSContactsManager.m
```diff
@@ -738,37 +738,49 @@ - (NSAttributedString *)attributedContactOrProfileNameForPhoneIdentifier:(NSStri
...
-        NSRange recipientIdRange = [numberAndProfileName rangeOfString:recipientId];
-        NSMutableAttributedString *attributedString =
-            [[NSMutableAttributedString alloc] initWithString:numberAndProfileName
-                                                   attributes:@{ NSFontAttributeName : secondaryFont }];
-        [attributedString addAttribute:NSFontAttributeName value:primaryFont range:recipientIdRange];
-
-        return [attributedString copy];
+        NSAttributedString *result =
+            [[NSAttributedString alloc] initWithString:recipientId attributes:primaryAttributes];
+        result = [result rtlSafeAppend:[[NSAttributedString alloc] initWithString:@" "] referenceView:[UIView new]];
```
charlesmchen-signal: IMO it's safer to use "RTL safe append" than to have translators juggle complicated and hard-to-read format string. WDYT?  If you approve, I'd like to remove all usage of `PROFILE_NAME_AND_PHONE_NUMBER_LABEL_FORMAT`.

charlesmchen-signal:
PTAL @michaelkirk

PR #3631, opened 2018-06-29
Tweak message contents.

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
```diff
@@ -76,25 +76,6 @@ - (void)configureWithConversationViewItem:(ConversationViewItem *)viewItem isOve
 
     [self configureLabelsWithConversationViewItem:viewItem];
 
-
-    // TODO: Constants
-    for (UIView *subview in @[
```
charlesmchen-signal: Remove shadows from message footers now that we have the gradient.

PR #3632, opened 2018-06-29
Charlesmchen/tweak quoted replies2

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -349,7 +349,7 @@ NS_ASSUME_NONNULL_BEGIN
 
             if (self.isQuotedReply) {
                 UIView *spacerView = [UIView containerView];
-                [spacerView autoSetDimension:ALDimensionHeight toSize:self.bodyMedaQuotedReplyVSpacing];
+                [spacerView autoSetDimension:ALDimensionHeight toSize:self.bodyMediaQuotedReplyVSpacing];
                 [spacerView setCompressionResistanceHigh];
                 [self.stackView addArrangedSubview:spacerView];
             }
@@ -606,7 +606,7 @@ NS_ASSUME_NONNULL_BEGIN
     return 2.f;
 }
 
-- (CGFloat)bodyMedaQuotedReplyVSpacing
+- (CGFloat)bodyMediaQuotedReplyVSpacing
 {
     return 8.f;
 }
@@ -1176,7 +1176,7 @@ NS_ASSUME_NONNULL_BEGIN
         }
 
         if (bodyMediaSize && quotedMessageSize && self.hasFullWidthMediaView) {
-            cellSize.height += self.bodyMedaQuotedReplyVSpacing;
+            cellSize.height += self.bodyMediaQuotedReplyVSpacing;
         }
     }
 
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -287,8 +287,6 @@ NS_ASSUME_NONNULL_BEGIN
         OWSAssert(quotedMessageSize);
         [self.viewConstraints addObject:[quotedMessageView autoSetDimension:ALDimensionHeight
                                                                      toSize:quotedMessageSize.CGSizeValue.height]];
-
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/Utils/MessageRecipientStatusUtils.swift
```diff
@@ -159,4 +159,24 @@ public class MessageRecipientStatusUtils: NSObject {
                                                               referenceView: referenceView)
         return status
     }
+
+    @objc
+    public class func description(forMessageReceiptStatus value: MessageReceiptStatus) -> String {
```
michaelkirk-signal: Is this just for logging?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -341,6 +346,13 @@ - (void)configureViews
             [self insertAnyTextViewsIntoStackView:textViews];
             [textViews removeAllObjects];
 
+            if (self.isQuotedReply) {
+                UIView *spacerView = [UIView containerView];
+                [spacerView autoSetDimension:ALDimensionHeight toSize:self.bodyMedaQuotedReplyVSpacing];
```
michaelkirk-signal: > bodyMed_a_QuotedReplyVSpacing

typo

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -532,10 +544,10 @@ - (BOOL)hasBottomFooter
     return !self.viewItem.shouldHideFooter && !shouldFooterOverlayMedia;
 }
 
-- (void)insertAnyTextViewsIntoStackView:(NSArray<UIView *> *)textViews
+- (BOOL)insertAnyTextViewsIntoStackView:(NSArray<UIView *> *)textViews
```
michaelkirk-signal: TIOLI: I think the code that calls this method could be clearer if this method returns a `UIStackView?` instead of a BOOL and the caller optionally does the insertion.

At that point, something like `buildTextStackView:` would probably be a better name.

Signal/src/ViewControllers/ConversationView/Cells/OWSQuotedMessageView.m
```diff
@@ -147,114 +220,47 @@ - (void)createContents
...
-        [stripeAndTextContainer addSubview:quotedTextLabel];
-        [quotedTextLabel autoPinTopToSuperviewMarginWithInset:self.quotedReplyStripeVExtension];
-        [quotedTextLabel autoPinBottomToSuperviewMarginWithInset:self.quotedReplyStripeVExtension];
-        [quotedTextLabel autoPinLeadingToTrailingEdgeOfView:quoteStripView offset:self.quotedReplyStripeHSpacing];
-        [quotedTextLabel autoPinTrailingToSuperviewMargin];
-        [quotedTextLabel setContentHuggingLow];
-        [quotedTextLabel setCompressionResistanceLow];
+        [hStackView addArrangedSubview:quotedAttachmentView];
+    } else {
+        UIView *emptyView = [UIView containerView];
```
michaelkirk-signal: Why do we need a zero space empty view in the stack?

Signal/src/ViewControllers/Utils/MessageRecipientStatusUtils.swift
charlesmchen-signal: Yup.  I try to keep around formatting methods for any enums that we might want to log, because... I often find myself (temporarily) logging them.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
charlesmchen-signal: Fixed.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
charlesmchen-signal: It's a tough choice.  `configureViews` is already _huge_.  If I get a chance, I'd like to decompose it a bit, but it's always going to be awkwardly large.  I'm wary of making it even longer.

Signal/src/ViewControllers/ConversationView/Cells/OWSQuotedMessageView.m
charlesmchen-signal: I've added a comment.

PR #3635, opened 2018-06-30
tweak navbar

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/SignalMessaging/ViewControllers/OWSTableViewController.m
+++ b/SignalMessaging/ViewControllers/OWSTableViewController.m
@@ -457,9 +457,7 @@ NSString *const kOWSTableCellIdentifier = @"kOWSTableCellIdentifier";
     [self.tableView applyScrollViewInsetsFix];
 
     [self.view addSubview:self.tableView];
-    [self.tableView autoPinWidthToSuperview];
-    [self.tableView autoPinToTopLayoutGuideOfViewController:self withInset:0];
-    [self.tableView autoPinEdgeToSuperviewEdge:ALEdgeBottom];
+    [self.tableView autoPinEdgesToSuperviewEdges];
 
     [self.tableView registerClass:[UITableViewCell class] forCellReuseIdentifier:kOWSTableCellIdentifier];
 }
--- a/SignalMessaging/utils/UIUtil.h
+++ b/SignalMessaging/utils/UIUtil.h
@@ -20,8 +20,6 @@ typedef void (^completionBlock)(void);
 + (void)applyRoundedBorderToImageView:(UIImageView *)imageView;
 + (void)removeRoundedBorderToImageView:(UIImageView *__strong *)imageView;
 
-+ (completionBlock)modalCompletionBlock;
-+ (void)applyDefaultSystemAppearence;
-+ (void)applySignalAppearence;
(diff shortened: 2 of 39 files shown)


SignalMessaging/ViewControllers/OWSTableViewController.m
```diff
@@ -457,9 +457,7 @@ - (void)loadView
     [self.tableView applyScrollViewInsetsFix];
 
     [self.view addSubview:self.tableView];
-    [self.tableView autoPinWidthToSuperview];
-    [self.tableView autoPinToTopLayoutGuideOfViewController:self withInset:0];
-    [self.tableView autoPinEdgeToSuperviewEdge:ALEdgeBottom];
+    [self.tableView autoPinEdgesToSuperviewEdges];
```
charlesmchen-signal: This seems like something we'll want to test in the 10+ places we use this vc on iOS 9, 10 and 11 and on iPhone X.

SignalMessaging/utils/UIUtil.h
```diff
@@ -20,8 +20,6 @@ typedef void (^completionBlock)(void);
 + (void)applyRoundedBorderToImageView:(UIImageView *)imageView;
 + (void)removeRoundedBorderToImageView:(UIImageView *__strong *)imageView;
 
-+ (completionBlock)modalCompletionBlock;
-+ (void)applyDefaultSystemAppearence;
-+ (void)applySignalAppearence;
++ (void)setupSignalAppearence;
```
charlesmchen-signal: So we _don't_ need to call this in the share extension?

SignalMessaging/utils/UIUtil.m
```diff
@@ -23,42 +24,26 @@ + (void)removeRoundedBorderToImageView:(UIImageView *__strong *)imageView
...
-
-+ (void)applyDefaultSystemAppearence
++ (void)setupSignalAppearence
 {
+    UINavigationBar.appearance.barTintColor = UIColor.ows_navbarBackgroundColor;
+    UINavigationBar.appearance.tintColor = UIColor.ows_navbarIconColor;
+    UIToolbar.appearance.barTintColor = UIColor.ows_navbarBackgroundColor;
+    UIToolbar.appearance.tintColor = UIColor.ows_navbarIconColor;
+
+    UIBarButtonItem.appearance.tintColor = UIColor.ows_navbarIconColor;
```
charlesmchen-signal: So in terms of not needing the default system appearance, we need to test:

* Camera roll / media picker.
* Email composer.
* SMS Message composer.
* Contact editor.
* Document Picker.
* Sharing UI.

Right?

SignalMessaging/utils/UIUtil.h
michaelkirk-signal: Right.

SignalMessaging/utils/UIUtil.m
michaelkirk-signal: Yup, I'll add this to the test plan.

SignalMessaging/ViewControllers/OWSTableViewController.m
michaelkirk-signal: I'll add this in the test plan.

PR #3642, opened 2018-07-02
Small media corners.

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -342,7 +342,7 @@ class MediaGalleryViewController: OWSNavigationController, MediaGalleryDataSourc
         detailView.backgroundColor = .clear
         self.view.backgroundColor = .clear
 
-        self.presentationView.layer.cornerRadius = kOWSMessageCellCornerRadius
+        self.presentationView.layer.cornerRadius = kOWSMessageCellCornerRadius_Large
```
michaelkirk-signal: Is this always correct? Can't the media bubbles sometimes have smaller, and not even symmetrical corder rounding?

This isn't a huge point, so I'm OK with merging this as is to converge closer to the design, but can you make sure this outstanding work is accounted for somewhere?

PR #3643, opened 2018-07-02
Fix quoted reply margin.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
@@ -47,7 +47,7 @@ static const CGFloat ConversationInputToolbarBorderViewHeight = 0.5;
 
 #pragma mark -
 
-@property (nonatomic, nullable) QuotedReplyPreview *quotedMessagePreview;
+@property (nonatomic, nullable) UIView *quotedMessagePreview;
 
 #pragma mark - Voice Memo Recording UI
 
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
```diff
@@ -289,6 +292,12 @@ - (void)setQuotedReply:(nullable OWSQuotedReplyModel *)quotedReply
 
     // TODO animate
     [self.contentStackView insertArrangedSubview:self.quotedMessagePreview atIndex:0];
+    self.contentStackView.layoutMargins = UIEdgeInsetsMake(self.quotedMessageTopMargin, 0, 0, 0);
```
michaelkirk-signal: So, usually we **don't** want the `contentStackView` to have a topMargin, but when it has a `quotedMessagePreview` we **do** want it to have a top margin, right?

It might be clearer to leave the contentStackView margins alone, and wrap the quotedMessagePreview in a view that includes the top margins. 

That way we don't have the inexplicable line 131 and we don't have to worry about cleaning it back up on line 311

Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
```diff
@@ -127,6 +128,8 @@ - (void)createContents
     _composeContainer = [UIView containerView];
     _contentStackView = [[UIStackView alloc] initWithArrangedSubviews:@[ _composeContainer ]];
     _contentStackView.axis = UILayoutConstraintAxisVertical;
+    _contentStackView.layoutMargins = UIEdgeInsetsZero;
```
michaelkirk-signal: I see now why we're doing this, but by itself this is kind of inexplicable, since enabling layoutMarginsRelativeArrangement with layoutMargins of 0 is a noop. See my comment below...

Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
charlesmchen-signal: Sure, that's simpler. Done.

PR #3647, opened 2018-07-03
misc cleanup

michaelkirk-signal:
PTAL @charlesmchen


Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -254,6 +254,7 @@ class MessageDetailViewController: OWSViewController, MediaGalleryDataSourceDele
                     // Table view cells don't layout properly outside the
                     // context of a table view.
                     let cellView = ContactCellView()
+                    cellView.isLayoutMarginsRelativeArrangement = true
```
charlesmchen-signal: I suggest `OWSAssert(!cellView.isLayoutMarginsRelativeArrangement);`.  Otherwise we might end up overwriting margins we set in that view if we later changed it.

michaelkirk-signal: Good idea, done.

PR #3650, opened 2018-07-03
Mkirk/fix initial contact group sync

michaelkirk-signal:
The server caches your device list on the websocket, so sending on the
websocket to a just-linked device will always fail. We could close/open the
websocket, but that might be disruptive in it's own way. Instead we'll closely
mirror the Android approach, where WebSocket sends are attempted only one time,
and failure is handled by falling back to the original REST approach.

So note: we don't do any special handling of failures on the websocket
(409/410). We simply retry it with REST which will handle the 409/410/etc.

Consequently, we don't want to decrement our retry count for websocket sends.

PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Messages/OWSMessageSender.m
+++ b/SignalServiceKit/src/Messages/OWSMessageSender.m
@@ -949,15 +949,13 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
             });
 
             return;
-        } else if (mayHaveLinkedDevices) {
+        } else if (mayHaveLinkedDevices && !hasDeviceMessages) {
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -946,15 +949,13 @@ - (void)sendMessageToService:(TSOutgoingMessage *)message
             });
 
             return;
-        } else if (mayHaveLinkedDevices) {
+        } else if (mayHaveLinkedDevices && !hasDeviceMessages) {
             // We may have just linked a new secondary device which is not yet reflected in
```
michaelkirk-signal: small logging bug here - previously we were hitting this in the happy path.

PR #3653, opened 2018-07-04
Tweak sync

michaelkirk-signal:
PTAL @charlesmchen 

tweak sync to desktop


Signal/src/Jobs/ConversationConfigurationSyncOperation.swift
```diff
@@ -0,0 +1,118 @@
...
+
+        self.messageSender.enqueueTemporaryAttachment(attachmentDataSource,
+                                                      contentType: OWSMimeTypeApplicationOctetStream,
+                                                      in: syncMessage,
+                                                      success: {
+                                                        self.reportSuccess()
+        },
+                                                      failure: { error in
+                                                        self.reportError(error)
+        })
```
charlesmchen-signal: NBD but it seems like we could DRY up these two sync methods a bit.

Signal/src/ViewControllers/DebugUI/DebugUISyncMessages.m
```diff
@@ -99,8 +99,13 @@ + (void)sendContactsSyncMessage
         [[OWSSyncContactsMessage alloc] initWithSignalAccounts:self.contactsManager.signalAccounts
                                                identityManager:self.identityManager
                                                 profileManager:self.profileManager];
-    DataSource *dataSource =
-        [DataSourceValue dataSourceWithSyncMessage:[syncContactsMessage buildPlainTextAttachmentData]];
+    __block DataSource *dataSource;
+    [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
```
charlesmchen-signal: NBD I'm trying to avoid these redundant `_Nonnull`.

michaelkirk-signal: Yeah... ObjC nullability handling is pretty shaky when it comes to blocks, so I don't mind the redundancy. In same cases it's helpful apparently.


"redundant" nonnull provides a helpful warning

screen shot 2018-07-05 at 1 36 51 pm

removing "redundant" nonnull produces no warning.

screen shot 2018-07-05 at 1 37 23 pm

This is all within an NS_ASSUME_NONULL_BEGIN block. 🤷‍♀️

SignalServiceKit/src/Devices/OWSContactsOutputStream.m
```diff
@@ -67,14 +68,8 @@ - (void)writeSignalAccount:(SignalAccount *)signalAccount
     // legacy client "not specifying".
     [contactBuilder setExpireTimer:0];
 
-    TSContactThread *_Nullable contactThread = [TSContactThread getThreadWithContactId:signalAccount.recipientId];
```
michaelkirk-signal: removed this sneaky transaction since we're already in a transaction.

SignalServiceKit/src/Messages/DeviceSyncing/OWSSyncContactsMessage.m
```diff
@@ -87,12 +87,24 @@ - (NSData *)buildPlainTextAttachmentDataWithTransaction:(YapDatabaseReadWriteTra
             [self.identityManager recipientIdentityForRecipientId:signalAccount.recipientId];
         NSData *_Nullable profileKeyData = [self.profileManager profileKeyDataForRecipientId:signalAccount.recipientId];
 
-        TSContactThread *contactThread = [TSContactThread getOrCreateThreadWithContactId:signalAccount.recipientId transaction:transaction];
+
+        OWSDisappearingMessagesConfiguration *_Nullable disappearingMessagesConfiguration;
+        NSString *conversationColorName;
+        
+        TSContactThread *_Nullable contactThread = [TSContactThread getThreadWithContactId:signalAccount.recipientId transaction:transaction];
+        if (contactThread) {
```
michaelkirk-signal: We now use a ReadTransaction which means contactThread can be nil. This is the primary change.

PR #3655, opened 2018-07-05
Use sending queue in websocket send failover to REST.

charlesmchen-signal:
PTAL @michaelkirk


SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -989,21 +989,23 @@ - (void)sendMessageToService:(TSOutgoingMessage *)message
...
-                // e.g. after linking a new device, sync messages will fail until the websocket re-opens.
-                [self sendMessageToService:message
-                                  recipient:recipient
-                                     thread:thread
-                                   attempts:remainingAttemptsParam
-                    useWebsocketIfAvailable:NO
-                                    success:successHandler
-                                    failure:failureHandler];
+                dispatch_async([OWSDispatch sendingQueue], ^{
+                    DDLogDebug(
```
charlesmchen-signal: Simpler to do this logging here, right?

PR #3656, opened 2018-07-05
Charlesmchen/tweak send failed

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -158,13 +170,42 @@ - (void)loadForDisplayWithTransaction:(YapDatabaseReadTransaction *)transaction
...
-                                                     withInset:self.conversationStyle.gutterLeading
-                                                      relation:NSLayoutRelationGreaterThanOrEqual],
-            [self.messageBubbleView autoPinEdgeToSuperviewEdge:ALEdgeTrailing
-                                                     withInset:self.conversationStyle.gutterTrailing],
-        ]];
+        if (self.shouldHavesendFailureBadge) {
+            self.sendFailureBadgeView = [UIImageView new];
+            self.sendFailureBadgeView.image =
+                [self.sendFailureBadge imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
+            self.sendFailureBadgeView.tintColor = [UIColor ows_destructiveRedColor];
```
michaelkirk-signal: No change needed now, but WRT earlier conversations, maybe we should default to more "component" named colors as opposed to spectrum, e.g. "failureBadgeColor" vs. "redColor"

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -184,6 +225,24 @@ - (void)loadForDisplayWithTransaction:(YapDatabaseReadTransaction *)transaction
     }
 }
 
+- (UIImage *)sendFailureBadge
+{
+    UIImage *image = [UIImage imageNamed:@"message_send_failed"];
+    OWSAssert(image);
+    OWSAssert(image.size.width == self.sendFailureBadgeSize && image.size.height == self.sendFailureBadgeSize);
```
michaelkirk-signal: nice canary.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
```diff
@@ -141,13 +144,32 @@ - (void)animateSpinningIcon
...
+
 - (void)configureLabelsWithConversationViewItem:(ConversationViewItem *)viewItem
 {
     OWSAssert(viewItem);
 
     [self configureFonts];
 
-    self.timestampLabel.text = [DateUtil formatTimestampAsTimeShort:viewItem.interaction.timestamp];
+    if ([self isFailedOutgoingMessage:viewItem]) {
+        self.timestampLabel.text
```
michaelkirk-signal: maybe: s/timestampLabel/deliveryStatusLabel?

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -10,7 +10,8 @@
 /* Label for 'invite' button in contact view. */
 "ACTION_INVITE" = "Invite to Signal";
 
-/* Label for 'sent message' button in contact view. */
+/* Label for 'sent message' button in contact view.
```
michaelkirk-signal: s/sent/send

And this comment doesn't provide very useful context, maybe "Button label on a contact share message"

Also might want to introduce a unique string, since we're using this in multiple places, or pull into CommonStrings

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1217,6 +1218,9 @@
 /* message status if message delivery to a recipient is skipped. We skip delivering group messages to users who have left the group or unregistered their Signal account. */
 "MESSAGE_STATUS_RECIPIENT_SKIPPED" = "Skipped";
 
+/* Label indicating that a message failed to send. */
+"MESSAGE_STATUS_SEND_FAILED" = "Send Failed";
```
michaelkirk-signal: This is in the footer, but the translation comments show "MESSAGE_STATUS_FAILED" as a message footer. Are they used in slightly different contexts? Or can one of them be deleted?

SignalMessaging/categories/UIColor+OWS.m
```diff
@@ -79,7 +79,7 @@ + (UIColor *)ows_reminderDarkYellowColor
 
 + (UIColor *)ows_destructiveRedColor
 {
-    return [UIColor colorWithRed:0.98639106750488281 green:0.10408364236354828 blue:0.33135244250297546 alpha:1.f];
+    return [UIColor colorWithRed:255.f / 255.f green:38.f / 255.f blue:31.f / 255.f alpha:1.0f];
```
michaelkirk-signal: For consistency, can we converge on using the hex constructor whenever reasonable?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -131,6 +134,15 @@ - (BOOL)isOutgoing
     return self.viewItem.interaction.interactionType == OWSInteractionType_OutgoingMessage;
 }
 
+- (BOOL)shouldHavesendFailureBadge
```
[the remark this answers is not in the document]

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: We only use this color once in this class, so I wasn't going to make a method - hence there's no local name for this color.  I like that `ows_destructiveRedColor` alludes to the fact that this is used for "errors & destructive actions" - although we could improve it.  I don't mind that the color "red" is in the name, since this color is the same in all themes.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: I'm trying to do this everywhere now.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
charlesmchen-signal: It's a tough choice, but I'm inclined to leave it since this label will have a timestamp 99% of teh time - the "sender failure" case is the only exception.

Signal/translations/en.lproj/Localizable.strings
charlesmchen-signal: > s/sent/send

Done.

> Also might want to introduce a unique string, since we're using this in multiple places, or pull into CommonStrings

I'm inclined not to introduce a new string, given that both places we're using have very similar semantics (Send a text message to a user) and the exact same copy, in order to minimize the burden on our translators and leverage our existing translations.  I wish Transifex defaulted to existing translations for strings with the same EN copy.

SignalMessaging/categories/UIColor+OWS.m
charlesmchen-signal: I'd like to use hex as a default, and I'll change it here.

However for non-palette colors delivered with Zeplin, I'm inclined to use this representation so it can be easily verified against the design.

PR #3657, opened 2018-07-05
Fix contact cell layout.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessageDetailViewController.swift
+++ b/Signal/src/ViewControllers/MessageDetailViewController.swift
@@ -254,14 +254,15 @@ class MessageDetailViewController: OWSViewController, MediaGalleryDataSourceDele
                     // Table view cells don't layout properly outside the
                     // context of a table view.
                     let cellView = ContactCellView()
-                    // canary in case we later add margins in the ContactCellView initializer.
-                    assert(!cellView.isLayoutMarginsRelativeArrangement)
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -254,11 +254,14 @@ class MessageDetailViewController: OWSViewController, MediaGalleryDataSourceDele
...
-                    cellView.layoutMargins = UIEdgeInsets(top: 8, left: 20, bottom: 8, right: 20)
                     // We use the "short" status message to avoid being redundant with the section title.
                     cellView.accessoryMessage = shortStatusMessage
                     cellView.configure(withRecipientId: recipientId, contactsManager: self.contactsManager)
-                    groupRows.append(cellView)
+                    let wrapper = UIView()
+                    wrapper.layoutMargins = UIEdgeInsets(top: 8, left: 20, bottom: 8, right: 20)
+                    wrapper.addSubview(cellView)
+                    cellView.autoPinToSuperviewMargins()
+                    groupRows.append(wrapper)
```
charlesmchen-signal: After some consideration, I favor using more wrappers.  Using the cell view's layout margins assumes that its internal layout respects margins.  Better to use composition and keep the code loosely coupled.

michaelkirk-signal: I agree.

PR #3658, opened 2018-07-05
Charlesmchen/relative timestamps

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -1343,6 +1343,7 @@ NS_ASSUME_NONNULL_BEGIN
 
     if (self.hasBottomFooter) {
         CGSize footerSize = [self.footerView measureWithConversationViewItem:self.viewItem];
+        footerSize.width = MIN(footerSize.width, self.conversationStyle.maxMessageWidth);
         [textViewSizes addObject:[NSValue valueWithCGSize:footerSize]];
     }
 
--- a/Signal/src/util/DateUtil.m
+++ b/Signal/src/util/DateUtil.m
@@ -4,6 +4,7 @@
 
 #import "DateUtil.h"
 #import <SignalMessaging/NSString+OWS.h>
+#import <SignalMessaging/OWSFormat.h>
 #import <SignalServiceKit/NSDate+OWS.h>
 
 NS_ASSUME_NONNULL_BEGIN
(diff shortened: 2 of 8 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -1312,6 +1312,7 @@ - (CGSize)measureSize
 
     if (self.hasBottomFooter) {
         CGSize footerSize = [self.footerView measureWithConversationViewItem:self.viewItem];
+        footerSize.width = MIN(footerSize.width, self.conversationStyle.maxMessageWidth);
```
michaelkirk-signal: can this logic go into the footerView?

charlesmchen-signal: I'll do this before I merge, but I don't want to conflict with unmerged PRs.

charlesmchen-signal:
PTAL @michaelkirk

Signal/src/util/DateUtil.m
```diff
@@ -187,19 +206,132 @@ + (NSString *)formatDateShort:(NSDate *)date
...
+
+    NSInteger yearsDiff = [self yearsFromFirstDate:date toSecondDate:nowDate];
+    NSInteger daysDiff = [self daysFromFirstDate:date toSecondDate:nowDate];
+    NSInteger minutesDiff
+        = MAX(0, [[calendar components:NSCalendarUnitMinute fromDate:date toDate:nowDate options:0] minute]);
+    NSInteger hoursDiff
+        = MAX(0, [[calendar components:NSCalendarUnitHour fromDate:date toDate:nowDate options:0] hour]);
+
+    NSString *result;
+    if (yearsDiff > 0) {
```
michaelkirk-signal: There are some existing methods that I think could clean this up.

e.g.     

if (![self dateIsThisYear:date]) {

} else if (![self dateIsThisWeek:date]) {

} else if (![self isOlderThanToday:date]) {

}

Signal/src/util/DateUtil.m
```diff
@@ -187,19 +206,132 @@ + (NSString *)formatDateShort:(NSDate *)date
...
+    NSCalendar *calendar = [NSCalendar currentCalendar];
+
+    NSInteger hoursDiff
+        = MAX(0, [[calendar components:NSCalendarUnitHour fromDate:date toDate:nowDate options:0] hour]);
+    return hoursDiff < 1;
+}
+
++ (NSString *)exemplaryNowTimeFormat
+{
+    return NSLocalizedString(@"DATE_NOW", @"The present; the current time.").uppercaseString;
```
michaelkirk-signal: This probably deserves a comment. e.g. why "Now" instead of "June 3, 2017"

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -469,7 +469,7 @@ - (void)ensureThumbnail
     }
 
     if (result == nil) {
-        OWSFail(@"%@ Unable to build thumbnail for attachmentId: %@", self.logTag, self.uniqueId);
+        DDLogError(@"%@ Unable to build thumbnail for attachmentId: %@", self.logTag, self.uniqueId);
```
michaelkirk-signal: What's this change about? I haven't yet seen this happen except in the case of bad data left over from a bug, and would prefer to keep it.

Signal/src/util/DateUtil.m
charlesmchen-signal: I've added a comment in the header.

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
charlesmchen-signal: Reverted. I didn't mean to include this.

Signal/src/util/DateUtil.m
charlesmchen-signal: Yes, we could do it either way.  I decided to be defensive here to ensure that we treat future dates as "now".  These dates methods are subtle.  I was debating changing the semantics of the existing methods but decided not to - at least, not yet.

PR #3660, opened 2018-07-05
Charlesmchen/tweak home view

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewCell.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewCell.m
@@ -198,11 +198,11 @@ NS_ASSUME_NONNULL_BEGIN
         = (overrideDate ? [self stringForDate:overrideDate] : [self stringForDate:thread.lastMessageDate]);
 
     if (hasUnreadMessages && overrideSnippet == nil) {
-        self.dateTimeLabel.textColor = [UIColor ows_blackColor];
-        self.dateTimeLabel.font = self.unreadFont.ows_mediumWeight;
+        self.dateTimeLabel.textColor = [UIColor ows_light90Color];
+        self.dateTimeLabel.font = self.dateTimeFont.ows_mediumWeight;
     } else {
-        self.dateTimeLabel.textColor = [UIColor lightGrayColor];
-        self.dateTimeLabel.font = self.unreadFont;
+        self.dateTimeLabel.textColor = [UIColor ows_light60Color];
+        self.dateTimeLabel.font = self.dateTimeFont;
     }
 
     NSUInteger unreadCount = thread.unreadCount;
@@ -303,7 +303,7 @@ NS_ASSUME_NONNULL_BEGIN
                                                           @"A label for conversations with blocked users.")
                                            attributes:@{
                                                NSFontAttributeName : self.snippetFont.ows_mediumWeight,
-                                               NSForegroundColorAttributeName : [UIColor ows_blackColor],
+                                               NSForegroundColorAttributeName : [UIColor ows_light90Color],
                                            }]];
     } else {
         if ([thread isMuted]) {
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/HomeView/HomeViewCell.m
```diff
@@ -95,35 +87,69 @@ - (void)commontInit
...
+
     self.topRowView = [[UIStackView alloc] initWithArrangedSubviews:@[
         self.nameLabel,
         self.dateTimeLabel,
+        self.messageStatusView,
     ]];
     self.topRowView.axis = UILayoutConstraintAxisHorizontal;
-    self.topRowView.alignment = UIStackViewAlignmentLastBaseline;
-    [self.payloadView addArrangedSubview:self.topRowView];
+    self.topRowView.alignment = UIStackViewAlignmentCenter;
```
michaelkirk-signal: Although these will look *very* close, we want the Name, Date, unread indicator label to be aligned by baseline, not vertical center.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
```diff
@@ -236,28 +266,54 @@ - (void)configureWithThread:(ThreadViewModel *)thread
...
-            // Vertically, badge is positioned vertically by aligning it's label *subview's* baseline.
-            // This allows us a single visual baseline of text across the top row across [name, dateTime,
-            // optional(unread count)]
-            [self.unreadLabel autoAlignAxis:ALAxisBaseline toSameAxisOfView:self.dateTimeLabel]
-        ]];
+    } else {
+        UIImage *_Nullable statusIndicatorImage = nil;
+        UIColor *messageStatusViewTintColor = textColor;
+        BOOL shouldAnimateStatusIcon = NO;
+        if (overrideSnippet == nil && [self.thread.lastMessageForInbox isKindOfClass:[TSOutgoingMessage class]]) {
```
michaelkirk-signal: overrideSnippet is always nil at this point based on line 233, should this be in assert?

Signal/src/ViewControllers/HomeView/HomeViewCell.m
```diff
@@ -236,28 +266,54 @@ - (void)configureWithThread:(ThreadViewModel *)thread
...
+            TSOutgoingMessage *outgoingMessage = (TSOutgoingMessage *)self.thread.lastMessageForInbox;
+
+            MessageReceiptStatus messageStatus =
+                [MessageRecipientStatusUtils recipientStatusWithOutgoingMessage:outgoingMessage];
+            switch (messageStatus) {
```
michaelkirk-signal: TIOLI: This is a pretty large block of code that seems like it would be largely redundant with what we'd want to use in the messages VC. Can you extract this into something like an OutgoingMessageStatusView?

SignalServiceKit/src/Contacts/TSThread.m
```diff
@@ -246,8 +246,10 @@ - (void)markAllAsReadWithTransaction:(YapDatabaseReadWriteTransaction *)transact
     OWSAssert([self unseenMessagesWithTransaction:transaction].count < 1);
 }
 
-- (TSInteraction *)lastInteractionForInboxWithTransaction:(YapDatabaseReadTransaction *)transaction
+- (nullable TSInteraction *)lastInteractionForInboxWithTransaction:(YapDatabaseReadTransaction *)transaction
```
michaelkirk-signal: good catch

Signal/src/ViewControllers/HomeView/HomeViewCell.m
charlesmchen-signal: Done.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
charlesmchen-signal: Good point, removed.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
charlesmchen-signal: They're _just_ different enough that IMO it'd be more awkward to de-dupe.

charlesmchen-signal:
PTAL @michaelkirk

PR #3661, opened 2018-07-05
tweak attachment icon

michaelkirk-signal:
PTAL @charlesmchen


Signal/src/ViewControllers/ConversationView/ConversationInputToolbar.m
```diff
@@ -119,7 +119,7 @@ - (void)createContents
     [self.attachmentButton addTarget:self
                               action:@selector(attachmentButtonPressed)
                     forControlEvents:UIControlEventTouchUpInside];
-    UIImage *attachmentImage = [UIImage imageNamed:@"btnAttachments--blue"];
```
charlesmchen-signal: Are we able to remove `btnAttachments--blue` now?

michaelkirk-signal: oh yeah! Will do.

PR #3666, opened 2018-07-06
Mkirk/fixup bubble shape

michaelkirk-signal:
tweaking our corner style, radius, height to more closely match the spec.

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
@@ -161,19 +161,52 @@ const CGFloat kOWSMessageCellCornerRadius_Small = 2;
     CGFloat bottomRounding
         = (useSmallCorners_Bottom ? kOWSMessageCellCornerRadius_Small : kOWSMessageCellCornerRadius_Large);
 
+    const CGFloat topAngle = 3.0f * M_PI / 2.0f;
+    const CGFloat rightAngle = 0.0f;
+    const CGFloat bottomAngle = M_PI / 2.0f;
+    const CGFloat leftAngle = M_PI;
+
     [bezierPath moveToPoint:CGPointMake(bubbleLeft + topRounding, bubbleTop)];
+
+    // top line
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
```diff
@@ -161,19 +161,52 @@ + (UIBezierPath *)maskPathForSize:(CGSize)size
     CGFloat bottomRounding
         = (useSmallCorners_Bottom ? kOWSMessageCellCornerRadius_Small : kOWSMessageCellCornerRadius_Large);
 
+    const CGFloat topAngle = 3.0f * M_PI / 2.0f;
+    const CGFloat rightAngle = 0.0f;
+    const CGFloat bottomAngle = M_PI / 2.0f;
+    const CGFloat leftAngle = M_PI;
```
charlesmchen-signal: This is fine, but FYI CG has a bunch of pre-existing but confusingly-named constants like `M_PI_2` for these purposes.

Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
```diff
@@ -161,19 +161,52 @@ + (UIBezierPath *)maskPathForSize:(CGSize)size
...
+
     [bezierPath moveToPoint:CGPointMake(bubbleLeft + topRounding, bubbleTop)];
+
+    // top line
     [bezierPath addLineToPoint:CGPointMake(bubbleRight - topRounding, bubbleTop)];
```
charlesmchen-signal: I wonder if the `addLineToPoint` calls are now redundant, e.g. whether `UIBezierPath` will "close" the shape between the arcs.

michaelkirk-signal: Hmm.. it appears that it does. I've made this change.

Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleView.m
michaelkirk-signal: updated, thanks!

PR #3677, opened 2018-07-09
Fix breakage in message details view.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.h
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.h
@@ -68,6 +68,8 @@ typedef NS_ENUM(NSUInteger, OWSMessageGestureLocation) {
 
 @property (nonatomic, weak) id<OWSMessageBubbleViewDelegate> delegate;
 
+@property (nonatomic) BOOL shouldSuppressHeaderFooterAndRounding;
+
 - (instancetype)init NS_UNAVAILABLE;
 
 - (instancetype)initWithFrame:(CGRect)frame NS_DESIGNATED_INITIALIZER;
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -640,6 +648,10 @@ - (BOOL)canFooterOverlayMedia
 
 - (BOOL)hasBottomFooter
 {
+    if (self.shouldSuppressHeaderFooterAndRounding) {
```
michaelkirk-signal: I thought we wanted the bubble to look more or less exactly the same in the message detail view as in the conversation.

I think it's a good call to avoid sharp corners in the detail view, but why don't we want the header and footer in message details view?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.h
```diff
@@ -68,6 +68,8 @@ typedef NS_ENUM(NSUInteger, OWSMessageGestureLocation) {
 
 @property (nonatomic, weak) id<OWSMessageBubbleViewDelegate> delegate;
 
+@property (nonatomic) BOOL shouldSuppressHeaderFooterAndRounding;
```
michaelkirk-signal: I find that the "mode flag" pattern often comes back to bite, especially as they start to interact.

What about creating a new ConversationViewItem, and setting its `isFirstInCluster` and `isLastInCluster` to true?

michaelkirk-signal: Sorry - I meant to say that this is TIOLI feedback. It might not be worth changing.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
michaelkirk-signal: I think we also always want to show the footer on the message detail screen.

PR #3682, opened 2018-07-09
Mkirk/date breaks

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
@@ -9,6 +9,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+const CGFloat OWSMessageCellDateHeaderVMargin = 23;
+
 @interface OWSMessageCell ()
 
 // The nullable properties are created as needed.
@@ -59,7 +61,7 @@ NS_ASSUME_NONNULL_BEGIN
     self.dateHeaderLabel.textColor = [UIColor ows_light60Color];
 
     self.dateHeaderView = [UIView new];
-    self.dateHeaderView.layoutMargins = UIEdgeInsetsMake(self.dateHeaderVMargin, 0, self.dateHeaderVMargin, 0);
+    self.dateHeaderView.layoutMargins = UIEdgeInsetsMake(0, 0, OWSMessageCellDateHeaderVMargin, 0);
     [self.dateHeaderView addSubview:self.dateHeaderLabel];
     [self.dateHeaderLabel autoPinToSuperviewMargins];
 
@@ -385,16 +387,11 @@ NS_ASSUME_NONNULL_BEGIN
     return cellSize;
 }
 
-- (CGFloat)dateHeaderVMargin
-{
-    return 23.f;
-}
-
 - (CGFloat)dateHeaderHeight
 {
     if (self.viewItem.shouldShowDate) {
         CGFloat textHeight = self.dateHeaderFont.lineHeight;
-        return (CGFloat)ceil(textHeight + self.dateHeaderVMargin * 2);
+        return (CGFloat)ceil(textHeight + OWSMessageCellDateHeaderVMargin);
     } else {
         return 0.f;
     }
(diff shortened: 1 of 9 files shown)


SignalMessaging/categories/UIFont+OWS.m
```diff
@@ -72,14 +72,19 @@ + (UIFont *)ows_dynamicTypeHeadlineFont
...
 
-+ (UIFont *)ows_dynamicTypeSubheadlineFont
++ (UIFont *)ows_dynamicTypeBodyFont
 {
-    return [UIFont preferredFontForTextStyle:UIFontTextStyleSubheadline];
+    return [UIFont preferredFontForTextStyle:UIFontTextStyleBody];
 }
 
-+ (UIFont *)ows_dynamicTypeBodyFont
++ (UIFont *)ows_dynamicTypeCalloutFont
```
michaelkirk-signal: No significant changes to this file - I mistakenly thought I needed to use this font, but as long as it's available, we might as well include it. 

I also tweaked the ordering to be monotonically decreasing in line with the HIG.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -425,21 +385,16 @@ - (CGSize)cellSizeWithTransaction:(YapDatabaseReadTransaction *)transaction
...
+    return 23.f;
 }
 
 - (CGFloat)dateHeaderHeight
 {
     if (self.viewItem.shouldShowDate) {
-        CGFloat textHeight = self.dateHeaderFont.capHeight;
-        return (CGFloat)ceil(self.dateHeaderStrokeThickness + textHeight + self.dateHeaderBottomMargin);
+        CGFloat textHeight = self.dateHeaderFont.lineHeight;
+        return (CGFloat)ceil(textHeight + self.dateHeaderVMargin * 2);
```
charlesmchen-signal: I thought we were converging on an approach wherein the cell contents began at the cell's edge and intra-cell spacing was a concern of `ConverasationViewItem`?

Signal/src/util/DateUtil.m
```diff
@@ -25,6 +25,64 @@ + (NSDateFormatter *)dateFormatter {
...
+
++ (NSDateFormatter *)dateBreakThisYearDateFormatter
+{
+    static NSDateFormatter *formatter;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        formatter = [NSDateFormatter new];
+        formatter.locale = [NSLocale currentLocale];
+        // Tue, Jun 6
+        formatter.dateFormat = @"EE, MMM d";
```
charlesmchen-signal: I wonder if we should use this kind of complicated date format - it won't necessarily localize well.

SignalMessaging/categories/UIFont+OWS.m
charlesmchen-signal: I thought the design was to only use four font styles?  When would we use "call out"?

michaelkirk-signal: We don't currently use it anywhere. We have literally every other font from the hig in this class, including ones we don't use. After accidentally adding it I thought I'd keep it since we might use it someday, but I didn't think it would be controversial.

Signal/src/util/DateUtil.m
michaelkirk-signal: Yeah, same issue with the message footer stuff we merged yesterday. Suggestion?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
michaelkirk-signal: Ok - I've done this, but it's a little awkward because the design wants symmetrical spacing above and below this header. 

So now, the cell adds the bottom spacing while the conversationViewItem adds the equivalent amount of top spacing.

michaelkirk-signal:
PTAL @charlesmchen

SignalMessaging/categories/UIFont+OWS.m
charlesmchen-signal: Not controversial, just asking.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
charlesmchen-signal: Thanks - IMO it'll make it easier to do the right thing.

Signal/src/util/DateUtil.m
charlesmchen-signal: Yup - that's why I did a hybrid thing there and tried to used a localize format as much as possible.  No change requested, but we should mention this to Myles.

PR #3683, opened 2018-07-09
fix color for outgoing failed

michaelkirk-signal:
Moving a bit closer to the spec

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
@@ -164,12 +164,15 @@ NS_ASSUME_NONNULL_BEGIN
 
     [self configureFonts];
 
+    NSString *timestampLabelText;
     if ([self isFailedOutgoingMessage:viewItem]) {
-        self.timestampLabel.text
+        timestampLabelText
             = NSLocalizedString(@"MESSAGE_STATUS_SEND_FAILED", @"Label indicating that a message failed to send.");
     } else {
-        self.timestampLabel.text = [DateUtil formatMessageTimestamp:viewItem.interaction.timestamp];
+        timestampLabelText = [DateUtil formatMessageTimestamp:viewItem.interaction.timestamp];
     }
+
+    self.timestampLabel.text = timestampLabelText.localizedUppercaseString;
 }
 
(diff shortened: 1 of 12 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
```diff
@@ -164,12 +164,15 @@ - (void)configureLabelsWithConversationViewItem:(ConversationViewItem *)viewItem
...
     if ([self isFailedOutgoingMessage:viewItem]) {
-        self.timestampLabel.text
+        timestampLabelText
             = NSLocalizedString(@"MESSAGE_STATUS_SEND_FAILED", @"Label indicating that a message failed to send.");
     } else {
-        self.timestampLabel.text = [DateUtil formatMessageTimestamp:viewItem.interaction.timestamp];
+        timestampLabelText = [DateUtil formatMessageTimestamp:viewItem.interaction.timestamp];
     }
+
+    self.timestampLabel.text = timestampLabelText.localizedUppercaseString;
```
charlesmchen-signal: Hmm, there's a lot of places we're just using `uppercaseString`.  Can we please replace them with `localizedUppercaseString` - or make a card?

SignalMessaging/utils/ConversationStyle.swift
```diff
@@ -137,7 +137,7 @@ public class ConversationStyle: NSObject {
     private static let defaultBubbleColorIncoming = UIColor.ows_messageBubbleLightGray
 
     @objc
-    public let bubbleColorOutgoingUnsent = UIColor.ows_light10
+    public let bubbleColorOutgoingFailed = UIColor.ows_darkSkyBlue
```
charlesmchen-signal: Huh, so it's now the same as the "sent" color?

michaelkirk-signal: According to zeplin.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
michaelkirk-signal: I've changed to localizedUppercaseString everywhere it makes sense.

PR #3685, opened 2018-07-10
Fix white status bar for call screen

michaelkirk-signal:
Transition to VC controlled status bar style.

PTAL @charlesmchen

### DIFF
--- a/SignalMessaging/ViewControllers/OWSNavigationController.m
+++ b/SignalMessaging/ViewControllers/OWSNavigationController.m
@@ -107,12 +107,22 @@ NS_ASSUME_NONNULL_BEGIN
     }
 }
 
-
 #pragma mark - NavBarLayoutDelegate
 
 - (void)navBarCallLayoutDidChangeWithNavbar:(OWSNavigationBar *)navbar
 {
     [self updateLayoutForNavbar:navbar];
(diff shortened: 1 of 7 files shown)


SignalMessaging/ViewControllers/OWSNavigationController.m
```diff
@@ -107,12 +107,22 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
...
+}
+
+- (UIStatusBarStyle)preferredStatusBarStyle
+{
+    if (OWSWindowManager.sharedManager.hasCall) {
+        // Status bar is overlaying the green "call banner"
+        return UIStatusBarStyleLightContent;
+    } else {
+        return super.preferredStatusBarStyle;
+    }
```
charlesmchen-signal: Will this do the right thing during screen lock/protection?

charlesmchen-signal: Also, let's verify the registration flow. And the upgrade splashes.

michaelkirk-signal: Screen lock is above the status bar, good call on the upgrade splashes and registration flow. 

This change works for them as well, but I neglected to update their header style - I'll do that now.

PR #3686, opened 2018-07-10
tweak selected color for homeview cells

michaelkirk-signal:
PTAL @charlesmchen


SignalMessaging/categories/UIColor+OWS.m
```diff
@@ -117,6 +117,11 @@ + (UIColor *)ows_messageBubbleLightGrayColor
     return [UIColor colorWithHue:240.0f / 360.0f saturation:0.02f brightness:0.92f alpha:1.0f];
 }
 
++ (UIColor *)ows_tableCellSelectedBackgroundColor
```
charlesmchen-signal: I don't think we should have every single color constant defined here, especially for constants that we only use in a single place.  It'll make this category enormous, and I don't see any benefit.

michaelkirk-signal: ok, I'll take it out.

PR #3687, opened 2018-07-10
Charlesmchen/tweak disappearing messages indicator

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
```diff
@@ -158,6 +183,19 @@ - (BOOL)isFailedOutgoingMessage:(ConversationViewItem *)viewItem
     return messageStatus == MessageReceiptStatusFailed;
 }
 
+- (BOOL)isDisappearingMessage:(ConversationViewItem *)viewItem
```
michaelkirk-signal: This is a confusing name - because even unstarted disappearing messages are still disappearing messages.

Would "shouldShowDisappearingTimer" be a better name?

Also, if it simplifies things, I see no reason we couldn't show the timer for all disappearing messages whether or not they've "started" the timer. In practice this should be a very small amount of time between being scrolled onto screen and hitting the read timer.

In fact, having the (un-animated) timer in place initially should help minimize vertical shifting that would otherwise happen as the timer is added when the cell is redrawn.

So, coming full circle, I think you could leave this method the same, but change the implementation to:

`return message.isExpiringMessage;`

charlesmchen-signal: Done.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4951,6 +4954,7 @@ - (void)reloadViewItems
             TSIncomingMessage *incomingMessage = (TSIncomingMessage *)viewItem.interaction;
             NSString *incomingSenderId = incomingMessage.authorId;
             OWSAssert(incomingSenderId.length > 0);
+            BOOL isDisappearingMessage = incomingMessage.shouldStartExpireTimer;
```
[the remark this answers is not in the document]
charlesmchen-signal: See below.

charlesmchen-signal:
PTAL @michaelkirk

Signal/src/ViewControllers/ConversationView/Cells/OWSCallMessageCell.m
```diff
@@ -255,6 +255,7 @@ - (NSString *)titleForCall:(TSCall *)call
             return NSLocalizedString(
                 @"CALL_DECLINED", @"Message recorded in conversation history when local user declined a call.");
     }
+    return nil;
```
michaelkirk-signal: make this method nullable?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
```diff
@@ -160,6 +185,19 @@ - (BOOL)isFailedOutgoingMessage:(ConversationViewItem *)viewItem
     return messageStatus == MessageReceiptStatusFailed;
 }
 
+- (BOOL)shouldShowDisappearingTimer:(ConversationViewItem *)viewItem
+{
+    OWSAssert(viewItem);
```
michaelkirk-signal: TIOLI: we could move this logic into the viewItem: `viewItem.isDisappearingMessage`

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageTimerView.m
```diff
@@ -0,0 +1,195 @@
...
+            break;
+        case 11:
+            image = [UIImage imageNamed:@"disappearing_message_55"];
+            break;
+        case 12:
+            image = [UIImage imageNamed:@"disappearing_message_60"];
+            break;
+    }
+    OWSAssert(image);
+    OWSAssert(image.size.width == kDisappearingMessageIconSize);
```
michaelkirk-signal: nice check 👍

michaelkirk-signal:
As a coup de grâce you could delete the `shouldStartExpireTimer` sneaky transaction method.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
charlesmchen-signal: Done.

Signal/src/ViewControllers/ConversationView/Cells/OWSCallMessageCell.m
charlesmchen-signal: This class has been removed.

charlesmchen-signal:
> As a coup de grâce you could delete the shouldStartExpireTimer sneaky transaction method.

Done.

PR #3689, opened 2018-07-11
Charlesmchen/tweak system messages2

charlesmchen-signal:
PTAL @michaelkirk


Signal.xcodeproj/project.pbxproj
```diff
@@ -43,7 +43,6 @@
 		340FC8D0205BF2FA007AEB0F /* OWSBackupIO.m in Sources */ = {isa = PBXBuildFile; fileRef = 340FC8CE205BF2FA007AEB0F /* OWSBackupIO.m */; };
 		341F2C0F1F2B8AE700D07D6B /* DebugUIMisc.m in Sources */ = {isa = PBXBuildFile; fileRef = 341F2C0E1F2B8AE700D07D6B /* DebugUIMisc.m */; };
 		34277A5E20751BDC006049F2 /* OWSQuotedMessageView.m in Sources */ = {isa = PBXBuildFile; fileRef = 34277A5C20751BDC006049F2 /* OWSQuotedMessageView.m */; };
-		3427C64020EFD43E00EEC730 /* OWSCallMessageCell.m in Sources */ = {isa = PBXBuildFile; fileRef = 3427C63E20EFD43D00EEC730 /* OWSCallMessageCell.m */; };
```
charlesmchen-signal: Backs out the "call bubbles."

Signal/Images.xcassets/system_message_group.imageset/Contents.json
```diff
@@ -1,23 +0,0 @@
...
-    {
-      "idiom" : "universal",
-      "filename" : "system_message_group@1x.png",
-      "scale" : "1x"
-    },
-    {
-      "idiom" : "universal",
-      "filename" : "system_message_group@2x.png",
-      "scale" : "2x"
-    },
```
charlesmchen-signal: Fewer system messages have icons now.

Signal/src/AppDelegate.m
```diff
@@ -596,6 +597,9 @@ - (void)handleActivation
                 // Mark all "attempting out" messages as "unsent", i.e. any messages that were not successfully
                 // sent before the app exited should be marked as failures.
                 [[[OWSFailedMessagesJob alloc] initWithPrimaryStorage:[OWSPrimaryStorage sharedManager]] run];
+                // Mark all "incomplete" calls as missed, e.g. any incoming or outgoing calls that were not
+                // connected, failed or hung up before the app existed should be marked as missed.
+                [[[OWSIncompleteCallsJob alloc] initWithPrimaryStorage:[OWSPrimaryStorage sharedManager]] run];
```
charlesmchen-signal: In call records, we use the "incomplete" state to distinguish "calls which are in progress but haven't yet connected" from the "calls which completed/failed/were missed."  

* We had a "missed" state for incoming calls but not for outgoing calls.  We just left them as "incomplete" forever.  We need to distinguish these states.  For example, now that we can leave ongoing calls and review the conversation history, we want "ongoing" calls to show up differently than "missed" calls.  They should have different copy and shouldn't have a "call again/call back" button.  So I've added a "missed" state for outgoing calls.
* We use `OWSFailedMessagesJob` to ensure that "sending" messages are marked as failed on startup.  We need to do something analogous for calls, marking "incomplete" calls as "missed."
* We might want to rename "incomplete" to "connecting or ringing".

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -284,24 +284,18 @@
 /* Button text to enable batch selection mode */
 "BUTTON_SELECT" = "Select";
 
+/* Label for button that lets users call a contact again. */
+"CALL_AGAIN_BUTTON_TITLE" = "Call Again";
```
charlesmchen-signal: Incoming calls have "call back" button; outgoing calls have "call again" button.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -880,7 +877,7 @@
 "ERROR_MESSAGE_INVALID_KEY_EXCEPTION" = "The recipient's key is not valid.";
 
 /* No comment provided by engineer. */
-"ERROR_MESSAGE_INVALID_MESSAGE" = "Received message was out of sync. Tap to reset your secure session.";
+"ERROR_MESSAGE_INVALID_MESSAGE" = "Received message was out of sync.";
```
charlesmchen-signal: Buttons are now separate from system message titles.

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1063,7 +1060,7 @@
 "INCOMING_DECLINED_CALL" = "You declined a call";
 
 /* No comment provided by engineer. */
-"INCOMING_INCOMPLETE_CALL" = "Incomplete incoming call from";
+"INCOMING_INCOMPLETE_CALL" = "Incoming call";
```
charlesmchen-signal: This wasn't a format, so this copy was just wrong.

SignalServiceKit/src/Messages/OWSIncompleteCallsJob.h
```diff
@@ -0,0 +1,29 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class OWSPrimaryStorage;
+@class OWSStorage;
+
+@interface OWSIncompleteCallsJob : NSObject
```
charlesmchen-signal: Very analogous to the existing "failed message job".

SignalServiceKit/src/Messages/OWSIncompleteCallsJob.m
```diff
@@ -0,0 +1,150 @@
...
+{
+    OWSAssert(transaction);
+
+    NSMutableArray<NSString *> *messageIds = [NSMutableArray new];
+
+    NSString *formattedString = [NSString stringWithFormat:@"WHERE %@ == %d OR  %@ == %d",
+                                          OWSIncompleteCallsJobCallTypeColumn,
+                                          (int)RPRecentCallTypeOutgoingIncomplete,
+                                          OWSIncompleteCallsJobCallTypeColumn,
+                                          (int)RPRecentCallTypeIncomingIncomplete];
```
charlesmchen-signal: This seems fine, but IIRC SQL has IN syntax so we could also do "WHERE call_type IN (incoming incomplete, outgoing incomplete)".  Didn't bother in this case, but if we were trying to match many enum values it'd probably be  worthwhile.

SignalServiceKit/src/Messages/TSCall.m
```diff
@@ -36,8 +58,11 @@ - (instancetype)initWithTimestamp:(uint64_t)timestamp
 
     _callSchemaVersion = TSCallCurrentSchemaVersion;
     _callType = callType;
-    if (_callType == RPRecentCallTypeIncomingMissed
-        || _callType == RPRecentCallTypeIncomingMissedBecauseOfChangedIdentity) {
+
+    // Ensure users are notified of missed calls.
+    BOOL isIncomingMissed = (_callType == RPRecentCallTypeIncomingMissed
```
charlesmchen-signal: No change; just trying to improve readability.

SignalServiceKit/src/Storage/OWSStorage.m
```diff
@@ -234,7 +234,11 @@ - (nullable Class)unarchiver:(NSKeyedUnarchiver *)unarchiver
     cannotDecodeObjectOfClassName:(NSString *)name
                   originalClasses:(NSArray<NSString *> *)classNames
 {
-    OWSProdLogAndFail(@"%@ Could not decode object: %@", self.logTag, name);
+    if ([name isEqualToString:@"TSRecipient"]) {
+        DDLogError(@"%@ Could not decode object: %@", self.logTag, name);
```
charlesmchen-signal: Removed `fail` for known missing types so that we can test them on debug devices and verify we're handling them properly.

Signal/translations/en.lproj/Localizable.strings
michaelkirk-signal: 👍

SignalServiceKit/src/Storage/OWSStorage.m
michaelkirk-signal: 👍

Signal/src/AppDelegate.m
michaelkirk-signal: 👍

Signal/src/ViewControllers/ConversationView/Cells/OWSSystemMessageCell.m
```diff
@@ -43,38 +71,71 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+    self.button.layer.cornerRadius = 4.f;
+    [self.button addTarget:self action:@selector(buttonWasPressed:) forControlEvents:UIControlEventTouchUpInside];
+    [self.button autoSetDimension:ALDimensionHeight toSize:self.buttonHeight];
+
+    self.vStackView = [[UIStackView alloc] initWithArrangedSubviews:@[
+        contentStackView,
+        self.button,
+    ]];
+    self.vStackView.axis = UILayoutConstraintAxisVertical;
+    self.vStackView.spacing = self.buttonVSpacing;
```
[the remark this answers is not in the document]
charlesmchen-signal: > An alternative is to keep them in one stack view but add a spacer view to the stack view, or wrap content in a padded view which accounts for the spacing. 

Yeah, but that breaks (or complicates) one of the nicest things about `UIStackView`: that the layout skips hidden views _and only adds spacing between non-hidden views_.

PR #3691, opened 2018-07-11
Charlesmchen/rework unread indicator

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3901,18 +3895,10 @@ - (void)updateLastVisibleTimestamp
...
+    BOOL hasUnreadIndicator = self.dynamicInteractions.unreadIndicator != nil;
     if (!hasUnreadIndicator) {
         return;
     }
-    __block BOOL hasUnseenInteractions = NO;
-    [self.uiDatabaseConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
-        hasUnseenInteractions =
-            [[transaction ext:TSUnreadDatabaseViewExtensionName] numberOfItemsInGroup:self.thread.uniqueId] > 0;
-    }];
-    if (hasUnseenInteractions) {
```
michaelkirk-signal: I'm not sure of the consequences of this in aggregate - but typically `ensureDynamicInteractions` is quite expensive, and now we've removed a (seemingly quite common) guard that avoids running it. I just want to make sure this isn't going to destroy perf.

Looking a bit closer, it seems like these guards are still in place, but now they happen within the context of a read/write transaction in ensureDynamicInteractions, is that right?

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -255,13 +269,8 @@ - (CGFloat)vSpacingWithPreviousLayoutItem:(ConversationViewItem *)previousLayout
...
 
-    if (self.interaction.interactionType == OWSInteractionType_UnreadIndicator
-        || previousLayoutItem.interaction.interactionType == OWSInteractionType_UnreadIndicator) {
-        return 20.f;
-    }
-
-    if (self.shouldShowDate) {
-        return OWSMessageCellDateHeaderVMargin;
+    if (self.hasCellHeader) {
+        return OWSMessageHeaderViewDateHeaderVMargin;
```
michaelkirk-signal: Nice little cleanups all over. 👏 👏

SignalMessaging/utils/ThreadUtil.m
```diff
@@ -318,13 +306,14 @@ + (ThreadDynamicInteractions *)ensureDynamicInteractionsForThread:(TSThread *)th
         // have been marked as read.
         //
         // IFF this variable is non-null, there are unseen messages in the thread.
-        if (firstUnseenInteractionTimestampParameter) {
-            result.firstUnseenInteractionTimestamp = firstUnseenInteractionTimestampParameter;
+        NSNumber *_Nullable firstUnseenInteractionTimestamp = nil;
+        if (lastUnreadIndicator) {
```
michaelkirk-signal: This is so we don't "shift" the unread indicator down while the conversation is open, right?

SignalMessaging/utils/ThreadUtil.m
```diff
@@ -527,137 +527,94 @@ + (void)ensureUnreadIndicator:(ThreadDynamicInteractions *)dynamicInteractions
...
+                          // If there are more unseen messages than can be displayed in the
+                          // messages view, show the unread indicator at the top of the
+                          // displayed messages.
+                          *stop = YES;
+                          hasMoreUnseenMessages = YES;
+                      }
+                  }];
+
+    if (!interactionAfterUnreadIndicator) {
+        // If we can't find an interaction after the unread indicator,
```
michaelkirk-signal: This comment seems out of date.

charlesmchen-signal: It mostly still applies, but I've reworded it a little bit for clarity.

* We want to have "unread indicator continuity", e.g. we want to leave the unread indicator "in place" even after we've marked messages after it as read.  We persist `firstUnseenInteractionTimestamp` to accomplish this.
* `firstUnseenInteractionTimestamp` can get stale - all messages after the unread indicator may have disappeared or been manually deleted, etc.  In that case, we don't want to show the unread indicator.

SignalMessaging/utils/ThreadUtil.m
charlesmchen-signal: Right - it's about continuity.

charlesmchen-signal:
> verifying the blocking/non-blocking SN change way back in message history behavior.

This hasn't changed.  See `missingUnseenSafetyNumberChangeCount`.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: I just realized that we can now remove this method `cleanUpUnreadIndicatorIfNecessary` entirely.  Since the unread indicator is no longer an interaction, there's no need to eagerly clean it up.

charlesmchen-signal:
> perf regressions - how does this affect initial conversation load time? (Check for the existing: "First viewWillAppear took:" log line as a start)

I did some testing on a 5c with a conversation with 10k messages.  Perf was similar.

PR #3697, opened 2018-07-12
message actions

michaelkirk-signal:
PATL @charlesmchen


Signal/src/Models/MessageActions.swift
```diff
@@ -0,0 +1,131 @@
...
+
+@objc
+protocol MessageActionsDelegate: class {
+    func messageActionsShowDetailsForItem(_ conversationViewItem: ConversationViewItem)
+    func messageActionsReplyToItem(_ conversationViewItem: ConversationViewItem)
+}
+
+struct MessageActionBuilder {
+    static func reply(conversationViewItem: ConversationViewItem, delegate: MessageActionsDelegate) -> MenuAction {
+        return MenuAction(image: #imageLiteral(resourceName: "ic_reply"),
```
charlesmchen-signal: RE: image literals: if you want to navigate from this usage of the image to the asset in xcassets, how do you do that?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1978,8 +1982,121 @@ - (void)handleCallTap:(TSCall *)call
...
+    [self populateReplyForViewItem:conversationViewItem];
+}
+
+#pragma mark - MenuActionsViewControllerDelegate
+
+- (void)menuActionsDidHide:(MenuActionsViewController *)menuActionsViewController
+{
+    [[OWSWindowManager sharedManager] hideMenuActionsWindow];
+
+    [self updateShouldObserveDBModifications];
```
charlesmchen-signal: If we're trying to ensure that `updateShouldObserveDBModifications` is called whenever the message actions UI is dismissed, this seems insufficient, since `menuActionsDidHide` is only called when the user initiates the dismissal.  But it should be fine, since the other causes of dismissal (app becoming inactive, incoming call, etc.) should have the side effect of calling  `updateShouldObserveDBModifications`, right?

I just tend to like to ensure invariants in a simpler way - e.g. by having this VC observe a "menu actions was dismissed" event (notification?) that is fired in _all_ cases.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1978,8 +1982,121 @@ - (void)handleCallTap:(TSCall *)call
...
+        self.logTag,
+        __PRETTY_FUNCTION__,
+        verticalChange,
+        NSStringFromUIEdgeInsets(oldInset),
+        NSStringFromUIEdgeInsets(newInset));
+
+    // Because we're in the context of the frame-changing animation, these adjustments should happen
+    // in lockstep with the messageActions frame change.
+    self.collectionView.contentOffset = newOffset;
+    self.collectionView.contentInset = newInset;
```
charlesmchen-signal: How do we ensure that `contentOffset` and `contentInset` are restored to their default values after dismissal?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4425,8 +4542,23 @@ - (void)updateCellsVisible
...
+        return;
+    }
+
+    if (OWSWindowManager.sharedManager.isPresentingMenuActions) {
+        self.shouldObserveDBModifications = NO;
+        return;
+    }
+
+    self.shouldObserveDBModifications = YES;
+    return;
```
charlesmchen-signal: Why `return`?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: e.g. NSNotificationName...isPresentingMenuActionsDidChange

Signal/src/ViewControllers/MediaDetailViewController.m
```diff
@@ -22,29 +22,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-// In order to use UIMenuController, the view from which it is
```
charlesmchen-signal: Does this PR eliminate all usage of `UIMenuController`?  🔥

Signal/src/ViewControllers/MenuActionsViewController.swift
```diff
@@ -0,0 +1,355 @@
...
+    // MARK: Actions
+
+    @objc
+    func didTapBackground() {
+        animateDismiss(action: nil)
+    }
+
+    @objc
+    func didSwipeBackground(gesture: UISwipeGestureRecognizer) {
+        animateDismiss(action: nil)
```
charlesmchen-signal: Don't we need to check the GR's state?

Signal/translations/en.lproj/Localizable.strings
```diff
@@ -1161,6 +1161,27 @@
...
+"MESSAGE_ACTION_COPY_MEDIA" = "Copy Media";
+
+/* Action sheet button title */
+"MESSAGE_ACTION_COPY_TEXT" = "Copy Message Text";
+
+/* Action sheet button title */
+"MESSAGE_ACTION_DELETE_MESSAGE" = "Delete this Message";
+
+/* Action sheet button subtitle */
+"MESSAGE_ACTION_DELETE_MESSAGE_SUBTITLE" = "This will delete it on this device only";
```
charlesmchen-signal: TIOLI "It will be deleted on this device only."

SignalMessaging/utils/OWSWindowManager.m
```diff
@@ -43,13 +42,40 @@ const UIWindowLevel UIWindowLevel_CallView(void)
...
     return UIWindowLevelStatusBar + 2.f;
 }
 
+// In front of everything
+const UIWindowLevel UIWindowLevel_MessageActions(void);
+const UIWindowLevel UIWindowLevel_MessageActions(void)
+{
+    // Note: To cover the keyboard, this is higher than the ScreenBlocking level,
+    // but this window is hidden when screen protection is shown.
+    return CGFLOAT_MAX;
```
charlesmchen-signal: Wow, so the highest level is necessary?  Couldn't we just `UIWindowLevel_ScreenBlocking() + 1`?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: Similarly, will `contentOffset` and `contentInset` be restored properly if something other than the user initiates dismissal?  e.g. an incoming call or app becoming inactive?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: Great point. 

To solve, I took a slightly different approach using the delegate, rather than a notification, which I think is a little easier to get right with it's explicit sender/receiver.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: (I solved this the same as above)

> To solve, I took a slightly different approach using the delegate, rather than a notification, which I think is a little easier to get right with it's explicit sender/receiver.

Signal/src/Models/MessageActions.swift
michaelkirk-signal: Although rendering the glyph into XCode is cute, I'm not really **that** into it. 

I primarily like it for:
- autocompletes image names
- it asserts that the file exists.

michaelkirk-signal:
PTAL @charlesmchen

SignalMessaging/utils/OWSWindowManager.m
charlesmchen-signal: I guess I'm just worried that UIKit might not handle such high levels correctly.  Let's do the `-100` if you don't mind.

Signal/src/ViewControllers/MenuActionsViewController.swift
```diff
@@ -0,0 +1,402 @@
...
+
+    var didInformDelegateThatDisappearenceCompleted = false
+    func ensureDelegateIsInformedThatDisappearenceCompleted() {
+        guard !didInformDelegateThatDisappearenceCompleted else {
+            Logger.debug("\(logTag) in \(#function) ignoring redundant 'disappeared' notification")
+            return
+        }
+        didInformDelegateThatDisappearenceCompleted = true
+
+        self.delegate?.menuActionsDidHide(self)
```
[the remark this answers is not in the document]
michaelkirk-signal: > Sorry, I meant to ask if isDismissingWithVerticalFocusChange could be called twice?

The new method: `ensureDelegateIsInformedOfDismissalAnimation` exists precisely to ensure that `isDismissingWithVerticalFocusChange` is only called once, (hence the `guard !didInformDelegateOfDismissalAnimation`)

Since, as you correctly pointed out, there are multiple ways to exit the MenuActions view (e.g. the user could explicitly dismiss it, or it could get implicitly dismissed by backgrounding the app) we can sprinkle around the `ensure...` methods at all possible exit points, and be confident the delegate will never be informed more than once.

SignalMessaging/utils/OWSWindowManager.m
michaelkirk-signal: Done!

michaelkirk-signal:
In 34be31b16

PR #3706, opened 2018-07-13
ScrollToBottom accounts for top inset

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -4286,13 +4286,27 @@ typedef enum : NSUInteger {
     // we use the collectionView bounds to determine where the "bottom" is.
     [self.view layoutIfNeeded];
 
-    CGFloat contentHeight = self.safeContentHeight;
+    const CGFloat topInset = ^{
+        if (@available(iOS 11, *)) {
+            return -self.collectionView.adjustedContentInset.top;
+        } else {
+            return -self.collectionView.contentInset.top;
+        }
+    }();
 
-    // bottomLayoutGuide accounts for extra offset needed on iPhoneX
+    const CGFloat bottomInset = ^{
+        if (@available(iOS 11, *)) {
+            return -self.collectionView.adjustedContentInset.bottom;
+        } else {
+            return -self.collectionView.contentInset.bottom;
+        }
+    }();
 
-    CGFloat dstY = MAX(0,
-        contentHeight + self.collectionView.contentInset.bottom + self.bottomLayoutGuide.length
-            - self.collectionView.bounds.size.height);
+    const CGFloat firstContentPageTop = topInset;
+    const CGFloat collectionViewUnobscuredHeight = self.collectionView.bounds.size.height + bottomInset;
+    const CGFloat lastContentPageTop = self.safeContentHeight - collectionViewUnobscuredHeight;
+
+    CGFloat dstY = MAX(firstContentPageTop, lastContentPageTop);
 
     [self.collectionView setContentOffset:CGPointMake(0, dstY) animated:NO];
     [self didScrollToBottom];


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4171,11 +4171,19 @@ - (void)scrollToBottomAnimated:(BOOL)animated
...
+        } else {
+            return -self.collectionView.contentInset.top;
+        }
+    }();
+
     // bottomLayoutGuide accounts for extra offset needed on iPhoneX
+    const CGFloat lastContentPageTop = contentHeight + self.collectionView.contentInset.bottom
+        + self.bottomLayoutGuide.length - self.collectionView.bounds.size.height;
 
-    CGFloat dstY = MAX(0,
```
michaelkirk-signal: Our max used to be zero, but it needs to be relative to the content insets.

charlesmchen-signal:
Is this PR intended to address all of the contentInset issues?  Because they're not just in conversation view.  Some other examples are documented in the card.

michaelkirk-signal:
> Is this PR intended to address all of the contentInset issues? Because they're not just in conversation view. Some other examples are documented in the card.

No, there's still some outstanding ones related to content not sliding under the navbar that I'm looking into.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: Will the following all play together nicely:

* UIViewController manipulating the insets (UIViewController.automaticallyAdjustsScrollViewInsets) (is this only when the view is loaded?)
* Conversation view manipulating the insets in response to keyboard frame events.
* Conversation view manipulating the insets in response to "custom action sheet" events.

charlesmchen-signal: Do we need to do a MIN() or MAX() to handle the edge cases where `contentHeight` is smaller than the viewport?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4171,11 +4171,19 @@ - (void)scrollToBottomAnimated:(BOOL)animated
...
-        contentHeight + self.collectionView.contentInset.bottom + self.bottomLayoutGuide.length
-            - self.collectionView.bounds.size.height);
+    CGFloat dstY = MAX(firstContentPageTop, lastContentPageTop);
```
charlesmchen-signal: Oh duh, I see the `MAX()` now.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
charlesmchen-signal: Forgive me for thinking out loud, but I think what we're trying to do here is something like:

```
 const CGFloat viewportHeight = 
(self.collectionView.bounds.size.height - (self.collectionView.contentInset.bottom + self.bottomLayoutGuide.length));
const CGFloat lastContentPageTop = contentHeight - viewportHeight;
```
Is that clearer?

michaelkirk-signal: > Forgive me for thinking out loud, but I think what we're trying to do here is something like:

Yup, that's what we're doing. Breaking out the temp variable as you have makes it clearer. I'll make that change.

michaelkirk-signal: Also to be clear, this diff is bigger because I reformatted code, but the only functional difference is, instead of using 0 for the max offset, we use the self.collectionView.contentInset.top (which used to == 0)

michaelkirk-signal:
> Sorry this took so long, I wanted to really understand it.

No problem. I appreciate the opportunity to make the code more literate. I'm going to merge now, but please take another look and let me know if you have concerns.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: > Will the following all play together nicely:

In theory, insets are insets, so scroll to bottom should respect them just fine.

In practice I've run through these scenarios on iOS9, 10, 11x and everything looks good.

PR #3707, opened 2018-07-13
fix contact cell

michaelkirk-signal:
On iOS9/10 the ContactCell is not respecting it's wrapper's margin.

We have some methods that are named very similarly to PureLayout methods.

`PureLayout#autoPinEdgesToSuperviewEdges` pins to leading/trailing/top/left
`OWS#autoPinToSuperviewEdges` now renamed to `ows_autoPinToSuperviewEdges` pinned to left/right/top/left.

I think it's probably rare that we want to use this one, but didn't want to make a broad sweeping change during stabilization.

`PureLayout#autoPinEdgesToSuperviewMargin` vs
`OWS#autoPinToSuperviewMargins` uses our own constraint adding logic,
and behaves differently in some cases. In the case of this PR, using PureLayout's flavor results in the margins being respected on iOS9 and 10 (both work with iOS11).

I haven't dug into exactly why this method is behaving differently, and again, was loathe to make a broad change during stabilization, so I've simply renamed our home-brew version to be more obvious when we're using one vs the other.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MessageDetailViewController.swift
+++ b/Signal/src/ViewControllers/MessageDetailViewController.swift
@@ -261,7 +261,7 @@ class MessageDetailViewController: OWSViewController, MediaGalleryDataSourceDele
                     let wrapper = UIView()
                     wrapper.layoutMargins = UIEdgeInsets(top: 8, left: 20, bottom: 8, right: 20)
                     wrapper.addSubview(cellView)
-                    cellView.autoPinToSuperviewMargins()
+                    cellView.autoPinEdgesToSuperviewMargins()
                     groupRows.append(wrapper)
                 }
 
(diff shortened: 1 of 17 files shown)


Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -261,7 +261,7 @@ class MessageDetailViewController: OWSViewController, MediaGalleryDataSourceDele
                     let wrapper = UIView()
                     wrapper.layoutMargins = UIEdgeInsets(top: 8, left: 20, bottom: 8, right: 20)
                     wrapper.addSubview(cellView)
-                    cellView.autoPinToSuperviewMargins()
+                    cellView.autoPinEdgesToSuperviewMargins()
```
michaelkirk-signal: This is the only substantive change, from the OWS flavor to the PureLayout flavor, everything else is just a rename.

PR #3708, opened 2018-07-13
decrease header spacing from 32->28.

michaelkirk-signal:
Previous spacing was based on applying the "full width" gutter spacing twice.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageHeaderView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageHeaderView.m
@@ -93,8 +93,8 @@ const CGFloat OWSMessageHeaderViewDateHeaderVMargin = 23;
         [self.strokeView autoSetDimension:ALDimensionHeight toSize:strokeThickness],
 
         [self.stackView autoPinEdgeToSuperviewEdge:ALEdgeTop],
-        [self.stackView autoPinEdgeToSuperviewEdge:ALEdgeLeading],
-        [self.stackView autoPinEdgeToSuperviewEdge:ALEdgeTrailing],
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageHeaderView.m
```diff
@@ -93,8 +93,8 @@ - (void)loadForDisplayWithViewItem:(ConversationViewItem *)viewItem
         [self.strokeView autoSetDimension:ALDimensionHeight toSize:strokeThickness],
 
         [self.stackView autoPinEdgeToSuperviewEdge:ALEdgeTop],
-        [self.stackView autoPinEdgeToSuperviewEdge:ALEdgeLeading withInset:conversationStyle.fullWidthGutterLeading],
-        [self.stackView autoPinEdgeToSuperviewEdge:ALEdgeTrailing withInset:conversationStyle.fullWidthGutterTrailing],
+        [self.stackView autoPinEdgeToSuperviewEdge:ALEdgeLeading],
+        [self.stackView autoPinEdgeToSuperviewEdge:ALEdgeTrailing],
```
charlesmchen-signal: Aren't there corresponding changes we also need to make in the measurement logic?  I suspect we'll use it in at least one place: When determining the max width of the header text.

If so, I suggest that we move the usage of `headerGutterLeading` into `OWSMessageHeaderView` so that we gather this concern.

michaelkirk-signal: You're right - I need to account for this when measuring cell width.


> If so, I suggest that we move the usage of headerGutterLeading into OWSMessageHeaderView so that we gather this concern.

Sure, I'll do that.

michaelkirk-signal:
PTAL @charlesmchen

PR #3710, opened 2018-07-16
Fix compose picker content scrolls behind navbar

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/NewContactThreadViewController.m
+++ b/Signal/src/ViewControllers/NewContactThreadViewController.m
@@ -54,7 +54,6 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) UILocalizedIndexedCollation *collation;
 
 @property (nonatomic, readonly) UISearchBar *searchBar;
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/NewContactThreadViewController.m
```diff
@@ -295,6 +283,30 @@ - (void)updateTableContents
 
     __weak NewContactThreadViewController *weakSelf = self;
 
+    if (self.contactsViewHelper.contactsManager.isSystemContactsDenied) {
```
charlesmchen-signal: How do we update this if the user leaves app to grant/deny access?  Should we reload the table on "become active"?

michaelkirk-signal: Changing contact permissions kills the app.

charlesmchen-signal: Can you add a comment to this effect please?

michaelkirk-signal: sure, done.

PR #3713, opened 2018-07-16
Mkirk/remove new message after send

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -2610,9 +2610,12 @@ typedef enum : NSUInteger {
         ConversationViewItem *oldIndicatorItem = [self viewItemForIndex:indexPathOfUnreadIndicator.row];
         OWSAssert(oldIndicatorItem);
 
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2605,6 +2605,19 @@ - (void)clearUnreadMessagesIndicator
 {
     OWSAssertIsOnMainThread();
 
+    NSIndexPath *_Nullable indexPathOfUnreadIndicator = [self indexPathOfUnreadMessagesIndicator];
+    if (indexPathOfUnreadIndicator) {
+        ConversationViewItem *oldIndicatorItem = [self viewItemForIndex:indexPathOfUnreadIndicator.row];
+        OWSAssert(oldIndicatorItem);
+
+        // TODO ideally this would be happening within the *same* transaction that caused the unreadMessageIndicator
+        // to be cleared.
```
charlesmchen-signal: Hmm, it seems like we could modify `[OWSMessageSender enqueueMessage]` to take an "enqueuedBlock" which would let us do some extra work in the same transaction.  NBD Just thinking out loud.

PR #3714, opened 2018-07-16
fixup audio/generic attachment margins

michaelkirk-signal:
PTAL @charlesmchen


Signal/src/ViewControllers/ConversationView/Cells/OWSAudioMessageView.m
```diff
@@ -164,7 +164,7 @@ - (CGFloat)bubbleHeight
 
 + (CGFloat)iconSize
 {
-    return 40.f;
+    return 48.f;
```
michaelkirk-signal: Until I get the new asset from Myles this change will (rightfully) cause an assertion failure, so I won't merge until I get the new asset, but take a look @charlesmchen and let me know if it otherwise looks good.

Signal/src/ViewControllers/ConversationView/Cells/OWSAudioMessageView.m
```diff
@@ -140,7 +140,7 @@ - (CGFloat)hSpacing
 
 + (CGFloat)vMargin
 {
-    return 0.f;
+    return 5.f;
```
michaelkirk-signal: Note we were already accounting for this value in the sizing methods, we just weren't using them in the layout (and until now it didn't matter because they were zero).

PR #3719, opened 2018-07-17
Charlesmchen/delta contact intersection2

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/contacts/OWSContactsManager.m
+++ b/SignalMessaging/contacts/OWSContactsManager.m
@@ -119,7 +119,7 @@ NSString *const OWSContactsManagerKeyNextFullIntersectionDate = @"OWSContactsMan
     [self requestSystemContactsOnceWithCompletion:nil];
 }
 
-- (void)requestSystemContactsOnceWithCompletion:(void (^_Nullable)(NSError *error))completion
+- (void)requestSystemContactsOnceWithCompletion:(void (^_Nullable)(NSError *_Nullable error))completion
 {
     [self.systemContactsFetcher requestOnceWithCompletion:completion];
 }
@@ -129,7 +129,7 @@ NSString *const OWSContactsManagerKeyNextFullIntersectionDate = @"OWSContactsMan
     [self.systemContactsFetcher fetchOnceIfAlreadyAuthorized];
 }
 
-- (void)userRequestedSystemContactsRefreshWithCompletion:(void (^)(NSError *error))completionHandler
+- (void)userRequestedSystemContactsRefreshWithCompletion:(void (^)(NSError *_Nullable error))completionHandler
 {
     [self.systemContactsFetcher userRequestedRefreshWithCompletion:completionHandler];
 }
@@ -164,7 +164,7 @@ NSString *const OWSContactsManagerKeyNextFullIntersectionDate = @"OWSContactsMan
         return nil;
     }
 
-    CNContact *cnContact;
+    CNContact *_Nullable cnContact;
     @synchronized(self.cnContactCache) {
         cnContact = [self.cnContactCache objectForKey:contactId];
         if (!cnContact) {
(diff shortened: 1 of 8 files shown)


michaelkirk-signal:
Seems like this needs a rebase?

SignalMessaging/contacts/OWSContactsManager.m
```diff
@@ -258,12 +380,13 @@ - (void)intersectContactsWithRetryDelay:(double)retryDelaySeconds
         // TODO: Abort if another contact intersection succeeds in the meantime.
```
charlesmchen-signal: The exponential backoff logic.  

Jeff as requested that we ensure that there's never more than one CDS "contact sync/request batch" in flight at a time.  De-bouncing will be non-trivial though.

SignalMessaging/contacts/OWSContactsManager.m
```diff
@@ -225,29 +234,142 @@ - (void)systemContactsFetcher:(SystemContactsFetcher *)systemContactsFetcher
...
+
+    dispatch_async(self.serialQueue, ^{
+        [self.dbReadConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
+            [transaction setObject:recipientIdsForIntersection
+                            forKey:OWSContactsManagerKeyLastKnownContactPhoneNumbers
+                      inCollection:OWSContactsManagerCollection];
+
+            if (isFullIntersection) {
+                // Don't do a full intersection more often than once every 6 hours.
+                const NSTimeInterval kFullIntersectionRate = 6 * kHourInterval;
```
michaelkirk-signal: nit:`kFullIntersectionInterval` might be better since "rate" implies *per* i.e. the numerator, but this var holds the denominator.

SignalMessaging/contacts/OWSContactsManager.m
```diff
@@ -225,29 +234,142 @@ - (void)systemContactsFetcher:(SystemContactsFetcher *)systemContactsFetcher
...
+                NSDate *nextFullIntersectionDate =
```
michaelkirk-signal: No biggie, but in case we change this interval later, it probably makes more sense to store the `lastFullIntersectionDate` and compute the `nextFullIntersection` wherever it's used.

SignalMessaging/contacts/OWSContactsManager.m
michaelkirk-signal: NSOperation/OWSOperation with a non-concurrent may be helpful in that future work.

SignalServiceKit/src/Contacts/SignalRecipient.m
```diff
@@ -3,7 +3,9 @@
 //
 
 #import "SignalRecipient.h"
+#import "PhoneNumber.h"
```
michaelkirk-signal: minor point, but I don't see any new code that uses these imports?

SignalServiceKit/src/Util/NSDate+OWS.mm
```diff
@@ -45,6 +45,16 @@ - (BOOL)isBeforeDate:(NSDate *)otherDate
     return [self compare:otherDate] == NSOrderedAscending;
 }
 
+- (BOOL)isAfterNow
```
michaelkirk-signal: nice!

SignalMessaging/contacts/OWSContactsManager.m
charlesmchen-signal: I went this way at Jeff's suggestion because there's future work here to honor the "next full intersection interval" hint which CDS will eventually include in responses.  But happy to do either way if it turns out to be easier.

SignalServiceKit/src/Contacts/SignalRecipient.m
charlesmchen-signal: Removed.

SignalMessaging/contacts/OWSContactsManager.m
michaelkirk-signal: Ah, that makes sense.

PR #3720, opened 2018-07-17
fix squashed cells on ios9/10

michaelkirk-signal:
must specify rowHeight automatic

However, mixing static and automatic sizing within a single tableView on iOS9/10 caused some our cells to be too small, since they were previously relying on their static size.  To address this, I've converted the affected cells to be sizeable using autolayout.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ThreadSettings/OWSConversationSettingsViewController.m
+++ b/Signal/src/ViewControllers/ThreadSettings/OWSConversationSettingsViewController.m
@@ -231,6 +231,9 @@ const CGFloat kIconViewLength = 24;
 {
     [super viewDidLoad];
 
+    self.tableView.estimatedRowHeight = 45;
+    self.tableView.rowHeight = UITableViewAutomaticDimension;
+
     _disappearingMessagesDurationLabel = [UILabel new];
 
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ThreadSettings/OWSConversationSettingsViewController.m
```diff
@@ -372,39 +375,30 @@ - (void)updateTableContents
...
-                                 [switchView autoVCenterInSuperview];
-                                 [switchView autoPinLeadingToTrailingEdgeOfView:rowLabel offset:weakSelf.iconSpacing];
-                                 [switchView autoPinTrailingToSuperviewMargin];
+
+                                 UIStackView *topRow =
+                                     [[UIStackView alloc] initWithArrangedSubviews:@[ iconView, rowLabel, switchView ]];
+                                 topRow.spacing = self.iconSpacing;
+                                 topRow.alignment = UIStackViewAlignmentCenter;
+                                 [cell.contentView addSubview:topRow];
+                                 [topRow autoPinEdgesToSuperviewMarginsExcludingEdge:ALEdgeBottom];
```
charlesmchen-signal: Nice.

PR #3721, opened 2018-07-17
fix bottom blur artifact after selecting from compose

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/SignalMessaging/Views/OWSNavigationBar.swift
+++ b/SignalMessaging/Views/OWSNavigationBar.swift
@@ -21,7 +21,7 @@ public class OWSNavigationBar: UINavigationBar {
 
     @objc
     public var callBannerHeight: CGFloat {
-        return OWSWindowManagerCallScreenHeight()
+        return OWSWindowManagerCallBannerHeight()
     }
 
     @objc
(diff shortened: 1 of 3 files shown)


SignalMessaging/Views/OWSNavigationBar.swift
```diff
@@ -62,7 +62,9 @@ public class OWSNavigationBar: UINavigationBar {
             // so we have to set a zindex to avoid obscuring navbar title/buttons.
             blurEffectView.layer.zPosition = -1
 
-            blurEffectView.autoPinEdgesToSuperviewEdges(with: UIEdgeInsets(top: -100, left: 0, bottom: 0, right: 0))
+            // navbar frame doesn't account for statusBar, so, same as the built-in navbar background, we need to exceed
+            // the navbar bounds to have the blur extend up and behind the status bar.
+            blurEffectView.autoPinEdgesToSuperviewEdges(with: UIEdgeInsets(top: -statusBarHeight, left: 0, bottom: 0, right: 0))
```
michaelkirk-signal: This is the change, everything else was an unrelated rename.

PR #3723, opened 2018-07-18
Fixup tests

michaelkirk-signal:
And address some compiler warnings.

~~For the record, I'm unable to get the tests to build from within XCode9.4 before and after this PR.~~

~~The build error happens while building Swift files in Signal.app, with "undefined symbols". The missing symbols are Swift classes defined in SignalMessaging (though from the build logs I can see that SignalMessaging builds successfully).~~

~~I haven't yet tracked down that issue, but I am still able to run the tests with `bundle exec fastlane scan` (which is also using the Xcode9.4 build chain 🤷‍♂️ )~~ 

After a clean rebuild I was able to run the tests in XCode.

PTAL @charlesmchen


Signal/test/Models/MesssagesBubblesSizeCalculatorTest.swift
```diff
@@ -1,117 +0,0 @@
...
-//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
-//
-
-import XCTest
-import SignalServiceKit
-@testable import Signal
-
-/**
- * This is a brittle test, which will break if our layout changes.
- *
```
charlesmchen-signal: RIP

SignalMessaging/contacts/OWSContactsManager.m
```diff
@@ -332,7 +332,7 @@ - (void)buildSignalAccountsAndClearStaleCache:(BOOL)shouldClearStaleCache
         NSMutableSet<NSString *> *seenRecipientIds = [NSMutableSet new];
         for (Contact *contact in contacts) {
             NSArray<SignalRecipient *> *signalRecipients = contactIdToSignalRecipientsMap[contact.uniqueId];
-            for (SignalRecipient *signalRecipient in [signalRecipients sortedArrayUsingSelector:@selector(compare:)]) {
+            for (SignalRecipient *signalRecipient in [signalRecipients sortedArrayUsingSelector:@selector((compare:))]) {
```
charlesmchen-signal: Just curious: why is this necessary?

SignalMessaging/environment/migrations/OWS105AttachmentFilePaths.m
```diff
@@ -32,7 +32,7 @@ - (void)runUpWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
                                               [attachmentStreams addObject:attachmentStream];
                                           }];
 
-    DDLogInfo(@"Saving %zd attachment streams.", attachmentStreams.count);
+    DDLogInfo(@"Saving %lu attachment streams.", (unsigned long)attachmentStreams.count);
```
charlesmchen-signal: Can't you use `%zu` and skip the cast?

michaelkirk-signal: Nope, that results in a warning on 32bit. 

I previously explored avoiding the cast with a %tu which sort of "happens" to work on 32/64, but opted not to continue that practice: https://github.com/signalapp/Signal-iOS/pull/3385#discussion_r186450824

SignalMessaging/contacts/OWSContactsManager.m
michaelkirk-signal: This suppresses the "Several methods with selector 'compare:' of mismatched types are found for the @selector expression"

All kinds of objects implement compare (with params of different types), and since so many objc API's use selectors (as opposed to an actual typed method reference like you could do in swift) this is the world we live in.

This is just me annotating to the compiler that we're aware in this case.

PR #3724, opened 2018-07-18
Use CurrentAppContext instead of compiler flag to affect test behavior

michaelkirk-signal:
- Gets rid of the SSK_BUILDING_FOR_TESTS flag
- uses "debug" build config instead of "test" build config, which was based on the debug test config.

In general, I'd like to minimize build configs if at all possible, since they tend to "drift apart" over time.

PTAL @charlesmchen


SignalServiceKit/src/Storage/OWSOrphanedDataCleaner.m
```diff
@@ -83,20 +75,18 @@ + (void)auditAndCleanup:(BOOL)shouldCleanup completion:(void (^_Nullable)(void))
...
 
-    CleanupLogDebug(@"%@ fileCount: %lu", self.logTag, (unsigned long)fileCount);
-    CleanupLogDebug(@"%@ totalFileSize: %lld", self.logTag, totalFileSize);
-    CleanupLogDebug(@"%@ attachmentStreams: %d", self.logTag, attachmentStreamCount);
-    CleanupLogDebug(
-        @"%@ attachmentStreams with file paths: %lu", self.logTag, (unsigned long)attachmentFilePaths.count);
+    DDLogDebug(@"%@ fileCount: %lu", self.logTag, (unsigned long)fileCount);
+    DDLogDebug(@"%@ totalFileSize: %lld", self.logTag, totalFileSize);
+    DDLogDebug(@"%@ attachmentStreams: %d", self.logTag, attachmentStreamCount);
+    DDLogDebug(@"%@ attachmentStreams with file paths: %lu", self.logTag, (unsigned long)attachmentFilePaths.count);
```
charlesmchen-signal: This logging won't work in tests now, right?

michaelkirk-signal: From what I can tell, it should work fine. The Signal-iOS test suite is logging without issue.

michaelkirk-signal:
In https://github.com/signalapp/Signal-iOS/commit/256ad75a4657bf84101dd9423a247cb0b2a83ebd

PR #3726, opened 2018-07-19
Mkirk/batch contact intersections

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
+++ b/SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
@@ -1,0 +1,196 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+extension Array {
+    func chunked(by chunkSize: Int) -> [[Element]] {
+        return stride(from: 0, to: self.count, by: chunkSize).map {
+            Array(self[$0..<Swift.min($0 + chunkSize, self.count)])
+        }
+    }
+}
+
+@objc
+class OWSContactDiscoveryOperation: OWSOperation {
+
+    // TODO verify proper batch size
+//    let batchSize = 2048
+    let batchSize = 10
+    let recipientIdsToLookup: [String]
+
+    @objc
+    var registeredRecipientIds: Set<String>
+
+    @objc
+    required init(recipientIdsToLookup: [String]) {
+        self.recipientIdsToLookup = recipientIdsToLookup
+        self.registeredRecipientIds = Set()
+
+        super.init()
+
+        Logger.debug("\(logTag) in \(#function) with recipientIdsToLookup: \(recipientIdsToLookup.count)")
+        for batchIds in recipientIdsToLookup.chunked(by: batchSize) {
+            let batchOperation = OWSContactDiscoveryBatchOperation(recipientIdsToLookup: batchIds)
+            self.addDependency(batchOperation)
+        }
+    }
+
+    // MARK: Mandatory overrides
+    // Called every retry, this is where the bulk of the operation's work should go.
+    override func run() {
+        Logger.debug("\(logTag) in \(#function)")
+
+        for dependency in self.dependencies {
+            guard let batchOperation = dependency as? OWSContactDiscoveryBatchOperation else {
+                owsFail("\(self.logTag) in \(#function) unexpected dependency: \(dependency)")
+                continue
+            }
+
+            self.registeredRecipientIds.formUnion(batchOperation.registeredRecipientIds)
+        }
+
+        self.reportSuccess()
+    }
+
+    // MARK: Optional Overrides
+
+    // Called one time only
+    override func checkForPreconditionError() -> Error? {
+        return super.checkForPreconditionError()
+    }
+
+    // Called at most one time.
+    override func didSucceed() {
+        super.didSucceed()
+    }
+
+    // Called at most one time, once retry is no longer possible.
+    override func didFail(error: Error) {
+        super.didFail(error: error)
+    }
+}
+
+class OWSContactDiscoveryBatchOperation: OWSOperation {
+
+    private let recipientIdsToLookup: [String]
+    var registeredRecipientIds: Set<String>
+
+    required init(recipientIdsToLookup: [String]) {
+        self.recipientIdsToLookup = recipientIdsToLookup
+        self.registeredRecipientIds = Set()
+
+        super.init()
+
+        Logger.debug("\(logTag) in \(#function) with recipientIdsToLookup: \(recipientIdsToLookup.count)")
+    }
+
+    private var networkManager: TSNetworkManager {
+        return TSNetworkManager.shared()
+    }
+
+    private func parse(response: Any?, phoneNumbersByHashes: [String: String]) throws -> Set<String> {
+
+        guard let responseDict = response as? [String: AnyObject] else {
+            let responseError: NSError = OWSErrorMakeUnableToProcessServerResponseError() as NSError
+            responseError.isRetryable = true
+
+            throw responseError
+        }
+
+        guard let contactDicts = responseDict["contacts"] as? [[String: AnyObject]] else {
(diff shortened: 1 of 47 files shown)


SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -132,21 +132,9 @@ - (instancetype)initWithMessage:(TSOutgoingMessage *)message
 
 - (nullable NSError *)checkForPreconditionError
 {
-    for (NSOperation *dependency in self.dependencies) {
```
michaelkirk-signal: I moved this "fail if a dependency failed" logic into the base OWSOperation class so we could use it elsewhere.

SignalServiceKit/src/Util/OWSOperation.h
```diff
@@ -24,22 +24,26 @@ typedef NS_ENUM(NSInteger, OWSOperationState) {
 // any of the errors were fatal. Fatal errors trump retryable errors.
 @interface OWSOperation : NSOperation
 
-@property (nullable) NSError *failingError;
-@property NSUInteger remainingRetries;
+@property (readonly, nullable) NSError *failingError;
```
michaelkirk-signal: These are all superficial ordering/commenting changes, except I made failingError private-write-only.

SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
```diff
@@ -0,0 +1,250 @@
...
+        Logger.debug("\(logTag) in \(#function) with recipientIdsToLookup: \(recipientIdsToLookup.count)")
+    }
+
+    // MARK: OWSOperationOverrides
+
+    // Called every retry, this is where the bulk of the operation's work should go.
+    override func run() {
+        Logger.debug("\(logTag) in \(#function)")
+
+        Logger.debug("\(logTag) in \(#function) FAKING intersection (TODO)")
```
michaelkirk-signal: stubbed.

SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
```diff
@@ -0,0 +1,250 @@
...
+        guard let cdsOperation = dependencies.first as? CDSBatchOperation else {
+            let error = OWSErrorMakeAssertionError("\(self.logTag) in \(#function) cdsOperation was unexpectedly nil")
+            self.reportError(error)
+            return
+        }
+
+        let cdsRegisteredRecipientIds = cdsOperation.registeredRecipientIds
+
+        if cdsRegisteredRecipientIds == legacyRegisteredRecipientIds {
+            Logger.debug("\(logTag) in \(#function) TODO: PUT /v1/directory/feedback/ok")
```
michaelkirk-signal: stubbed

SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
```diff
@@ -0,0 +1,250 @@
...
+                                        failure: { (task, error) in
+                                            guard let response = task.response as? HTTPURLResponse else {
+                                                let responseError: NSError = OWSErrorMakeUnableToProcessServerResponseError() as NSError
+                                                responseError.isRetryable = true
+                                                self.reportError(responseError)
+                                                return
+                                            }
+
+                                            guard response.statusCode != 413 else {
+                                                let rateLimitError = OWSErrorWithCodeDescription(OWSErrorCode.contactsUpdaterRateLimit, "Contacts Intersection Rate Limit")
```
charlesmchen-signal: FYI this response will eventually include a hint about when to retry.

charlesmchen-signal: Wait, I meant the CDS rate limit errors.

SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
```diff
@@ -0,0 +1,250 @@
...
+
+    // Called every retry, this is where the bulk of the operation's work should go.
+    override func run() {
+        Logger.debug("\(logTag) in \(#function)")
+
+        var phoneNumbersByHashes: [String: String] = [:]
+
+        for recipientId in recipientIdsToLookup {
+            let hash = Cryptography.truncatedSHA1Base64EncodedWithoutPadding(recipientId)
+            phoneNumbersByHashes[hash] = recipientId
```
charlesmchen-signal: assert that there aren't any collisions?

SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
```diff
@@ -0,0 +1,250 @@
...
+
+        Logger.debug("\(logTag) in \(#function) FAKING intersection (TODO)")
+        self.registeredRecipientIds = Set(self.recipientIdsToLookup)
+        self.reportSuccess()
+    }
+}
+
+class CDSFeedbackOperation: OWSOperation {
+
+    static let operationQueue = OperationQueue()
```
charlesmchen-signal: Shouldn't we configure the concurrency level on this queue?

SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
```diff
@@ -0,0 +1,250 @@
...
+    required init(recipientIdsToLookup: [String]) {
+        self.recipientIdsToLookup = recipientIdsToLookup
+        self.registeredRecipientIds = Set()
+
+        super.init()
+
+        Logger.debug("\(logTag) in \(#function) with recipientIdsToLookup: \(recipientIdsToLookup.count)")
+        for batchIds in recipientIdsToLookup.chunked(by: batchSize) {
+            let batchOperation = LegacyContactDiscoveryBatchOperation(recipientIdsToLookup: batchIds)
+            self.addDependency(batchOperation)
```
charlesmchen-signal: So, if any of the operations in the batch fail, we treat the entire batch as a failure, right?  So if any fail, shouldn't we cancel any outstanding operations?

Also, `OWSOperation` does retry, but without exponential backoff.  Don't we need to change that now?

SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
michaelkirk-signal: Good catch! Fixed.

SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
michaelkirk-signal: OWSOperation **can** do retry, but only if you set `remainingRetries` to something greater than 0.

For now, none of these options retry, and I've left all retry logic in the callers (OWSContactManager#intersectContacts:retryDelaySexconds...)

michaelkirk-signal: > So, if any of the operations in the batch fail, we treat the entire batch as a failure, right? So if any fail, shouldn't we cancel any outstanding operations?

Yeah, good idea. I'll make this change.

SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
michaelkirk-signal: done!

SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
michaelkirk-signal: 👍

SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
michaelkirk-signal: I implemented the dependent cancel logic here:

https://github.com/signalapp/Signal-iOS/pull/3726/commits/2e30b33415a3cad161632f9f63f26c467c5f1c87

michaelkirk-signal:
rebased and merged in 0513077d2

PR #3728, opened 2018-07-20
Mkirk/fix overzealous error

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Contacts/ContactsUpdater.m
+++ b/SignalServiceKit/src/Contacts/ContactsUpdater.m
@@ -88,11 +88,11 @@ NS_ASSUME_NONNULL_BEGIN
 
     [self contactIntersectionWithSet:[NSSet setWithArray:identifiers]
                              success:^(NSSet<SignalRecipient *> *recipients) {
-                                 if (recipients.count > 0) {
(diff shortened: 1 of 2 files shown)


SignalServiceKit/src/Contacts/ContactsUpdater.m
```diff
@@ -88,11 +88,11 @@ - (void)lookupIdentifiers:(NSArray<NSString *> *)identifiers
 
     [self contactIntersectionWithSet:[NSSet setWithArray:identifiers]
                              success:^(NSSet<SignalRecipient *> *recipients) {
-                                 if (recipients.count > 0) {
-                                     success(recipients.allObjects);
-                                 } else {
-                                     failure(OWSErrorMakeNoSuchSignalRecipientError());
+                                 if (recipients.count == 0) {
```
michaelkirk-signal: This condition is an error for lookup `lookupIdentifier` but not for `lookupIdentifier**s**`

michaelkirk-signal: Actually, this PR isn't ready yet. Some callers expect this list to be none empty... 

Closing for a minute.

PR #3741, opened 2018-07-23
Charlesmchen/byte parser

charlesmchen-signal:
PTAL @michaelkirk


Signal/test/util/ByteParserTest.swift
```diff
@@ -0,0 +1,168 @@
...
+//
+
+import XCTest
+@testable import Signal
+
+class ByteParserTest: XCTestCase {
+
+    override func setUp() {
+        super.setUp()
+//         self.imageCache = ImageCache()
```
michaelkirk-signal: vestigial?

Signal/test/util/ByteParserTest.swift
```diff
@@ -0,0 +1,168 @@
...
+        XCTAssertNotNil(parser)
+        XCTAssertFalse(parser.hasError)
+
+        XCTAssertEqual(0, parser.nextShort())
+        XCTAssertTrue(parser.hasError)
+    }
+
+    func testGetShort() {
+        let data = Data(bytes: [0x01, 0x00, 0x00, 0x01, 0x01, 0x01 ])
+        let parser = ByteParser(data: data, littleEndian: true)
```
michaelkirk-signal: a `littleEndian: false` test might be nice.

Signal/test/util/ByteParserTest.swift
charlesmchen-signal: Removed.

Signal/test/util/ByteParserTest.swift
charlesmchen-signal: Done.

PR #3743, opened 2018-07-23
Mkirk/fixup navbar

michaelkirk-signal:
Fixes some visual artifacts resulting from recent navbar changes.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/NewGroupViewController.m
+++ b/Signal/src/ViewControllers/NewGroupViewController.m
@@ -102,6 +102,8 @@ const NSUInteger kNewGroupViewControllerAvatarWidth = 68;
 
     self.title = [MessageStrings newGroupDefaultTitle];
 
+    self.view.backgroundColor = UIColor.whiteColor;
+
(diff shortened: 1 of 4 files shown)


Signal/src/ViewControllers/NewGroupViewController.m
```diff
@@ -102,6 +102,8 @@ - (void)loadView
 
     self.title = [MessageStrings newGroupDefaultTitle];
 
+    self.view.backgroundColor = UIColor.whiteColor;
```
charlesmchen-signal: I suspect this should be:

```
    self.view.backgroundColor = Theme.backgroundColor;
```
Maybe I'll move that into `OWSViewController` so we can dry this up though.

charlesmchen-signal: Oh sorry, this is for v2.28. Forget it.

michaelkirk-signal: You're right - updated.

michaelkirk-signal:
In 58e2e1383

PR #3746, opened 2018-07-23
Fix NPE using mock for unknown database objects.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Contacts/TSThread.m
+++ b/SignalServiceKit/src/Contacts/TSThread.m
@@ -97,12 +97,14 @@ NS_ASSUME_NONNULL_BEGIN
     NSMutableArray<NSString *> *interactionIds = [NSMutableArray new];
     YapDatabaseViewTransaction *interactionsByThread = [transaction ext:TSMessageDatabaseViewExtensionName];
     OWSAssert(interactionsByThread);
-    [interactionsByThread
-        enumerateKeysInGroup:self.uniqueId
-                  usingBlock:^(
-                      NSString *_Nonnull collection, NSString *_Nonnull key, NSUInteger index, BOOL *_Nonnull stop) {
-                      [interactionIds addObject:key];
-                  }];
+    [interactionsByThread enumerateKeysInGroup:self.uniqueId
+                                    usingBlock:^(NSString *collection, NSString *key, NSUInteger index, BOOL *stop) {
+                                        if (key.length < 1) {
+                                            OWSProdLogAndFail(@"%@ invalid key in thread interactions.", self.logTag);
+                                            return;
+                                        }
+                                        [interactionIds addObject:key];
+                                    }];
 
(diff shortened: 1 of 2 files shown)


SignalServiceKit/src/Storage/OWSStorage.m
```diff
@@ -457,7 +482,8 @@ + (YapDatabaseDeserializer)logOnFailureDeserializer
 
     return ^id(NSString __unused *collection, NSString __unused *key, NSData *data) {
         if (!data || data.length <= 0) {
-            return nil;
+            OWSProdLogAndFail(@"%@ can't deserialize null object: %@", self.logTag, collection);
+            return [OWSUnknownDBObject new];
```
charlesmchen-signal: Just being defensive and trying to eliminate other possible causes of NPE.  But I don't think this is the issue after all.

SignalServiceKit/src/Storage/OWSStorage.m
```diff
@@ -205,17 +205,42 @@ @interface OWSUnknownDBObject : NSObject <NSCoding>
...
+}
+
+- (void)touchWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
+{
+    OWSProdLogAndFail(@"%@ Tried to touch unknown object", self.logTag);
+
+    // No-op.
+}
+
+- (void)removeWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
```
charlesmchen-signal: This should make our mock slightly safer.

SignalServiceKit/src/Contacts/TSThread.m
```diff
@@ -97,12 +97,14 @@ - (void)removeAllThreadInteractionsWithTransaction:(YapDatabaseReadWriteTransact
...
     OWSAssert(interactionsByThread);
-    [interactionsByThread
-        enumerateKeysInGroup:self.uniqueId
-                  usingBlock:^(
-                      NSString *_Nonnull collection, NSString *_Nonnull key, NSUInteger index, BOOL *_Nonnull stop) {
-                      [interactionIds addObject:key];
-                  }];
+    [interactionsByThread enumerateKeysInGroup:self.uniqueId
+                                    usingBlock:^(NSString *collection, NSString *key, NSUInteger index, BOOL *stop) {
+                                        if (key.length < 1) {
```
charlesmchen-signal: I noticed that its the _key_ not the _deserialized object_ that is invalid (probably nil).  Which looks a lot like database corruption.

michaelkirk-signal: ah, good catch. Specifically - seems like our view is out of sync with the backing collection. 

Should we trigger a rebuild?

charlesmchen-signal: > Should we trigger a rebuild?

Done.

PR #3750, opened 2018-07-25
Swift protos for Envelope

michaelkirk-signal:
PTAL @charlesmchen


Podfile.lock
```diff
@@ -60,6 +60,7 @@ PODS:
     - Reachability
     - SAMKeychain
     - SocketRocket
+    - SwiftProtobuf
```
charlesmchen-signal: Aren't there corresponding changes to `BUILDING.md` around using brew to install a dependency or two of `SwiftProtobuf`?

Signal/src/Jobs/MessageFetcherJob.swift
```diff
@@ -38,10 +38,15 @@ public class MessageFetcherJob: NSObject {
 
         Logger.info("\(self.logTag) fetching messages via REST.")
 
-        let promise = self.fetchUndeliveredMessages().then { (envelopes: [OWSSignalServiceProtosEnvelope], more: Bool) -> Promise<Void> in
+        let promise = self.fetchUndeliveredMessages().then { (envelopes: [SSKEnvelope], more: Bool) -> Promise<Void> in
```
charlesmchen-signal: Hmm, `SSK` prefix is new.  I like it.

Signal/src/Jobs/MessageFetcherJob.swift
```diff
@@ -113,56 +118,57 @@ public class MessageFetcherJob: NSObject {
...
+                if let content = Data(base64Encoded: encodedContent) {
+                    return content
+                }
             }
-        }
+            return nil
+        }()
 
-        return builder.build()
+        return SSKEnvelope(timestamp: timestamp, source: source, sourceDevice: sourceDevice, type: type, content: content, legacyMessage: legacyMessage)
```
charlesmchen-signal: It's a shame to lose the builder pattern.  It's so nice when there are optional or nested protos.

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -3372,23 +3372,32 @@ + (void)sendRandomAttachment:(TSThread *)thread uti:(NSString *)uti length:(NSUI
...
-
-    [builder setTimestamp:[NSDate ows_millisecondTimeStamp]];
+    uint64_t timestamp = [NSDate ows_millisecondTimeStamp];
+    NSString *source = ^{
+        if ([thread isKindOfClass:[TSGroupThread class]]) {
+            TSGroupThread *gThread = (TSGroupThread *)thread;
+            return gThread.groupModel.groupMemberIds[0];
+        } else if ([thread isKindOfClass:[TSContactThread class]]) {
+            TSContactThread *contactThread = (TSContactThread *)thread;
+            return contactThread.contactIdentifier;
```
charlesmchen-signal: Should this be a method on `TSThread`?

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -3542,9 +3551,6 @@ + (OWSSignalServiceProtosEnvelope *)createEnvelopeForThread:(TSThread *)thread
                                                           withTransaction:transaction]];
         [result addObject:[TSErrorMessage invalidVersionWithEnvelope:[self createEnvelopeForThread:thread]
                                                      withTransaction:transaction]];
-        [result addObject:[TSInvalidIdentityKeyReceivingErrorMessage
-                              untrustedKeyWithEnvelope:[self createEnvelopeForThread:thread]
-                                       withTransaction:transaction]];
```
charlesmchen-signal: IMO it'd be nice to keep this so we can use this to ensure that we don't break users who have these old messages.

Signal/src/Jobs/MessageFetcherJob.swift
michaelkirk-signal: On the other hand, this ensures we're setting all required params.

If we wanted something more like the builder pattern, we could make the properties `var` instead of `let`.

Podfile.lock
michaelkirk-signal: IMO the instructions for generating protobufs aren't really relevant to people building the app, which is why there is a separate README in the protobufs subdir (which I updated in this PR).

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
michaelkirk-signal: I don't think so - In general we can't assume the sender is `groupMemberIds[0]`,  it's just expedient debug code.

SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -313,19 +314,19 @@ - (void)handleEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
...
+    } else if (envelope.legacyMessage != nil) { // DEPRECATED - Remove after all clients have been upgraded.
         OWSSignalServiceProtosDataMessage *dataMessage =
             [OWSSignalServiceProtosDataMessage parseFromData:plaintextData];
         DDLogInfo(
             @"%@ handling message: <DataMessage: %@ />", self.logTag, [self descriptionForDataMessage:dataMessage]);
 
         [self handleIncomingEnvelope:envelope withDataMessage:dataMessage transaction:transaction];
     } else {
-        OWSProdInfoWEnvelope([OWSAnalyticsEvents messageManagerErrorEnvelopeNoActionablePayload], envelope);
+        DDLogWarn(@"%@ Ignoring envelope with neither Content nor LegacyMessage", self.logTag);
```
charlesmchen-signal: `OWSProdInfoWEnvelope()` was a placeholder for any eventual analytics.  I think this is the only place we're using it, so if you want to remove it, we should remove that macro.  But maybe put in a TODO: here to add analytics?

SignalServiceKit/src/Protos/ProtoWrappers/SSKEnvelope.swift
```diff
@@ -0,0 +1,149 @@
...
+    @objc
+    public init(timestamp: UInt64, source: String, sourceDevice: UInt32, type: SSKEnvelopeType, content: Data?, legacyMessage: Data?) {
+        self.source = source
+        self.type = type
+        self.timestamp = timestamp
+        self.sourceDevice = sourceDevice
+        self.relay = nil
+        self.content = content
+        self.legacyMessage = legacyMessage
+    }
```
charlesmchen-signal: Shouldn't we explicitly invoke `super.init()` in these initializers?

SignalServiceKit/protobuf/OWSSignaliOSProtos.proto
```diff
@@ -4,11 +4,11 @@
...
 
-package signalios;
+// iOS - since we use a modern proto-compiler, we must specify
+// the legacy proto format.
+syntax = "proto2";
 
-// Signal-iOS
-import "objectivec-descriptor.proto";
-option (google.protobuf.objectivec_file_options).class_prefix = "OWSSignaliOSProtos";
+package signalios;
```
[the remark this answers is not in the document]
charlesmchen-signal: I also wonder if we should include "Proto" in these package names, e.g. ServiceProto_X, IOSProto_Y, etc. Just an idea.

SignalServiceKit/src/Protos/OWSSignalServiceProtos.pb.swift
```diff
@@ -0,0 +1,4199 @@
+// DO NOT EDIT.
```
charlesmchen-signal: I see that you included OWSSignalServiceProtos.pb.swift.  What about the other 3 generated swift files?

charlesmchen-signal:
Also, should we add unit tests that explore what happens when we receive various kinds of degenerate envelopes?

SignalServiceKit/src/Protos/OWSSignalServiceProtos.pb.swift
michaelkirk-signal: I verified that they generated. I could include them, but they aren't being used yet.

SignalServiceKit/src/Protos/ProtoWrappers/SSKEnvelope.swift
michaelkirk-signal: Swift does this implicitly as necessary. You only need to do it explicitly if you want to do something *after* the super initializer.

In general, Swift makes it pretty hard to do the wrong thing with initializers.

SignalServiceKit/protobuf/OWSSignaliOSProtos.proto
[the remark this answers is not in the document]
michaelkirk-signal: I changed the SignalService.proto package to be camel case so that the generated class names would be like SignalService_Envelope instead of Signalservice_Envelope. 

Beyond ensuring that they compile, I haven't messed with the other .proto files, but I'd imagine a similar change.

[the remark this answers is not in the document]
michaelkirk-signal: > I also wonder if we should include "Proto" in these package names, e.g. ServiceProto_X, IOSProto_Y, etc. Just an idea.

Yeah, I like that.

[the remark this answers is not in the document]
michaelkirk-signal: Ok - I've included the yet-to-be-used protos and modified the class prefix to include the word Proto, e.g. SignalServiceProtos_Envelope

SignalServiceKit/src/Messages/OWSMessageManager.m
michaelkirk-signal: Sorry - this was unintentional. I've reverted this change.

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
michaelkirk-signal: Good point - I've added this back.

PR #3751, opened 2018-07-25
Dry up Parameter parsing logic

michaelkirk-signal:
This depends on #3750, so review that first and if there's major feedback hold off on reviewing this one.

This works towards removing boiler plate while doing strict parsing of server responses.

PTAL @charlesmchen


SignalServiceKit/src/Util/ParamParser.swift
```diff
@@ -0,0 +1,126 @@
...
+
+import Foundation
+
+// A DSL for parsing expected and optional values from a Dictionary, appropriate for
+// validating a service response.
+//
+// Additionally it includes some helpers to DRY up common conversions.
+//
+// Rather than exhaustively enumerate accessors for types like `requireUInt32`, `requireInt64`, etc.
+// We instead leverage generics at the call site.
```
charlesmchen-signal: Nice.

SignalServiceKit/src/Util/ParamParser.swift
```diff
@@ -0,0 +1,126 @@
...
+//
+//     do {
+//         // Required
+//         let name: String = try paramParser.require(key: "name")
+//         let count: UInt32 = try paramParser.require(key: "count")
+//
+//         // Optional
+//         let last_seen: Date? = try paramParser.allow(key: "last_seen")
```
charlesmchen-signal: I find "require/allow" unintuitive, especially "allow".  Personally, I'd prefer "parseRequired/parseOptional" or "required/optional".  Strictly TIOLI

michaelkirk-signal: Since a primary goal of this is to allow concise calling, I chose pretty terse names, but I can live with `required`/`optional`, since I think "parse" is somewhat superfluous given it's a Parser class with two method flavors.

michaelkirk-signal:
In https://github.com/signalapp/Signal-iOS/commit/2c8a64a48f6a2ca6afcaf6bb761972fdb25cf5e4

PR #3752, opened 2018-07-25
Fixup parser: Robust to servers various "empty" types

michaelkirk-signal:
For empty base64Encoded data, sometimes the server sends "null" (which is represented as NSNull) and sometimes the server sends an empty string. 🤷‍♂️ 

In both cases we represent this as nil - which will fail if it's a required field.

PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Util/ParamParser.swift
+++ b/SignalServiceKit/src/Util/ParamParser.swift
@@ -64,6 +64,10 @@ public class ParamParser {
             return nil
         }
 
+        guard !(someValue is NSNull) else {
+            return nil
+        }
+
         guard let typedValue = someValue as? T else {
             throw invalid(key: key)
         }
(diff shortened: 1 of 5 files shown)


SignalServiceKit/src/Util/ParamParser.swift
```diff
@@ -121,6 +125,10 @@ public class ParamParser {
             throw ParseError.invalidFormat(key)
         }
 
+        guard data.count > 0 else {
+            return nil
```
charlesmchen-signal: Wait, I'm not sure this is a good idea.  IMO there's a difference between a "missing/null" value and empty byte string.  Even if the field is optional.

michaelkirk-signal: It's true that they are not necessarily the same thing.

I'm failing to think of a scenario where we "need" a data parameter, but accept a data parameter of 0 length.

michaelkirk-signal: e.g. in this case the service is sending an empty string for receipts, which have no body.

PR #3754, opened 2018-07-26
Log when we change logging state.

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/AppSettings/AdvancedSettingsTableViewController.m
```diff
@@ -267,10 +267,12 @@ - (void)syncPushTokens
 - (void)didToggleEnableLogSwitch:(UISwitch *)sender
 {
     if (!sender.isOn) {
+        DDLogInfo(@"%@ disabling logging.", self.logTag);
```
michaelkirk-signal: heh.

PR #3760, opened 2018-07-30
Mkirk/message actions ux

michaelkirk-signal:
PTAL @charlesmchen 

Mostly written opportunistically on a plane.

I could be persuaded to target master instead of release if you think any of this seems risky.

### DIFF
--- a/Signal/src/ViewControllers/MenuActionsViewController.swift
+++ b/Signal/src/ViewControllers/MenuActionsViewController.swift
@@ -299,8 +299,10 @@ class MenuActionSheetView: UIView, MenuActionViewDelegate {
 
         self.clipsToBounds = true
 
-        let panGesture: UIPanGestureRecognizer = UIPanGestureRecognizer(target: self, action: #selector(didPan(gesture:)))
-        self.addGestureRecognizer(panGesture)
+        let touchGesture = UILongPressGestureRecognizer(target: self, action: #selector(didTouch(gesture:)))
+        touchGesture.minimumPressDuration = 0.0
+        touchGesture.allowableMovement = CGFloat.greatestFiniteMagnitude
+        self.addGestureRecognizer(touchGesture)
     }
 
     required init?(coder aDecoder: NSCoder) {
@@ -308,7 +310,7 @@ class MenuActionSheetView: UIView, MenuActionViewDelegate {
     }
 
     @objc
-    public func didPan(gesture: UIPanGestureRecognizer) {
+    public func didTouch(gesture: UIGestureRecognizer) {
         switch gesture.state {
         case .possible:
             break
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/MenuActionsViewController.swift
```diff
@@ -288,29 +283,58 @@ class MenuActionSheetView: UIView, MenuActionViewDelegate {
...
 
         actions = []
+        actionViews = []
+        hapticFeedback = HapticFeedback()
 
         super.init(frame: frame)
 
         backgroundColor = UIColor.ows_light10
         addSubview(actionStackView)
-        actionStackView.ows_autoPinToSuperviewEdges()
```
charlesmchen-signal: Do we eventually want to purge our usage of `ows_autoPinToSuperviewEdges` entirely?  Is there a card for this?

Signal/src/ViewControllers/MenuActionsViewController.swift
```diff
@@ -329,6 +353,47 @@ class MenuActionSheetView: UIView, MenuActionViewDelegate {
...
+    }
+
+    private func highlightActionView(location: CGPoint, fromView: UIView) {
+        guard let touchedView = actionView(touchedBy: location, fromView: fromView) else {
+            unhighlightAllActionViews()
+            return
+        }
+
+        if hasEverHighlightedAction, !touchedView.isHighlighted {
+            self.hapticFeedback.selectionChanged()
```
charlesmchen-signal: So we use haptic feedback if the selection _changes_ but not if you just tap on an item.

Signal/test/util/HapticFeedback.swift
```diff
@@ -0,0 +1,50 @@
+//
```
charlesmchen-signal: Why is this file in the test directory?

Signal/test/util/HapticFeedback.swift
```diff
@@ -0,0 +1,50 @@
...
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+protocol HapticAdapter {
```
charlesmchen-signal: newline between imports and other file contents?

Signal/test/util/HapticFeedback.swift
michaelkirk-signal: 🤦‍♂️ 

Thanks for noticing.

Signal/src/ViewControllers/MenuActionsViewController.swift
michaelkirk-signal: Yup, exactly. That's consistent with how Apple uses (and specifies) this to be used.

Signal/src/ViewControllers/MenuActionsViewController.swift
michaelkirk-signal: I'd like to get rid of it entirely, but I consider it really low priority. I think with the name change (adding the `ows_` prefix) we'll naturally phase it out.

Signal/test/util/HapticFeedback.swift
michaelkirk-signal: Done.

PR #3761, opened 2018-07-31
Fix too-much-space when call banner is present

michaelkirk-signal:
Reconciles safeAreaInsets with *additional*SafeAreaInsets now that we have a translucent navbar.

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -40,6 +40,23 @@ typedef NS_ENUM(NSInteger, HomeViewMode) {
     HomeViewMode_Inbox,
 };
 
+// The bulk of the content in this view is driven by a YapDB view/mapping.
+// However, we also want to optionally include ReminderView's at the top
+// and an "Archived Conversations" button at the bottom. Rather than introduce
+// index-offsets into the Mapping calculation, we introduce two pseudo groups
+// to add a top and bottom section to the content, and create cells for those
+// sections without consulting the YapMapping.
+// This is a bit of a hack, but it consolidates the hacks into the Reminder/Archive section
+// and allows us to leaves the bulk of the content logic on the happy path.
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -734,31 +748,30 @@ - (NSInteger)numberOfSectionsInTableView:(UITableView *)tableView
     return (NSInteger)[self.threadMappings numberOfSections];
 }
 
-- (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger)section
+- (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger)aSection
 {
-    NSInteger result = (NSInteger)[self.threadMappings numberOfItemsInSection:(NSUInteger)section];
-    if (self.hasArchivedThreadsRow) {
-        // Add the "archived conversations" row.
-        result++;
```
michaelkirk-signal: Rather than offset indexes in a section managed by a YapDB mapping, I've added a new (potentially empty) section. This let's the yap logic work untouched in that section, and we can manually handle the other non-content sections (reminder views, archive button), which I think is a bit more flexible / less confusing.

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -418,7 +418,11 @@ - (void)viewDidLoad
     [self addChildViewController:searchResultsController];
     [self.view addSubview:searchResultsController.view];
     [searchResultsController.view autoPinEdgesToSuperviewEdgesWithInsets:UIEdgeInsetsZero excludingEdge:ALEdgeTop];
-    [searchResultsController.view autoPinTopToSuperviewMarginWithInset:56];
+    if (@available(iOS 11, *)) {
+        [searchResultsController.view autoPinTopToSuperviewMarginWithInset:56];
+    } else {
+        [searchResultsController.view autoPinToTopLayoutGuideOfViewController:self withInset:40];
```
charlesmchen-signal: Ugh

PR #3764, opened 2018-08-01
Charlesmchen/proto wrappers

charlesmchen-signal:
Uses code generation to generate Swift wrappers for the "Apple Swift Proto" classes.

PTAL @michaelkirk


Scripts/ProtoWrappers.py
```diff
@@ -0,0 +1,902 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import os
+import sys
```
charlesmchen-signal: Protobuf schemas have a pretty simple syntax, and we only use a narrow subset of it.

In the interest of pragmatism, this script only handles the schema syntax that we use and expects us to format the schema using our conventions.

Signal/src/Jobs/MessageFetcherJob.swift
```diff
@@ -38,7 +38,7 @@ public class MessageFetcherJob: NSObject {
 
         Logger.info("\(self.logTag) fetching messages via REST.")
 
-        let promise = self.fetchUndeliveredMessages().then { (envelopes: [SSKEnvelope], more: Bool) -> Promise<Void> in
+        let promise = self.fetchUndeliveredMessages().then { (envelopes: [SSKProtoEnvelope], more: Bool) -> Promise<Void> in
```
charlesmchen-signal: In this PR, I've generated wrapper for all of the "SSK" protos.  In a follow-up PR, I'll generate for our 4 other schemas as well.

`SSKProtoEnvelope` is the generated replacement for `SSKEnvelope`.

SignalServiceKit/protobuf/Makefile
```diff
@@ -2,12 +2,16 @@
 
 PROTOC=protoc \
 		--proto_path='./'
+WRAPPER_SCRIPT=../../Scripts/ProtoWrappers.py \
+		--proto-dir='./' --verbose
```
charlesmchen-signal: Before merging this, I'll probably remove the `--verbose` switch.

SignalServiceKit/src/Messages/InvalidKeyMessages/TSInvalidIdentityKeyReceivingErrorMessage.m
```diff
@@ -66,11 +66,11 @@ - (nullable instancetype)initForUnknownIdentityKeyWithTimestamp:(uint64_t)timest
     return self;
 }
 
-- (nullable SSKEnvelope *)envelope
+- (nullable SSKProtoEnvelope *)envelope
 {
     if (!_envelope) {
         NSError *error;
-        SSKEnvelope *_Nullable envelope = [[SSKEnvelope alloc] initWithSerializedData:self.envelopeData error:&error];
+        SSKProtoEnvelope *_Nullable envelope = [SSKProtoEnvelope parseData:self.envelopeData error:&error];
```
charlesmchen-signal: One departure from the `SSKEnvelope` is that I'm doing the parsing in a factory method rather than an initializer.

Signal/src/Jobs/MessageFetcherJob.swift
charlesmchen-signal: Note that I'm using the prefix `SSKProto` instead of `SSK`.  The intention is to avoid namespace collisions and improve readability.

SignalServiceKit/src/Protos/Generated/SSKProto.swift
```diff
@@ -0,0 +1,3163 @@
...
+		}
+	}
+
+	@objc public let type: SSKProtoEnvelopeType
+	@objc public let source: String?
+	@objc public let sourceDevice: UInt32
+	@objc public let relay: String?
+	@objc public let timestamp: UInt64
+	@objc public let legacyMessage: Data?
+	@objc public let content: Data?
```
charlesmchen-signal: One of the limitations is that our proto schemas don't use the `required` "rule" to indicate which fields are required.  One option would be to use schema comments to decorate required fields, e.g.:

```
// @required
optional Type   type          = 1;
// @required
optional string source        = 2;
// @required
optional uint32 sourceDevice  = 7;
optional string relay         = 3;
```

But I'm inclined to leave things as they stand. I suspect that the "optionality" of class properties won't require much change if any in our parsing code.

SignalServiceKit/src/Protos/Generated/SSKProto.swift
```diff
@@ -0,0 +1,3163 @@
...
+		if proto.hasLegacyMessage {
+			legacyMessage = proto.legacyMessage
+		}
+
+		var content: Data? = nil
+		if proto.hasContent {
+			content = proto.content
+		}
+
+		// MARK: - Begin Validation Logic for SSKProtoEnvelope -
```
charlesmchen-signal: The generate code leaves boundary markers for custom validation code.  Custom code is preserved when updating the generated code.

SignalServiceKit/src/Protos/Generated/SSKProto.swift
```diff
@@ -0,0 +1,3163 @@
...
+		                              source: source,
+		                              sourceDevice: sourceDevice,
+		                              relay: relay,
+		                              timestamp: timestamp,
+		                              legacyMessage: legacyMessage,
+		                              content: content)
+		return result
+	}
+
+	fileprivate var asProtobuf: SignalServiceProtos_Envelope {
```
charlesmchen-signal: Hopefully much of this code looks very familiar.

SignalServiceKit/src/Protos/Generated/SSKProto.swift
```diff
@@ -0,0 +1,3163 @@
...
+        }
+    }
+
+    @objc public let type: SSKProtoEnvelopeType
+    @objc public let source: String?
+    @objc public let sourceDevice: UInt32
+    @objc public let relay: String?
+    @objc public let timestamp: UInt64
+    @objc public let legacyMessage: Data?
+    @objc public let content: Data?
```
charlesmchen-signal: One of the limitations is that our proto schemas don't use the `required` "rule" to indicate which fields are required.  One option would be to use schema comments to decorate required fields, e.g.:

```
// @required
optional Type   type          = 1;
// @required
optional string source        = 2;
// @required
optional uint32 sourceDevice  = 7;
optional string relay         = 3;
```

But I'm inclined to leave things as they stand. I suspect that the "optionality" of class properties won't require much change if any in our parsing code.

SignalServiceKit/src/Protos/Generated/SSKProto.swift
```diff
@@ -0,0 +1,3163 @@
...
+        if proto.hasLegacyMessage {
+            legacyMessage = proto.legacyMessage
+        }
+
+        var content: Data? = nil
+        if proto.hasContent {
+            content = proto.content
+        }
+
+        // MARK: - Begin Validation Logic for SSKProtoEnvelope -
```
charlesmchen-signal: The generate code leaves boundary markers for custom validation code. Custom code is preserved when updating the generated code.

SignalServiceKit/src/Protos/Generated/SSKProto.swift
```diff
@@ -0,0 +1,3163 @@
...
+                                      sourceDevice: sourceDevice,
+                                      relay: relay,
+                                      timestamp: timestamp,
+                                      legacyMessage: legacyMessage,
+                                      content: content)
+        return result
+    }
+
+    fileprivate var asProtobuf: SignalServiceProtos_Envelope {
+        let proto = SignalServiceProtos_Envelope.with { (builder) in
```
charlesmchen-signal: Hopefully much of this code looks very familiar.

charlesmchen-signal:
oh I forgot to add an important comment around optionality of numerics.  I'll add it in a bit.

Signal/src/Jobs/MessageFetcherJob.swift
```diff
@@ -110,38 +110,51 @@ public class MessageFetcherJob: NSObject {
...
             more: moreMessages
         )
     }
 
-    private func buildEnvelope(messageDict: [String: Any]) -> SSKEnvelope? {
+    private func buildEnvelope(messageDict: [String: Any]) -> SSKProtoEnvelope? {
         do {
             let params = ParamParser(dictionary: messageDict)
 
+            let builder = SSKProtoEnvelope.SSKProtoEnvelopeBuilder()
```
charlesmchen-signal: I added in support for the builder pattern.  This ought to be very useful when we build complex, nested messages.

SignalServiceKit/protobuf/SignalService.proto
```diff
@@ -23,10 +23,14 @@ message Envelope {
     RECEIPT       = 5;
   }
 
+  // @required
```
charlesmchen-signal: So far, I've stuck with using only a single "decorator', `@required`.

charlesmchen-signal: Note that thanks to `@required`, I haven't need the "custom validation" logic at all so far.  It's still there in case we want to use it.

charlesmchen-signal: Note that `@required` doesn't just address the "nullability" questions - we can also enforce it in the parsing process.

SignalServiceKit/src/Protos/Generated/SSKProto.swift
```diff
@@ -0,0 +1,3588 @@
...
+
+    @objc public var content: Data? {
+        guard proto.hasContent else {
+            return nil
+        }
+        return proto.content
+    }
+    @objc public var hasContent: Bool {
+        return proto.hasContent
+    }
```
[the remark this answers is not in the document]
charlesmchen-signal: That is: for strings, we can just `guard let someString = wrapper.someString else {` but for primitive types like `UInt64`, we _have to_ consult `has...` since the accessors will return a default value if the field isn't set. This is how our old obj-c and new Swift proto parsing code works, and I think it's better than trying to use `nullable NSNumber *` which loses the numeric type info.

SignalServiceKit/src/Protos/Generated/SSKProto.swift
```diff
@@ -0,0 +1,3588 @@
...
+        switch value {
+        case .delivery: return .delivery
+        case .read: return .read
+        }
+    }
+
+    private class func SSKProtoReceiptMessageTypeUnwrap(_ value: SSKProtoReceiptMessageType) -> SignalServiceProtos_ReceiptMessage.TypeEnum {
+        switch value {
+        case .delivery: return .delivery
+
```
michaelkirk-signal: Indentation seems to have gone awry at some point

SignalServiceKit/src/Protos/Generated/SSKProto.swift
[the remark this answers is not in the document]
michaelkirk-signal: I understand why that would be true when the caller is in objc, but I'm not sure why it would be true for swift, where all types can be optional.

SignalServiceKit/src/Protos/Generated/SSKProto.swift
```diff
@@ -0,0 +1,3588 @@
...
+        }
+        let sourceDevice = proto.sourceDevice
+
+        guard proto.hasTimestamp else {
+            throw SSKProtoError.invalidProtobuf(description: "\(logTag) missing required field: timestamp")
+        }
+        let timestamp = proto.timestamp
+
+        // MARK: - Begin Validation Logic for SSKProtoEnvelope -
+
```
michaelkirk-signal: It seems like it would be really easy to blow away our customizations to this generated file. 

Could we separate any customizations into a separate file?

I guess since we're not actually using it anywhere at this point, I'll defer my bike-shedding on this for now.

Signal/src/ViewControllers/DebugUI/DebugUIMessages.m
```diff
@@ -3390,13 +3390,17 @@ + (SSKEnvelope *)createEnvelopeForThread:(TSThread *)thread
...
     }();
 
-    SSKEnvelope *envelope = [[SSKEnvelope alloc] initWithTimestamp:timestamp
-                                                            source:source
-                                                      sourceDevice:1
-                                                              type:SSKEnvelopeTypeCiphertext
-                                                           content:nil
-                                                     legacyMessage:nil];
-
+    SSKProtoEnvelopeBuilder *envelopeBuilder = [SSKProtoEnvelopeBuilder new];
```
michaelkirk-signal: There might be some circumstance where it's useful, but in general I'm opposed to swapping an initializer which (e.g. if we add a new required field) guarantees validity at compile time vs. this builder which won't fail until you run it (if you run it 😬 )

SignalServiceKit/src/Messages/InvalidKeyMessages/TSInvalidIdentityKeyReceivingErrorMessage.m
charlesmchen-signal: I suppose it's mostly a matter of taste.  I prefer initializers that are "dumb" and don't do a bunch of work.

SignalServiceKit/protobuf/SignalService.proto
charlesmchen-signal: `@required` have many effects, e.g.:

* The wrapper parsing methods will throw an exception if the value is not set. We use the same checking logic when parsing incoming protos and constructing outgoing protos, so this will affect both cases.
* Wrapper instances capture a non-optional reference to the value as a property, allowing other code to access with value without checking for nullability.
* Wrapper instances don't have a `has...` method.

SignalServiceKit/src/Protos/Generated/SSKProto.swift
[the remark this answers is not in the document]
charlesmchen-signal: These wrapper classes need to present a single API to Obj-c and Swift code.  I suppose we could offer a Swift-only accessor that is optional.  Almost all of our code that uses these wrappers is Obj-C, so I haven't bothered but we could add that.

SignalServiceKit/src/Protos/Generated/SSKProto.swift
charlesmchen-signal: Fixed.

SignalServiceKit/src/Protos/Generated/SSKProto.swift
charlesmchen-signal: Yup, I'm considering removing support for customizations actually.  Let's see how it plays out.

PR #3766, opened 2018-08-02
Mkirk/upstream yap db

michaelkirk-signal:
Applied branches:

- mkirk/invalidFinalIndices
- mkirk/upstream-unencrypted-headers

Also fixed up test suite.

PTAL @charlesmchen

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -8,7 +8,7 @@ def shared_pods
   # pod 'SQLCipher', path: '../sqlcipher2'
   pod 'SQLCipher', :git => 'https://github.com/sqlcipher/sqlcipher.git', :commit => 'd5c2bec'
   # pod 'YapDatabase/SQLCipher', path: '../YapDatabase'
-  pod 'YapDatabase/SQLCipher', :git => 'https://github.com/signalapp/YapDatabase.git', branch: 'release/unencryptedHeaders'
+  pod 'YapDatabase/SQLCipher', :git => 'https://github.com/signalapp/YapDatabase.git', branch: 'signal-release'
   # pod 'AxolotlKit',   path: '../SignalProtocolKit'
   pod 'SignalServiceKit', path: '.'
   pod 'AxolotlKit', git: 'https://github.com/signalapp/SignalProtocolKit.git'
--- a/SignalServiceKit/src/Storage/FullTextSearchFinder.swift
+++ b/SignalServiceKit/src/Storage/FullTextSearchFinder.swift
@@ -7,14 +7,14 @@ import Foundation
 // Create a searchable index for objects of type T
 public class SearchIndexer<T> {
 
-    private let indexBlock: (T) -> String
+    private let indexBlock: (T, YapDatabaseReadTransaction) -> String
 
-    public init(indexBlock: @escaping (T) -> String) {
+    public init(indexBlock: @escaping (T, YapDatabaseReadTransaction) -> String) {
         self.indexBlock = indexBlock
     }
 
-    public func index(_ item: T) -> String {
-        return normalize(indexingText: indexBlock(item))
+    public func index(_ item: T, transaction: YapDatabaseReadTransaction) -> String {
+        return normalize(indexingText: indexBlock(item, transaction))
     }
 
     private func normalize(indexingText: String) -> String {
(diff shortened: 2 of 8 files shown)


SignalServiceKit/src/Storage/FullTextSearchFinder.swift
```diff
@@ -193,58 +193,58 @@ public class FullTextSearchFinder: NSObject {
...
     }
 
-    private static func oversizeText(forMessage message: TSMessage) -> String? {
+    private static func oversizeText(forMessage message: TSMessage, transaction: YapDatabaseReadTransaction) -> String? {
         guard message.hasAttachments() else {
             return nil
         }
-        let dbConnection = OWSPrimaryStorage.shared().dbReadConnection
-        var oversizeText: String?
-        dbConnection.read({ (transaction) in
```
michaelkirk-signal: In this update YapDB passes a transaction to the indexing method, so we now pass it through instead of building this sneaky transaction.

SignalServiceKit/src/Storage/OWSPrimaryStorage.m
```diff
@@ -223,22 +180,48 @@ - (void)runSyncRegistrations
...
+    [OWSIncomingMessageFinder asyncRegisterExtensionWithPrimaryStorage:self];
+    [TSDatabaseView asyncRegisterSecondaryDevicesDatabaseView:self];
+    [OWSDisappearingMessagesFinder asyncRegisterDatabaseExtensions:self];
+    [OWSFailedMessagesJob asyncRegisterDatabaseExtensionsWithPrimaryStorage:self];
+    [OWSIncompleteCallsJob asyncRegisterDatabaseExtensionsWithPrimaryStorage:self];
+    [OWSFailedAttachmentDownloadsJob asyncRegisterDatabaseExtensionsWithPrimaryStorage:self];
+    [OWSMediaGalleryFinder asyncRegisterDatabaseExtensionsWithPrimaryStorage:self];
+    [TSDatabaseView asyncRegisterLazyRestoreAttachmentsDatabaseView:self];
+
+    [self.database flushExtensionRequestsWithCompletionQueue:nil
```
michaelkirk-signal: In this update YapDB introduced this method which we were approximating by having a completion handler on our last registration. 

I think this approach is a little more future proof.

Podfile
```diff
@@ -8,7 +8,7 @@ def shared_pods
   # pod 'SQLCipher', path: '../sqlcipher2'
   pod 'SQLCipher', :git => 'https://github.com/sqlcipher/sqlcipher.git', :commit => 'd5c2bec'
   # pod 'YapDatabase/SQLCipher', path: '../YapDatabase'
-  pod 'YapDatabase/SQLCipher', :git => 'https://github.com/signalapp/YapDatabase.git', branch: 'release/unencryptedHeaders'
```
michaelkirk-signal: Proposed upgrade process:

- reset `signal-release` to the upstream stable release
- for each patch, rebase and merge

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
charlesmchen-signal: Fantastic.

charlesmchen-signal: Hopefully that fixes (or at least reduces) some of the inconsistency issues that we required the guards below.

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
```diff
@@ -193,58 +193,58 @@ public class FullTextSearchFinder: NSObject {
...
-                owsFail("Could not load oversize text attachment")
-                return
-            }
-            oversizeText = text
-        })
-        return oversizeText
+
+        guard let attachment = message.attachment(with: transaction) else {
+            // This can happen during the initial save of incoming messages.
+            Logger.warn("Could not load attachment for search indexing.")
```
charlesmchen-signal: Can this become an error now?

SignalServiceKit/src/Storage/OWSPrimaryStorage.m
charlesmchen-signal: Great.  Since the stakes are high, I wonder how we can test this.  The cautious thing would be to add a `__block` flag that is set by the last view's completion block and is asserted by this methods' completion block, but that's probably a waste of time.

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
michaelkirk-signal: Oh good call. I think you're right.

michaelkirk-signal: It appears that it can. Changed this log to an owsFail

PR #3769, opened 2018-08-03
Charlesmchen/redesign optimizations

charlesmchen-signal:
PTAL @michaelkirk 

Contains 3 separate changes in 3 commits.  They can be reviewed separately.

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageTextView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageTextView.m
@@ -93,7 +93,7 @@ NS_ASSUME_NONNULL_BEGIN
         return;
     }
     [super setTextColor:textColor];
-    self.cachedSize = nil;
+    // No need to clear cached size here.
 }
 
 - (void)setFont:(nullable UIFont *)font
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageTextView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageTextView.m
@@ -7,6 +7,14 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@interface OWSMessageTextView ()
+
+@property (nonatomic, nullable) NSValue *cachedSize;
+
+@end
+
+#pragma mark -
+
 @implementation OWSMessageTextView
 
 // Our message text views are never used for editing;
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageTextView.m
```diff
@@ -7,6 +7,14 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@interface OWSMessageTextView ()
+
+@property (nonatomic, nullable) NSValue *cachedSize;
```
charlesmchen-signal: Cache the text view's size and ignore redundant updates.

Signal/src/util/DateUtil.m
```diff
@@ -259,14 +259,20 @@ + (NSString *)formatTimestampShort:(uint64_t)timestamp
 
 + (NSString *)formatDateShort:(NSDate *)date
 {
+    OWSAssertIsOnMainThread();
     OWSAssert(date);
 
+    NSDate *now = [NSDate date];
+    NSInteger dayDifference = [self daysFromFirstDate:date toSecondDate:now];
+    BOOL dateIsOlderThanToday = dayDifference > 0;
+    BOOL dateIsOlderThanOneWeek = dayDifference > 6;
```
charlesmchen-signal: Date math is expensive; this cuts the cost in half.

charlesmchen-signal:
> Contains 3 separate changes in 3 commits. They can be reviewed separately.

Make that 2 changes.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageTextView.m
```diff
@@ -61,6 +69,61 @@ - (BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *_Nullable)event
...
+    self.cachedSize = nil;
+}
+
+- (void)setTextColor:(nullable UIColor *)textColor
+{
+    if ([NSObject isNullableObject:textColor equalTo:self.textColor]) {
+        return;
+    }
+    [super setTextColor:textColor];
+    self.cachedSize = nil;
```
michaelkirk-signal: Text color changes the size?

Signal/src/util/DateUtil.m
michaelkirk-signal: Hmm... any insight into why it's expensive?

It looks like this change would approximately cut it in half?

charlesmchen-signal: It's the usage of `NSCalendar` to slice the date into components that is surprisingly expensive. Due to the relative (to now) nature of these methods, we can't cache.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageTextView.m
charlesmchen-signal: Yeah, we can skip that here.  Done, and added a comment.  Note that we still want this method; setting text color is not free.

PR #3770, opened 2018-08-03
Charlesmchen/defer message resizing

charlesmchen-signal:
PTAL @michaelkirk 

A proposal for how to reduce the "animation churn" around footer collapse/resizing.

Essentially, we reload all cells every minute in order to update footer timestamps.  We only change footer sizing & collapse in those updates, once per minute.  This will group and reduce these animations.

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
@@ -228,29 +228,6 @@ NS_ASSUME_NONNULL_BEGIN
     // Measure the actual current width, to be safe.
     CGFloat timestampLabelWidth = [self.timestampLabel sizeThatFits:CGSizeZero].width;
 
-    // Measuring the timestamp label's width is non-trivial since its
-    // contents can be relative the current time.  We avoid having
-    // message bubbles' "visually vibrate" as their timestamp labels
-    // vary in width.  So we try to leave enough space for all possible
-    // contents of this label _for the first hour of its lifetime_, when
-    // the timestamp is particularly volatile.
-    if ([DateUtil isTimestampFromLastHour:viewItem.interaction.timestamp]) {
-        // Measure the "now" case.
-        self.timestampLabel.text = [DateUtil exemplaryNowTimeFormat];
-        timestampLabelWidth = MAX(timestampLabelWidth, [self.timestampLabel sizeThatFits:CGSizeZero].width);
-        // Measure the "relative time" case.
-        // Since this case varies with time, we multiply to leave
-        // space for the worst case (whose exact value, due to localization,
-        // is unpredictable).
-        self.timestampLabel.text = [DateUtil exemplaryMinutesTimeFormat];
-        timestampLabelWidth = MAX(timestampLabelWidth,
-            [self.timestampLabel sizeThatFits:CGSizeZero].width + self.timestampLabel.font.lineHeight * 0.5f);
-
-        // Re-configure the labels with the current appropriate value in case
-        // we are configuring this view for display.
-        [self configureLabelsWithConversationViewItem:viewItem];
-    }
-
     result.width = timestampLabelWidth;
     if (viewItem.interaction.interactionType == OWSInteractionType_OutgoingMessage) {
         if (![self isFailedOutgoingMessage:viewItem]) {
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -448,7 +448,6 @@ typedef enum : NSUInteger {
     // Cache the cell media for ~24 cells.
     self.cellMediaCache.countLimit = 24;
     _conversationStyle = [[ConversationStyle alloc] initWithThread:thread];
-    self.collapseCutoffDate = [NSDate new];
 
     // We need to update the "unread indicator" _before_ we determine the initial range
     // size, since it depends on where the unread indicator is placed.
@@ -724,6 +723,8 @@ typedef enum : NSUInteger {
     [self updateBarButtonItems];
     [self updateNavigationTitle];
 
+    [self resetContentAndLayout];
+
     // We want to set the initial scroll state the first time we enter the view.
     if (!self.viewHasEverAppeared) {
         [self scrollToDefaultPosition];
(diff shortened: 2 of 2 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -5071,6 +5080,10 @@ - (void)reloadViewItems
             }
         }
 
+        if (viewItem.interaction.timestampForSorting > collapseCutoffTimestamp) {
+            shouldHideFooter = NO;
```
charlesmchen-signal: Only collapse footers for any new messages (message created since we enter the conversation) once per minute.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
```diff
@@ -228,29 +228,6 @@ - (CGSize)measureWithConversationViewItem:(ConversationViewItem *)viewItem
     // Measure the actual current width, to be safe.
     CGFloat timestampLabelWidth = [self.timestampLabel sizeThatFits:CGSizeZero].width;
 
-    // Measuring the timestamp label's width is non-trivial since its
-    // contents can be relative the current time.  We avoid having
-    // message bubbles' "visually vibrate" as their timestamp labels
-    // vary in width.  So we try to leave enough space for all possible
-    // contents of this label _for the first hour of its lifetime_, when
-    // the timestamp is particularly volatile.
-    if ([DateUtil isTimestampFromLastHour:viewItem.interaction.timestamp]) {
```
michaelkirk-signal: To make sure I understand - is this change intentionally re-introducing "width vibration" because it seems less bad than having overly wide bubbles with short messages (e.g. jumbomoji)

charlesmchen-signal: Right.

We'll now update the cell (footer collapse = "height vibration") every ~1 minute _or_ whenever we re-enter the conversation view/app.  If re-entry triggers the update, we reset the ~1 minute countdown.  

"width vibration" should occur with around the same cadence, since the time formatting's finest resolution is 1 minute.  Most of the time, the width and height vibrations will occur _together for all_ visible cells, rather then a series of sporadic and uncoordinated animations in different cells.

Ideally, we'd also do this kind of update whenever we send/receive a message, but these updates (e.g. calling `resetContentAndLayout`) are expensive (and cost grows with, and I haven't yet found an acceptable way to do that.  This would align us with Android behavior and be a better experience.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageFooterView.m
```diff
@@ -228,29 +228,6 @@ - (CGSize)measureWithConversationViewItem:(ConversationViewItem *)viewItem
...
-        // space for the worst case (whose exact value, due to localization,
-        // is unpredictable).
-        self.timestampLabel.text = [DateUtil exemplaryMinutesTimeFormat];
-        timestampLabelWidth = MAX(timestampLabelWidth,
-            [self.timestampLabel sizeThatFits:CGSizeZero].width + self.timestampLabel.font.lineHeight * 0.5f);
-
-        // Re-configure the labels with the current appropriate value in case
-        // we are configuring this view for display.
-        [self configureLabelsWithConversationViewItem:viewItem];
-    }
```
charlesmchen-signal: Width vibration is okay, it should usually occur in a coordinated way with the ~1 minute timer.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -235,6 +235,7 @@ @interface ConversationViewController () <AttachmentApprovalViewControllerDelega
 @property (nonatomic) ContactShareViewHelper *contactShareViewHelper;
 @property (nonatomic) NSTimer *reloadTimer;
 @property (nonatomic, nullable) NSDate *lastReloadDate;
+@property (nonatomic, nullable) NSDate *collapseCutoffDate;
```
charlesmchen-signal: Don't collapse footers of "very new" messages, e.g. messages:

* Messages created since we entered the view _and_ since we last did `resetContentAndLayout`, e.g. we re-entered the view or the "refresh timer" fired.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -722,6 +723,8 @@ - (void)viewWillAppear:(BOOL)animated
     [self updateBarButtonItems];
     [self updateNavigationTitle];
 
+    [self resetContentAndLayout];
```
charlesmchen-signal: Refresh every time we re-enter the view.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3384,10 +3388,9 @@ - (void)uiDatabaseDidUpdate:(NSNotification *)notification
         // These errors seems to be very rare; they can only be reproduced
         // using the more extreme actions in the debug UI.
         OWSProdLogAndFail(@"%@ hasMalformedRowChange", self.logTag);
-        [self reloadViewItems];
-        [self.collectionView reloadData];
-        self.lastReloadDate = [NSDate new];
+        [self resetContentAndLayout];
```
charlesmchen-signal: DRY this up into `resetContentAndLayout`.  Note that we were neglecting to invalidate the layout here.  Shouldn't have mattered since I don't know that we're hitting this case in production. This check is defensive.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4350,8 +4353,7 @@ - (void)conversationColorWasUpdated
 {
     [self.conversationStyle updateProperties];
     [self.headerView updateAvatar];
-    [self.collectionView reloadData];
-    self.lastReloadDate = [NSDate new];
+    [self resetContentAndLayout];
```
charlesmchen-signal: Similarly, we should refresh everything if the color is updated.

charlesmchen-signal:
Okay, this PR has been updated to reflect the discussion.  

If I can find an "affordable" way to refresh (all content) after every message send/receive, I'll do that too, but it doesn't look good.

PTAL @michaelkirk

PR #3771, opened 2018-08-03
Mkirk/fix remaining navbar

michaelkirk-signal:
PTAL @charlesmchen


Signal/src/ViewControllers/ContactsPicker.xib
```diff
@@ -1,59 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
michaelkirk-signal: our last xib 😢

charlesmchen-signal: yahoooooo

PR #3772, opened 2018-08-03
Charlesmchen/redundant content insets

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4161,7 +4161,9 @@ - (void)handleKeyboardNotification:(NSNotification *)notification
     BOOL wasScrolledToBottom = [self isScrolledToBottom];
 
     void (^adjustInsets)(void) = ^(void) {
-        self.collectionView.contentInset = newInsets;
+        if (!UIEdgeInsetsEqualToEdgeInsets(self.collectionView.contentInset, newInsets)) {
+            self.collectionView.contentInset = newInsets;
```
charlesmchen-signal: This turns out to be expensive enough that we can benefit from skipping it.

PR #3781, opened 2018-08-06
Fix lazy collapse of message cell footers.

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4820,8 +4823,6 @@ - (void)collectionViewDidChangeLayout
 // cell view models.
 - (void)reloadViewItems
 {
-    self.collapseCutoffDate = [NSDate new];
```
charlesmchen-signal: We can't set `collapseCutoffDate` here; this is called every time we get a database update notification.

PR #3786, opened 2018-08-07
Fix crash during CollectionView thrash

michaelkirk-signal:
PTAL @charlesmchen


michaelkirk-signal:
To be clear - I don't think this is the more widespread crash we've been experiencing. That seems to affect the animated code path. I haven't been able to reproduce that one yet.

PR #3788, opened 2018-08-07
Add missing header file

michaelkirk-signal:
Fixes build failure on Xcode10b5

PTAL @charlesmchen


Signal/SignalUBSan.supp
```diff
@@ -7,6 +7,10 @@ bool:YapDatabaseAutoViewTransaction.m
 # SocketRocket
 bool:SRWebSocket.m
 
+# SQLCipher
```
michaelkirk-signal: Not actually related to iOS12/Xcode10b5, but came upon this while doing some stress testing.

PR #3792, opened 2018-08-08
Remove 'has' accessors from proto wrappers.

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/call/CallService.swift
```diff
@@ -1366,11 +1366,9 @@ private class SignalCallData: NSObject {
         }
         let call = callData.call
 
-        if message.hasConnected {
+        if let connected = message.connected {
```
michaelkirk-signal: nice little cleanup. 👍

PR #3794, opened 2018-08-08
Mkirk/log owsfail

michaelkirk-signal:
PTAL @charlesmchen 

For auditing coherency, I split the process into commits, but by design many of the changes were "undone", which is why you see some spurious formatting changes.

tldr; I ended up introducing production logging in all occurrences of OWSFail, though I maintained the OWS(C)FailWithoutLogging macros in case we want them in the future.

### DIFF
--- a/SignalServiceKit/src/Util/OWSAsserts.h
+++ b/SignalServiceKit/src/Util/OWSAsserts.h
@@ -19,44 +19,49 @@ NS_ASSUME_NONNULL_BEGIN
 // OWSCAssert() and OWSCFail() should be used in free functions.
 
 #define OWSAssert(X)                                                                                                   \
-    if (!(X)) {                                                                                                        \
-        DDLogError(@"%s Assertion failed: %s", __PRETTY_FUNCTION__, CONVERT_EXPR_TO_STRING(X));                        \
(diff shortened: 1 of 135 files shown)


SignalServiceKit/src/Util/OWSAsserts.h
```diff
@@ -19,55 +19,56 @@ NS_ASSUME_NONNULL_BEGIN
...
-        [DDLog flushLog];                                                                                              \
-        NSAssert(0, @"Assertion failed: %s", CONVERT_EXPR_TO_STRING(X));                                               \
-    }
+    do {                                                                                                               \
+        if (!(X)) {                                                                                                    \
+            DDLogError(@"%s Assertion failed: %s", __PRETTY_FUNCTION__, CONVERT_EXPR_TO_STRING(X));                    \
+            [DDLog flushLog];                                                                                          \
+            NSAssert(0, @"Assertion failed: %s", CONVERT_EXPR_TO_STRING(X));                                           \
+        }                                                                                                              \
+    } while (NO)
```
michaelkirk-signal: I like this construct which enforces the caller add a semicolon to the macro invocation.

PR #3796, opened 2018-08-09
Charlesmchen/cache label sizes

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSLabel.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSLabel.m
@@ -61,6 +61,42 @@ NS_ASSUME_NONNULL_BEGIN
     self.cachedSize = nil;
 }
 
+- (void)setNumberOfLines:(NSInteger)numberOfLines
+{
+    if (self.numberOfLines == numberOfLines) {
+        return;
+    }
+    [super setNumberOfLines:numberOfLines];
+    self.cachedSize = nil;
+}
+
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -85,7 +86,7 @@ - (void)commontInit
     self.stackView = [UIStackView new];
     self.stackView.axis = UILayoutConstraintAxisVertical;
 
-    self.senderNameLabel = [UILabel new];
+    self.senderNameLabel = [OWSLabel new];
```
charlesmchen-signal: Caching size of sender name is a significant win on slower devices.

Signal/src/ViewControllers/ConversationView/Cells/OWSLabel.m
```diff
@@ -0,0 +1,76 @@
...
+}
+
+- (void)setLineBreakMode:(NSLineBreakMode)lineBreakMode
+{
+    if (self.lineBreakMode == lineBreakMode) {
+        return;
+    }
+    [super setLineBreakMode:lineBreakMode];
+    self.cachedSize = nil;
+}
```
michaelkirk-signal: They maybe not currently relevant, there's a few other properties we might want to consider as cache-invalidating for future proofing:

- numberOfLines
- adjustsFontSizeToFitWidth
- minimumScaleFactor
- minimumFontSize

charlesmchen-signal: Done.

PR #3797, opened 2018-08-09
Skip default implementation of preferredLayoutAttributesFittingAttrib…

charlesmchen-signal:
…utes.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
@@ -40,6 +40,13 @@ NS_ASSUME_NONNULL_BEGIN
     }
 }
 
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/ConversationView/Cells/ConversationViewCell.m
```diff
@@ -40,6 +40,13 @@ - (void)setIsCellVisible:(BOOL)isCellVisible
     }
 }
 
+// For perf reasons, skip the default implementation which is only relevant for self-sizing cells.
+- (UICollectionViewLayoutAttributes *)preferredLayoutAttributesFittingAttributes:
```
charlesmchen-signal: Small change with significant benefit on slower devices.

PR #3802, opened 2018-08-09
Fix unused warnings in proto wrappers.

charlesmchen-signal:
PTAL @michaelkirk


charlesmchen-signal:
PTAL @michaelkirk

michaelkirk-signal:
@charlesmchen - I'm proposing an alternative here: 
https://github.com/signalapp/Signal-iOS/pull/3834

Which ensures we can't call these methods outside of DEBUG

michaelkirk-signal:
obviated by https://github.com/signalapp/Signal-iOS/pull/3834

PR #3803, opened 2018-08-09
Tweak theme

michaelkirk-signal:
PTAL @charlesmchen 

before:
old-profile

after:
new-profile

### DIFF
--- a/SignalMessaging/categories/Theme.m
+++ b/SignalMessaging/categories/Theme.m
@@ -57,7 +57,7 @@ NSString *const ThemeKeyThemeEnabled = @"ThemeKeyThemeEnabled";
 + (UIColor *)offBackgroundColor
 {
     return (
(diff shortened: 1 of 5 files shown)


SignalMessaging/categories/Theme.m
```diff
@@ -110,7 +110,12 @@ + (UIColor *)toolbarBackgroundColor
 
 + (UIColor *)cellSelectedColor
 {
-    return (Theme.isDarkThemeEnabled ? UIColor.ows_whiteColor : UIColor.ows_blackColor);
+    return (Theme.isDarkThemeEnabled ? [UIColor colorWithWhite:0.2 alpha:1] : [UIColor colorWithWhite:0.92 alpha:1]);
+}
+
++ (UIColor *)cellSeparatorColor
```
charlesmchen-signal: Why is this in the Theme if it doesn't depend on the theme state?  I suggest `UIColor+OWS`.

PR #3805, opened 2018-08-09
Use dark blur for navbar

michaelkirk-signal:
PTAL @charlesmchen 

Before this change the navbar is always visible in dark theme, and the nav bar item icons become low contrast.
navbar-light-blur

After this change the navbar is invisible if there's no content behind it - in line with recent light-theme changes. 

And when there is content behind it, the icons still have relatively high contrast.

navbar-dark-blur


michaelkirk-signal:
In 9ab447a3d

PR #3806, opened 2018-08-09
Avoid incremental theme-redraws

michaelkirk-signal:
Before this change, the settings table redraws several times during the theme switch, which feels clunky - this feels cleaner.

PTAL @charlesmchen


michaelkirk-signal:
Merged in 71cb90b57

PR #3809, opened 2018-08-15
Unify log tags.

charlesmchen-signal:
Pretty mechanical.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/Jobs/MultiDeviceProfileKeyUpdateJob.swift
+++ b/Signal/src/Jobs/MultiDeviceProfileKeyUpdateJob.swift
@@ -14,8 +14,6 @@ import SignalMessaging
  */
 @objc public class MultiDeviceProfileKeyUpdateJob: NSObject {
 
-    let TAG = "[MultiDeviceProfileKeyUpdateJob]"
-
     private let profileKey: OWSAES256Key
     private let identityManager: OWSIdentityManager
     private let messageSender: MessageSender
@@ -38,7 +36,7 @@ import SignalMessaging
 
     func run(retryDelay: TimeInterval = 1) {
         guard let localNumber = TSAccountManager.localNumber() else {
-            owsFail("\(self.TAG) localNumber was unexpectedly nil")
+            owsFail("\(self.logTag) localNumber was unexpectedly nil")
             return
         }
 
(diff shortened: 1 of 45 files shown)


Signal/src/Jobs/MultiDeviceProfileKeyUpdateJob.swift
```diff
@@ -14,8 +14,6 @@ import SignalMessaging
  */
 @objc public class MultiDeviceProfileKeyUpdateJob: NSObject {
 
-    let TAG = "[MultiDeviceProfileKeyUpdateJob]"
```
charlesmchen-signal: Change 1: remove antiquated `TAG`s.

Signal/src/ViewControllers/ContactViewController.swift
```diff
@@ -573,7 +573,7 @@ class ContactViewController: OWSViewController, ContactShareViewHelperDelegate {
         Logger.info("\(self.logTag) \(#function)")
 
         guard let url = NSURL(string: "tel:\(phoneNumber.phoneNumber)") else {
-            owsFail("\(ContactViewController.logTag) could not open phone number.")
+            owsFail("\(ContactViewController.logTag()) could not open phone number.")
```
charlesmchen-signal: 2. Ensure all usages of class `logTag` property are invoked.

SignalServiceKit/src/Util/NSObject+OWS.h
```diff
@@ -10,7 +10,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, readonly) NSString *logTag;
 
-+ (NSString *)logTag;
+@property (class, nonatomic, readonly) NSString *logTag;
```
charlesmchen-signal: 3. Convert class method to class property.

Signal/src/ViewControllers/ContactViewController.swift
michaelkirk-signal: hmmm... isn't this obviated by:

```
3. Convert class method to class property.
```

charlesmchen-signal: Unfortunately, no.  I haven't looked into why yet.

PR #3810, opened 2018-08-15
Fix nominal race in "is registered" state.

charlesmchen-signal:
PTAL @michaelkirk


SignalServiceKit/src/Account/TSAccountManager.m
```diff
@@ -120,8 +120,8 @@ - (BOOL)isRegistered
             // Cache this once it's true since it's called alot, involves a dbLookup, and once set - it doesn't change.
             _isRegistered = [self storedLocalNumber] != nil;
         }
+        return _isRegistered;
     }
-    return _isRegistered;
```
charlesmchen-signal: Was safe in production, but better to do all access to this ivar in synchronized block.

PR #3818, opened 2018-08-16
Sync conversation colors.

charlesmchen-signal:
PTAL @michaelkirk


charlesmchen-signal:
> Is our unit test still passing?

I'll confirm that.

PR #3819, opened 2018-08-16
Improve message timestamp formatting.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/util/DateUtil.m
+++ b/Signal/src/util/DateUtil.m
@@ -357,46 +357,50 @@ static NSString *const DATE_FORMAT_WEEKDAY = @"EEEE";
 
     NSCalendar *calendar = [NSCalendar currentCalendar];
 
+    NSDateComponents *relativeDiffComponents =
+        [calendar components:NSCalendarUnitMinute | NSCalendarUnitHour fromDate:date toDate:nowDate options:0];
+
+    NSInteger minutesDiff = MAX(0, [relativeDiffComponents minute]);
+    if (minutesDiff < 1) {
+        return NSLocalizedString(@"DATE_NOW", @"The present; the current time.");
+    }
+
+    NSInteger hoursDiff = MAX(0, [relativeDiffComponents hour]);
+    if (hoursDiff < 1) {
+        NSString *minutesString = [OWSFormat formatInt:(int)minutesDiff];
+        return [NSString stringWithFormat:NSLocalizedString(@"DATE_MINUTES_AGO_FORMAT",
+                                              @"Format string for a relative time, expressed as a certain number of "
+                                              @"minutes in the past. Embeds {{The number of minutes}}."),
+                         minutesString];
+    }
+
     // Note: we are careful to treat "future" dates as "now".
     NSInteger yearsDiff = [self yearsFromFirstDate:date toSecondDate:nowDate];
-    NSInteger daysDiff = [self daysFromFirstDate:date toSecondDate:nowDate];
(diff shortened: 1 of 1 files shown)


Signal/src/util/DateUtil.m
```diff
@@ -357,46 +357,50 @@ + (NSString *)formatMessageTimestamp:(uint64_t)timestamp
 
     NSCalendar *calendar = [NSCalendar currentCalendar];
 
+    NSDateComponents *relativeDiffComponents =
```
charlesmchen-signal: Calculate the "relative time difference" minutes and hours as a single calculation, not as two separate calculations.

Signal/src/util/DateUtil.m
```diff
@@ -357,46 +357,50 @@ + (NSString *)formatMessageTimestamp:(uint64_t)timestamp
...
+        [calendar components:NSCalendarUnitMinute | NSCalendarUnitHour fromDate:date toDate:nowDate options:0];
+
+    NSInteger minutesDiff = MAX(0, [relativeDiffComponents minute]);
+    if (minutesDiff < 1) {
+        return NSLocalizedString(@"DATE_NOW", @"The present; the current time.");
+    }
```
charlesmchen-signal: Exit early in the "now" case.

Signal/src/util/DateUtil.m
```diff
@@ -357,46 +357,50 @@ + (NSString *)formatMessageTimestamp:(uint64_t)timestamp
...
+
+    NSInteger hoursDiff = MAX(0, [relativeDiffComponents hour]);
+    if (hoursDiff < 1) {
+        NSString *minutesString = [OWSFormat formatInt:(int)minutesDiff];
+        return [NSString stringWithFormat:NSLocalizedString(@"DATE_MINUTES_AGO_FORMAT",
+                                              @"Format string for a relative time, expressed as a certain number of "
+                                              @"minutes in the past. Embeds {{The number of minutes}}."),
+                         minutesString];
+    }
```
charlesmchen-signal: Exit early in the "N minutes ago" ago case.

Signal/src/util/DateUtil.m
```diff
@@ -357,46 +357,50 @@ + (NSString *)formatMessageTimestamp:(uint64_t)timestamp
...
     if (yearsDiff > 0) {
         // "Long date" + locale-specific "short" time format.
         NSString *dayOfWeek = [self.otherYearMessageFormatter stringFromDate:date];
         NSString *formattedTime = [[self timeFormatter] stringFromDate:date];
-        result = [[dayOfWeek rtlSafeAppend:@" "] rtlSafeAppend:formattedTime];
-    } else if (daysDiff >= 7) {
+        return [[dayOfWeek rtlSafeAppend:@" "] rtlSafeAppend:formattedTime];
+    }
+
+    NSInteger daysDiff = [self daysFromFirstDate:date toSecondDate:nowDate];
```
charlesmchen-signal: Only calculate `daysDiff` if necessary.

Signal/src/util/DateUtil.m
charlesmchen-signal: Note that this has perf benefits _and_ is a design refinement.   If someone sends or receives a message at Monday 11:59pm and looks at it at Tuesday 12:01 am, we want to see "2 minutes ago", not "Monday".  Myles has confirmed.

michaelkirk-signal: 👍

PR #3820, opened 2018-08-16
Tap to copy sender timestamp to clipboard.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MessageDetailViewController.swift
+++ b/Signal/src/ViewControllers/MessageDetailViewController.swift
@@ -284,9 +284,12 @@ class MessageDetailViewController: OWSViewController, MediaGalleryDataSourceDele
             }
         }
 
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -703,6 +706,11 @@ class MessageDetailViewController: OWSViewController, MediaGalleryDataSourceDele
         // no - op
     }
 
+    @objc func didTapSent(sender: UIGestureRecognizer) {
+        let messageTimestamp = "\(message.timestamp)"
+        UIPasteboard.general.string = messageTimestamp
```
michaelkirk-signal: I think we should be a little more careful about clobbering someone's pasteboard. A 'tap' can easily happen inadvertently.

Since this is a debug feature, maybe we should make it like 2 taps with two fingers or something.

charlesmchen-signal: I made it a long press.

PR #3821, opened 2018-08-16
Charlesmchen/theme review

charlesmchen-signal:
PTAL @michaelkirk 

Addresses 90% of the feedback from Myles.

### DIFF
--- a/Signal/src/ViewControllers/AppSettings/OWSLinkedDevicesTableViewController.m
+++ b/Signal/src/ViewControllers/AppSettings/OWSLinkedDevicesTableViewController.m
@@ -48,6 +48,7 @@ int const OWSLinkedDevicesTableViewControllerSectionAddDevice = 1;
     self.isExpectingMoreDevices = NO;
     self.tableView.rowHeight = UITableViewAutomaticDimension;
     self.tableView.estimatedRowHeight = 60;
+    self.tableView.separatorColor = Theme.hairlineColor;
 
     [self.tableView applyScrollViewInsetsFix];
 
--- a/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
+++ b/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
@@ -60,6 +60,7 @@ class ConversationSearchViewController: UITableViewController {
 
         tableView.rowHeight = UITableViewAutomaticDimension
         tableView.estimatedRowHeight = 60
+        tableView.separatorColor = Theme.hairlineColor
 
         tableView.register(EmptySearchResultCell.self, forCellReuseIdentifier: EmptySearchResultCell.reuseIdentifier)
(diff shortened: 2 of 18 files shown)


Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -276,6 +277,37 @@ class ConversationSearchViewController: UITableViewController {
...
+        guard let title = self.tableView(tableView, titleForHeaderInSection: section) else {
+            return nil
+        }
+
+        let label = UILabel()
+        label.textColor = Theme.secondaryColor
+        label.text = title
+        label.font = UIFont.ows_dynamicTypeBody.ows_mediumWeight()
+        label.tag = section
+        label.sizeToFit()
```
michaelkirk-signal: Do we need this since we're using autolayout for the label?

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -276,6 +277,37 @@ class ConversationSearchViewController: UITableViewController {
...
+
+        let hMargin: CGFloat = 15
+        let vMargin: CGFloat = 4
+        let wrapper = UIView()
+        wrapper.backgroundColor = Theme.offBackgroundColor
+        wrapper.addSubview(label)
+        label.autoPinWidthToSuperview(withMargin: hMargin)
+        label.autoPinHeightToSuperview(withMargin: vMargin)
+        wrapper.frame = CGRect(x: 0, y: 0, width: label.width() + 2 * hMargin, height: label.height() + 2 * vMargin)
```
michaelkirk-signal: Huh, we've got kind of a confusing mix of autolayout and manual sizing here, and typically the sizing logic would go in the `heightForHeaderInSection` method. 

Can we instead do all the layout with autolayout and have `heightForHeaderInSection` return `UITableViewAutomaticDimension `?

Not a deal breaker for me, it just seems like it'll be easier if we need to come back and tweak the layout later to avoid this manual sizing logic.

Signal/src/ViewControllers/AppSettings/OWSLinkedDevicesTableViewController.m
```diff
@@ -48,6 +48,7 @@ - (void)viewDidLoad
     self.isExpectingMoreDevices = NO;
     self.tableView.rowHeight = UITableViewAutomaticDimension;
     self.tableView.estimatedRowHeight = 60;
+    self.tableView.separatorColor = Theme.hairlineColor;
```
michaelkirk-signal: Can we instead use/adjust the existing `Theme.cellSeparatorColor` ?

charlesmchen-signal: Done.

PR #3822, opened 2018-08-17
Add custom themed search bar.

charlesmchen-signal:
PTAL @michaelkirk


SignalMessaging/Views/OWSTextField.m
```diff
@@ -32,6 +32,24 @@ - (void)ows_applyTheme
     self.keyboardAppearance = (Theme.isDarkThemeEnabled ? UIKeyboardAppearanceDark : UIKeyboardAppearanceDefault);
 }
 
+- (BOOL)becomeFirstResponder
+{
+    BOOL result = [super becomeFirstResponder];
+
+    [self.ows_delegate textFieldDidBecomeFirstResponder:self];
```
michaelkirk-signal: Should this be:

if (result) {
  [self.ows_delegate textFieldDidBecomeFirstResponder:self];
}

PR #3824, opened 2018-08-18
fix overzealous asserts

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/SignalShareExtension/utils/ShareAppExtensionContext.m
+++ b/SignalShareExtension/utils/ShareAppExtensionContext.m
@@ -145,7 +145,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (CGFloat)statusBarHeight
 {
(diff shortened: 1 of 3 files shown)


SignalShareExtension/utils/ShareAppExtensionContext.m
```diff
@@ -145,7 +145,6 @@ - (void)setStatusBarHidden:(BOOL)isHidden animated:(BOOL)isAnimated
 
 - (CGFloat)statusBarHeight
 {
-    OWSFail(@"%@ in %s unexpected for share extension", self.logTag, __PRETTY_FUNCTION__);
```
charlesmchen-signal: What calls this now?

michaelkirk-signal: SAE uses statusBarHeight via OWSNavigationController via OWSNavbar

PR #3825, opened 2018-08-18
faster conversation presentation after contact picking

michaelkirk:
Current after picking a conversation we first dismiss the keyboard, the current view (contact picker) then animate pushing the conversation view.

This PR pushes (without animation) the conversation view *before* dismissing the contact picker, allowing us to skip:

1. the keyboard dismiss animation
2. the conversation push animation

PTAL @charlesmchen 

You might want to build and run this and compare it to prod to get a feel for the difference.

### DIFF
--- a/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
+++ b/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
@@ -137,7 +137,7 @@ class ConversationSearchViewController: UITableViewController {
             }
 
             let thread = searchResult.thread
-            SignalApp.shared().presentConversation(for: thread.threadRecord, action: .compose)
+            SignalApp.shared().presentConversation(for: thread.threadRecord, action: .compose, animated: true)
 
         case .contacts:
             let sectionResults = searchResultSet.contacts
@@ -146,7 +146,7 @@ class ConversationSearchViewController: UITableViewController {
                 return
             }
 
-            SignalApp.shared().presentConversation(forRecipientId: searchResult.recipientId, action: .compose)
+            SignalApp.shared().presentConversation(forRecipientId: searchResult.recipientId, action: .compose, animated: true)
 
         case .messages:
             let sectionResults = searchResultSet.messages
@@ -157,8 +157,9 @@ class ConversationSearchViewController: UITableViewController {
 
             let thread = searchResult.thread
             SignalApp.shared().presentConversation(for: thread.threadRecord,
-                                                   action: .compose,
-                                                   focusMessageId: searchResult.messageId)
+                                                   action: .none,
+                                                   focusMessageId: searchResult.messageId,
+                                                   animated: true)
         }
     }
 
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.h
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.h
@@ -12,19 +12,15 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface HomeViewController : OWSViewController
 
-- (void)presentThread:(TSThread *)thread action:(ConversationViewAction)action;
+- (void)presentThread:(TSThread *)thread action:(ConversationViewAction)action animated:(BOOL)isAnimated;
+
 - (void)presentThread:(TSThread *)thread
                action:(ConversationViewAction)action
-       focusMessageId:(nullable NSString *)focusMessageId;
+       focusMessageId:(nullable NSString *)focusMessageId
+             animated:(BOOL)isAnimated;
 
+// Used by force-touch Springboard icon shortcut
 - (void)showNewConversationView;
-
-- (void)presentTopLevelModalViewController:(UIViewController *)viewController
-                          animateDismissal:(BOOL)animateDismissal
-                       animatePresentation:(BOOL)animatePresentation;
-- (void)pushTopLevelViewController:(UIViewController *)viewController
-                  animateDismissal:(BOOL)animateDismissal
-               animatePresentation:(BOOL)animatePresentation;
 
 @end
 
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -554,9 +554,8 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
         //
         // We just want to make sure contact access is *complete* before showing the compose
         // screen to avoid flicker.
-        OWSNavigationController *navigationController =
-            [[OWSNavigationController alloc] initWithRootViewController:viewController];
-        [self presentTopLevelModalViewController:navigationController animateDismissal:YES animatePresentation:YES];
+        OWSNavigationController *modal = [[OWSNavigationController alloc] initWithRootViewController:viewController];
+        [self.navigationController presentViewController:modal animated:YES completion:nil];
     }];
 }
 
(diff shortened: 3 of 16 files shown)


Signal/src/ViewControllers/HomeView/HomeViewController.h
```diff
@@ -12,6 +12,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface HomeViewController : OWSViewController
 
+@property (nonatomic) TSThread *lastThread;
```
charlesmchen-signal: Shouldn't this public property be readonly?

Signal/src/ViewControllers/NewContactThreadViewController.m
```diff
@@ -823,11 +823,8 @@ - (void)newConversationWithRecipientId:(NSString *)recipientId
 - (void)newConversationWithThread:(TSThread *)thread
 {
     OWSAssert(thread != nil);
-    [self dismissViewControllerAnimated:YES
-                             completion:^() {
-                                 [SignalApp.sharedApp presentConversationForThread:thread
-                                                                            action:ConversationViewActionCompose];
-                             }];
+    [SignalApp.sharedApp withoutAnimationPresentConversationForThread:thread action:ConversationViewActionCompose];
+    [[self presentingViewController] dismissViewControllerAnimated:YES completion:nil];
```
charlesmchen-signal: Is there an analogous change we could make in the "new group thread" view?  I'm assuming no change is need in the "home view search" view?

Signal/src/environment/SignalApp.m
```diff
@@ -220,6 +220,20 @@ - (void)presentConversationForThread:(TSThread *)thread
     });
 }
 
+- (void)withoutAnimationPresentConversationForThread:(TSThread *)thread action:(ConversationViewAction)action
```
charlesmchen-signal: This is fine, but we could also: `presentConversationForThread:animated:`.

Signal/src/environment/SignalApp.m
```diff
@@ -220,6 +220,20 @@ - (void)presentConversationForThread:(TSThread *)thread
...
+{
+    OWSAssertIsOnMainThread();
+    DDLogInfo(@"%@ %s", self.logTag, __PRETTY_FUNCTION__);
+
+    ConversationViewController *conversationVC = [ConversationViewController new];
+    [conversationVC configureForThread:thread action:action focusMessageId:nil];
+
+    UIViewController *homeVC = self.homeViewController;
+    homeVC.lastThread = thread;
```
charlesmchen-signal: Oh huh... so previously we've home view has always had responsibility for presenting its subview and it's `pushTopLevelViewController:...` and `presentViewControllerWithBlock:...` ensure that any existing subviews are cleared.  We've historically had bugs wherein home view ends up with multiple conversations pushed.  I'm wondering if we should:

* move this logic into home view 
* have home view's `withoutAnimationPresentConversationForThread` (or whatever) method also ensure that only one conversation is pushed at a time.

WDYT?

michaelkirk-signal:
OK - I've largely re-written this PR to address some of your concerns. Can you please re-review this PR @charlesmchen ?

My first approach only optimized the NewContactThreadVC -> ConversationVC animation

My second approach instead modified `presentConversation`, which is used in myriad places, to accommodate efficient animations wherever we use it.

I then audited each call site, applied the desired animation and verified that it worked.

I've captured my rationale in the commit message, copied here:

  Faster conversation presentation.
    
    There are multiple places in the codebase we present a conversation.
    
    We used to have some very conservative machinery around how this was done, for
    fear of failing to present the call view controller, which would have left a
    hidden call in the background. We've since addressed that concern more
    thoroughly via the separate calling UIWindow.
    
    As such, the remaining presentation machinery is overly complex and inflexible
    for what we need.
    
    Sometimes we want to animate-push the conversation. (tap on home, tap on "send message" in contact card/group members)
    Sometimes we want to dismiss a modal, to reveal the conversation behind it (contact picker, group creation)
    Sometimes we want to present the conversation with no animation (becoming active from a notification)
    
    We also want to ensure that we're never pushing more than one conversation view
    controller, which was previously a problem since we were "pushing" a newly
    constructed VC in response to these myriad actions. It turned out there were
    certain code paths that caused multiple actions to be fired in rapid succession
    which pushed multiple ConversationVC's.
    
    The built-in method: `setViewControllers:animated` easily ensures we only have
    one ConversationVC on the stack, while being composable enough to faciliate the
    various more efficient animations we desire.
    
    The only thing lost with the complex methods is that the naive
    `presentViewController:` can fail, e.g. if another view is already presented.
    E.g. if an alert appears *just* before the user taps compose, the contact
    picker will fail to present.
    
    Since we no longer depend on this for presenting the CallViewController, this
    isn't catostrophic, and in fact, arguable preferable, since we want the user to
    read and dismiss any alert explicitly.

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -157,8 +157,9 @@ class ConversationSearchViewController: UITableViewController {
 
             let thread = searchResult.thread
             SignalApp.shared().presentConversation(for: thread.threadRecord,
-                                                   action: .compose,
```
michaelkirk-signal: Note: I changed this one - I think when you are searching for a particular message you are usually not intending to compose a response. WDYT?

(I did leave the compose action for the "conversation" and "contact" results)

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -744,8 +743,6 @@ - (void)displayAnyUnseenUpgradeExperience
         ExperienceUpgradesPageViewController *experienceUpgradeViewController =
             [[ExperienceUpgradesPageViewController alloc] initWithExperienceUpgrades:unseenUpgrades];
         [self presentViewController:experienceUpgradeViewController animated:YES completion:nil];
-    } else if (!self.hasEverAppeared && [ProfileViewController shouldDisplayProfileViewOnLaunch]) {
```
michaelkirk-signal: At this point this just seems like a liability. Happy to put it back if you think it's useful.

Signal/src/environment/SignalApp.h
```diff
@@ -35,16 +35,24 @@ NS_ASSUME_NONNULL_BEGIN
 
 + (instancetype)sharedApp;
 
-#pragma mark - View Convenience Methods
+#pragma mark - Conversation Presentation
+
+- (void)presentConversationForRecipientId:(NSString *)recipientId animated:(BOOL)isAnimated;
```
michaelkirk-signal: just added an `animated` flag to each of these, which ultimately gets passed to `setViewControllers:@[self, conversationVC] animated:**isAnimated**` in the HomeViewController

Signal/src/environment/SignalApp.m
```diff
@@ -152,24 +152,24 @@ - (AccountManager *)accountManager
...
-    [self presentConversationForRecipientId:recipientId action:ConversationViewActionNone];
+    [self presentConversationForRecipientId:recipientId action:ConversationViewActionNone animated:(BOOL)isAnimated];
 }
 
-- (void)presentConversationForRecipientId:(NSString *)recipientId action:(ConversationViewAction)action
+- (void)presentConversationForRecipientId:(NSString *)recipientId
+                                   action:(ConversationViewAction)action
+                                 animated:(BOOL)isAnimated
 {
-    DispatchMainThreadSafe(^{
```
michaelkirk-signal: There's already a `DispatchMainThreadSafe` check farther down the stack frame before manipulating UIKit. The OWSPrimaryStorage stuff can happen off the main thread (though in practice these methods all already assert is main thread, down the stack frame)

PR #3826, opened 2018-08-18
don't block main thread during search

michaelkirk:
The search doesn't take too long, but there's a very perceptible momentary hang while typing in your search term.

I don't think it's slow enough to introduce a progress indicator, in fact I think doing so would be more damaging than helpful since it would blip in and out.

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
+++ b/Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
@@ -379,17 +379,21 @@ class ConversationSearchViewController: UITableViewController {
         }
 
         var searchResults: SearchResultSet?
-        self.uiDatabaseConnection.asyncRead({ transaction in
-            searchResults = self.searcher.results(searchText: searchText, transaction: transaction, contactsManager: self.contactsManager)
+        self.uiDatabaseConnection.asyncRead({[weak self] transaction in
+            guard let strongSelf = self else { return }
+            searchResults = strongSelf.searcher.results(searchText: searchText, transaction: transaction, contactsManager: strongSelf.contactsManager)
         },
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -378,12 +378,19 @@ class ConversationSearchViewController: UITableViewController {
...
 
-        self.uiDatabaseConnection.read { transaction in
-            self.searchResultSet = self.searcher.results(searchText: searchText, transaction: transaction, contactsManager: self.contactsManager)
-        }
-
-        // TODO: more performant way to do this?
-        self.tableView.reloadData()
+        var searchResults: SearchResultSet?
+        self.uiDatabaseConnection.asyncRead({ transaction in
+            searchResults = self.searcher.results(searchText: searchText, transaction: transaction, contactsManager: self.contactsManager)
```
charlesmchen-signal: Should we do a `weakSelf` thing here?

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -378,12 +378,19 @@ class ConversationSearchViewController: UITableViewController {
...
+        },
+                                             completionBlock: {
+                                                guard let results = searchResults else {
```
charlesmchen-signal: And again, separately, here?

charlesmchen-signal: let's assert that we're on the main thread here?

michaelkirk-signal:
In aabf9e79e

PR #3828, opened 2018-08-18
"Tap to retry" retries, rather than maybe deletes.

michaelkirk:
Attachment downloads can fail on slow networks or if the app crashes.
It's unlikely that the users response to this would be to delete the
attachment pointer, so it doesn't make sense to surface that as a
primary action.

If a user does want to delete the attachment, as always they can
long-press to delete.

PTAL @charlesmchen


michaelkirk-signal:
In 5907cd2a0

PR #3830, opened 2018-08-20
Disable CDS.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
+++ b/SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
@@ -7,6 +7,8 @@ import Foundation
 @objc(OWSLegacyContactDiscoveryOperation)
 class LegacyContactDiscoveryBatchOperation: OWSOperation {
 
+    private let isCDSEnabled = false
+
     @objc
     var registeredRecipientIds: Set<String>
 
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Contacts/OWSContactDiscoveryOperation.swift
```diff
@@ -83,13 +85,15 @@ class LegacyContactDiscoveryBatchOperation: OWSOperation {
...
     // Called at most one time.
     override func didSucceed() {
-        // Compare against new CDS service
-        let modernCDSOperation = CDSOperation(recipientIdsToLookup: self.recipientIdsToLookup)
-        let cdsFeedbackOperation = CDSFeedbackOperation(legacyRegisteredRecipientIds: self.registeredRecipientIds)
-        cdsFeedbackOperation.addDependency(modernCDSOperation)
-
-        let operations = modernCDSOperation.dependencies + [modernCDSOperation, cdsFeedbackOperation]
-        CDSOperation.operationQueue.addOperations(operations, waitUntilFinished: false)
+        if isCDSEnabled {
```
michaelkirk-signal: nit: for more semantic diffs, a guard might be better, but TIOLI obvs

charlesmchen-signal: Done.

PR #3833, opened 2018-08-21
Theme quoted replies.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/utils/ConversationStyle.swift
+++ b/SignalMessaging/utils/ConversationStyle.swift
@@ -228,7 +228,7 @@ public class ConversationStyle: NSObject {
     @objc
     public func quotedReplyBubbleColor(isIncoming: Bool) -> UIColor {
         if Theme.isDarkThemeEnabled {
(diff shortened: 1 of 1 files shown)


SignalMessaging/utils/ConversationStyle.swift
```diff
@@ -227,7 +227,10 @@ public class ConversationStyle: NSObject {
 
     @objc
     public func quotedReplyBubbleColor(isIncoming: Bool) -> UIColor {
-        if isIncoming {
+        if Theme.isDarkThemeEnabled {
+            let alpha: CGFloat = (isIncoming ?0.6 :0.5)
```
michaelkirk-signal: format nit: missing space after `?`

charlesmchen-signal: Done. I don't know why clang-format doesn't work better.  When we get a chance, I'd be up for switching formatters.

michaelkirk-signal: clang-format doesn't format swift code, does it?

PR #3835, opened 2018-08-22
Mkirk/search bar theme 2

michaelkirk-signal:
PTAL @charlesmchen 

Rather than our custom class, we can get pretty far using existing UISearchBar
styling, and then apply the final tweaks, the search bar text field background
color by traversing the view.

This fixes two small issues:

1. There was a small wavering in height of the custom search bar
2. When your table header view is a UISearchBar you get nice scroll "snapping"
   when showing/hiding the search bar.

screen shot 2018-08-22 at 10 50 37 am

screen shot 2018-08-22 at 10 50 35 am


Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
```diff
@@ -429,5 +428,8 @@ class EmptySearchResultCell: UITableViewCell {
         let format = NSLocalizedString("HOME_VIEW_SEARCH_NO_RESULTS_FORMAT", comment: "Format string when search returns no results. Embeds {{search term}}")
         let messageText: String = NSString(format: format as NSString, searchText) as String
         self.messageLabel.text = messageText
+
+        messageLabel.textColor = Theme.primaryColor
+        messageLabel.font = UIFont.ows_dynamicTypeBody
```
charlesmchen-signal: This is totally unrelated, right?

SignalMessaging/utils/UIUtil.m
```diff
@@ -41,6 +41,9 @@ + (void)setupSignalAppearence
     //                                                 ? UIKeyboardAppearanceDark
     //                                                 : UIKeyboardAppearanceDefault);
 
+    //    [[UIBarButtonItem appearanceWhenContainedIn:[UISearchBar class], nil] setTintColor:[UIColor
+    //    ows_materialBlueColor]];
```
charlesmchen-signal: ?

SignalMessaging/utils/UIImage+OWS.swift
```diff
@@ -0,0 +1,30 @@
...
+            return nil
+        }
+
+        view.layer.render(in: context)
+        let image = UIGraphicsGetImageFromCurrentImageContext()
+        return image
+    }
+
+    @objc
+    public func templated(color: UIColor) -> UIImage? {
```
charlesmchen-signal: I'd propose `asTintedView`?

Signal/src/ViewControllers/HomeView/ConversationSearchViewController.swift
michaelkirk-signal: Yup - unrelated, just noticed while testing that the cell text was not visible in dark theme.

SignalMessaging/utils/UIUtil.m
michaelkirk-signal: Ah, came back with the 'Revert "Add custom themed search bar."'

I'll remove it.

SignalMessaging/utils/UIImage+OWS.swift
michaelkirk-signal: Yeah, that's better. I struggled with this one. =)

PR #3836, opened 2018-08-22
Clean up Swift main thread assertion

michaelkirk-signal:
PTAL @charlesmchen 

For 2.30

### DIFF
--- a/SignalServiceKit/src/Util/OWSAsserts.h
+++ b/SignalServiceKit/src/Util/OWSAsserts.h
@@ -151,9 +151,6 @@ __attribute__((annotate("returns_localized_nsstring"))) static inline NSString *
     return s;
 }
 
-// This function is intended for use in Swift.
-void SwiftAssertIsOnMainThread(NSString *functionName);
-
 #define OWSRaiseException(name, formatParam, ...)                                                                      \
     do {                                                                                                               \
         DDLogError(@"Exception: %@ %@", name, [NSString stringWithFormat:formatParam, ##__VA_ARGS__]);                 \
--- a/SignalServiceKit/src/Util/OWSSwiftUtils.swift
+++ b/SignalServiceKit/src/Util/OWSSwiftUtils.swift
@@ -20,3 +20,15 @@ public func owsFail(_ message: String) {
     Logger.flush()
     assertionFailure(message)
 }
+
+// Once we're on Swift4.2 we can mark this as inlineable
+// @inlinable
+public func AssertIsOnMainThread(file: String = #file,
+                                 function: String = #function,
+                                 line: Int = #line) {
+    guard Thread.isMainThread else {
+        let filename = (file as NSString).lastPathComponent
+        owsFail("\(filename):\(line) in \(function): Must be on main thread.")
+        return
+    }
+}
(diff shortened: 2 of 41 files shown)


SignalServiceKit/src/Util/OWSSwiftUtils.swift
```diff
@@ -20,3 +20,16 @@ public func owsFail(_ message: String) {
     Logger.flush()
     assertionFailure(message)
 }
+
+// Once we're on Swift4.2 we can mark this as inlineable
+// @inlinable
+@objc
+public func AssertIsOnMainThread(file: String = #file,
```
michaelkirk-signal: substantive change 2: add this swift version. (all other changes are either a straight search/replace auto-formatting)

We can leverage the `#file`/`#function`/`#line` as default attributes to make a cleaner call site.

(i.e. `AssertIsOnMainThread()` vs `AssertIsOnMainThread(#function)`)

I'm considering doing something similar for `owsFail` and `Logger.foo` methods so we can stop doing the `"\(logTag) in \(#function)"` dance, but that would require a slightly more hands on approach. WDYT @charlesmchen?

SignalServiceKit/src/Util/OWSAsserts.h
```diff
@@ -151,9 +151,6 @@ __attribute__((annotate("returns_localized_nsstring"))) static inline NSString *
     return s;
 }
 
-// This function is intended for use in Swift.
-void SwiftAssertIsOnMainThread(NSString *functionName);
```
michaelkirk-signal: substantive change 1: remove this function

SignalServiceKit/src/Util/OWSSwiftUtils.swift
charlesmchen-signal: Noice

> I'm considering doing something similar for owsFail and Logger.foo methods so we can stop doing the "\(logTag) in \(#function)" dance, but that would require a slightly more hands on approach. WDYT @charlesmchen?

I support this.  

I assume `#file/#function/#line` work in free functions and don't capture a reference to self if used in a block.

charlesmchen-signal: Can this method be called from obj-c code?  If so, I suggest we call it `SwiftAssertIsOnMainThread` to help visually distinguish it from `OWSAssertIsOnMainThread()`.

charlesmchen-signal:
> I'm considering doing something similar for owsFail and Logger.foo methods so we can stop doing the "\(logTag) in \(#function)" dance, but that would require a slightly more hands on approach. WDYT @charlesmchen?

I'm going to work on the first part of this (`owsFail()`) since I'm already touching that function.

SignalServiceKit/src/Util/OWSSwiftUtils.swift
```diff
@@ -20,3 +20,16 @@ public func owsFail(_ message: String) {
     Logger.flush()
     assertionFailure(message)
 }
+
+// Once we're on Swift4.2 we can mark this as inlineable
+// @inlinable
+@objc
```
charlesmchen-signal: @michaelkirk I'm getting a build error doing something similar in the "prod fail" methods.

> @objc can only be used with members of classes, @objc protocols, and concrete extensions of classes

SignalServiceKit/src/Util/OWSSwiftUtils.swift
michaelkirk-signal: I removed the `@objc`. That was a mistake, I only intended it for Swift.

PR #3837, opened 2018-08-22
Cannot block self

michaelkirk-signal:
For 2.30

PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/AppSettings/AddToBlockListViewController.m
+++ b/Signal/src/ViewControllers/AppSettings/AddToBlockListViewController.m
@@ -6,7 +6,9 @@
 #import "BlockListUIUtils.h"
 #import "ContactsViewHelper.h"
 #import <SignalMessaging/OWSContactsManager.h>
+#import <SignalMessaging/SignalMessaging-Swift.h>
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/AppSettings/AddToBlockListViewController.m
```diff
@@ -78,6 +80,12 @@ - (void)signalAccountWasSelected:(SignalAccount *)signalAccount
         OWSFail(@"%@ Cannot add already blocked user to block list.", self.logTag);
         return;
     }
+
+    if ([signalAccount.recipientId isEqual:TSAccountManager.localNumber]) {
+        [self showCannotBlockSelfAlert];
+        return;
+    }
+
     [BlockListUIUtils showBlockSignalAccountActionSheet:signalAccount
```
charlesmchen-signal: Weird. It looks like `BlockListUIUtils showBlockSignalAccountActionSheet` was already checking for this.  Was that check not working?

michaelkirk-signal:
Seems like we're already doing this, so I'm going to close this PR.

PR #3842, opened 2018-08-23
Charlesmchen/reworking logging

charlesmchen-signal:
Streamlines Swift logging. 

* We should probably do something similar for Obj-C, e.g. replace `DDLogError()` with `OWSLogError()` which uses `__PRETTY_FUNCTION__`.

There's two (tiny) commits that have the real changes.  Everything else is "mechanical".

* https://github.com/signalapp/Signal-iOS/commit/89ee28721cd44cf13ef2a3c22c9dee0147ac8c45
* https://github.com/signalapp/Signal-iOS/commit/df2d3db27ccf899055c38569e5b26f3acd669d1e

PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/ViewControllers/AttachmentApprovalViewController.swift
+++ b/SignalMessaging/ViewControllers/AttachmentApprovalViewController.swift
@@ -67,7 +67,7 @@ public class AttachmentApprovalViewController: OWSViewController, CaptioningTool
         let navController = OWSNavigationController(rootViewController: vc)
 
         guard let navigationBar = navController.navigationBar as? OWSNavigationBar else {
-            owsFail("\(self.logTag) in \(#function) navigationBar was nil or unexpected class")
+            owsFail("navigationBar was nil or unexpected class")
             return navController
         }
         navigationBar.makeClear()
@@ -76,7 +76,7 @@ public class AttachmentApprovalViewController: OWSViewController, CaptioningTool
     }
 
     override public func viewWillLayoutSubviews() {
-        Logger.debug("\(logTag) in \(#function)")
+        Logger.debug("")
         super.viewWillLayoutSubviews()
 
         // e.g. if flipping to/from landscape
@@ -84,22 +84,20 @@ public class AttachmentApprovalViewController: OWSViewController, CaptioningTool
     }
 
     override public func viewWillAppear(_ animated: Bool) {
-        Logger.debug("\(logTag) in \(#function)")
+        Logger.debug("")
         super.viewWillAppear(animated)
 
         CurrentAppContext().setStatusBarHidden(true, animated: animated)
     }
 
     override public func viewDidAppear(_ animated: Bool) {
-        Logger.debug("\(logTag) in \(#function)")
-        DispatchQueue.global().async {
-            AssertIsOnMainThread()
-        }
+        Logger.debug("")
+
         super.viewDidAppear(animated)
     }
 
     override public func viewWillDisappear(_ animated: Bool) {
-        Logger.debug("\(logTag) in \(#function)")
+        Logger.debug("")
         super.viewWillDisappear(animated)
 
         // Since this VC is being dismissed, the "show status bar" animation would feel like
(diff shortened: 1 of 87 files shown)


charlesmchen-signal:
> Can you clean up these commits?

Done.

charlesmchen-signal:
> I'm not convinced that this is helpful.

Yeah, I've removed that.

charlesmchen-signal:
> Can you make all the logString/message parameters utilize autoclosure?

I've done this for the "verbose" and "debug" level log methods and used the preprocessor to ensure that we don't evaluate those autoclosures in production.  

But there's no value in using `@autoclosure` for the other log levels since we'll always evaluate them, right?  And `@autoclosure` isn't free - you'll notice that we now have to be explicit about references to self in our logging.

charlesmchen-signal:
PTAL @michaelkirk

michaelkirk-signal:
> But there's no value in using @autoclosure for the other log levels since we'll always evaluate them, right? And @autoclosure isn't free - you'll notice that we now have to be explicit about references to self in our logging.

There's the value of consistency - in my mind that's worth more than not having to write `self` in top-level logging (we'll still have to use `self` when logging in blocks anyway), but I'm willing to let that one go.

SignalServiceKit/src/Util/OWSLogger.swift
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * A minimal DDLog wrapper for swift.
+ */
+open class Logger: NSObject {
+
+    open class func verbose(_ logString: @escaping @autoclosure () -> String,
+                            file: String = #file,
+                            function: String = #function,
+                            line: Int = #line) {
+        #if DEBUG
```
michaelkirk-signal: My proposal is that this looks something like:
```
guard ddLogLevel <= verbose else {
  return
}
let message = owsFormatLogMessage(logString(), file: file, function: function, line: line)
OWSLogger.verbose(message)
```

Then remove the #if DEBUG and change the params of the OWSLogger to be a String, not a (()->String)

Also, can you rename this file to Logger.swift?

SignalServiceKit/src/Util/OWSLogger.swift
```diff
@@ -21,20 +21,24 @@ public func owsFormatLogMessage(_ logString: String,
  */
 open class Logger: NSObject {
 
-    open class func verbose(_ logString: String,
+    open class func verbose(_ logString: @autoclosure () -> String,
                             file: String = #file,
                             function: String = #function,
                             line: Int = #line) {
-        let message = owsFormatLogMessage(logString, file: file, function: function, line: line)
+        #if DEBUG
```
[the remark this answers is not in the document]
charlesmchen-signal: Done, but we should do a perf analysis.  The advantage of the preprocessor is that the code compiles out entirely.

charlesmchen-signal:
> There's the value of consistency - in my mind that's worth more than not having to write self in top-level logging (we'll still have to use self when logging in blocks anyway), but I'm willing to let that one go.

Done.

SignalServiceKit/src/Util/OWSLogger.swift
charlesmchen-signal: Done, Done, Done, Done.

charlesmchen-signal:
PTAL @michaelkirk

michaelkirk-signal:
Have you pushed your latest changes? I'm still seeing the old code (e.g. you said you renamed OWSLogger.swift -> Logger.swift)

charlesmchen-signal:
> Have you pushed your latest changes? I'm still seeing the old code (e.g.
you said you renamed OWSLogger.swift -> Logger.swift)

Whoops. I've pushed.

SignalServiceKit/src/Util/OWSLogger.m
```diff
@@ -8,27 +8,27 @@
 
 @implementation OWSLogger
 
-+ (void)verbose:(NSString *)logString
++ (void)verbose:(NSString *)logString;
```
michaelkirk-signal: please remove these semicolons

SignalMessaging/ViewControllers/AttachmentApprovalViewController.swift
```diff
@@ -76,30 +76,28 @@ public class AttachmentApprovalViewController: OWSViewController, CaptioningTool
...
-        Logger.debug("\(logTag) in \(#function)")
+        Logger.debug("")
         super.viewWillAppear(animated)
 
         CurrentAppContext().setStatusBarHidden(true, animated: animated)
     }
 
     override public func viewDidAppear(_ animated: Bool) {
-        Logger.debug("\(logTag) in \(#function)")
-        DispatchQueue.global().async {
```
michaelkirk-signal: Sorry - left this in while testing assertion failures. I guess you found it. =)

SignalServiceKit/src/Util/OWSLogger.m
charlesmchen-signal: Done.

PR #3843, opened 2018-08-23
Charlesmchen/prod fail

charlesmchen-signal:
I didn't apply these changes, since I think we should discuss exactly how this works.

PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Util/OWSAsserts.h
+++ b/SignalServiceKit/src/Util/OWSAsserts.h
@@ -15,8 +15,8 @@ NS_ASSUME_NONNULL_BEGIN
 #define CONVERT_TO_STRING(X) #X
 #define CONVERT_EXPR_TO_STRING(X) CONVERT_TO_STRING(X)
 
-// OWSAssert() and OWSFail() should be used in Obj-C methods.
-// OWSCAssert() and OWSCFail() should be used in free functions.
+// OWSAssert() and OWSFailDebug() should be used in Obj-C methods.
+// OWSCAssert() and OWSCFailDebug() should be used in free functions.
 
 #define OWSAssert(X)                                                                                                   \
     do {                                                                                                               \
@@ -76,7 +76,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 #endif
 
-#define OWSAbstractMethod() OWSFail(@"%@ Method needs to be implemented by subclasses.", self.logTag)
+#define OWSAbstractMethod() OWSFailDebug(@"%@ Method needs to be implemented by subclasses.", self.logTag)
 
 #pragma mark - Singleton Asserts
 
@@ -124,14 +124,14 @@ NS_ASSUME_NONNULL_BEGIN
 // This macro is intended for use in Objective-C.
 #define OWSAssertIsOnMainThread() OWSCAssert([NSThread isMainThread])
 
-#define OWSFail(_messageFormat, ...)                                                                                   \
+#define OWSFailDebug(_messageFormat, ...)                                                                              \
     do {                                                                                                               \
         DDLogError(_messageFormat, ##__VA_ARGS__);                                                                     \
         [DDLog flushLog];                                                                                              \
         OWSFailWithoutLogging(_messageFormat, ##__VA_ARGS__);                                                          \
     } while (0)
 
-#define OWSCFail(_messageFormat, ...)                                                                                  \
+#define OWSCFailDebug(_messageFormat, ...)                                                                             \
     do {                                                                                                               \
         DDLogError(_messageFormat, ##__VA_ARGS__);                                                                     \
         [DDLog flushLog];                                                                                              \
@@ -139,15 +139,15 @@ NS_ASSUME_NONNULL_BEGIN
     } while (NO)
 
 
-#define OWSProdExit(_messageFormat, ...)                                                                               \
+#define OWSFail(_messageFormat, ...)                                                                                   \
     do {                                                                                                               \
-        OWSFail(_messageFormat, ##__VA_ARGS__);                                                                        \
+        OWSFailDebug(_messageFormat, ##__VA_ARGS__);                                                                   \
         exit(0);                                                                                                       \
     } while (0)
 
-#define OWSCProdExit(_messageFormat, ...)                                                                              \
+#define OWSCFail(_messageFormat, ...)                                                                                  \
     do {                                                                                                               \
-        OWSCFail(_messageFormat, ##__VA_ARGS__);                                                                       \
+        OWSCFailDebug(_messageFormat, ##__VA_ARGS__);                                                                  \
         exit(0);                                                                                                       \
     } while (NO)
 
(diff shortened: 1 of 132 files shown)


SignalServiceKit/src/Util/OWSAsserts.h
```diff
@@ -139,6 +139,18 @@ NS_ASSUME_NONNULL_BEGIN
...
+    do {                                                                                                               \
+        OWSFail(_messageFormat, ##__VA_ARGS__);                                                                        \
+        exit(0);                                                                                                       \
+    } while (0)
+
+#define OWSCProdExit(_messageFormat, ...)                                                                              \
+    do {                                                                                                               \
+        OWSCFail(_messageFormat, ##__VA_ARGS__);                                                                       \
+        exit(0);                                                                                                       \
+    } while (NO)
```
charlesmchen-signal: Options:

* We could add these new `OWSExit()` macros and then migrate (most?) `OWSFail()` macros to use them.  
* Same, but we could call them `OWSProdFail()` and rename the existing analytics macros with those names to something else.
* We could simply call the exiting macros `OWSFail()` and rename the existing macros to `OWSFailWithoutExit()` or whatever.  WDYT?

SignalServiceKit/src/Util/OWSSwiftUtils.swift
```diff
@@ -15,8 +15,41 @@ public func assertOnQueue(_ queue: DispatchQueue) {
...
+// Once we're on Swift4.2 we can mark this as inlineable
+// @inlinable
+public func owsProdFail(_ rawMessage: String,
+                        file: String = #file,
+                        function: String = #function,
+                        line: Int = #line) {
+    let message = "\(file) \(function) \(line): \(rawMessage)"
+    Logger.error(message)
+    Logger.flush()
+    // TODO: Should we use this or fatalError()?
```
charlesmchen-signal: I haven't researched this question yet.

SignalServiceKit/src/Util/OWSSwiftUtils.swift
```diff
@@ -15,8 +15,41 @@ public func assertOnQueue(_ queue: DispatchQueue) {
...
+                    line: Int = #line) {
+    let message = "\(file) \(function) \(line): \(rawMessage)"
+    Logger.error(message)
+    Logger.flush()
+    assertionFailure(message)
+}
+
+// Once we're on Swift4.2 we can mark this as inlineable
+// @inlinable
+public func owsProdFail(_ rawMessage: String,
```
charlesmchen-signal: Presumably the naming of these methods will align with our decisions for the Obj-C macros.

SignalServiceKit/src/Util/OWSSwiftUtils.swift
```diff
@@ -15,8 +15,41 @@ public func assertOnQueue(_ queue: DispatchQueue) {
...
     assertionFailure(message)
 }
+
+// Once we're on Swift4.2 we can mark this as inlineable
+// @inlinable
+public func owsFail(_ rawMessage: String,
+                    file: String = #file,
+                    function: String = #function,
+                    line: Int = #line) {
+    let message = "\(file) \(function) \(line): \(rawMessage)"
```
charlesmchen-signal: We'll actually format these messages using the `owsFormatLogMessage()` function defined in:

https://github.com/signalapp/Signal-iOS/pull/3842

michaelkirk-signal:
It seems like these changes are also in https://github.com/signalapp/Signal-iOS/pull/3843?

I left some comments in that PR related to this. Can you separate them? or base one on the other so it's clear what I'm reviewing?

charlesmchen-signal:
> It seems like these changes are also in #3843?

Sorry about that. PTAL @michaelkirk

SignalServiceKit/src/Util/OWSSwiftUtils.swift
```diff
@@ -32,3 +26,27 @@ public func AssertIsOnMainThread(file: String = #file,
...
+    // TODO: Format using owsFormatLogMessage() once it is merged.
+    let message = "\(file) \(function) \(line): \(rawMessage)"
+    Logger.error(message)
+    Logger.flush()
+    assertionFailure(message)
+}
+
+// Once we're on Swift4.2 we can mark this as inlineable
+// @inlinable
+public func owsProdExit(_ rawMessage: String,
```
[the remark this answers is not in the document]
michaelkirk-signal: Swift actually provides a nice set of assertions for debug and production, but as far as I can tell there's no way to plug them into our logging.

SignalServiceKit/src/Util/OWSSwiftUtils.swift
```diff
@@ -32,3 +26,27 @@ public func AssertIsOnMainThread(file: String = #file,
...
+                        file: String = #file,
+                        function: String = #function,
+                        line: Int = #line) {
+
+    owsFail(rawMessage, file: file, function: function, line: line)
+    fatalError()
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

SignalServiceKit/src/Util/OWSSwiftUtils.swift
```diff
@@ -32,3 +26,27 @@ public func AssertIsOnMainThread(file: String = #file,
...
 }
+
+// Once we're on Swift4.2 we can mark this as inlineable
+// @inlinable
+public func owsFail(_ rawMessage: String,
+                    file: String = #file,
+                    function: String = #function,
+                    line: Int = #line) {
+    // TODO: Format using owsFormatLogMessage() once it is merged.
+    let message = "\(file) \(function) \(line): \(rawMessage)"
```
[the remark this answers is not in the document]
charlesmchen-signal: Yeah this is just temporary until `owsFormatLogMessage()` is merged in the logging PR.

SignalServiceKit/src/Util/OWSSwiftUtils.swift
charlesmchen-signal: Done.

SignalServiceKit/src/Util/OWSSwiftUtils.swift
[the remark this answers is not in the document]
charlesmchen-signal: Great question. It turns out we can return `-> Never`.  I've resurrected `notImplemented()` to take advantage of this and applied this in the "exiting" Swift fail.

SignalServiceKit/src/Util/OWSAsserts.h
charlesmchen-signal: Done, with `OWSFailDebug()`.

michaelkirk-signal:
>  charlesmchen-signal closed this an hour ago

Was this merged? I don't see the commit.

charlesmchen-signal:
@michaelkirk do you see it now?

77711df27

michaelkirk:
Yes, thanks!

PR #3845, opened 2018-08-27
Charlesmchen/objc logging

charlesmchen-signal:
This is the interesting commit:  https://github.com/signalapp/Signal-iOS/commit/d81dea1d84eff5d78683e04843b2b7b3a8d20f25

95% of the other changes were done with regular expressions.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -78,23 +78,23 @@ static NSTimeInterval launchStartedAt;
 @synthesize window = _window;
 
 - (void)applicationDidEnterBackground:(UIApplication *)application {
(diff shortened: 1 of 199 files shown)


charlesmchen-signal:
Two things:

* `self.logTag` was always a risk for introducing memory leaks via retain cycles.  The new log macros don't have that risk.

* The new log prefix has both the `file:line` prefix (for "Open Quickly..." and to disambiguate similar logging within a function or method) and the "pretty function" for readability.  The logs are more chatty now, but I think it's worth it.  WDYT?

michaelkirk-signal:
LGTM

PR #3855, opened 2018-08-31
Add test app context; use mock "document" and "shared data container"…

charlesmchen-signal:
… directories in tests, use mock keychain storage in tests.

* The test app context lets us simplify and control the test environment.
* Unit tests benefit from running in isolation.  By using mock file system (these are the two key directories used by the app) and keychain, we can isolate _most_ of the data context for the tests.  In a follow-on PR we'll let tests replace & mock their environments.

PTAL @michaelkirk


Signal/src/ViewControllers/Registration/RegistrationViewController.m
```diff
@@ -580,8 +581,10 @@ - (void)setDebugValue:(NSString *)value forKey:(NSString *)key
     OWSCAssert(value.length > 0);
 
     NSError *error;
-    [SAMKeychain setAccessibilityType:kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly];
```
charlesmchen-signal: One small side benefit: we DRY up usage of `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`.

SignalShareExtension/utils/ShareAppExtensionContext.m
```diff
@@ -212,6 +214,26 @@ - (void)runNowOrWhenMainAppIsActive:(AppActiveBlock)block
...
+
+- (NSString *)appDocumentDirectoryPath
+{
+    NSFileManager *fileManager = [NSFileManager defaultManager];
+    NSURL *documentDirectoryURL =
+        [[fileManager URLsForDirectory:NSDocumentDirectory inDomains:NSUserDomainMask] lastObject];
+    return [documentDirectoryURL path];
+}
+
+- (NSString *)appSharedDataDirectoryPath
```
charlesmchen-signal: I haven't yet added mocking of the `NSUserDefaults`.

SignalServiceKit/src/Storage/KeychainStorage.swift
```diff
@@ -0,0 +1,103 @@
...
+    @objc public static let sharedInstance = SSKKeychainStorage()
+
+    // Force usage as a singleton
+    override private init() {
+        super.init()
+
+        SwiftSingletons.register(self)
+    }
+
+    @objc public func string(forKey key: String, service: String) throws -> String {
```
michaelkirk-signal: TIOLI, but I'd prefer to maintain the order of params that SAMKeychain uses, so `service` first then `key`

The service represents a "bag" of potentially multiple keys.

e.g. if you think about it as a dictionary access, you might write something like:

```
let s3Password = allPasswords["amazon"]["s3"]
```

Not:

```
let s3Password = allPasswords["s3"]["amazon"]
```

SignalServiceKit/src/Storage/KeychainStorage.swift
```diff
@@ -0,0 +1,103 @@
...
+        var error: NSError?
+        let result = SAMKeychain.password(forService: service, account: key, error: &error)
```
michaelkirk-signal: huh - I'm surprised this didn't get interpreted by Swift as a throwable method, but I'm glad our wrapper throws.

SignalServiceKit/src/Storage/KeychainStorage.swift
```diff
@@ -0,0 +1,103 @@
...
+        if let error = error {
+            throw SSKProtoError.invalidProtobuf(description: "\(logTag) error retrieving string: \(error)")
```
michaelkirk-signal: I don't think these are supposed to be SSKProtoError's, are they?

TIOLI: I often make a private factory method like:

```
private func keychainError(_ description: String) -> KeyChainError {
    return KeyChainError.failure(description: description)
}
```

SignalServiceKit/src/Tests/TestKeychainStorage.m
```diff
@@ -0,0 +1,113 @@
...
+    *error = nil;
+
+    NSLog(@"setWithString:%@ service:%@ -> %@", key, service, string);
+
+    NSString *mapKey = [NSString stringWithFormat:@"%@-%@", service, key];
+    self.dataMap[mapKey] = [string dataUsingEncoding:NSUTF8StringEncoding];
+    return YES;
+}
+
+- (NSData *_Nullable)dataForKey:(NSString *)key service:(NSString *)service error:(NSError *_Nullable *_Nullable)error
```
michaelkirk-signal: we don't require an error param to be passed in? We are dereferencing it on line 70.

SignalServiceKit/src/Util/AppContext.m
```diff
@@ -43,6 +43,11 @@ void SetCurrentAppContext(id<AppContext> appContext)
     currentAppContext = appContext;
 }
 
+void ClearCurrentAppContextForTests()
```
michaelkirk-signal: #ifdef DEBUG

Signal/src/ViewControllers/Registration/RegistrationViewController.m
```diff
@@ -566,11 +566,12 @@ - (NSString *_Nullable)debugValueForKey:(NSString *)key
     OWSCAssert(key.length > 0);
 
     NSError *error;
-    NSString *value = [SAMKeychain passwordForService:kKeychainService_LastRegistered account:key error:&error];
-    if (value && !error) {
-        return value;
+    NSString *_Nullable value =
+        [CurrentAppContext().keychainStorage stringForKey:key service:kKeychainService_LastRegistered error:&error];
+    if (error || !value) {
+        DDLogWarn(@"Could not retrieve 'last registered' value from keychain: %@.", error);
```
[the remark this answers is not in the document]
charlesmchen-signal: Done. Note that this is DEBUG-only code.

SignalServiceKit/src/Util/AppContext.m
charlesmchen-signal: Done.

SignalServiceKit/src/Tests/TestKeychainStorage.m
charlesmchen-signal: I've just added conditions, but there are already non-nil asserts at the start of each of these methods.  We always want to check for errors in our tests, but this class needs to comply with the method signatures in the (generated) interface.

SignalServiceKit/src/Storage/KeychainStorage.swift
charlesmchen-signal: > I don't think these are supposed to be SSKProtoError's, are they?

Done.

SignalServiceKit/src/Storage/KeychainStorage.swift
charlesmchen-signal: Done. I think `TestAppContext` is now broken, but until more of these branches are merged I can't fix it.

SignalServiceKit/src/Tests/TestAppContext.m
```diff
@@ -0,0 +1,153 @@
...
+{
+    return nil;
+}
+
+- (nullable UIAlertAction *)openSystemSettingsAction
+{
+    return nil;
+}
+
+- (void)doMultiDeviceUpdateWithProfileKey:(OWSAES256Key *)profileKey
```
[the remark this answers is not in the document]
charlesmchen-signal: I agree.

SignalServiceKit/src/Tests/TestAppContext.m
```diff
@@ -0,0 +1,153 @@
...
+
+#import "TestAppContext.h"
+#import "TestKeychainStorage.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface TestAppContext ()
+
+@property (nonatomic) TestKeychainStorage *testKeychainStorage;
+@property (nonatomic) NSString *appDocumentDirPath;
```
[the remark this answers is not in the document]
charlesmchen-signal: I renamed these properties to include a `mock` prefix.

SignalServiceKit/src/Tests/TestKeychainStorage.m
```diff
@@ -0,0 +1,113 @@
...
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+#import "TestKeychainStorage.h"
+#import "NSData+OWS.h"
+#import <SignalServiceKit/SignalServiceKit-Swift.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@implementation TestKeychainStorage
```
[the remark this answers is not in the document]
charlesmchen-signal: I'll convert it to Swift if I get a chance once the tests are building again.

Signal/src/ViewControllers/Registration/RegistrationViewController.m
```diff
@@ -566,11 +566,12 @@ - (NSString *_Nullable)debugValueForKey:(NSString *)key
     OWSCAssert(key.length > 0);
 
     NSError *error;
-    NSString *value = [SAMKeychain passwordForService:kKeychainService_LastRegistered account:key error:&error];
-    if (value && !error) {
-        return value;
+    NSString *_Nullable value =
+        [CurrentAppContext().keychainStorage stringForKey:key service:kKeychainService_LastRegistered error:&error];
```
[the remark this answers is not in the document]
charlesmchen-signal: That's a good question.  I did this before we settled (if we have) on the new approach of "build all singletons explicitly, in order of dependency, mostly in `AppSetup`, injecting dependencies into initializers, and populate the environments when they're all ready."  

Now that we have that approach, it'd be better to hang the "keychain storage" on an environment.  Let's make that change once the existing work is done in order to avoid rebase hell.

charlesmchen-signal:
PTAL @michaelkirk

charlesmchen-signal:
PTAL @michaelkirk

charlesmchen-signal:
PTAL @michaelkirk =)

SignalServiceKit/src/Tests/TestAppContext.m
```diff
@@ -0,0 +1,153 @@
...
+}
+
+- (id<KeychainStorage>)keychainStorage
+{
+    return self.testKeychainStorage;
+}
+
+- (NSString *)appDocumentDirectoryPath
+{
+    return self.mockAppDocumentDirectoryPath
```
michaelkirk-signal: missing semicolon - can you fix and make sure any other build breakages are addressed?

SignalServiceKit/src/Storage/KeychainStorage.swift
michaelkirk-signal: In general, I'd prefer that you make these PR's not predicated on the future presumption that future PR's are going to merged in a timely way. I'm hoping that each PR can be a useful atomic step forward. 

I know it takes more of *your* time to prepare PR's in that way, but I believe it allows for a better review process and cleaner git history.

SignalServiceKit/src/Storage/KeychainStorage.swift
```diff
@@ -0,0 +1,103 @@
...
+    case failure(description: String)
+}
+
+// MARK: -
+
+@objc public protocol KeychainStorage: class {
+
+    @objc func string(forKey key: String, service: String) throws -> String
+
+    @objc func set(string: String, forKey key: String, service: String) throws
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

SignalServiceKit/src/Tests/TestKeychainStorage.m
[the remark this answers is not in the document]
charlesmchen-signal: I will do this, but I want to do it when the tests are building again, and we're a few PRs away from that.  https://trello.com/c/VnXo73E8/3547-convert-testkeychainstorage-to-swift

SignalServiceKit/src/Tests/TestAppContext.m
charlesmchen-signal: Done.

SignalServiceKit/src/Storage/KeychainStorage.swift
```diff
@@ -0,0 +1,103 @@
...
+import Foundation
+import SAMKeychain
+
+public enum KeychainStorageError: Error {
+    case failure(description: String)
+}
+
+// MARK: -
+
+@objc public protocol KeychainStorage: class {
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

SignalServiceKit/src/Tests/TestKeychainStorage.m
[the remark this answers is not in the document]
charlesmchen-signal: Okay, I did it now.

SignalServiceKit/src/Storage/KeychainStorage.swift
```diff
@@ -15,7 +15,7 @@ public enum KeychainStorageError: Error {
 
     @objc func string(forService service: String, key: String) throws -> String
 
-    @objc func set(string: String, service: String, key: String) throws
+    @objc(setString:service:key:error:) func set(string: String, service: String, key: String) throws
```
michaelkirk-signal: `@objc` on it's own line please.

PR #3856, opened 2018-08-31
Charlesmchen/fix ssk tests d

charlesmchen-signal:
Very mechanical changes.  

* Rename `TextSecureKitEnv` to `SSKEnvironment`.
* Modify environment accessors to use our 'shared' convention.

First steps towards making `Environment` and `SSKEnvironment` more consistent and consistent with our other code.

PTAL @michaelkirk

### DIFF
--- a/SignalMessaging/environment/Environment.h
+++ b/SignalMessaging/environment/Environment.h
@@ -34,8 +34,9 @@
 @property (nonatomic, readonly) OWSMessageSender *messageSender;
 @property (nonatomic, readonly) OWSPreferences *preferences;
 
-+ (Environment *)current;
-+ (void)setCurrent:(Environment *)environment;
+@property (class, readonly, nonatomic) Environment *shared;
+
++ (void)setShared:(Environment *)environment;
 // Should only be called by tests.
 + (void)clearCurrentForTests;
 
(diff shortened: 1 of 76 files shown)


SignalMessaging/environment/Environment.h
```diff
@@ -34,8 +34,9 @@
 @property (nonatomic, readonly) OWSMessageSender *messageSender;
 @property (nonatomic, readonly) OWSPreferences *preferences;
 
-+ (Environment *)current;
-+ (void)setCurrent:(Environment *)environment;
+@property (class, readonly, nonatomic) Environment *shared;
+
++ (void)setShared:(Environment *)environment;
```
michaelkirk-signal: It seems like we'd get the same behavior by removing the `readonly` up a couple lines.

I'd argue doing so is less surprising to the reader since it's really **not** `readonly` - we've manually declared the exact same setter that would be generated by omitting `readonly`.

SignalServiceKit/src/SSKEnvironment.h
```diff
@@ -20,8 +22,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype)init NS_UNAVAILABLE;
 
-+ (instancetype)sharedEnv;
-+ (void)setSharedEnv:(TextSecureKitEnv *)env;
+@property (class, readonly, nonatomic) SSKEnvironment *shared;
```
michaelkirk-signal: same comment re: not actually `readonly`.

PR #3857, opened 2018-08-31
Fix build breakage in our tests.

charlesmchen-signal:
PTAL @michaelkirk


michaelkirk-signal:
Is this actually required for anything?

PR #3870, opened 2018-09-07
Charlesmchen/fix ssk tests k

charlesmchen-signal:
This PR does three things:

* Add base class for tests.  They don't do much yet, but they will soon when the Environments are mock-able.
* Move fakes to SSK/tests.  This way we can use them in Signal and SSK tests.
* Get some (not all) broken tests compiling (not passing).  Lots of little fixes.

These changes are in three separate commits.

PTAL @michaelkirk


charlesmchen-signal:
> These changes are in three separate commits.

If this is hard to review, I'll chop this into 3 PRs.

PR #3871, opened 2018-09-07
Add base class for tests.

charlesmchen-signal:
PTAL @michaelkirk 

These base classes don't do much yet - but they will soon when the environments are mockable.

### DIFF
--- a/SignalServiceKit/tests/Contacts/TSContactThreadTest.m
+++ b/SignalServiceKit/tests/Contacts/TSContactThreadTest.m
@@ -19,6 +19,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)setUp
 {
+    [super setUp];
+
     self.contactThread = [TSContactThread getOrCreateThreadWithContactId:@"fake-contact-id"];
     [OWSRecipientIdentity removeAllObjectsInCollection];
 }
--- a/SignalServiceKit/tests/Contacts/TSContactThreadTest.m
+++ b/SignalServiceKit/tests/Contacts/TSContactThreadTest.m
@@ -1,15 +1,15 @@
 //
-//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
 #import "TSContactThread.h"
+#import "MockSSKEnvironment.h"
 #import "OWSIdentityManager.h"
-#import "OWSUnitTestEnvironment.h"
-#import <XCTest/XCTest.h>
+#import "SSKBaseTest.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface TSContactThreadTest : XCTestCase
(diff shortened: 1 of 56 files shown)


SignalServiceKit/tests/Util/SSKSwiftTests.swift
```diff
@@ -0,0 +1,8 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import XCTest
+
+class SSKSwiftTests: XCTestCase {
```
michaelkirk-signal: Why is there a separate swift base class? And don't we want it to do the same things as the ObjC base class re: environment setup/teardown?

SignalServiceKit/tests/Contacts/TSContactThreadTest.m
```diff
@@ -19,7 +19,6 @@ @implementation TSContactThreadTest
 
 - (void)setUp
 {
```
michaelkirk-signal: This doesn't call super. Can you check the others to make sure they all do?

Maybe we can take advantage of `__attribute__((objc_requires_super));`

Also, this method, and presumably others are doing things to ensure no existing state lingers. That should be unnecessary now, right?

In this example, we could get rid of `[OWSRecipientIdentity removeAllObjectsInCollection]`?

SignalServiceKit/tests/Util/SSKSwiftTests.swift
charlesmchen-signal: > Why is there a separate swift base class?

Sorry, I should have left a comment.  This was just to workaround to get things building.  It's been 2+ weeks since I did this and I can't recall the details, but it's something like this: if you don't have any .swift tests, no Swift-to-ObjC bridging doesn't work.  Once I get things building again, I'll improve the comment.

SignalServiceKit/tests/Contacts/TSContactThreadTest.m
charlesmchen-signal: > This doesn't call super. Can you check the others to make sure they all do?

Done.

> Maybe we can take advantage of __attribute__((objc_requires_super));

I've added this to the card.

> Also, this method, and presumably others are doing things to ensure no existing state lingers. That should be unnecessary now, right?

Yup, that's right.  Since we're re-doing this work in small steps, I'm trying to first work towards a) getting everything building b) mechanical changes.

> In this example, we could get rid of [OWSRecipientIdentity removeAllObjectsInCollection]?

Right, there's a lot of that kind of "teardown" that will no longer be necessary and which I'll eventually remove.

charlesmchen-signal:
PTAL @michaelkirk

charlesmchen-signal:
PTAL @michaelkirk

charlesmchen-signal:
PTAL @michaelkirk

michaelkirk-signal:
LGTM

PR #3877, opened 2018-09-10
Treat failure to delete a non-existent password as success

mattrubin:

### Contributor checklist

 * iPhone X (simulator), iOS 11.4 (15F79)
 * iPhone X, iOS 11.4.1 (15G77)

- - - - - - - - - -

### Description

If `SSKDefaultKeychainStorage.remove(service:key:)` fails to remove a keychain item because the specified item does not exist, treat the operation as having succeeded instead of throwing an error.

Fixes https://github.com/signalapp/Signal-iOS/issues/3876. With these changes, the app no longer crashes on launch when installed on a clean device and run for the first time.

### DIFF
--- a/SignalServiceKit/src/Storage/KeychainStorage.swift
+++ b/SignalServiceKit/src/Storage/KeychainStorage.swift
@@ -94,6 +94,10 @@ public class SSKDefaultKeychainStorage: NSObject, SSKKeychainStorage {
         var error: NSError?
         let result = SAMKeychain.deletePassword(forService: service, account: key, error: &error)
         if let error = error {
+            // If deletion failed because the specified item could not be found in the keychain, consider it success.
+            if error.code == errSecItemNotFound {
+                return
+            }
             throw KeychainStorageError.failure(description: "\(logTag) error removing data: \(error)")
         }
         guard result else {


SignalServiceKit/src/Storage/KeychainStorage.swift
```diff
@@ -94,6 +94,10 @@ public class SSKDefaultKeychainStorage: NSObject, SSKKeychainStorage {
         var error: NSError?
         let result = SAMKeychain.deletePassword(forService: service, account: key, error: &error)
         if let error = error {
+            // If deletion failed because the specified item could not be found in the keychain, consider it success.
+            if error.code == errSecItemNotFound {
```
michaelkirk-signal: A guard statement would better capture this pattern.

mattrubin: Thanks for the feedback. Are you suggesting replacing the new `if` statement with a `guard` statement like this?
```swift
guard error.code != errSecItemNotFound else {
    return
}
```

-----

There's also the possibility of condensing this whole section to
```swift
if let error = error {
    guard error.code == errSecItemNotFound else {
        throw KeychainStorageError.failure(description: "\(logTag) error removing data: \(error)")
    }
}
```
The problem with the second approach would be the fall-through to the `guard result` statement below, which would require further refactoring to avoid mistakenly throwing that error instead.

michaelkirk-signal:
PTAL @charlesmchen

charlesmchen-signal:
Merged; thanks @mattrubin

PR #3879, opened 2018-09-12
fix group avatar updates and quote generation

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
+++ b/SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
@@ -695,9 +695,9 @@ typedef void (^OWSLoadedThumbnailSuccess)(OWSLoadedThumbnail *loadedThumbnail);
 {
     OWSLoadedThumbnail *_Nullable loadedThumbnail;
     loadedThumbnail = [self loadedThumbnailWithThumbnailDimensionPoints:thumbnailDimensionPoints
-        success:^(OWSLoadedThumbnail *loadedThumbnail) {
+        success:^(OWSLoadedThumbnail *thumbnail) {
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -749,9 +749,9 @@ - (nullable OWSLoadedThumbnail *)loadedThumbnailSmallSync
     __block OWSLoadedThumbnail *_Nullable asyncLoadedThumbnail = nil;
     OWSLoadedThumbnail *_Nullable syncLoadedThumbnail = nil;
     syncLoadedThumbnail = [self loadedThumbnailWithThumbnailDimensionPoints:kThumbnailDimensionPointsSmall
-        success:^(OWSLoadedThumbnail *asyncLoadedThumbnail) {
```
michaelkirk-signal: This is the fix - we were assigning to the local param, which shadowed the name of the outer variable.

SignalServiceKit/src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -695,9 +695,9 @@ - (nullable UIImage *)thumbnailImageWithThumbnailDimensionPoints:(NSUInteger)thu
 {
     OWSLoadedThumbnail *_Nullable loadedThumbnail;
     loadedThumbnail = [self loadedThumbnailWithThumbnailDimensionPoints:thumbnailDimensionPoints
-        success:^(OWSLoadedThumbnail *loadedThumbnail) {
```
michaelkirk-signal: This change is cosmetic, just wanted to avoid shadowing the external variable to avoid similar problems.

PR #3884, opened 2018-09-15
Mkirk/fix test run

michaelkirk-signal:
Get the iOS test suite compiling.

PTAL @charlesmchen


Signal/test/util/OWSOrphanDataCleanerTest.m
```diff
@@ -43,7 +43,9 @@ - (void)tearDown
 
 - (NSUInteger)numberOfItemsInAttachmentsFolder
 {
-    return [OWSOrphanDataCleaner filePathsInAttachmentsFolder].count;
```
michaelkirk-signal: The underlying class has changed so much, that updating this test is a large project, but in the meanwhile it was preventing other tests from running / being written.

I arrived at this approach because I didn't want to lose the existing implementation or forget that it was "commented out".

PR #3888, opened 2018-09-19
New grayscale palette.

charlesmchen-signal:
PTAL @michaelkirk 

Note that hairlines are now more subtle in "light theme".

### DIFF
--- a/SignalMessaging/categories/Theme.m
+++ b/SignalMessaging/categories/Theme.m
@@ -70,13 +70,11 @@ NSString *const ThemeKeyThemeEnabled = @"ThemeKeyThemeEnabled";
 
 + (UIColor *)boldColor
 {
-    // TODO: Review with design.
     return (Theme.isDarkThemeEnabled ? UIColor.ows_whiteColor : UIColor.blackColor);
 }
 
 + (UIColor *)middleGrayColor
 {
-    // TODO: Review with design.
     return [UIColor colorWithWhite:0.5f alpha:1.f];
 }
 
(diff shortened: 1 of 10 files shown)


SignalMessaging/categories/Theme.m
```diff
@@ -49,7 +49,7 @@ + (void)setIsDarkThemeEnabled:(BOOL)value
 
 + (UIColor *)backgroundColor
 {
-    return (Theme.isDarkThemeEnabled ? UIColor.ows_blackColor : UIColor.ows_whiteColor);
+    return (Theme.isDarkThemeEnabled ? UIColor.ows_gray95Color : UIColor.ows_whiteColor);
```
michaelkirk-signal: Is this right?

I see "Dark: **Mobile** Background" should be "#000000"
Whereas gray95 is for "Dark: **Desktop** Background"

SignalMessaging/categories/UIColor+OWS.m
```diff
@@ -142,72 +142,54 @@ + (UIColor *)ows_redColor
...
-    return [UIColor colorWithRGBHex:0x8B8E91];
-}
-
-+ (UIColor *)ows_light60Color
-{
-    return [UIColor colorWithRGBHex:0x62656A];
-}
-
-+ (UIColor *)ows_light90Color
++ (UIColor *)ows_gray15Color
```
michaelkirk-signal: I don't see this one in the spec

SignalMessaging/categories/Theme.m
michaelkirk-signal: This was my only real concern - feel free to merge after addressing.

charlesmchen-signal: This is pending a final decision from Myles - he's still mulling - but this is supposedly the new design.

SignalMessaging/categories/UIColor+OWS.m
charlesmchen-signal: Ah yes, we just eliminated that one after Myles reviewed this work.

michaelkirk:
Ok, as long as it’s Myles approved, LGTM

charlesmchen-signal:
Myles has committed to changing the background color.

Also, he's refined gray 95: https://github.com/signalapp/Signal-iOS/pull/3888/commits/6a712366aecbddf74da6a99778fa1d95803852be

PR #3892, opened 2018-09-20
Fix proto stream test.

charlesmchen-signal:
Fixes the only non-passing test in `Signal` target.

PTAL @michaelkirk


Signal/test/util/ProtoParsingTest.m
```diff
@@ -128,8 +128,9 @@ - (void)testProtoStreams
     XCTAssertNotNil(streamData);
 
     XCTAssertEqualObjects(streamData.hexadecimalString,
```
michaelkirk-signal: This test is a bit weird. It's good for detecting regressions, but how do we know these new bytes are the right bytes?  👀 

I guess we just verify that contact syncing of names, profile key, identity key, etc. are working after updating the test?

Additionally we could do some kind of round tripping, e.g. re-parsing the output data.

charlesmchen-signal: Agreed.  

Once we get the tests all passing, I'd like to have a separate conversation about in which sections of the codebase we'd like to improve our test coverage.  IMO it isn't urgent to improve the tests in this area.

PR #3895, opened 2018-09-20
Move more singletons to SSKEnvironment.

charlesmchen-signal:
Pretty straightforward.  This enables the next PR which gets the SSK tests building and running (but not passing).

PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Contacts/Threads/TSContactThread.h
+++ b/SignalServiceKit/src/Contacts/Threads/TSContactThread.h
@@ -22,6 +22,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 + (NSString *)contactIdFromThreadId:(NSString *)threadId;
 
+// This is only exposed for tests.
+#ifdef DEBUG
++ (NSString *)threadIdFromContactId:(NSString *)contactId;
+#endif
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/SignalServiceKit/src/Messages/OWSBlockingManager.m
+++ b/SignalServiceKit/src/Messages/OWSBlockingManager.m
@@ -31,7 +31,6 @@ NSString *const kOWSBlockingManager_SyncedBlockedGroupIdsKey = @"kOWSBlockingMan
 @interface OWSBlockingManager ()
 
 @property (nonatomic, readonly) YapDatabaseConnection *dbConnection;
-@property (nonatomic, readonly) OWSMessageSender *messageSender;
 
 // We don't store the phone numbers as instances of PhoneNumber to avoid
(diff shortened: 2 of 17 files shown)


SignalServiceKit/src/Contacts/Threads/TSContactThread.h
```diff
@@ -22,6 +22,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 + (NSString *)contactIdFromThreadId:(NSString *)threadId;
 
+// This is only exposed for tests.
```
charlesmchen-signal: There's two ways we tend to expose private functionality or state to tests:

* Through categories defined in the tests.
* Through DEBUG-only methods/properties on the class itself.

Previously I preferred the former since it kept the production code clean and neat.  Lately I prefer the latter because we're more likely to detect breakage (through build warnings or errors) even if don't build/run the tests.  

WDYT?

michaelkirk-signal: I'm not sure I'm following... can we discuss offline?

SignalServiceKit/src/Messages/OWSBlockingManager.m
```diff
@@ -73,17 +60,11 @@ - (instancetype)initWithPrimaryStorage:(OWSPrimaryStorage *)primaryStorage
...
-    OWSAssertDebug(messageSender);
 
     _dbConnection = primaryStorage.newDatabaseConnection;
-    _messageSender = messageSender;
 
     OWSSingletonAssert();
 
-    // Register this manager with the message sender.
-    // This is a circular dependency.
-    [messageSender setBlockingManager:self];
```
michaelkirk-signal: Happy to see this go!

SignalServiceKit/src/Messages/OWSBlockingManager.m
```diff
@@ -100,8 +81,14 @@ - (void)observeNotifications
                                                object:nil];
 }
 
-#pragma mark -
+- (OWSMessageSender *)messageSender
```
michaelkirk-signal: By convention I've been putting these into a `#pragma mark - Dependencies` section, but no biggie.

SignalServiceKit/src/Contacts/Threads/TSContactThread.h
charlesmchen-signal: Yup, happy to.

SignalServiceKit/src/Messages/OWSBlockingManager.m
charlesmchen-signal: Okay

PR #3897, opened 2018-09-21
Add contact utils class.

charlesmchen-signal:
PTAL @michaelkirk  

Pulled out some debug functionality so that it can be accessed programmatically.


Signal/src/ViewControllers/DebugUI/ContactsUtils.h
```diff
@@ -0,0 +1,25 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class CNContact;
+
+@interface ContactsUtils : NSObject
```
michaelkirk-signal: Maybe DebugContactUtils? TIOLI, but we usually include the term "Debug" in these kinds of classes.

charlesmchen-signal: Done.

PR #3898, opened 2018-09-21
Charlesmchen/dark theme reminders

charlesmchen-signal:
PTAL @michaelkirk 

Fixed the dark theme for "nag" reminders.  While doing it I found and fixed two other bugs.

simulator screen shot - iphone x - 2018-09-21 at 11 34 08

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -157,7 +157,10 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
 #pragma GCC diagnostic ignored "-Wunused-result"
     [ExperienceUpgradeFinder sharedManager];
 #pragma GCC diagnostic pop
+}
 
+- (void)observeNotifications
+{
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(signalAccountsDidChange:)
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -157,7 +157,10 @@ - (void)commonInit
 #pragma GCC diagnostic ignored "-Wunused-result"
     [ExperienceUpgradeFinder sharedManager];
 #pragma GCC diagnostic pop
+}
 
+- (void)observeNotifications
```
charlesmchen-signal: We don't want to observe notifications until the view has loaded, since we want to "reload the table" in response to some notifications and that's not safe to do until the mappings are created.

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -349,6 +350,16 @@ - (void)updateReminderViews
         || !self.deregisteredView.isHidden || !self.outageView.isHidden;
 }
 
+- (void)setHasVisibleReminders:(BOOL)hasVisibleReminders
+{
+    if (_hasVisibleReminders == hasVisibleReminders) {
+        return;
+    }
+    _hasVisibleReminders = hasVisibleReminders;
+    // If the reminders show/hide, reload the table.
```
charlesmchen-signal: If the reminders show/hide, reload the table since a cell needs to be added/removed.

PR #3902, opened 2018-09-24
Conversation Colors vs. Avatars, Part 1

charlesmchen-signal:
In order to unblock work/avoid divergence, I'm opening this PR even though there's more changes to be made related to avatars.

### What's included:

* Update avatar colors
* Refactor avatar creation to DRY up the contact avatars, group avatars and emoticon avatars.
* Add shaded conversation color constants These are stubbed out and don't have the right names/values yet.
* Modify color picker to be color-name-based, not color-based.
* Use shaded conversation colors
* Remove JSQ.

### What's not included:

* Update the actual conversation colors constants (names and hex values).
* Update group avatars (waiting on icon asset from Myles).


Signal/src/ViewControllers/ColorPickerViewController.swift
```diff
@@ -46,15 +46,15 @@ class ColorPickerViewController: UIViewController, UIPickerViewDelegate, UIPicke
 
     private let pickerView: UIPickerView
     private let thread: TSThread
-    private let colors: [UIColor]
+    private let colorNames: [String]
```
charlesmchen-signal: This picker should work in light and dark theme.  Makes more sense for it to be "color name"-based than "color" based.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
```diff
@@ -269,11 +269,10 @@ - (BOOL)updateAvatarView
     }
 
     TSIncomingMessage *incomingMessage = (TSIncomingMessage *)self.viewItem.interaction;
-    OWSAvatarBuilder *avatarBuilder = [[OWSContactAvatarBuilder alloc] initWithSignalId:incomingMessage.authorId
-                                                                                  color:self.conversationStyle.primaryColor
-                                                                               diameter:self.avatarSize
-                                                                        contactsManager:contactsManager];
-    self.avatarView.image = [avatarBuilder build];
+    TSThread *authorThread = [TSContactThread getOrCreateThreadWithContactId:incomingMessage.authorId];
```
charlesmchen-signal: We could look up the thread for this author in the `ConversationViewItem`, so that we a) use the right db connection b) avoid sneaky read transaction here.

SignalMessaging/categories/UIColor+OWS.h
```diff
@@ -31,12 +31,10 @@ NS_ASSUME_NONNULL_BEGIN
 #pragma mark - ConversationColor
 
 + (nullable UIColor *)ows_conversationColorForColorName:(NSString *)colorName
-    NS_SWIFT_NAME(ows_conversationColor(colorName:));
-+ (nullable NSString *)ows_conversationColorNameForColor:(UIColor *)color
-    NS_SWIFT_NAME(ows_conversationColorName(color:));
+                                               isShaded:(BOOL)isShaded
+    NS_SWIFT_NAME(ows_conversationColor(colorName:isShaded:));
```
charlesmchen-signal: We could have the impl check for dark theme, but I thought it'd be good to force the call site to specify whether or not the color should be shaded.

SignalMessaging/categories/UIColor+OWS.m
```diff
@@ -273,26 +273,46 @@ + (UIColor *)ows_darkSkyBlueColor
...
 
-    return [self.ows_conversationColorMap objectForKey:colorName];
+    return (isShaded ? self.ows_conversationColorMapDark : self.ows_conversationColorMapLight)[colorName];
 }
 
-+ (nullable NSString *)ows_conversationColorNameForColor:(UIColor *)color
++ (NSString *)ows_defaultConversationColorName
 {
-    return [self.ows_conversationColorMap allKeysForObject:color].firstObject;
+    return @"teal";
```
charlesmchen-signal: I need to ask Myles about this.

SignalMessaging/utils/OWSAvatarBuilder.h
```diff
@@ -1,24 +1,33 @@
...
 //
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class TSThread;
 @class OWSContactsManager;
+@class TSThread;
 @class UIImage;
 
 @interface OWSAvatarBuilder : NSObject
```
charlesmchen-signal: The goal here was to DRY up the avatar creation logic so that we had consistent colors, gradients, text sizing, etc.

SignalMessaging/utils/OWSAvatarBuilder.m
```diff
@@ -50,34 +51,194 @@ + (UIImage *)buildRandomAvatarWithDiameter:(NSUInteger)diameter
...
+        NSFontAttributeName : font,
+        NSForegroundColorAttributeName : textColor,
+    };
+    CGSize textSize =
+        [initials boundingRectWithSize:frame.size
+                               options:(NSStringDrawingUsesLineFragmentOrigin | NSStringDrawingUsesFontLeading)
+                            attributes:textAttributes
+                               context:nil]
+            .size;
+    // Ensure that the text fits within the avatar bounds, with a margin.
```
charlesmchen-signal: This isn't in the spec, but IMO we should ensure that unexpected glyphs (e.g. in other languages) always fit neatly within the avatar bounds.

SignalMessaging/utils/OWSContactAvatarBuilder.m
```diff
@@ -87,10 +90,15 @@ - (nullable UIImage *)buildSavedImage
...
+- (id)cacheKey
+{
+    return [NSString stringWithFormat:@"%@-%d", self.signalId, Theme.isDarkThemeEnabled];
+}
+
+- (nullable UIImage *)buildDefaultImage
 {
     UIImage *cachedAvatar =
-        [self.contactsManager.avatarCache imageForKey:self.signalId diameter:(CGFloat)self.diameter];
+        [self.contactsManager.avatarCache imageForKey:self.cacheKey diameter:(CGFloat)self.diameter];
```
charlesmchen-signal: Note that we now use a cache key so that light and dark avatars don't collide.

charlesmchen-signal:
Oh, this also doesn't include switching to using an enum for conversation colors instead of a string.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m
michaelkirk-signal: This is a sneaky read/__write__ transaction - seems like something we should avoid in UI code.

SignalMessaging/categories/UIColor+OWS.h
michaelkirk-signal: Yeah - I prefer this rather than have UIColor depend on `Theme` state implicitly.

SignalMessaging/categories/UIColor+OWS.m
```diff
@@ -273,26 +273,46 @@ + (UIColor *)ows_darkSkyBlueColor
     return colorMap;
 }
 
-+ (NSArray<NSString *> *)ows_conversationColorNames
++ (NSDictionary<NSString *, UIColor *> *)ows_conversationColorMapDark
```
michaelkirk-signal: maybe `colorMapShaded` so as not to be confused with "dark" theme?

SignalMessaging/utils/OWSAvatarBuilder.m
```diff
@@ -50,34 +51,194 @@ + (UIImage *)buildRandomAvatarWithDiameter:(NSUInteger)diameter
     NSString *randomEyebrow = eyebrows[arc4random_uniform((uint32_t)eyebrows.count)];
     NSString *face = [NSString stringWithFormat:@"%@%@%@", randomEyebrow, randomEye, randomMouth];
 
-    CGFloat fontSize = (CGFloat)(diameter / 2.4);
+    UIColor *backgroundColor = [UIColor colorWithRGBHex:0xaca6633];
+
+    return [self avatarImageWithDiameter:diameter
+                         backgroundColor:backgroundColor
+                               drawBlock:^(CGContextRef context) {
+                                   CGContextTranslateCTM(context, diameter / 2, diameter / 2);
```
michaelkirk-signal: heh, I don't know if it's worth abstracting this to yield the drawing context if it's only used by this debug code.

Admittedly the old debug code was ugly, but I prefer ugly debug code vs. uglifying the production code to make the debug code look nicer.

SignalMessaging/utils/OWSAvatarBuilder.m
```diff
@@ -50,34 +51,194 @@ + (UIImage *)buildRandomAvatarWithDiameter:(NSUInteger)diameter
...
+        NSForegroundColorAttributeName : textColor,
+    };
+    CGSize textSize =
+        [initials boundingRectWithSize:frame.size
+                               options:(NSStringDrawingUsesLineFragmentOrigin | NSStringDrawingUsesFontLeading)
+                            attributes:textAttributes
+                               context:nil]
+            .size;
+    // Ensure that the text fits within the avatar bounds, with a margin.
+    if (textSize.width > 0 && textSize.height > 0) {
```
michaelkirk-signal: else OWSDebugFail?

SignalMessaging/utils/OWSContactAvatarBuilder.m
```diff
@@ -121,19 +129,19 @@ - (UIImage *)buildDefaultImage
...
-                                                                            font:[UIFont ows_boldFontWithSize:fontSize]
-                                                                        diameter:self.diameter] avatarImage];
+    UIColor *color = [UIColor ows_conversationColorForColorName:self.colorName isShaded:Theme.isDarkThemeEnabled];
+    OWSAssertDebug(color);
 
-    [self.contactsManager.avatarCache setImage:image forKey:self.signalId diameter:self.diameter];
+    UIImage *_Nullable image =
+        [OWSAvatarBuilder avatarImageWithInitials:initials backgroundColor:color diameter:self.diameter];
+    if (!image) {
+        return nil;
```
michaelkirk-signal: Do we ever expect this to actually be nil? or is this just in line with the CGImage drawing API?

SignalMessaging/categories/UIColor+OWS.m
```diff
@@ -273,26 +273,46 @@ + (UIColor *)ows_darkSkyBlueColor
...
 {
-    return self.ows_conversationColorMap.allKeys;
+    static NSDictionary<NSString *, UIColor *> *colorMap;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        colorMap = @{
+            @"red" : self.ows_red700Color,
```
michaelkirk-signal: Shouldn't there be 12 colors?

I also suspect that instead of passing around string arguments, we'd benefit from passing around strongly typed ConversationColor enum members, and have the ability to map it to/from strings for the purposes of serialization.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageCell.m

SignalMessaging/utils/OWSAvatarBuilder.m
charlesmchen-signal: This isn't just to DRY up the debug code, I'm also working towards DRYing up with the group avatars. We're missing an asset but when that's delivered we'll also use this there.

SignalMessaging/utils/OWSAvatarBuilder.m
charlesmchen-signal: Done.

SignalMessaging/utils/OWSContactAvatarBuilder.m
charlesmchen-signal: It shouldn't ever be nil.

SignalMessaging/categories/UIColor+OWS.m
charlesmchen-signal: This is the PR that actually updates these constants: https://github.com/signalapp/Signal-iOS/pull/3906

Yes, we could definitely use an enum.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -94,6 +93,9 @@ - (instancetype)initWithInteraction:(TSInteraction *)interaction
     _interaction = interaction;
     _isGroupThread = isGroupThread;
     _conversationStyle = conversationStyle;
+    _incomingMessageAuthorThread = ([interaction isKindOfClass:[TSIncomingMessage class]]
+            ? [TSContactThread getOrCreateThreadWithContactId:((TSIncomingMessage *)interaction).authorId]
```
michaelkirk-signal: Now we're opening a sneaky read/write transaction. I equally don't think we should do this here or anywhere else on the main thread, especially for hot code like ConversationViewItem initialization.

Why do we now need a thread when we didn't before? I think it was part of collapsing the avatar factory methods behind a single unified method. Is that part of the change necessary?

michaelkirk-signal: Keep in mind, we're already within the context of a read transaction here. 

Don't we already have access to a thread when ConversationViewItems are built? Could we just tag it on in the initializer?

charlesmchen-signal: Oops, I was using the wrong method. Fixed.

>  I think it was part of collapsing the avatar factory methods behind a single unified method.

Nope, this is necessary for the new design.  If the conversation color for your 1:1 conversation with Alice is "red", Alice's default avatar should be red, even in group conversations.  This isn't something we were doing before, hence the need for the thread.

charlesmchen-signal: Hmm, actually, this isn't good enough. I'm going to revisit tomorrow.

michaelkirk-signal: > Nope, this is necessary for the new design. If the conversation color for your 1:1 conversation with Alice is "red", Alice's default avatar should be red, even in group conversations. This isn't something we were doing before, hence the need for the thread.

Oh right, that makes sense. 

Passing the thread in as an initializer param (e.g. instead of isGroupThread) seems reasonable.

charlesmchen-signal: Okay, I just pushed a commit that addresses this issue.

SignalMessaging/utils/OWSContactAvatarBuilder.m
michaelkirk-signal: OWSFailDebug then? If nothing else it's good documentation.

charlesmchen-signal: Done.

PR #3905, opened 2018-09-24
Use local avatar as app settings button.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -197,6 +197,10 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
                                              selector:@selector(themeDidChange:)
                                                  name:ThemeDidChangeNotification
                                                object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(localProfileDidChange:)
+                                                 name:kNSNotificationName_LocalProfileDidChange
+                                               object:nil];
 }
 
 - (void)dealloc
@@ -225,6 +229,13 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
     OWSAssertIsOnMainThread();
 
     [self updateReminderViews];
+}
+
+- (void)localProfileDidChange:(id)notification
+{
+    OWSAssertIsOnMainThread();
+
+    [self updateBarButtonItems];
 }
 
 #pragma mark - Theme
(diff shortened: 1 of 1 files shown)


Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -472,10 +483,24 @@ - (void)updateBarButtonItems
...
+    AvatarImageView *avatarView = [[AvatarImageView alloc] initWithImage:avatarImage];
+    if (!localProfileAvatarImage) {
+        avatarView.tintColor = Theme.middleGrayColor;
+    }
+    [avatarView autoSetDimension:ALDimensionWidth toSize:kAvatarSize];
+    [avatarView autoSetDimension:ALDimensionHeight toSize:kAvatarSize];
+    UIBarButtonItem *settingsButton = [[UIBarButtonItem alloc] initWithCustomView:avatarView];
+    avatarView.userInteractionEnabled = YES;
+    [avatarView addGestureRecognizer:[[UITapGestureRecognizer alloc] initWithTarget:self
+                                                                             action:@selector(settingsButtonPressed:)]];
```
charlesmchen-signal: We need to verify that this doesn't make it any harder to hit the app settings button.

One alternative would be to pre-render the avatar view to a `UIImage`.

michaelkirk-signal:
Hmm... does this need a rebase? I see the colorNames changes in here, but it looks like you're targeting that PR already.

charlesmchen-signal:
> Hmm... does this need a rebase? I see the colorNames changes in here, but it looks like you're targeting that PR already.

Rebased. PTAL @michaelkirk

Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -485,10 +496,24 @@ - (void)updateBarButtonItems
     }
 
     //  Settings button.
-    //
-    // TODO: Theme
-    UIImage *image = [UIImage imageNamed:@"button_settings_white"];
-    UIBarButtonItem *settingsButton = [[UIBarButtonItem alloc] initWithImage:image style:UIBarButtonItemStylePlain target:self action:@selector(settingsButtonPressed:)];
+    const NSUInteger kAvatarSize = 28;
+    // TODO: Replace this icon.
```
michaelkirk-signal: re: this todo, do you mean instead of using the profile_avatar_default needs to change? Is that already on Myles's radar?

Signal/src/ViewControllers/HomeView/HomeViewController.m
michaelkirk-signal: We also lose the "touch down" effect this way. You could try doing something like `DisappearingTimerConfigurationView` which simulates this state.

Also, AvatarImageView is a pretty tiny class - did you explore re-implementing it's features in a UIButton?

Signal/src/ViewControllers/HomeView/HomeViewController.m
charlesmchen-signal: Myles has clarified the design.  I'll address in a follow-up PR.

Signal/src/ViewControllers/HomeView/HomeViewController.m
michaelkirk-signal: > DisappearingTimerConfigurationView uses long press which has multiple states. Tap gesture recognizers don't provide anything except the "tapped" state.

Right. I was suggesting we do more than a simple tap gesture, which is insufficient for providing the UX feedback that UIButtons offer.

PR #3911, opened 2018-09-26
Charlesmchen/converation colors vs bubbles

charlesmchen-signal:
PTAL @michaelkirk


Signal/src/ViewControllers/ColorPickerViewController.swift
```diff
@@ -82,8 +82,8 @@ class ColorPickerViewController: UIViewController, UIPickerViewDelegate, UIPicke
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        if let colorName = thread.conversationColorName,
-            let index = colorNames.index(of: colorName) {
+        let colorName = thread.conversationColorName
```
charlesmchen-signal: `thread.conversationColorName` is no longer nullable.

SignalMessaging/categories/Theme.h
```diff
@@ -56,4 +57,12 @@ extern NSString *const ThemeDidChangeNotification;
 
 @end
 
+#pragma mark -
+
+@interface OWSConversationColors (Theme)
+
+@property (nonatomic, readonly) UIColor *themeColor;
```
charlesmchen-signal: I think `OWSConversationColors` will make for more readable code.

This category lets us pull out the theme-specific functionality.

SignalMessaging/categories/UIColor+OWS.h
```diff
@@ -59,19 +64,66 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+// If the conversation color name is valid, return its colors.
+// Otherwise return the "default" conversation colors.
++ (OWSConversationColors *)ows_conversationColorsOrDefaultForColorName:(NSString *)conversationColorName
+    NS_SWIFT_NAME(ows_conversationColorsOrDefault(colorName:));
+
+@property (class, readonly, nonatomic) NSArray<NSString *> *ows_conversationColorNames;
 
 + (NSString *)ows_defaultConversationColorName;
++ (OWSConversationColors *)ows_defaultConversationColors;
```
charlesmchen-signal: The "default" conversation colors should only be relevant if a conversation has an invalid (obsolete?) conversation color.

Signal/src/ViewControllers/ConversationView/Cells/OWSQuotedMessageView.m
```diff
@@ -122,7 +122,7 @@ - (BOOL)hasQuotedAttachmentThumbnailImage
 - (UIColor *)highlightColor
 {
     BOOL isQuotingSelf = [NSObject isNullableObject:self.quotedMessage.authorId equalTo:TSAccountManager.localNumber];
-    return (isQuotingSelf ? self.conversationStyle.bubbleColorOutgoingSent
+    return (isQuotingSelf ? [self.conversationStyle bubbleColorWithIsIncoming:NO]
```
michaelkirk-signal: It seems more straight forward / readable to have both `incomingBubbleColor` and `outgoingBubbleColor` methods than a "mode flag" on this single method.

SignalMessaging/categories/UIColor+OWS.h
michaelkirk-signal: > obsolete?

It was my understanding that we'd be migrating everyone to a modern color, so this really is just a fail-safe, right?

SignalMessaging/categories/Theme.h
michaelkirk-signal: I like that you've pulled these into an object - though I'm not crazy about the name. I try to avoid having a single object expressed as a plural because inevitably, what do you call more than one of these? It's just a little readability/predictability hurdle. Would OWSConversationPalette be acceptable?

Also - what was the reasoning behind having a separate category? 

I pretty much always put things into the primary interface declaration unless:

1. I'm adding methods to classes I don't control (NSString+OWS) 
2. Or if I *privately* want to access some extra functionality on a class.
3. If I want to add Swift features to an Objc class without re-writing the entire class, I'll  create a swift extension of the class.

SignalMessaging/categories/UIColor+OWS.m
```diff
@@ -2,12 +2,41 @@
...
+#import "Theme.h"
+#import "UIColor+OWS.h"
 #import <SignalServiceKit/Cryptography.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
+
+@interface OWSConversationColors ()
+
+@property (nonatomic) UIColor *defaultColor;
```
michaelkirk-signal: Maybe `primaryColor` or `mainColor` would be better than `defaultColor` so as not to conflict with the other usage of `default` as in `ows_defaultConversationColors`

SignalMessaging/categories/UIColor+OWS.m
```diff
@@ -196,123 +225,311 @@ + (UIColor *)ows_blackColor
...
+}
+
++ (OWSConversationColors *)ows_conversationColorsOrDefaultForColorName:(NSString *)conversationColorName
+{
+    OWSConversationColors *_Nullable conversationColors =
+        [self ows_conversationColorsForColorName:conversationColorName];
+    if (conversationColors) {
+        return conversationColors;
+    }
+    return [self ows_defaultConversationColors];
```
michaelkirk-signal: Like Android, I think we want to map legacy colors to their new color value here, otherwise won't we have a whole lot of defaultConversationColor'd avatars?

Signal/src/ViewControllers/ThreadSettings/OWSConversationSettingsViewController.m
```diff
@@ -289,8 +289,7 @@ - (void)updateTableContents
                              itemWithCustomCellBlock:^{
                                  NSString *colorName = self.thread.conversationColorName;
                                  UIColor *currentColor =
-                                     [UIColor ows_conversationColorForColorName:colorName
-                                                                       isShaded:Theme.isDarkThemeEnabled];
+                                     [UIColor ows_conversationColorsForColorName:colorName].themeColor;
```
michaelkirk-signal: This is a nullable method - should this be ows_conversationColorsOrDefaultForColorName?

SignalMessaging/categories/Theme.h
charlesmchen-signal: > Would OWSConversationPalette be acceptable?

I think I'm going to go with `OWSConversationColor` (singular) since "conversation colors" is the term we're using within the team, in the UI, etc.

> Also - what was the reasoning behind having a separate category?

I was trying to avoid having `UIColor+OWS` depend on `Theme`, as you asked.  

https://github.com/signalapp/Signal-iOS/pull/3902#discussion_r220323636

I'll merge this category into the class.

SignalMessaging/categories/UIColor+OWS.h
charlesmchen-signal: Right.

SignalMessaging/categories/UIColor+OWS.m
charlesmchen-signal: This will be in a forthcoming PR.

SignalMessaging/categories/Theme.h
michaelkirk-signal: Since theme depends on UIColor, I do think it would be good for UIColor not to depend on Theme. 

In my experience, having strictly hierarchical dependencies causes less problems on average than things with circular dependencies. Splitting the circular dependency into a category does not remove the dependency, it just occludes it.

It's not always possible/feasible, but here it seems pretty arbitrary that the ConversationColor getters live on UIColor. e.g. why not ConversationColor.forColorName(String) (e.g.) rather than a map on UIColor?

I don't beat this to death... it's fine to leave the circular dependency as is, but I wanted to explain my rationale, since your usage of a category here, apparently only to satisfy me, tells me we had a misunderstanding.

charlesmchen-signal: I see.  Is this what you have in mind?

https://github.com/signalapp/Signal-iOS/pull/3921

PR #3913, opened 2018-09-26
Charlesmchen/conversation colors profile shadows

charlesmchen-signal:
PTAL @michaelkirk


charlesmchen-signal:
> I thought we were doing a stroke - but if this is Myle's desire, then LGTM!

I confirmed with Myles.  I think the intent here is for the "border" to be so subtle that it only shows when absolutely needed, e.g. when the content at the edge closely matches the background against which it is presented.

PR #3914, opened 2018-09-26
Add inner shadows to media thumbnails.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleShapeView.h
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleShapeView.h
@@ -24,11 +24,16 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, nullable) UIColor *strokeColor;
 @property (nonatomic) CGFloat strokeThickness;
 
+@property (nonatomic, nullable) UIColor *innerShadowColor;
+@property (nonatomic) CGFloat innerShadowRadius;
+@property (nonatomic) float innerShadowOpacity;
+
 - (instancetype)init NS_UNAVAILABLE;
 
 + (OWSBubbleShapeView *)bubbleDrawView;
 + (OWSBubbleShapeView *)bubbleShadowView;
 + (OWSBubbleShapeView *)bubbleClipView;
++ (OWSBubbleShapeView *)bubbleInnerShadowView;
 
 @end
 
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -352,13 +352,14 @@ - (void)configureViews
...
-                strokeView.strokeThickness = CGHairlineWidth();
-                strokeView.strokeColor = (Theme.isDarkThemeEnabled ? [UIColor colorWithWhite:1.f alpha:0.2f]
-                                                                   : [UIColor colorWithWhite:0.f alpha:0.2f]);
-                [bodyMediaView addSubview:strokeView];
-                [self.bubbleView addPartnerView:strokeView];
-                [self.viewConstraints addObjectsFromArray:[strokeView ows_autoPinToSuperviewEdges]];
+                OWSBubbleShapeView *innerShadowView = [OWSBubbleShapeView bubbleInnerShadowView];
+                innerShadowView.innerShadowColor
+                    = (Theme.isDarkThemeEnabled ? UIColor.ows_whiteColor : UIColor.ows_blackColor);
+                innerShadowView.innerShadowRadius = 0.5f;
```
michaelkirk-signal: I'm not sure what the perf implications for this are, but if we're setting three properties, and each triggers `updateLayers`, maybe we should instead have a single three-arg-setter which does a single `updateLayers`

TIOLI though - maybe this method isn't a hotspot.

charlesmchen-signal: Replaced the factory methods with initializers that take the required configuration for each mode to avoid this.  I'm not worried about the perf though.

Signal/src/ViewControllers/ConversationView/Cells/OWSBubbleShapeView.h
```diff
@@ -24,11 +24,21 @@ NS_ASSUME_NONNULL_BEGIN
...
+@property (nonatomic) float innerShadowOpacity;
+
 - (instancetype)init NS_UNAVAILABLE;
 
-+ (OWSBubbleShapeView *)bubbleDrawView;
-+ (OWSBubbleShapeView *)bubbleShadowView;
-+ (OWSBubbleShapeView *)bubbleClipView;
+- (instancetype)initDraw NS_DESIGNATED_INITIALIZER;
+- (instancetype)initShadow NS_DESIGNATED_INITIALIZER;
+;
```
michaelkirk-signal: format? what are these extra semicolons?

PR #3915, opened 2018-09-27
String cleanup

jlund-signal:
* Made our Title Case usage a little more consistent.
* Standardized on the ellipses special character.
* Rephrased a few buttons and descriptions.

### DIFF
--- a/Signal/translations/en.lproj/Localizable.strings
+++ b/Signal/translations/en.lproj/Localizable.strings
@@ -69,7 +69,7 @@
 "APN_MESSAGE_IN_GROUP_DETAILED" = "%@ in group %@: %@";
 
 /* Message for the 'app launch failed' alert. */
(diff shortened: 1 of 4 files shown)


Signal/translations/en.lproj/Localizable.strings
```diff
@@ -300,7 +300,7 @@
 "CALL_AGAIN_BUTTON_TITLE" = "Call Again";
 
 /* Alert message when calling and permissions for microphone are missing */
-"CALL_AUDIO_PERMISSION_MESSAGE" = "Signal requires access to your microphone to make calls and record voice messages. You can grant this permission in the Settings app.";
+"CALL_AUDIO_PERMISSION_MESSAGE" = "You can enable microphone access in the iOS Settings app to make calls and record voice messages in Signal.";
```
charlesmchen-signal: This is fine.  I wonder if "Enable microphone access..." would be better & shorter?  Just throwing it out there.

Also, I wonder if "record voice messages" or "send voice messages" is better?

charlesmchen-signal: I take it back, it's better as you have it.

PR #3918, opened 2018-09-27
Map the legacy conversation color names.

charlesmchen-signal:
PTAL @michaelkirk


SignalMessaging/categories/UIColor+OWS.m
```diff
@@ -488,13 +488,51 @@ + (UIColor *)ows_steelShadeColor
...
+}
+
 + (NSArray<NSString *> *)ows_conversationColorNames
 {
     return self.ows_conversationColorMap.allKeys;
 }
 
 + (nullable OWSConversationColor *)ows_conversationColorForColorName:(NSString *)conversationColorName
 {
+    NSString *_Nullable mappedColorName = self.ows_legacyConversationColorMap[conversationColorName.lowercaseString];
```
charlesmchen-signal: Note that the new color names values are all lowercase.

PR #3925, opened 2018-09-28
Mkirk/picker bubble preview

michaelkirk-signal:
PTAL @charlesmchen - it's a big diff, but the majority of changes are the mechanical ones in ConversationViewItem

Basically, our views are really tightly coupled to our db models. Introducing the ConversationViewItem class a while back moved us in the right direction - but there's still a couple of problems.

First problem: ConversationViewItems are initialized with a transaction, and read a bunch of state from the db

I've addressed this by moving the CVI to a protocol.

Second problem: lots of view logic calls methods on `ConversationViewItem.interaction`, which sort of sabotages our attempts to decouple view logic from the database models.

e.g. a canonical example:

To get delivery/read state, the FooterView grabs viewItem.interaction and passes it to MessageStatusUtils.

If instead FooterView just asked for `viewItem.deliveryStatus` we could easily stub this behind the protocol.

In the interest of stability and not changing more production code, I opted *not* to do the second part, and instead juggled some awkward MockDatabaseObjects that will result in the proper behavior, but it's somewhat brittle IMO.

screen shot 2018-09-28 at 1 26 25 pm


Signal/src/Models/MessageActions.swift
```diff
@@ -67,66 +67,67 @@ struct MessageActionBuilder {
     }
 }
 
-extension ConversationViewItem {
+@objc
```
michaelkirk-signal: This was an expedient, but ultimately not very good use of `extension` that I regret. 

It was pretty straight forward, but was the only non-mechanical change related to moving ConversationViewItem to be a protocol.

charlesmchen-signal: I'm curious why you regret it?

Signal/src/ViewControllers/ColorPickerViewController.swift
```diff
@@ -78,15 +78,18 @@ class ColorPicker: NSObject, ColorPickerViewDelegate {
     private let currentConversationColor: OWSConversationColor
 
     @objc
-    init(currentConversationColor: OWSConversationColor) {
+    init(thread: TSThread) {
+        let colorName = thread.conversationColorName
+        let currentConversationColor = OWSConversationColor.conversationColorOrDefault(colorName: colorName)
         self.currentConversationColor = currentConversationColor
```
charlesmchen-signal: NBD but why bother with `currentConversationColor` local variable?

Signal/src/ViewControllers/MessageDetailViewController.swift
```diff
@@ -663,7 +663,7 @@ class MessageDetailViewController: OWSViewController, MediaGalleryDataSourceDele
 
         if let audioAttachmentPlayer = self.audioAttachmentPlayer {
             // Is this player associated with this media adapter?
-            if audioAttachmentPlayer.owner as? ConversationViewItem == viewItem {
+            if audioAttachmentPlayer.owner === viewItem {
```
michaelkirk-signal: This was also not a mechanical change - ultimately what we want is reference equality, so this is simpler.

Signal/src/ViewControllers/ColorPickerViewController.swift
```diff
@@ -166,6 +191,58 @@ class ColorPickerView: UIView, ColorViewDelegate {
         return headerView
     }
 
+    private func updateMockConversationView() {
+        conversationStyle.viewWidth = max(bounds.size.width, kMinimumConversationWidth)
+        mockConversationView.subviews.forEach { $0.removeFromSuperview() }
+
+        // outgoing
+        outgoingMessageView = OWSMessageBubbleView(forAutoLayout: ())
+        outgoingMessageView.prepareForReuse()
```
charlesmchen-signal: I'm surprised that `prepareForReuse` is necessary here since we just made the view.

Signal/src/ViewControllers/ColorPickerViewController.swift
```diff
@@ -166,6 +191,58 @@ class ColorPickerView: UIView, ColorViewDelegate {
...
+        outgoingMessageView.loadContent()
+        let outgoingCell = UIView()
+        outgoingCell.addSubview(outgoingMessageView)
+        outgoingMessageView.autoPinEdgesToSuperviewEdges(with: .zero, excludingEdge: .leading)
+        let outgoingSize = outgoingMessageView.measureSize()
+        outgoingMessageView.autoSetDimensions(to: outgoingSize)
+
+        // incoming
+        incomingMessageView = OWSMessageBubbleView(forAutoLayout: ())
+        incomingMessageView.prepareForReuse()
```
charlesmchen-signal: Ditto.

Signal/src/Models/MessageActions.swift
michaelkirk-signal: I think extensions work best when the methods are dead simple and used all over the place.

But they aren't super flexible - e.g. you can't add methods to a protocol exported to objc, because objc doesn't have "default protocol implementations" like swift does.

Signal/src/ViewControllers/ColorPickerViewController.swift
michaelkirk-signal: You're right, we no longer need it. Removed.

Signal/src/ViewControllers/ColorPickerViewController.swift
michaelkirk-signal: Thanks for catching that - I added it during some earlier debugging. I've removed it.

Signal/src/ViewControllers/ColorPickerViewController.swift
```diff
@@ -185,3 +262,200 @@ class ColorPickerView: UIView, ColorViewDelegate {
...
+        return ConversationViewCell(forAutoLayout: ())
+    }
+
+    func replace(_ interaction: TSInteraction, transaction: YapDatabaseReadTransaction) {
+        owsFailDebug("unexpected invocation")
+        return
+    }
+
+    func clearCachedLayoutState() {
+        owsFailDebug("unexpected invocation")
```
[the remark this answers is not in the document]
michaelkirk-signal: doesn't seem production crash worthy to me.

PR #3930, opened 2018-10-01
fixup some tests

michaelkirk-signal:
PTAL @charlesmchen-signal


Signal/test/util/CDSSigningCertificateTest.m
```diff
@@ -17,25 +17,32 @@ @implementation CDSSigningCertificateTest
...
+    XCTAssertNil(error);
 }
 
 - (void)testParsing_bad
 {
     NSString *pem = [self certificatesPem_bad];
-
-    XCTAssertThrows([CDSSigningCertificate parseCertificateFromPem:pem]);
+    NSError *error;
+    CDSSigningCertificate *_Nullable certificate = [CDSSigningCertificate parseCertificateFromPem:pem error:&error];
```
michaelkirk-signal: firstly, this method doesn't throw anymore, but secondly, but checking specific error types we can have a better sense about which paths are tested.

Signal/test/util/UtilTest.m
```diff
@@ -35,7 +35,6 @@ - (NSString *)filterUnsafeFilenameCharacters;
 @implementation UtilTest
 
 -(void) testRemoveAllCharactersIn {
-    testThrows([@"" removeAllCharactersIn:nil]);
```
michaelkirk-signal: This is not a nullable param, and we assert it's presence which was causing the test to crash.

SignalServiceKit/src/Contacts/CDSSigningCertificate.h
```diff
@@ -2,11 +2,28 @@
 //  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
 //
 
+#import "OWSError.h"
+
 NS_ASSUME_NONNULL_BEGIN
 
+typedef NS_ENUM(NSUInteger, CDSSigningCertificateErrorCode) {
```
michaelkirk-signal: This is a diversion from our usual approach of piling everything into OWSError, but I'm not sure there's much value in having everything piled into OWSError - the downside of doing it that way is I think error codes get - reused for different things, and having smaller enums defined within the context of their usage gives a better hint as to what they mean.

Instead, I'm proposing that we  (roughly) use the classname (or related class name in the case of small internal classes that produce errors) as the domain and defining error codes scoped to that.

charlesmchen-signal: That seems reasonable.

SignalServiceKit/src/Util/Cryptography.m
```diff
@@ -461,7 +461,7 @@ + (nullable NSData *)decryptAttachment:(NSData *)dataToDecrypt
                                                         matchingHMAC:hmac
                                                               digest:digest];
     if (!paddedPlainText) {
-        OWSFailDebug(@"couldn't decrypt attachment.");
+        OWSFailDebugUnlessRunningTests(@"couldn't decrypt attachment.");
```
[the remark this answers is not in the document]
michaelkirk-signal: We have red tests which intentionally fail decrypt - e.g. testDecryptAttachmentWithBadDigest

It'd be nice if `decryptCBCMode:encryptedAttachment` took an **error param and surfaced the specific failure.

SignalServiceKit/tests/Util/CryptographyTests.m
```diff
@@ -105,11 +107,14 @@ - (void)testDecryptAttachmentWithBadKey
     NSData *badKey = [Cryptography generateRandomBytes:64];
 
     NSError *error;
-    XCTAssertThrows([Cryptography decryptAttachment:cipherText
```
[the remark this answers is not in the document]
michaelkirk-signal: Yup, I can wait until thats merged so I can rebase.

PR #3934, opened 2018-10-01
Fix: group color change doesn't immediately apply

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -3239,17 +3239,7 @@ typedef enum : NSUInteger {
 
     if (self.isGroupConversation) {
         [self.uiDatabaseConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3239,17 +3239,7 @@ - (void)uiDatabaseDidUpdate:(NSNotification *)notification
...
-            if (gThread.groupModel) {
-                TSGroupThread *_Nullable updatedThread =
-                    [TSGroupThread threadWithGroupId:gThread.groupModel.groupId transaction:transaction];
-                if (updatedThread) {
-                    self.thread = updatedThread;
-                } else {
-                    OWSFailDebug(@"Could not reload thread.");
-                }
-            }
+            [self.thread reloadWithTransaction:transaction];
```
michaelkirk-signal: Because we were fetching a new thread instance, instead of updating the
existing thread instance, classes which were bound to the old thread instance
weren't updating. This affected the HeaderView.AvatarView.thread and the
ConversationStyle.thread.

PR #3937, opened 2018-10-04
Apply new colors to new conversation

michaelkirk-signal:
We were using the old color names, which would get mapped to the new colors, but precludes a couple of options from being selected, and would have a skewed distribution, since multiple legacy colors map to some new colors.

PTAL @charlesmchen


michaelkirk-signal:
Actually... not ready for review yet.

PR #3944, opened 2018-10-08
fix archive nav stack

michaelkirk-signal:
fixes https://github.com/signalapp/Signal-iOS/issues/3943

PTAL @charlesmchen-signal


michaelkirk-signal:
Without it, there is no way to return to your inbox after viewing an archived conversation, short of killing the app.

So I think it definitely needs to go into this release.

charlesmchen-signal:
@michaelkirk Okay, let me know when you're merged up and I'll cut a new build.

PR #3947, opened 2018-10-09
Fix settings button glitch on iOS 10.

charlesmchen-signal:
PTAL @michaelkirk


michaelkirk-signal:
LGTM

Fixes https://github.com/signalapp/Signal-iOS/issues/3945

PR #3950, opened 2018-10-11
dark theme toggle in app settings

michaelkirk-signal:
PTAL @charlesmchen-signal 

dark-theme-toggle-off
dark-theme-toggle-on


charlesmchen-signal:
Probably not worth it, but IMO the icon feels a little (2pt?) vertically low due to its asymmetry.

PR #3967, opened 2018-10-31
Fix failed reloads in SignalRecipient.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Contacts/SignalRecipient.m
+++ b/SignalServiceKit/src/Contacts/SignalRecipient.m
@@ -175,7 +175,7 @@ NS_ASSUME_NONNULL_BEGIN
     OWSAssertDebug(devices.count > 0);
 
     OWSLogDebug(@"removing devices: %@, from registered recipient: %@", devices, self);
-    [self reloadWithTransaction:transaction];
+    [self reloadWithTransaction:transaction ignoreMissing:YES];
     [self removeDevices:devices];
     [self saveWithTransaction_internal:transaction];
 }
--- a/SignalServiceKit/src/Storage/TSYapDatabaseObject.h
+++ b/SignalServiceKit/src/Storage/TSYapDatabaseObject.h
@@ -94,8 +94,9 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  * Assign the latest persisted values from the database.
  */
-- (void)reloadWithTransaction:(YapDatabaseReadTransaction *)transaction;
 - (void)reload;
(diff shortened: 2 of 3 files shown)


SignalServiceKit/src/Contacts/SignalRecipient.m
```diff
@@ -170,7 +170,7 @@ - (void)removeDevicesFromRecipient:(NSSet *)devices transaction:(YapDatabaseRead
     OWSAssertDebug(devices.count > 0);
 
     OWSLogDebug(@"removing devices: %@, from registered recipient: %@", devices, self);
-    [self reloadWithTransaction:transaction];
+    [self reloadWithTransaction:transaction ignoreMissing:YES];
```
charlesmchen-signal: We sometimes use unsaved `SignalRecipient` instances when sending.

charlesmchen-signal:
> To be clear, this isn't a change in production behavior right, 

Yes, it isn't a change in production behavior, but...

> it just might help find issues in debug?

Not quite - this is intended to fix an existing issue here.  

`reloadWithTransaction:transaction` isn't intended to be called on unsaved entities.  We sometimes use unsaved instances of `SignalRecipient` (built by `getOrBuildUnsavedRecipientForRecipientId`) when message sending for the first time to users not in our `SignalRecipient` cache.  If that user is unregistered, we'll get a 409/410 to remove their primary.  In debug, we'll hit an assert in `reloadWithTransaction:transaction`.  

This PR modifies `reloadWithTransaction:transaction` to take an argument that indicates whether or not we expect the entity to already exist in the database.

michaelkirk-signal:
>> it just might help find issues in debug?
>
> Not quite - this is intended to fix an existing issue here.
> ...
> This PR modifies reloadWithTransaction:transaction to take an argument that indicates whether or not we expect the entity to already exist in the database.

But that argument doesn't affect anything other than whether or not to fire a debug assertion. So how could it fix anything? Am I misreading the diff?

charlesmchen-signal:
> But that argument doesn't affect anything other than whether or not to fire a debug assertion. So how could it fix anything? Am I misreading the diff?

Right.  We shouldn't be hitting the assert in that case - it's _expected_ in some cases.

e.g. clean install, send a message to a group with an unregistered user.  Send will use an "unsaved SignalRecipient", will fail with 409/410 to remove primary, bang. 

> it just might help find issues in debug?

All I'm trying to say is that change isn't to find some unknown issue, it's to fix an overzealous assert.

PR #3968, opened 2018-10-31
Add ConversationViewModel.

charlesmchen-signal:
Pulls out a VM for the conversation view.  It corresponds to "the conversation".  

For clarity, we could eventually rename the existing `ConversationViewItem` to `ConversationViewModelItem`, since they are sub-viewmodels, representing items in the conversation.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -14,6 +14,7 @@
 #import "ConversationViewCell.h"
 #import "ConversationViewItem.h"
 #import "ConversationViewLayout.h"
+#import "ConversationViewModel.h"
 #import "DateUtil.h"
 #import "DebugUITableViewController.h"
 #import "FingerprintViewController.h"
@@ -91,24 +92,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-// Always load up to n messages when user arrives.
-//
-// The smaller this number is, the faster the conversation can display.
-// To test, shrink you accessability font as much as possible, then count how many 1-line system info messages (our shortest cells) can
-// fit on screen at a time on an iPhoneX
-//
-// PERF: we could do less messages on shorter (older, slower) devices
-// PERF: we could cache the cell height, since some messages will be much taller.
-static const int kYapDatabasePageSize = 25;
-
-// Never show more than n messages in conversation view when user arrives.
-static const int kConversationInitialMaxRangeSize = 300;
-
-// Never show more than n messages in conversation view at a time.
-static const int kYapDatabaseRangeMaxLength = 25000;
-
-static const int kYapDatabaseRangeMinLength = 0;
-
 static const CGFloat kLoadMoreHeaderHeight = 60.f;
 
 static const CGFloat kToastInset = 10;
@@ -146,38 +129,19 @@ typedef enum : NSUInteger {
     UITextViewDelegate,
     ConversationCollectionViewDelegate,
     ConversationInputToolbarDelegate,
-    GifPickerViewControllerDelegate>
+    GifPickerViewControllerDelegate,
+    ConversationViewModelDelegate>
 
 @property (nonatomic) TSThread *thread;
-@property (nonatomic, readonly) YapDatabaseConnection *editingDatabaseConnection;
+@property (nonatomic, readonly) ConversationViewModel *conversationViewModel;
+
 @property (nonatomic, readonly) OWSAudioActivity *recordVoiceNoteAudioActivity;
 @property (nonatomic, readonly) NSTimeInterval viewControllerCreatedAt;
-
-// These two properties must be updated in lockstep.
-//
-// * The first (required) step is to update uiDatabaseConnection using beginLongLivedReadTransaction.
-// * The second (required) step is to update messageMappings.
-// * The third (optional) step is to update the messageMappings range using
-//   updateMessageMappingRangeOptions.
-// * The fourth (optional) step is to update the view items using reloadViewItems.
-// * The steps must be done in strict order.
-// * If we do any of the steps, we must do all of the required steps.
-// * We can't use messageMappings or viewItems after the first step until we've
-//   done the last step; i.e.. we can't do any layout, since that uses the view
-//   items which haven't been updated yet.
-// * If the first and/or second steps changes the set of messages
-//   their ordering and/or their state, we must do the third and fourth steps.
-// * If we do the third step, we must call resetContentAndLayout afterward.
-@property (nonatomic, readonly) YapDatabaseConnection *uiDatabaseConnection;
-@property (nonatomic) YapDatabaseViewMappings *messageMappings;
 
 @property (nonatomic, readonly) ConversationInputToolbar *inputToolbar;
 @property (nonatomic, readonly) ConversationCollectionView *collectionView;
 @property (nonatomic, readonly) ConversationViewLayout *layout;
 @property (nonatomic, readonly) ConversationStyle *conversationStyle;
-
-@property (nonatomic) NSArray<id<ConversationViewItem>> *viewItems;
-@property (nonatomic) NSMutableDictionary<NSString *, id<ConversationViewItem>> *viewItemCache;
 
 @property (nonatomic, nullable) AVAudioRecorder *audioRecorder;
 @property (nonatomic, nullable) OWSAudioPlayer *audioAttachmentPlayer;
@@ -194,9 +158,7 @@ typedef enum : NSUInteger {
 @property (nonatomic, readonly) UILabel *backButtonUnreadCountLabel;
 @property (nonatomic, readonly) NSUInteger backButtonUnreadCount;
 
-@property (nonatomic) NSUInteger lastRangeLength;
 @property (nonatomic) ConversationViewAction actionOnOpen;
-@property (nonatomic, nullable) NSString *focusMessageIdOnOpen;
 
 @property (nonatomic) BOOL peek;
 
@@ -205,8 +167,6 @@ typedef enum : NSUInteger {
 @property (nonatomic) BOOL userHasScrolled;
 @property (nonatomic, nullable) NSDate *lastMessageSentDate;
 
-@property (nonatomic, nullable) ThreadDynamicInteractions *dynamicInteractions;
-@property (nonatomic) BOOL hasClearedUnreadMessagesIndicator;
 @property (nonatomic) BOOL showLoadMoreHeader;
 @property (nonatomic) UILabel *loadMoreHeader;
 @property (nonatomic) uint64_t lastVisibleTimestamp;
@@ -222,16 +182,19 @@ typedef enum : NSUInteger {
 
 @property (nonatomic) BOOL isViewCompletelyAppeared;
 @property (nonatomic) BOOL isViewVisible;
-@property (nonatomic) BOOL shouldObserveDBModifications;
+@property (nonatomic) BOOL shouldObserveVMUpdates;
 @property (nonatomic) BOOL viewHasEverAppeared;
 @property (nonatomic) BOOL hasUnreadMessages;
 @property (nonatomic) BOOL isPickingMediaAsDocument;
 @property (nonatomic, nullable) NSNumber *previousLastTimestamp;
+@property (nonatomic, nullable) NSNumber *previousViewItemCount;
+@property (nonatomic, nullable) NSNumber *previousViewTopToContentBottom;
 @property (nonatomic, nullable) NSNumber *viewHorizonTimestamp;
 @property (nonatomic) ContactShareViewHelper *contactShareViewHelper;
 @property (nonatomic) NSTimer *reloadTimer;
 @property (nonatomic, nullable) NSDate *lastReloadDate;
-@property (nonatomic, nullable) NSDate *collapseCutoffDate;
+
+@property (nonatomic) CGFloat scrollDistanceToBottomSnapshot;
 
 @end
 
@@ -354,18 +317,6 @@ typedef enum : NSUInteger {
                                                  name:UIContentSizeCategoryDidChangeNotification
                                                object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(uiDatabaseDidUpdateExternally:)
-                                                 name:OWSUIDatabaseConnectionDidUpdateExternallyNotification
-                                               object:OWSPrimaryStorage.sharedManager.dbNotificationObject];
-    [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(uiDatabaseWillUpdate:)
-                                                 name:OWSUIDatabaseConnectionWillUpdateNotification
-                                               object:OWSPrimaryStorage.sharedManager.dbNotificationObject];
-    [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(uiDatabaseDidUpdate:)
-                                                 name:OWSUIDatabaseConnectionDidUpdateNotification
-                                               object:OWSPrimaryStorage.sharedManager.dbNotificationObject];
-    [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(applicationWillEnterForeground:)
                                                  name:OWSApplicationWillEnterForegroundNotification
                                                object:nil];
@@ -394,10 +345,6 @@ typedef enum : NSUInteger {
                                                  name:kNSNotificationName_ProfileWhitelistDidChange
                                                object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(signalAccountsDidChange:)
-                                                 name:OWSContactsManagerSignalAccountsDidChangeNotification
-                                               object:nil];
-    [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(keyboardWillChangeFrame:)
                                                  name:UIKeyboardWillChangeFrameNotification
                                                object:nil];
@@ -410,12 +357,6 @@ typedef enum : NSUInteger {
     return self.thread.isGroupThread;
 }
 
-- (void)signalAccountsDidChange:(NSNotification *)notification
-{
-    OWSAssertIsOnMainThread();
-
-    [self ensureDynamicInteractions];
-}
 
 - (void)otherUsersProfileDidChange:(NSNotification *)notification
 {
@@ -445,11 +386,11 @@ typedef enum : NSUInteger {
     NSString *_Nullable recipientId = notification.userInfo[kNSNotificationKey_ProfileRecipientId];
     NSData *_Nullable groupId = notification.userInfo[kNSNotificationKey_ProfileGroupId];
     if (recipientId.length > 0 && [self.thread.recipientIdentifiers containsObject:recipientId]) {
-        [self ensureDynamicInteractions];
+        [self.conversationViewModel ensureDynamicInteractions];
     } else if (groupId.length > 0 && self.thread.isGroupThread) {
         TSGroupThread *groupThread = (TSGroupThread *)self.thread;
         if ([groupThread.groupModel.groupId isEqualToData:groupId]) {
-            [self ensureDynamicInteractions];
+            [self.conversationViewModel ensureDynamicInteractions];
             [self ensureBannerState];
         }
     }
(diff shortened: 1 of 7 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -90,24 +91,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-// Always load up to n messages when user arrives.
```
charlesmchen-signal: There's a lot of red in this diff.  I'd love to continue pulling out other pieces of CVC into separate modules.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -321,18 +283,6 @@ - (void)addNotificationListeners
...
-                                                 name:OWSUIDatabaseConnectionDidUpdateExternallyNotification
-                                               object:OWSPrimaryStorage.sharedManager.dbNotificationObject];
-    [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(uiDatabaseWillUpdate:)
-                                                 name:OWSUIDatabaseConnectionWillUpdateNotification
-                                               object:OWSPrimaryStorage.sharedManager.dbNotificationObject];
-    [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(uiDatabaseDidUpdate:)
-                                                 name:OWSUIDatabaseConnectionDidUpdateNotification
-                                               object:OWSPrimaryStorage.sharedManager.dbNotificationObject];
```
charlesmchen-signal: CVC no longer responds to DB change events - that's the CVM's job.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -413,11 +353,11 @@ - (void)profileWhitelistDidChange:(NSNotification *)notification
     NSString *_Nullable recipientId = notification.userInfo[kNSNotificationKey_ProfileRecipientId];
     NSData *_Nullable groupId = notification.userInfo[kNSNotificationKey_ProfileGroupId];
     if (recipientId.length > 0 && [self.thread.recipientIdentifiers containsObject:recipientId]) {
-        [self ensureDynamicInteractions];
+        [self.conversationViewModel ensureDynamicInteractions];
```
charlesmchen-signal: Eventually, I hope we can do away with "dynamic interactions" entirely and make them:

* "ephemeral/transient" (never persisted in DB)
* created/updated by the CVM.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1144,7 +1081,7 @@ - (void)viewDidAppear:(BOOL)animated
     [self updateBackButtonUnreadCount];
     [self autoLoadMoreIfNecessary];
 
-    self.focusMessageIdOnOpen = nil;
+    self.conversationViewModel.focusMessageIdOnOpen = nil;
```
charlesmchen-signal: In the long run, I think we can achieve a cleaner & simpler relationship between CVC and CVM.  For example, I'd rather that (eventually) CVM not know about `focusMessageIdOnOpen` but instead know about `messageIdToTryToIncludeInLoadWindow` (or whatever).  But I'm trying to take this PR as a first step toward disentangling the CVC and CVM.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3201,433 +2871,88 @@ - (YapDatabaseConnection *)editingDatabaseConnection
...
-- (void)uiDatabaseDidUpdateExternally:(NSNotification *)notification
+- (BOOL)isScrolledToBottom
 {
-    OWSAssertIsOnMainThread();
+    CGFloat contentHeight = self.safeContentHeight;
 
-    OWSLogVerbose(@"");
+    // This is a bit subtle.
+    //
+    // The _wrong_ way to determine if we're scrolled to the bottom is to
```
charlesmchen-signal: This method isn't new; it just got moved around.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4578,84 +3903,58 @@ - (void)updateCellsVisible
...
+        self.shouldObserveVMUpdates = NO;
         return;
     }
 
-    self.shouldObserveDBModifications = YES;
+    self.shouldObserveVMUpdates = YES;
 }
 
-- (void)setShouldObserveDBModifications:(BOOL)shouldObserveDBModifications
+- (void)setShouldObserveVMUpdates:(BOOL)shouldObserveVMUpdates
```
charlesmchen-signal: This is an important change.  Previously the CVC observed DB changes, but paused observation while it was hidden or the app was in the background.

Now, the CVM observes DB changes, and the CVC observes CVM updates.  When the view is hidden or the app is in the background, the CVM _continues to update_ to reflect DB changes, but the CVC ignores CVM updates.  

Ignoring changes when not visible was important for perf, but introduced some painful complications.  I think the arrangement will be considerably more convenient.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -4749,107 +4067,6 @@ - (nullable NSIndexPath *)firstIndexPathAtViewHorizonTimestamp
     }
 }
 
-// We stop observing database modifications when the app or this view is not visible
-// (see: shouldObserveDBModifications).  When we resume observing db modifications,
-// we want to extend the "range" of this view to include any items added to this
-// thread while we were not observing.
-- (void)extendRangeToIncludeUnobservedItems
```
charlesmchen-signal: We no longer need to do this, since the VM continues to update when the view is hidden.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -5384,6 +4290,241 @@ - (void)presentViewController:(UIViewController *)viewController
     [super presentViewController:viewController animated:animated completion:completion];
 }
 
+#pragma mark - ConversationViewModelDelegate
```
charlesmchen-signal: These are the key additions to the class - the CVC is the delegate for the CVM.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -5384,6 +4290,241 @@ - (void)presentViewController:(UIViewController *)viewController
...
+
+- (void)conversationViewModelWillUpdate
```
charlesmchen-signal: Corresponds to the old "DB will update" handler.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -5384,6 +4290,241 @@ - (void)presentViewController:(UIViewController *)viewController
...
+    // This was our #2 crash, and much exacerbated by the refactoring somewhere between 2.6.2.0-2.6.3.8
+    //
+    // NOTE: It's critical we do this before beginLongLivedReadTransaction.
+    //       We want to relayout our contents using the old message mappings and
+    //       view items before they are updated.
+    [self.collectionView layoutIfNeeded];
+    // ENDHACK to work around radar #28167779
+}
+
+- (void)conversationViewModelDidUpdate:(ConversationUpdate *)conversationUpdate
```
charlesmchen-signal: Corresponds to the old "db did update" handler.  Note that we can receive 3 kinds of updates:

* Minor (no changes to conversation items, but may need to update other subviews like the nav bar).
* Diff (an incremental update; some items were added/removed/updated).
* Reload (incremental update isn't possible, reload the entire view).

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -5384,6 +4290,241 @@ - (void)presentViewController:(UIViewController *)viewController
...
+    // a) already was at the bottom of the conversation.
+    // b) is inserting new interactions.
+    __block BOOL scrollToBottom = wasAtBottom;
+
+    void (^batchUpdates)(void) = ^{
+        OWSAssertIsOnMainThread();
+
+        const NSUInteger section = 0;
+        BOOL hasInserted = NO, hasUpdated = NO;
+        for (ConversationUpdateItem *updateItem in conversationUpdate.updateItems) {
```
charlesmchen-signal: Much of this logic around handling incremental updates should look familiar.  The difference is that we now use `ConversationUpdateItem` instead of `YapDatabaseViewRowChange`.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.h
```diff
@@ -121,8 +121,14 @@ NSString *NSStringForOWSMessageCellType(OWSMessageCellType cellType);
 
 - (BOOL)canSaveMedia;
 
+// For view items that correspond to interactions, this is the interaction's unique id.
+// For other view views (like the typing indicator), this is a unique, stable string.
+- (NSString *)itemId;
```
charlesmchen-signal: I've introduce the concept of an "item id".  In the past, all conversation items corresponded to a `TSInteraction` persisted in the DB.  Soon, we'll have "transient" conversation items (like typing indicators, unread indicators, date breaks, etc.).

Signal/src/ViewControllers/ConversationView/ConversationViewModel.h
```diff
@@ -0,0 +1,108 @@
...
+@class ThreadDynamicInteractions;
+
+@protocol ConversationViewItem;
+
+typedef NS_ENUM(NSUInteger, ConversationUpdateType) {
+    // No view items in the load window were effected.
+    ConversationUpdateType_Minor,
+    // A subset of view items in the load window were effected;
+    // the view should be updated using the update items.
+    ConversationUpdateType_Diff,
```
charlesmchen-signal: Maybe `ConversationUpdateType_Incremental` would be a better name.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -0,0 +1,1288 @@
...
+// PERF: we could cache the cell height, since some messages will be much taller.
+static const int kYapDatabasePageSize = 25;
+
+// Never show more than n messages in conversation view when user arrives.
+static const int kConversationInitialMaxRangeSize = 300;
+
+// Never show more than n messages in conversation view at a time.
+static const int kYapDatabaseRangeMaxLength = 25000;
+
+static const int kYapDatabaseRangeMinLength = 0;
```
charlesmchen-signal: Much of the contents of the CVM were drawn from the CVC, without radical changes.  There's a bit more work to use the `delegate` to coordinate certain activities.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -0,0 +1,1284 @@
...
+        [self resetMappings];
+    }
+}
+
+- (void)uiDatabaseWillUpdate:(NSNotification *)notification
+{
+    [self.delegate conversationViewModelWillUpdate];
+}
+
+- (void)uiDatabaseDidUpdate:(NSNotification *)notification
```
charlesmchen-signal: Like the old CVC, the CVM uses this method to respond to DB modifications.  Specifically, it:

* Updates its internal state (the view items).  This logic is almost identical.
* Fires update events to the CVC.  This is new.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -0,0 +1,1284 @@
...
+        // These errors are rare.
+        OWSFailDebug(@"could not reload view items; hard resetting message mappings.");
+        // resetMappings will call delegate.conversationViewModelDidUpdate.
+        [self resetMappings];
+        return;
+    }
+
+    OWSLogVerbose(@"self.viewItems.count: %zd -> %zd", oldViewItemCount, self.viewItems.count);
+
+    [self updateViewWitholdItemIdList:oldItemIdList
```
charlesmchen-signal: Here's the bulk of the new logic, where we create a incremental/diff update for the CVC.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -0,0 +1,1284 @@
...
+    NSMutableSet<NSString *> *deletedItemIdSet = [oldItemIdSet mutableCopy];
+    [deletedItemIdSet minusSet:newItemIdSet];
+    NSMutableSet<NSString *> *insertedItemIdSet = [newItemIdSet mutableCopy];
+    [insertedItemIdSet minusSet:oldItemIdSet];
+    NSArray<NSString *> *deletedItemIdList = [deletedItemIdSet.allObjects sortedArrayUsingSelector:@selector(compare:)];
+    NSArray<NSString *> *insertedItemIdList =
+        [insertedItemIdSet.allObjects sortedArrayUsingSelector:@selector(compare:)];
+
+    // Try to generate a series of "update items" that safely transform
+    // the "old item list" into the "new item list".
```
charlesmchen-signal: This is the core task.  The guiding principle is that `UICollectionView` incremental updates work best if you do deletes first, inserts second and updates last (because deletes and inserts affect the indices of subsequent operations).

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -0,0 +1,1284 @@
...
+        NSUInteger newIndex = [newItemIdList indexOfObject:itemId];
+        if (newIndex == NSNotFound) {
+            OWSFailDebug(@"Can't find new index of updated view item.");
+            return [self.delegate conversationViewModelDidUpdate:ConversationUpdate.reloadUpdate];
+        }
+        id<ConversationViewItem> _Nullable viewItem = newViewItemMap[itemId];
+        if (!viewItem) {
+            OWSFailDebug(@"Can't find inserted view item.");
+            return [self.delegate conversationViewModelDidUpdate:ConversationUpdate.reloadUpdate];
+        }
```
charlesmchen-signal: We're super defensive when constructing these updates because of the bugs we've seen in the YDB mappings.  One of the benefits to the CVM is that any inconsistencies/bugs in the YDB state/logic are _discarded_ by the CVM.  The CVC only knows about the CVM state (a sequence of items) and learns about changes to that state.  If YDB gets into a bad state, we might omit certain items from the CVM state, but that's hidden from the POV of the CVC.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -0,0 +1,1284 @@
...
+
+    return [self.delegate
+        conversationViewModelDidUpdate:[ConversationUpdate diffUpdateWithUpdateItems:updateItems
+                                                                shouldAnimateUpdates:shouldAnimateUpdates]];
+}
+
+- (BOOL)shouldAnimateUpdateItems:(NSArray<ConversationUpdateItem *> *)updateItems
+                oldViewItemCount:(NSUInteger)oldViewItemCount
+          updatedNeighborItemSet:(nullable NSMutableSet<NSString *> *)updatedNeighborItemSet
+{
```
charlesmchen-signal: I moved this here because the CVM had the necessary state to make this decision and I was reluctant to plumb it through the "CVM update" event.  However, animations are a view concern and probably don't belong in the CVM.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -0,0 +1,1284 @@
...
+}
+
+- (void)dealloc
+{
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
+}
+
+- (BOOL)canLoadMoreItems
+{
+    if (self.lastRangeLength >= kYapDatabaseRangeMaxLength) {
```
charlesmchen-signal: The vast majority of this code in the CVM was drawn from the CVC unchanged.  I'll try to call out big differences.

charlesmchen-signal:
PTAL @michaelkirk

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -0,0 +1,1284 @@
...
+        // resetMappings will call delegate.conversationViewModelDidUpdate.
+        [self resetMappings];
+        return;
+    }
+
+    OWSLogVerbose(@"self.viewItems.count: %zd -> %zd", oldViewItemCount, self.viewItems.count);
+
+    [self updateViewWitholdItemIdList:oldItemIdList
+                       updatedItemSet:updatedItemSet
+                     oldViewItemCount:oldViewItemCount
```
michaelkirk-signal: How is this different from `oldItemIdList.count`?

It reads like redundant state.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
michaelkirk-signal: 👍

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -0,0 +1,1284 @@
...
+        @(+1),
+    ]]
+                                                 forGroup:self.thread.uniqueId];
+
+    [self.uiDatabaseConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
+        [self.messageMappings updateWithTransaction:transaction];
+    }];
+    // We need to impose the range restrictions on the mappings immediately to avoid
+    // doing a great deal of unnecessary work and causing a perf hotspot.
+    [self updateMessageMappingRangeOptions];
```
michaelkirk-signal: Should this happen *before* `updateWithTransaction` to avoid said "perf hotspots"?

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -0,0 +1,1284 @@
...
+}
+
+- (void)resetMappings
+{
+    // If we're entering "active" mode (e.g. view is visible and app is in foreground),
```
michaelkirk-signal: I don't think this comment is relevant any longer - we no longer stop observing, right?

michaelkirk-signal: I'm assuming everything after here was moved unchanged from the CVC, LMK if there's something I should look at.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
charlesmchen-signal: Done.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3238,457 +2909,112 @@ - (YapDatabaseConnection *)editingDatabaseConnection
...
+    CGFloat distanceFromBottom = contentOffsetYBottom - self.collectionView.contentOffset.y;
+    BOOL isScrolledToBottom = distanceFromBottom <= kIsAtBottomTolerancePts;
+
+    return isScrolledToBottom;
 }
 
-- (void)uiDatabaseWillUpdate:(NSNotification *)notification
+#pragma mark - Audio
+
+- (void)requestRecordingVoiceMemo
```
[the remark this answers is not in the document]
charlesmchen-signal: No.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -3238,457 +2909,112 @@ - (YapDatabaseConnection *)editingDatabaseConnection
...
+        } else {
+            OWSLogInfo(@"we do not have recording permission.");
+            [strongSelf cancelVoiceMemo];
+            [OWSAlerts showNoMicrophonePermissionAlert];
+        }
+    }];
 }
 
-- (void)uiDatabaseDidUpdate:(NSNotification *)notification
+- (void)startRecordingVoiceMemo
```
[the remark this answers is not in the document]
charlesmchen-signal: No.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.h
```diff
@@ -0,0 +1,108 @@
...
+@class ConversationViewModel;
+@class OWSQuotedReplyModel;
+@class TSThread;
+@class ThreadDynamicInteractions;
+
+@protocol ConversationViewItem;
+
+typedef NS_ENUM(NSUInteger, ConversationUpdateType) {
+    // No view items in the load window were effected.
+    ConversationUpdateType_Minor,
```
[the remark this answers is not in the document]
charlesmchen-signal: CVC does certain things in response to every DB modification notification:

```

    if (self.isGroupConversation) {
        [self.uiDatabaseConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
            [self.thread reloadWithTransaction:transaction];
        }];
        [self updateNavigationTitle];
    }

    [self updateDisappearingMessagesConfiguration];
```
These update the UI to reflect DB changes that don't affect the conversation items.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
charlesmchen-signal: You're right, that's simpler. Done.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
charlesmchen-signal: The mappings have to be updated before you update their range.  That was very counter-intuitive to me and I never found docs the clearly set out the correct order of these operations, but the the results are wrong if you do it the other way.  

We solved bugs around this a while back.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
charlesmchen-signal: > I don't think this comment is relevant any longer - we no longer stop observing, right?

Removed.

> I'm assuming everything after here was moved unchanged from the CVC, LMK if there's something I should look at.

`ensureLoadWindowContainsQuotedReply` is a new method, but it was derived from similar logic in CVC.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.h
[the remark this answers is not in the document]
michaelkirk-signal: Makes sense. Thanks!

PR #3970, opened 2018-10-31
Add typing indicators setting.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Util/TypingIndicators.swift
+++ b/SignalServiceKit/src/Util/TypingIndicators.swift
@@ -204,7 +204,7 @@ public class TypingIndicatorsImpl: NSObject, TypingIndicators {
                 // If the user types a character into the compose box, and the sendRefresh timer isn’t running:
 
                 // Send a ACTION=TYPING message.
-                sendTypingMessage(forThread: thread, action: .started)
+                sendTypingMessageIfNecessary(forThread: thread, action: .started)
                 // Start the sendRefresh timer for 10 seconds
                 sendRefreshTimer?.invalidate()
                 sendRefreshTimer = Timer.weakScheduledTimer(withTimeInterval: 10,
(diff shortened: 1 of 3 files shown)


SignalServiceKit/src/Util/TypingIndicators.swift
```diff
@@ -27,14 +27,66 @@ public protocol TypingIndicators: class {
...
 }
 
 // MARK: -
 
 @objc(OWSTypingIndicatorsImpl)
 public class TypingIndicatorsImpl: NSObject, TypingIndicators {
+
     @objc public static let typingIndicatorStateDidChange = Notification.Name("typingIndicatorStateDidChange")
 
+    private let kDatabaseCollection = "TypingIndicators"
```
michaelkirk-signal: I know we've done this in multiple places, but it seems weird to have an entire collection (table) which exists just to hold one row with one boolean. In my mind it makes sense to use the existing preferences collection for pretty much all singleton data.

I guess it's less straight forward because SSK doesn't know about OWSPreferences?

SignalServiceKit/src/Util/TypingIndicators.swift
```diff
@@ -238,6 +292,14 @@ public class TypingIndicatorsImpl: NSObject, TypingIndicators {
         private func sendTypingMessage(forThread thread: TSThread, action: TypingIndicatorAction) {
             Logger.verbose("\(TypingIndicatorMessage.string(forTypingIndicatorAction: action))")
 
+            guard let delegate = delegate else {
+                owsFailDebug("Missing delegate.")
+                return
+            }
+            guard delegate.areTypingIndicatorsEnabled() else {
+                return
```
michaelkirk-signal: `sendTypingMessage` is a bit misleading of a method name since we'll bail out here.

or is this just a safety, in which cases this should have an owsfailDebug?

SignalServiceKit/src/Util/TypingIndicators.swift
charlesmchen-signal: Right, and I'm not eager to move `OWSPreferences ` to SSK.

SignalServiceKit/src/Util/TypingIndicators.swift
charlesmchen-signal: `areTypingIndicatorsEnabled` reflects the user-facing setting in the app preferences.  If it's disabled we don't want to emit these messages or show the indicators for other users.  

I'll rename the method and improve the comments.

PR #3972, opened 2018-10-31
fix media-gallery doesn't respect call banner

michaelkirk-signal:
PTAL @charlesmchen 

Came across this while testing the landscape stuff, but actually unrelated.

### DIFF
--- a/Signal/src/ViewControllers/MediaGalleryViewController.swift
+++ b/Signal/src/ViewControllers/MediaGalleryViewController.swift
@@ -168,44 +168,7 @@ protocol MediaGalleryDataSourceDelegate: class {
     func mediaGalleryDataSource(_ mediaGalleryDataSource: MediaGalleryDataSource, deletedSections: IndexSet, deletedItems: [IndexPath])
 }
 
-@objc
-class MediaGalleryViewController: OWSNavigationController, MediaGalleryDataSource, MediaTileViewControllerDelegate {
-
-    private var pageViewController: MediaPageViewController?
-
-    private let uiDatabaseConnection: YapDatabaseConnection
-    private let editingDatabaseConnection: YapDatabaseConnection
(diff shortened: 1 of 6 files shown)


michaelkirk-signal:
This needs some more work. Closing for now.

Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -168,43 +168,10 @@ protocol MediaGalleryDataSourceDelegate: class {
...
-        super.init(nibName: nil, bundle: nil)
-    }
-
-    required init?(coder aDecoder: NSCoder) {
-        notImplemented()
-    }
+class MediaGalleryNavigationController: OWSNavigationController {
 
+    var presentationView: UIImageView!
+    var retainUntilDismissed: MediaGallery?
```
charlesmchen-signal: I don't see this being cleared anywhere.  Will this by a retain cycle?  Are these entities being released?

michaelkirk-signal: Yes, they're being released.

Ownership is basically inverted.

The MediaGallery has a *weak* reference to the MediaGalleryNavigationController it creates, which in turn has a *strong* reference back to the MediaGallery that created it.

So in effect, the lifetime of the media gallery is only as long as something *else* holds a reference to the MediaGalleryNavigationController. 

An alternative would be that everywhere we show the MediaGalleryNavigationController we also hold a reference to the MediaGallery, and then whenever we dismiss the MediaGalleryNavigationController, we also manually clear the reference to the MediaGallery.

While more explicit, that approach seemed more error prone.

I'm not crazy about the current approach, but I'm open to suggestion.

michaelkirk-signal: If you prefer, I could also just revert c5bb924, and merge 97e2c33 which also works.

PR #3976, opened 2018-11-01
Charlesmchen/typing indicators5 

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/TypingIndicatorCell.swift
+++ b/Signal/src/ViewControllers/ConversationView/Cells/TypingIndicatorCell.swift
@@ -56,7 +56,6 @@ public class TypingIndicatorCell: ConversationViewCell {
 
         bubbleView.bubbleColor = conversationStyle.bubbleColor(isIncoming: true)
         typingIndicatorView.startAnimation()
-        typingIndicatorView.addBackgroundView(withBackgroundColor: UIColor.red)
 
         viewConstraints.append(contentsOf: [
             bubbleView.autoPinEdge(toSuperviewEdge: .leading, withInset: conversationStyle.gutterLeading),
--- a/Signal/src/ViewControllers/ConversationView/Cells/TypingIndicatorCell.swift
+++ b/Signal/src/ViewControllers/ConversationView/Cells/TypingIndicatorCell.swift
@@ -43,10 +43,6 @@ public class TypingIndicatorCell: ConversationViewCell {
         avatarView.autoSetDimension(.height, toSize: kAvatarSize)
     }
 
-    deinit {
-        NotificationCenter.default.removeObserver(self)
-    }
-
     @objc
     public override func loadForDisplay() {
         guard let conversationStyle = self.conversationStyle else {
@@ -99,7 +95,7 @@ public class TypingIndicatorCell: ConversationViewCell {
         }
         guard let authorAvatarImage =
             OWSContactAvatarBuilder(signalId: typingIndicators.recipientId,
-                                    colorName: ConversationColorNameForString(colorName),
+                                    colorName: ConversationColorName(rawValue: colorName),
                                     diameter: UInt(kAvatarSize)).build() else {
                                         owsFailDebug("Could build avatar image")
                                         return nil
@@ -148,7 +144,5 @@ public class TypingIndicatorCell: ConversationViewCell {
         avatarView.removeFromSuperview()
 
         typingIndicatorView.stopAnimation()
-
-        NotificationCenter.default.removeObserver(self)
     }
 }
--- a/Signal/src/ViewControllers/ConversationView/Cells/TypingIndicatorCell.swift
+++ b/Signal/src/ViewControllers/ConversationView/Cells/TypingIndicatorCell.swift
@@ -10,28 +10,146 @@ public class TypingIndicatorCell: ConversationViewCell {
     @objc
     public static let cellReuseIdentifier = "TypingIndicatorCell"
 
-        @available(*, unavailable, message:"use other constructor instead.")
+    @available(*, unavailable, message:"use other constructor instead.")
     @objc
     public required init(coder aDecoder: NSCoder) {
         notImplemented()
     }
 
+    private let kAvatarSize: CGFloat = 36
+    private let kAvatarHSpacing: CGFloat = 8
+
+    private let avatarView = AvatarImageView()
+    private let bubbleView = OWSBubbleView()
+    private let typingIndicatorView = TypingIndicatorView()
+    private var viewConstraints = [NSLayoutConstraint]()
+
     override init(frame: CGRect) {
         super.init(frame: frame)
+
+        commonInit()
+    }
+
+    private func commonInit() {
+        self.layoutMargins = .zero
+        self.contentView.layoutMargins = .zero
+
+        bubbleView.layoutMargins = .zero
+
+        bubbleView.addSubview(typingIndicatorView)
+        contentView.addSubview(bubbleView)
+
+        avatarView.autoSetDimension(.width, toSize: kAvatarSize)
+        avatarView.autoSetDimension(.height, toSize: kAvatarSize)
+    }
+
+    deinit {
+        NotificationCenter.default.removeObserver(self)
     }
 
     @objc
     public override func loadForDisplay() {
+        guard let conversationStyle = self.conversationStyle else {
+            owsFailDebug("Missing conversationStyle")
+            return
+        }
 
+        bubbleView.bubbleColor = conversationStyle.bubbleColor(isIncoming: true)
+        typingIndicatorView.startAnimation()
+        typingIndicatorView.addBackgroundView(withBackgroundColor: UIColor.red)
+
+        viewConstraints.append(contentsOf: [
+            bubbleView.autoPinEdge(toSuperviewEdge: .leading, withInset: conversationStyle.gutterLeading),
+            bubbleView.autoPinEdge(toSuperviewEdge: .trailing, withInset: conversationStyle.gutterTrailing, relation: .greaterThanOrEqual),
+            bubbleView.autoPinTopToSuperviewMargin(withInset: 0),
+            bubbleView.autoPinBottomToSuperviewMargin(withInset: 0),
+
+            typingIndicatorView.autoPinEdge(toSuperviewEdge: .leading, withInset: conversationStyle.textInsetHorizontal),
+            typingIndicatorView.autoPinEdge(toSuperviewEdge: .trailing, withInset: conversationStyle.textInsetHorizontal),
+            typingIndicatorView.autoPinTopToSuperviewMargin(withInset: conversationStyle.textInsetTop),
+            typingIndicatorView.autoPinBottomToSuperviewMargin(withInset: conversationStyle.textInsetBottom)
+            ])
+
+        if let avatarView = configureAvatarView() {
+            contentView.addSubview(avatarView)
+            viewConstraints.append(contentsOf: [
+                bubbleView.autoPinLeading(toTrailingEdgeOf: avatarView, offset: kAvatarHSpacing),
+                bubbleView.autoAlignAxis(.horizontal, toSameAxisOf: avatarView)
+                ])
+
+        } else {
+            avatarView.removeFromSuperview()
+        }
+    }
+
+    private func configureAvatarView() -> UIView? {
+        guard let viewItem = self.viewItem else {
+            owsFailDebug("Missing viewItem")
+            return nil
+        }
+        guard let typingIndicators = viewItem.interaction as? TypingIndicatorInteraction else {
+            owsFailDebug("Missing typingIndicators")
+            return nil
+        }
+        guard shouldShowAvatar() else {
+            return nil
+        }
+        guard let colorName = viewItem.authorConversationColorName else {
+            owsFailDebug("Missing authorConversationColorName")
+            return nil
+        }
+        guard let authorAvatarImage =
+            OWSContactAvatarBuilder(signalId: typingIndicators.recipientId,
+                                    colorName: ConversationColorNameForString(colorName),
+                                    diameter: UInt(kAvatarSize)).build() else {
+                                        owsFailDebug("Could build avatar image")
+                                        return nil
+        }
+        avatarView.image = authorAvatarImage
+        return avatarView
+    }
+
+    private func shouldShowAvatar() -> Bool {
+        guard let viewItem = self.viewItem else {
+            owsFailDebug("Missing viewItem")
+            return false
+        }
+        return viewItem.isGroupThread
     }
 
     @objc
     public override func cellSize() -> CGSize {
-        return .zero
+        guard let conversationStyle = self.conversationStyle else {
+            owsFailDebug("Missing conversationStyle")
+            return .zero
+        }
+
+        let insetsSize = CGSize(width: conversationStyle.textInsetHorizontal * 2,
+                                height: conversationStyle.textInsetTop + conversationStyle.textInsetBottom)
+        let typingIndicatorSize = typingIndicatorView.sizeThatFits(.zero)
+        let bubbleSize = CGSizeAdd(insetsSize, typingIndicatorSize)
+
+        if shouldShowAvatar() {
+            let avatarSize = CGSize(width: kAvatarSize, height: kAvatarSize)
+            return CGSizeCeil(CGSize(width: avatarSize.width + kAvatarHSpacing + bubbleSize.width,
+                                     height: max(avatarSize.height, bubbleSize.height)))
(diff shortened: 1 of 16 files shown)


Signal/src/ViewControllers/ConversationView/Cells/TypingIndicatorCell.swift
```diff
@@ -0,0 +1,154 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+@objc(OWSTypingIndicatorCell)
+public class TypingIndicatorCell: ConversationViewCell {
```
charlesmchen-signal: A simple cell that looks consistent with the message bubble cells, but is vastly simpler.  

The "real" message bubble cells are already way too complicated.  I didn't want to modify them to handle yet another special case.

Signal/src/ViewControllers/ConversationView/Cells/TypingIndicatorCell.swift
```diff
@@ -0,0 +1,154 @@
...
+    public static let cellReuseIdentifier = "TypingIndicatorCell"
+
+    @available(*, unavailable, message:"use other constructor instead.")
+    @objc
+    public required init(coder aDecoder: NSCoder) {
+        notImplemented()
+    }
+
+    private let kAvatarSize: CGFloat = 36
+    private let kAvatarHSpacing: CGFloat = 8
```
charlesmchen-signal: These two constants are use mimic the appearance of the "real" message bubbles.  We use a single constant, but I didn't bother.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -143,14 +143,24 @@ - (void)updateAuthorConversationColorNameWithTransaction:(YapDatabaseReadTransac
 {
     OWSAssertDebug(transaction);
 
-    if (self.interaction.interactionType != OWSInteractionType_IncomingMessage) {
-        _authorConversationColorName = nil;
-        return;
+    switch (self.interaction.interactionType) {
```
charlesmchen-signal: Adding a new "transient" cell type turned out to be incredibly easy.  I had to:

* Add a new interaction type.
* Add a new "stub" `TSInteraction` subclass.  This is regrettable, but it seemed simpler than modifying view items to only sometimes have a `TSInteraction`.
* The changes you see in this file.

I've very optimistic about converting unread indicators, date breaks, etc. to transient view items.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -574,6 +585,32 @@ - (void)uiDatabaseDidUpdate:(NSNotification *)notification
                updatedNeighborItemSet:updatedNeighborItemSet];
 }
 
+// A simpler version of the update logic we use when
+// only transient items have changed.
+- (void)updateForTransientItems
```
charlesmchen-signal: I'm not sure it's worth DRY'ing up the similar code between this and the method we use to respond to DB modification notifications.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -863,6 +900,25 @@ - (BOOL)reloadViewItems
             OWSAssertDebug(!viewItemCache[interaction.uniqueId]);
             viewItemCache[interaction.uniqueId] = viewItem;
         }
+
+        if (self.typingIndicatorsRecipient) {
+            id<ConversationViewItem> _Nullable lastViewItem = viewItems.lastObject;
+            uint64_t typingIndicatorTimestamp = (lastViewItem ? lastViewItem.interaction.timestamp + 1 : 1);
```
charlesmchen-signal: The typing indicator use a fake timestamp after the last interaction in the conversation - or 1 if none exists.

Signal/src/ViewControllers/ConversationView/TypingIndicatorInteraction.swift
```diff
@@ -0,0 +1,44 @@
+//
+//  Copyright (c) 2018 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+@objc(OWSTypingIndicatorInteraction)
+public class TypingIndicatorInteraction: TSInteraction {
+    @objc
+    public static let TypingIndicatorId = "TypingIndicator"
```
charlesmchen-signal: This is the "stub" `TSInteraction".  Note that it uses a fixed unique id.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
```diff
@@ -111,15 +118,25 @@ - (void)commontInit
...
+    self.typingIndicatorView = [TypingIndicatorView new];
+
+    self.previewStackView = [[UIStackView alloc] initWithArrangedSubviews:@[
         self.snippetLabel,
+        self.typingIndicatorView,
+    ]];
+    self.previewStackView.axis = UILayoutConstraintAxisVertical;
+    self.previewStackView.alignment = UIStackViewAlignmentLeading;
+    [self.previewStackView setContentHuggingHorizontalLow];
+    [self.previewStackView setCompressionResistanceHorizontalLow];
```
charlesmchen-signal: The typing indicator replaces the "snippet" we use to preview the conversation contents.  The only tricky thing was ensuring that the cell had a consistent layout between these two cases.

Signal/src/views/TypingIndicatorView.swift
```diff
@@ -0,0 +1,180 @@
...
+                let margin = (TypingIndicatorView.kMaxRadiusPt - radius) * 0.5
+                let bezierPath = UIBezierPath(ovalIn: CGRect(x: margin, y: margin, width: radius, height: radius))
+                pathValues.append(bezierPath.cgPath)
+
+                keyTimes.append(keyFrameTime)
+                animationDuration = max(animationDuration, keyFrameTime)
+            }
+
+            // All animations in the group apparently need to have the same number
+            // of keyframes, and use the same timing.
```
charlesmchen-signal: This deliberately corresponds to how the animation is spec'd.

SignalServiceKit/src/Contacts/TSThread.h
```diff
@@ -11,6 +11,9 @@ NS_ASSUME_NONNULL_BEGIN
 @class TSInvalidIdentityKeyReceivingErrorMessage;
 
 typedef NSString *ConversationColorName NS_STRING_ENUM;
+
+ConversationColorName ConversationColorNameForString(NSString *value);
```
charlesmchen-signal: Swift can't convert `ConversationColorName` to `String` and back, so I made this convenience method.

SignalServiceKit/src/Util/TypingIndicators.swift
```diff
@@ -310,27 +332,32 @@ public class TypingIndicatorsImpl: NSObject, TypingIndicators {
 
     // MARK: -
 
-    // Map of (thread id and recipient id)-to-(device id)-to-IncomingIndicators.
-    private var incomingIndicatorsMap = [String: [UInt: IncomingIndicators]]()
+    // Map of (thread id)-to-(recipient id and device id)-to-IncomingIndicators.
+    private var incomingIndicatorsMap = [String: [String: IncomingIndicators]]()
```
charlesmchen-signal: There's a bit of complexity here.  We want to track the typing state for each (thread, sender, device), but we want to query that in this was: "which recipient (if any) is typing in thread X?  If more than one recipient is typing, give me the one that started typing first".  

Therefore, I used a map-of-maps.

SignalServiceKit/src/Util/TypingIndicators.swift
```diff
@@ -375,43 +407,37 @@ public class TypingIndicatorsImpl: NSObject, TypingIndicators {
...
         }
 
         func didReceiveIncomingMessage() {
             AssertIsOnMainThread()
 
-            // If the client receives a message:
-            //
-            // Cancel the displayTyping timer for that (sender, device)
-            // Hide the typing indicator for that (sender, device)
+            clearTyping()
```
charlesmchen-signal: I DRY'd up this logic.

Signal/src/ViewControllers/ConversationView/Cells/TypingIndicatorCell.swift
michaelkirk-signal: :+1:

Signal/src/ViewControllers/ConversationView/Cells/TypingIndicatorCell.swift
```diff
@@ -0,0 +1,154 @@
...
+
+        NSLayoutConstraint.deactivate(viewConstraints)
+        viewConstraints = [NSLayoutConstraint]()
+
+        avatarView.image = nil
+        avatarView.removeFromSuperview()
+
+        typingIndicatorView.stopAnimation()
+
+        NotificationCenter.default.removeObserver(self)
```
michaelkirk-signal: What notifications are we no longer observing here?

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
michaelkirk-signal: 🎉

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -143,14 +143,24 @@ - (void)updateAuthorConversationColorNameWithTransaction:(YapDatabaseReadTransac
...
+        case OWSInteractionType_TypingIndicator: {
+            OWSTypingIndicatorInteraction *typingIndicator = (OWSTypingIndicatorInteraction *)self.interaction;
+            _authorConversationColorName =
+                [TSContactThread conversationColorNameForRecipientId:typingIndicator.recipientId
```
michaelkirk-signal: Hmm... why do we need ` _authorConversationColorName` for a typing indicator?

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -1276,6 +1333,28 @@ - (nullable NSNumber *)findGroupIndexOfThreadInteraction:(TSInteraction *)intera
...
 }
 
+- (void)typingIndicatorStateDidChange:(NSNotification *)notification
+{
+    OWSAssertIsOnMainThread();
+
+    self.typingIndicatorsRecipient = [self.typingIndicators typingIndicatorsForThread:self.thread];
+}
+
+- (void)setTypingIndicatorsRecipient:(nullable NSString *)typingIndicatorsRecipient
```
michaelkirk-signal: recipient->recipientId? 

No biggied, but it's kind of confusing terminology, because it's the *sender* of the typing message, not the recipient. The local user is actually the recipient, right?

Signal/src/ViewControllers/ConversationView/TypingIndicatorInteraction.swift
michaelkirk-signal: Should we do more to make sure this is never saved? e.g. override saveWithTransaction?

Signal/src/ViewControllers/HomeView/HomeViewCell.m
michaelkirk-signal: stacks on stacks on stacks on stacks. Let's make sure we verify this layout on legacy iOS versions as part of the test plan, since iOS9/10 can behave somewhat differently with complex stack views.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
```diff
@@ -111,15 +118,25 @@ - (void)commontInit
...
+
+    UIStackView *bottomRowView = [[UIStackView alloc] initWithArrangedSubviews:@[
+        self.previewStackView,
         self.messageStatusView,
     ]];
+    
     bottomRowView.axis = UILayoutConstraintAxisHorizontal;
-    bottomRowView.alignment = UIStackViewAlignmentLastBaseline;
+    bottomRowView.alignment = UIStackViewAlignmentCenter;
```
michaelkirk-signal: I know it's subtle point but we were intentionally aligning on baseline. It's a couple px different. 

To keep the existing behavior you might have to do something like add a [viewForLastBaseLineLayout](https://developer.apple.com/documentation/uikit/uiview/1622633-viewforlastbaselinelayout?language=objc) to the typingIndicatorView.

SignalServiceKit/src/Util/TypingIndicators.swift
```diff
@@ -428,8 +454,11 @@ public class TypingIndicatorsImpl: NSObject, TypingIndicators {
             guard delegate.areTypingIndicatorsEnabled() else {
                 return
             }
-
-            NotificationCenter.default.postNotificationNameAsync(TypingIndicatorsImpl.typingIndicatorStateDidChange, object: recipientId)
+            guard let threadId = thread.uniqueId else {
+                owsFailDebug("Thread is missing id.")
+                return
+            }
+            NotificationCenter.default.postNotificationNameAsync(TypingIndicatorsImpl.typingIndicatorStateDidChange, object: threadId)
```
michaelkirk-signal: Huh - isn't `object` observation scoping based on reference (not value) equality?

SignalServiceKit/src/Contacts/TSThread.h
michaelkirk-signal: Huh... what happens when you use `ConversationColorName(rawValue: colorName)` and (eg) `ConversationColorNameCrimson.rawValue`?

SignalServiceKit/src/Util/TypingIndicators.swift
```diff
@@ -24,9 +24,14 @@ public protocol TypingIndicators: class {
...
 
+    // Returns the recipient id of the user who should currently be shown typing for a given thread.
+    //
+    // If no one is typing in that thread, returns nil.
+    // If multiple users are typing in that thread, returns the user to show.
+    //
     // TODO: Use this method.
     @objc
-    func areTypingIndicatorsVisible(inThread thread: TSThread, recipientId: String) -> Bool
+    func typingIndicators(forThread thread: TSThread) -> String?
```
michaelkirk-signal: maybe: `typingRecipientId(forThread:) -> String` ?

SignalServiceKit/src/Util/TypingIndicators.swift
michaelkirk-signal: TIOLI but typealiases might make this clearer:

```
typealias ThreadId = String
typealias ReceipientDeviceKey = String
private var incomingIndicatorsMap = [ThreadId: [ReceipientDeviceKey: IncomingIndicators]]()

```

Signal/src/ViewControllers/ConversationView/TypingIndicatorInteraction.swift
charlesmchen-signal: Good idea.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
charlesmchen-signal: Noted.

SignalServiceKit/src/Contacts/TSThread.h
charlesmchen-signal: Oh nice, thanks.

Signal/src/ViewControllers/ConversationView/Cells/TypingIndicatorCell.swift
charlesmchen-signal: This is obsolete, removed.

Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
charlesmchen-signal: Because in a group, we show the avatar of the "typing" user next to the typing indicator and we want to honor their conversation color in the default avatar icon.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
charlesmchen-signal: Done.  We've mostly standardized around "recipient id" for user/account ids in the codebase, so I was trying to be consistent with that.

SignalServiceKit/src/Util/TypingIndicators.swift
charlesmchen-signal: Done.

Signal/src/ViewControllers/HomeView/HomeViewCell.m
charlesmchen-signal: Okay.  I ended up reworking this entirely to hide the snippet label by changing its text.  IMO that'll be simplest.

PR #3981, opened 2018-11-02
Charlesmchen/debug ui multi image sends

charlesmchen-signal:
Fixes the "happy path" for multi-image paths and adds Debug UI to exercise it.

* Multiple attachments are uploaded & downloaded.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
@@ -481,6 +481,7 @@ NSString *NSStringForOWSMessageCellType(OWSMessageCellType cellType)
     if (message.attachmentIds.count == 0) {
         return nil;
     }
+    // TODO: Support multi-image messages.
     NSString *_Nullable attachmentId = message.attachmentIds.firstObject;
     if (attachmentId.length == 0) {
         return nil;
--- a/SignalMessaging/utils/ThreadUtil.m
+++ b/SignalMessaging/utils/ThreadUtil.m
@@ -93,35 +93,36 @@ NS_ASSUME_NONNULL_BEGIN
                                            inThread:(TSThread *)thread
                                    quotedReplyModel:(nullable OWSQuotedReplyModel *)quotedReplyModel
 {
-    return [self enqueueMessageWithAttachment:attachment
-                                     inThread:thread
-                             quotedReplyModel:quotedReplyModel
-                                 ignoreErrors:NO];
+    return [self enqueueMessageWithAttachment:attachment inThread:thread quotedReplyModel:quotedReplyModel];
 }
 
-+ (TSOutgoingMessage *)enqueueMessageWithAttachment:(SignalAttachment *)attachment
-                                           inThread:(TSThread *)thread
-                                   quotedReplyModel:(nullable OWSQuotedReplyModel *)quotedReplyModel
-                                       ignoreErrors:(BOOL)ignoreErrors
++ (TSOutgoingMessage *)enqueueMessageWithAttachments:(NSArray<SignalAttachment *> *)attachments
+                                            inThread:(TSThread *)thread
+                                    quotedReplyModel:(nullable OWSQuotedReplyModel *)quotedReplyModel
 {
     OWSAssertIsOnMainThread();
-    OWSAssertDebug(attachment);
-    OWSAssertDebug(!attachment.hasError);
-    OWSAssertDebug(attachment.mimeType.length > 0);
+    OWSAssertDebug(attachments.count > 0);
     OWSAssertDebug(thread);
+    for (SignalAttachment *attachment in attachments) {
+        OWSAssertDebug(!attachment.hasError);
+        OWSAssertDebug(attachment.mimeType.length > 0);
+    }
 
     OWSDisappearingMessagesConfiguration *configuration =
         [OWSDisappearingMessagesConfiguration fetchObjectWithUniqueID:thread.uniqueId];
 
     uint32_t expiresInSeconds = (configuration.isEnabled ? configuration.durationSeconds : 0);
+    BOOL isVoiceMessage = (attachments.count == 1 && attachments.lastObject.isVoiceMessage);
+    // TODO: Support multi-image captions.
+    NSString *_Nullable messageBody = attachments.lastObject.captionText;
     TSOutgoingMessage *message =
         [[TSOutgoingMessage alloc] initOutgoingMessageWithTimestamp:[NSDate ows_millisecondTimeStamp]
                                                            inThread:thread
-                                                        messageBody:attachment.captionText
+                                                        messageBody:messageBody
                                                       attachmentIds:[NSMutableArray new]
                                                    expiresInSeconds:expiresInSeconds
                                                     expireStartedAt:0
-                                                     isVoiceMessage:[attachment isVoiceMessage]
+                                                     isVoiceMessage:isVoiceMessage
                                                    groupMetaMessage:TSGroupMetaMessageUnspecified
                                                       quotedMessage:[quotedReplyModel buildQuotedMessageForSending]
                                                        contactShare:nil];
--- a/SignalMessaging/utils/ThreadUtil.m
+++ b/SignalMessaging/utils/ThreadUtil.m
@@ -127,11 +127,11 @@ NS_ASSUME_NONNULL_BEGIN
                                                       quotedMessage:[quotedReplyModel buildQuotedMessageForSending]
                                                        contactShare:nil];
 
-    [self.messageSenderJobQueue addMediaMessage:message
-                                     dataSource:attachment.dataSource
-                                    contentType:attachment.mimeType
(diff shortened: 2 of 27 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewItem.m
```diff
@@ -481,6 +481,7 @@ - (nullable TSAttachment *)firstAttachmentIfAnyOfMessage:(TSMessage *)message
     if (message.attachmentIds.count == 0) {
         return nil;
     }
+    // TODO: Support multi-image messages.
```
charlesmchen-signal: There's a bunch of these TODOs identifying possible follow-up work.  I tried to make them identical to facilitate pursuing them.

SignalMessaging/utils/ThreadUtil.m
```diff
@@ -93,44 +93,45 @@ + (TSOutgoingMessage *)enqueueMessageWithAttachment:(SignalAttachment *)attachme
...
+        OWSAssertDebug(!attachment.hasError);
+        OWSAssertDebug(attachment.mimeType.length > 0);
+    }
 
     OWSDisappearingMessagesConfiguration *configuration =
         [OWSDisappearingMessagesConfiguration fetchObjectWithUniqueID:thread.uniqueId];
 
     uint32_t expiresInSeconds = (configuration.isEnabled ? configuration.durationSeconds : 0);
+    BOOL isVoiceMessage = (attachments.count == 1 && attachments.lastObject.isVoiceMessage);
+    // TODO: Support multi-image captions.
```
charlesmchen-signal: This PR doesn't include any proto schema changes.

SignalServiceKit/src/Messages/Interactions/TSQuotedMessage.m
```diff
@@ -253,7 +253,8 @@ + (nullable TSAttachmentStream *)tryToDeriveLocalThumbnailWithAttachmentInfo:(OW
         return nil;
     }
 
-    TSAttachment *attachment = [quotedMessage attachmentWithTransaction:transaction];
+    // We quote _the first_ attachment, if any.
```
charlesmchen-signal: I've got a checklist of open Qs around this work.  This is one.

SignalServiceKit/src/Messages/OWSMessageSender.h
```diff
@@ -33,6 +33,23 @@ typedef void (^RetryableFailureHandler)(NSError *_Nonnull error);
 
 #pragma mark -
 
+NS_SWIFT_NAME(OutgoingAttachmentInfo)
+@interface OWSOutgoingAttachmentInfo : NSObject
```
charlesmchen-signal: Unlike `SignalAttachment` (which I'd love to rename to `OutgoingAttachment`), this new class:

* Is a super lightweight struct with no logic.
* Resides in SSK (`SignalAttachment` resides in SM).
* Is MIME-type/`contentType`-oriented, not UTI-oriented.

This class just simplified the APIs as we pass around outgoing attachment metadata.

SignalServiceKit/src/Storage/OWSMediaGalleryFinder.m
```diff
@@ -147,17 +147,20 @@ + (YapDatabaseAutoView *)mediaGalleryDatabaseExtension
             return nil;
         }
         TSMessage *message = (TSMessage *)object;
-        
-        OWSAssertDebug(message.attachmentIds.count <= 1);
-        NSString *attachmentId = message.attachmentIds.firstObject;
-        if (attachmentId.length == 0) {
-            return nil;
+
+        BOOL allAttachmentsAreMedia = message.attachmentIds.count > 0;
```
charlesmchen-signal: I did the simplest possible changes to get the media views working; there's more follow-up work.  Right now:

* We only treat messages as media messages if _all_ of their attachments are media (should be true).
* I think the media views will only use the first attachment from multi-image messages, but I haven't tested it yet.

SignalMessaging/utils/ThreadUtil.m
```diff
@@ -93,44 +93,45 @@ + (TSOutgoingMessage *)enqueueMessageWithAttachment:(SignalAttachment *)attachme
                                            inThread:(TSThread *)thread
                                    quotedReplyModel:(nullable OWSQuotedReplyModel *)quotedReplyModel
 {
-    return [self enqueueMessageWithAttachment:attachment
-                                     inThread:thread
-                             quotedReplyModel:quotedReplyModel
-                                 ignoreErrors:NO];
+    return [self enqueueMessageWithAttachment:attachment inThread:thread quotedReplyModel:quotedReplyModel];
```
michaelkirk-signal: This looks like infinite recursion. 

I was expecting something like: `enqueueMessageWithAttachment**s**:@[ attachment ] ...`

SignalMessaging/utils/ThreadUtil.m
```diff
@@ -93,44 +93,45 @@ + (TSOutgoingMessage *)enqueueMessageWithAttachment:(SignalAttachment *)attachme
...
         [OWSDisappearingMessagesConfiguration fetchObjectWithUniqueID:thread.uniqueId];
 
     uint32_t expiresInSeconds = (configuration.isEnabled ? configuration.durationSeconds : 0);
+    BOOL isVoiceMessage = (attachments.count == 1 && attachments.lastObject.isVoiceMessage);
+    // TODO: Support multi-image captions.
+    NSString *_Nullable messageBody = attachments.lastObject.captionText;
     TSOutgoingMessage *message =
         [[TSOutgoingMessage alloc] initOutgoingMessageWithTimestamp:[NSDate ows_millisecondTimeStamp]
                                                            inThread:thread
-                                                        messageBody:attachment.captionText
```
michaelkirk-signal: So presumably the caption-text will eventually be a new field on the TSAttachment?

SignalServiceKit/src/Messages/Attachments/OWSAttachmentsProcessor.h
```diff
@@ -41,11 +41,11 @@ extern NSString *const kAttachmentDownloadAttachmentIDKey;
 
 - (void)fetchAttachmentsForMessage:(nullable TSMessage *)message
                     primaryStorage:(OWSPrimaryStorage *)primaryStorage
-                           success:(void (^)(TSAttachmentStream *attachmentStream))successHandler
```
michaelkirk-signal: As I understand it, despite this "aggregate" completion handler, we're still saving each attachment and `touch`ing it's message with each individual downloaded attachment, meaning we'll still be able to update the UI incrementally, rather than having to wait for the entire album to be downloaded. :+1:

SignalServiceKit/src/Messages/OWSMessageSender.h
michaelkirk-signal: This seems very similar to `OWSAttachmentInfo`, which so far was only used for `TSQuotedMessage`.

The difference I guess is that OWSAttachmentInfo is intended to be serialized into a YapDB object, whereas OWSOutgoingAttachmentInfo is an ephemeral object, maintaining a reference to a DataSource.

Even though they're very similar, I **think** it makes sense to have both.

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
```diff
@@ -204,29 +204,7 @@ public class FullTextSearchFinder: NSObject {
...
-            return nil
-        }
-
-        guard let text = attachmentStream.readOversizeText() else {
-            owsFailDebug("Could not load oversize text attachment")
-            return nil
-        }
-
-        return text
+        return message.oversizeText(with: transaction)
```
michaelkirk-signal: TIOLI: Now that you've deduped this logic, can we just inline this method call?

SignalServiceKit/src/Messages/OWSMessageSender.h
charlesmchen-signal: > Even though they're very similar, I think it makes sense to have both.

Good point.  Yeah, I think it makes sense to have both.

SignalMessaging/utils/ThreadUtil.m
charlesmchen-signal: Done.

SignalMessaging/utils/ThreadUtil.m
charlesmchen-signal: It sounded like that might happen.

SignalServiceKit/src/Messages/Attachments/OWSAttachmentsProcessor.h
charlesmchen-signal: That's the idea.  We can revisit as we implement the view changes.

SignalServiceKit/src/Storage/FullTextSearchFinder.swift
charlesmchen-signal: Done and I actually found a way to simplify this even further.

PR #3982, opened 2018-11-02
Charlesmchen/app will become ready 

charlesmchen-signal:
I've wanted to do something like this for a while, but I'm doing it now to resolve a real race on startup around the new durable message queues.

The proposal is to avoid races on startup by distinguishing "will become ready" (which doesn't depend on any other components except primary storage) and "did become ready" (which might use other components) work.  

After the "will" blocks, all of the components in the system should be ready to interact with other components - they can ensure that with a "will" block.  

```
// If the app is ready, the block is called immediately;
// otherwise it is called when the app becomes ready.
//
// This method should only be called on the main thread.
// The block will always be called on the main thread.
//
// * The "will become ready" blocks are called before the "did become ready" blocks.
// * The "will become ready" blocks should be used for internal setup of components
//   so that they are ready to interact with other components of the system.
// * The "did become ready" blocks should be used for any work that should be done
//   on app launch, especially work that uses other components.
// * We should usually use "did become ready" blocks since they are safer.
+ (void)runNowOrWhenAppWillBecomeReady:(AppReadyBlock)block NS_SWIFT_NAME(runNowOrWhenAppWillBecomeReady(_:));
+ (void)runNowOrWhenAppDidBecomeReady:(AppReadyBlock)block NS_SWIFT_NAME(runNowOrWhenAppDidBecomeReady(_:));
```

It also lets us gather concerns.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -1130,17 +1130,9 @@ static NSTimeInterval launchStartedAt;
 
     [AppVersion.sharedInstance mainAppLaunchDidComplete];
 
-    [Environment.shared.contactsManager setup];
-    [Environment.shared.contactsManager startObserving];
-
-    // If there were any messages in our local queue which we hadn't yet processed.
-    [SSKEnvironment.shared.messageReceiver handleAnyUnprocessedEnvelopesAsync];
-    [SSKEnvironment.shared.batchMessageProcessor handleAnyUnprocessedEnvelopesAsync];
     [Environment.shared.audioSession setup];
 
     [SSKEnvironment.shared.reachabilityManager setup];
-    [SSKEnvironment.shared.messageSenderJobQueue setup];
-    [AppEnvironment.shared.sessionResetJobQueue setup];
 
     if (!Environment.shared.preferences.hasGeneratedThumbnails) {
         [self.primaryStorage.newDatabaseConnection
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -550,7 +550,7 @@ static NSTimeInterval launchStartedAt;
 
     if (!AppReadiness.isAppReady) {
         OWSLogWarn(@"Ignoring openURL: app not ready.");
-        // We don't need to use [AppReadiness runNowOrWhenAppIsReady:];
+        // We don't need to use [AppReadiness runNowOrWhenAppDidBecomeReady:];
         // the only URLs we handle in Signal iOS at the moment are used
         // for resuming the verification step of the registration flow.
         return NO;
@@ -596,7 +596,7 @@ static NSTimeInterval launchStartedAt;
 
     [self ensureRootViewController];
 
-    [AppReadiness runNowOrWhenAppIsReady:^{
+    [AppReadiness runNowOrWhenAppDidBecomeReady:^{
         [self handleActivation];
     }];
 
@@ -616,7 +616,7 @@ static NSTimeInterval launchStartedAt;
 
 - (void)enableBackgroundRefreshIfNecessary
 {
-    [AppReadiness runNowOrWhenAppIsReady:^{
+    [AppReadiness runNowOrWhenAppDidBecomeReady:^{
         if (OWS2FAManager.sharedManager.is2FAEnabled && [self.tsAccountManager isRegistered]) {
             // Ping server once a day to keep-alive 2FA clients.
             const NSTimeInterval kBackgroundRefreshInterval = 24 * 60 * 60;
@@ -732,7 +732,7 @@ static NSTimeInterval launchStartedAt;
     OWSAssertIsOnMainThread();
 
     [SignalApp clearAllNotifications];
-    [AppReadiness runNowOrWhenAppIsReady:^{
+    [AppReadiness runNowOrWhenAppDidBecomeReady:^{
         [OWSMessageUtils.sharedManager updateApplicationBadgeCount];
     }];
 }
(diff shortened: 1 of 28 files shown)


Signal/src/AppDelegate.m
```diff
@@ -596,7 +596,7 @@ - (void)applicationDidBecomeActive:(UIApplication *)application {
 
     [self ensureRootViewController];
 
-    [AppReadiness runNowOrWhenAppIsReady:^{
+    [AppReadiness runNowOrWhenAppDidBecomeReady:^{
```
charlesmchen-signal: Almost all of our "on app ready" blocks become "_did_ become ready" blocks.

Signal/src/AppDelegate.m
```diff
@@ -1130,17 +1130,9 @@ - (void)checkIfAppIsReady
 
     [AppVersion.sharedInstance mainAppLaunchDidComplete];
 
-    [Environment.shared.contactsManager setup];
```
charlesmchen-signal: These deletions are key changes - each of these components has taken over responsibility for "queuing up" their own setup methods.

Signal/src/AppDelegate.m
```diff
@@ -1130,17 +1130,9 @@ - (void)checkIfAppIsReady
...
-    [Environment.shared.contactsManager startObserving];
-
-    // If there were any messages in our local queue which we hadn't yet processed.
-    [SSKEnvironment.shared.messageReceiver handleAnyUnprocessedEnvelopesAsync];
-    [SSKEnvironment.shared.batchMessageProcessor handleAnyUnprocessedEnvelopesAsync];
     [Environment.shared.audioSession setup];
 
     [SSKEnvironment.shared.reachabilityManager setup];
-    [SSKEnvironment.shared.messageSenderJobQueue setup];
-    [AppEnvironment.shared.sessionResetJobQueue setup];
```
charlesmchen-signal: I think we were neglecting to `setup` up these last two queues in the SAE.  I think that was a bug on `messageSenderJobQueue`.

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
```diff
@@ -458,6 +458,12 @@ - (instancetype)initWithPrimaryStorage:(OWSPrimaryStorage *)primaryStorage
 
     _processingQueue = processingQueue;
 
+    [AppReadiness runNowOrWhenAppDidBecomeReady:^{
+        if (CurrentAppContext().isMainApp) {
```
charlesmchen-signal: Some components (like this one) only need to do their "on startup" work in the main app.

SignalServiceKit/src/Util/JobQueue.swift
```diff
@@ -109,7 +109,11 @@ public extension JobQueue {
         jobRecord.save(with: transaction)
 
         transaction.addCompletionQueue(.global()) {
-            self.workStep()
+            AppReadiness.runNowOrWhenAppDidBecomeReady {
```
charlesmchen-signal: Being a bit more defensive here.

charlesmchen-signal:
> One concern I have is that it is now more difficult to reason about the order in which the setup blocks run, but I guess the implied design constraints are:

Exactly right.  This approach _is_ limited, but I think it'll be beneficial to remain within those limits since they force us to keep our code decoupled.  In the big picture, our dependencies should look like this:

* Bootstrapping (setting up logging, AppVersions, etc.)
* Setting up the environments.
* Waiting for primary storage to be ready.
* Migrations
* All modules "inner" setup (no use of other modules allowed).
* All modules "on startup logic" (and, simultaneously...)
* Present UI.

SignalServiceKit/src/Messages/OWSBatchMessageProcessor.m
charlesmchen-signal: Good point. Made the same change in `OWSMessageReceiver`.

SignalServiceKit/src/Util/JobQueue.swift
charlesmchen-signal: But this is `JobQueue`, not `MessageSenderJobQueue`.  I could see us queuing jobs before the app is ready, e.g. I could see a "migration" enqueuing a "durable account attributes update", for example.

I'm adding an assert `MessageSenderJobQueue`, but I could even see a migration enqueuing a sync message? Maybe?

SignalMessaging/contacts/OWSContactsManager.m
```diff
@@ -82,6 +82,13 @@ - (id)initWithPrimaryStorage:(OWSPrimaryStorage *)primaryStorage
 
     OWSSingletonAssert();
 
+    [AppReadiness runNowOrWhenAppWillBecomeReady:^{
+        [self setup];
+    }];
+    [AppReadiness runNowOrWhenAppDidBecomeReady:^{
```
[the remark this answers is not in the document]
charlesmchen-signal: You're right - it could race.  I've modified the contact manager to start observing on _will_ but to respond to those notifications using _did_.

SignalServiceKit/src/Messages/OWSMessageReceiver.m
```diff
@@ -248,7 +248,7 @@ - (instancetype)initWithDBConnection:(YapDatabaseConnection *)dbConnection finde
     _finder = finder;
     _isDrainingQueue = NO;
 
-    [AppReadiness runNowOrWhenAppIsReady:^{
+    [AppReadiness runNowOrWhenAppDidBecomeReady:^{
```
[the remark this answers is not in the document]
charlesmchen-signal: Done.

PR #3983, opened 2018-11-02
Fix race in CVM startup.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
@@ -211,18 +211,6 @@ static const int kYapDatabaseRangeMinLength = 0;
 - (void)addNotificationListeners
 {
     [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(uiDatabaseDidUpdateExternally:)
-                                                 name:OWSUIDatabaseConnectionDidUpdateExternallyNotification
-                                               object:self.primaryStorage.dbNotificationObject];
-    [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(uiDatabaseWillUpdate:)
-                                                 name:OWSUIDatabaseConnectionWillUpdateNotification
-                                               object:self.primaryStorage.dbNotificationObject];
-    [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(uiDatabaseDidUpdate:)
-                                                 name:OWSUIDatabaseConnectionDidUpdateNotification
-                                               object:self.primaryStorage.dbNotificationObject];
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -259,6 +247,19 @@ - (void)configure
     if (![self reloadViewItems]) {
         OWSFailDebug(@"failed to reload view items in configureForThread.");
     }
+
+    [[NSNotificationCenter defaultCenter] addObserver:self
```
charlesmchen-signal: Unlike the old CVC, the CVM builds a mapping immediately and continually updates it to reflect all db modifications.  Consequently, we need to start observing DB modifications notification immediately.  Previous to this PR we were vulnerable to a race wherein a DB modification occurred between the view/VM being configured and the view loading.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -458,19 +459,24 @@ - (void)uiDatabaseDidUpdateExternally:(NSNotification *)notification
...
+    // We don't need to do this if we're not observing db modifications since we'll
+    // do it when we resume.
+    [self resetMappings];
 }
 
 - (void)uiDatabaseWillUpdate:(NSNotification *)notification
 {
+    if (!self.delegate.isObservingVMUpdates) {
+        return;
+    }
```
charlesmchen-signal: As an optimization, we don't need to fire will/did update events to the delegate if the view isn't observing VM updates.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
```diff
@@ -618,6 +624,13 @@ - (void)updateViewWithOldItemIdList:(NSArray<NSString *> *)oldItemIdList
     OWSAssertDebug(oldItemIdList);
     OWSAssertDebug(updatedItemSet);
 
+    if (!self.delegate.isObservingVMUpdates) {
+        OWSFailDebug(@"Skipping VM update.");
+        // We fire this event, but it will be ignored.
+        [self.delegate conversationViewModelDidUpdate:ConversationUpdate.minorUpdate];
```
charlesmchen-signal: Calculating the incremental VM update diffs is non-trival, we can skip this work.

Signal/src/ViewControllers/ConversationView/ConversationViewModel.m
charlesmchen-signal: I was thinking about this over the weekend and I think this race might have existed in the pre-CVM CVC.  

There's a number of ways in which these changes _might_ significantly reduce our crashes in production.

PR #3987, opened 2018-11-05
Custom photo picker, respects theme/call banner

michaelkirk-signal:
Custom photo picker, respects theme/call banner
    
- share GridViewCell
- multi-select. Since the approval/send flow doesn't yet support multiple attachments, this is behind a feature flag.

multiselect

PTAL @charlesmchen-signal


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2677,6 +2678,14 @@ - (void)resetFrame
     self.view.frame = frame;
 }
 
+- (void)imagePicker:(OWSImagePickerGridController *)imagePicker
+    didPickImageAttachments:(NSArray<SignalAttachment *> *)attachments
+{
+    // TODO support approving multiple attachments.
+    SignalAttachment *firstAttachment = attachments.firstObject;
+    [self tryToSendAttachmentIfApproved:firstAttachment];
```
charlesmchen-signal: Are we going to use one UI/VC for "approving" media galleries (with multiple captions, etc.), and other messages (e.g. mp3s, pdfs, single image, single video, etc.)?  Or will we separate VCs?

michaelkirk-signal:
I haven’t seen the final design spec yet, but my understanding is that, from an interface perspective:

We will be reusing the existing captioning interface once for each item. When only one item is picked, approval will be identical to the current experience. When multiple items have been picked, there will be a series of these captioning controllers organized in a page view controller, which can be swiped through, with a visible photo rail if you want to skip around.

So I’m assuming a single coordinating controller which has the list of attachments, and builds a captioning controller for each picked media item page.

Signal/src/ViewControllers/ImagePickerController.swift
```diff
@@ -0,0 +1,439 @@
...
+    @objc
+    func didTapSelect(_ sender: Any) {
+        isInBatchSelectMode = true
+
+        // disabled until at least one item is selected
+        self.doneButton.isEnabled = false
+
+        // Don't allow the user to leave mid-selection, so they realized they have
+        // to cancel (lose) their selection if they leave.
+        self.navigationItem.hidesBackButton = true
```
charlesmchen-signal: So if a user selects a photo, they can't hit back until they unselect that photo? That doesn't sound right, but I think I'm misunderstanding.

Signal/src/ViewControllers/ImagePickerController.swift
```diff
@@ -0,0 +1,439 @@
...
+        super.init()
+        PHPhotoLibrary.shared().register(self)
+    }
+
+    deinit {
+        PHPhotoLibrary.shared().unregisterChangeObserver(self)
+    }
+
+    func albumForAllPhotos() -> PhotoLibraryAlbum {
+        let allPhotosOptions = PHFetchOptions()
```
charlesmchen-signal: I suspect we'll want to only load a "window" into the album and raise the load window size to reflect scroll state. Otherwise, won't we be loading all assets into memory for users with large libraries?

Signal/src/ViewControllers/MediaTileViewController.swift
```diff
@@ -68,7 +68,7 @@ public class MediaTileViewController: UICollectionViewController, MediaGalleryDa
 
         collectionView.backgroundColor = Theme.backgroundColor
 
-        collectionView.register(MediaGalleryCell.self, forCellWithReuseIdentifier: MediaGalleryCell.reuseIdentifier)
+        collectionView.register(PhotoGridViewCell.self, forCellWithReuseIdentifier: PhotoGridViewCell.reuseIdentifier)
```
charlesmchen-signal: Media cells can represent images, videos and animated images, right?

charlesmchen-signal: My point is - is photo the best name?

Signal/src/ViewControllers/ImagePickerController.swift
michaelkirk-signal: Sorry, this comment is confusing because it was lifted from the MediaTileViewController. 

It's currently irrelevant anyway, because there is no back button, only a dismiss button, which isn't impeded by this behavior, but I'm going to remove it.

Signal/src/ViewControllers/MediaTileViewController.swift
michaelkirk-signal: I personally don't think it's surprising that the **Photo**Grid includes videos and animated gifs. e.g. see the **Photo**s app.

Signal/src/ViewControllers/ImagePickerController.swift
michaelkirk-signal: It does look reckless, but my understanding is that this is the correct way to load even very large collections, and that the Photo's framework does the work to make sure this performs well:

[excerpt from Photos framework](https://developer.apple.com/documentation/photokit/phfetchresult)

> You access the contents of a fetch result with the same methods and conventions used by the NSArray class. Unlike an NSArray object, however, a PHFetchResult object dynamically loads its contents from the Photos library as needed, providing optimal performance even when handling a large number of results.

michaelkirk-signal: Furthermore, from what I can tell there is no way to do a "load window" or pagination even if we wanted to.

PR #3994, opened 2018-11-07
multi media send approval page view controller

michaelkirk-signal:
PTAL @charlesmchen-signal 

Continuing the multisend work, this is a page view controller for the captioning flow.

Here's a screen shot mid-swipe.

image

The caption field updates as you pan to reflect the caption for *that* Attachment.

Conspicuously missing are:
1. photo rail at the bottom of the picker view for summary and quick navigation
2. integration with the SAE

I'm tracking both of those items separately, but I think they'll be largely additive changes.


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2770,22 +2768,21 @@ - (void)imagePickerController:(UIImagePickerController *)picker
...
-    // TODO: Should we assume non-nil or should we check for non-nil?
-    OWSAssertDebug(attachment != nil);
-    OWSAssertDebug(![attachment hasError]);
-    OWSAssertDebug([attachment mimeType].length > 0);
-
-    OWSLogVerbose(@"Sending attachment. Size in bytes: %lu, contentType: %@",
-        (unsigned long)[attachment dataLength],
-        [attachment mimeType]);
+    for (SignalAttachment *attachment in attachments) {
+        // TODO: Should we assume non-nil or should we check for non-nil?
```
charlesmchen-signal: `NSArray` can't hold nil; we can assume non-nil.

michaelkirk-signal:
> So this doesn't support editing the "album title", right?

Yup, that's right. The "send flow" section of the design spec is blank so far. At this point I'm assuming it'll be omitted from the initial sending experience. Though if we added it later, my understanding is that it would live on the `message.body` field so we're already in position to render that in the conversation view as a "caption" below the album 1/2/3/4/5 grid.

 /cc @myles-signal

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: Good point.

charlesmchen-signal:
> Though if we added it later, my understanding is that it would live on the message.body field so we're already in position to render that in the conversation view as a "caption" below the album 1/2/3/4/5 grid.

That what we're doing in conversation view - using `message.body` as the album title.  Similarly, `ConversationViewItem.hasBodyText/displayableBodyText` are set as usual to reflect album title.

michaelkirk-signal:
In db3e5f608

PR #3995, opened 2018-11-07
Charlesmchen/typing indicators vs self

charlesmchen-signal:
PTAL @michaelkirk


SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -667,6 +667,10 @@ - (void)handleIncomingEnvelope:(SSKProtoEnvelope *)envelope
         OWSFailDebug(@"typingMessage has invalid timestamp.");
         return;
     }
+    NSString *localNumber = self.tsAccountManager.localNumber;
+    if ([localNumber isEqualToString:envelope.source]) {
+        OWSLogVerbose(@"Ignoring typing indicators from self or linked device.");
```
michaelkirk-signal: I think we're missing a `return`?

charlesmchen-signal: Done. 

I can't test yet. ¯\_(ツ)_/¯

PR #3996, opened 2018-11-07
Display caption indicators for media albums in conversation view.

charlesmchen-signal:
PTAL @michaelkirk 

img_0141


Signal/src/ViewControllers/ConversationView/Cells/MediaGalleryCellView.swift
```diff
@@ -142,11 +146,39 @@ public class MediaGalleryCellView: UIStackView {
                 moreLabel.autoCenterInSuperview()
             }
         }
+
+        for itemView in itemViews {
+            guard !moreItemViews.contains(itemView) else {
```
michaelkirk-signal: TIOLI: Since we're skipping everything that's not in `moreItemViews`, would it make more sense to just iterate over the `moreItemViews`?

`for itemView in moreItemViews {`

Signal/src/ViewControllers/ConversationView/Cells/MediaGalleryCellView.swift
```diff
@@ -142,11 +146,39 @@ public class MediaGalleryCellView: UIStackView {
...
+            }
+            guard caption.count > 0 else {
+                continue
+            }
+            guard let icon = UIImage(named: "media_album_caption") else {
+                owsFailDebug("Couldn't load icon.")
+                continue
+            }
+            let iconView = UIImageView(image: icon.withRenderingMode(.alwaysTemplate))
+            iconView.tintColor = .ows_white
```
michaelkirk-signal: How does this look over a white image? I wonder if it'd be better to have an asset with a built in shadow, but that might be more a question for @myles-signal

Signal/src/ViewControllers/ConversationView/Cells/MediaGalleryCellView.swift
charlesmchen-signal: It's the opposite.  We're skipping everything that's _IN_ `moreItemViews`.

Signal/src/ViewControllers/ConversationView/Cells/MediaGalleryCellView.swift
charlesmchen-signal: I was going to ask Myles the same thing.  I've added it to the media album Qs card so you can track his answer.

PR #4003, opened 2018-11-07
album gallery view

michaelkirk-signal:
Adapts our gallery views (page view and tile view) to support the new multi-attachment schema.

The gist is that the galleries are now "attachment" based, not "message" based. 

To achieve this, attachment's now maintain a back-reference to their messageId, and everything else is a consequence of that.

No screenshots since there's no visual changes here.

PTAL @charlesmchen-signal


Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -1647,16 +1647,18 @@ - (void)handleFailedDownloadTapForMessage:(TSMessage *)message
...
-                                                    networkManager:self.networkManager];
-    [processor fetchAttachmentsForMessage:message
-        primaryStorage:self.primaryStorage
-        success:^(NSArray<TSAttachmentStream *> *attachmentStreams) {
-            OWSLogInfo(@"Successfully redownloaded attachment in thread: %@", message.thread);
-        }
-        failure:^(NSError *error) {
-            OWSLogWarn(@"Failed to redownload message with error: %@", error);
-        }];
+        [[OWSAttachmentsProcessor alloc] initWithAttachmentPointers:@[ attachmentPointer ]];
```
charlesmchen-signal: Huh, I've been touching this too.

Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -39,13 +39,13 @@ public class MediaGalleryItem: Equatable, Hashable {
...
     public static func == (lhs: MediaGalleryItem, rhs: MediaGalleryItem) -> Bool {
-        return lhs.message.uniqueId == rhs.message.uniqueId
+        return lhs.attachmentStream.uniqueId == rhs.attachmentStream.uniqueId
     }
 
     // MARK: Hashable
 
     public var hashValue: Int {
-        return message.hashValue
+        return attachmentStream.hashValue
```
charlesmchen-signal: `attachmentStream.uniqueId.hashValue`?

SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
```diff
@@ -78,25 +78,27 @@ - (void)runWithAttachmentHandler:(void (^)(NSArray<TSAttachmentStream *> *attach
...
                                                       quotedMessage:transcript.quotedMessage
                                                        contactShare:transcript.contact];
 
+    NSArray<TSAttachmentPointer *> *attachmentPointers =
+        [TSAttachmentPointer attachmentPointersFromProtos:transcript.attachmentPointerProtos
+                                             albumMessage:outgoingMessage];
+    for (TSAttachmentPointer *pointer in attachmentPointers) {
+        [pointer saveWithTransaction:transaction];
+        [outgoingMessage.attachmentIds addObject:pointer.uniqueId];
+    }
```
charlesmchen-signal: The conflicts here should be straightforward to resolve.

SignalServiceKit/src/Messages/Interactions/TSInteraction.m
```diff
@@ -98,7 +98,7 @@ - (instancetype)initInteractionWithTimestamp:(uint64_t)timestamp inThread:(TSThr
 {
     OWSAssertDebug(timestamp > 0);
 
-    self = [super initWithUniqueId:nil];
+    self = [super initWithUniqueId:[[NSUUID UUID] UUIDString]];
```
charlesmchen-signal: So these uniqueIds aren't compatible with the uniqueIds returned by `getAndIncrementMessageIdWithTransaction`?  Isn't the an issue?

SignalServiceKit/src/Messages/OWSMessageManager.m
```diff
@@ -1245,14 +1242,6 @@ - (TSIncomingMessage *_Nullable)handleReceivedEnvelope:(SSKProtoEnvelope *)envel
                     return nil;
                 }
 
-                if (body.length == 0 && attachmentIds.count < 1 && !contact) {
-                    OWSLogWarn(@"ignoring empty incoming message from: %@ for group: %@ with timestamp: %lu",
```
charlesmchen-signal: Don't we want to preserve this logic - perhaps in a new location?

charlesmchen-signal:
My only blocking concern is around the uniqueIds?

SignalServiceKit/src/Messages/Interactions/TSInteraction.m
michaelkirk-signal: I'm not sure what you mean by "compatible" they're all strings. 

The issue is that the `incremented id` doesn't get assigned until the item is saved.

But because there is now a circular reference between an interaction and its attachments, we need the id *before* it's been saved.

charlesmchen-signal: If we're going to _sometimes_ use UUIDs for interaction ids, let's migrate to always using them... and remove `getAndIncrementMessageIdWithTransaction`.

michaelkirk-signal: > If we're going to sometimes use UUIDs for interaction ids, let's migrate to always using them... and remove getAndIncrementMessageIdWithTransaction.

SGTM, I'll make that change.

Signal/src/ViewControllers/MediaGalleryViewController.swift
michaelkirk-signal: Yeah, I think that'd make more sense. Updated!

michaelkirk-signal:
> LGTM Although IMO we should remove getAndIncrementMessageIdWithTransaction.

I agree that if we're going to sometimes *not* use it, that we should probably never use it, but I'm inclined to to leave it in for now with an OWSFailDebug, since there are a lot of code paths that go through that method.

charlesmchen-signal:
> I agree that if we're going to sometimes not use it, that we should probably never use it, but I'm inclined to to leave it in for now with an OWSFailDebug, since there are a lot of code paths that go through that method.

But couldn't we replace:
```
 if (!self.uniqueId) {
        self.uniqueId = [OWSPrimaryStorage getAndIncrementMessageIdWithTransaction:transaction];
    }
```
With:
```
 if (!self.uniqueId) {
        self.uniqueId = [[NSUUID UUID] UUIDString]];
    }
```

PR #4004, opened 2018-11-08
Charlesmchen/attachment downloads

charlesmchen-signal:
Replaces `OWSAttachmentsProcessor` with new singleton `OWSAttachmentDownloads`.  

* Handles multiple attachments per message.
* Enforces "max simultaneous attachments download" limit.
* Will (soon, not in this PR) provide a simple API for querying/observing the download status of any incoming attachment.
* IMO simpler, cleaner API.  For example, "initial download" and "retry failed downloads" now use the same code path.

PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
+++ b/Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
@@ -234,7 +234,6 @@ public class ConversationMediaView: UIView {
         AssertIsOnMainThread()
 
         guard let loadBlock = loadBlock else {
-            owsFailDebug("Missing loadBlock")
             return
         }
         loadBlock()
@@ -245,7 +244,6 @@ public class ConversationMediaView: UIView {
         AssertIsOnMainThread()
 
         guard let unloadBlock = unloadBlock else {
-            owsFailDebug("Missing unloadBlock")
             return
         }
         unloadBlock()
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.h
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.h
@@ -39,8 +39,7 @@ extern const UIDataDetectorTypes kOWSAllowedDataDetectorTypes;
 
 - (void)didTapTruncatedTextMessage:(id<ConversationViewItem>)conversationItem;
 
-- (void)didTapFailedIncomingAttachment:(id<ConversationViewItem>)viewItem
-                     attachmentPointer:(TSAttachmentPointer *)attachmentPointer;
+- (void)didTapFailedIncomingAttachment:(id<ConversationViewItem>)viewItem;
 
 - (void)didTapConversationItem:(id<ConversationViewItem>)viewItem quotedReply:(OWSQuotedReplyModel *)quotedReply;
 - (void)didTapConversationItem:(id<ConversationViewItem>)viewItem
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController.m
@@ -67,7 +67,7 @@
 #import <SignalServiceKit/NSTimer+OWS.h>
 #import <SignalServiceKit/OWSAddToContactsOfferMessage.h>
 #import <SignalServiceKit/OWSAddToProfileWhitelistOfferMessage.h>
-#import <SignalServiceKit/OWSAttachmentsProcessor.h>
+#import <SignalServiceKit/OWSAttachmentDownloads.h>
 #import <SignalServiceKit/OWSBlockingManager.h>
 #import <SignalServiceKit/OWSDisappearingMessagesConfiguration.h>
 #import <SignalServiceKit/OWSIdentityManager.h>
@@ -295,6 +295,11 @@ typedef enum : NSUInteger {
 - (id<OWSTypingIndicators>)typingIndicators
 {
     return SSKEnvironment.shared.typingIndicators;
+}
+
+- (OWSAttachmentDownloads *)attachmentDownloads
+{
+    return SSKEnvironment.shared.attachmentDownloads;
 }
 
 #pragma mark -
@@ -1644,13 +1649,11 @@ typedef enum : NSUInteger {
 #pragma mark Bubble User Actions
 
 - (void)handleFailedDownloadTapForMessage:(TSMessage *)message
-                        attachmentPointer:(TSAttachmentPointer *)attachmentPointer
 {
-    OWSAttachmentsProcessor *processor =
-        [[OWSAttachmentsProcessor alloc] initWithAttachmentPointers:@[ attachmentPointer ]];
+    OWSAssert(message);
 
-    [self.editingDatabaseConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
-        [processor fetchAttachmentsForMessage:message
+    [self.uiDatabaseConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
+        [self.attachmentDownloads downloadAttachmentsForMessage:message
             transaction:transaction
             success:^(NSArray<TSAttachmentStream *> *attachmentStreams) {
                 OWSLogInfo(@"Successfully redownloaded attachment in thread: %@", message.thread);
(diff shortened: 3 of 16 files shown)


Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
```diff
@@ -234,7 +234,6 @@ public class ConversationMediaView: UIView {
         AssertIsOnMainThread()
 
         guard let loadBlock = loadBlock else {
-            owsFailDebug("Missing loadBlock")
```
charlesmchen-signal: Previously, this view assumed that it was only used to render `TSAttachmentStream`.  Now it's also used for `TSAttachmentPointer` (e.g. un-downloaded incoming attachments).

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.h
```diff
@@ -39,8 +39,7 @@ extern const UIDataDetectorTypes kOWSAllowedDataDetectorTypes;
 
 - (void)didTapTruncatedTextMessage:(id<ConversationViewItem>)conversationItem;
 
-- (void)didTapFailedIncomingAttachment:(id<ConversationViewItem>)viewItem
-                     attachmentPointer:(TSAttachmentPointer *)attachmentPointer;
+- (void)didTapFailedIncomingAttachment:(id<ConversationViewItem>)viewItem;
```
charlesmchen-signal: We'll be touching this plumbing again (for tap to retry albums), but not in this PR.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2202,27 +2205,22 @@ - (void)didTapConversationItem:(id<ConversationViewItem>)viewItem
...
                         [message saveWithTransaction:postSuccessTransaction];
                     }];
             }
             failure:^(NSError *error) {
                 OWSLogWarn(@"Failed to redownload thumbnail with error: %@", error);
                 [self.editingDatabaseConnection
-                    asyncReadWriteWithBlock:^(YapDatabaseReadWriteTransaction *postSuccessTransaction) {
-                        [message touchWithTransaction:transaction];
+                    readWriteWithBlock:^(YapDatabaseReadWriteTransaction *postSuccessTransaction) {
+                        [message touchWithTransaction:postSuccessTransaction];
```
charlesmchen-signal: This was a bug.

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
```diff
@@ -2202,27 +2205,22 @@ - (void)didTapConversationItem:(id<ConversationViewItem>)viewItem
...
-
-    [self.editingDatabaseConnection asyncReadWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
-        [processor fetchAttachmentsForMessage:nil
-            transaction:transaction
+    [self.editingDatabaseConnection asyncReadWithBlock:^(YapDatabaseReadTransaction *transaction) {
+        [self.attachmentDownloads downloadAttachmentPointer:attachmentPointer
             success:^(NSArray<TSAttachmentStream *> *attachmentStreams) {
                 OWSAssertDebug(attachmentStreams.count == 1);
                 TSAttachmentStream *attachmentStream = attachmentStreams.firstObject;
                 [self.editingDatabaseConnection
```
charlesmchen-signal: `OWSAttachmentDownloads` ensures that success/failure callbacks are always dispatched to a worker queue, so these callbacks don't need to do "async" any more.

SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
```diff
@@ -78,18 +95,17 @@ - (void)runWithAttachmentHandler:(void (^)(NSArray<TSAttachmentStream *> *attach
         return;
     }
 
-    OWSAttachmentsProcessor *attachmentsProcessor =
-        [[OWSAttachmentsProcessor alloc] initWithAttachmentProtos:transcript.attachmentPointerProtos
-                                                   networkManager:self.networkManager
-                                                      transaction:transaction];
-
+    NSArray<TSAttachmentPointer *> *attachmentPointers =
+        [self.attachmentDownloads saveAttachmentPointersForAttachmentProtos:transcript.attachmentPointerProtos
```
charlesmchen-signal: I've pulled out `saveAttachmentPointersForAttachmentProtos:transcript` into a separate method so that we can be explicit about doing this work separately from the downloads.

SignalServiceKit/src/Messages/Attachments/OWSAttachmentDownloads.m
```diff
@@ -35,42 +36,104 @@
...
+@end
+
+#pragma mark -
+
+@interface OWSAttachmentDownloads ()
+
+// This property should only be accessed while synchronized on this class.
+@property (nonatomic, readonly) NSMutableSet<NSString *> *downloadingAttachmentIdSet;
+// This property should only be accessed while synchronized on this class.
+@property (nonatomic, readonly) NSMutableArray<OWSAttachmentDownloadJob *> *attachmentDownloadJobQueue;
```
charlesmchen-signal: I decided to use an explicit (non-durable) queue so that we could ensure that we never have two downloads for the same attachment at a time.  Maybe this was overkill?

SignalServiceKit/src/Messages/Attachments/OWSAttachmentDownloads.m
```diff
@@ -80,99 +143,233 @@ - (instancetype)initWithAttachmentProtos:(NSArray<SSKProtoAttachmentPointer *> *
...
+    [self startDownloadIfPossible];
+}
+
+- (void)startDownloadIfPossible
+{
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        OWSAttachmentDownloadJob *_Nullable job;
+
+        @synchronized(self) {
+            const NSUInteger kMaxSimultaneousDownloads = 4;
```
charlesmchen-signal: WDYT?

Signal/src/ViewControllers/ConversationView/ConversationViewController.m
michaelkirk-signal: 👀

SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
michaelkirk-signal: Right, as you mentioned, this is going to conflict with #4003.

SignalServiceKit/src/Messages/Attachments/OWSAttachmentDownloads.m
michaelkirk-signal: > WDYT 

Specific to this line, I think 4 concurrent downloads is a reasonable first guess.

More generally, the public API makes sense, and it's clear how it solves the problem. 

My instincts would have been to use NSOperation/NSOperationQueue/maxConcurrentOperations, rather than write our own synchronized throttling logic from scratch, but this should also work.

SignalServiceKit/src/Messages/Attachments/OWSAttachmentDownloads.m
michaelkirk-signal: If we eventually wanted downloads to restart after the app is killed, we might consider porting it to the durable queue.

But this seems fine for now, as we've always previously required manual restart of failed downloads.

SignalServiceKit/src/Messages/Attachments/OWSAttachmentDownloads.m
```diff
@@ -35,42 +36,104 @@
...
 
     NSMutableArray<NSString *> *attachmentIds = [NSMutableArray new];
+    for (TSAttachment *attachment in attachments) {
+        [attachmentIds addObject:attachment.uniqueId];
+    }
+    return [attachmentIds copy];
+}
+
+- (NSArray<TSAttachmentPointer *> *)
+    saveAttachmentPointersForAttachmentProtos:(NSArray<SSKProtoAttachmentPointer *> *)attachmentProtos
```
michaelkirk-signal: As we discussed this AM (I think), it makes sense for most of the logic in this method to live on some kind of `TSAttachmentPointer#fromProto` method.

SignalServiceKit/src/Devices/OWSRecordTranscriptJob.m
charlesmchen-signal: These conflicts have been resolved.  I found some issues with `OWSMessageManager` wherein some parts of the code have to be restructured a bit.

SignalServiceKit/src/Messages/Attachments/OWSAttachmentDownloads.m
charlesmchen-signal: Yup!

PR #4005, opened 2018-11-08
Charlesmchen/album progress

charlesmchen-signal:
PTAL @michaelkirk 

simulator screen shot - iphone x - 2018-11-08 at 11 15 21

### DIFF
--- a/Signal/src/ViewControllers/ColorPickerViewController.swift
+++ b/Signal/src/ViewControllers/ColorPickerViewController.swift
@@ -429,7 +429,7 @@ private class MockConversationViewItem: NSObject, ConversationViewItem {
         return 2
     }
 
-    func firstValidGalleryAttachment() -> TSAttachmentStream? {
+    func firstValidAlbumAttachment() -> TSAttachmentStream? {
         owsFailDebug("unexpected invocation")
         return nil
     }
--- a/Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
+++ b/Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
@@ -6,9 +6,19 @@ import Foundation
 
 @objc
 public class ConversationMediaView: UIView {
+
+    // MARK: - Dependencies
+
+    private var attachmentDownloads: OWSAttachmentDownloads {
+        return SSKEnvironment.shared.attachmentDownloads
+    }
+
+    // MARK: -
+
     private let mediaCache: NSCache<NSString, AnyObject>
     private let attachment: TSAttachment
     private let isOutgoing: Bool
+    private let maxMessageWidth: CGFloat
     private var loadBlock : (() -> Void)?
     private var unloadBlock : (() -> Void)?
     private var didFailToLoad = false
@@ -16,10 +26,12 @@ public class ConversationMediaView: UIView {
     @objc
     public required init(mediaCache: NSCache<NSString, AnyObject>,
                          attachment: TSAttachment,
-                         isOutgoing: Bool) {
+                         isOutgoing: Bool,
+                         maxMessageWidth: CGFloat) {
         self.mediaCache = mediaCache
         self.attachment = attachment
         self.isOutgoing = isOutgoing
+        self.maxMessageWidth = maxMessageWidth
 
         super.init(frame: .zero)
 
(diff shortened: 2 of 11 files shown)


Signal/src/ViewControllers/ColorPickerViewController.swift
```diff
@@ -429,7 +429,7 @@ private class MockConversationViewItem: NSObject, ConversationViewItem {
         return 2
     }
 
-    func firstValidGalleryAttachment() -> TSAttachmentStream? {
+    func firstValidAlbumAttachment() -> TSAttachmentStream? {
```
charlesmchen-signal: There's a bunch of places I missed during the "gallery" -> "album" rename.  I've fixed them here.

charlesmchen-signal:
Note: The design isn't final, so this approximates where we'll land.

Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
```diff
@@ -48,13 +62,53 @@ public class ConversationMediaView: UIView {
...
+            configureForMissingOrInvalid()
+            return
+        }
+        guard let attachmentId = attachmentPointer.uniqueId else {
+            owsFailDebug("Attachment stream missing unique ID.")
+            configureForMissingOrInvalid()
+            return
+        }
+
+        guard let progress = attachmentDownloads.downloadProgress(forAttachmentId: attachmentId) else {
```
michaelkirk-signal: Seems like we're not doing anything with this `progress`?

Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
```diff
@@ -117,7 +174,10 @@ public class ConversationMediaView: UIView {
         stillImageView.layer.minificationFilter = kCAFilterTrilinear
         stillImageView.layer.magnificationFilter = kCAFilterTrilinear
         stillImageView.backgroundColor = Theme.offBackgroundColor
-        addMediaSubview(stillImageView)
+        addSubview(stillImageView)
+        stillImageView.autoPinEdgesToSuperviewEdges()
+        addUploadProgressIfNecessary(stillImageView) { (_) in
```
michaelkirk-signal: TIOLI: Maybe this callback handler should be optional, since we're not always doing anything with it?

SignalServiceKit/src/Messages/Attachments/OWSAttachmentDownloads.m
```diff
@@ -593,8 +615,10 @@ - (void)downloadFromLocation:(NSString *)location
                 return;
             }
 
+            job.progress = progress.fractionCompleted;
```
michaelkirk-signal: I wonder if it'd be beneficial to just pass back the entire NSProgress rather than a number... but maybe YAGNI, since we're not (yet) doing anything like aggregating the results.

Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
```diff
@@ -48,13 +62,53 @@ public class ConversationMediaView: UIView {
...
+        }
+
+        backgroundColor = UIColor.ows_gray05
+        let progressView = AttachmentDownloadView(attachmentId: attachmentId, radius: maxMessageWidth * 0.1)
+        self.addSubview(progressView)
+        progressView.autoPinEdgesToSuperviewEdges()
+    }
+
+    private func addUploadProgressIfNecessary(_ subview: UIView,
+                                                    progressCallback: @escaping ProgressCallback) {
```
michaelkirk-signal: indent seems off

Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
charlesmchen-signal: Modified guard.

Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
charlesmchen-signal: Done.

Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
charlesmchen-signal: Fixed.

PR #4008, opened 2018-11-08
Show captions in gallery page view

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/Signal/src/ViewControllers/MediaGalleryViewController.swift
+++ b/Signal/src/ViewControllers/MediaGalleryViewController.swift
@@ -4,12 +4,6 @@
 
 import Foundation
 
-extension String {
-    var filterForDisplay: String? {
-        return (self as NSString).filterStringForDisplay()
-    }
-}
-
 public enum GalleryDirection {
     case before, after, around
 }
--- a/Signal/src/ViewControllers/MediaGalleryViewController.swift
+++ b/Signal/src/ViewControllers/MediaGalleryViewController.swift
@@ -31,6 +31,10 @@ public class MediaGalleryItem: Equatable, Hashable {
         return attachmentStream.isImage
     }
 
+    var caption: String? {
+        return attachmentStream.caption
+    }
+
     public typealias AsyncThumbnailBlock = (UIImage) -> Void
     func thumbnailImage(async:@escaping AsyncThumbnailBlock) -> UIImage? {
         return attachmentStream.thumbnailImageSmall(success: async, failure: {})
--- a/Signal/src/ViewControllers/MediaGalleryViewController.swift
+++ b/Signal/src/ViewControllers/MediaGalleryViewController.swift
@@ -4,6 +4,12 @@
 
 import Foundation
 
+extension String {
+    var filterForDisplay: String? {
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -31,6 +31,10 @@ public class MediaGalleryItem: Equatable, Hashable {
         return attachmentStream.isImage
     }
 
+    var caption: String? {
+        return attachmentStream.caption
```
charlesmchen-signal: We need to apply `filterStringForDisplay`.  Maybe we should add `[TSAttachment captionForDisplay]`?

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -654,3 +755,38 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
...
+        addSubview(gradientView)
+        gradientView.autoPinEdgesToSuperviewEdges()
+
+        addSubview(label)
+        label.font = UIFont.ows_dynamicTypeBody
+        label.textColor = .white
+
+        // Usually captions are short, but they can be as long as 2k.
+        // We don't have UI for viewing infinitely large captions, so
+        // we do some not-ideal things to broaden the lenght of the
```
charlesmchen-signal: s/lenght/length/.

`minimumScaleFactor` a) reduces accessibility by rendering the text smaller, making it harder to read. b) still doesn't guarantee us that all text will fit in a reasonable way.  

What if instead we set `numberOfLines` to something like 5 and if users tap on the caption we push a simple view to display all of the text, e.g. `LongTextViewController`.  LGTV isn't pretty but it ensures that all of the caption is readable by all.

michaelkirk-signal: Good point. Long captions are now in a scroll view. Getting a scroll view to grow to fit it's content, but only up to a point, was a little tricky.

Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -4,6 +4,12 @@
 
 import Foundation
 
+extension String {
+    var filterForDisplay: String? {
```
michaelkirk-signal: whoops, I'm going to move this to Strings+OWS.swift

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -654,3 +750,102 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
...
+            } else if isScrollEnabled {
+                isScrollEnabled = false
+            }
+        }
+
+        override var intrinsicContentSize: CGSize {
+            var size = super.intrinsicContentSize
+
+            if size.height == UIViewNoIntrinsicMetric {
+                size.height = layoutManager.usedRect(for: textContainer).height + textContainerInset.top + textContainerInset.bottom
```
charlesmchen-signal: TIL

michaelkirk-signal:
in cc068b0b7

PR #4009, opened 2018-11-08
 Render single media as albums.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
+++ b/Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
@@ -187,14 +187,6 @@ const UIDataDetectorTypes kOWSAllowedDataDetectorTypes
     return (TSMessage *)self.viewItem.interaction;
 }
 
-- (CGSize)mediaSize
-{
-    // This should always be valid for the appropriate cell types.
-    OWSAssertDebug(self.viewItem.mediaSize.width > 0 && self.viewItem.mediaSize.height > 0);
-
-    return self.viewItem.mediaSize;
-}
-
 - (BOOL)isQuotedReply
 {
     // This should always be valid for the appropriate cell types.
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -795,47 +774,22 @@ - (UIView *)loadViewForMediaAlbum
...
     };
-    for (UIView *itemView in albumView.itemViews) {
-        OWSBubbleShapeView *strokeView = [[OWSBubbleShapeView alloc] initDraw];
-        strokeView.strokeColor = [UIColor colorWithWhite:0.5f alpha:0.4f];
-        strokeView.strokeThickness = 1.f;
-        [itemView addSubview:strokeView];
-        [self.bubbleView addPartnerView:strokeView];
-        [self.viewConstraints addObjectsFromArray:[strokeView ows_autoPinToSuperviewEdges]];
+
+    // Only apply "inner shadow" for single media, not albums.
```
michaelkirk-signal: I thought the spec had shadows in albums too, was that an intentional design change?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
```diff
@@ -1421,10 +1350,33 @@ - (void)handleMediaTapGesture
...
+        case OWSMessageCellType_MediaAlbum: {
             OWSAssertDebug(self.bodyMediaView);
             OWSAssertDebug(self.viewItem.mediaAlbumItems.count > 0);
 
+            if (self.viewItem.mediaAlbumHasFailedAttachment) {
+                [self.delegate didTapFailedIncomingAttachment:self.viewItem];
+                return;
+            }
+
+            // TODO: We might be able to get rid of this.
```
michaelkirk-signal: Just for my own understanding re: this TODO, do you mean to get rid of it and replace it with the "did tap on specific album cell" behavior?

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
charlesmchen-signal: I couldn't find out a good way to do it, so Myles approved taking them out.

Signal/src/ViewControllers/ConversationView/Cells/OWSMessageBubbleView.m
charlesmchen-signal: > Just for my own understanding re: this TODO, do you mean to get rid of it and replace it with the "did tap on specific album cell" behavior?

That's half of it. The other half of it is whether or not we need to special-case tapping on "lone" media v. "album" media.

PR #4015, opened 2018-11-09
Discard receipts for unregistered users.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Messages/OWSOutgoingReceiptManager.m
+++ b/SignalServiceKit/src/Messages/OWSOutgoingReceiptManager.m
@@ -4,6 +4,7 @@
 
 #import "OWSOutgoingReceiptManager.h"
 #import "AppReadiness.h"
+#import "OWSError.h"
 #import "OWSMessageSender.h"
 #import "OWSPrimaryStorage.h"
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Messages/OWSOutgoingReceiptManager.m
```diff
@@ -202,6 +203,11 @@ - (void)process {
                 failure:^(NSError *error) {
                     OWSLogError(@"Failed to send %@ receipts to sender with error: %@", receiptName, error);
 
+                    if (error.domain == OWSSignalServiceKitErrorDomain
+                        && error.code == OWSErrorCodeNoSuchSignalRecipient) {
+                        [self dequeueReceiptsWithRecipientId:recipientId timestamps:timestamps receiptType:receiptType];
```
michaelkirk-signal: Is it reasonable that we're still rejecting the promise in this case? It seems like it doesn't really make a difference based on how it's currently being used, so TIOLI.

charlesmchen-signal: Right.

PR #4016, opened 2018-11-09
Prevent long version strings from being scrubbed in the logs.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit/src/Util/AppVersion.m
+++ b/SignalServiceKit/src/Util/AppVersion.m
@@ -68,11 +68,16 @@ NSString *const kNSUserDefaults_LastCompletedLaunchAppVersion_SAE
     [[NSUserDefaults appUserDefaults] setObject:self.currentAppVersion forKey:kNSUserDefaults_LastAppVersion];
     [[NSUserDefaults appUserDefaults] synchronize];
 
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Util/AppVersion.m
```diff
@@ -68,11 +68,16 @@ - (void)configure {
...
+    NSString *longVersionString = [[[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleVersion"]
+        stringByReplacingOccurrencesOfString:@"."
+                                  withString:@"_"];
+
     OWSLogInfo(@"firstAppVersion: %@", self.firstAppVersion);
     OWSLogInfo(@"lastAppVersion: %@", self.lastAppVersion);
-    OWSLogInfo(@"currentAppVersion: %@ (%@)",
-        self.currentAppVersion,
-        [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleVersion"]);
+    OWSLogInfo(@"currentAppVersion: %@ (%@)", self.currentAppVersion, longVersionString);
```
charlesmchen-signal: Alternately we could format this line like so:

```
currentAppVersion: 2.30.0 (.32)
```

PR #4018, opened 2018-11-09
Tap on album item.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/MediaGalleryViewController.swift
+++ b/Signal/src/ViewControllers/MediaGalleryViewController.swift
@@ -264,10 +264,7 @@ class MediaGallery: NSObject, MediaGalleryDataSource, MediaTileViewControllerDel
 
         self.options = options
         self.mediaGalleryFinder = OWSMediaGalleryFinder(thread: thread)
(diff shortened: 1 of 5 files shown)


Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -264,10 +264,7 @@ class MediaGallery: NSObject, MediaGalleryDataSource, MediaTileViewControllerDel
 
         self.options = options
         self.mediaGalleryFinder = OWSMediaGalleryFinder(thread: thread)
-        let navController = MediaGalleryNavigationController()
-        self.navigationController = navController
         super.init()
-        navController.retainUntilDismissed = self
```
charlesmchen-signal: We can't safely allocate `MediaGalleryNavigationController` in the initializer, since we only hold a weak reference to it.  We need to allocate it right before present it.

michaelkirk-signal: Yeah, you're right, that was dumb.

PR #4019, opened 2018-11-09
Mkirk/gallery pager fixups

michaelkirk-signal:
PTAL @charlesmchen-signal 

Handful of small fixes, highlight in individual commits.


Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -801,7 +803,7 @@ class CaptionView: UIView {
 
     class CaptionTextView: UITextView {
 
-        var kMaxHeight: CGFloat = ScaleFromIPhone5(100)
+        var kMaxHeight: CGFloat = ScaleFromIPhone5(200)
```
michaelkirk-signal: I'd accidentally lowered this to 100 while testing the scroll view, but I think this is closer to where we'd want to be.

PR #4022, opened 2018-11-12
partial revert of 4c5d46e8f89a512832d6d304fad9208e63f93a96

michaelkirk-signal:
Restore old UIImagePicker until new UIImagePicker supports browse-by-album

PTAL @charlesmchen-signal


michaelkirk-signal:
> Would it make sense to put this behind a feature flag?

Sure, that would be simple enough. I'll do that before I merge.

michaelkirk-signal:
merged in bf81f00c0

PR #4031, opened 2018-11-13
fix disable earpiece

michaelkirk-signal:
PTAL @charlesmchen

### DIFF
--- a/SignalMessaging/utils/ProximityMonitoringManager.swift
+++ b/SignalMessaging/utils/ProximityMonitoringManager.swift
@@ -11,7 +11,6 @@ public protocol OWSProximityMonitoringManager: class {
 @objc
 public class OWSProximityMonitoringManagerImpl: NSObject, OWSProximityMonitoringManager {
     var lifetimes: [Weak<AnyObject>] = []
-    let serialQueue = DispatchQueue(label: "ProximityMonitoringManagerImpl")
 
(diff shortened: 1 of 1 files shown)


SignalMessaging/utils/ProximityMonitoringManager.swift
```diff
@@ -39,20 +49,44 @@ public class OWSProximityMonitoringManagerImpl: NSObject, OWSProximityMonitoring
...
+        // `device.proximityState` is true (while the device is held to the ear)
+        // then `device.proximityState` remains true, even after we bring the phone
+        // away from the ear and re-enable monitoring.
+        //
+        // To resolve this, we wait to disable proximity monitoring until `proximityState`
+        // is false.
+        if self.device.proximityState {
+            self.add(lifetime: self)
+        } else {
+            self.remove(lifetime: self)
```
charlesmchen-signal: `proximitySensorStateDidChange(...)` is presumably called on the main thread.  `add()`/`remove()`/`reconcile()` do their work on the serial queue.  I think there's a (rare, possibly inconsequential) race wherein `reconcile()` is called _after_ `proximityState` becomes true but before the `add(lifetime:self)` is done.  

I'm going to modify this PR to use `NSLock` or `objc_sync_enter()` to avoid this race.

PR #4032, opened 2018-11-13
Fix glitches in conversation media view.

charlesmchen-signal:
PTAL @michaelkirk

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
+++ b/Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
@@ -61,7 +61,7 @@ public class ConversationMediaView: UIView {
             return
         }
         guard !isFailedDownload else {
-            configureForMissingOrInvalid(.failed)
+            configure(forError: .failed)
             return
         }
         if attachmentStream.isAnimated {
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
```diff
@@ -7,6 +7,12 @@ import Foundation
 @objc(OWSConversationMediaView)
 public class ConversationMediaView: UIView {
 
+    private enum MediaError {
+        case missing
+        case invalid
+        case failed
```
charlesmchen-signal: We need to distinguish these cases more carefully.  In particular, we want to avoid "flashing" the "invalid media indicator" when media start/complete downloading.

Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
```diff
@@ -270,25 +282,28 @@ public class ConversationMediaView: UIView {
...
         let icon: UIImage
-        if isFailedDownload {
-            guard let asset = UIImage(named: "media_retry") else {
-                owsFailDebug("Missing image")
-                return
-            }
-            icon = asset
-        } else {
+        switch (error) {
+            case .failed:
```
michaelkirk-signal: weird indent

Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
```diff
@@ -270,25 +282,28 @@ public class ConversationMediaView: UIView {
         return attachmentPointer.state == .failed
     }
 
-    private func configureForMissingOrInvalid() {
+    private func configureForMissingOrInvalid(_ error: MediaError) {
```
michaelkirk-signal: TIOLI: maybe `configure(error: )`

Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
charlesmchen-signal: Fixed.

Signal/src/ViewControllers/ConversationView/Cells/ConversationMediaView.swift
charlesmchen-signal: Better, done.

PR #4033, opened 2018-11-14
Album rail in Gallery

michaelkirk-signal:
PTAL @charlesmchen-signal 

gallery_pager

Note: this doesn't reflect the design's myles just sent.

### DIFF
--- a/Signal/src/ViewControllers/MediaPageViewController.swift
+++ b/Signal/src/ViewControllers/MediaPageViewController.swift
@@ -3,6 +3,7 @@
 //
 
 import UIKit
+import PromiseKit
 
 // Objc wrapper for the MediaGalleryItem struct
 @objc
@@ -53,10 +54,11 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
             return
         }
 
-        self.updateTitle(item: item)
-        self.updateCaption(item: item)
-        self.setViewControllers([galleryPage], direction: direction, animated: isAnimated)
-        self.updateFooterBarButtonItems(isPlayingVideo: false)
+        updateTitle(item: item)
+        updateCaption(item: item)
+        setViewControllers([galleryPage], direction: direction, animated: isAnimated)
+        updateFooterBarButtonItems(isPlayingVideo: false)
+        updateMediaRail()
     }
 
     private let uiDatabaseConnection: YapDatabaseConnection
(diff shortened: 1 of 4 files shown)


charlesmchen-signal:
This PR is against `master`.  Shouldn't it be against `release/2.32.0`?

Signal/src/ViewControllers/MediaPageViewController.swift
```diff
@@ -108,6 +110,26 @@ class MediaPageViewController: UIPageViewController, UIPageViewControllerDataSou
     var currentCaptionView: CaptionView!
     var pendingCaptionView: CaptionView!
 
+    // MARK: ImageRail
+
+    var galleryRailView: GalleryRailView!
```
charlesmchen-signal: Is the intention to re-use this rail class for the attachment approval changes?

Signal/src/views/GalleryRailView.swift
```diff
@@ -0,0 +1,274 @@
...
+        let animationDuration: TimeInterval = 0.2
+
+        guard let itemProvider = itemProvider else {
+            UIView.animate(withDuration: animationDuration) {
+                self.isHidden = true
+            }
+            return
+        }
+
+        let sameRailItems = { (lhs: [GalleryRailItem], rhs: [GalleryRailItem]) -> Bool in
```
charlesmchen-signal: s/sameRailItems/areRailItemsEqual/ ?

charlesmchen-signal: areRailItemsIdentical?

Signal/src/views/GalleryRailView.swift
```diff
@@ -0,0 +1,274 @@
...
+                self.isHidden = true
+            }
+            return
+        }
+
+        UIView.animate(withDuration: animationDuration) {
+            self.isHidden = false
+        }
+
+        let cellViews = buildCellViews(items: itemProvider.railItems)
```
charlesmchen-signal: So we always maintain a `GalleryRailCellView` for each asset in the the album, but not for each asset in the entire "all media gallery".  

Each `GalleryRailCellView` eagerly loads a thumbnail image.  This seems like it will work fine if our max album size is 32, but it won't scale to album sizes of 1000.  I suppose we could use a UICollectionView for horizontal scrolling if we needed to?

michaelkirk-signal: Right - I considered using an embedded collectionview, but I thought given that our limited album size allowed it, that it was simpler to do it this way.

Signal/src/ViewControllers/MediaPageViewController.swift
michaelkirk-signal: It's not clear to me yet how much will be reusable. Originally the designs were quite different so most of the complicated code wasn't relevant. I'll try to re-use where it makes sense.

Signal/src/views/GalleryRailView.swift
michaelkirk-signal: Done.

michaelkirk-signal:
in 8eff8966b

> LGTM although I wish there were screenshots!

There was just the one, sorry. More screenshots in the followup PR which is closer to the proposed design. #4035

PR #4035, opened 2018-11-14
Gallery pager style changes

michaelkirk-signal:
PTAL @charlesmchen-signal 

**no caption, no album**
gallery_style2

**short caption, in an album**
gallery_style4

**no caption, in an album**
gallery_style3

**long caption, no album**
gallery_style1

### DIFF
--- a/Signal/src/ViewControllers/MediaGalleryViewController.swift
+++ b/Signal/src/ViewControllers/MediaGalleryViewController.swift
@@ -277,9 +277,9 @@ class MediaGalleryNavigationController: OWSNavigationController {
 
         navigationBar.respectsTheme = false
         navigationBar.barStyle = .black
-        navigationBar.titleTextAttributes = [NSAttributedStringKey.foregroundColor: Theme.galleryIconColor]
-        navigationBar.barTintColor = Theme.galleryBackgroundColor.withAlphaComponent(0.6)
-        navigationBar.tintColor = Theme.galleryIconColor
+        navigationBar.titleTextAttributes = [NSAttributedStringKey.foregroundColor: Theme.darkThemePrimaryColor]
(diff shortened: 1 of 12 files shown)


Signal/src/ViewControllers/MediaGalleryViewController.swift
```diff
@@ -437,7 +450,7 @@ class MediaGallery: NSObject, MediaGalleryDataSource, MediaTileViewControllerDel
 
                             // fade out content behind the pageViewController
                             // and behind the presentation view
-                            self.navigationController.view.backgroundColor = Theme.backgroundColor
+                            self.navigationController.view.backgroundColor = Theme.galleryBackgroundColor
```
charlesmchen-signal: Huh, so these "gallery..." properties are essentially "always dark theme"?  I wonder if we should call them "dark..." instead so that we can reuse them in places like the image picker, etc.

Signal/src/views/CaptionView.swift
```diff
@@ -0,0 +1,182 @@
...
+            return self.bounds.height == kMaxHeight
+        }
+
+        override func layoutSubviews() {
+            super.layoutSubviews()
+
+            // Enable/disable scrolling depending on wether we've clipped
+            // content in `intrinsicContentSize`
+            if doesContentNeedScroll {
+                if !isScrollEnabled {
```
charlesmchen-signal: Why do we need this condition?

michaelkirk-signal: Hmm... I don't recall. I'm not sure if setting isScrollEnabled has any non-obvious side effects I'll try removing it and see if anything seems broken - it might just be cruft.

Signal/src/ViewControllers/MediaGalleryViewController.swift
michaelkirk-signal: Ok, I'll rename.

PR #4036, opened 2018-11-15
Mkirk/fix all media view

michaelkirk-signal:
The "All Media" button was broken in the conversation settings

While working on this I noticed we were never dealloc'ing ConversationSettings.


Signal/src/ViewControllers/ThreadSettings/OWSConversationSettingsViewController.m
```diff
@@ -121,6 +121,7 @@ - (void)commonInit
 
 - (void)dealloc
 {
+    OWSLogDebug(@"");
```
charlesmchen-signal: `OWSConversationSettingsViewController` is a `OWSTableViewController` which is as `OWSViewController` which also logs deallocation. Is this necessary?

michaelkirk-signal: I'll remove it.

PR #4038, opened 2018-11-15
sender rail

michaelkirk-signal:
PTAL @charlesmchen-signal 

** rail for multimage **
multi_image_shows_rail

** no rail for single image **

single_image_shows_no_rail

**keyboard covers rails**
keyboard_covers_rail


SignalMessaging/ViewControllers/AttachmentApprovalViewController.swift
```diff
@@ -124,6 +224,77 @@ public class AttachmentApprovalViewController: UIPageViewController, UIPageViewC
 
     // MARK: - View Helpers
 
+    func buildRemoveButton(attachmentItem: SignalAttachmentItem) -> RemoveAttachmentButton {
+        let button = RemoveAttachmentButton()
+        button.setImage(#imageLiteral(resourceName: "ic_small_x"), for: .normal)
+        button.attachmentItem = attachmentItem
+        return button
+    }
```
charlesmchen-signal: Should we write a (cliché) UIButton subclass (`OWSButton`?) that is block-based to avoid the need for this kind of subclass?

PR #4039, opened 2018-11-15
Charlesmchen/image picker add more

charlesmchen-signal:
PTAL @michaelkirk


Signal/Images.xcassets/album_add_more.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
+  "images" : [
```
charlesmchen-signal: This asset is temporary.

Signal/src/ViewControllers/PhotoLibrary/ImagePickerController.swift
```diff
@@ -26,6 +26,11 @@ class ImagePickerGridController: UICollectionViewController, PhotoLibraryDelegat
 
     private let titleLabel = UILabel()
 
+    private var selectedIds = Set<String>()
+
+    // This variable should only be accessed on the main thread.
+    private var assetIdToCommentMap = [String: String]()
```
charlesmchen-signal: Two new pieces of state: the selection and the current caption state.

Signal/src/ViewControllers/PhotoLibrary/ImagePickerController.swift
```diff
@@ -52,16 +57,26 @@ class ImagePickerGridController: UICollectionViewController, PhotoLibraryDelegat
...
+        if let navBar = self.navigationController?.navigationBar as? OWSNavigationBar {
+            navBar.makeClear()
+        } else {
+            owsFailDebug("Invalid nav bar.")
+        }
+
+        let cancelButton = UIBarButtonItem(barButtonSystemItem: .stop,
+                                           target: self,
+                                           action: #selector(didPressCancel))
+        cancelButton.tintColor = .ows_gray05
```
charlesmchen-signal: In hindsight, I wish I had added "dark..." methods to `Theme`, sort of like you did.

Signal/src/ViewControllers/PhotoLibrary/ImagePickerController.swift
```diff
@@ -92,16 +107,38 @@ class ImagePickerGridController: UICollectionViewController, PhotoLibraryDelegat
...
+    }
+
+    private func reloadDataAndRestoreSelection() {
+        guard let collectionView = collectionView else {
+            owsFailDebug("Missing collectionView.")
+            return
+        }
+
+        collectionView.reloadData()
+        collectionView.layoutIfNeeded()
```
charlesmchen-signal: `reloadData` clears the selection, so we need to restore it.

SignalMessaging/attachments/SignalAttachment.swift
```diff
@@ -160,6 +160,10 @@ public class SignalAttachment: NSObject {
     @objc
     public let dataUTI: String
 
+    // Can be used by views to link this SignalAttachment with an arbitrary source.
+    @objc
+    public var sourceId: String?
```
charlesmchen-signal: This was a _lot_ more expedient & clean than any other linking mechanism I could come up with.

Signal/src/ViewControllers/PhotoLibrary/ImagePickerController.swift
```diff
@@ -26,6 +26,11 @@ class ImagePickerGridController: UICollectionViewController, PhotoLibraryDelegat
 
     private let titleLabel = UILabel()
 
+    private var selectedIds = Set<String>()
```
charlesmchen-signal: There's an interesting question about the ordering of the selected assets.  Order of selection?  Order within the collection?

SignalMessaging/attachments/SignalAttachment.swift
michaelkirk-signal: That seems fine, but can we call this field `assetId` instead? or `sourceAssetId`

charlesmchen-signal: Done.

Signal/src/ViewControllers/PhotoLibrary/ImagePickerController.swift
VanillaRex: I think at some point in the commit we have gone from default scroll position being the last item in the collection to the first. This means the user would now have to scroll down through potentially thousands of images to get to the most recent. Any idea where this might have occurred?

PR #4045, opened 2018-11-16
Mkirk/ud error handling

michaelkirk-signal:
PTAL @charlesmchen-signal 

(depends on https://github.com/signalapp/SignalMetadataKit/pull/3)

### DIFF
--- a/SignalServiceKit/src/Messages/OWSMessageDecrypter.m
+++ b/SignalServiceKit/src/Messages/OWSMessageDecrypter.m
@@ -37,6 +37,7 @@ NSError *EnsureDecryptError(NSError *_Nullable error, NSString *fallbackErrorDes
     if (error) {
         return error;
     }
+    OWSCFailDebug(@"Caller should provide specific error");
     return OWSErrorWithCodeDescription(OWSErrorCodeFailedToDecryptUDMessage, fallbackErrorDescription);
 }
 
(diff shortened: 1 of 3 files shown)


SignalServiceKit/src/Messages/OWSMessageDecrypter.m
```diff
@@ -434,89 +435,137 @@ - (void)decryptUnidentifiedSender:(SSKProtoEnvelope *)envelope
...
-                OWSLogError(@"%@", errorDescription);
-                NSError *error = OWSErrorWithCodeDescription(OWSErrorCodeFailedToDecryptMessage, errorDescription);
-                failureBlock(error);
-            });
+            OWSFailDebug(@"Could not decrypt UD message: %@", underlyingError);
+            failureBlock(underlyingError);
+            return;
+        }
+
+        if (decryptResult.messageType == SMKMessageTypePrekey) {
```
michaelkirk-signal: The remainder of the changes in the method are indent only, because it's no longer in a try/catch.

SignalServiceKit/src/Messages/OWSMessageDecrypter.m
```diff
@@ -434,89 +435,137 @@ - (void)decryptUnidentifiedSender:(SSKProtoEnvelope *)envelope
...
+
+                NSNumber *senderDeviceId = decryptError.userInfo[SecretSessionKnownSenderError.kSenderDeviceIdKey];
+                OWSAssert(senderDeviceId);
+
+                SSKProtoEnvelopeBuilder *identifiedEnvelopeBuilder = envelope.asBuilder;
+                identifiedEnvelopeBuilder.source = senderRecipientId;
+                identifiedEnvelopeBuilder.sourceDevice = senderDeviceId.unsignedIntValue;
+                NSError *identifiedEnvelopeBuilderError;
+
+                identifiedEnvelope = [identifiedEnvelopeBuilder buildAndReturnError:&identifiedEnvelopeBuilderError];
```
charlesmchen-signal: I don't see any checking of `identifiedEnvelopeBuilderError`.  At the very least we should log it if it's non-nil.

michaelkirk-signal: Note that we *do* `OWSAssert(identifiedEnvelope)`. 

It would probably be a bug if we returned both an error and a result, but as I understand it, the recommended approach is to only consult the error if and only if the result was not successful.

But, I'm happy to add the logging you've requested.

PR #4050, opened 2018-11-19
Avoid spurious crash reporting

michaelkirk-signal:
PTAL @charlesmchen-signal

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -306,7 +306,8 @@ static NSTimeInterval launchStartedAt;
         [[UIApplication sharedApplication] scheduleLocalNotification:notification];
         [UIApplication.sharedApplication setApplicationIconBadgeNumber:1];
 
-        OWSFail(@"!isDatabasePasswordAccessible.");
+        [DDLog flushLog];
+        exit(0);
     }
 }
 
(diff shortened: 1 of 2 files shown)


Signal/src/AppDelegate.m
```diff
@@ -306,7 +306,8 @@ - (void)verifyDBKeysAvailableBeforeBackgroundLaunch
         [[UIApplication sharedApplication] scheduleLocalNotification:notification];
         [UIApplication.sharedApplication setApplicationIconBadgeNumber:1];
 
-        OWSFail(@"!isDatabasePasswordAccessible.");
+        [DDLog flushLog];
```
charlesmchen-signal: Shouldn't we log something before we exit?

Signal/src/environment/SignalApp.m
```diff
@@ -140,7 +140,7 @@ + (void)resetAppData
     [self clearAllNotifications];
 
     [DebugLogger.sharedLogger wipeLogs];
-    OWSFail(@"App data reset.");
```
charlesmchen-signal: Ditto.

charlesmchen-signal: Well, maybe not here.

Signal/src/AppDelegate.m
michaelkirk-signal: Existing logging is up a couple lines, we just want to make sure it (and anything else) flushes before exiting.

Signal/src/environment/SignalApp.m
michaelkirk-signal: see line 133 above

PR #4052, opened 2018-11-19
Fix crash around account attributes update when unregistered.

charlesmchen-signal:
PTAL @michaelkirk


SignalServiceKit/src/Account/TSAccountManager.m
```diff
@@ -649,6 +649,10 @@ - (AnyPromise *)updateAccountAttributes {
 }
 
 - (AnyPromise *)updateAccountAttributesIfNecessary {
+    if (!self.isRegistered) {
+        return [AnyPromise promiseWithValue:@(1)];
+    }
```
charlesmchen-signal: Only update account attributes if we are registered.

* I've verified that the other places we do account attributes updates already have this check.
* We're not "skipping" any necessary account attributes update due to this change.  If we're unregistered, we'll update account attributes when we register.

PR #4077, opened 2018-12-07
fix settings accessibility

michaelkirk-signal:
PTAL @charlesmchen 

fixes https://github.com/signalapp/Signal-iOS/issues/4072

### DIFF
--- a/Signal/src/ViewControllers/HomeView/HomeViewController.m
+++ b/Signal/src/ViewControllers/HomeView/HomeViewController.m
@@ -504,14 +504,16 @@ NSString *const kArchivedConversationsReuseIdentifier = @"kArchivedConversations
         UIImage *avatarImage = (localProfileAvatarImage
                 ?: [[[OWSContactAvatarBuilder alloc] initForLocalUserWithDiameter:kAvatarSize] buildDefaultImage]);
         OWSAssertDebug(avatarImage);
-        AvatarImageView *avatarView = [[AvatarImageView alloc] initWithImage:avatarImage];
-        [avatarView autoSetDimension:ALDimensionWidth toSize:kAvatarSize];
-        [avatarView autoSetDimension:ALDimensionHeight toSize:kAvatarSize];
-        avatarView.userInteractionEnabled = YES;
-        [avatarView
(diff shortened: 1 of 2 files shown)


Signal/src/ViewControllers/HomeView/HomeViewController.m
```diff
@@ -504,14 +504,16 @@ - (void)updateBarButtonItems
         UIImage *avatarImage = (localProfileAvatarImage
                 ?: [[[OWSContactAvatarBuilder alloc] initForLocalUserWithDiameter:kAvatarSize] buildDefaultImage]);
         OWSAssertDebug(avatarImage);
-        AvatarImageView *avatarView = [[AvatarImageView alloc] initWithImage:avatarImage];
```
michaelkirk-signal: I'm not entirely sure why, but when voiceover is activated, you cannot select a custom non-button view, even though it's "tappable". 

Using a UIButton instead works.

SignalMessaging/Views/AvatarImageView.swift
```diff
@@ -178,3 +178,59 @@ public class ConversationAvatarImageView: AvatarImageView {
...
+        // This can be any value large enough to cast a sufficiently large shadow.
+        let shadowInset: CGFloat = -3
+        shadowPath.append(UIBezierPath(rect: shadowBounds.insetBy(dx: shadowInset, dy: shadowInset)))
+        // This can be any color since the fill should be clipped.
+        shadowLayer.fillColor = UIColor.black.cgColor
+        shadowLayer.path = shadowPath.cgPath
+        shadowLayer.fillRule = kCAFillRuleEvenOdd
+        shadowLayer.shadowColor = (Theme.isDarkThemeEnabled ? UIColor.white : UIColor.black).cgColor
+        shadowLayer.shadowRadius = 0.5
+        shadowLayer.shadowOpacity = 0.15
```
charlesmchen-signal: Should the radius value reflect the size of the avatar?

charlesmchen-signal: Hmm, I guess not.

PR #4096, opened 2019-01-05
Convert to just using CocoaPods for dependencies

orta:

- [x] I have signed the [Contributor Licence Agreement](https://whispersystems.org/cla/) ( I did this a few years ago )

### Contributor checklist

 * Simulator

- - - - - - - - - -

### Description

Removes Carthage as a dependency manager, it looks like it was still around for legacy reasons and was only being used to handle a single dependency that was available on CocoaPods. This simplifies the setup process, removes a build phase script and requires less maintenance.

I also took a quick look over the Podfile and simplified that a bit by using the [implicit `abstract_target`](https://guides.cocoapods.org/using/the-podfile.html#what-is-a-podfile). 

Fixes https://github.com/signalapp/Signal-iOS/issues/1471


orta:
You could use CocoaPods Rome, https://github.com/CocoaPods/Rome to pre-generate the frameworks in the same way and manually link the frameworks (like how it works with WebRTC) - it'd probably require adding a Podfile to a subfolder and compiling etc in there.

I've always wondered why people see these kind of issues but we don't at Artsy, my guess is it's because projects like this do a lot in work in dev pods and I wonder if https://github.com/CocoaPods/CocoaPods/issues/8253 would allow scoping those to their own project meaning the more static pods would live in their own.

orta:
Interesting, looks like there's this too: https://github.com/leavez/cocoapods-binary

orta:
Alright, I've updated this PR to use cocoapods-binary which automatically inlines a static version of your library for XZing. I added some docs on it on the CocoaPods guides: https://guides.cocoapods.org/plugins/pre-compiling-dependencies.html

michaelkirk-signal:
> Alright, I've updated this PR to use cocoapods-binary which automatically inlines a static version of your library for XZing. I added some docs on it on the CocoaPods guides: https://guides.cocoapods.org/plugins/pre-compiling-dependencies.html

Just wanted to say thanks for looking into this - I haven't forgotten about it, but wanted to do some more internal testing with it.

Working to have less setup steps is definitely appreciated.

breznak:
Bump. This looks like a great PR, is it going to be mainlined?

Imperiopolis-Signal:
Thanks for your contribution @orta! Sorry it has taken us so long to circle back to this, but we have integrated this into our release/3.2.0 branch here 8f45f579c0ec29d1da951ee24547ba9a8c427a4a

PR #4113, opened 2019-01-18
Add accessibility identifiers to registration view.

charlesmchen-signal:
PTAL @michaelkirk


Imperiopolis-Signal:

PR #4122, opened 2019-02-01
Update Localizable.strings

PentagramTr:
Fixed some incorrect translations.


michaelkirk-signal:
Hiya @PentagramTr. Thanks for the PR, but we input translations in transifex, not in GitHub.

Could you sign up and make these changes there?

https://www.transifex.com/signalapp/signal-ios

PR #4212, opened 2019-06-21
Only escape % signs in notifications on iOS 9

### LINKED ISSUE #4195: Push notifications for messages with % character are rendered with %%

### Bug description
When I receive a Signal message that contains a `%` character the push notification renders it as `%%`. 

#### For Example
I got this message: 
message in signal
But the push notification I saw was: 
signal push iOS notification

I'm pretty sure it stems from [line 276 of DisplayableText.swift](https://github.com/signalapp/Signal-iOS/blob/0b638f4/SignalMessaging/utils/DisplayableText.swift#L276), a change that was made in #2996. I didn't dig any deeper but the reference to [Apple's documentation of alertBody](https://developer.apple.com/documentation/uikit/uilocalnotification/1616646-alertbody) indicates that it was deprecated at some point around iOS 10—is a newer API being used on devices running a newer version of iOS?

### Steps to reproduce
- receive a message containing a `%` without the chat open

**Actual result:** observe that the push notification has `%%` where `%` should be

**Expected result:** the push notification exactly matches the original message

### Device info
**Device**: iPhone XS

**iOS version**: 12.3

**Signal version:** 2.39.0.14

### Link to debug log
**n/a** (but I can produce one if requested)

mateoeh:
Fixes #4195

### Contributor checklist

 * iPhone X, iOS 12.3.1

- - - - - - - - - -

### Description
`UILocalNotification` (used on iOS 9) requires literal percent signs to be escaped, while `UNUserNotifications` (used on 10+) does not. Don't have an iOS 9 device to test on, but behavior there should be unchanged

### DIFF
--- a/Signal/src/UserInterface/Notifications/AppNotifications.swift
+++ b/Signal/src/UserInterface/Notifications/AppNotifications.swift
@@ -352,12 +352,7 @@ public class NotificationPresenter: NSObject, NotificationsProtocol {
         // While batch processing, some of the necessary changes have not been commited.
         let rawMessageText = incomingMessage.previewText(with: transaction)
 
-        // iOS strips anything that looks like a printf formatting character from
-        // the notification body, so if we want to dispay a literal "%" in a notification
-        // it must be escaped.
-        // see https://developer.apple.com/documentation/uikit/uilocalnotification/1616646-alertbody
-        // for more details.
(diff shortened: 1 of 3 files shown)


michaelkirk-signal:
Thanks for this! We're going to include it in the next release 2.41.

PR #4237, opened 2019-08-09
Photolibrary performance

### LINKED ISSUE #4236: Image picker is very slow for large photo libraries

### Bug description

I have a photo library of ± 36000 items on my iPhone SE.
Trying to select a photo from my library often takes multiple seconds.

### Steps to reproduce
- go to a conversation
- tap the + button
- choose Photo Library

**Actual result:** 

- it now takes ±5s before an image can be selected
- if you want to select a second photo by tapping the "back" button or the + button at the bottom, it again takes a long time.

**Expected result:** 

Apps using Apple's standard UIImagePickerController work pretty much instantly on my device.

### Screenshots

If needed, I have a screen recording of me trying to send an image as fast as possible. Between selecting "Photo Library" and landing back in the conversation, there's ±16 seconds.

### Device info
**Device**: iPhone SE
**iOS version**:  iOS 12.4
**Signal version:** 2.41.0

lieven:

### Contributor checklist

 * iPhone SE, iOS 12.4

- - - - - - - - - -

### Description

Fixes #4236 by:
- not doing photo library fetch operations for determining if a collection is empty
- not doing fetch operations in advance for determining if an asset is selected, but rather check it when it is being displayed

### DIFF
--- a/Signal/src/ViewControllers/Photos/ImagePickerController.swift
+++ b/Signal/src/ViewControllers/Photos/ImagePickerController.swift
@@ -225,7 +225,7 @@ class ImagePickerGridController: UICollectionViewController, PhotoLibraryDelegat
         let cellSize = collectionViewFlowLayout.itemSize
         photoMediaSize.thumbnailSize = CGSize(width: cellSize.width * scale, height: cellSize.height * scale)
 
-        reloadDataAndRestoreSelection()
+        reloadData()
         if !hasEverAppeared {
             scrollToBottom(animated: false)
         }
@@ -318,7 +318,7 @@ class ImagePickerGridController: UICollectionViewController, PhotoLibraryDelegat
         }
     }
 
-    public func reloadDataAndRestoreSelection() {
+    public func reloadData() {
         guard let collectionView = collectionView else {
             owsFailDebug("Missing collectionView.")
             return
@@ -409,7 +409,7 @@ class ImagePickerGridController: UICollectionViewController, PhotoLibraryDelegat
 
     func photoLibraryDidChange(_ photoLibrary: PhotoLibrary) {
         photoCollectionContents = photoCollection.contents()
-        reloadDataAndRestoreSelection()
+        reloadData()
     }
 
     // MARK: - PhotoCollectionPicker Presentation
@@ -476,7 +476,7 @@ class ImagePickerGridController: UICollectionViewController, PhotoLibraryDelegat
         photoCollectionContents = photoCollection.contents()
 
         // Any selections are invalid as they refer to indices in a different collection
-        reloadDataAndRestoreSelection()
+        reloadData()
 
         titleView.text = photoCollection.localizedTitle()
 
--- a/Signal/src/ViewControllers/Photos/ImagePickerController.swift
+++ b/Signal/src/ViewControllers/Photos/ImagePickerController.swift
@@ -55,7 +55,7 @@ class ImagePickerGridController: UICollectionViewController, PhotoLibraryDelegat
             owsFailDebug("collectionView was unexpectedly nil")
             return
         }
-
+        clearsSelectionOnViewWillAppear = false
         collectionView.register(PhotoGridViewCell.self, forCellWithReuseIdentifier: PhotoGridViewCell.reuseIdentifier)
 
         // ensure images at the end of the list can be scrolled above the bottom buttons
@@ -332,15 +332,6 @@ class ImagePickerGridController: UICollectionViewController, PhotoLibraryDelegat
 
         collectionView.reloadData()
         collectionView.layoutIfNeeded()
-
-        let count = photoCollectionContents.assetCount
-        for index in 0..<count {
-            let asset = photoCollectionContents.asset(at: index)
-            if delegate.imagePicker(self, isAssetSelected: asset) {
-                collectionView.selectItem(at: IndexPath(row: index, section: 0),
-                                          animated: false, scrollPosition: [])
-            }
-        }
     }
 
     // MARK: - Actions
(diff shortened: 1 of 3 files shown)


Signal/src/ViewControllers/Photos/ImagePickerController.swift
```diff
@@ -325,22 +324,8 @@ class ImagePickerGridController: UICollectionViewController, PhotoLibraryDelegat
             return
         }
 
-        guard let delegate = delegate else {
-            owsFailDebug("delegate was unexpectedly nil")
-            return
-        }
-
         collectionView.reloadData()
         collectionView.layoutIfNeeded()
```
michaelkirk-signal: The name of this method should be updated to reflect it's new behavior.

Signal/src/ViewControllers/Photos/PhotoLibrary.swift
```diff
@@ -244,6 +244,10 @@ class PhotoCollection {
 
         return PhotoCollectionContents(fetchResult: fetchResult, localizedTitle: localizedTitle())
     }
+    
+    var isEmpty: Bool {
+    	return collection.estimatedAssetCount == 0
```
michaelkirk-signal: It's not clear to me that `estimatedAssetCount` is guaranteed to be accurate. What's your rationale?

lieven: According to the documentation, it's indeed not guaranteed to always be up-to-date or it might not always be loaded in time:

> This count may not match the current number of assets in the collection. To get the most recent count, fetch the collection’s assets with the fetchAssets(in:options:) method and read the count property of the fetch result.
If asset count information is not available for the collection, this property’s value is NSNotFound

False positives wil hopefully self-correct in time, and false negatives might display e.g. an empty Favourites collection. It's not perfect, but it seemed like a good trade-off.

I tried using `fetchAssets(in:options:)` with a `limit` of `1`, but that was as slow as without limit.

The factor slowing this down is actually the number of albums I have in my Photo Library (automatically converted from iPhoto's Events)

michaelkirk-signal: To make sure that I understand, the implication is that we might (rarely) show an empty collection in the album list. Is that right?

Have you tested what happens when the user gets in that state? (e.g. change the code to list empty collections). Can you send a screenshot of what the user sees in that case? And what happens when the users taps on that collection?

If all that's reasonable, then I'd agree that's a preferable exchange for a much snappier load time.

One small thing, apparently `estimatedAssetCount` is one of the things we can filter by in our query. No biggie, but it seems like it might be more to the point to just filter on that in our query rather than checking each one. What do you think?

https://developer.apple.com/documentation/photokit/phfetchoptions

lieven: > To make sure that I understand, the implication is that we might (rarely) show an empty collection in the album list. Is that right?

Yes, and right after favouriting your first photo ever, Favourites might still be hidden until `estimatedAssetCount` catches up (same for shortly after adding the first photo to an empty album)

> Have you tested what happens when the user gets in that state? (e.g. change the code to list empty collections). Can you send a screenshot of what the user sees in that case?

IMG_7771


> And what happens when the users taps on that collection?

Nothing happens, the currently selected photos remain the same.

 
> If all that's reasonable, then I'd agree that's a preferable exchange for a much snappier load time.

I think so too, but to lower the risks even more, you could not filter out the empty collections and add an empty state.

> One small thing, apparently `estimatedAssetCount` is one of the things we can filter by in our query. No biggie, but it seems like it might be more to the point to just filter on that in our query rather than checking each one. What do you think?
> 
> https://developer.apple.com/documentation/photokit/phfetchoptions

That's a possibility indeed, but I'd rather keep my changes as limited as possible, since this is my first PR ;-)

lieven:
> Other than my inline feedback, some edge cases to test would be navigating back and forth between the approval screen, the media library view, and the camera capture view while adding and removing items, both from the media library and from the approval screen - ensuring that what is sent is ultimately accurately reflected in the UI.

Did some more monkey testing and didn't manage to break it. Is there a workflow for having other people test bugfixes some more before they're merged in?

Signal/src/ViewControllers/Photos/PhotoLibrary.swift
Imperiopolis-Signal: I played around with this a bit and it seems like `estimatedAssetCount` always returns `NSNotFound` for smart albums, which means things like `time lapse`, `favorites`, etc. will always show up even if they're empty. I'm not sure if there is a good work around for this, but it'd be nice to investigate it a little further.

lieven: Aha, thanks for testing. I'll try to find a better solution then when I'm back from my holidays.

Imperiopolis-Signal: Awesome! Thanks so much for your contributions. I've incorporated all of your work, other than the `estimatedAssetCount` change (6ecc8c33f98204591d11628cfa65771af16deb91), into the upcoming 2.42 release.

michaelkirk-signal:
Hey @lieven - thanks for this PR. 

We pulled some of these changes into the 2.42.0 release with 67fa45ef9.

Would you mind rebasing?

AbroMansoor:
it is not good, we need that when we select different photos we also need different option like to SAVE PHOTOS, FILES and other sending option it just give us option to delete Picture is very difficult for us to manage our Media in Signal App

Imperiopolis-Signal:
Hey @lieven – since several months have passed since we integrated a portion of this PR, I'm going to close this out. Feel free to rebase and open an new PR with further changes in the future. Thanks for your contribution!

PR #4424, opened 2020-04-05
Fix 'guard' body fall through when contacts are unexpectedly not set after a fetch

golivera:

### Contributor checklist

- - - - - - - - - -

### Description
This pull request removes a compiler error in `SignalMessaging` that is caused by allowing a `guard` body to fall through, instead of returning or throwing an error.

The fix is simply to add a `return` to follow the same completion/return pattern found in the file.

I don't believe that this issue will be encountered in production. In the switch case of `fetchContacts` returning `.success`, the payload will at least be an empty array and in the case of `.error` we successfully `return` before the `guard`.


Imperiopolis-Signal:
Thanks, we have already resolved this internally and it will be present on master soon.

Issue #4459, opened 2020-05-22
Share Link from any 3rd app


frdonn:
When I want to share a link from a third app (example: YouTube, go to a Chan, ... menu, share, more, select Signal)
It try to go to Signal but it comes back to YouTube app.

Genis-Sage:
hey @frdonn 

could you fill out the issue template, and give more information:

- [ ] I have searched open and closed issues for duplicates
- [ ] I am submitting a bug report for existing functionality that does not work as intended
- [ ] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
Describe here the issue that you are experiencing.

### Steps to reproduce
- using hyphens as bullet points
- list the steps
- that reproduce the bug

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)

**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)

### Screenshots



### Device info

**Device**: iDevice X

**iOS version**: X.Y.Z

**Signal version:** Z.Y

### Link to debug log

frdonn:
Hi,

Please find the form filled up.

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

Issue #4316, opened 2019-11-29
Share Sheet screen unresponsive


TempSpas:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
Signal's Share Sheet is unresponsive to input and scrolling within it is laggy, rendering the feature nonfunctional.

### Steps to reproduce
- Open an app that can use Signal's Share Sheet (e.g. Photos)
- Trigger Signal's Share Sheet
- Tap one of the "Recent Chats" and observe the app does not respond to the input
- Attempt to scroll on the screen and observe laggy scrolling and flickering

**Actual result:** Share Sheet functionality is broken.

**Expected result:** Tapping on the recent chats should immediately bring up Signal's message compose screen. Scrolling should be smooth and not cause flickering.

### Device info

**Device**: iPhone XR

**iOS version**: 13.2.3

**Signal version:** 3.0.2.1

### Link to debug log

[Debug log.](https://debuglogs.org/ff6923448caacee49274573a75d44c579240e579c3de2938f487818fe00b92e0.zip)

I reproduced the bug multiple times – repeatedly tapping a chat in order to trigger a share without any response from the app – so hopefully this is helpful.

agottardo:
I am experiencing the same issue on 3.0.2.1 on an iPhone XS running iOS 13.2.3. I’d like to add that tapping repeatedly on the contact cell eventually fires the selection handler.

esstroud:
I'm experiencing a similar issue with the share extension. When triggered, the Signal share sheet loads contacts and then crashes. 

### Device info
- iPhone 7 Plus
- Signal app v 3.0.4
- iOS v. 13.2.3

### Debug log
[Debug log](https://debuglogs.org/8070145efc3d06d8a5971b7c26e470b8a10eb79878fbbdc484062bbdca66f78f.zip
)

denver-s:
If you are experiencing a crash, you should send a crash report.

https://support.signal.org/hc/en-us/articles/360007318591-Debug-Logs-and-Crash-Reports

> **Steps to share a crash report**
> iOS 11 & 12
> 
> Go to iOS Settings > Privacy > Analytics > Analytics Data.
> Locate and select the latest Signal-(DateTime).ips file.
> Tap Share in the top right.
> Send the email to support@signal.org.

TempSpas:
[...]

Issue #4435, opened 2020-04-13
iPadOS share sheet « send » button missing in landscape 


tutis:
- [X] I have searched open and closed issues for duplicates
- [X] I am submitting a bug report for existing functionality that does not work as intended
- [X] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
When using the share sheet to signal from any app with an iPad running iPad OS 13 in landscape mode the UI is broken and the send button unavailable 

### Steps to reproduce
-  open any webpage in Safari (or any other iPadOS app)
- turn ipad to landscape 
- open share sheet to Signal

**Actual result:** No send button is available, message cannot be sent

**Expected result:** send button should be available, as when the ipad is turned in portrait mode

### Screenshots

![IMG_0726](https://user-images.githubusercontent.com/5681030/79107660-19758800-7d75-11ea-8573-a9c808049ea7.jpeg)

### Device info

**Device**: iPad Pro 9.7

**iOS version**: 13.4.1

**Signal version:** 3.7.1.0

XxBlackHatxX:
I’ve been having this issue on my iPad Pro 2018 for a while as well. It doesn’t matter if it is landscape or portrait. Bluetooth keyboard or virtual. Mine never shows the send button or the text field. This even happens on my iPhone sometimes.

ventz:
I am very surprised and saddened at how long this has been an issue, and how many github issue tickets/support emails there have been around it.

It seems like a very significant/critical app problem that really should be prioritized in some way over "miscellaneous" and "fun features".

Here are a few more screenshots showing the problem in landscape mode (the way most users use ipads):

![IMG_0006](https://user-images.githubusercontent.com/901168/79898854-ecab1b80-83d9-11ea-8f15-3e7f89d59998.jpg)

and

![IMG_0003](https://user-images.githubusercontent.com/901168/79898849-eb79ee80-83d9-11ea-979f-1ee4a69ccc6c.jpg)



vs, here is a "working" version of it in portrait mode:

![IMG_0004](https://user-images.githubusercontent.com/901168/79898953-0e0c0780-83da-11ea-9d66-807993f518e2.PNG)


ping @Imperiopolis-Signal @charlesmchen-signal @michaelkirk-signal

ventz:
A list of other threads:

https://github.com/signalapp/Signal-iOS/issues/4392
and
https://github.com/signalapp/Signal-iOS/issues/4345
and
[...]

Issue #4392, opened 2020-02-24
[BUG] Sharing in landscape mode


Mihahn:
- [ x] I have searched open and closed issues for duplicates
- [ x] I am submitting a bug report for existing functionality that does not work as intended
- [ x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
When I try to share a link/an image on a linked iPad (same problem on iPhone) the layout glitches and I can't open the keyboard.

<img width="1024" alt="signal-2020-02-24-182328" src="https://user-images.githubusercontent.com/3447340/75175624-ff370a80-5732-11ea-8529-3f93d5d82f39.png">

The issue may be related to #4391 .
### Steps to reproduce
Scenario A (link)
- Open any app (e.g. news)
- Click the share icon and select Signal
- Select a contact (often you have to hit it 3-5 times until it gets selected, see #4376 )

**Actual result:** The layout glitches and when you try to open the keyboard to add some message to your link, it doesn't come up or is off screen.

**Expected result:** The layout shouldn't glitch and a keyboard should come up.

Scenario B (photo)
- Open photos app
- Click the share icon and select Signal
- Select a contact

**Actual result:** The layout doesn't glitch, but when you try to open the keyboard to add some message to your photo, it doesn't come up or is off screen.

**Expected result:** The keyboard should come up.

### Device info
**Device**: iPhone 7
**iOS version**: 13.3.1
**Signal version:** 3.5.0.10

TempSpas:
> Select a contact (often you have to hit it 3-5 times until it gets selected, see #4376 )

Documented with issue #4316 as well

igorkulman:
Same problem on iPad Air 2 and probably all the iPad (which people probably mostly use in landscape).

The share button is not even visible breaking the sharing functionality completely.

![IMG_F3073364E8FD-1](https://user-images.githubusercontent.com/4746070/78356202-64daaa00-75af-11ea-8746-0845b751aab1.jpeg)

scottcarterco:
It appears that this issue is especially persistent in Landscape mode when the keyboard option of “Shortcuts” is turned off. I wonder if it normally anchors above the shortcuts on-screen bar and when it’s disabled, it has nothing to anchor to and disappears.

ch1phead:
Just experienced the same issue as Scenario B. More specifically I can’t even send it, before adding text.  Rotating the device to Portrait makes the edit/send bar show up again.

ignaspa:
[...]

PR #4500, opened 2020-07-08
Fixes the issue around cleaning up the share extension 

fumiakiy:
### Contributor checklist

 * iPhone XR, iOS 12.4
 * iPhone SE 2nd gen, iOS 13.5

- - - - - - - - - -

### Description

This PR attempts to fix an issue found both in iOS 12 and in iOS 13. The symptom is the same between the iOS versions but the cause was different. The first commit fixes the issue in iOS 12 while the second commit is specifically to fix it in iOS 13.

#### iOS 12
1. Open Safari, Map or any app that shares a text
1. Start sharing by tapping the share icon
1. Tap the Share icon
1. Select Signal
1. Select a thread to share the photo
1. On the attachment approval view, tap [Cancel], or tap Send button to send it
1. In the share sheet, select Signal again

**Actual behavior:**

Nothing happens. Behind it, the app context swaps, that ultimately calls the previous instance's `deinit`, that kills the process.

Either canceling or sending at the step 6 yields the same result.

Note that if you tap on [X] at the step 5 above, it behaves as expected.

Also note that if you share images from Photo app or any imaging apps, it behaves as expected too.

**Expected behavior:**

Signal share extension opens again. Behind it, at the tap of [Cancel], the process should exit (by calling `ExitShareExtension`).

#### iOS 13
The steps are the same, and the actual behaviors are also the same for the buggy behaviors.

In addition to those, tapping [X] at the step 5 yields the buggy behavior as described in *Actual behavior* above.

Also in iOS 13, image sharing from Photo app yields the same buggy behavior as described in *Actual behavior* above.

----

The PR _may_ fix the issues reported in #4459 and #4316 , but the description of those issues do not exactly match the steps above.


Imperiopolis-Signal:
I haven't had a chance to review yet, but I will note that #4316 is a different issue that only effects iPad in landscape.

fumiakiy:
> but I will note that #4316 is a different issue that only effects iPad in landscape.

I was able to reproduce the similar (though not exactly the same) issue described in #4316 in my iPhone. The iPad landscape issue you mentioned is #4435 and #4392 I guess? Those issues are not related to this PR.

Imperiopolis-Signal:
Yep, you're right I meant to flag #4435 and #4392 as the ones that were iPad only, but at least with iOS 14 they happen on phones as well.

Imperiopolis-Signal:
In general the changes here look OK, we'll probably get them incorporated in a future release.

fumiakiy:
The gist of the pull request was implemented in [c073d8e204].

I found that in iOS 14 even `viewWillDisappear` isn't called at least in the simulator, so the issue raised in this PR still persists in iOS 14. But I was not able to find a solution for the issue in iOS 14 and the issue in the previous versions have been fixed by the commit.

Closing.

Issue #4393, opened 2020-02-24
"New message" shortcut doesn't open compose screen if app was fully closed


u32i64:
If the app was e.g. force closed from the app switcher, the "New message" app icon shortcut (the same that #4288 is about) doesn't open compose screen, instead it just shows the main app screen (conversation list).

### Steps to reproduce
- Make sure Signal is fully closed (via app switcher);
- Haptic touch on Signal icon;
- Choose *New message*.

**Actual result:** Conversation list is shown.
**Expected result:** The "compose" / "choose contact to message" screen appears.

### Device info
**Device**: iPhone SE
**iOS version**: 13.4 Public beta 2
**Signal version:** 3.5.0.10

moritzluedtke:
Still happens on an iPhone XS running iOS 14.6 using Signal 5.14.0.35

u32i64:
This is still reproducible.

**Device:** iPhone XR
**iOS version:** 15.2.1
**Signal version:** `5.27.0.102`

u32i64:
This is still reproducible.

Device: iPhone XR
iOS version: 15.4
Signal version: 5.34.0.8

u32i64:
This is still reproducible.

**Device:** iPhone XR
**iOS version:** 16.0 Public Beta (20A5358a)
**Signal version:** 5.51.0.10

u32i64:
This is still reproducible.

**Device:** iPhone XR
**iOS version:** iOS 16.2 (20C5049e) public beta + iOS 16.2 (b) (20C7750490f)
**Signal version:** 6.4.0.8

EvanHahn-Signal:
Just reproduced this and filed a ticket in our internal tracker.

PadraigK:
I just tested this in version `6.50.1.0` and it seems to be behaving correctly now.

moritzluedtke:
I can confirm that on version `6.50.0.13`.

PR #4634, opened 2020-11-25
Fixes a crash happens when cold starting the app by a home screen quick action

fumiakiy:
### Contributor checklist

 * iPhone 11, iOS 13.5
 * iPhone 12, iOS 14.2

- - - - - - - - - -

### Description

When Signal is **not** even in the background of the device so the next launch will be a cold-start, using the home screen quick action i.e. long pressing the app icon and select `New Message` doesn't open the list of Signal Contacts popup.

When the PR was created in November 2020 the result of the home screen quick action is a crash. But after a few code updates the app seems to stop crashing as was described in #4393. But the bug is still there in that the block of code was not executed and the popup never appears as it should.

This was because of the order of execution in which the code assumes something should be there already when it is not in this case.

The diff may look large but really the only change was to wrap the existing code in a `dispatch_async` block.

### DIFF
--- a/Signal/src/AppDelegate.m
+++ b/Signal/src/AppDelegate.m
@@ -731,28 +731,38 @@ void uncaughtExceptionHandler(NSException *exception)
     }
 
     AppReadinessRunNowOrWhenAppDidBecomeReadySync(^{
-        if (![self.tsAccountManager isRegisteredAndReady]) {
-            ActionSheetController *controller = [[ActionSheetController alloc]
-                initWithTitle:NSLocalizedString(@"REGISTER_CONTACTS_WELCOME", nil)
-                      message:NSLocalizedString(@"REGISTRATION_RESTRICTED_MESSAGE", nil)];
+        // We need a root view controller set up before we run the block that presents a view controller.
+        // In most cases of launching the app, `ensureRootViewController` is called in `checkIfAppIsReady`
+        // or in some other ways.
+        // But in the case of the home screen quick action when the app is not running, in which case is
+        // handled here, it doesn't have a chance to call `ensureRootViewController` because this block is
+        // executed before `ensureRootViewController` in `setAppIsReady`.
+        // The workaround is to place the execution of the block after the call to `checkIfAppIsReady`,
+        // so the root view controller is set by then.
+        dispatch_async(dispatch_get_main_queue(), ^{
+            if (![self.tsAccountManager isRegisteredAndReady]) {
+                ActionSheetController *controller = [[ActionSheetController alloc]
+                    initWithTitle:NSLocalizedString(@"REGISTER_CONTACTS_WELCOME", nil)
+                          message:NSLocalizedString(@"REGISTRATION_RESTRICTED_MESSAGE", nil)];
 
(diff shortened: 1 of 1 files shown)


Signal/src/AppDelegate.m
```diff
@@ -731,28 +731,38 @@ - (void)application:(UIApplication *)application
...
-        }
+        // We need a root view controller set up before we run the block that presents a view controller.
+        // In most cases of launching the app, `ensureRootViewController` is called in `checkIfAppIsReady`
+        // or in some other ways.
+        // But in the case of the home screen quick action when the app is not running, in which case is
+        // handled here, it doesn't have a chance to call `ensureRootViewController` because this block is
+        // executed before `ensureRootViewController` in `setAppIsReady`.
+        // The workaround is to place the execution of the block after the call to `checkIfAppIsReady`,
+        // so the root view controller is set by then.
+        dispatch_async(dispatch_get_main_queue(), ^{
```
Imperiopolis-Signal, line 742: I think an additional async dispatch is not the cleanest way to handle this as it's basically just introducing a delay, if any timing changed the problem would be reintroduced. The best thing to do would likely be similar to how we handle URLs and intents where we store it, and then process them only when the UI is ready.

fumiakiy:
First of all, thank you for the review!

## Tl;dr;

After some deeper dive into code to find the reasons of the bug, I don't think an external dev like me could/should write the code needed to fix the issue "in a right way". I will see if I may be able to put together something that may be short and sweet enough for the dev team to consider. But for now I will close this PR.

## What I found

I did deeper dive in code on how these app-launching behaviors from external actions do in Signal and found that the only one that worked fine was `signal.me` link. I wasn't able to find "where we store it, and then process them only when the UI is ready" though.

- signal.me to a chat
- signal.group to a group
- sticker pack
- home screen quick action (what this PR was about)

Opening a `signal.group` or sticker pack url cold start the app, then briefly opens a modal sheet but the sheet abruptly closes to open the default chat list view. I wasn't able to see a view briefly opens and abruptly closing in doing Home Screen quick action but the behavior looks pretty much the same; it skips animation and just opens the default chat list view.

What was happening was that the app first starts `LoadingViewController`. After doing whatever it has to do, it switches the AppDelegate's root view controller to an appropriate one (ConversationSplitViewController, OnboardingNavigationController, ...). But the handler of each app-launching action has already been started to process by the time when the AppDelegate's root view controller switches. They have already been doing some UI works using LoadingViewController as their root view controller. Root view controller switches to something else eventually, that abruptly closes what has been shown (Group or sticker sheet, or the list of contacts in Home Screen quick action) and switches the view to the default ConversationSplitViewController.

Signal.me link works fine because the code to open the chat UI doesn't use the window's root view controller; instead it uses the split view controller that has already been loaded and stored in a property of SignalApp class.

I would fix these issues by changing how we declare "App is ready", basically by delaying to declare it after we switch root view controller to the "real" one, so that we can delay all actions during the time the root view controller is LoadingViewController. But it sounds like it could touch a lot of core functionalities of the app which I personally don't think an external dev like me should do.

I would appreciate if the dev team let me know anything that I may be missing or overlooking. For now I'll close the PR. Thank you again for the code review!

Issue #4427, opened 2020-04-09
Uploading from Files application to Signal deletes source file


temal-:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description

When sharing certain files via the Files application on iOS, the file is deleted after it is sent.

### Steps to reproduce

- Open Files application on iOS
- Select the file you want to send
- Click share icon
- Select Signal
- Select group, member or "note to self"
- Click send icon
- Close the file you just sent

**Actual result:** File is now deleted

**Expected result:** File is not deleted

### Screenshots

![signal](https://user-images.githubusercontent.com/5236825/78926575-9666e000-7a9d-11ea-8024-a73072762193.gif)

I cut out the parts where it would show personal data.

### Device info

**Device**: iPhone 11 Pro

**iOS version**: 13.4.1

**Signal version:** 3.7.1.0

### Link to debug log

https://debuglogs.org/9b3e5742bfd3fbaf246f806ea4871d59509aa6854811f2f5310adb76bdc04f86.zip

exchgr:
at the risk of spamming, i am also having this issue

fumiakiy:
I believe this issue has been resolved by now. See my comment in the PR for more information.

PR #4635, opened 2020-11-25
Check if it can be removed before moving a file upon sharing it

fumiakiy:
### Contributor checklist

 * iPhone 11, iOS 13.5
 * iPhone 12, iOS 14.2

- - - - - - - - - -

### Description

When sharing files that are *not* images or movies, the app *moves* the file to a destination when it sends the file to the contact of the chat. It effectively removes the file from where it was before. It was originally reported in #4427 .

Note that after the PR was submitted, the commit [c799f663c807101111ebed0faf6ef06aa0dc1427] was made which effectively got rid of the problem described in the Github issue that was about a movie file. But the problem is still there for the files that are not encoded by Signal (thus to be copied in the temporary directory) even after the commit.

### Steps to verify the fix
1. Prepare a file that is not an image (bin or txt or something else)
1. Open Files app
1. Select the file prepared at the step 1
1. Share it to Signal
1. Choose a contact (or Note to Self) and send it
1. Go back to the Files app
1. You may still see the file there. Go back up one directory in the Files app
1. Come back to the original directory again

**Behavior before the fix**

The file is now gone.

**Behavior after the fix**

The file remains there.

### DIFF
--- a/SignalServiceKit/src/Util/DataSource.m
+++ b/SignalServiceKit/src/Util/DataSource.m
@@ -487,12 +487,16 @@ NS_ASSUME_NONNULL_BEGIN
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wblock-capture-autoreleasing"
                                self.isConsumed = YES;
-                               if ([[NSFileManager defaultManager] isWritableFileAtPath:self.fileUrl.path]) {
+                               // We generally prefer moving the file when the file is writable (thus removable) because
+                               // it is faster. But we should not move a file when the file is not in our container.
+                               // So we check if the file's path is under our temporary directory instead of checking
+                               // if the file is writable to decide whether we are going to move the file or copy the file.
+                               if ([self.fileUrl.path hasPrefix:OWSTemporaryDirectory()]) {
                                    success = [NSFileManager.defaultManager moveItemAtURL:self.fileUrl
                                                                                    toURL:dstUrl
                                                                                    error:error];
                                } else {
-                                   OWSLogError(@"File was not writeable. Copying instead of moving.");
+                                   OWSLogInfo(@"File was not in our temporary directory or not writeable. Copying instead of moving.");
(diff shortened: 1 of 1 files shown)


SignalServiceKit/src/Util/DataSource.m
```diff
@@ -487,12 +487,16 @@ - (BOOL)moveToUrlAndConsume:(NSURL *)dstUrl error:(NSError **)error;
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wblock-capture-autoreleasing"
                                self.isConsumed = YES;
-                               if ([[NSFileManager defaultManager] isWritableFileAtPath:self.fileUrl.path]) {
+                               // We generally prefer moving the file when the file is writable (thus removable) because
+                               // it is faster. But we should not move a file when the file is not in our container.
+                               // So we check if the file's path is under our temporary directory instead of checking
+                               // if the file is writable to decide whether we are going to move the file or copy the file.
+                               if ([self.fileUrl.path hasPrefix:OWSTemporaryDirectory()]) {
```
Imperiopolis-Signal, line 494: I think we'll need to check if it's in our container too to be comprehensive, such as OWSFileSystem.appSharedDataDirectoryPath()

fumiakiy:
Thanks for the review! It took me more time than I thought it would but in the end it turned out that the issue that this PR was to fix was no longer relevant.

The issue was about movie files being deleted by Signal by moving the file instead of copying the file before wrapping it as an attachment. Since the PR was submitted though, the movie files would now be encoded, thus effectively copied first to the temporary directory, that makes the original issue for which the PR was created, would never be reproduced any more.

Other types of files such as text files would typically be picked up by UIDocumentPickerViewController, but the URL it gives Signal is already scoped so the file contents are copied from the original location. So those types of files won't be abruptly deleted upon sending them as an attachment either.

Issue #4546, opened 2020-09-05
Image artifacts next to emoji response


211217613:
- [ ] I have searched open and closed issues for duplicates
- [ ] I am submitting a bug report for existing functionality that does not work as intended
- [ ] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
Oval image artifacts next to emoji responses.

### Screenshots

![725759C7-11EF-4F91-AE6F-FC9E60A168DC](https://user-images.githubusercontent.com/7006850/92311580-1457e600-ef6d-11ea-9662-0bf1b20618c9.JPG)

### Device info

**Device**: iPhone 8

**iOS version**: 14b7

**Signal version:** 3.16.0.13

The artifacts go away on screen refresh or when I scroll the emoji out of view then back into view.

PR #4636, opened 2020-11-25
Remove a strange artifact that appears to the right of a reaction to a message

fumiakiy:
### Contributor checklist

 * iPhone 11, iOS 13.5
 * iPhone 12, iOS 12.4

- - - - - - - - - -

### Description

The bug was originally reported in #4546 .

This strange behavior is was due to a bug in `UIStackView` that behaves strangely when its subviews are hidden and then shown. There are a few other places that has already handled this bug by applying `alpha` value to the hidden view. The code of the PR is to apply the trick to the reaction views too.

Note that a similar issue is fixed in #4637 but in a different way; this PR uses the `hide by alpha=0` trick. It works for this PR because it does not involve stretching subviews. For the other PR it had to remove and insert subviews because it involves stretching subviews in the `UIStackView`.

### Steps to verify the fix of the issue
It's probably easier to see the gif below to understand the steps.
1. Have a chat long enough so you must scroll to reach the bottom of it
1. Add an emoji reaction to a message that you receive, that is a few messages before the bottom of the chat
1. Scroll up the chat so the message that has the reaction emoji is the next one to appear on screen
1. Go back to the conversation list screen
1. Come back to the chat again
1. Scroll down to the bottom

**Actual behavior before the fix**

A strange artifact appears to the right of the emoji reaction.

**Expected behavior after the fix**

The artifact does not appear.

### Screenshots

|Before|After|
|---|---|
|artifact-bug|artifact-fixed|
|artifact-bug-ios12|artifact-fixed-ios12|


fumiakiy:
This PR is no longer relevant because the files touched in the PR are removed by the commit [111338b59dea9d9604441b45f1ec9236cfe6f27b]. The issue the PR was resolving seems to be fixed by the commit too.

Closing.

Issue #4847, opened 2021-01-28
Contacts are sorted by nickname instead of considering the system preferences


christianapel:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
The contacts on my phone are sorted by last name and Signal is using the same basic sort order. Some of my contacts have a nickname in the iOS contact list, and this is where the sort order is different.

While Signal is sorting contacts with nickname into the bucket of the nicknames first character, the iOS contacts list doesn't consider the nickname. In the iOS contacts list all names are sorted by last name, no matter if s nickname is configured or not.

### Steps to reproduce
- Configure the iOS contacts to be sorted by "last name, first name"
- Add contacts with nicknames having a different first character than their last name

**Actual result:**
Contacts with nicknames are not sorted by their last name, but by their nickname instead.

**Expected result:** 
Contacts should always be sorted as it's configured for the iOS contacts.

### Device info

**Device**: iPhone 11 Pro

**iOS version**: 14.4

**Signal version:** 5.3.1.0

sbilling:
@christianapel the sorting is a function of the OS. The device favors the nickname in the CNContact object returned from fetching the device contacts.

christianapel:
> @christianapel the sorting is a function of the OS. The device favors the nickname in the CNContact object returned from fetching the device contacts. 

In that case I'm wondering why the sort order in the iOS contacts list isn't considering the nickname. Beside that, Signal is the first app which is using my contacts and is sorting like that.

Which function is used to sort the contacts? In CNContacts, the following enum is referenced which doesn't contain the nickname:
[CNContactSortOrder](https://developer.apple.com/documentation/contacts/cncontactsortorder)

sbilling:
@christianapel after further investigation it looks like we are using `.userDefault` as the CNContactSortOrder. I don't see anything that would point to the nickname playing a role in that so this is either an OS bug or I am missing something. I notice a bunch of weird behavior with the "nickname" attribute within the Signal app - sometimes it shows as the name, sometimes it doesn't. Very odd.

farovitus:
[...]

PR #4848, opened 2021-01-28
Fix contact editing bug / replace use of hashvalue with equatable

sbilling:
- [ x] I have read the [README](https://github.com/WhisperSystems/Signal-iOS/blob/master/README.md) and [CONTRIBUTING](https://github.com/WhisperSystems/Signal-iOS/blob/master/CONTRIBUTING.md) documents
- [x ] I have signed the [Contributor Licence Agreement](https://whispersystems.org/cla/)

### Contributor checklist
- [ x] I'm following the [code, UI and style conventions](https://github.com/WhisperSystems/Signal-iOS/blob/master/CONTRIBUTING.md#code-conventions)
- [x ] My commits are rebased on the latest master branch
- [x ] My commits are in nice logical chunks
- [x ] My contribution is fully baked and is ready to be merged as is
- [x ] I have tested my contribution on these devices:
 * iPhone 11 Pro 14.0

- - - - - - - - - -

### Description
I discovered a bug in the edit contact info screen white investigating #4847 which occurs in the latest release: when editing contact info we were using the deprecated `hashValue` variable on two arrays to compare their likeness and update our views if the hash's were not equal. This is an insufficient way to compare two arrays as it is deprecated by Apple (see: https://developer.apple.com/documentation/swift/hashable/1540917-hashvalue), furthermore the hashValue was only hashing first, last and phone number(s) so if I edit the nickname, birthday, or any other field the hashes remain the same and the contact info is not updated. My fix leverages the equatable protocol which `Contact` conforms to allowing us to compare entire arrays of contacts between fetches.


sbilling (commented):
@charlesmchen-signal curious about your thoughts on this one. I see you did some interesting stuff with the hash in your recent work. Does that resolve the bug im describing? Not sure why we are hashing and not leveraging equatable?

Imperiopolis-Signal:
I believe this issue has been resolved in a slightly different way. Please let us know if you're still seeing issues.

PR #4966, opened 2021-04-22
Security Enhancement: Enable Certificate Transparency in Signal iOS app (RFC6962)

ajinabraham:

### Contributor checklist

 * Simulator iOS 14.4

- - - - - - - - - -

### Description

Certificate Transparency (CT) is a protocol that iOS App Transport Security can use to identify mistakenly or maliciously issued X.509 certificates. Set the value for the NSRequiresCertificateTransparency key to YES to require that for a given domain, server certificates are supported by valid, signed CT timestamps from at least two CT logs trusted by Apple. **Unlike most other ATS exceptions, using a non-default value in this case tightens security requirements.**

Reference: https://developer.apple.com/documentation/bundleresources/information_property_list/nsrequirescertificatetransparency

In addition to this, it will be great if signal maintainers can setup CT log alerts for signal domains like 
`signal.org` , `signalcaptchas.org` and `whispersystems.org`.

### DIFF
--- a/NotificationServiceExtension/Info.plist
+++ b/NotificationServiceExtension/Info.plist
@@ -37,16 +37,12 @@
 				<true/>
 				<key>NSIncludesSubdomains</key>
 				<true/>
-				<key>NSRequiresCertificateTransparency</key>
-				<true/>
 			</dict>
 			<key>whispersystems.org</key>
 			<dict>
 				<key>NSExceptionAllowsInsecureHTTPLoads</key>
 				<true/>
 				<key>NSIncludesSubdomains</key>
-				<true/>
-				<key>NSRequiresCertificateTransparency</key>
 				<true/>
 			</dict>
 			<key>signalcaptchas.org</key>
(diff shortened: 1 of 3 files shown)


u32i64:
I don't think this is needed, because Signal pins the needed certificate in the app (see the [Certifiably "F"ine](https://signal.org/blog/certifiably-fine/) blog post). Moreover, it's probably not even going to work because of this, as the certificate in question are *not trusted* by the certificate authorities (by design).

ajinabraham:
That makes sense. 

I was looking at
https://crt.sh/?q=whispersystems.org&exclude=expired
and https://crt.sh/?q=signal.org&exclude=expired 

and saw that CT logs are available for the domains. But it seems that certain subdomains that Signal uses like `https://textsecure-service.whispersystems.org`, `https://cdn.signal.org` etc. are on self signed root CA due to reasons mentioned in the link you shared and hence CT is not applicable.

However `https://signalcaptchas.org/registration/generate.html` is loaded in a WebView and appears to have Amazon Root CA. I will update my PR to only include that.

Imperiopolis-Signal:
Thank you for your contribution. We're not going to incorporate this for now, as we may change how certs are managed on this domain / how captchas in general work soon.

PR #5053, opened 2021-07-12
Allow sharing `.pkpass` files from Wallet and adding received passes to Wallet

u32i64:

### Contributor checklist

- [ ] I'm following the [code, UI and style conventions](https://github.com/WhisperSystems/Signal-iOS/blob/master/CONTRIBUTING.md#code-conventions) *(looks like that section was removed in https://github.com/signalapp/Signal-iOS/commit/5d312220ab7083a64a0f4c2d47e44566ab06bb7a)*
 * Simulator: iPhone 12, iOS 14.5

- - - - - - - - - -

### Description

This PR aims to implement the [Support Sharing of iOS Wallet Passes (PKPASS files)](https://community.signalusers.org/t/34811) feature request by:

1. Allowing to share `.pkpass` files from the Wallet app;
2. Presenting a `PKAddPassesViewController` when an attachment containing a pass is tapped,

While this PR already works, I have a few questions, mostly regarding some specifics of the implementation.

#### Finding the UTI for `.pkpass` files

<details><summary>Seems to be `com.apple.pkpass`</summary>

> The universal type identifier for `.pkpass` doesn't seem to be very clearly documented (and there doesn't seem to be a constant for it), but because of the recent addition of multi-pass downloads in `.pkpasses` files (which notably are not supported by this PR, since iOS 15 is still in beta and I don't think it's possible to share multiple passes from Wallet at a time), [Distributing and Updating a Pass](https://developer.apple.com/documentation/walletpasses/distributing_and_updating_a_pass) says the MIME type for `.pkpasses` is `application/vnd.apple.pkpasses`.
> 
> So it seems the `.pkpass`'s MIME type is `application/vnd.apple.pkpass` then. To get the UTI type, I ran the following code:
> 
> ```swift
> if #available(iOS 14, *) {
>     let mimeType = "application/vnd.apple.pkpass"
>     let utTypes = UTType.types(tag: mimeType, tagClass: .mimeType, conformingTo: nil)
>     print("\(mimeType) -> \(utTypes)")
> }
> ```
> 
> Which outputs something like:
> 
> ```
> application/vnd.apple.pkpass -> [<_UTCoreType 0x123456789012> com.apple.pkpass]
> ```
> 
> So the UTI seems to be `com.apple.pkpass`.

</details>

On an iPhone XR with iOS 15.0 Public Beta 2, when I tap on a `.pkpass` file in the Files app, it literally says `com.apple.pkpass-data`... However, sharing that file from the Files app already works in Signal anyway (perhaps because Files gives a file URL? So the fact that this works might not be related to the UTI), and also if the implementation of *Part 1* is changed to `com.apple.pkpass-data`, it doesn't work when sharing a pass from Wallet. So I chose `com.apple.pkpass`.

**Edit:** see https://github.com/signalapp/Signal-iOS/pull/5053#issuecomment-880563253 below regarding this.

#### Part 1. Allowing to share `.pkpass` files from the Wallet app

The attachment is automatically named `signal-yyyy-mm-dd-hhmmss.pkpass`. Perhaps it makes sense to parse the file into a [`PKPass`](https://developer.apple.com/documentation/passkit/pkpass/) and grab some string from there, although I'm leaning more towards *not* doing that.

#### Part 2. Presenting a `PKAddPassesViewController` when an attachment containing a pass is tapped

Perhaps the method `representedPKPass` should actually be a `lazy` stored property instead in order to not read from disk multiple times (e.g. in case the user taps on it multiple times while being in the chat - maybe they don't want to add the pass but just preview it multiple times while composing some message)? The `handleTap` method seems like it wouldn't try to Quick Look / share the attachment if the `attachmentStream` is `nil`, so by the point this property would be initialized the attachment should be available?

Regarding the `contentType` check there, I'm not sure it's needed, but it seems like (as long as it itself doesn't read the file from disk?) it would allow to quickly tell if the attachment is not a pass, to save time in that case.

Also not sure if better error handling than just `try?` is needed here.

#### Testing

I've added a pass to Simulator's Wallet (by dropping a `.pkpass` file into the Simulator, which I got by sharing a pass via iMessage from a physical iPhone's Wallet) and tested:
- Sharing the pass from Wallet to Signal in the Simulator;
- Adding it from Signal to Wallet in the Simulator:
    - The one that I just shared to Signal;
    - When sent from Signal iOS (`5.16.0.43`);
    - When sent from Signal Desktop;
- Tested that (Quick Look)-able and non-(Quick Look)-able generic attachments still behave just as before when tapped.

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/CV/CVComponents/CVComponentGenericAttachment.swift
+++ b/Signal/src/ViewControllers/ConversationView/CV/CVComponents/CVComponentGenericAttachment.swift
@@ -4,6 +4,7 @@
 
 import Foundation
 import QuickLook
+import PassKit
 
 @objc
 public class CVComponentGenericAttachment: CVComponentBase, CVComponent {
@@ -315,6 +316,17 @@ public class CVComponentGenericAttachment: CVComponentBase, CVComponent {
             return false
         }
         return QLPreviewController.canPreview(url as NSURL)
+    }
+
+    /// Returns the `PKPass` represented by this attachment, if any.
+    public func representedPKPass() -> PKPass? {
+        guard attachmentStream?.contentType == "application/vnd.apple.pkpass" else {
+            return nil
+        }
+        guard let data = try? attachmentStream?.readDataFromFile() else {
+            return nil
+        }
+        return try? PKPass(data: data)
     }
 
     @objc(showShareUIFromView:)
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController+CVComponentDelegate.swift
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController+CVComponentDelegate.swift
@@ -5,6 +5,7 @@
 import Foundation
 import QuickLook
 import PromiseKit
+import PassKit
 
 extension ConversationViewController: CVComponentDelegate {
 
@@ -237,6 +238,11 @@ extension ConversationViewController: CVComponentDelegate {
             let previewController = QLPreviewController()
             previewController.dataSource = attachment
             self.present(previewController, animated: true, completion: nil)
+            return .handledByDelegate
+        } else if PKAddPassesViewController.canAddPasses(),
+                  let pkPass = attachment.representedPKPass(),
+                  let addPassesVC = PKAddPassesViewController(pass: pkPass) {
+            self.present(addPassesVC, animated: true, completion: nil)
             return .handledByDelegate
         } else {
             return .default
--- a/SignalShareExtension/Info.plist
+++ b/SignalShareExtension/Info.plist
@@ -66,6 +66,7 @@
                 $attachment,
                 ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO "public.data"
                 || ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO "public.url"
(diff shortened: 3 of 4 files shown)


Imperiopolis-Signal:
Thanks for opening this, will take a look soon.

u32i64:
Realized that I can run something like:

```swift
if #available(iOS 14, *) {
    let utStrings = [
        "com.apple.pkpass",
        "com.apple.pkpass-data"
    ]
    
    for utString in utStrings {
        let utType = UTType(utString)!
        let supertypes = utType.supertypes
        print("\(utType) conforms to \(supertypes)")
    }
}
```

which outputs something like:

```
com.apple.pkpass conforms to [<_UTCoreType ...> public.item, <_UTCoreType ...> com.apple.package, <_UTCoreType ...> com.apple.bundle, <_UTCoreType ...> public.directory]
com.apple.pkpass-data conforms to [<_UTCoreType ...> public.item, <_UTCoreType ...> public.data]
```

Particularly, the fact that `com.apple.pkpass-data` conforms to `public.data` is likely the reason that sharing `com.apple.pkpass-data` already works in Signal (as `public.data` in general is already supported). I think this resolves the doubts I had regarding the UTI that needs to be used in this PR (so `com.apple.pkpass` should be correct).

Imperiopolis-Signal:
Thanks, this will be in our next release.

PR #5075, opened 2021-08-01
Turn off camera when the view is not in the foreground

fumiakiy:
### Contributor checklist

 * iPhone 7+ iOS 14.6

- - - - - - - - - -

### Description
Turn off the camera capture session when the view is not in the foreground and resume the session again when the view becomes foreground. This way the green indicator on the status bar turns off too when the camera is not actually in use.


Imperiopolis-Signal:
Thanks, we will incorporate this in our next release.

PR #5225, opened 2021-12-25
Center stage

RowanG1:

- [x ] I have read the [README](https://github.com/WhisperSystems/Signal-iOS/blob/master/README.md) and [CONTRIBUTING](https://github.com/WhisperSystems/Signal-iOS/blob/master/CONTRIBUTING.md) documents
- [x ] I have signed the [Contributor Licence Agreement](https://whispersystems.org/cla/)

### Contributor checklist

- [ X] I'm following the [code, UI and style conventions](https://github.com/WhisperSystems/Signal-iOS/blob/master/CONTRIBUTING.md#code-conventions)
- [ X] My commits are in nice logical chunks
- [ X] My contribution is fully baked and is ready to be merged as is
 * Please help with test on physical device. I tested on emulator, and could not load on physical device. I created an issue for help on this:
 https://community.signalusers.org/t/build-on-physical-ios-device-3rd-party-developer/40097/1

- - - - - - - - - -


RowanG1:
> Is this still relevant? If so, what is blocking it? Is there anything you can do to help move it forward?
> 
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.

I created a pull request, and there was no response. I noticed there were other PRs open for some time without any activity. I made a comment in the PR about test help required.

Signal/src/Calls/UserInterface/CallControls.swift
```diff
@@ -192,8 +201,8 @@ class CallControls: UIView {
         hangUpButton.isHidden = call.groupCall.localDeviceState.joinState != .joined
 
         // Use small controls if video is enabled and we have external
-        // audio inputs, because we have five buttons now.
-        [audioSourceButton, flipCameraButton, videoButton, muteButton, hangUpButton].forEach {
+        // audio inputs, because we have six buttons now (generally only on iPad, and 5 buttons for iPhone, due to center stage).
+        [audioSourceButton, flipCameraButton, videoButton, centerStageButton, muteButton, hangUpButton].forEach {
```
Imperiopolis-Signal, line 205: I think you'll want to check if center stage is supported to see how many buttons are actually being rendered (you can check the `isHidden` property on the button). The buttons probably don't need to be small to fit if the `centerStageButton` is hidden.

Signal/src/util/CenterStageUtil.swift
```diff
@@ -0,0 +1,57 @@
+//
+//  Copyright (c) 2020 Open Whisper Systems. All rights reserved.
+//
+
+import Foundation
+
+class CenterStageUtil {
```
Imperiopolis-Signal, line 7: I'm not really understanding why this class uses KVO, or why it's a class at all. It should probably be a `struct` (or even an `enum` since it's never initialized) and it could have something like

```swift
static var centerStageEnabled: Bool {
didSet {  do what you need to do }
}

Imperiopolis-Signal:
Your contribution is appreciated and it'd be great to add center stage support to the app, but I think this PR could use some cleanup.

Signal/src/util/CenterStageUtil.swift
RowanG1: There are two ways for user to switch on/off center stage: via control center, or via in app UI. While the user is on a call, potentially the user could access the control center, and press the Center stage button. The KVO listening is to keep the control center and app UI in sync.

RowanG1: For class vs struct, I was under the impression that structs are meant for data structures. The class here is merely a means to group functionality, in a way that clearly communicates to the point of call what the method relates to (Center stage).

RowanG1:
> Your contribution is appreciated and it'd be great to add center stage support to the app, but I think this PR could use some cleanup.

I think I read in the Signal docs that generally the Signal developers handle the finer clean-up on PRs. I assume I don't need to handle anything further?

Signal/src/Calls/UserInterface/CallControls.swift
RowanG1: I think I checked the hidden property in another place in this PR?

PR #5246, opened 2022-01-17
Add Emoji Search to Reaction Picker

kinoroy:

### Contributor checklist

 * iPhone X, iOS 14.7.1
 * iPhone 13 Pro, iOS 15.1.1

- - - - - - - - - -

## Introduction 

Signal currently allows users to react to messages with emojis, and implements an emoji picker for them to select a reaction. This PR implements a search bar for emojis within the emoji picker

## Motivation

Reacting to a message with an emoji is a frequently used use case in a messaging app. The current implementation can make it slow to find an emoji that the user would like to use.
Other messaging apps implement a search bar for searching the available emojis. Examples: **Facebook Messenger**, **Signal for Desktop**, **Signal for Android**. This is also a requested feature on Signal Community: https://community.signalusers.org/t/add-emoji-search-to-reaction-picker/17975 

## Proposed solution

The proposed solution is to implement a search bar above the emoji picker, and to change the emoji data source to one which provides keywords

## Detailed Design

Note that the number of changes may seem large, however the vast majority of it is auto-generated Emoji files. 

On the UI end of it, its fairly simple, we just add a `UISearchBar` to the top of the existing `EmojiPickerSheet`

The current data source for generating the emoji swift files, does not include any "keyword" data which would allow us to search the emojis by anything other than their exact name. I propose changing the data source to EmojiBase (https://emojibase.dev), a Unicode spec-complient dataset which is autogenerated from the official Unicode data files. This dataset is licenced under the **MIT licence**, which is the **same licence** as the existing data set used by Signal (https://github.com/iamcal/emoji-data) and is served by a CDN making it more reliable to fetch than the exiting dataset, which was often unavailable.

A new file is generated by `EmojiGenerator.swift`, the `EmojiTags.swift` contains a dictionary mapping which maps each possible keyword in the dataset to an array of emoji which match that keyword. 

For implementing the actual search, I created an `EmojiKeywordMatcher` class, which upon initialization generates a trie (https://en.wikipedia.org/wiki/Trie) data structure which is stored in memory, allowing for fast, O(1) lookup of the matching emoji for a given search string. The trie is generated once in O(n+m) time where n is the number of emoji and m is the number possible keywords. In my testing there was no noticable UI delays, and the search is very snappy. 

## Alternatives considered

Instead of replacing the entire dataset and changing exising code in the `EmojiGenerator.swift`, I considered referencing a new dataset only for the keyword data. However I decided against this for simplicity, and all the information we need can be found in the EmojiBase dataset.

I also considered decreasing the size of this PR by creating a PR with firstly only the change in data source to EmojiBase, and then follow up with another PR to add the search functionality. This is still a possible option if the reviewer would like a smaller PR. 

## Future Directions

At the moment, the search is only implemented in English. However the EmojiBase dataset does have datasets for many languages as it's derrived from the Unicode CLDR data. A future improvment could be to generate data for multiple languages and thus allow the user to search emojis in any language. I decided against this in this PR as there are already a number of changes. 

## Demo

https://user-images.githubusercontent.com/22554212/149709843-decc2649-e974-4644-859c-b245369525c6.MP4


SignalServiceKit/src/Messages/Reactions/ReactionManager.swift
```diff
@@ -7,7 +7,7 @@ import Foundation
 @objc(OWSReactionManager)
 public class ReactionManager: NSObject {
 
-    public static let emojiSet = ["❤️", "👍", "👎", "😂", "😮", "😢"]
+    public static let emojiSet = ["❤️", "👍️", "👎️", "😂", "😮", "😢"]
```
kinoroy, line 10: What changed here? Some emoji can have a textual representation or an "emoji"/picture presentation, this is due to some emoticons predating emoji. Which one is used can be decided using a [variant selector](https://character.construction/emoji-vs-text). In the old hardcoded emoji, no variant selector was included, so the display of the emoji was system dependant. On macOS this often just defaults to the emoji/picture presentation. The EmojiBase dataset includes the emoji variant selector on its emoji therefore these hardcoded emoji must be changed to match. And in this case I think it is more correct to use the emoji version anyways, as we never want the textural representation. 

Example showing the difference in codepoints from the Swift REPL:
```
  8> let oldEmoji = "👍"
oldEmoji: String = "👍"
  9> let newEmoji = "👍️"
newEmoji: String = "👍️"
 10> print(Array(oldEmoji.unicodeScalars))
["\u{0001F44D}"]
 11> print(Array(newEmoji.unicodeScalars))
["\u{0001F44D}", "\u{FE0F}"]
```

Imperiopolis-Signal:
Thanks for your contributions, but we added emoji search in a slightly different way.

Issue #5257, opened 2022-02-05
Broken new feature: Reaction picker configuration (where you can set your default emojis): Changing an emoji replaces not selected emoji


meyermarcel:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
In the reaction picker configuration (where you can set your default emojis): When I change emoji to the same emoji the assignment gets confused and replacing the selected emoji does not work. This breaks the reaction picker configuration and makes it unusable.

### Steps to reproduce
- Open reaction picker configuration
- Set same emoji multiple times in the reaction picker configuration
- Change emojis

**Actual result:** Not selected emoji gets replaced.

**Expected result:** The emoji that I selected should be replaced.

### Screenshots

I select always the 🆗 emoji for 🙌🏼 and 🧠:

https://user-images.githubusercontent.com/3398801/152642956-8b90e11d-580c-4c84-9306-339f83c40438.mp4

### Device info

**Device**: iPhone SE 1. Generation

**iOS version**: 15.3

**Signal version:** 5.27.1

### Link to debug log
https://debuglogs.org/79c06d7071691dbd8b70f4c1a317a093bfc1725c6a4b762471a1c81ad6992947.zip

And thank you for making this free and awesome App happen! :)

meyermarcel:
It works on my iPhone SE 1st gen with iOS 15.4.1 and Signal 5.34.1.0
Thanks to @Imperiopolis-Signal and @DeVitoC !

PR #5269, opened 2022-02-17
Fix Issue #5257 - Broken new feature: Reaction picker configuration (where you can set your default emojis): Changing an emoji replaces not selected emoji 

DeVitoC:

### Contributor checklist

 * iPhone 8, iOS 15.2.1 (Physical device)
 * iPhone 13, iOS 15.2 (Simulator)

- - - - - - - - - -

### Description

This fix addresses Issue [#5257](https://github.com/signalapp/Signal-iOS/issues/5257) where it was reported that when multiple emojis in the favorites bar were set to the same emoji, it would lock you into just changing the last of those and you would be unable to change from the multiple emoji's. 

As I ran through the method in the MessageReactionPicker to see how the values change, I found that the issue was the dictionary buttonForEmoji. As a duplicate emoji (i.e. Thumbs Up) was added, it would add the button from the old emoji (i.e. Thumbs Down) to the emoji that matches the new emoji in the dictionary. This meant that Thumbs Up in the dictionary now has the button for the Thumbs Down, and Thumbs Down also does, and the button for the original Thumbs Up is lost. 

I had two ideas for fixing this, one was simply to have it return out of the method without changing the emoji if it was a duplicate, and the other was to change it from a dictionary to a structure that would maintain duplicates, namely an array of tuples (closely mirroring the key value pairing of the dictionary). I felt that the latter would be less noticeable a change and opted to do it this way. However, this did require quite a bit more changed lines of code. If this is not the preferred change, I'm happy to submit a PR with a simple few lines returning out of the method without changing the emoji in the case of a duplicate to simply prevent duplicates in the first place. 

As a note, I did in a separate branch, add a little bit more code that changed the functionality slightly more to swap emoji's in the case of a duplicate in case that would be a more desirable functionality that having duplicates. This would allow users to swap the positions on the emoji picker and rearrange their emojis simply that way. At this point I am not submitting that PR as it is enough of a change that I wasn't sure that it would be a welcome change. 

If the emoji swap change would be desired, I am happy to submit that and if you would prefer staying with a dictionary and simply not doing anything if a duplicate is chosen, I am happy to submit that instead also.

### DIFF
--- a/Signal.xcodeproj/project.pbxproj
+++ b/Signal.xcodeproj/project.pbxproj
@@ -4845,7 +4845,6 @@
 						ProvisioningStyle = Automatic;
 					};
 					D221A088169C9E5E00537ABF = {
-						DevelopmentTeam = U68MSDN6DR;
 						LastSwiftMigration = 1020;
 						ProvisioningStyle = Automatic;
 						SystemCapabilities = {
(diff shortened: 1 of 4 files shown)


Imperiopolis-Signal:
c63fe9f9f4

PR #5373, opened 2022-06-03
Photo taken in landscape mode didn't rotate when sent (signalapp#4334)

blujan:

- [x] I have signed the [Contributor License Agreement](https://signal.org/cla/)

### Contributor checklist

 * iPhone 6s iOS 15.5

- - - - - - - - - -

### Description
This is fix for issue where, when the rotation lock is enabled, a photo will not be properly rotated upon capture. I also noticed, under this scenario, that the UI buttons also did not rotate in the same manner as the built-in iOS camera with the lock enabled.

#### Background
This occurs because UIDevice.current.orientation becomes locked to portrait when the rotation lock is enabled.

#### Implementation
The only way to get device orientation data regardless of the orientation lock is to pull the data directly from the accelerometer. I have added three functions to act as drop-in replacements to the current functions handling notifications from beginGeneratingDeviceOrientationNotifications().

**startOrientationUpdates()** -> This sets up CoreMotion to generate updates from the accelerometer. If CoreMotion or the accelerometer is not available or restricted from use, then the original method of using beginGeneratingDeviceOrientationNotifications() is used.

**stopOrientationUpdates()** -> Stops updates from CoreMotion or GeneratingDeviceOrientationNotifications if they are active.

**orientationDidChange(orientation: UIDeviceOrientation)** -> Overload of the existing orientationDidChange and basically does the same thing aside from the translation of data types.

The accelerometer update code was based on references from this post:
https://stackoverflow.com/questions/49164302/get-current-ios-device-orientation-even-if-devices-orientation-locked

#### Result
With the rotation lock enabled, photos now capture the correct rotation and the camera UI buttons rotate with the device as expected.

### DIFF
--- a/Signal/src/ViewControllers/Photos/PhotoCapture.swift
+++ b/Signal/src/ViewControllers/Photos/PhotoCapture.swift
@@ -7,6 +7,7 @@ import CoreServices
 import Foundation
 import SignalCoreKit
 import UIKit
+import CoreMotion
 
 protocol PhotoCaptureDelegate: AnyObject {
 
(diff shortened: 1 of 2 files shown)


blujan:
Hey, just a quick update on this - I saw that the recent UI refactor (looks great btw) now causes conflicts in the merge. Moreover, the issue described still exists when the rotation lock is enabled. My initial quick check made it look like it shouldn't be too much effort to resolve the conflict and re-test everything. I've been particularly busy as of late, but hopefully I'll get to this in the next week or so.

PR #5407, opened 2022-07-26
Fix missing Change Photos option when limited access provides no photos

### LINKED ISSUE #5401: Unable to select more photos within the app to send

### Bug description
I remember at some point that I was able to edit the list of selected photos, that sign al had access to, from within the app. But this hasn’t been the case for a while. Assuming this is a bug, as it feels like intention should be to allow users to select the photos they want to share through signal, from an interface printed within the signal app.

### Steps to reproduce
- make sure signal only has “selected photos” photo access
- Open a chat
- Go to camera
- Press photo library

**Actual result:** There is no way to select photos

**Expected result:** that a button to prompt the iOS “select more photos” dialog was present

### Device info
iPhone 12 mini iOS 15.4

**Signal version:** 5.44.0.10

QuaqSim:

### Contributor checklist
 * iPhone 13 Pro Simulator , iOS 15.5

- - - - - - - - - -

### Description
This pull requests fixes #5401 by displaying the standard RecentPhotosCollectionView instead of the RecentPhotosErrorView when it has no photos but library access is limited.

I tested this by running the modified code using the iPhone simulator and confirming that when pressing the + icon to the left of the message field in a chat, while Signal has limited access to the photos library and none selected, it displays the Change Photos message and button correctly, as well as immediately after changing the selection to have 1 or 0 photos.


emreberge:
Just checked the code, haven't tested it my self, but wondering if the change should go into [`checkPermissions`](https://github.com/signalapp/Signal-iOS/blob/3e505e58cbc7fafee8355feb60840cd431563697/Signal/src/ViewControllers/Attachment%20Keyboard/AttachmentKeyboard.swift#L186) function instead?

Suggestion:
```
        case .authorized:
            showRecentPhotos()
        case .denied, .restricted, .limited:
            showRecentPhotosError()
```

QuaqSim:
I think the `checkPermissions` change would have the opposite effect and worse, it would result in the error view being displayed even when the user has selected photos. My proposed change does make an assumption about `RecentPhotosCollectionView` which perhaps it shouldn't.

---

An alternative might be to reuse the view code of `SelectMorePhotosCell` as a subview of `RecentPhotosErrorView` in the situation of having limited access to the library but not having any images selected.

I see `UICollectionViewCell` is eventually a subclass of `UIView` but I'm not very familiar with `UICollectionView`, I'm not sure if it would be appropriate to directly add `SelectMorePhotosCell` as a subview in `RecentPhotosErrorView`. If it isn't, the view code might need to be lifted into say a `SelectMorePhotosView` class that both `SelectMorePhotosCell` and `RecentPhotosErrorView` can use.

In this alternative, `AttachmentKeyboard` would also need to conform to `PhotoLibraryDelegate` so that it can be notified of a change in library access and (via a call to `checkPermissions`), do `showRecentPhotos()` or `showRecentPhotosError()` as appropriate.

PR #5412, opened 2022-07-30
Improve VoiceOver support for the settings button

meherkasam:

### Contributor checklist

 * iPhone 11 Pro, iOS 15.6
 * iPhone 13 Pro Simulator, iOS 15.5

- - - - - - - - - -

## Description

Previously, when a user would tap the Settings button in VoiceOver, the cursor would not jump to the context menu. All of the elements in the chat list screen would remain visible to VoiceOver which is not the desired behaviour. With the changes in this PR, the context menu experience on VoiceOver is similar to `UIPopoverController` or `UIAlertController`. Since, unlike a `UIAlertController`, there's no explicit cancel button, I've added support for the [two-finger scrub](https://dequeuniversity.com/screenreaders/voiceover-ios-shortcuts) gesture.

### Before
| Screen Recording  | Reveal Accessibility Inspector |
| ------------- | ------------- |
| <video src="https://user-images.githubusercontent.com/450030/182002622-9ec1a25a-59fe-4311-9115-63800a395fbb.mov">  | <image src="https://user-images.githubusercontent.com/450030/182002947-f06c27de-1a29-48fe-8c48-04009e55442c.png" height=800> |

### After
| Screen Recording  | Reveal Accessibility Inspector |
| ------------- | ------------- |
| <video src="https://user-images.githubusercontent.com/450030/182002626-dede8885-4643-4d4d-a81a-5fffc2b00745.mov">  | <image src="https://user-images.githubusercontent.com/450030/182002988-d4c4059f-e5ad-4a43-8cc7-f188d8caf75b.png" height=800> |

### DIFF
--- a/Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController+Multiselect.swift
+++ b/Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController+Multiselect.swift
@@ -181,8 +181,13 @@ extension ChatListViewController {
         v.frame = CGRect(x: 0, y: 0, width: size.width, height: size.height)
         v.delegate = self
 
-        viewState.multiSelectState.contextMenuView = ContextMenuActionsViewContainer(v)
-        viewState.multiSelectState.contextMenuView?.addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(hideMenu)))
+        viewState.multiSelectState.contextMenuView = ContextMenuActionsViewContainer(
+            v,
+            dismissCallback: { [weak self] in
+                self?.hideMenu()
+            }
+        )
+        
         window.addSubview(viewState.multiSelectState.contextMenuView!)
         viewState.multiSelectState.contextMenuView!.frame = window.bounds
         adjustContextMenuPosition()
@@ -212,6 +217,8 @@ extension ChatListViewController {
             self?.viewState.multiSelectState.parentButton?.alpha = 1
             self?.viewState.multiSelectState.parentButton = nil
             self?.viewState.multiSelectState.contextMenuView = nil
+            // ensures that the VoiceOver cursor jumps to the parent button upon dismissal of the modal
+            UIAccessibility.post(notification: .screenChanged, argument: self?.viewState.multiSelectState.parentButton)
             handler?()
         }
         if animated {
@@ -462,12 +469,14 @@ extension ChatListViewController: ContextMenuActionsViewDelegate {
 private class ContextMenuActionsViewContainer: UIView {
     static let offset = CGPoint(x: 8, y: 0)
     fileprivate let menuContainer: UIView
+    private let dismissCallback: () -> Void
 
-    required init(_ target: UIView) {
+    required init(_ target: UIView, dismissCallback: @escaping () -> Void) {
         let container = UIView(frame: target.bounds)
         container.autoresizingMask = [.flexibleRightMargin, .flexibleBottomMargin]
         self.menuContainer = container
-
+        self.dismissCallback = dismissCallback
+        
         super.init(frame: .zero)
 
         let radius = target.layer.cornerRadius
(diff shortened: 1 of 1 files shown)


meherkasam:
I separately recommend changing the accessibility label for the button on the top left from `Settings` to `Menu`. It is currently a bit confusing to tap on a "Settings" button only to be shown yet another "Settings" button.

Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController+Multiselect.swift
```diff
@@ -181,8 +181,7 @@ extension ChatListViewController {
         v.frame = CGRect(x: 0, y: 0, width: size.width, height: size.height)
         v.delegate = self
 
-        viewState.multiSelectState.contextMenuView = ContextMenuActionsViewContainer(v)
-        viewState.multiSelectState.contextMenuView?.addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(hideMenu)))
+        viewState.multiSelectState.contextMenuView = ContextMenuActionsViewContainer(v, dismissCallback: hideMenu)
```
harry-signal, line 184: I think you'll be creating a retain cycle here where hideMenu has a strong reference to the ChatListViewController instance which has a strong reference to its viewState which has a strong reference to contextMenuView, which now has a strong reference to hideMenu.  You should be able to fix it by making the callback' self reference weak (but then you'll want to pass an explicit block like:
dismissCallback: { [weak self] in self?.hideMenu() }

harry-signal (changes requested):
Thank you for writing this! Special thanks for the detailed description and before/after screen captures, very helpful! :)

Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController+Multiselect.swift
meherkasam: Thank you for reviewing this PR, @harry-signal. Without changing the code, I added a `deinit` block inside `ContextMenuActionsViewContainer` and set a breakpoint inside of it. When I dismiss the menu, it hits the breakpoint. I believe that happens because of this [line](https://github.com/signalapp/Signal-iOS/blob/main/Signal/src/ViewControllers/HomeView/Chat%20List/ChatListViewController%2BMultiselect.swift#L214). Setting the `contextMenuView` to `nil` breaks the retain cycle. Please let me know if you'd still like me to make the suggested change.

harry-signal: Thanks for the detailed reply (and sorry for my own delay). Yeah, you're right that removing the menu view on dismiss resolves the retain cycle. Still, it feels dangerous to have a temporary retain cycle like this, because its easy to make a future change that forgets to remove the circular reference and leaves things in a bad state. Its much safer to make the reference weak and never create a cycle in the first place. Of course in practice the weak reference isn't going to nil out, so it should be an easy change. Does that all make sense?

meherkasam: I've made the change you requested. Please take a look and let me know if you have any other concerns.

harry-signal: Thank you so much! I'll get this merged in and close this PR out, should be out in the next beta!

meherkasam: Thank you. I noticed that you started the CI run and there's a linter failure. The failed script is called `Scripts/precommit.py` but it did not trigger a linter check for me prior to committing. Am I missing something in my environment setup?

harry-signal: Yeah you can run the precommit script manually, or add it to your git configs precommit so it auto-lints things for you. Regardless, the way it works is I'll make a copy of this PR to merge into our internal repo, and in ~1 week the private repo will get pushed here to the public one and I will link this PR to the final commit and close it. I can run lint checks when I do that (they'll just run automatically) so no need to do anything on your end!

meherkasam: @harry-signal This PR is still open. Wanted to follow up with you to see if everything went okay or if there's something I can do on my end to help resolve.

harry-signal: Oh my bad, I merged this already in this PR: https://github.com/signalapp/Signal-iOS-Private/pull/4713. Github has weird UI

image

I'll explicitly mark this one closed. Thanks for sending this!

meherkasam: Sweet! Thank you for the update.

PR #5423, opened 2022-08-24
Simplify process for external developers to sign local Xcode builds

mlinington:
## Motivation

The Signal organization in ADC has claimed the `org.whispersystems.signal` bundleID. This means that any developer that's not a member of this organization must swap out the project bundleID for something else in order for Apple to sign their local build. This requires developers to chase down any usage of `org.whispersystems...` and swap it with their own version.

image

## Proposal

Define two new Xcode build settings: `SIGNAL_BUNDLEID_PREFIX` and `SIGNAL_MERCHANTID`. These build settings will be consulted throughout the project to derive each executable's bundleID and their entitlements. They'll also be recorded in the Info.plist so they're accessible at runtime.

Just parameterizing the bundleID prefix is insufficient, since Apple also validates that merchantIDs don't collide. Only updating bundleIDs while leaving merchantIDs as-is results in the following error:
image

By default, their values will be:
```
SIGNAL_BUNDLEID_PREFIX = org.whispersystems;
SIGNAL_MERCHANTID = org.signalfoundation;
```

In order to test on my device, I'm able to redefine these build settings to be:
```
SIGNAL_BUNDLEID_PREFIX = com.mlinington;
SIGNAL_MERCHANTID = com.mlinington;
```

By changing this setting and rebuilding, everything remaps appropriately:
`org.whispersystems.signal` -> `com.mlinington.signal`
`org.whispersystems.signal.SignalNSE` -> `com.mlinington.signal.SignalNSE`
`group.org.whispersystems.signal.group.staging` -> `group.com.mlinington.signal.group.staging`
etc. etc.

This greatly simplifies the work required to get a local build running on a device.

## Result

Now, an external developer is able to deploy Signal to their own test devices in two steps.

Step one: Update the Development Team for each executable target. These must be updated independently since the current project has per-target Development Teams. (Though, I'm not sure why the per-target settings are necessary. It's likely this could be simplified further by having the Signal project use per-project Development Team settings as well).
Screen Shot 2022-08-23 at 5 18 13 PM
Screen Shot 2022-08-23 at 5 22 16 PM

Step two: 
Set SIGNAL_BUNDLEID_PREFIX and SIGNAL_MERCHANTID to something unique to you.
Screen Shot 2022-08-23 at 5 27 25 PM

## Things I haven't changed

After this change, there's still a handful of results for "org.whispersystems" when searching the workspace. They are:
1. The Signalyzer instrument. As best I can tell, this is the identifier of the instrument package itself. I think it's fine to leave as-is.
2. UTI and Activity identifiers. These should certainly be left as-is to maintain compatibility between native-Signal and homegrown-Signal.
3. Miscellaneous comments

I also haven't changed anything in the build scripts. Those will continue to use the hardcoded bundleIDs as before. I can't imagine external developers will need to use these scripts anyway.

## Testing

This change necessarily introduces some risk to Signal's build process that I'm not able to test on my own. Before merging, it's probably worthwhile for a Signal engineer to kick off a test build to make sure the build server is able to handle this change properly and that the built product runs correctly.

Additionally, there are a couple other things I'd suggest a Signal engineer verifies:
1. The built product entitlements should be byte-for-byte identical. This can verified this by diffing the output of `codesign -d -vvv --entitlements - Signal.app` from the app in your built products directory before and after this change.
2. Similarly, the Info.plist for each target should be *mostly* identical. The only two changes we should expect to see are two new keys: `OWSBundleIDPrefix` and `OWSMerchantID`


SignalMessaging/contacts/OWSSyncManager.m
```diff
@@ -317,7 +317,10 @@ - (dispatch_queue_t)serialQueue
     static dispatch_queue_t _serialQueue;
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
-        _serialQueue = dispatch_queue_create("org.whispersystems.contacts.syncing", DISPATCH_QUEUE_SERIAL);
+        NS_VALID_UNTIL_END_OF_SCOPE NSString *label = [OWSDispatch createLabel:@"contacts.syncing"];
+        const char *cStringLabel = [label cStringUsingEncoding:NSUTF8StringEncoding];
+
+        _serialQueue = dispatch_queue_create(cStringLabel, DISPATCH_QUEUE_SERIAL_WITH_AUTORELEASE_POOL);
```
mlinington, line 323: Bonus change: `DISPATCH_QUEUE_SERIAL` -> `DISPATCH_QUEUE_SERIAL_WITH_AUTORELEASE_POOL`

This is usually wise idea anyway. The dispatch docs seem to imply that objects autoreleased on this queue won't have a pool.
```
 * @const DISPATCH_AUTORELEASE_FREQUENCY_INHERIT
 * Dispatch queues with this autorelease frequency inherit the behavior from
 * their target queue. This is the default behavior for manually created queues.
 ...
 * @const DISPATCH_AUTORELEASE_FREQUENCY_NEVER
 * Dispatch queues with this autorelease frequency never set up an individual
 * autorelease pool around the execution of a block that is submitted to it
 * asynchronously. This is the behavior of the global concurrent queues.
```

From man dispatch_queue_create:
```
The default target queue of all dispatch objects created by the application
     is the default priority global concurrent queue
```

SignalServiceKit/src/Util/DispatchQueue+OWS.swift
```diff
@@ -87,4 +87,14 @@ extension OWSDispatch {
...
         return DispatchQueue.sharedQueue(at: qos)
     }
+
+    /// Returns a reverse-DNS queue label namespaced by appending the `suffix` to the current executable's bundleId
+    public static func createLabel(_ suffix: String) -> String {
+        let prefix = Bundle.main.bundleIdentifier ?? {
+            owsFailDebug("Missing bundleID")
+            return "UNDEFINED"
+        }()
+
```
EvanHahn-Signal, line 97: Nitpick: It feels strange to have it return `UNDEFINED` as if it's trying to recover from an invalid state. I feel like this should either (1) return `org.whispersystems.signal.` instead of "UNDEFINED" (2) fully crash. I think I prefer the second option.

```suggestion
        guard let prefix = Bundle.main.bundleIdentifier else {
            owsFail("Missing bundleID")
        }
```

SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -85,7 +85,10 @@ + (dispatch_queue_t)serialQueue
     static dispatch_queue_t queue = nil;
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
-        queue = dispatch_queue_create("org.whispersystems.disappearing.messages", DISPATCH_QUEUE_SERIAL);
+        NS_VALID_UNTIL_END_OF_SCOPE NSString *label = [OWSDispatch createLabel:@"disappearingMessages"];
```
EvanHahn-Signal, line 88: Newbie question: why did this change from `disappearing.messages` to `disappearingMessages`?

EvanHahn-Signal (changes requested):
This LGTM (though I admit I'm a bit of a newbie here) other than two small nitpicks. Could you address those?

SignalServiceKit/src/Util/DispatchQueue+OWS.swift
mlinington: I'm fine with that! Though, to be clear: dispatch queue labels are [optional and really only used to aid in debugging](https://developer.apple.com/documentation/dispatch/1642205-dispatch_queue_create_with_targe). So in the worst case, this would just result in something like the following showing up in your backtrace:

```
Thread 14 name:   Dispatch queue: UNDEFINED.PKPushRegistry
Thread 14 Crashed:
*** Backtrace details ***
```

mlinington: It's also worth noting that Signal isn't really consistent with naming conventions. In the link above, Apple suggests using a reverse-DNS label format to make it easier to differentiate who created the queue. However, in some old crash I have saved locally:
```
Thread 14 name:   Dispatch queue: GRDB Storage.reader.1
Thread 14 Crashed:
0   libsystem_kernel.dylib        	       0x20a232200 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x21a46f1ac pthread_kill + 268
2   libsystem_c.dylib             	       0x1d5029ca0 abort + 180
3   libc++abi.dylib               	       0x21a3b1b8c abort_message + 132
```

Who created this queue? Did it come from Signal? Or some linked framework? Something namespaced like "org.whispersystems.signal.grdbStorage.reader.1" would make it much more clear who the originator is.

mlinington: Sorry to keep commenting but just one more thing I want to note. I don't think it's even possible for an iOS app to run without a bundleId. I think the optionality comes from macOS land. So your suggestion to hard assert is almost certainly safe.

SignalServiceKit/src/Messages/OWSDisappearingMessagesJob.m
mlinington: I figured this was a typo from before but I can change it back if you'd prefer. Like I mentioned above, queue labels are really only used for debugging. I think it'd make sense to have it called `org.whispersystems.signal.disappearing.messages` if there were a family of `org.whispersystems.signal.disappearing.*` queues to go along with it.

The whole subsystem that handles disappearing messages is called just that: "Disappearing Messages". There's no separate "Disappearing Photos" or whatever. So to me, it makes sense for the queue to be namespaced as if it's a singular concept. But if you'd prefer the other format I'm happy to undo it!

EvanHahn-Signal (approved):
Thanks! Could you sign the CLA and we can get this merged?

mlinington:
> Thanks! Could you sign the CLA and we can get this merged?

Signed!

EvanHahn-Signal:
Thanks! We'll get this merged this into our internal repo. I'm going to close this, but expect to see your change in the public repo soon!!

PR #5432, opened 2022-09-11
Bug fix: Stale thread selection in ChatListViewController

mlinington:
~Branched off of existing PR #5423. Relevant changes start with 25d8f76d39bd97f141757faef292947e57fce18e~
Rebased off main as of 10/5/22

This change corrects an issue where presented threads may not update the selected thread in a split view. See the following video for a couple of examples:

https://user-images.githubusercontent.com/3432567/189549645-8eaf07ef-144a-4078-83c9-5c130d6963fc.mov

After this change:

https://user-images.githubusercontent.com/3432567/189549659-48715760-a43a-4cab-8312-e30d96824787.mov

### Implementation Details

When updating the DetailVC, the split view will notify the ChatListVC to update its selection for the newly presented thread. More details in the commit messages.

It's worth noting that `CLVTableDataSource.buildConversationCell(tableView:indexPath:)` will also update the table's selection when building a cell if it observes that the SplitView's selected thread matches the cell currently being built. One alternative approach could be for the SplitView to dirty the selected cell to force the ChatList to rebuild this cell and force a selection update.

I think a great future direction (not included in this change) would be for the ChatList to automatically reconfigure its mode based on the selected thread. If the user opens an archived thread, should the ChatList automatically switch to archived view? When searching threads, should selections persist themselves? (They don't currently, see the first line of `ConversationSearchViewController.tableView(_:didSelectRowAt:)`)

### Testing Details
I've only tested on iOS 15 since that's all of the devices I have available to me. I've tested my changes both when the split view is collapsed and expanded. I've tested conversation presentation from the ChatList inbox, archive, and search modes. I've also tested conversation presentation originating from outside the ChatList, such as notification handlers and contact sheets.

### DIFF
--- a/Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController+Actions.swift
+++ b/Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController+Actions.swift
@@ -21,7 +21,6 @@ extension ChatListViewController {
         if let previousIndexPath = renderState.indexPath(beforeThread: currentThread),
            let thread = self.thread(forIndexPath: previousIndexPath) {
             self.present(thread, action: .compose, animated: true)
-            tableView.selectRow(at: previousIndexPath, animated: true, scrollPosition: .none)
         }
     }
 
@@ -40,7 +39,6 @@ extension ChatListViewController {
         if let nextIndexPath = renderState.indexPath(afterThread: currentThread),
            let thread = self.thread(forIndexPath: nextIndexPath) {
             self.present(thread, action: .compose, animated: true)
-            tableView.selectRow(at: nextIndexPath, animated: true, scrollPosition: .none)
         }
     }
 }
--- a/Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController+Helpers.swift
+++ b/Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController+Helpers.swift
@@ -145,8 +145,6 @@ public extension ChatListViewController {
             owsFailDebug("Missing threadViewModel.")
             return nil
         }
-        self.lastViewedThread = threadViewModel.threadRecord
-
         let vc = ConversationViewController(threadViewModel: threadViewModel,
                                             action: .none,
                                             focusMessageId: nil)
--- a/Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.h
+++ b/Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.h
@@ -28,7 +28,9 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)unarchiveSelectedConversation;
 
 @property (nonatomic, readonly) CLVViewState *viewState;
-@property (nonatomic) TSThread *lastViewedThread;
+
+/// Used to update the selected cell for split view and maintain scroll positions for reappearing collapsed views.
+- (void)updateLastViewedThread:(TSThread *)thread animated:(BOOL)animated;
 
 // For use by Swift extension.
 - (void)updateBarButtonItems;
--- a/Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.swift
+++ b/Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.swift
@@ -498,17 +498,28 @@ public extension ChatListViewController {
     }
 
     /// Verifies that the currently selected cell matches the provided thread's uniqueId.
-    /// If it does: Do nothing.
-    /// If it doesn't: Select the first cell matching the provided thread, if one exists.
+    /// If it does or if the user's in multi-select: Do nothing.
+    /// If it doesn't: Select the first cell matching the provided thread, if one exists. Otherwise, deselect the current row.
+    @objc
     func ensureSelectedThread(_ targetThread: TSThread, animated: Bool) {
+        // Ignore any updates if we're in multiselect mode. I don't think this can happen,
+        // but if it does let's avoid stepping over the user's manual selection.
         let currentSelection = tableView.indexPathsForSelectedRows ?? []
-        let isThreadCurrentlySelected = currentSelection.lazy
-            .map { self.tableDataSource.thread(forIndexPath: $0) }
-            .contains { $0?.uniqueId == targetThread.uniqueId }
+        guard viewState.multiSelectState.isActive == false, currentSelection.count < 2 else {
+            return
+        }
 
-        if !isThreadCurrentlySelected,
-           let targetPath = tableDataSource.renderState.indexPath(forUniqueId: targetThread.uniqueId) {
-            tableView.selectRow(at: targetPath, animated: animated, scrollPosition: .none)
+        let currentlySelectedThread = currentSelection.first.flatMap {
+            self.tableDataSource.thread(forIndexPath: $0)
+        }
+
+        if currentlySelectedThread?.uniqueId != targetThread.uniqueId {
+            if let targetPath = tableDataSource.renderState.indexPath(forUniqueId: targetThread.uniqueId) {
+                tableView.selectRow(at: targetPath, animated: animated, scrollPosition: .none)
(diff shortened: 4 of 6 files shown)


Signal.xcodeproj/project.pbxproj
```diff
@@ -1094,6 +1094,7 @@
 		E1368CBE18A1C36B00109378 /* MessageUI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = B9EB5ABC1884C002007CBB57 /* MessageUI.framework */; };
 		E44AD4E624E98F440035D7B8 /* PhotoCaptureDismiss.swift in Sources */ = {isa = PBXBuildFile; fileRef = E44AD4E524E98F430035D7B8 /* PhotoCaptureDismiss.swift */; };
 		E75DD3E02810CDBD00E32C36 /* SubscriptionManagerTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = E75DD3DF2810CDBD00E32C36 /* SubscriptionManagerTest.swift */; };
+		E7D7C93F28B580AC003F043B /* Bundle+OWS.swift in Sources */ = {isa = PBXBuildFile; fileRef = E7D7C93E28B580AC003F043B /* Bundle+OWS.swift */; };
```
mlinington, line 1097: Inherited from #5423. Changes relevant to this PR start with https://github.com/signalapp/Signal-iOS/commit/25d8f76d39bd97f141757faef292947e57fce18e.

Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController+Actions.swift
```diff
@@ -21,7 +21,6 @@ extension ChatListViewController {
         if let previousIndexPath = renderState.indexPath(beforeThread: currentThread),
            let thread = self.thread(forIndexPath: previousIndexPath) {
             self.present(thread, action: .compose, animated: true)
-            tableView.selectRow(at: previousIndexPath, animated: true, scrollPosition: .none)
```
mlinington, line 24: There were some existing paths that would update the table selection when presenting a thread. Now that all presentations will automatically update the table selection, this is no longer necessary.

Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController+Helpers.swift
```diff
@@ -145,8 +145,6 @@ public extension ChatListViewController {
             owsFailDebug("Missing threadViewModel.")
             return nil
         }
-        self.lastViewedThread = threadViewModel.threadRecord
```
mlinington, line 148: This didn't seem right. I'm not sure why we'd want to adjust our scroll offset based on our last thread that we displayed in a preview platter.

Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.h
```diff
@@ -28,7 +28,9 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)unarchiveSelectedConversation;
 
 @property (nonatomic, readonly) CLVViewState *viewState;
-@property (nonatomic) TSThread *lastViewedThread;
+
+/// Used to update the selected cell for split view and maintain scroll positions for reappearing collapsed views.
+- (void)updateLastViewedThread:(TSThread *)thread animated:(BOOL)animated;
```
mlinington, line 33: Since `lastViewedThread` is only every accessed internally, I replaced the public property with a private property and a public update method.

Bonus fix: This property is always initially nil but the declaration marks it nonnull. Now the property is nullable, but its public update method still requires a nonnull thread argument.

Signal/src/environment/PushRegistrationManager.swift
```diff
@@ -32,7 +32,7 @@ public class PushRegistrationManager: NSObject, PKPushRegistryDelegate {
     // Private callout queue that we can use to synchronously wait for our call to start
     // TODO: Rewrite call message routing to be able to synchronously report calls
     private static let calloutQueue = DispatchQueue(
-        label: "org.whispersystems.signal.PKPushRegistry",
```
mlinington, line 35: Everything below here was inherited from https://github.com/signalapp/Signal-iOS/pull/5423 and can be ignored for this PR.

Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.swift
```diff
@@ -496,6 +496,32 @@ public extension ChatListViewController {
         navigationController.setViewControllers(viewControllers, animated: false)
         presentFormSheet(navigationController, animated: true, completion: completion)
     }
+
+    /// Verifies that the currently selected cell matches the provided thread's uniqueId.
```
sashaweiss-signal, line 500: nit: `matches the provided thread`, to avoid leaking an implementation detail (the `uniqueId`). I don't care that much about this.

Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.swift
```diff
@@ -496,6 +496,32 @@ public extension ChatListViewController {
...
+    @objc
+    func ensureSelectedThread(_ targetThread: TSThread, animated: Bool) {
+        // Ignore any updates if we're in multiselect mode. I don't think this can happen,
+        // but if it does let's avoid stepping over the user's manual selection.
+        let currentSelection = tableView.indexPathsForSelectedRows ?? []
+        guard viewState.multiSelectState.isActive == false, currentSelection.count < 2 else {
+            return
+        }
+
+        let currentlySelectedThread = currentSelection.first.flatMap {
```
sashaweiss-signal, line 512: nit: since if this ends up `nil` we'll basically just return anyway, would prefer we do above as:

```swift
guard
    !viewState.multiSelectState.isActive,
    currentSelection.count < 2,
    let currentlySelectedThread = currentSelection.first
else {
    return
}

let ... = self.tableDataSource.thread(forIndexPath...)
```

or similar.

sashaweiss-signal: Can then reuse `currentlySelectedThread` below, and not need an `else if`

Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController+Helpers.swift
sashaweiss-signal: Seems okay to me, given how we're using `lastViewedThread` elsewhere.

Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.swift
mlinington: > nit: since if this ends up `nil` we'll basically just return anyway, would prefer we do above as:

We may need to do some work if this ends up being nil. If there's no currently selected thread but the target thread is in the render state, then:
- `currentlySelectedThread?.uniqueId != targetThread.uniqueId` returns true
- `if let targetPath = tableDataSource.renderState.indexPath(...` returns a valid path
- The targetPath gets selected

Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.swift
mlinington: Updated!

Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.swift
sashaweiss-signal: Ah, yeah I mixed up my `==` and `!=`. We can leave this as-is I think.

sashaweiss-signal:
Thanks for the contribution! I'll take it from here, and get these changes into a commit in our internal workflows (maintaining authorship) so they can get into our next release.

sashaweiss-signal:
@mlinington, this was merged internally in `809a792`, which will make its way to public soon. Thank you!

Issue #5440, opened 2022-09-25
Lower voice message audio encoding bitrate to lower storage usage and be consistent with other Signal devices


Totorrr:
- [X] I have searched open and closed issues for duplicates
- [~] I am submitting a bug report for existing functionality that does not work as intended
- [~] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
This bug is the mirror bug of https://github.com/signalapp/Signal-Android/issues/12475 (please read that one to know more). 

Still, here is a brief explanation. 

Signal on iOS sends voice messages with audio encoded as AAC @ 128kbps. This is inconsistent with Android (AAC @ 36kbps approx.), and Signal-Desktop for Linux (MP3 @ 160kbps). Also, it makes a huge difference in storage usage to have a ratio of almost ~4 in bitrate. 

### My request:

 - Step 1, the quick fix: **please lower the AAC voice message bitrate used in the iOS app from 128kbps to ~36kbps** (or 48kbps) so that Android and iOS are on the same page (enough for voice & that can save a lot of storage space), see PR #5442 
 - Step 2, the bigger change: **please switch voice message encoding to Opus @ 16 kbps** (or between 16 and 32) from every device.

### See also

 - the reference bug report https://github.com/signalapp/Signal-Android/issues/12475 
 - https://github.com/signalapp/Signal-iOS/issues/4539 (to handle playing Opus audio file/message received on iOS)
 - https://github.com/signalapp/Signal-iOS/issues/3156 (suggesting using Opus for audio recording)

PR #5442, opened 2022-09-27
Lower AAC bitrate for voice message from 128 to 48 kbps

Totorrr:
Good enough quality and less space/network used.

Initial intent is be consistent with Android (~36kbps), but as this seems to be stereo, give it a little more, so 48kbps. This quick change would lower storage usage by a factor > 2.6. Setting it to mono with even lower bitrate can also be an option.

### Contributor checklist

- [ ] I have tested my contribution on these devices: => tiny change, to illustrate the issue.
 * iDevice A, iOS X.Y.Z
 * iDevice B, iOS Z.Y

- - - - - - - - - -

### Description

This contributes to #5440 which aims at lowering the bitrate used by voice messages Signal (and provide more consistency between platforms). That would end up in using less storage and network resources.

### DIFF
--- a/SignalUI/ViewModels/VoiceMessageModel.swift
+++ b/SignalUI/ViewModels/VoiceMessageModel.swift
@@ -131,7 +131,7 @@ public class VoiceMessageModel: NSObject {
                     AVFormatIDKey: kAudioFormatMPEG4AAC,
                     AVSampleRateKey: 44100,
                     AVNumberOfChannelsKey: 2,
-                    AVEncoderBitRateKey: 128 * 1024
+                    AVEncoderBitRateKey: 48 * 1024
                 ]
             )
             self.audioRecorder = audioRecorder


jonathancross:
Failing: https://github.com/signalapp/Signal-iOS/actions/runs/3132366048/jobs/5090578235

skylar-oler:
Extremely new to GitHub, sorry if this is the wrong place to put this.

I've actually always considered that high bitrate to be a large pro for Signal. I'm in a GC with a few musicians who like sending voice notes to collaborate, and it's been really nice that the recordings actually come across sounding good.

I certainly understand that it might be necessary, I just wanted to try to express that for some of the userbase, this isn't a negligible loss, so I hope there could be thought put into that before it's pushed.

Totorrr:
> I've actually always considered that high bitrate to be a large pro for Signal. I'm in a GC with a few musicians who like sending voice notes to collaborate, and it's been really nice that the recordings actually come across sounding good.

Hello @skylar-oler, 

Hmm, that's nice indeed. Even though, now, the behaviour in inconsistent between devices, because the Android app, for some reason, uses far lower bitrate when recording & sending audio than the iOS app. 

I would definitely support a high bitrate mode, that would be off by default, but could be enabled for use cases like yours.

skylar-oler:
Thanks for the response! I think that's a really great idea.

PR #5498, opened 2022-12-04
A handful of fixes for ConversationView gesture recognizers

mlinington:
- Fixes an issue where swipe-to-reply was broken for trackpads. This is from a conflict between the UINavigationController's interactive pop gesture and the collection view's pan gesture recognizer.

  Traditionally, the interactive pop gesture requires an edge swipe from the left edge. With indirect input like a trackpad, UIKit will also activate this gesture recognizer with a two-finger swipe anywhere in the view controller.

  The result here is that any initial rightward swipe would be recgonized as a pop gesture instead of a swipe-to-reply gesture. Reversing the gesture recognizer failure ordering fixed this.

- Constrain the recognition area for the swipe-to-reply pan gesture to only the cell content that gets translated during the reply gesture. Now, when the user pans over a bubble, it activates the swipe-to-reply gesture. But when the user pans over blank space, it's a interpreted as a navigation stack pop.

- Fixes an issue where swiping to reply on an audio message was difficult. The vertical recognition area of the audio waveform was occupying the entirety of audio message view. Now, it only recognizes pans on the waveform itself.

- Fixes a UI glitch with the interactive MessageDetail transition that occurs on iPad. The DetailNavController in the split view wasn't setting the SplitViewController as its delegate like the ChatListNavController. This interrupted a delegation chain from the ConversationVC to the MessageDetailVC in the DetailNavController. This only occurred on iPad since the compact size class collapses everything down to only use the ChatListNavController which was always properly configured.

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/CV/CVComponents/CVComponentMessage.swift
+++ b/Signal/src/ViewControllers/ConversationView/CV/CVComponents/CVComponentMessage.swift
@@ -2079,11 +2079,13 @@ public class CVComponentMessage: CVComponentBase, CVRootComponent {
                                                            renderItem: renderItem,
                                                            messageSwipeActionState: messageSwipeActionState) {
             return panHandler
+        } else if componentView.contentViewSwipeToReplyWrapper.containsGestureLocation(sender) {
+            return CVPanHandler(delegate: componentDelegate,
+                                panType: .messageSwipeAction,
+                                renderItem: renderItem)
+        } else {
+            return nil
         }
-
-        return CVPanHandler(delegate: componentDelegate,
-                            panType: .messageSwipeAction,
-                            renderItem: renderItem)
     }
 
     public override func startPanGesture(sender: UIPanGestureRecognizer,
--- a/Signal/src/ViewControllers/ConversationView/Cells/AudioMessageView.swift
+++ b/Signal/src/ViewControllers/ConversationView/Cells/AudioMessageView.swift
@@ -426,7 +426,7 @@ class AudioMessageView: ManualStackView {
         }
 
         let locationInSlider = convert(point, to: waveformProgress)
-        return locationInSlider.x >= 0 && locationInSlider.x <= waveformProgress.width
+        return waveformProgress.bounds.contains(locationInSlider)
     }
 
     func progressForLocation(_ point: CGPoint) -> CGFloat {
--- a/Signal/src/ViewControllers/ConversationView/ConversationViewController+GestureRecognizers.swift
+++ b/Signal/src/ViewControllers/ConversationView/ConversationViewController+GestureRecognizers.swift
@@ -42,13 +42,6 @@ extension ConversationViewController: UIGestureRecognizerDelegate {
 
         // Allow panning with trackpad
         if #available(iOS 13.4, *) { collectionViewPanGestureRecognizer.allowedScrollTypesMask = .continuous }
-
-        // There are cases where we don't have a navigation controller, such as if we got here through 3d touch.
-        // Make sure we only register the gesture interaction if it actually exists. This helps the swipe back
-        // gesture work reliably without conflict with audio scrubbing or swipe-to-repy.
-        if let interactivePopGestureRecognizer = navigationController?.interactivePopGestureRecognizer {
-            collectionViewPanGestureRecognizer.require(toFail: interactivePopGestureRecognizer)
-        }
     }
 
     // TODO: Revisit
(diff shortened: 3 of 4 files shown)


mlinington:
Apologies for the terrible quality. Had to super compress these to fit them in Github's 10MB cap

Before: 
https://user-images.githubusercontent.com/3432567/205477760-892282e2-bb22-47df-a129-312633547c1a.mp4

After:
https://user-images.githubusercontent.com/3432567/205477767-176c4133-6936-40b9-af33-8c8935538f02.mp4

Signal/src/ViewControllers/ConversationView/CV/CVComponents/CVComponentMessage.swift
```diff
@@ -2079,11 +2079,13 @@ public class CVComponentMessage: CVComponentBase, CVRootComponent {
                                                            renderItem: renderItem,
                                                            messageSwipeActionState: messageSwipeActionState) {
             return panHandler
+        } else if componentView.contentViewSwipeToReplyWrapper.containsGestureLocation(sender) {
```
mlinington, line 2082: There's probably a better way to do this like maybe installing the gesture recognizer on the bubble itself or something. That's a larger change than I'm wanting to make so I think this is a pretty good heuristic.

The contentViewSwipeToReplyWrapper appears to contain all of the message content that's translated in the reply swipe. I think it makes sense to use that as the gesture target? Anything outside of that area should not be recognized.

Signal/src/ViewControllers/ConversationView/Cells/AudioMessageView.swift
```diff
@@ -426,7 +426,7 @@ class AudioMessageView: ManualStackView {
         }
 
         let locationInSlider = convert(point, to: waveformProgress)
-        return locationInSlider.x >= 0 && locationInSlider.x <= waveformProgress.width
+        return waveformProgress.bounds.contains(locationInSlider)
```
mlinington, line 429: There is another bug here that I thought about looking at but decided against it since it might require design input. Currently, the location of the scrubber isn't consulted here. This means that if the progress is at 0% but you start panning at 90% the scrubber will immediately jump to 90%. Kinda weird.

Here's what I think might make sense but feel free to ignore. Just my $0.02:
- If you are physically touching the screen to drag, you should start the pan reasonably close to the scrubber. Ignore anything that's outside of some target area.
- If you're panning with the trackpad, you can start a pan anywhere in the waveform but the scrubber will move linearly with the pan distance. It shouldn't immediately jump to the cursor location.

Signal/src/ViewControllers/ConversationView/ConversationViewController+GestureRecognizers.swift
```diff
@@ -42,13 +42,6 @@ extension ConversationViewController: UIGestureRecognizerDelegate {
 
         // Allow panning with trackpad
         if #available(iOS 13.4, *) { collectionViewPanGestureRecognizer.allowedScrollTypesMask = .continuous }
-
-        // There are cases where we don't have a navigation controller, such as if we got here through 3d touch.
-        // Make sure we only register the gesture interaction if it actually exists. This helps the swipe back
-        // gesture work reliably without conflict with audio scrubbing or swipe-to-repy.
-        if let interactivePopGestureRecognizer = navigationController?.interactivePopGestureRecognizer {
-            collectionViewPanGestureRecognizer.require(toFail: interactivePopGestureRecognizer)
```
mlinington, line 50: This priority is inverted. I'm not sure if this used to work or this is just due to the very confusing UIKit API. The pan gesture recognizer should be prioritized over the interactivePop.

I decided against flipping the receiver/argument since it felt weird tweaking settings on a gesture recognizer we don't own. I accomplished something similar with a UIGestureRecognizerDelegate method below.

Signal/src/ViewControllers/ConversationView/ConversationViewController+GestureRecognizers.swift
```diff
@@ -72,11 +65,18 @@ extension ConversationViewController: UIGestureRecognizerDelegate {
...
+        if gestureRecognizer === collectionViewPanGestureRecognizer {
             // Only allow the pan gesture to recognize horizontal panning,
             // to avoid conflicts with the conversation view scroll view.
             let translation = collectionViewPanGestureRecognizer.translation(in: view)
-            return abs(translation.x) > abs(translation.y)
+            guard abs(translation.x) > abs(translation.y) else { return false }
+
+            // If there's nothing available to handle the pan, don't bother recognizing.
+            // This allows other recognizers like the navigationController popGesture to begin.
+            guard findPanHandler(sender: collectionViewPanGestureRecognizer) != nil else { return false }
```
mlinington, line 76: I don't love this, but it accomplishes what it needs to without changing a bunch of code. Basically, the pan should win if there's something to pan. Otherwise, it should defer elsewhere and allow another recognizer (e.g. interactive pop) to win.

PR #5583, opened 2023-05-03
Show anchored emoji picker search results first

skjiisa:
### Contributor checklist

 * 2018 iPad Pro 11-inch, iOS 16.2

- - - - - - - - - -

### Description

Shows anchored emoji search results before unanchored results. "Anchored" in this case referring to results where the text matches the search term from the start, as opposed to elsewhere in the text. This is done to make it easier for the user to find the emoji they are looking for. For example, there are currently 56 emoji that appear when searching for "rat". If the user wants the rat emoji (🐀) this change causes it to be sorted to the top.
Rat before after

Other examples where this helps:

| "Ita"  | "Cat" |
| ----- | ------ |
| Ita before after  | Cat  |
| Shows union jack before Italian flag  | Cats are disconnected and far down the list  |

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/Emoji Picker/EmojiPickerCollectionView.swift
+++ b/Signal/src/ViewControllers/ConversationView/Emoji Picker/EmojiPickerCollectionView.swift
@@ -220,15 +220,30 @@ class EmojiPickerCollectionView: UICollectionView {
             return []
         }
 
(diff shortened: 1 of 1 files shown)


sashaweiss-signal:
@skjiisa, this is merged in 83196cb2e8d59c7ff442c54419852d9416b5f62a. Thanks again!

PR #5586, opened 2023-05-04
Sort images by recent instead of creation date

skjiisa:
### Contributor checklist

 * 2018 iPad Pro 11-inch, iOS 16.4.1

- - - - - - - - - -

### Description

Photos in the app are currently always sorted by the creation date. While this is how the "Library" tab of the photos app works, this is not how the rest of Photos app works and can make it difficult to find the right image. The "Recents" album in the photos app sorts by most recently added, not most recently created, which can be different for certain images downloaded from the internet, or can be manually set by the user.

#### Issues
+ If a user downloads a new image or video and it has metadata about a creation date in the past, it will not appear at the front of the Recents picker in Signal as the user would expect.
+ If a user manually sorts an album, the manual sorting does not appear in the Signal photo picker.

#### Solution

By removing the sort descriptors from asset fetch, PhotoKit will return assets sorted the same way they are in the Photos app.

#### Examples

The photos with a year printed have their creation date metadata set to that year, but that is not the order they were saved to the device.

| Before  | After |
| --- | --- |
| Before Recents | After Recents |
| Before album | After album |

#### Notes on code changes

None of the [PHAsset fetch options](https://developer.apple.com/documentation/photokit/phfetchoptions) are for sorting "the way the Photos app does". The way to do that is to have no sort descriptors on the fetch request. But once the sort descriptor is removed, we can't pass `ascending` into it anymore, which means we have to manually flip the results if we want to sort by descending (the changes to `PhotoCollectionContents.asset(at:)`). _That_ then means that setting a fetch limit will be limiting the results from the wrong end of the results, so the fetch limit also has to be handled by `PhotoCollectionContents`. Based on the [documentation for `PHFetchResult`](https://developer.apple.com/documentation/photokit/phfetchresult), it caches accessed `PHAssets` which implies that they are fetched lazily, so not having a limit on the original fetch and instead limiting the accessed assets inside of `PhotoCollectionContents` should not have much of a performance hit.

I decided to make `PhotoCollectionContents.fetchResult` private as it was not accessed anywhere else in the app, and it could lead to confusion if it were to be, as the assets would not be sorted descending as expected if it were initialized with `ascending` as false. While I was at it, I removed `localizedTitle` which was never used.

### DIFF
--- a/Signal/src/ViewControllers/Photos/MediaControls.swift
+++ b/Signal/src/ViewControllers/Photos/MediaControls.swift
@@ -1270,12 +1270,8 @@ class MediaPickerThumbnailButton: UIButton {
 
         // Async Fetch last image
         DispatchQueue.global(qos: .userInteractive).async {
-            let fetchOptions = PHFetchOptions()
-            fetchOptions.sortDescriptors = [NSSortDescriptor(key: "creationDate", ascending: false)]
-            fetchOptions.fetchLimit = 1
-
-            let fetchResult = PHAsset.fetchAssets(with: PHAssetMediaType.image, options: fetchOptions)
-            if fetchResult.count > 0, let asset = fetchResult.firstObject {
+            let fetchResult = PHAsset.fetchAssets(with: PHAssetMediaType.image, options: nil)
+            if let asset = fetchResult.lastObject {
                 let targetImageSize = CGSize(square: MediaPickerThumbnailButton.visibleSize)
                 PHImageManager.default().requestImage(for: asset, targetSize: targetImageSize, contentMode: .aspectFill, options: nil) { (image, _) in
                     if let image = image {
--- a/Signal/src/ViewControllers/Photos/PhotoLibrary.swift
+++ b/Signal/src/ViewControllers/Photos/PhotoLibrary.swift
@@ -78,8 +78,9 @@ class PhotoPickerAssetItem: PhotoGridItem {
 
 class PhotoCollectionContents {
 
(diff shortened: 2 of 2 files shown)


sashaweiss-signal:
Thank you again for the contribution, @skjiisa! I checked with our design team, and they confirmed this is something we'd like to change. I did a quick audit and I think your change will cover both the full-screen photo picker and the "photo strip" in `AttachmentKeyboard`, which is great.

The one other place we'll need to handle is the image thumbnail on the in-app camera view, specifically `MediaPickerThumbnailButton.configure()`. I'm pretty sure the change will be similarly simple there. Would you mind updating to handle that site, and confirming the behavior is as we'd expect in those three sites?

1. Full screen photo picker
2. Photo strip
3. In-app camera thumbnail

Thanks again.

skjiisa:
Updated the camera view thumbnail to show the most recent image. I double checked by searching the project for other `PHFetchOptions` and the only other instance of it is for sorting albums, not photos, so this should be everything 👍.

Screenshots (taken on a simulator, but I did test on a physical device too):
| Before  | After |
| --- | --- |
| Before | After |

sashaweiss-signal:
Awesome, thanks @skjiisa! Similarly to the other PRs, I'll post here when the internal commit goes public. (When I make the internal commit I'll preserve your authorship!)

sashaweiss-signal:
@skjiisa, this is merged in 6271655. Thanks again!

PR #5628, opened 2023-07-17
[ISSUE-5608] fix stuck after iPad keyboard dismiss

### LINKED ISSUE #5608: Signal could get stuck and needs to be killed when taking a photo inside the app
- [ x] I have searched open and closed issues for duplicates
- [x ] I am submitting a bug report for existing functionality that does not work as intended
- [ x] This isn't a feature request or a discussion topic

### Bug description
Tested on iPad, not on iPhone.
When taking a photo inside the app, it could enter a state which it cannot leave. The app needs to be killed to continue.

### Steps to reproduce
- Take a photo inside the app with the photo button next to the message input area.
- Tap the pencil button in the lower left corner
- One may draw stuff on the photo here. In the lower left corner there is now an X button to leave and discard; in the lower right corner there is an r button to confirm and leave.
- Tap the Aa button in the lower center. Now one may enter text onto the photo.
- Close the keyboard with the keyboard button in the lower right.
- The app is stuck now. One may continue to draw stuff onto the image, but there is now way to leave this stage. The lower button bar is gone. The only thing left to do is to kill the app, which also forfeits the photo taken.

The app is stuck (not dead), but it needs to be killed because there is no way to send the photo or continue using the app.

After closing the keyboard the lower button bar should be recreated to be able to leave this stage and send the photo.

### Device info

**Device**: iPad 8

**iOS version**: 16.5

**Signal version:** 6.25.0.13

Serguitar:

### Contributor checklist

 * Simulator iPad Pro (11-inch) A, iOS 16.4

- - - - - - - - - -

### Description
After closing the keyboard using the keyboard button (bottom right button on the iPad), the ImageEditorBottomBar was not being shown again. As a result, the user was unable to navigate away from the screen and had to force quit the app. 
`fixes #5608`

### Video
Before the fix
https://drive.google.com/file/d/1etIlN4yJWoij3qSezFRRU4goN_laRluO/view?usp=sharing

After the fix
https://drive.google.com/file/d/1EcWw6es3bOaMOzStyAnDGm4UKJWjd91G/view?usp=sharing

### Screenshots
Simulator Screenshot - iPad Pro (11-inch) (4th generation) - 2023-07-17 at 11 15 48

### DIFF
--- a/SignalUI/Views/ImageEditor/ImageEditorViewController+Text.swift
+++ b/SignalUI/Views/ImageEditor/ImageEditorViewController+Text.swift
@@ -294,6 +294,11 @@ extension ImageEditorViewController: UITextViewDelegate {
             self.textViewContainer.alpha = 0
         }
     }
+    
+    func textViewShouldEndEditing(_ textView: UITextView) -> Bool {
+        finishTextEditing(applyEdits: true)
+        return true
+    }
 }
 
 // MARK: - ImageEditorViewDelegate


Serguitar:
up

PR #5703, opened 2023-11-16
Fix for constraint SafeArea loop crash.

johnsonajason:

### Contributor checklist
 * iPhone 15 Pro, iOS 17.0
 * iPhone 15 Pro Max, iOS 17.0
 * iPhone 15 SE, iOS 17.0

- - - - - - - - - -

### Description
This is to fix a crash that occurs on specific devices in landscape. There was an edge case  issue with constraints and safeArea causing the views layout to loop and eventually crash.

Steps to reproduce the crash:
- Use iOS 16.6.1 or 17.0 (I did not test other versions).
- Use a Max device (ex. iPhone 12 Pro Max or iPhone 15 Pro Max)
1) Rotate the device to be in landscape
2) Compose a new message from the Chat Tab
3) Tap New Group
4) Tap Skip
5) Enter a name for the group (The name does not matter)
6) Tap Create
7) Tap "Enable and Share Link"

The share sheet will not present and memory will constantly grow. The app will eventually crash.

To Verify the reason for the crash you can add `-UIViewLayoutFeedbackLoopDebuggingThreshold 50` to the apps run arguments and set an all exceptions breakpoint with the following action `po [_UIViewLayoutFeedbackLoopDebugger layoutFeedbackLoopDebugger]`

Issue:
The constraints on the actionSheets scrollView subviews would adjust to to avoid the safeArea, this in turn would resize the scrollview. In most cases this works correctly but if the size of the subview was within a small range of the maximum scrollView size a constraint loop would occur where resizing the scrollView would relayout the subviews which would no longer be affected by the safeArea causing the scrollView to resize again. This would continue to infinitely resize these views in a layout loop.

Solution:
To solve this issue I moved the handling of the safeArea to the scrollView and altered a few layout parameters to compensate for this change.

### DIFF
--- a/SignalUI/ViewControllers/ActionSheetController.swift
+++ b/SignalUI/ViewControllers/ActionSheetController.swift
@@ -151,11 +151,16 @@ open class ActionSheetController: OWSViewController {
         // Depending on the number of actions, the sheet may need
         // to scroll to allow access to all options.
         view.addSubview(scrollView)
+        
+        // Set the top margin for the scrollview to avoid hugging the top of the view
+        let topMargin: CGFloat = 59
         scrollView.clipsToBounds = false
         scrollView.showsVerticalScrollIndicator = false
-        scrollView.autoPinEdge(toSuperviewEdge: .bottom)
+        
+        // Adjust ScrollView to fit within the safeArea
+        scrollView.autoPinEdge(toSuperviewSafeArea: .bottom)
         scrollView.autoHCenterInSuperview()
-        scrollView.autoMatch(.height, to: .height, of: view, withOffset: 0, relation: .lessThanOrEqual)
+        scrollView.autoMatch(.height, to: .height, of: view, withOffset: -topMargin, relation: .lessThanOrEqual)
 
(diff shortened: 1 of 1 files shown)


Miguel-Signal:
Hi @johnsonajason, thanks for bringing this crash to our attention.

I've gone ahead and filed this with our engineering team. While our engineers are working on a fix, I unfortunately can't give out a specific timeline for a resolution.

gabefair:
@johnsonajason  is this bug still an issue?

ExtRIELICi:
@gabefair It appears to be fixed now, although I’ve noticed a bug when I rotate back to vertical mode after completing the entire process. The back gesture doesn’t seem to work, but tapping on the “<“ key takes me back just fine. So, it’s not entirely fixed, but it’s significantly less buggy. I’m using iOS 26, if that matters.

johnsonajason:
I retested on a iPhone 17 Pro Max, iOS 26.0.1, app version 7.8.2 and I was not able to reproduce the issue.

PR #5772, opened 2024-03-16
Double tap message to show context menu

kevinmbeaulieu:

### Contributor checklist

 * iPhone 14 Pro, iOS 17.4

- - - - - - - - - -

### Description

In the conversation view, long pressing a message presents the context view for that cell, allowing the user to add reactions, reply, delete the message, etc. To match the behavior that iOS users are accustomed to in iMessage, Signal should also allow double-tapping messages to present this menu. This can also be faster, since double-tapping doesn't have the same time delay that long press gestures have.

The code changes implemented all followed the same patterns as the tap and long press gesture recognizers on that view, so even though it's a few hundred lines it's mostly copy-paste from that and just tweaking the relevant parts.

### DIFF
--- a/Signal/src/ViewControllers/ConversationView/CV/CVCell.swift
+++ b/Signal/src/ViewControllers/ConversationView/CV/CVCell.swift
@@ -334,6 +334,21 @@ public extension CVRootComponentHost {
                                                              renderItem: renderItem)
     }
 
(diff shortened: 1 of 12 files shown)


tohrxyz:
can't comment on the code, but this would be a good feature to have for sure

great work

PR #5848, opened 2024-08-08
Add Mutipath TCP (MPTCP) support

Aperence:

### Contributor checklist

 * iDevice A, iOS X.Y.Z
 * iDevice B, iOS Z.Y

- - - - - - - - - -

### Description

MPTCP is a TCP extension allowing to improve network reliabilty by using mutiple network interface for the same TCP connection. Check https://www.mptcp.dev and https://developer.apple.com/documentation/foundation/urlsessionconfiguration/improving_network_reliability_using_multipath_tcp for details. Changes to this repository consists in setting the multipathServiceType of URLSessionConfiguration to Handover, allowing a seamless transition of interface in case of deterioration of the connection.

### DIFF
--- a/SignalServiceKit/Network/ContentProxy.swift
+++ b/SignalServiceKit/Network/ContentProxy.swift
@@ -25,6 +25,7 @@ public class ContentProxy: NSObject {
             "HTTPSProxy": proxyHost,
             "HTTPSPort": proxyPort
         ]
+        configuration.multipathServiceType = .handover
         return configuration
     }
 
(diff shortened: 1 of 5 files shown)


211217613:
Would this change help call quality when moving away from WiFi to cellular data?

Whenever I move from WiFi access to cellular data i always get a "signal reconnecting" message.

Aperence:
As long as the calls use an URLSession with the MultipathServiceType set to Handover mode (which is the suggested changes here), yes there should be an automatic transition to the cellular interface if the Wi-Fi connection becomes poor. The user won’t even notice that the application changed of interface (so there shouldn’t be anymore a message like the connection was lost if you have cellular access).

Note however that it requires server support, so in the end it will also depend about wether the server support mptcp or not (it is suggested to enable it on server side also by default https://www.mptcp.dev/faq.html#why--when-should-mptcp-be-enabled-by-default)

matttbe:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

It looks like this PR didn't get any reviews, but it was not blocked because the developer had to change something.

Issue #5846, opened 2024-08-02
"Something went wrong" after captcha


lillian-b:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
I am presented with a captcha on registration. After I complete the captcha, I get a "something went wrong" popup, in the logs I can see the Signal server returning a 403 to the captcha.

I am not using a VPN, but I have tried using my cell connection to register and my home internet.

-> This might be an issue specific to my phone number, I tried emailing support@signal.org last week but did not hear back.

**Actual result:** "Something went wrong" popup.

**Expected result:** it sends me a SMS code to complete registration.

### Screenshots
![image](https://github.com/user-attachments/assets/4b968204-2d2f-4f4f-b58d-8b4772cdbf0e)

### Device info
**Device**: iPhone SE 3rd generation

**iOS version**: 17.5.1

**Signal version:** 7.22

### Link to debug logs
https://debuglogs.org/ios/7.22.0/2c14f6d1deb45ab90a1157b7379da48bfb962fa3f0e81401410cd1364b8a2a2a.zip

Notably the below:

```
2024/08/02 22:44:37:422  💛 [RegistrationCoordinatorImpl.swift:2464 submit(challengeFulfillment:for:retriesLeft:)]: Submitting CAPTCHA challenge fulfillment
2024/08/02 22:44:37:423  💛 [RegistrationNavigationController.swift:57 pushNextController(_:loadingMode:)]: Pushing loading controller
2024/08/02 22:44:38:185  🧡 [OWSUrlSession.swift:423 baseCompletionPromise(requestConfig:responseData:monitorId:)]: Request failed: HTTP 403; <OWSHttpHeaders: [content-length; content-type; date; vary; x-signal-
timestamp: 1722638678102]>; https://chat.signal.org/v1/verification/session/l_t7o75VSi2_fswVPY2NAw==; nil
2024/08/02 22:44:38:186  🧡 [OWSUrlSession.swift:568 promiseForTSRequest(_:)]: Failure: { PATCH: v1/verification/session/[REDACTED] }, error: HTTP 403; <OWSHttpHeaders: [content-length; content-type; date; vary;
 x-signal-timestamp: 1722638678102]>; https://chat.signal.org/v1/verification/session/l_t7o75VSi2_fswVPY2NAw==; nil
[...]

PR #5865, opened 2024-09-04
[#5846] Remove custom user agent to fix captcha failure loop

g1a55er:
### Contributor checklist
 * iPhone 15 Pro, iOS 17.6.1

- - - - - - - - - -

### Description
I was also hit by the bug filed as #5846, where I get stuck in an infinite "Something went wrong" captcha failure loop on registration with notifications unavailable. 

After [some debugging](https://github.com/signalapp/Signal-iOS/issues/5846#issuecomment-2329881697), comparing iOS and Android, it turns out this traces back to hCaptcha not liking the custom user agent of the WebView used for the CAPTCHA view 

Once I removed the custom user agent so that the WebView goes back to the default user agent, I am able to successfully get out of the failure loop.

### DIFF
--- a/SignalUI/Views/CaptchaView.swift
+++ b/SignalUI/Views/CaptchaView.swift
@@ -71,7 +71,6 @@ public class CaptchaView: UIView {
 
         let webView = WKWebView(frame: .zero, configuration: configuration)
         webView.allowsBackForwardNavigationGestures = false
-        webView.customUserAgent = OWSURLSession.userAgentHeaderValueSignalIos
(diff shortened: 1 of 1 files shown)


luoye1101:
I deleted the custom user agent in your way, but the problem still exists

OscillateWildly:
I'd like to raise the profile of this proposed fix, as if this does indeed resolve the described issue, it will resolve an inability to change phone numbers currently being experienced/reported by myself and others.

PR #5884, opened 2024-10-01
Replace recursive call with while loop to prevent stack overflow.

VasilErema:

### Contributor checklist

 * Simulator iPhone 16 PRO , iOS 18.0

- - - - - - - - - -

### Description
- Recursive call might cause stack overflow if called to many times.
- Improvement: replace recursive call with while loop, that won't be affected by the issue.

P.S. I had no stack overflow with old/new code when tested - there was no recursive calls. I picked it as a small first contribution.

### DIFF
--- a/Signal/Attachments/AttachmentDownloadRetryRunner.swift
+++ b/Signal/Attachments/AttachmentDownloadRetryRunner.swift
@@ -81,10 +81,10 @@ public class AttachmentDownloadRetryRunner {
         }
 
         private func run() async {
-            do {
-                self.isRunning = true
-                defer { self.isRunning = false }
-
+            self.isRunning = true
+            defer { self.isRunning = false }
+            // Run while there is a next timestamp.
+            while true {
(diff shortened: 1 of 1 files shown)


Signal/Attachments/AttachmentDownloadRetryRunner.swift
```diff
@@ -81,10 +81,10 @@ public class AttachmentDownloadRetryRunner {
...
 
         private func run() async {
-            do {
-                self.isRunning = true
-                defer { self.isRunning = false }
-
+            self.isRunning = true
+            defer { self.isRunning = false }
+            // Run while there is a next timestamp.
+            while true {
```
ankushkushwaha, line 87: This loop never breaks?

VasilErema: It can break. If there is no more timestamps to handle it returns on line 92. 
Functionally it is the same as before.

ankushkushwaha (approved):

PR #5959, opened 2025-02-04
#5837 Disappearing Navigation Bar

### LINKED ISSUE #5837: Chat Name Field Vanishes on iPad

### Bug description
On some stale app launches (opening the app after a previous fresh launch), the name for the active chat window is missing at the top. Attached a screenshot of what this looks like. Note that I have not experienced this on iPhone/iOS.

### Steps to reproduce
- Restart iPad and open Signal, open a chat, then leave the app without force exiting.
- A while later, reopen the app. If Signal didn’t close in the background, you might see the chat with a missing chat name.

On some stale launches, the chat name is missing. This does not happen every time. A workaround to get it displayed again is to switch to another chat, and then switch back to the previously open one.

The chat name is displayed at the top as expected.

### Screenshots

IMG_1231

### Device info

**Device**: iPad (9th generation)

**iOS version**: 17.5.1

**Signal version:** 7.19.1 (208)

### Link to debug log

Does not occur every time.

MarlowBrown:

### Contributor checklist

 * iPhone 16, iOS 18.2
 * iPad Pro 13-inch M4, iOS 18.2

- - - - - - - - - -

### Description

This proposal `fixes #5837` by provided a workaround the navigation bar disappearing. Exploring the codebase, there is no location where the navigation bar title view is hidden; however, as explained in the code comment in this PRs only commit, I am confident that this is a UIKit bug. This PR is a workaround until that UIKit bug is patched.

### DIFF
--- a/Signal/ConversationView/ConversationViewController+UI.swift
+++ b/Signal/ConversationView/ConversationViewController+UI.swift
@@ -13,6 +13,12 @@ extension ConversationViewController {
 
         self.title = nil
 
+        //HACK: If you look at issue #5837, the navigation title view is disappering. You can confirm this by looking at the view hierarchy; however, the navigation titleview says that its title view is not nil. Setting active breakpoints in this function prevents the bug from happening. I'm confident that this is a UIKit bug. To get around this we simply remove and add the headerview back again. This seems to only happen with iPad
+        DispatchQueue.main.async { [weak self] in
+            self?.navigationItem.titleView = nil
+            self?.navigationItem.titleView = self?.headerView
+        }
+        
         if thread.isNoteToSelf {
             headerView.titleIcon = Theme.iconImage(.official)
             headerView.titleIconSize = 16


sashaweiss-signal:
I'm not wild about adding this workaround.

I can't trivially reproduce the issue on iOS 18.3 by rotating to landscape and back, so clearly something unreliable is going on. There's already a lot of unstable and hacky things already around Signal's nav bars, and I'm not inclined to add yet another instance without understanding the root cause. I worry that this will create more subtle problems down the line.

Signal/ConversationView/ConversationViewController+UI.swift
```diff
@@ -13,6 +13,12 @@ extension ConversationViewController {
 
         self.title = nil
 
+        //HACK: If you look at issue #5837, the navigation title view is disappering. You can confirm this by looking at the view hierarchy; however, the navigation titleview says that its title view is not nil. Setting active breakpoints in this function prevents the bug from happening. I'm confident that this is a UIKit bug. To get around this we simply remove and add the headerview back again. This seems to only happen with iPad
```
ankushkushwaha, line 16:
```suggestion
        // HACK: If you look at issue #5837, the navigation title view is disappearing. You can confirm this by looking at the view hierarchy; however, the navigation titleview says that its title view is not nil. Setting active breakpoints in this function prevents the bug from happening. I'm confident that this is a UIKit bug. To get around this we simply remove and add the headerview back again. This seems to only happen with iPad
```

Signal/ConversationView/ConversationViewController+UI.swift
```diff
@@ -13,6 +13,12 @@ extension ConversationViewController {
...
+        DispatchQueue.main.async { [weak self] in
```
ankushkushwaha, line 17: Do we really need [weak self] here ?

sashaweiss-signal:
Unfortunately, I don't think this is the way we'd want to address this, so I'm going to close this PR. Thanks for looking into it, and I'll keep this approach in mind the next time I take a look into this issue.

PR #5965, opened 2025-02-09
Clamped Disppearing Messages to 4 Weeks

### LINKED ISSUE #5947: Disappearing messages: setting them above 4 weeks create inconsistent behavior

### Bug description
Disappearing messages: I understand that it's not possible to set the disappearing message to a value above 4 weeks ([official doc source](https://support.signal.org/hc/en-us/articles/360007320771-Set-and-manage-disappearing-messages)).
On the custom parameter view, turning the number wheel of that parameter to a value above 4 **and then** the unit wheel to weeks always makes the number wheel back to 4.
**However**, I noticed that putting the value at 15 and the unit wheel on weeks **and then** going back to the conversation view, we are told that disappearing messages will last 15 weeks.

### Steps to reproduce
- Go on the disappearing messages setting panel
- Click on "custom value"
- Put the number wheel on 15 (arbitrary, above 4)
- Put the unit values to weeks

**Actual result:** 
The number is pushed back to 4 by the app.
But when back to the conversation, a small message says that disappearing messages will last **15 weeks**.
Not sure which is the actual truth but the message is different from the setting.

**Expected result:** 
Disappearing messages should last 4 weeks, as per the official documentation.

### Screenshots

In French

https://github.com/user-attachments/assets/6ff2d8c9-46a3-4579-bd39-8813e3b1267a

### Device info

**Device**: iPhone 13 Mini

**iOS version**: 18.2.1

**Signal version:** 7.4.3 (573)

### Link to debug log

https://debuglogs.org/ios/7.43.0/56d00ff58e5f1df9cbe6bbdedddac8348e1adfc71ea951db8be08951235b4aff.zip

MarlowBrown:
### Contributor checklist

 * iPhone 16, iOS 18.2

- - - - - - - - - -

### Description

This pull request `fixes #5947` by clamping an internal value in the Disappearing Messages Timers picker to 4 weeks.

### DIFF
--- a/Signal/src/ViewControllers/ThreadSettings/DisappearingMessagesTimerSettingsViewController.swift
+++ b/Signal/src/ViewControllers/ThreadSettings/DisappearingMessagesTimerSettingsViewController.swift
@@ -317,9 +317,17 @@ private class CustomTimePicker: UIPickerView, UIPickerViewDataSource, UIPickerVi
         didSet {
             guard oldValue != selectedUnit else { return }
             reloadComponent(Component.duration.rawValue)
+            clampSelectedTimeIfSelectedUnitIsWeeks(selectedUnit)
             durationChangeCallback(selectedDuration)
         }
     }
+    
+    private func clampSelectedTimeIfSelectedUnitIsWeeks(_ unit: CustomTimePicker.Unit) {
+        if unit.interval == .week {
+            selectedTime = unit.maxValue
+        }
+    }
+    
     var selectedTime: Int = 1 {
         didSet {
             guard oldValue != selectedTime else { return }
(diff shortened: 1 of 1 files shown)


sashaweiss-signal:
Hi @MarlowBrown,

While I'm guessing this workaround works for the specific case of handling too-many weeks, it feels like we shouldn't need to be clamping. Instead, it feels like when the unit is changed (e.g., from days to weeks), and the maximum selectable number of units is clamped in the UI (e.g., from some number of days to "4", for weeks), that the new "selected" unit should be propagated back to the view controller by the `CustomTimePicker`.

MarlowBrown:
Update for you @sashaweiss-signal 

I changed the code to avoid clamping. The change is a little weird as technically until the user has moved the picker wheel, it doesn't count as "selected". I think it might be by design, so the code for force selecting is warranted.

dihv:
Looks good to me

sashaweiss-signal:
Hey @MarlowBrown, thanks for the contribution. I ended up rebuilding this view, and in the process addressed the clamping issue.

The rebuild is in 65dccf9f42, which should become public with an upcoming release.

MarlowBrown:
Sounds good. Thank you for keeping me updated! Check out my other PRs btw ;)

PR #5977, opened 2025-02-20
Properly wait for dismissal before presenting a new view controller

### LINKED ISSUE #5976: Disappearing Bottom Controls in ImageEditorCropViewController

### Bug description
There appears to be a very rare case where the bottom controls when cropping an image can entirely disappear, leaving you unable to get out of the crop tool without force closing the application. Behavior seems almost like a race condition in that it only happens very rarely and is difficult to reliably reproduce. Somehow the buttons are getting stuck in the hidden state I think, and is related to the transitionUI function.

I've also had this happen with the markup tool which appears to use the same animation methods for the bottom row of buttons.

### Steps to reproduce
- Open and close the crop editor and spam inputs. Really tough to get it to happen. It will naturally happen to me about once a day or every other day.

**Actual result:** Navigation controls will sometimes be absent, leaving the user unable to exit the crop tool.

**Expected result:** Navigation controls should ALWAYS be present when in the crop tool. Disabling animations seems to fix this, but I'd like to come up with a solution that preserves the animations and prevents the potential race condition.

### Device info

**Device**: iPhone 16 Pro Max

**iOS version**: 18.3.1 (and previous releases)

**Signal version:** 7.47 (625)

### Link to debug log

https://debuglogs.org/ios/7.47.0/e321bc0df8a08148136ed43d6dcc607a27ca9f8e02c3db11681d37fd8cc74dd8.zip

This error stands out to me in relation to the issue: 2025/02/20 00:54:25:091  ERR❤️ [AttachmentPrepViewController.swift:221 presentFullScreen(viewController:)]: Already has presented view controller. [<SignalUI.ImageEditorCropViewController: 0x127aef700>]

ChronicLynx:

### Contributor checklist

 * iPhone 16 Pro (Xcode Simulator - 18.2)

- - - - - - - - - -

### Description
Potentially fixes #5976 

My theory, given the error in #5976, is that because the original code failed to wait for the dismissal to complete, it would occasionally try to present a new view while an existing view was present. This caused the animation for the navigation buttons in ImageEditorCropViewController and similar image editing views to be left in a state with no bottom navigation buttons. Therefore, the user has to force quit the app in order to exit the view. I think this may stabilize that flow and prevent such situations from occurring.

ImageEditorCropViewController:

I would imagine when encountering the error

`2025/02/20 00:54:25:091 ERR❤️ [AttachmentPrepViewController.swift:221 presentFullScreen(viewController:)]: Already has presented view controller. [<SignalUI.ImageEditorCropViewController: 0x127aef700>`

that it possibly short circuits or resets the logic responsible for summoning the bottom buttons. I was originally fixated on the logic summoning the buttons until I noticed a seemingly correlated error with the occurrence of the bug in AttachmentPrepViewController.

```
override func viewDidAppear(_ animated: Bool) {
        super.viewDidAppear(animated)

        transitionUI(toState: .final, animated: true)
    }
```

```
private func transitionUI(toState state: UIState, animated: Bool, completion: ((Bool) -> Void)? = nil) {
        let layoutGuide: UILayoutGuide = {
            switch state {
            case .initial: return initialStateContentLayoutGuide
            case .final: return finalStateContentLayoutGuide
            }
        }()

        let hideControls = state == .initial
        let setControlsHiddenBlock = {
            let alpha: CGFloat = hideControls ? 0 : 1
            self.footerView.alpha = alpha
            self.cropView.setState(state == .initial ? .initial : .normal, animated: false)
            self.bottomBar.setControls(hidden: hideControls)
        }

        let animationDuration: TimeInterval = 0.15

        let imageCornerRadius: CGFloat = state == .initial ? ImageEditorView.defaultCornerRadius : 0
        if animated {
            let animation = CABasicAnimation(keyPath: #keyPath(CALayer.cornerRadius))
            animation.fromValue = imageView.layer.cornerRadius
            animation.toValue = imageCornerRadius
            animation.duration = animationDuration
            imageView.layer.add(animation, forKey: "cornerRadius")
        }
        imageView.layer.cornerRadius = imageCornerRadius

        if animated {
            UIView.animate(withDuration: animationDuration,
                           animations: {
                setControlsHiddenBlock()
                self.constrainContent(to: layoutGuide)
                self.updateImageViewTransform()
                // Animate layout changes made within bottomBar.setControls(hidden:).
                self.view.setNeedsDisplay()
                self.view.layoutIfNeeded()
            },
                           completion: completion)
        } else {
            setControlsHiddenBlock()
            constrainContent(to: layoutGuide)
            updateImageViewTransform()
            completion?(true)
        }
    }
```

### DIFF
--- a/SignalUI/AttachmentApproval/AttachmentPrepViewController.swift
+++ b/SignalUI/AttachmentApproval/AttachmentPrepViewController.swift
@@ -219,12 +219,15 @@ public class AttachmentPrepViewController: OWSViewController {
     private func presentFullScreen(viewController: UIViewController) {
         if let presentedViewController = presentedViewController {
             owsAssertDebug(false, "Already has presented view controller. [\(presentedViewController)]")
-            presentedViewController.dismiss(animated: false)
+            presentedViewController.dismiss(animated: false) { [weak self] in
+                self?.presentFullScreen(viewController: viewController)
+            }
+            return
         }
 
         viewController.modalPresentationStyle = .fullScreen
         zoomOut(animated: true) { [weak self] in
-            self?.presentFullScreen(viewController, animated: false)
+            self?.present(viewController, animated: false, completion: nil)
         }
     }
 


SignalUI/AttachmentApproval/AttachmentPrepViewController.swift
```diff
@@ -219,12 +219,15 @@ public class AttachmentPrepViewController: OWSViewController {
...
+            presentedViewController.dismiss(animated: false) { [weak self] in
+                self?.presentFullScreen(viewController: viewController)
+            }
+            return
         }
 
         viewController.modalPresentationStyle = .fullScreen
         zoomOut(animated: true) { [weak self] in
-            self?.presentFullScreen(viewController, animated: false)
+            self?.present(viewController, animated: false, completion: nil)
```
sashaweiss-signal, line 230: Especially since this isn't related to the fix, I'm more inclined to drop the `viewController.modalPresentationStyle = .fullScreen` here, and continue to call `self.presentFullScreen` here.

sashaweiss-signal:
This was merged internally as `d7e9a88`, which should become public tomorrow with our upcoming release! Thanks for the contribution.

PR #5981, opened 2025-02-24
#5890 Added in text message with single file sending functionality.

### LINKED ISSUE #5890: Accept both the File and Text (message) from iOS share sheet

### Bug description
When the sharing activity item array contains both the message and the attached file (as URL) the file is missed during the parsing and only the message will be shown. 

### Steps to reproduce
- open the [OrganicMaps app](https://github.com/organicmaps/organicmaps)
- try to export the Bookmarks category (as KMZ or GPX)
- share to the Signal
- only the message will be parsed and the attached to the share sheet file will not appear in the app 

**Actual result:** The attached file is missed

**Expected result:** Both the message and the shared file should be parsed and added to the chat

### Screenshots
image
image
image

How the Telegram App handles this sharing:
image

MarlowBrown:
### Contributor checklist

 * iPhone 16, iOS 18.2

- - - - - - - - - -

### Description

This PR `fixes #5890` by allowing a text message to have a file accompanying it. I tested this by performing the steps in #5890 to ensure the functionality works.

https://github.com/user-attachments/assets/92bbf44b-8990-4786-a5f4-c83144da1522

### DIFF
--- a/SignalShareExtension/ShareViewController.swift
+++ b/SignalShareExtension/ShareViewController.swift
@@ -675,8 +675,9 @@ public class ShareViewController: UIViewController, ShareViewDelegate, SAEFailed
         }
 
         // only 1 attachment is supported unless it's visual media so select just the first or just the visual media elements with a preference for visual media
-        let visualMediaCandidates = candidates.filter { $0.isVisualMedia }
-        return visualMediaCandidates.isEmpty ? Array(candidates.prefix(1)) : visualMediaCandidates
+//        let visualMediaCandidates = candidates.filter { $0.isVisualMedia }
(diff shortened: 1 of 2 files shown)


sashaweiss-signal:
Unfortunately, in the flow you're proposing it's not obvious that there's an attached file at all – it just looks like there's text to be shown. Any improvement here would need to indicate that both the file and text are attached.

MarlowBrown:
@sashaweiss-signal 

To make it obvious that there is a file, we should include a little banner with the file's name and a paperclip on it, right around where I circled right here.

Screenshot 2025-05-30 at 12 51 57 PM

Of course, this would require talking with the UI/UX engineer, so it's understandable if you want to have this internally done.

Issue #5980, opened 2025-02-22
"Manage" button to permit media access is very difficult to find when already having a lot of granted items


dabo248:
I didn't knew there was a possibility to allow access to other photos except via the iOS settings until I got a tip in this issue: 

> To add more photos, they can simply scroll to the side and select "manage." However, this experience is not currently available when creating stories.
> 
> [...]
>
> Oh wow, didn't knew that either. This is super hidden, especially when already having 300+ granted items in there. I think it would be much better to have the "manage" button as the first item in the selection and not the last one, so it's always visible.

 _Originally posted by @dabo248 in [#5921](https://github.com/signalapp/Signal-iOS/issues/5921#issuecomment-2613265793)_

I'm talking about the "Verwalten" button here:

![Image](https://github.com/user-attachments/assets/502fde18-3d47-4a96-82bb-2e2ed4c8c7a4)

As quoted I think it would be much easier to find the "Manage" button while having a lot of existing permitted media when it would reside on the first position of the list, not on the last one. 

**Suggested fix: Move "Manage" button to permit media access to first instead of last position in list**

With this change the button would be directly visible when attaching new images or videos after pressing the plus icon.

sashaweiss-signal:
Thanks for highlighting. Improvements to the "Manage" button have been merged in 7fe9aef364, which will be rolling out in the next release.

MarlowBrown:
Hey @sashaweiss-signal. Thanks for keeping us updated.

On another note, check your emails from me. I triaged all of Signal's open issues. 🗣️🔥🔥

PR #5984, opened 2025-02-25
#5980 Manage Photos Button is First

MarlowBrown:
### Contributor checklist

 * iPhone 16, iOS 18.2

- - - - - - - - - -

### Description

This PR proposes to enhance the RecentPhotoCollectionView, and move the manage button to be the first entry. Many users may have 10s or 100s of photos selected for limited access and having the button all the way at the end is impractical to reach. Moving it to be the first button would be the solution.

I tested this fix by selecting multiple photos in addition to toggling the selection for "photo limitations"

https://github.com/user-attachments/assets/27342e03-688c-41b3-8503-7150cca55197


sashaweiss-signal:
Hi Talon,

This is something we're interested in improving long-term by moving away from a custom photo picker more broadly. While it's true the "manage photos" button is fairly inaccessible if you've got lots of photos in the quick-picker, and the design could be improved, I don't think this is the solution we're going to go with.

Thanks for taking a look at this.

MarlowBrown:
Sounds good! It was a small change, so I'm not too bummed about it. If plans change, this PR and code changes will always be here.

PR #5988, opened 2025-02-27
#5876 Switched file decryption priority to be original file extension

### LINKED ISSUE #5876: .fpl Files Sent as Text Instead of Downloadable Attachments in Signal (iPadOS)

### Bug description
When sending `.fpl` files (ForeFlight Plan files, a file with a XML format) from one iPhone to another iPad via Signal, the files now appear in the chat with their content displayed, rather than as a downloadable file. Previously, these files could be sent and received in Signal as downloadable attachments. After a recent update, the content of the `.fpl` files is shown as text in the chat, which alters their functionality and requires a workaround (compressing and decompressing the files) to make them usable again in ForeFlight. Sending the same files through Airdrop for example, does not alter their functionality. 

### Steps to reproduce
- Create a `.fpl` file on iPhone 1 using ForeFlight.
- Send the `.fpl` file via Signal to iPad 2.
- On iPad 2, open Signal to view the sent file.

**Actual result:** The `.fpl` file's content is displayed in the chat rather than appearing as a downloadable file.

**Expected result:** The `.fpl` file should appear as a downloadable file in the chat, as it did in previous Signal versions, so it can be directly downloaded and opened in ForeFlight without requiring additional steps. In the images below you can see the old behaviour when clicking on the file from the chat. Only the filename is displayed with the extension and when clicking to the top to do sth with that file, the dropdown shows the ForeFlight app to open it with it. This has changed as you can see in the 2nd image, where the content of that file is displayed and the app ForeFlight does not longer appear in the dropdown. 

### Screenshots
old behavior
new_behavior

### Device info
**Device**: iPhone SE 2022

**iOS version**: 17.6.1

**Signal version (new behaviour):** 7.28 (328)
**Signal version (with old behaviour):** 7.22 (246)

MarlowBrown:
### Contributor checklist

 * iPhone 16, iOS 18.2

- - - - - - - - - -

### Description

This PR `fixes #5876` by changing the file decryption priority to be the files original extension instead of the detected files mime type.

However I do see some value in decrypting the file as its mime type to see the raw file; I think losing the ability to send the file to the respective application its used for comes second to that. Perhaps in the future the UI/UX will be redesigned to accommodate path viewing the raw file and being able to use the original files extension.

https://github.com/user-attachments/assets/9c3c14a2-14e1-4759-a00e-aacb2e53d9be

### DIFF
--- a/SignalServiceKit/Messages/Attachments/V2/AttachmentStream.swift
+++ b/SignalServiceKit/Messages/Attachments/V2/AttachmentStream.swift
@@ -87,13 +87,13 @@ public class AttachmentStream {
     /// will instead be inferred from the file contents) and made url-safe AND user-friendly. If nil, a random file name is used.
     public func makeDecryptedCopy(filename: String?) throws -> URL {
         var pathExtension: String = {
-            if let pathExtension = MimeTypeUtil.fileExtensionForMimeType(mimeType) {
-                return pathExtension
-            } else if
+            if
                 let filename,
                 let filenameUrl = URL(string: filename),
                 let pathExtension = filenameUrl.pathExtension.nilIfEmpty
             {
+                return pathExtension
+            } else if let pathExtension = MimeTypeUtil.fileExtensionForMimeType(mimeType) {
                 return pathExtension
             } else {
                 return "bin"


tomduese:
Thank you very much for looking into this @MarlowBrown 
What is the timeframe of having this merged and on production?

MarlowBrown:
Hey @tomduese. Since I'm just a contributor to this repository it will be when one of the signal devs takes a look at it to merge it into the codebase. Depending on how busy their day is I can't give you a concrete timeline. All that I hope for is within a month.

sashaweiss-signal:
Hey @MarlowBrown. This is a deceptively simple change, and before we take it I'll need to do a thorough evaluation of what the upstream effects would be. This method is very explicit in its preference order, so while it may be safe to change this I'd want to be sure it won't inadvertently break other things.

MarlowBrown:
Keep open

MarlowBrown:
Keep open

Issue #5955, opened 2025-01-31
Snapped selfie picture is horizontally flipped


stianalexan:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
When taking a picture with the selfie camera it is horizontally the way I would expect (like a mirror). When I snap the picture it gets flipped horizontally.

### Steps to reproduce
- Open a chat
- Open selfie camera to send image
- Snap picture

**Actual result:** The image gets flipped horizontally.

**Expected result:** The snapped picture is the same as in the camera preview.

### Device info

**Device**: iPhone X/iPhone 16

**iOS version**: 16.7.10/18.1.1

**Signal version:** 7.44 (587)

guidobouman:
But that is how the rest of the world sees you, no? I'd say it's expected? Otherwise your facial features become inconsistent, based on which camera you use.

jonasrydningen:
It is true that that is how the rest of the world sees you. This is the way iOS does it, flipped preview (like a mirror) and unflipped result. However this is not how selfie heavy apps like Snapchat or Instagram deals with it, they keep the mirrored orientation. I think because seeing your face flip in real time is jarring, especially for people with less than perfect symetrical faces (most people).

Only reporting this as my friend who I introduced Signal to manually flips each and every selfie she sends. The reciever doesn't know and probably doesn't care.

PR #5991, opened 2025-02-28
Fix Video and Photo selfie output to mirrored

flashyhuckle:
Fixing a bug where photo and video from selfie camera was mirrored to 'true to life' instead of staying as a mirror representation

### Contributor checklist

 * I have tested isolated settings I'm changing on iPhone 12 mini on iOS 18.3 and 18.3.1, however I wasn't able to test it as a whole app due to my apple developer account lacking features required to run the app on real device, as simulator doesn't support camera functions, so if someone else could verify it on a real device, it would be much appreciated.

- - - - - - - - - -

### Description
This is related to issue #5955. I was debating myself whether this is an issue at all, as outputing 'true to life' photo could be a preference for some, but then I remembered the ideology "There are no power users.", so I asked few people what would be their preferred outcome, and 6 answered with "I want the photo to be mirrored", while only 1 said they would want a 'real' photo. Given that user can flip the photo afterwards, the app should present user with selfie photo captured as the preview shows it - mirrored.
I have also changed code for video output, as keeping photo output mirrored while flipping video would be illogical.

### DIFF
--- a/Signal/src/ViewControllers/Photos/CameraCaptureSession.swift
+++ b/Signal/src/ViewControllers/Photos/CameraCaptureSession.swift
@@ -343,7 +343,7 @@ class CameraCaptureSession: NSObject {
 
     func updateVideoCaptureOrientation() {
         sessionQueue.async {
-            self.videoCapture.setVideoOrientation(self.captureOrientation)
+            self.videoCapture.setVideoOrientation(self.captureOrientation, outputMirrored: self.desiredPosition == .front)
         }
     }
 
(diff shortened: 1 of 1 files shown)


MarlowBrown:
Hey @flashyhuckle!

I have a paid developer account and a physical device to test on now so I can help you out soon. I'm preoccupied with other matters, but expect a review near the middle to end of this week. :)

sashaweiss-signal:
Hi @flashyhuckle, thanks for looking into this. The current behavior is intentional, although we understand it's not everyone's preference. It's a decision that we do revisit periodically, but for the time being we don't plan to change it.

Thanks again for looking into this, and for being a Signal user!

sashaweiss-signal:
As this is the current preferred behavior, I'm going to close this PR. Thanks again for looking into it.

PR #6000, opened 2025-03-07
Fix UI getting stuck when transition animation is interrupted

ChronicLynx:

- [X] I have signed the [Contributor License Agreement](https://signal.org/cla/)

### Contributor checklist

 * iPhone 16 Pro (iOS 18.2, Xcode Simulator)

- - - - - - - - - -

### Description
This is a continuation of the issue in #5976. I encountered two scenarios over the week where after tapping done or cancel the buttons animated away, but the view failed to dismiss. I believe #5977 fixed part of this issue, as I haven't had the same error logged, but it's still possible via this route which doesn't generate an error. I think the frequency at which this occurs on my device is much less than pre-#5977 at least.

The theory here is that the animation is interrupted partway through the cycle and returns false even though the buttons have already been dismissed. I think these problems in general are indicative of a larger issue, but we may be able to provide relief with these workarounds.

### DIFF
--- a/SignalUI/ImageEditor/ImageEditorCropViewController.swift
+++ b/SignalUI/ImageEditor/ImageEditorCropViewController.swift
@@ -948,8 +948,7 @@ extension ImageEditorCropViewController {
 
     @objc
     private func didTapCancel() {
-        transitionUI(toState: .initial, animated: true) { finished in
-            guard finished else { return }
+        transitionUI(toState: .initial, animated: true) { _ in
             self.dismiss(animated: false)
         }
     }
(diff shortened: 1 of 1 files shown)


ChronicLynx:
@sashaweiss-signal any thoughts on this? Low priority I'm sure, but it would be good to have a handler of some sort since it just hangs currently.

sashaweiss-signal:
Sorry for the delay. We'll keep the PR open so this doesn't get lost, but unfortunately as you noted due to relative priority I can't guarantee when we'll have time to engage.

PR #6007, opened 2025-03-21
Places the manage-your-selected-photos at start of collection

robvandijk:

### Contributor checklist

 * iPhone 16 Pro, iOS 18.2

- - - - - - - - - -

### Description
When access to photos is limited, the UI for selecting additional photos places the "Manage" button at the end of the collection of already selected photos. When hundreds of photos have already been added, it takes a very long time to scroll to the far right, and after selecting a photo you have to scroll left again (see first 2 screenshots). This has been a nuisance for myself for a couple of years now, but it is also an unpleasant surprise for people who nowadays switch to Signal.

This PR simply places the "Manage" button at the start of the collection.

Now, I know Signal is planning a better solution to this problem using more regular iOS photo library components. But I believe you are underestimating the problem that end-users face. For me it is an argument not to promote Signal to friends and collegues too strongly. I therefore urge you to consider implementing a simple solution like this for now, and in the meantime work on a better one. 

old-situation-after-open

old-situation-after-scrolling

new-situation-after-open

### DIFF
--- a/Signal/src/ViewControllers/Attachment Keyboard/RecentPhotoCollectionView.swift
+++ b/Signal/src/ViewControllers/Attachment Keyboard/RecentPhotoCollectionView.swift
@@ -51,6 +51,21 @@ class RecentPhotosCollectionView: UICollectionView {
     private lazy var collection = photoLibrary.defaultPhotoAlbum()
     private lazy var collectionContents = collection.contents(ascending: false, limit: RecentPhotosCollectionView.maxRecentPhotos)
 
+    private func collectionIndexForIndexPath(_ indexPath: IndexPath) -> Int {
+        if isAccessToPhotosLimited {
+            return indexPath.row - 1
+        } else {
+            return indexPath.row
+        }
+    }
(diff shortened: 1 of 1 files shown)


sashaweiss-signal:
Hi, as you noted and as discussed on some other PRs and issues, this isn't the way we want to improve this UX. I understand that it's frustrating in the meantime, in that the "select more photos" UX is potentially inaccessible from within Signal, and so instead needs to be accessed directly via iOS Settings.

Consequently, I'm going to close this PR. Thanks for looking into this, and I'm sorry for the inconvenience while we make separate improvements to the photo picker. Thanks also, as always, for being a Signal user.

robvandijk:
Can you please tell me the time frame that this new UX will be available?

Because even if it is just 2 months away, I do not understand closing this PR. The PR is a no-brainer, technically, and a life-saver for end-users. You could really make a huge difference for end-users without any effort.

The fact this is not being picked up by the team leads to only one possible conclusion: there is not a single member of the Signal team using the iOS app with limited photo access. Because if there had been just 1 member using it this way, the pain would have been felt and this would have been solved more than a year ago.

Issue #6008, opened 2025-03-24
Get started banner and first conversation message placement bug


armartinez:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
After opening a conversation after a first install of the app, the get started banner moves up and can't be interacted with on the top half on each element. Also, the first conversation message moves down from the starting position.

### Steps to reproduce
- Do a fresh install
- Get to the chat list
- Open a Note to Self chat
- Go back to chat list

**Actual result:**

The get started banner is moved up and the first conversation message is moved down.

**Expected result:** 

Everything should look the same.

### Video

https://github.com/user-attachments/assets/f59894ef-4363-4447-bda7-7f3d1209403e

### Device info
**Device**: iDevice X

**iOS version**: 18.3.2

**Signal version:** 7.51

PR #6009, opened 2025-03-24
#6008 Fixed placement errors for get started banner and first conversation

armartinez:
### Contributor checklist

 * iPhone 16 Pro

- - - - - - - - - -

### Description
This fixes issue #6008 for the placement of the get started banner and first conversation cue. The get started banner was inheriting the margins from the chat list view, which after opening a conversation gets modified by the navigation, for instance when showing the search bar. This change just breaks that link and passes a fixed value for the leading a trailing margins which are changed by matching the width of the superview, necessary to create the sliding from the right effect of the banner. The conversation cue position was being modified in the same way, so this change places the cue as a subview of the navigation, preventing it from being moved or getting below the navigation. 

https://github.com/user-attachments/assets/09c3bc76-ace1-4c32-8ea7-ec0b7dc417a3

### DIFF
--- a/Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.swift
+++ b/Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.swift
@@ -76,7 +76,7 @@ public class ChatListViewController: OWSViewController, HomeTabViewController {
             navigationController.view.addSubview(firstConversationCueView)
             
             // This inset bakes in assumptions about UINavigationBar layout, but I'm not sure
-            // there's a better way to do it, since it isn't safe to use iOS auto layout with
+            // if there's a better way to do it, since it isn't safe to use iOS auto layout with
             // UINavigationBar contents.
             firstConversationCueView.autoPinEdge(toSuperviewMargin: .top, withInset: 35)
             firstConversationCueView.autoPinEdge(toSuperviewEdge: .trailing, withInset: 6)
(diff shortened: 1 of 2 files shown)


ImAwsumm (approved):

Signal/src/ViewControllers/HomeView/Chat List/ChatListViewController.swift
```diff
@@ -72,14 +72,16 @@ public class ChatListViewController: OWSViewController, HomeTabViewController {
...
-        // UINavigationBar contents.
-        firstConversationCueView.autoPinEdge(toSuperviewEdge: .trailing, withInset: 6)
-        firstConversationCueView.autoPinEdge(toSuperviewEdge: .leading, withInset: 10, relation: .greaterThanOrEqual)
-        firstConversationCueView.autoPinEdge(toSuperviewMargin: .bottom, relation: .greaterThanOrEqual)
+        if let navigationController = self.navigationController {
+            navigationController.view.addSubview(firstConversationCueView)
+            
+            // This inset bakes in assumptions about UINavigationBar layout, but I'm not sure
+            // there's a better way to do it, since it isn't safe to use iOS auto layout with
+            // UINavigationBar contents.
```
ankushkushwaha, line 80:
```suggestion
            // This inset bakes in assumptions about UINavigationBar layout, but I'm not sure
            // if there's a better way to do it, since it isn't safe to use iOS auto layout with
            // UINavigationBar contents.
```

ankushkushwaha (approved):

sashaweiss-signal:
Hi, thanks for your interest in contributing. There are indeed known layout issues with the "Get Started" cards, but unfortunately I don't think baking in hardcoded margins is the way we'd want to go. I'm going to close this for now, but I've linked it to our internal tracking so we can revisit this approach in the future.

Issue #6023, opened 2025-04-10
Jump To Message When Viewing "All Media"


stevenwalton:
- [x] I have searched open and closed issues for duplicates
- [ ] I am submitting a bug report for existing functionality that does not work as intended
- [ ] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
It is not possible to jump to the message associated with a media message. The options are "save", "delete" (under the 3 dots), export (with typical Apple options) and "Forward". The result of this is that users are unable to[^0] reply to image messages from the past[^1].

[^0]: I mean "effectively unable to". If there is no text associated with the image users have no ability to search to get near the image, leaving only the option to brute force scroll or excessive guessing of search terms. 

[^1]: A useful conversation tool as we can *also* jump to the conversation and regain context. Just as one might want to reply to any other message.

The obviously solution to this is to "Forward" the message, which ends up duplicating these images, resulting in increased storage on both user devices as well as increased data transmission which costs both users and Signal money[^3].

[^3]: This is why I consider this a bug. Data duplication and excessive data transmissions are not desired results. 

 This issue was resolved on Android in commit [21bd8a3
](https://github.com/signalapp/Signal-Android/commit/21bd8a308ba2c02d066bbe445a65dc9f8a74aff4) (associated [community post](https://community.signalusers.org/t/add-jump-to-message-button-when-viewing-media/5422/12)[^4]). This issue was not resolved in either iOS or Desktop. It should be noted that the solution there is quite unintuitive. It is nice that you can click on the date (keep?) but it should also be included in the 3 dot dropdown or somewhere else so that a user can discover it themselves. 

[^4]: A community forum is not an effective means to handle these issues. Especially considering the culture that has been cultivated over there. It can be helpful to connect users with other knowledgeable users but it creates a lot of noise and requiring registration leads to significantly inaccurate measurements of what features users actually want.

### Steps to reproduce
- Open any Signal conversation containing media messages
- Click on User's name at the top of the app
- Either:
    - Select "See All" under "All messages"
        - Click on a picture
    - Select one of 4 recent images
- Click on date, 3 dot icon, or any other button 

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)

No operation

**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)

To be able to forward messages

### Screenshots
N/A

### Device info
Currently affects all iOS devices and all versions of Signal at least prior to 7.53.1

[...]

PR #6028, opened 2025-04-13
#6023: Add "Go to Message" Feature to Media Viewer

kp198:

### Contributor checklist

- [ ] My contribution is fully baked and is ready to be merged as is
 * iDevice A, iOS X.Y.Z
 * iDevice B, iOS Z.Y

- - - - - - - - - -

### Description

This fixes https://github.com/signalapp/Signal-iOS/issues/6023 by adding the view media in conversation functionality.

- Added new "Go to Message" option in media viewer context menu
- Implemented navigation from media viewer to the source message in conversation
- Added automatic scrolling to the message position when returning to conversation

- [Needs Translations]

https://github.com/user-attachments/assets/75fdcd94-c45b-4947-86e0-23242fac2302

### DIFF
--- a/Signal/src/ViewControllers/MediaGallery/MediaPageViewController.swift
+++ b/Signal/src/ViewControllers/MediaGallery/MediaPageViewController.swift
@@ -345,6 +345,16 @@ class MediaPageViewController: UIPageViewController {
         ))
         contextMenuActions.append(UIAction(
             title: OWSLocalizedString(
+                "MEDIA_VIEWER_GO_TO_MESSAGE",
+                comment: "Context menu item in media viewer. Refers to going to the position in conversation of that photo/video."
+            ),
+            image: Theme.iconImage(.arrowRight),
+            handler: { [weak self] _ in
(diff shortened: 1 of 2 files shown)


sashaweiss-signal:
Hi, thanks for looking into this! I'm planning to merge a version of this change. The way we do that is I'll create an internal copy of this PR, preserving your original author info, and merge it as normal. It'll then become public with an upcoming release. Once I merge it internally, I'll circle back here and close this PR with the commit hash of the internal version.

I'm going to make some small tweaks in the internal version, but it'll use the same author info as your original commit here.

sashaweiss-signal:
I merged this internally as ac31946 mostly as-is, and added a little cleanup in fda33e0. (I was surprised we needed to specifically make a call to scroll the conversation view, because the `focusMessageId` parameter should've covered that, which led me to a bug in the `presentConversation` method that ignored `focusMessageId` if the conversation was already presented, as it always is in this case.)

Thanks again for the contribution! I'm going to close this PR now.

PR #6055, opened 2025-05-20
#5769 - Pause all audio attachments if receiving an incoming call.

### LINKED ISSUE #5769: Audio output randomly disabled on calls until app restart

### Bug description
Describe here the issue that you are experiencing.

I'm on Android. This only seems to happen with my peers on iOS. It's being going on for months if not years. For some reason, when they answer my call, they can't hear anything. I can hear them, though. They can call me back, or I can call them back, and the silence persists. It's only when they close Signal (swipe it away) and reopen it by tapping on the icon that they can again hear me. This has happened regardless of which of my Android phones I've been on, and it seems unpredictable, affecting maybe 5% of calls.

### Steps to reproduce

- Open Signal on Android (or maybe iOS).
- Voice call your friend on iOS.
- Let them answer.
- Randomly, they won't be able to hear you.
- There is probably something else necessary to reproduce this. It might be, for instance, that the receiving phone needs to have been idle long enough to fall asleep first. I've just never been able to nail it.

I'm pretty sure this has happened when the recipients were not actually listening to a voice message. Nevertheless there might be a common cause here:

https://github.com/signalapp/Signal-iOS/issues/5493

Also, I haven't asked my iOS peers to try muting and then unmuting, as described in this vaguely related issue:

https://github.com/signalapp/Signal-iOS/issues/5689

If this occurs again, I'll try asking them to do that.

Randomly, my peer can't hear me.

They would be able to hear me.

**Device**: iPhone 13, at least, and I believe others, but it's been too long to remember.

**iOS version**: Various over many months.

**Signal version:**: Latest over many months.

MarlowBrown:
### Contributor checklist

 * iPhone 12, iOS 18.4.1
 * iPhone 16, iOS 18.4.1

- - - - - - - - - -

### Description

This PR `fixes #5769` by making sure that audio attachments don't interfere with signal calls. This problem stems from audio attachments that are playing while an incoming call happens, have their audio controls captured by NowPlayingInforCenter and RemoteCommandCenter, switch subsequently blocks audio coming from a signal call.

I tested this by playing audio attachments and then initiating an incoming call. Additionally, I tested to make sure that audio attachments still interrupt music playing in the background, and subsequently resume upon the audio attachment pausing or completing.

### DIFF
--- a/Signal/Calls/UserInterface/CallUIAdapter.swift
+++ b/Signal/Calls/UserInterface/CallUIAdapter.swift
@@ -97,6 +97,9 @@ public class CallUIAdapter: NSObject {
         }
         Logger.info("remoteAddress: \(caller)")
 
+        // make sure we dont have any audio playing as that will bug audio when it interacts with the call service
+        AppEnvironment.shared.cvAudioPlayerRef.pauseAll()
+        
         // make sure we don't terminate audio session during call
         _ = SUIEnvironment.shared.audioSessionRef.startAudioActivity(call.commonState.audioActivity)
 
--- a/Signal/ConversationView/CVAudioPlayback.swift
+++ b/Signal/ConversationView/CVAudioPlayback.swift
@@ -248,6 +248,14 @@ public class CVAudioPlayer: NSObject {
         audioPlayback.stop()
         self.audioPlayback = nil
     }
+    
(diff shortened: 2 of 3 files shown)


MarlowBrown:
Keep open

PR #6058, opened 2025-05-26
#5886-Prevent Crash on Airplaying Audio Attachment

### LINKED ISSUE #5886: Signal crashes if you try to play a voice message while airplay is active

### Bug description
If you are AirPlay-ing something running in the background (e.g. youtube) and try to play a voice message in Signal, the app will crash as soon as you click the play button.

### Steps to reproduce
- Have a Mac or Apple TV you can AirPlay to
- AirPlay something to your device (eg a youtube video)
- Find/send a voice message in Signal
- Play it

**Actual result:** The Signal app attempts to play the voice message through AirPlay. As soon as it tries to do so, it crashes.

**Expected result:** The Signal app doesn’t try to play the voice message through AirPlay

### Screenshots
N/A

### Device info
**Device**: iPhone 11

**iOS version**: 17.7

**Signal version:** 7.32

### Link to debug log

https://debuglogs.org/ios/7.32.0/2d176bab1f008086e7e7fbdf1f08df0b0aed0746be72f77444f9809163d08a38.zip

MarlowBrown:
### Contributor checklist

 * iPhone 12, iOS 18.4.1

- - - - - - - - - -

### Description

This PR `fixes #5886` by preventing AudioPlayers AVPlayer from externally playing. Weirdly (this will be explained later) this doesn't prevent audio from from being airplayed, but it prevents the dataRequests requested data length from being bigger than a UInt32, which was causing this crash to happen. I tested this by airplaying Spotify from my iPhone 12 to my Mac Mini and playing an audio attachment.

My hypothesis:

While working on this bug, I noticed that in the AudioPlayer class, we are using the _AVPlayer_ class instead of the _AVAudioPlayer_ class. I think the root cause of what is causing this bug to happen is that when we switch to using AirPlay, the system sees that the ResourceLoadingRequest is coming from an AVPlayer, something that is designed to play both video and audio, and then subsequently requests too much space for video content, something that will never come as were using only audio in our AVAssets. By disallowing external playback, we prevent any video content from being played outside the app (even though we don't have any) and subsequently the space for video content never gets allocated. I think by switching the AudioPlayers AVPlayer to _AVAudioPlayer_ will solve this problem.

### DIFF
--- a/SignalUI/AV/AudioPlayer.swift
+++ b/SignalUI/AV/AudioPlayer.swift
@@ -224,7 +224,9 @@ public class AudioPlayer: NSObject {
                     asset = AVURLAsset(url: symlinkUrl)
                 }
             }
-            return AVPlayer(playerItem: .init(asset: asset))
+            let avPlayer = AVPlayer(playerItem: .init(asset: asset))
+            avPlayer.allowsExternalPlayback = false
+            return avPlayer
         }
 
         let audioPlayer: AVPlayer
@@ -235,6 +237,7 @@ public class AudioPlayer: NSObject {
             case .attachment(let attachment):
                 let asset = try attachment.decryptedAVAsset()
                 audioPlayer = .init(playerItem: .init(asset: asset))
+                audioPlayer.allowsExternalPlayback = false
             }
         } catch let error as NSError {
             Logger.error("Error: \(error)")


sashaweiss-signal:
Hi, thanks again for looking into this. We investigated on our end and found a simpler root cause and fix, which will become public soon in 4378304.

---

> Even if attachment size is capped to 32 bits, AVAssetResourceLoadingDataRequest.requestedLength can be set to NSIntegerMax and the cast can cause a crash. According to comments in the header, its set to NSIntegerMax in the following scenario:
>
> `When requestsAllDataToEndOfResource is YES and the content length has not yet been provided by you via a prior finished loading request, the value of requestedLength is set to NSIntegerMax`
> 
> This seems to be the case when a user tries playing a voice memo when airplay is happening.

```.../Messages/Attachments/V2/Playback/AVAsset+Attachment.swift   | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/SignalServiceKit/Messages/Attachments/V2/Playback/AVAsset+Attachment.swift b/SignalServiceKit/Messages/Attachments/V2/Playback/AVAsset+Attachment.swift
index 2f8e16e679d..7cc2c00d212 100644
--- a/SignalServiceKit/Messages/Attachments/V2/Playback/AVAsset+Attachment.swift
+++ b/SignalServiceKit/Messages/Attachments/V2/Playback/AVAsset+Attachment.swift
@@ -150,7 +150,7 @@ extension AVAsset {
             }
 
             let requestedOffset = UInt32(dataRequest.requestedOffset)
-            var requestedLength = UInt32(dataRequest.requestedLength)
+            var requestedLength = UInt32(clamping: dataRequest.requestedLength)
             if dataRequest.requestsAllDataToEndOfResource {
                 requestedLength = fileHandle.plaintextLength - requestedOffset
             }
```

MarlowBrown:
Thank you for giving me an update on this! It didn't mention this in the Apple documentation at all. Very nice find! 👍

PR #6081, opened 2025-07-11
#5399 Do not use .defaultToSpeaker for shared AudioSession

### LINKED ISSUE #5399: Voice Message Recording does not work with Headphones (AirPods)
> * [x]  I have searched open and closed issues for duplicates
> * [x]  I am submitting a bug report for existing functionality that does not work as intended
> * [x]  This isn't a feature request or a discussion topic
> 
> ### Bug description
> Voice message recording does not use the headphone microphone for recording
> 
> ### Steps to reproduce
> * Connect Headphones, such as AirPods and open Signal
> * Record a Voice Message from a Signal Chat with your AirPods, walk away from your iphone
> * Send the Voice Message
> * Play the voice message, you will recognize that the microphone from the AirPods are not used, so your voice couldn't be recorded
> 
> **Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)
> 
> * A Voice Message gets recorded but the microphone from the iPhone is used, not the headphone/airpods microphone
> * This does work without a problem in e.g. WhatsApp
> * I also tested this with Signal for MacOS. In this case the AirPods are used as a Microphone, working away from the Mac does not result in not recorded voice
> * I dont know if this is a AirPods issue or Headphone issue in general
> 
> **Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)
> 
> * The microphone from the connected headphones are used, not the iPhone Microphone
> 
> **iOS version**: iOS 15.5
> 
> **Signal version:** 5.42.0.12

ospfranco:
It seems using .defaultToSpeaker causes issues when recording with AirPods. By disabling this line it audio recording is automatically routed to the appropriate device

- [X] I have signed the [Contributor License Agreement](https://signal.org/cla/)

### Contributor checklist

 * iPhone 16 Pro, iOS 18.5

- - - - - - - - - -

### Description
Fixes #5399 where audio is routed through the device microphone despite it being connected to other Bluetooth device. Such as AirPods. If the session is re-used maybe this could mess playback? Not super familiar with the codebase, but at least the behavior would be consistent with WhatsApp where even when connected to Bluetooth speaker, messages are played there.

The alternative would be to use two AudioSessions maybe? One for recording and one for playback

### DIFF
--- a/SignalUI/AV/AudioSession.swift
+++ b/SignalUI/AV/AudioSession.swift
@@ -172,7 +172,6 @@ public class AudioSession: NSObject {
                 .playAndRecord,
                 mode: .default,
                 options: [
-                    .defaultToSpeaker,
                     .allowBluetoothHFP,
                     .allowBluetoothA2DP
                 ]


ospfranco:
I just tested again and it seems to be working **without** my changes. Maybe someone can confirm this?

ospfranco:
Yeah, this seems to have been fixed. Going to close this.

PR #6088, opened 2025-07-26
#6079 - Double tapping on a quotedmessageview using voiceover

### LINKED ISSUE #6079: VoiceOver users cannot jump to the quoted message

### Bug description
VoiceOver users cannot jump to an original message if the current message is a reply to said original. Sighted users can just tap on the quote and jump directly to the original message.

### Steps to reproduce
- Open a chat that has multiple messages, and at least one is a reply to another message.
- Turn on VoiceOver
- Focus the reply
- Try to jump to the original. Double-tapping doesn't work, and there is no action in the action rotor to jump to the original message, either. It won't work.

**Actual result:** Nothing happens, current message stays in VoiceOver focus, and the screen doesn't scroll

**Expected result:** Choosing an appropriate action, VoiceOver users should be able to jump directly to the original message. Screen should scroll, and VoiceOver focus should be placed on the message item for the original message.

### Device info

**Device**: iPhone 16 Pro max

**iOS version**: 26 Beta 2, but also happens in 18 (any release)

**Signal version:** Any current or recent version, this has never worked

MarlowBrown:
### Contributor checklist

 * iPhone 12, iOS 18.4.1

- - - - - - - - - -

### Description

This PR `fixes #6079` by adding additional text to the accessibility string for a quoted message view. This extra text informs the user that by double-tapping on the view, they will be taken to the original message.

https://github.com/user-attachments/assets/46604490-4bcd-4c95-b05c-e1a6be2585c3

(Edit)
Since my change involves touching some of the locale files, what is the process for updating all locale files to be accurate? Do a series of human translators do it, or is there a script for using an AI translator? Can you give me some insight, Sasha?

### DIFF
--- a/Signal/ConversationView/Components/CVComponentMessage.swift
+++ b/Signal/ConversationView/Components/CVComponentMessage.swift
@@ -2232,6 +2232,10 @@ public class CVComponentMessage: CVComponentBase, CVRootComponent {
             }
         }
 
+        func focusAccessibility() {
+            UIAccessibility.post(notification: .layoutChanged, argument: self.hInnerStack)
+        }
+        
         // MARK: - Flashing Message Bubble
 
         func performMessageBubbleHighlightAnimation() {
--- a/Signal/ConversationView/ConversationViewController+CVC.swift
+++ b/Signal/ConversationView/ConversationViewController+CVC.swift
@@ -4,6 +4,7 @@
 //
 
 public import Foundation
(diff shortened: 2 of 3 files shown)


MarlowBrown:
Keep open

MarlowBrown:
Keep open

MarlowBrown:
Keep open

PR #6096, opened 2025-08-15
Add Russia to censoredCountryCodes

wukko:
- [ ] I have not signed the [Contributor Licence Agreement](https://signal.org/cla/), but I agree to all terms listed there.

### Contributor checklist
 * iPhone 8, iOS 16.7.11

- - - - - - - - - -

### Description
This PR enables censorship circumvention by default for accounts with a Russian (+7) phone number. At the time of writing, it's impossible to use Signal in Russia due to the local government's network restrictions. This small fix is required for Signal to work independently and without additional workarounds.

Additionally, I noticed that Signal doesn't call `updateHasCensoredPhoneNumberDuringProvisioning()` during the registration/login flow, so the user still has to use a VPN or proxy service to create an account or log in. This seems like an oversight, so I'll try to make an additional PR that fixes this in the near future (I have to come up with a clean way to do this first).

### DIFF
--- a/SignalServiceKit/Network/OWSCensorshipConfiguration.swift
+++ b/SignalServiceKit/Network/OWSCensorshipConfiguration.swift
@@ -170,6 +170,8 @@ struct OWSCensorshipConfiguration {
         "+998": "UZ",
         // Pakistan
         "+92": "PK",
+        // Russia
+        "+7": "RU",
     ]
 
     /// Returns nil if the phone number is not known to be censored


sashaweiss-signal:
Hi, thanks for your interest. We monitor some metrics internally to maintain this list, and at this time we don't believe that adding +7 to the default list is the right step. Consequently, I'm going to close this PR.

We do periodically make updates to this list based on our monitoring, though, so this may come in a later release.

Issue #6021, opened 2025-04-08
Voice recording stops when changing orientation CW only, and keyboard is hidden; plus other cases


tudy7:
- [ x ] I have searched open and closed issues for duplicates
- [ x ] I am submitting a bug report for existing functionality that does not work as intended
- [ x ] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description

This is discussed in [Issue 4359](https://github.com/signalapp/Signal-iOS/issues/4359) & [Issue 5404](https://github.com/signalapp/Signal-iOS/issues/5404) and it was marked as solved, but the situation is still reported by some users. 

I have found some conditions under which the bug still appears, and it has to do with the visibility of the keyboard, the direction in which you turn the phone, as well as the number of 90 deg rotations when starting recording from landscape orientation.

This issue is NOT present if you press the mic button and start recording an audio message from portrait orientation while the keyboard is visible: you can rotate the device in any direction, the message keeps recording. Nor does the bug appear when the keyboard is hidden, but you rotate  the device CCW.

However, with the keyboard hidden, if you press the mic button to record a message at the bottom of the screen, then rotate your phone clockwise to landscape, the audio recording cancels and what you've recorded gets lost. Very frustrating corner case.

### Steps to reproduce
- make sure your device's rotation lock is off, so the screen changes orientation,
- go to a Signal conversation,
- make sure the keyboard is not visible, or hide it by swiping down in the conversation,
- start recording an audio message by pressing the mic button at the bottom right of the screen,
- tilt your phone TO THE RIGHT (clockwise) so that the screen changes orientation.

**Actual result:** The voice recording stops and the recorded voice message disappears. Two haptic feedback buzzez.

**Expected result:** The voice recording continues, as it does when the keyboard is visible or when you rotate your device to the left. One haptic feedback buzz.

### Additional cases in which this issue occurs:
Actually, after additional fiddling around, it turns out recording an audio message also cancels in the following situations:
- When keyboard is visible and you start recording in left landscape, rotate right to portrait, then rotate right again to right landscape, after the second rotation, the recording stops;
-- the bug is present in the opposite sense as well, when you start recording in right landscape, and rotate phone to upright, then to left landscape,
-- weirdly, the bug does not show up when rotating the phone directly 180 deg from right landscape to left landscape, or vice-versa,
- When keyboard is hidden, in the situations described above with two consecutive 90 deg rotations.

[...]

Issue #4359, opened 2020-01-14
Rotating phone aborts voice message


jonas7654:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
Hello, 
I don't know if it is a bug or intentional but when recording a voice message it stops recording when rotating the screen.

### Steps to reproduce
-Start recording a Voice Message without locking the recording.
-Turn your phone in order to rotate the screen.

**Actual result:** The current recording will stop.

**Expected result:** It would be great if the Signal app would turn of the screen rotation feature while recording in order to prevent the cancelling mentioned above.

### Device info

**Device**: iPhone X
**iOS version**: 13.3

**Signal version:** 3.1.1.1

### Link to debug log
https://debuglogs.org/8d6a4175c8cc8269613888514aa85bf1268a8ca8188a7a7841e9a42f18eaaccb.zip

thegesuser:
+1. Expiriencing the same issue since the iPad App was released.

SawkeeReemo:
Same here, on iPhone 11 Pro Max running iOS 13.3. Hopefully a fix is in the works. Also, if I'm interrupted by a call or something, instead of sending an incomplete message (like Voxer does), it just cancels the voice message.

vjcharles:
This has been a longtime issue with Signal iOS. A workaround is to lock the orientation before recording, which is cumbersome to remember to do and unlock for other apps.
It may go without saying, but losing a recorded message can be frustrating and lead to less trust the app is able to properly handle content.

axldd:
Having the same issue on iPhone XR (iOS 13.3.1). Already lost many 5 minutes + recordings due to that bug.

collysucker:
I have the same issue on an iPhone 11, iPhone 7, iPhone SE (always latest iOS, currently iOS 14.3). I and others already lost many minutes of voice messages due to this bug, which is really annoying, after mostly it happens when you record long voice messages and you accidentily rotate your iPhone while doing that. This also already happend often to some friends and my family. Please fix this issue :) 👍

major-mayer:
A friend of mine told me just today that He experienced the Same issue in signal ios multiple times, which was very annoying.

I was suprised, bc on Android this isn't a problem and IT shouldn't be one on ios too

vjcharles:
[...]

PR #6109, opened 2025-09-09
**Commit message:**

mogwai:
**Changes:**
1. Added `isVoiceInteractionActive` computed property to `ConversationInputToolbar` that checks if voice memo is in active state (not idle/draft)
2. Added orientation lock logic to `ConversationViewController.shouldAutorotate` that prevents rotation during voice interactions
3. Both changes work together to fix autorotation issues during voice memo usage

### Contributor checklist

 * iPhone 15 Pro Max
- - - - - - - - - -

### Description
Fixing #6021 #4359

### DIFF
--- a/Signal/ConversationView/ConversationInputToolbar.swift
+++ b/Signal/ConversationView/ConversationInputToolbar.swift
@@ -1440,6 +1440,11 @@ public class ConversationInputToolbar: UIView, LinkPreviewViewDraftDelegate, Quo
             ensureButtonVisibility(withAnimation: true, doLayout: true)
         }
     }
+    
+    var isVoiceInteractionActive: Bool {
+        return voiceMemoRecordingState != .idle && voiceMemoRecordingState != .draft
+    }
+    
     private var voiceMemoGestureStartLocation: CGPoint?
 
     private var isShowingVoiceMemoUI: Bool = false {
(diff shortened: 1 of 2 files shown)


sashaweiss-signal:
Hi, thanks for your interest in contributing. I'm not able to reproduce voice message recording or playback being interrupted by orientation change on either iOS 18 or iOS 26. Can you please provide more details as to what issue exactly you're experiencing?

sashaweiss-signal:
Actually, I am able to reproduce specifically with a clockwise rotation, but not counter-clockwise.

sashaweiss-signal:
This should already be prevented by the block of code just above what you added:
```swift
// Don't allow orientation changes while recording voice messages.
if viewState.inProgressVoiceMessage?.isRecording == true {
    return false
}
```
I confirmed we're trying to return `false` there correctly on rotation. However, even though we're returning `false` there, the orientation changes. I think a more appropriate fix here would be to investigate why autorotation is happening even when we indicate it shouldn't.

tr1ng0:
@sashaweiss-signal are you sure this only happens when rotating in a specific direction?
For me, rotating in both directions (clockwise & counterclockwise) stops the audio message. 

Device: iPhone 11
iOS: 18.6.2
Signal Version: 7.74

mogwai:
Yeah still an issue and has been for quite a while

PR #6130, opened 2025-10-17
Fix message list visibility when dismissing keyboard

### LINKED ISSUE #6126: Message list goes "invisible" when hiding the keyboard

### Bug description
Hiding a virtual keyboard by tapping anywhere outside it on the chat screen, or sending media (e.g. a picture), causes the message list to shift down by over a half of the device's screen height.

It is still possible to scroll through messages in this broken state, but the message field, including buttons adjacent to it, is not rendered.

The keyboard can be hidden by scrolling the message list up, which prevents the issue from manifesting. However, once the issue has manifested, re-opening the chat works around the issue until one of the conditions is met.

### Steps to reproduce
- Open any chat
- Tap the text field to show a keyboard
- Tap outside the keyboard in order to hide it

**Actual result:** The message list appears invisible--no messages are displayed.

**Expected result:** The keyboard gets hidden, freeing up screen real estate for the app.

### Screenshots

https://github.com/user-attachments/assets/fd7c9649-c57a-409d-be88-9b1446b1ca2e

### Device info

**Device**: iPhone 6S+

**iOS version**: 15.8.5

**Signal version:** 7.80

### Link to debug log

https://debuglogs.org/ios/7.80.0/4b9737dcbc05eecb604a8bb2e3b974f6fdfc2fbeb285ca15d7c5e1e14ea627c5.zip

its-bash33r-here:
Fixes #6126

## Changes
- Updated collectionView bottom constraint to reference bottomBar instead of superview
- [If needed] Adjusted content insets calculation for consistency

### DIFF
--- a/Signal/ConversationView/ConversationViewController+OWS.swift
+++ b/Signal/ConversationView/ConversationViewController+OWS.swift
@@ -95,7 +95,10 @@ extension ConversationViewController {
         let oldInsets = collectionView.contentInset
         var newInsets = oldInsets
 
-        newInsets.bottom = keyboardLayoutGuide.layoutFrame.height - view.safeAreaInsets.bottom + (bottomBar.frame.height - bottomBar.safeAreaInsets.bottom)
+        // With collectionView constrained to bottomBar.top, we only need to account for
+        // the bottomBar height and safe area insets. The keyboard layout guide height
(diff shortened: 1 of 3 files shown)


max-signal:
I believe we resolved this in v7.82.

PR #6145, opened 2025-11-06
Enable sending Original Quality images with metadata preservation

erikackermann:

### Contributor checklist

 * iPhone 17 Pro Simulator, iOS 26.0.1
 * iPad Pro 13-inch (M4) Simulator, iPad OS 26.0.1
 * iPhone 17 Pro, iOS 26.1.0
 
 Tested sending and receiving from the above devices, plus the following flows:
 - sending to multiple recipients (with varying toggle states)
 - copying images into the chat, sending with full metadata
 - sending from the share extension
 - sending from the Home screen of the app (CameraFirstCaptureSendFlow)

- - - - - - - - - -

### Description

This PR adds an opt-in, per-thread setting that allows users to send images in their original format with full metadata (EXIF, GPS, etc.) preserved. When enabled for a conversation, images are sent as original files (HEIC, PNG, JPEG) up to 100 MiB without compression or metadata stripping.

#### User-Facing Changes

**Thread Settings**

Users can now enable "Original Quality" for individual conversations through a new Image Quality settings screen accessible from the conversation settings menu.

Image Quality Settings Screen

The setting is a simple toggle:
- **Default**: Uses system-standard quality with metadata removed
- **Original Quality**: Sends images in original format with all metadata preserved

A warning message informs users that Original Quality preserves location data and other metadata that could inadvertently reveal where photos were taken.

**Attachment Approval Screen**

When sending images, users see a quality selector that respects the per-thread setting:

Attachment Approval Quality Selector

- For conversations with Original Quality enabled, the selector defaults to "Original"
- In multi-recipient sends, Original Quality applies only to threads where it's enabled
- The UI shows "Standard" and "Original" options (or "Standard" and "High" when Original isn't available)
- Accessible via VoiceOver with adjustable trait support

#### Technical Implementation

**Storage Layer**
- `ImageQualitySettingStore`: Per-thread key-value storage for quality preferences (default vs original)
- `ImageQualityLevel.original`: New enum case with 100 MiB max file size

**Attachment Pipeline**
- `SignalAttachment.preserveMetadata`: Flag to control metadata stripping
- `preparedForOutput(qualityLevel:)`: Defers image conversion; returns original files when quality is `.original`
- `PhotoLibrary`: Always exports original file bytes via `PHAssetResourceManager`; metadata stripping happens in `preparedForOutput()` based on quality level

**UI Integration**
- `ImageQualitySettingsViewController`: New per-thread settings screen with toggle switch
- `AttachmentApprovalViewController`: 2-button quality selector with per-thread default
- `MediaItemViewController`: Fixed preview to use `attachmentStream.decryptedImage()` so "Save from preview" preserves EXIF metadata

**Multi-Recipient Support**
- `AttachmentMultisend`: Processes attachments once at the highest quality needed across all recipients
- Threads with Original enabled receive original files; others receive high/standard quality

#### Testing Notes

- Original Quality setting persists per-thread
- Images sent with Original Quality retain EXIF/GPS metadata
- Preview → Share → Save preserves metadata
- Multi-recipient sends respect individual thread settings
- Standard quality still strips metadata as before
- File size limit enforced at 100 MiB for original quality

#### Migration

No database migration required. The setting is additive and defaults to system behavior for all existing conversations.

### DIFF
--- a/SignalUI/AttachmentMultisend/AttachmentMultisend.swift
+++ b/SignalUI/AttachmentMultisend/AttachmentMultisend.swift
@@ -55,6 +55,7 @@ public class AttachmentMultisend {
 
                 let segmentedAttachments = try await segmentAttachmentsIfNecessary(
                     for: conversations,
+                    destinations: destinations,
                     approvedAttachments: approvedAttachments,
                     hasNonStoryDestination: hasNonStoryDestination,
                     hasStoryDestination: hasStoryDestination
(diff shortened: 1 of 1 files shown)


sashaweiss-signal:
Hi, thanks for your interest in contributing. This is something we've considered in the past, and there are things we want to improve in this space, but unfortunately I don't think this is the direction we'd take. (Due to the surface area and product/UX implications, this type of complex feature is also generally difficult for us to integrate from a public contribution.)

Metadata preservation is something we intentionally don't support, and we believe it's important that that be a uniform decision.

With regards to image quality, we already expose a "Sent Media Quality" setting in Settings > Data Usage and on a per-send basis. We understand that regardless of the Sent Media Quality setting, media sending over Signal on iOS often has subpar performance. Fortunately, we have significant improvements to media quality and performance under active development, as it's something we care a lot about and friction we ourselves experience all the time.

erikackermann:
Hey thanks for your reply.  I understand, this is a large change with many UX implications 😄 

> Metadata preservation is something we intentionally don't support, and we believe it's important that that be a uniform decision.

Can you explain what you mean by uniform decision?  This option could also be implemented similar to disappearing messages (shared setting for all users in a thread) or as a global per-sender setting, but both of these are less privacy-preserving options.

Photo metadata (plus full image quality) is nice for photo archiving and to utilize the full features of photo organization apps.  And it is frustrating that when chatting with trusted contacts (close friends / family), metadata is always removed if sending with the nice photo picker UI.  This can already be worked around by sending photos as attachments, but this UX is terrible; it would be nice if users could enable sending exact byte-for-byte images to select chats.

In any case, this PR demonstrates that this can be achieved with a simple and clean implementation.  I hope you will reconsider this as part of your photo sharing improvements.

PR #6166, opened 2025-11-30
Remove duplicate assets from SignalShareExtension and consolidate resizable imagesets

jsj:

### Contributor checklist
  * iPhone 16 Pro, iOS 18

- - - - - - - - - -

### Description

SignalShareExtension was embedding its own copy of Images.xcassets and Symbols.xcassets, duplicating assets already in the main app bundle. This PR loads images from the containing app instead using the existing `Bundle.app` extension.

Also removes *-resizable imagesets (group, person, note, number, invite) which were byte-for-byte identical copies of their base imagesets with the same Contents.json.

### Changes

- Remove Images.xcassets and Symbols.xcassets from SignalShareExtension build phase
- Use `Bundle.app` extension to load images from the containing app
- Delete duplicate *-resizable imagesets  
- Update code references to use base imageset names

### Impact

Reduces app bundle size by ~10MB.


Signal/Symbols.xcassets/group/group-resizable.imageset/Contents.json
```diff
@@ -1,16 +0,0 @@
...
-    }
-  ],
-  "info" : {
-    "author" : "xcode",
-    "version" : 1
-  },
-  "properties" : {
-    "preserves-vector-representation" : true,
-    "template-rendering-intent" : "template"
-  }
```
sashaweiss-signal: Importantly, this "Use Vector Data" setting was removed and is not present for the other asset. When these were added, presumably this is what distinguished the "-resizable" variants.

It may be possible for us to replace the non-resizable one with the resizable, because we should be able to enable this for all PDF assets, but we shouldn't implicitly disable this.

PR #6178, opened 2025-12-25
Fix dimming animation flicker

### LINKED ISSUE #6179: Dimming animation flickers when tapping on parent message quickly

### Bug description

When you reply to a message, it includes a link to the parent message in the reply. Tapping on this link triggers a dimming animation on the parent, but if you tap quickly this leaves some flickering after you stop tapping.

### Steps to reproduce
- Reply to a message
- Tap the link to the parent message multiple times quickly and then stop

**Actual result:** The parent message continues to flicker for a second or two
**Expected result:** The final tap effect fades out normally

### Screenshots

Unfortunately a bit difficult to see since you can't tell when the taps trigger a flicker and when it's due to this bug:

https://github.com/user-attachments/assets/7bf46cdb-d29e-4025-9b52-9027652e8665

### Device info

**Device**: iPhone 12

**iOS version**: 26.1

**Signal version:** 7.89

### Link to debug log
N/A

smchase:

### Contributor checklist

 * iPhone 16, iOS 26.2 (xcode simulator)

https://github.com/user-attachments/assets/1bb90889-e7bf-4e0f-9ef5-8b1d894a5894

- - - - - - - - - -

### Description

Fixes #6179, see it for context on the issue.

This bug is due to the fact that the parent message dimming animation is composed of a `fadeIn` and separate async `fadeOut`. When we tap in quick succession we call `dimmerLayer.removeAllAnimations()` to remove any old animation, but this doesn't include the `fadeOut` if the async hasn't fired yet. This means we end up with multiple old async `fadeOut`s causing the parent message to flicker once the tapping stops.

The fix is to make this into one sync animation that contains the fade in and fade out. This way when the link is tapped again we clear the entire old animation including the fade out. This seems simpler and cleaner than tracking the async queue and clearing from there, and more logically correct to treat the animation as atomic.

### DIFF
--- a/Signal/ConversationView/CellViews/CVColorOrGradientView.swift
+++ b/Signal/ConversationView/CellViews/CVColorOrGradientView.swift
@@ -337,25 +337,13 @@ public class CVColorOrGradientView: ManualLayoutViewWithLayer {
 
         dimmerLayer.removeAllAnimations()
 
-        // Animate fade-in.
-        let fadeIn = CABasicAnimation(keyPath: #keyPath(CALayer.opacity))
-        fadeIn.fromValue = 0
-        fadeIn.toValue = 1
-        fadeIn.duration = stepDuration
-        fadeIn.fillMode = .forwards
-        fadeIn.isRemovedOnCompletion = false
-        dimmerLayer.add(fadeIn, forKey: "fadeIn")
-
-        // Schedule fade-out after delay.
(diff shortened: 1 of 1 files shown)


smchase:
first PR here so apologies if I missed anything, but hopefully this is a small enough change with enough context on the issue to be easy to review :)

PR #6194, opened 2026-01-16
Fixes #6193, fixes typing: IMPORTANT this is not (only) a translation change, this…

benediktjohannes:
Fixes #6193, fixes typing: IMPORTANT this is not (only) a translation change, this is an internal change of the value field for the translations

Contributor checklist
[ x] I am following the [Code Style Guidelines](https://github.com/signalapp/Signal-Android/wiki/Code-Style-Guidelines)
[ x] My contribution is fully baked and ready to be merged as is
[ x] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the Fixes #1234 [syntax](https://help.github.com/articles/closing-issues-via-commit-messages/)

**Description**

Fixes typing: Important this is not (only) a translation change, this is an internal change of the value field for the translations - another important thing to mention: This (probably) has to be changed in all language translation files as well and also in the general value context if it's wrong there as well and copied

PR #6195, opened 2026-01-16
Fix typing error in class

benediktjohannes:
This (for example) has to be fixed as well as a consequence of #6194 

But there are still several more issues with those typos inside classes, unforetunately, but I'll try to fix them as soon as possible


ankushkushwaha (approved):

sashaweiss-signal:
Thanks for your interest in contributing. While this kind of typo is awkward, it doesn't actually hurt anything – these string keys are arbitrary, and only a reference to our `Localizable.strings` files. Given that, we're going to keep the keys as they are. (Changing string keys also has implications for our translation pipeline, and generally ends up costing us money unnecessarily.) Consequently, I'm going to close this PR.

PR #6202, opened 2026-02-04
Show most recent messages in 3D Touch conversation preview

aivazashvilitekla:
Previously, when using 3D Touch to preview a conversation from the chat list, the preview would scroll to the first unread message. This made it difficult to quickly check the current state of a conversation without opening it.

This change adds a `scrollToNewestMessages` parameter to `ConversationViewController.load()` and uses it for 3D Touch previews, so users can now see the most recent messages when previewing a chat.

### Contributor checklist
 * Build verified on iOS Simulator (iPhone 17 Pro Max)
 * Note: Server-connected testing requires official signing certificates (per CONTRIBUTING.md known limitations)

- - - - - - - - - -

### Description

**Problem:**
When using 3D Touch (or long-press on newer devices) to preview a conversation from the chat list, the preview scrolls to the first unread message instead of showing the most recent messages. This prevents users from quickly glancing at the current state of a conversation.

**Solution:**
Added a `scrollToNewestMessages` parameter to `ConversationViewController.load()` with a default value of `false` to maintain backward compatibility. For 3D Touch previews, this is set to `true` so the preview displays the most recent messages.

**Changes:**
- `ConversationViewController.swift`: Added `scrollToNewestMessages: Bool = false` parameter
- `ChatListViewController+Helpers.swift`: Pass `scrollToNewestMessages: true` when creating preview

**Testing:**
- Build compiles and runs successfully on iOS Simulator
- Code logic verified through review
- Default parameter value ensures no impact on existing callers

### DIFF
--- a/Signal/ConversationView/ConversationViewController.swift
+++ b/Signal/ConversationView/ConversationViewController.swift
@@ -67,6 +67,7 @@ public final class ConversationViewController: OWSViewController {
         threadViewModel: ThreadViewModel,
         action: ConversationViewAction,
         focusMessageId: String?,
+        scrollToNewestMessages: Bool = false,
         tx: DBReadTransaction,
     ) -> ConversationViewController {
         let thread = threadViewModel.threadRecord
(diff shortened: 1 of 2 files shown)


sashaweiss-signal:
Hi, thanks for your interest in contributing. We've discussed this internally, and have decided to keep the behavior as-is for the time being.

It's an interesting idea, and I appreciate that your changes relied on existing infrastructure, but there are some complicated UX implications for the change. For example, it could be confusing if the preview showed you the latest messages, but when you opened the chat you were dropped back at the last-unread message. (We wouldn't want to fast-forward you to the latest messages on opening the chat, since that has implications for "read" state.)

If we do decide to change this behavior in the future, we'll come back and revive these changes for attribution purposes. Thanks again for your interest in contributing, and as always for being a Signal user.

PR #6203, opened 2026-02-04
Secreat Gesture

Vasanthadabala:
Added Secreat Gesture aduio recording feature


sashaweiss-signal:
Hi, thanks for your interest in contributing. It's not clear to me what the intent of this secret gesture is, or what problem it solves. We also generally avoid "secret" or alternative app behaviors, since we'd prefer to either improve a feature for all users, and/or have restrictions on what features we're willing or able to build. As a result, I'm going to close this PR.

PR #6204, opened 2026-02-10
fixes device transfer

decanus:
<!-- You can remove this first section if you have contributed before -->
<!-- replace the empty checkboxes [ ] below with checked ones [x] accordingly -->

### Contributor checklist
<!-- replace the empty checkboxes [ ] below with checked ones [x] accordingly -->
- [ ] My contribution is fully baked and is ready to be merged as is
 * iDevice A, iOS X.Y.Z
 * iDevice B, iOS Z.Y

- - - - - - - - - -

### Description

This pull requests potentially resolves issues around device transfers, what would be the best way to fully test them end to end?

  Summary                                                                                                                                                                                                            
                                                                                                                                                                                                                     
  - Add checkpoint-based resume capability for interrupted device transfers, allowing transfers to continue from where they left off rather than restarting from scratch                                             
  - Implement automatic connection retry with exponential backoff (up to 3 retries) when connections are lost during transfer                                                                                        
  - Add background grace period (30 seconds) instead of immediately terminating when the app enters background, giving users time to return to the app                                                               
  - Extend peer invitation timeout from 30 to 120 seconds to accommodate slower connection establishment

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -128,11 +128,11 @@ CHECKOUT OPTIONS:
     :tag: v4.6.1-f_barrierfsync
 
 SPEC CHECKSUMS:
-  blurhash: f588a9d659f4c742a325bbbbd438d1ef3c34af19
+  blurhash: 4b3b2b8909cf41ab59fa349788654b8387284ff4
   BonMot: fb2b6a2209cb3149aca37b7131d49c051c04ae86
   CocoaLumberjack: 543c79c114dadc3b1aba95641d8738b06b05b646
   GRDB.swift: 1395cb3556df6b16ed69dfc74c3886abc75d2825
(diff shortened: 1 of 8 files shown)


sashaweiss-signal:
Hi, thanks for your interest in contributing. You're right that Device Transfer is a complex and painful feature to use, and we have plans to overhaul it in the medium-term future, especially now that we have complementary features like Backups.

At the same time, we don't generally accept large PRs, especially ones that touch complex features and will require lots of manual testing and support from Signal Staff. If you're able to reproduce particular issues, and make changes that you're able to demonstrate fix those issues, we could consider those fixes individually – but this kind of large speculative fix isn't realistically something we can merge.

PR #6205, opened 2026-02-12
Reset keyboard to ABC layout after send

ddxor:
### Contributor checklist
 * iPhone 17, iOS 26.3
 * iPhone 17 Pro Max, iOS 26.3

- - - - - - - - - -

### Description
It is standard behavior on iOS messaging apps (iMessage, WhatsApp, etc) that when you send a message, the current input pane is returned to the default "ABC" pane.

For example: in iMessage, if I write the message "Hello, World!", and tap send, my keyboard will be set back to "ABC" ready to start my next message. However, on Signal, when I write "Hello, World!", and tap send, the input pane remains on punctuation, and I have to tap the "ABC" button to start typing my next message.

This adds friction, and is counter to the natural flow of iOS. Further, Signal for Android does not exhibit this behavior: it works as iMessage does and resets to "ABC" after send.

I contributed a PR for this in 2023, but it was reverted, because that change reset the keyboard to `.alphabet` which caused non-latin languages to change to English unexpectedly. This PR uses `.default` to avoid that issue.

### DIFF
--- a/Signal/ConversationView/ConversationInputToolbar.swift
+++ b/Signal/ConversationView/ConversationInputToolbar.swift
@@ -1583,6 +1583,22 @@ public class ConversationInputToolbar: UIView, QuotedReplyPreviewDelegate {
         editTarget = nil
         setMessageBody(nil, animated: animated)
         inputTextView.undoManager?.removeAllActions()
+        resetKeyboardLayout()
+    }
+
+    /// Resets the iOS keyboard from the symbols/numbers pane back to the
+    /// default alphabetic layout by toggling ``keyboardType``. Each
+    /// reload is required — the first forces the keyboard to tear down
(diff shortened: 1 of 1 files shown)


ddxor:
@harry-signal @jim-signal @amy-signal @gte-signal @jkt-signal would appreciate if you'd be able to do a brief test of a non-Latin language on this before merging it. I have tested, but would appreciate a 2nd pair of eyes.

sashaweiss-signal:
Thanks for this contribution! I've merged it internally as d022cfd5a12444ec7cad17b0a9bd817372a9372d (preserving the authorship as it is here), which will become available publicly soon (probably next Wednesday).

Thanks again, and as always for being a Signal user!

PR #6226, opened 2026-03-18
Fix accessibility of attachment keyboard buttons

meherkasam:
### Contributor checklist
 * iPhone 17 Pro Simulator, iOS 26.0 using Accessibility Inspector. I also verified my changes using Reveal App.

- - - - - - - - - -

### Description
This change fixes the accessibility of the buttons in the attachment keyboard by combining the icon and the label into a single accessibility element.

### Before
Before

### After
After


max-signal:
@meherkasam we made a similar patch (9257a77e008936df908d04a4e5797acbca655f3e), and I think that one may make this one obsolete.

PR #6229, opened 2026-03-25
Implement reacting to messages and stories with stickers

diaruga777:

### Contributor checklist

 * iPhone 17, iOS 26.2
 * iPad A16, iOS 26.2

- - - - - - - - - -

Hello Signal team! This is a speculative implementation of Discord-like or Slack-like custom reactions in Signal. It adds the ability to react with stickers in addition to emoji across the board*. This includes animated stickers!

The implementation is, in my opinion, production-ready. By necessity, I have made opinionated decisions on some details both from an engineering and a UX design perspective. I’ve tried to document the main ones below.

It works by adding a Sticker proto field to Reaction proto(s). Reaction sticker sending and receiving works the same as for sticker messages; sticker image data is uploaded to cdn and sent as an attachment alongside sticker identifier info, and the receiver either uses a locally installed sticker pack if available or downloads from CDN.

The protocol changes are backwards compatible; clients without sticker reaction proto fields silently fall back to treating the sticker as its associated emoji. Of course implementations on the other client platforms would be necessary to actually ship this, but I hope these changes are simple enough at a protocol level, and compelling enough at the iOS client level, to inspire that work to happen! :)

That said, I will not be offended if this goes nowhere, please take it or leave it! I just want this feature as a user, and had my own extrinsic reasons for working on an open source app feature.

Some decisions I made which a reasonable person could disagree with:
* ```Reaction stickers’ Attachments are owned by the TSMessage being reacted to, not the OWSReaction```
_This was to avoid the overhead of a new ReactionAttachmentReference table. But it does mean the cascade delete when an OWSReaction is deleted is handled in Swift, not using SQL triggers._
* ```Relatedly: no index on MessageAttachmentReference reactionRowId```
_In the typical case there aren’t that many reactions on a single message; doing O(n) lookup should be fine. But it would also be okay to add an index at any point in the future._
* ```Reaction stickers always auto-download```
_I didn't add a UX hook for the user to manually download reaction stickers. Failures fall back to the sticker’s emoji.
This is probably the most frail part of the system, though I’m not sure how to do it properly, since tapping reaction pill views already opens the reaction sheet. I made tapping stickers in the sheet launch the sticker pack install sheet; perhaps tapping undownloaded stickers can trigger a download? That's buried several taps deep, which is why I didn't do it._
* ```Use 💌 as the emoji for stickers with no emoji```
_Emoji are used as fallbacks at client and protocol level, so I needed something when no emoji is present._
* ```Call reactions only have access to emoji```
_I think a different protocol would be necessary to handle upload and transmission of stickers during a call. (Would it upload at all, or just send the sticker image data in the call audio/video stream?) So for now its untouched and just uses emoji._

Here’s some screenshots of all of it working (lightly edited for privacy):

https://github.com/user-attachments/assets/7eef433e-0140-4ec5-898e-4f23fa959a65

https://github.com/user-attachments/assets/073c8fbb-c190-44a4-9202-2acb3a92cea6

Simulator Screenshot - iPhone 17 - 2026-03-25 at 21 30 13
Simulator Screenshot - iPhone 17 - 2026-03-25 at 21 24 20

https://github.com/user-attachments/assets/be81adeb-0036-418c-a349-0be51e5ad5d5

https://github.com/user-attachments/assets/5fd760e5-1c4b-4271-9da0-f4dedf4a3337

### DIFF
--- a/SignalServiceKit/Messages/Interactions/Quotes/QuotedReplyManager.swift
+++ b/SignalServiceKit/Messages/Interactions/Quotes/QuotedReplyManager.swift
@@ -531,7 +531,7 @@ class QuotedReplyManagerImpl: QuotedReplyManager {
                         sticker: attachment,
                         stickerInfo: originalMessage.messageSticker?.info
                     ),
-                    stickerThumbnail: resizedThumbnailImage
+                    stickerThumbnail: UIImage(cgImage: thumbnailImage)
                 ))
             }
 
--- a/SignalServiceKit/Messages/Reactions/OutgoingReactionMessage.swift
+++ b/SignalServiceKit/Messages/Reactions/OutgoingReactionMessage.swift
@@ -152,14 +152,8 @@ final class OutgoingReactionMessage: TransientOutgoingMessage {
                     for: .messageReactionSticker(messageRowId: messageRowId, reactionRowId: reactionRowId),
                     tx: tx,
                 ),
-                let transitPointer = referencedAttachment.attachment.asTransitTierPointer(),
-                case let .digestSHA256Ciphertext(digest) = transitPointer.info.integrityCheck
+                let attachmentProto = referencedAttachment.asProtoForSending()
             {
-                let attachmentProto = DependenciesBridge.shared.attachmentManager.buildProtoForSending(
-                    from: referencedAttachment.reference,
-                    pointer: transitPointer,
-                    digestSHA256Ciphertext: digest,
-                )
                 let stickerBuilder = SSKProtoDataMessageSticker.builder(
                     packID: sticker.packId,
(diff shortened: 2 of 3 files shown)


sashaweiss-signal:
Thanks for your interest in contributing! This is a neat feature, and it's something we've discussed periodically internally.

Unfortunately, this is too large of a change for us to accept as an external contribution. Beyond just the changes on iOS, as you noted this would require substantial changes on other clients and rigorous evaluation of the opinionated decisions you listed.

I appreciate the effort you put into this, and hopefully we'll have space in our roadmap to implement this in the future! Thanks again, and as always for being a Signal user.

PR #6241, opened 2026-04-16
Remove exclamation point from GENERIC_INCOMING_MESSAGE_NOTIFICATION, …

WyattJChalifoux:
…lowercase "message" in English.

- [x ] I have read the [README](https://github.com/signalapp/Signal-iOS/blob/main/README.md) and [CONTRIBUTING](https://github.com/signalapp/Signal-iOS/blob/main/CONTRIBUTING.md) documents

### Contributor checklist

! I don't have an Apple device, I am making this change on behalf of my iOS user friends who suggested it. Since the change is straightforward, I am hoping that either someone else would test it or that this step can be bypassed.

- - - - - - - - - -

### Description
The Android version of the Signal app displays "New message" for generic message notifications, whereas the iOS version displays "New Message!", which is less tone-neutral and incorrectly capitalized (an inconsistency with English vs other languages). In order to make the app more consistent across platforms, and at the suggestion of multiple iOS users I know who find the exclamation point "stressful" or "annoying", I have removed the exclamation point and fixed the capitalization. I have also removed the exclamation point from the message in other languages for consistency. I did not change the capitalization of any other languages (i.e German).

### DIFF
--- a/Signal/translations/ar.lproj/Localizable.strings
+++ b/Signal/translations/ar.lproj/Localizable.strings
@@ -3680,7 +3680,7 @@
 "GENERIC_ATTACHMENT_LABEL" = "ملف مُرفَق";
 
 /* notification title indicating the user generically has a new message */
-"GENERIC_INCOMING_MESSAGE_NOTIFICATION" = "رسالة جديدة!";
+"GENERIC_INCOMING_MESSAGE_NOTIFICATION" = "رسالة جديدة";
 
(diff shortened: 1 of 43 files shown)


max-signal:
I merged an English-only variant of this that should be available in our v8.9 beta. (We only need to modify the English variant because other languages will pick up the change via our standard localization pipeline.)

PR #6245, opened 2026-04-25
Fix iPhone calling UI not rotating with device

### LINKED ISSUE #4239: [video call] no display rotation

### Bug description
During a video call, the screen will not rotate when turning the iPhone from portrait to landscape orientation. 

### Steps to reproduce
- start a video call with someone (in portrait orientation)
- turn the iPhone to landscape 

**Actual result:** the video of your partner does not rotate correspondingly. Thus, your conversation partner is not displayed upright but 90° turned. 

**Expected result:** the video of your partner should be rotated in correspondence with the orientation of the iPhone.

Note: interestingly, the tiny video of yourself that is taken by the front camera is rotated correctly.

### Device info
**Device**: iPhone X, iPhone Air 2

**iOS version**: 12.4

**Signal version:** 2.41.0.21

timomak:
- [x] I have read the README and CONTRIBUTING documents
- [x] I have signed the Contributor Licence Agreement

### Contributor checklist
   * iPhone 14 Pro, iOS 26.0.1
   * iPad Pro 12.9" (5th generation, M1), iPadOS 26.4

- - - - - - - - - -

### Description
Fixes #4239 — iPhone video-call UI not rotating with the device.

#### Root cause

`AppDelegate.application(_:supportedInterfaceOrientationsFor:)` had an unconditional early return of `.portrait` for every iPhone UIWindow whenever any call was active. The comment at the call site said the lock was for the call-banner (return-to-call) window only, but the implementation ignored the `window` parameter and locked every window — overriding the per-VC orientation policies the rest of the calling UI was already configured for. This made the iPhone call UI portrait-locked while iPad's call UI rotated freely. Around this lock, several workarounds had accumulated to compensate for it; those are removed here as well.

#### Recording
https://github.com/user-attachments/assets/9a18396c-5a4c-4fc5-8a55-f5121f9226a2

#### Changes

1. `AppDelegate.swift` — Remove the unconditional `.portrait` early return; honor the `window` parameter in the rootVC fallback so each window reports its own rootVC's allowed orientations.
2. `IndividualCallViewController.swift` — Replace the iPhone `.portrait` override with `defaultSupportedOrientations` so the per-VC value is consistent with what `WindowRootNavigationViewController` advertises for the call window. Also remove the `sendInitialPhoneOrientationNotification` call (no longer needed; see #6 below).
3. `LocalVideoView.swift` — Remove the `CGAffineTransform` compensation that existed only as a workaround for the now-removed portrait lock. `RTCCameraPreviewView` already updates its own `connection.videoOrientation` on `UIDeviceOrientationDidChangeNotification`, so the preview rotates correctly without the manual transform.
4. `RemoteVideoView.swift` — Remove the `shouldOverrideRotation` branch and its per-orientation rotation override. With the parent view now rotating with the device, `RTCMTLVideoView` reads `frame.rotation` natively and renders correctly in any orientation. The `videoContentMode` aspect-match logic at the bottom of `renderFrame` is unchanged.
5. `ReturnToCallViewController.swift`:
   - `supportedInterfaceOrientations` allows iPhone to rotate (was portrait-only); the PIP banner now rotates with the device, like the rest of the calling UI.
   - `inherentPipSize` is orientation-aware on iPhone (it was already on iPad); the PIP tile reflows between 9:16 (90×160) and 16:9 (160×90).
   - Added a `UIDevice.orientationDidChangeNotification` observer that drives a layout pass; the small PIP window's frame doesn't change size automatically on device rotation, so `viewWillTransition` and `viewWillLayoutSubviews` aren't fired by UIKit. The observer is the explicit hook that keeps the PIP frame in sync with device orientation.
6. `CallService.swift`, `CallButton.swift`, `CallMemberView.swift`, `CallMemberCameraOffView.swift`, `CallMemberChromeOverlayView.swift`, `CallMemberVideoView.swift`, `CallMemberWaitingAndErrorView.swift`, `GroupCallViewController.swift` — Remove the `phoneOrientationDidChange` notification system (`shouldReorientUI`, `sendPhoneOrientationNotification`, `sendInitialPhoneOrientationNotification`, the `rotateForPhoneOrientation` protocol method and its four implementations, and the corresponding `NotificationCenter` observers in `CallButton` and `CallMemberView`). This system manually rotated icons (mute/end/flip-camera buttons, mute indicator, avatar) by ±π/2 to compensate for the portrait-locked window. With UIKit auto-rotation now in effect, those transforms compounded with the parent's rotation and rendered the icons sideways. The `begin/endGeneratingDeviceOrientationNotifications` lifecycle in `CallService` is preserved because `ReturnToCallViewController` and `LocalVideoView` still rely on `UIDevice.orientationDidChangeNotification`.
7. `WindowManager.swift` — Remove the `ows_setOrientation(.portrait)` call in `startCall(viewController:)`. It existed because the call UI used to be portrait-locked and would not auto-switch from a landscape orientation; with the call UI now reporting `defaultSupportedOrientations`, UIKit picks the appropriate orientation on its own.
8. `GroupCallVideoOverflow.swift` — Remove the manual cell-rotation override (`OrientationOverride`, `updateOrientationOverride`, the `UIDevice.orientationDidChangeNotification` observer, and the `cell.transform` branch in `willDisplay`). It only triggered during screen-share-on-iPhone-in-landscape and existed because the cells were in a portrait-locked parent. With UIKit rotation, cells follow the parent naturally; the manual transform was double-rotating them.

#### Test plan

Tested with Signal Desktop on macOS as the second party in calls. Devices:
- iPhone 14 Pro, iOS 26.0.1
- iPad Pro 12.9" (5th generation, M1), iPadOS 26.4

iPhone:

- 1:1 call to Mac, both portrait: looks correct.
- 1:1 call, rotate to landscape: full UI rotates, remote video remains upright (no double-rotation), call-control buttons and on-tile mute indicator are upright (no icon double-rotation), Mac peer sees iPhone user upright throughout (CVO via RTP carries rotation correctly).
- 1:1 call started while device is already in landscape: call opens in landscape (no forced flip to portrait).
- 1:1 audio-only call, rotate to landscape: UI rotates, controls upright.
- Group call (group thread), both portrait: looks correct.
- Group call (group thread), rotate to landscape: grid rotates, tiles reflow, controls upright.
- Call Link (group) to Mac, both portrait: looks correct.
- Call Link (group), rotate to landscape: grid rotates, tiles reflow.
- In landscape call, minimize to PIP: banner appears in landscape (160×90).
- Rotate while in PIP: banner reshapes between 9:16 and 16:9.
- Tap PIP banner to expand: returns to full call in current orientation.
- Local self-preview tile (in-call): reflows from 9:16 to 16:9 on rotation.
- Front and back camera both behave correctly.

iPad:

- 1:1 call to Mac, both portrait: looks correct.
- 1:1 call, rotate to landscape: full UI rotates, remote video upright, controls upright.
- 1:1 call started while device is already in landscape: opens in landscape.
- 1:1 audio-only call, rotate to landscape: UI rotates, controls upright.
- Group call (group thread), both portrait: looks correct.
- Group call (group thread), rotate to landscape: grid rotates, tiles reflow, controls upright.
- Call Link (group) to Mac, both portrait: looks correct.
- Call Link (group), rotate to landscape: grid rotates, tiles reflow.
- In landscape call, minimize to PIP: banner appears in landscape (272×204).
- Rotate while in PIP: banner reshapes between 4:3 and 3:4 (272×204 ↔ 204×272).
- Tap PIP banner to expand: returns to full call in current orientation.
- Local self-preview tile (in-call): reflows on rotation.
- Front and back camera both behave correctly.

A short screen recording of the rotation behavior is attached as a comment below.

Known limitations:
- `GroupCallVideoGridLayout` prefers more rows than columns when sizing the grid (a longstanding choice that suits portrait). On iPhone in landscape with two participants, that produces a vertical 2×1 stack of wide-and-short tiles, which is functional but not visually ideal. This is a pre-existing layout choice that the rotation fix simply exposes on iPhone; better handled in a separate PR that revisits per-orientation column/row preferences across iPhone and iPad.


max-signal:
It’s intentional that calls don’t rotate on iPhone right now. We will likely address this at some point in the future, but we’re not planning on changing it right now.

timomak:
@max-signal thanks for taking a look. Before I drop this, a few things worth weighing.

#4239 has been open since 2019, so this isn't a hypothetical concern. The bigger issue is technical:

1. The comment at the AppDelegate call site scopes the portrait lock to `ReturnToCallWindow` only, but the implementation ignores the `window` parameter and applies it to every window. That suggests the iPhone-wide effect is implementation drift, not the original intent.

2. Current behavior isn't "iPhone calls don't rotate." It's a mix:
   - Local self-preview rotates (via `RTCCameraPreviewView`'s built-in handling).
   - Remote video doesn't.
   - iPad's call UI rotates freely.
   - RTP CVO already sends device orientation to peers, so a Mac peer sees the iPhone user upright in landscape. The iPhone user just can't see the peer the same way.

   That asymmetry reads like a bug rather than a product decision.

3. The codebase carries several pieces whose sole purpose is to compensate for the portrait lock: `shouldOverrideRotation` in `RemoteVideoView`, `OrientationOverride` in `GroupCallVideoOverflow`, the `CGAffineTransform` in `LocalVideoView`, and the `phoneOrientationDidChange` notification system that hand-rotates control icons by ±π/2. Code written to fight a constraint usually means the constraint wasn't load-bearing.

I know this PR is larger than your "smaller is better" guidance prefers. The fix and the workaround removals are coupled (removing the lock without removing the workarounds leaves icons double-rotated). If splitting it would help review even at the cost of a transient broken state on the intermediate commit, I can do that.

Is there a redesign in flight that depends on portrait-only iPhone calls? If so, that context would help. If not, I think this is worth a second look. The current state isn't "iPhone calls don't rotate"; it's a half-rotated state that users are living with today, and any redesign will inherit it.

Either way, looking forward to whatever you have planned here. Landscape video calls on iPhone becoming a reality on Signal will be great whenever it ships.

PR #6280, opened 2026-05-29
Fix immediate sends with pending link previews

bytware-alpha:
## Why

Sending a link should behave like sending any other message. The bubble should appear immediately; the preview can finish a beat later and still belong to that same bubble.

[#4550](https://github.com/signalapp/Signal-iOS/issues/4550#issuecomment-689978354) already ruled out the cheap fixes: blocking Send is bad UX, and sending the preview later is not what the user chose to send. So this does the actual thing: persist the outgoing row now, keep the preview attached to that send, and only hand the finalized payload to the sender.

This is better than the earlier durable-job shape because it does not teach the job queue about half-built sends. No runnable job exists until the message is final. Ordering is held with a small in-memory barrier, which is the sort of short-lived queueing pattern maintainers have accepted before in [#4004](https://github.com/signalapp/Signal-iOS/pull/4004#discussion_r232011159) and [#2529](https://github.com/signalapp/Signal-iOS/pull/2529#discussion_r138754365). Less machinery, cleaner contract.

## How

- Keep an in-flight link preview alive when Send wins the race.
- Insert the outgoing message row immediately so the conversation stays stable.
- Hold sender ordering with an in-memory deferred job barrier.
- Attach the preview to the same row, then create the real sender job.
- Keep edit/share approval paths using the same preview handoff.

## Test Plan

- `xcodebuild test -workspace Signal.xcworkspace -scheme Signal -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.5' -only-testing:SignalServiceKitTests/MessageSenderJobQueueTest/test_deferredJobBlocksLaterJobsUntilStarted -only-testing:SignalTests/ThreadUtilSignalUITest/testPendingLinkPreviewDoesNotBlockOutgoingRowPersistence`
- `xcodebuild test -workspace Signal.xcworkspace -scheme Signal -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.5' -only-testing:SignalUITests/LinkPreviewFetchStateTest/testLinkPreviewDraftForSendingTaskSurvivesClearingCurrentText`
- `PATH="/tmp/signal-swiftformat/SwiftFormat/.build/release:$PATH" Scripts/precommit.py --ref origin/main`
- `git diff --check`

### DIFF
--- a/Signal.xcodeproj/project.pbxproj
+++ b/Signal.xcodeproj/project.pbxproj
@@ -594,6 +594,7 @@
 		4C6E6C6924241C00009DE948 /* ConversationViewControllerTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C6E6C6824241C00009DE948 /* ConversationViewControllerTest.swift */; };
 		4C751BE523FA0284002A8AF1 /* ContactSupportActionSheet.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C751BE423FA0284002A8AF1 /* ContactSupportActionSheet.swift */; };
 		4C83AC4223C55D9C00D4F2E6 /* SignalBaseTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C83AC4123C55D9C00D4F2E6 /* SignalBaseTest.swift */; };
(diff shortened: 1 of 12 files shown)


sashaweiss-signal:
Thanks for your interest in contributing. Unfortunately, we don't generally accept external contributions of this size since they're difficult for us to validate and maintain. Running the unit tests is not sufficient validation.

bytware-alpha:
Wild.

Issue #6281, opened 2026-05-29
Sent message sound is not the same as iOS default


prolud:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
The message sent audio is not actually in the best quality and it is not the same as iMessage sound.

### Steps to reproduce
- send a message in the chat

**Actual result:** the sent_message.aiff audio is not the same as iPhone default

**Expected result:** the sent_message.aiff should play ios default sent message audio

### Device info

**Device**: iPhone 13

**iOS version**: 26.5

**Signal version:** 8.12.1 (1616)

prolud:
https://github.com/user-attachments/assets/ebaa4263-f8a8-4683-907d-d6eb89c22f5f

prolud:
my pull request to fix this:
https://github.com/signalapp/Signal-iOS/pull/6282

sashaweiss-signal:
Hi, thanks for reporting. We plan to keep the outgoing-message sound as it for now.

PR #6282, opened 2026-05-29
[bugfix/improve-message-sent] fix: change message_sent.aiff to the default one #6281

prolud:
this pull request just changes de message_sent.aiff audio to the iOS default one.


prolud:
this pull request is associated with issue https://github.com/signalapp/Signal-iOS/issues/6281

sashaweiss-signal:
Hi, thanks for your interest in contributing. We plan to keep the outgoing-message send as it is for now.

PR #6285, opened 2026-05-31
Add live audio waveform animation

Chaus:
### Problem Statement

When recording a voice message, the UI previously lacked visual feedback indicating the level of audio being captured. This made it difficult for users to confirm that their microphone was picking up their voice clearly and at the right volume before the message was actually sent. Common scenario: the bluetooth headset did not disconnect from the phone while the user left the headset in another room and is recording an audio using their phone only.

### Solution

This PR introduces a brand new **Live Audio Waveform** that animates in real-time while a voice message is being recorded.

**Key Features:**
1. **Real-time Visualization:** A dynamic, scrolling waveform that maps directly to the microphone's input power levels, providing immediate and responsive visual feedback to the user.
2. **Seamless Lock State Support:** The waveform automatically expands and anchors itself between the `Cancel` button and the audio message time - taking only the space that has previously been empty and unused.
3. **Smooth Aesthetics:** Uses hardware-accelerated `CABasicAnimation` for smooth 60fps scrolling, along with a `CAGradientLayer` mask on the leading edge so that older audio samples gracefully fade into transparency as they scroll out of view.

- - - - - - - - - -

### Contributor checklist
 * iPhone 12 Mini, iOS 17.4

### DIFF
--- a/Signal.xcodeproj/project.pbxproj
+++ b/Signal.xcodeproj/project.pbxproj
@@ -1741,6 +1741,7 @@
 		A1A018521805C5E800A052A6 /* QuartzCore.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = A11CD70C17FA230600A2D1B1 /* QuartzCore.framework */; };
 		A1A018531805C60D00A052A6 /* CoreGraphics.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = D221A091169C9E5E00537ABF /* CoreGraphics.framework */; };
 		A5E7C675248C5443007C949A /* InfoPlist.strings in Resources */ = {isa = PBXBuildFile; fileRef = A5E7C673248C5442007C949A /* InfoPlist.strings */; };
+		A68C407AAE86CE67FAC11F79 /* LiveAudioWaveformView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0484486DA7CD91AB4D8E771E /* LiveAudioWaveformView.swift */; };
 		B60EDE041A05A01700D73516 /* AudioToolbox.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = B60EDE031A05A01700D73516 /* AudioToolbox.framework */; };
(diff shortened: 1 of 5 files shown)


sashaweiss-signal:
Thanks for your interest in contributing! Unfortunately, we don't generally accept external contributions of "features" like this because of the review and ongoing maintenance cost. This is a neat idea though, and I'll pass it along to our design team as we have improvements planned for Voice Notes overall! If we do decide to adopt this along with our other improvements, I'll reopen this PR.

Chaus:
Thank you for your response @sashaweiss-signal. 

For the context, the situation when you record a 3-minute voice message - only to realize your headset was still connected inside your bag, and Signal was picking up the mic from in there the whole time - happens to me, my girlfriend and virtually all my friends every other day. It's quite frustrating so we all would really love to have this kind of feature. 

And the space between the recording timer and the Cancel button is currently empty and unused, so it seemed like a very intuitive place for the live waveform. If there is anything I can do for this feature to be adopted, please let me know!

Best wishes to you and the team, thank you for everything you've done for Signal - I love it.

image

sashaweiss-signal:
Makes sense – there are definitely voice message papercuts we need to address. Can you provide some demo videos of how this looks, which I can pass along and will help us prioritize?

Chaus:
@sashaweiss-signal sure, here are the demos for two use cases:

1. Pull-up the mic button
 
https://github.com/user-attachments/assets/5798cb81-5c37-4564-9340-8e491c4ea557

4. Press-and-hold the mic button

https://github.com/user-attachments/assets/75952f9e-a74f-47f9-9ed6-54e67ff24ea7

sashaweiss-signal:
Thanks @Chaus, I'll make sure this gets looked at!

sashaweiss-signal:
We recently added something like this to Desktop, so I'm going to reopen the PR and mark it as Acknowledged for now, and I've linked it to our internal tracking system. I can't guarantee when we'll take this up, but we'll circle back here when we do!

Chaus:
Thank you @sashaweiss-signal!

PR #6326, opened 2026-08-13
Fix Screen Lock authentication while device is locked

JustAlex0000:
## Contributor checklist

Administrative:

- [x] I have read the [README](https://github.com/signalapp/Signal-iOS/blob/main/README.md).
- [x] I have read the [CONTRIBUTING](https://github.com/signalapp/Signal-iOS/blob/main/CONTRIBUTING.md) document and understand the caveats in the Pull Requests section.
- [x] I have signed the [Contributor Licence Agreement](https://signal.org/cla/).

Commits and testing:

- [x] My commits are rebased on the latest main branch.
- [x] My commits are well-structured for review.
- [x] My change has been thoroughly tested, and I am not aware of any regressions to existing features or behaviors.
 * iPhone 15 Pro, iOS 27.0 (Public Beta 3)

---

## Description

When Signal is open on an iPhone with Always-On Display enabled, locking the phone causes Signal's Face ID prompt to appear on the Lock Screen. After unlocking the phone, Signal asks for Face ID again.

This change prevents Signal from starting Screen Lock authentication while the device is locked. Once the device is unlocked, Signal checks the Screen Lock state again and presents authentication normally.

## Testing

- Tested with Screen Lock set to immediately and Always-On Display enabled. The Signal prompt no longer appeared while the phone was locked, and appeared once after unlocking.
- Repeated the test with Low Power Mode enabled.
- Verified that backgrounding and reopening Signal still requires authentication.
- Built the Signal scheme with code signing disabled.
- Ran SwiftFormat on the changed file.

### DIFF
--- a/Signal/util/ScreenLockUI.swift
+++ b/Signal/util/ScreenLockUI.swift
@@ -161,6 +161,12 @@ class ScreenLockUI {
         )
         NotificationCenter.default.addObserver(
             self,
+            selector: #selector(protectedDataDidBecomeAvailable),
+            name: UIApplication.protectedDataDidBecomeAvailableNotification,
+            object: nil,
+        )
+        NotificationCenter.default.addObserver(
+            self,
             selector: #selector(screenLockDidChange),
             name: ScreenLock.ScreenLockDidChange,
             object: nil,
@@ -292,6 +298,11 @@ class ScreenLockUI {
 
         guard !appIsInactiveOrBackground else {
             // Never show the auth UI unless active.
+            return
+        }
+
+        guard UIApplication.shared.isProtectedDataAvailable else {
+            // Never show the auth UI while the device is locked.
             return
         }
 
@@ -445,6 +456,11 @@ class ScreenLockUI {
     @objc
     private func applicationDidEnterBackground(_ notification: Notification) {
         appIsInBackground = true
+    }
+
+    @objc
+    private func protectedDataDidBecomeAvailable(_ notification: Notification) {
+        ensureUI()
     }
 }
 


sashaweiss-signal:
Hi, thanks for your interest in contributing. I'm not able to reproduce the issue you're describing; when I lock the device (with Always-On Display enabled) with Signal foregrounded and Signal Screen Lock enabled, I simply see the device lock. Pressing the power button to then wake the device causes it to use FaceID to unlock (as normal), and when I then swipe the Lock Screen away and Signal is again foregrounded it goes through the Signal Screen Lock (as normal).

Could I ask you for a screen recording of the issue you're experiencing, so I can make sure we're talking about the same things?

JustAlex0000:
Thanks for replying to me, let me try to clarify the issue a bit more.
I cannot really screen record since the issue involves locking the screen and it turns off my screen record.

The issue for me is when i turn off the phone with Signal opened it waits about 3 seconds and then the iphone displays the face id widget to authenticate (see the attached picture) even though the phone was turned off, and in that state i can either look at my phone so it unlocks, then it unlocks the screen and then it unlocks the Signal app again or i can decline the face id in the lock screen and the phone dims the screen normally.

Here are all potentially relevant settings for signal and iOS for reproducing this issue:
Signal:
Screen Lock Enabled
Screen Lock Timeout: Instant
iOS:
Always On Display Enabled
Raise to Wake Enabled
Auto-Lock 2 minutes
Low Power Mode Off (it turns off Always On display and that way "fixes" the issue)
Focus mode cannot be set to Sleep (it turns off Always On display and that way "fixes" the issue)

I tested with these exact settings on 2 phones:
iPhone 17 Pro with iOS 26.6 
iPhone 15 Pro with iOS 27 Public Beta 3
(latest Signal App Store version)
Both phones had the same issue (see photo) with these settings

The version with my code fixes the issue completly. 
I hope this clears up any confusion about my PR.

Screenshot shows my screen about 3 seconds after locking my phone with Signal opened and unlocked, happens 100% of the time after locking the phone.
image

5 seconds after locking the phone:
699EC54F-E9D1-4271-B074-6F456BE3AD12

PR #6335, opened 2026-08-28
Stop logging absolute filepaths and call root key

trufae:
## Contributor checklist

Administrative:

- [x] I have read the [README](https://github.com/signalapp/Signal-iOS/blob/main/README.md).
- [x] I have read the [CONTRIBUTING](https://github.com/signalapp/Signal-iOS/blob/main/CONTRIBUTING.md) document and understand the caveats in the Pull Requests section.
- [ ] I have signed the [Contributor Licence Agreement](https://signal.org/cla/).
- [x] I have reported this vulnerability privately and I was tired of waiting for a response

Commits and testing:

- [x] My commits are rebased on the latest main branch.
- [x] My commits are well-structured for review.
- [x] My change has been thoroughly tested, and I am not aware of any regressions to existing features or behaviors.
 * iPhone 15 pro iOS 26.6

## Description

The logging api calls have a regex to filter sensitive info like email addresses, phone numbers and other stuff that shouldn't be stored. ideally apple provides the proper private logging apis that redact all that using swift primitives.

Using a regex may somehow work but the current implementation is not covering all possible cases, and in this case there's a root key logged and it's bypassing the regex checks. And also absolute paths can be used as a reference for other exploits to pivot the app home directory if the attacker have access to the logs.

Ideally none of this information should be exposed in the logs, no matter if its usb, icloud or device backups. So my proposed solution is to catch those absolute paths in the regex, and just don't log the root key.

Feel free to pick those commits with tests to proof the regex cornercase or just adjust the patch the way it works better for you, no need for mentions/merits/cves from my side. I just want to raise aware of these issues and being solved upstream asap.

thanks


sashaweiss-signal:
Thanks for your interest in contributing. Per our SECURITY.md and CONTRIBUTING.md, security reports should be directed to security@signal.org rather than on Github. If you've already written in and haven't yet received a reply, please know we're looking into it!

Thanks again, and as always for being a Signal user.

